Issue #6027, opened 2022-07-12
Throw better error when failing to fetch entity on ometa


pmbrull:
For example, when we had some issues with JWT tokens:

```
  InsecureRequestWarning,
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for Table
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for Topic
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for Dashboard
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for Pipeline
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for MlModel
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for User
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for Team
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for GlossaryTerm
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for Glossary
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for Policy
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for TagCategory
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for MessagingService
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for DatabaseService
[2022-07-11 16:33:30] ERROR    {Ingestion:metadata:233} - Fetching entities failed for PipelineService
```

we should better show the stacktrace and understand the issue

Issue #5835, opened 2022-07-04
Provide meaningful errors on Airflow APIs


harshach:
**Affected module**
Ingestion, Airflow

**Describe the bug**
Currently, we are returning full stack traces of errors when deployment does't work.
We should capture these full stack traces and provide meaningful errors
Errors like these are not helpful for our end users
![image](https://user-images.githubusercontent.com/38649/177200213-ebc337fb-1bdc-4cbc-b3ed-3fb1c8f7673d.png)

**To Reproduce**

Screenshots or steps to reproduce

**Expected behavior**

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8]
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

pmbrull:
We need to:
- follow the flow of the code when the dag gets built
- validate the different points where we can get an exception
- make sure that explicit exceptions are being thrown
- capture those exceptions at the root and send comprehensive error messages back to the users

The logic is already different in main, but for example in `0.11.x` a couple of errors can happen:
1. we have an error retrieving the service, and yet we throw a "verify your config is correct" when the problem is in the communication between Airflow and OM
2. we have an error retrieving the token in the ometa_api. We still throw the "verify your config is correct" instead of explaining what is going on.

Double-check all these cases and return proper error messages

PR #6920, opened 2022-08-25
Fix#6027: Improve logging in OpenMetadata Airflow APIs

nahuelverdugo:
Part of: #6027 and #5835

### Describe your changes :

We are trying to improve and align logging by following these rules:

- Use `logger.debug(traceback.format_exc())` after catching an exception
- Log the exception/error as part of the logger message. Tried to follow the convention `{message}: {err/exc}`
- Use `warning` instead of `warn` and log it when the application encounters a problem that interrupts normal operation flow, but it has a fallback or can fix the problem itself or can continue
- Log an error if the `except` is at the end of the method or if we are rolling back

### Type of change :
- [x] Improvement

### Checklist:

### Reviewers

### DIFF
--- a/openmetadata-airflow-apis/openmetadata_managed_apis/api/error_handlers.py
+++ b/openmetadata-airflow-apis/openmetadata_managed_apis/api/error_handlers.py
@@ -12,23 +12,24 @@
 Register error handlers
 """
 
-import logging
-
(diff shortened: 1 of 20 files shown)


pmbrull (approved):
great changes, thank you both @nahuelverdugo @ulixius9

PR #6960, opened 2022-08-26
Added Kafka publisher

parthp2107:
### Describe your changes :

Added Kafka publisher

#
### Type of change :

- [x] Improvement

#
### Checklist:

#
### Reviewers
@mohitdeuex @harshach


catalog-rest-service/src/main/java/org/openmetadata/catalog/CatalogApplication.java
```diff
@@ -267,6 +270,15 @@ private void registerEventPublisher(CatalogApplicationConfig catalogApplicationC
     }
   }
 
+  public void registerKafkaPublisher(CatalogApplicationConfig catalogApplicationConfig) {
+    // register Kafka Event publisher
```
harshach, line 274: lets add UI way of controlling this instead of config. We should model similar to web hook.

catalog-rest-service/src/main/java/org/openmetadata/catalog/kafka/KafkaEventPublisher.java
```diff
@@ -0,0 +1,82 @@
...
+public class KafkaEventPublisher extends AbstractEventPublisher {
+
+  protected final KafkaEventConfiguration kafkaEventConfiguration;
+  private static KafkaProducer<String, String> producer;
+  Properties properties = new Properties();
+
+  public KafkaEventPublisher(KafkaEventConfiguration kafkaConfig) {
+    super(kafkaConfig.getBatchSize(), new ArrayList<>());
+    this.kafkaEventConfiguration = kafkaConfig;
+    if (kafkaConfig.getSecurityProtocol().equals(KafkaEventConfiguration.SecurityProtocol.SSL)) {
```
harshach, line 27: there are way too many combinations of Kafka security. SASL, SASL_SSL etc.. we should take key/value pairs as the config and pass them in as it is into kafkaConfig

catalog-rest-service/src/main/java/org/openmetadata/catalog/kafka/KafkaEventPublisher.java
```diff
@@ -0,0 +1,82 @@
...
+    producer.close();
+  }
+
+  @Override
+  public void publish(EventResource.ChangeEventList events) throws JsonProcessingException {
+    if (kafkaEventConfiguration.getTopics() != null) {
+      for (String topic : kafkaEventConfiguration.getTopics()) {
+        for (ChangeEvent event : events.getData()) {
+          String eventJson = JsonUtils.pojoToJson(event);
+          producer.send(new ProducerRecord<>(topic, eventJson));
```
harshach, line 75: this is in the path of returning a response back. Lets make sure this is async producer and we are not waiting for any ack

conf/openmetadata.yaml
```diff
@@ -165,6 +165,24 @@ elasticsearch:
   socketTimeoutSecs: ${ELASTICSEARCH_SOCKET_TIMEOUT_SECS:-60}
   batchSize: ${ELASTICSEARCH_BATCH_SIZE:-10}
 
+kafka:
+  topics: ${KAFKA_TOPICS:-[]}
```
harshach, line 169: there should only be one topic

harshach: we should take the minimum required configs, such as bootstrap servers and topic. Rest of them should be key/value pairs that users can chose what they want to pass

conf/openmetadata.yaml
```diff
@@ -165,6 +165,24 @@ elasticsearch:
...
+  bootstrapServers: ${KAFKA_BOOTSTRAP_SERVERS:-"http://localhost:9092"}
+  acks: ${KAFKA_ACKS:-"all"}
```
harshach, line 171: The defaults are already in place in Kafka producer client we don't need to mention them again

harshach (changes requested):

parthp2107:
@harshach, addressed all the requested changes along with comments.

catalog-rest-service/src/main/java/org/openmetadata/catalog/kafka/KafkaEventPublisher.java
parthp2107: The producer here is not waiting for any ack, it is directly publishing the message.

parthp2107:
Hey @harshach, can you please give your views on this pr? I have addressed all your comments.

Issue #6189, opened 2022-07-19
UI | Improve Search/Explore page - allow complex searching and offer more suggestions


mosiac1:
**Is your feature request related to a problem? Please describe.**
Users are reporting the search page is not good enough. The issues they run into are either: 
 - a slight difference in names cause a resource not to be displayed at all
 -  searching for a service/database/schema does not yield any relevant results (neither a link to the service/database/schema or tables/topic/etc. under it)
 - getting to a search only in a particular database/schema is still convoluted. Either navigate to it, start typing in search and click "Search in this database/schema" or try setting it the complex search.

**Describe the solution you'd like**
Some points I would like to tackle:
- [ ] The "Complex Search" section should always be below the search in `/explore`. Users have a hard time spotting the dropdown and using it.
- [ ] Users should be able to add multiple filters on the same resource (database/schema/service etc.). Filters for the same resource should be ORd together and ANDed with all other filters.
- [ ] Users should be able to freely enter any text in the complex search, not be limited to options.
- [ ] Users should receive suggestions for the complex search fields. These suggestions should be aware of the context - the value of other filters - so they offer relevant suggestions.
- [ ] Text search should also by default search in a table's database, schema, and service.
- [ ] Text search for tables should display as possible results Services, Databases, and Schemas along with the tables (in the main column). They should be clickable to go to the page of the resource or add that resource as a filter.

Another way of handling some of the issues here would be to add search bars on the left-hand side panel of `/explore` so users can search through resources.

**Additional context**

I would really appreciate feedback on these suggestions, maybe a design session to find the best way to tackle these issues.

mosiac1:
Inspiration:

Jira's search is based on a relatively simple search syntax and offers a simple search (UI-based, easy to use, intuitive, lots of suggestions, etc.) and an advanced search that is a text box, allowing any queries.

Creating a fully custom search syntax would be a very big feature - we should investigate initial implementations.

mosiac1:
Regarding the point 
> - [ ] Users should be able to add multiple filters on the same resource (database/schema/service etc.). Filters for the same resource should be ORd together and ANDed with all other filters.

Forcing users into this AND and OR behavior is not a good idea. It makes sense for some use cases and not for others. This is mainly what sparked the Jira query language discussion - allowing users the flexibility to mix and match filters.

mosiac1:
Yeah I will look into creating a POC this week

mosiac1:
Update:

[...]

PR #6965, opened 2022-08-26
UI (#6189) | PR 1 |  Extends Search api to support custom Elasticsearch filters

mosiac1:
### Describe your changes :
This is the first PR for the query builder feature tracked by #6189.

API changes:

###  - `/search` now accepts `query_filter` and `post_filter` parameters

These accept Elasticsearch queries in JSON format (e.g: Boolean query https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html). These are used to filter and post filter the results and aim to replace the functionality currently achieved through string queries (EQL),

`query_filter` is used as a filter on top of the string search query creator from the `q` parameter.  This the how most searches should filter their results. These filters affect the scoring of hits. (more info: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-filter-context.html)

`post_filter` is used as a final filter, its functionality is closely tied with Elasticseach-specific functionality (https://www.elastic.co/guide/en/elasticsearch/reference/current/filter-search-results.html#post-filter). This is slower than `query_filter`, but it does not affect scoring and __aggregations__. The main use case is the facet filter used on the Explore page. (this is not implemented in this PR).

### - UI calls to `/seach` and `/suggest` consolidated to 2 function in `searchAPI.ts`

Calls to the search API were spread out to various functions. `searchQuery` and `suggestQuery` are now the 2 points of contact with the API. These include functionality to adapt the results to a format well suited to the UI.

Added types to correctly represent the data returned. Cleaned up `Formatted*` types, now using generated types.

### - Added `fetch_source` and `include_fields` to `/search` and `/suggest`

With these parameters, callers can control if they need the body of objects and how much of that body they need. In many cases in the UI, there is no need to pull entire objects from Elasticsearch, this should decrease load times.

This is also fully supported by the type system in the UI.

#
### Type of change :

- [x] Improvement
- [x] New feature

#
query builder

#
### Checklist:

#
### Reviewers

Frontend: @open-metadata/ui


harshach:
Thanks @mosiac1 . We are a week away from 0.12.0 release. Will review and merge but will make it part of 0.12.1 as we need to do Advanced Search UX.
cc @chirag-madlani @devyani-kaushik  we need to look at the mocks for the advanced search design.

mosiac1:
I think this could go in 0.12.0 - it doesn't change functionality and its well tested

On the UX part, yeah we'll probably not get it done in one week.

harshach:
@mosiac1 mostly the UI part is what stopping getting this into the 0.12.0 release. If you want to separate this into backend and UI PRs we can get the search backend changes in for 0.12

harshach:
one help would be great from @mosiac1 is to add integration tests similar to what we are doing with mysql/postgres and om . It can be a separate PR all together but we are missing similar coverage that we have else where in search

mosiac1:
I've split the feature into multiple PRs. This one functionally is just backend changes, with small refactoring and typescript changes to the UI.

The next PR will have the bulk of the changes - this is just a setup.

I split it up because it was ~10k lines long, which is impossible to review effectively.

mosiac1:
I agree with the tests, they are very needed. I think they should go to another PR.

mosiac1:
@harshach @ShaileshParmar11 

Could we get some 👀 on this, please? I would really appreciate if we could get this moving since most new changes to the UI cause conflicts and I have to keep rebasing. Thank you!

harshach:
@mosiac1 can you do one last merge with latest main. We will merge it in

Sachin-chaurasiya (approved):

harshach:
@mosiac1 can you please help us why the advanced search that you have in the image is not available here.

mosiac1:
> @mosiac1 can you please help us why the advanced search that you have in the image is not available here.

I had to split the feature into a few PRs because of the size, i tried to outline it here: https://github.com/open-metadata/OpenMetadata/issues/6189#issuecomment-1228407492

This is the first one and doesn't have the query builder, it is a preparation step. As soon as we are done with this one I can open the next PR

mosiac1:
@harshach @Sachin-chaurasiya Rebased on the changes on main and tested locally. Would really appreciate a review.

mosiac1:
Closing PR, will reopen another one. Reasoning:  https://github.com/open-metadata/OpenMetadata/issues/6189#issuecomment-1271571158

PR #6970, opened 2022-08-26
Fix#6027: Fix logging in metadata package modules except ingestion and orm_profiler

nahuelverdugo:
Fixes: #6027

### Describe your changes :

We are trying to improve and align logging by following these rules:

- Use `logger.debug(traceback.format_exc())` after catching an exception
- Log the exception/error as part of the logger message. Tried to follow the convention `{message}: {err/exc}`
- Use `warning` instead of `warn` and log it when the application encounters a problem that interrupts normal operation flow, but it has a fallback or can fix the problem itself or can continue
- Log an error if the `except` is at the end of the method or if we are rolling back

### Type of change :
- [x] Improvement

### Checklist:

### Reviewers

### DIFF
--- a/ingestion/src/metadata/cmd.py
+++ b/ingestion/src/metadata/cmd.py
@@ -13,6 +13,7 @@ import logging
 import os
 import pathlib
 import sys
+import traceback
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from typing import List, Optional, Tuple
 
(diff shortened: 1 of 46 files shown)


ingestion/src/metadata/cmd.py
```diff
@@ -114,8 +115,10 @@ def profile(config: str) -> None:
     try:
         logger.debug(f"Using config: {workflow_config}")
         workflow = ProfilerWorkflow.create(workflow_config)
-    except ValidationError as e:
-        click.echo(e, err=True)
+    except ValidationError as err:
+        logger.debug(traceback.format_exc())
```
harshach, line 119: @nahuelverdugo even when I enable debug the traceback wasn't getting printed. Lot of times I needed to change the logger.debug to print to get the full trace. Can you check why that may be happening

harshach (approved):

ingestion/src/metadata/cmd.py
nahuelverdugo: I will check that before merging

nahuelverdugo: It is working fine. The only problem is that we call `set_loggers_level` in `logger.py` from different places. I will create a new issue for reviewing that.

Issue #6929, opened 2022-08-25
docker-compose services should wait until OpenMetadata DB is created


nahuelverdugo:
**Issue**
`mysql` service can take a while to create the OpenMetadata DB, which could end up in a race condition where dependent services start before the DB is created and fail.

**Solution**
- Improve `mysql` service health check to check that OpenMetadata DB has been created.
- Reorder dependency between services: `mysql <- om_server <- ingestion`

PR #6598, opened 2022-08-05
Added the docker compose for postgres

Vj-L:
Added the docker-compose file for Postgres

PR #7025, opened 2022-08-29
Fix#6929: docker-compose services should wait until OpenMetadata DB is created

nahuelverdugo:
Fixes: #6929 
Based on PR: #6598  

### Describe your changes :
- Added health checks in `docker-compose` files to the following services:
  - `mysql`
  - `postgresql`
  - `openmetadata-server`
- Removed unnecessary waits for services inside `sh` scripts used in `Dockerfile` files
- Aligned names for services and containers

### Type of change :
- [x] Bug fix
- [x] Improvement

### Checklist:

### Reviewers
@akash-jain-10 @Vj-L @ulixius9 @vivekratnavel @harshach


docker/local-metadata/docker-compose-dev-postgres.yml
```diff
@@ -11,18 +11,11 @@
 
 version: "3.9"
 services:
-  postgres:
-    build:
-      context: ../../.
-      dockerfile: docker/local-metadata/Dockerfile_postgres
-    container_name: openmetadata_postgres
```
pmbrull, line 18: Why are we not using our image? In there we should be having the latest postgres already but with the users ready

pmbrull (approved):
Great changes, thanks for the PR. Left a qq if you have the time. Thanks!

docker/local-metadata/docker-compose-dev-postgres.yml
nahuelverdugo: I have aligned this file with what we have in `docker-compose-dev.yml`. I can roll back this change and align it with `docker-compose-dev.yml`. What should I do?

pmbrull: LGTM then, thanks for raising

nahuelverdugo: We have agreed offline to leave this change.

docker/airflow/Dockerfile
```diff
@@ -4,10 +4,9 @@ RUN curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add -
 RUN curl https://packages.microsoft.com/config/debian/11/prod.list > /etc/apt/sources.list.d/mssql-release.list
 RUN apt-get update \
     && apt-get install -y \
-        gcc libsasl2-dev gnupg build-essential libssl-dev libffi-dev \
-        librdkafka-dev unixodbc-dev libxml2 libevent-dev --no-install-recommends  \
+        build-essential gcc gnupg libevent-dev libffi-dev libpq-dev librdkafka-dev libsasl2-dev libsasl2-modules  \
```
akash-jain-10, line 7: looks like we are missing on `openjdk-11-jre` that is required for deltalake connector! Let's add that in here!

docker/local-metadata/docker-compose-dev-postgres.yml
```diff
@@ -11,18 +11,11 @@
...
+  postgresql:
+    image: openmetadata/postgresql:latest
```
akash-jain-10, line 15: Let's instead build the postgresql using `build`, `context` and `dockerfile` for local development.

docker/metadata/openmetadata-start.sh
```diff
@@ -10,10 +10,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-MYSQL="${MYSQL_HOST:-mysql}":"${MYSQL_PORT:-3306}"
-while ! nc -z -w 5 "${MYSQL}";
-  do echo "Trying to connect to ${MYSQL}"; sleep 5;
-done
+echo "Starting OpenMetadata Server";
```
akash-jain-10, line 13: I will need to test out this change with OpenMetadata Helm Charts, creating a local docker image! Let me get back here testing post that!

akash-jain-10 (changes requested):

docker/local-metadata/docker-compose-dev-postgres.yml
nahuelverdugo: Should I update the same for `docker/local-metadata/docker-compose-dev.yml`?

docker/airflow/Dockerfile
nahuelverdugo: Should I add also the same dependencies that we have in `ingestion/Dockerfile`?

akash-jain-10: we can ignore `python3.9-dev`, `netcat` and `wget`

docker/local-metadata/docker-compose-dev-postgres.yml
akash-jain-10: Looks like it's already update there!

docker/metadata/openmetadata-start.sh
nahuelverdugo: Feel free to put it back if OpenMetadata Helm Charts test fails after this change.

akash-jain-10: Tested out this change! Looks good on Kubernetes as well! Let's roll in these changes! :)

akash-jain-10 (approved):

Issue #6816, opened 2022-08-19
Review auto query comment in Hive & Athena


pmbrull:
From https://openmetadata.slack.com/archives/C02B6955S4S/p1660852909848949

In all our connections we send a comment in the queries, something like

```
/* {\"app\": \"OpenMetadata\", \"version\": \"0.11.3.1\"} */ SELECT 42
```

And it is possible that these comments are giving issues.

We need to validate that the queries can properly run on Hive and, if not, we will need to update when we are adding them when creating the Engine

pmbrull:
Same thing is happening in Athena:

```
metadata.utils.connections.SourceConnectionException: Unknown error connecting with Engine(awsathena+rest://...:***@athena.us-east-2.amazonaws.com:443?s3_staging_dir=s3%3A%2F%2Fathena-postgres%2Foutput%2F&work_group=primary) - (pyathena.error.DatabaseError) An error occurred (UnrecognizedClientException) when calling the StartQueryExecution operation: The security token included in the request is invalid.
[SQL: /* {"app": "OpenMetadata", "version": "0.12.0.dev0"} */
SELECT 42]
(Background on this error at: https://sqlalche.me/e/14/4xp6).
```

pmbrull:
IMO we can add a new property to all JSON Schemas for databases named `supportsQueryComment` with default value to true. Then, if this value is True, we configure the Engine in SQLAlchemy to send the comment. If it is false, we disable that.

Let's validate that this works for Testing the connection as well

MilanBariya:
> metadata.utils.connections.SourceConnectionException: Unknown error connecting with Engine(awsathena+rest://...:***@athena.us-east-2.amazonaws.com:443?s3_staging_dir=s3%3A%2F%2Fathena-postgres%2Foutput%2F&work_group=primary) - (pyathena.error.DatabaseError) An error occurred (UnrecognizedClientException) when calling the StartQueryExecution operation: The security token included in the request is invalid.
> [SQL: /* {"app": "OpenMetadata", "version": "0.12.0.dev0"} */
> SELECT 42]
> (Background on this error at: https://sqlalche.me/e/14/4xp6).

Okay, @pmbrull let me try that way.

PR #7054, opened 2022-08-30
Fix: add a new property to all JSON Schemas for databases

MilanBariya:
### Describe your changes :

I worked on the add a new property to all JSON Schemas for databases named supportsQueryComment with default value to true.
Fix: #6816 

#
### Type of change :

- [x] Bug fix
- [x] Improvement

### Checklist:

#

### DIFF
--- a/ingestion/examples/workflows/athena.yaml
+++ b/ingestion/examples/workflows/athena.yaml
@@ -4,7 +4,6 @@ source:
   serviceConnection:
     config:
       type: Athena
-      supportsQueryComment: false
       database: database_name
       awsConfig:
         awsAccessKeyId: access key id
--- a/ingestion/examples/workflows/athena.yaml
+++ b/ingestion/examples/workflows/athena.yaml
@@ -4,6 +4,7 @@ source:
   serviceConnection:
     config:
       type: Athena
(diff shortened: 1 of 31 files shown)


ingestion/examples/workflows/athena.yaml
```diff
@@ -4,6 +4,7 @@ source:
   serviceConnection:
     config:
       type: Athena
+      supportsQueryComment: false
```
nahuelverdugo, line 7: I think we should only add this property to the connections which support query comments and then check in the python code like this:
```python
 if hasattr(self.config.source.serviceConnection.__root__.config, "supportsQueryComment")
```
at least is how we are checking other `support` properties for example with the [supportDatabase](https://github.com/open-metadata/OpenMetadata/blob/7863f040d1486d7920d0e4293631d973d66fd7b2/ingestion/src/metadata/orm_profiler/api/workflow.py#L357) one.

MilanBariya: Okay, sure@nahuelverdugo.

pmbrull:
closing in favour of https://github.com/open-metadata/OpenMetadata/pull/7079

PR #7106, opened 2022-09-01
Add Query Tag in Snowflake Profiler

ulixius9:
### Describe your changes :

Add Query Tag in Snowflake Profiler

#
### Type of change :

- [x] Improvement

#
### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/interfaces/sqa_interface.py
+++ b/ingestion/src/metadata/interfaces/sqa_interface.py
@@ -26,6 +26,9 @@ from sqlalchemy.engine.row import Row
 from sqlalchemy.orm import DeclarativeMeta, Session
 
 from metadata.generated.schema.entity.data.table import Table
+from metadata.generated.schema.entity.services.connections.database.snowflakeConnection import (
+    SnowflakeType,
+)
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/interfaces/sqa_interface.py
```diff
@@ -127,7 +131,27 @@ def _session_factory(service_connection_config):
         garbage collected once the application thread ends
         """
         engine = get_connection(service_connection_config)
-        return create_and_bind_thread_safe_session(engine)
+        session = create_and_bind_thread_safe_session(engine)
+        SQAInterface.set_session_tag(service_connection_config, engine)
+        return session
+
+    @staticmethod
+    def set_session_tag(service_connection_config, engine):
```
TeddyCr, line 139: We'll need to pass the `session` instance here (vs `engine`) and call that function l. 95 and l. 255

ulixius9: @TeddyCr, we are calling this function inside session factory, do we still need to do that then 🤔 ?

TeddyCr (approved):

PR #7269, opened 2022-09-06
Profiler close session after metric computation

TeddyCr:
### Describe your changes :

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


ingestion/src/metadata/interfaces/sqa_interface.py
```diff
@@ -275,6 +275,8 @@ def compute_metrics_in_thread(
         if column is not None:
             column = column.name
 
+        session.close()
```
harshach, line 278: do we need to wrap it around try/except here. What if session.close() throws a exception

TeddyCr: @harshach what kind of exception are you thinking of? We open the session just above so that instance will have a close() attribute.

pmbrull: maybe we can check if our sqlalchemy support something like

```python
with Session() as session:
    ...
```

and use use the internal context management

pmbrull: we're able to do that on the Airflow REST APIs on 2.3.3

TeddyCr: @pmbrull yes should be much cleaner. Just pushed a change.

nahuelverdugo (approved):

PR #7321, opened 2022-09-07
Fix #7239 Backend: Add validations for teamType patch

### LINKED ISSUE #7239: Backend: Add validations for teamType patch
While updating a team's type, add the following validations at the API level

- A team of type 'Group' cannot be updated 
- A team cannot be changed to 'Group' if it contains children

vivekratnavel:
### Describe your changes :
See #7239 

#
### Type of change :

- [x] Improvement

#
### Checklist:

#


nahuelverdugo (approved):

catalog-rest-service/src/main/java/org/openmetadata/catalog/exception/CatalogExceptionMessage.java
```diff
@@ -148,6 +148,14 @@ public static String unexpectedParent() {
     return "Team of type Organization can't have a parent team";
   }
 
+  public static String invalidGroupTeamUpdate() {
```
harshach, line 151: do we need a method call, lets make them static final strings no?

harshach (changes requested):
lets remove method calls

harshach (approved):

PR #7390, opened 2022-09-12
[BasicAuth] Added Basic Authentication 

mohitdeuex:
### Describe your changes :

#
### Type of change :

- [x] Improvement
- [x] New feature

#
### Checklist:

#
### Reviewers

#
### TODOs:

- [x] Refresh Token Logic (use refreshToken sent to generate new AccessToken)
- [x] Should remove old tokens while generating the new password Reset tokens
- [x] How should we remove the tokens , when the tokens are consumed , should be just go ahead and delete those, or we need to maintaining that information asking the admin to clear those details later 
- [x] Check Responses properly and , also scenarios in case of missing SMTP settings
- [x] Migrate Registration and changePassword to userResource from RegistrationResource
- [x] Update SQL Files with new table details and make user email as unique
- [ ] E2E testing


catalog-rest-service/pom.xml
```diff
@@ -454,6 +455,27 @@
       <artifactId>dropwizard-kafka</artifactId>
       <version>1.8.0</version>
     </dependency>
+    <dependency>
+      <groupId>at.favre.lib</groupId>
+      <artifactId>bcrypt</artifactId>
+      <version>0.9.0</version>
```
harshach, line 461: @mohitdeuex please move the versions to properties in main pom.xml

catalog-rest-service/src/main/java/org/openmetadata/catalog/email/EmailUtil.java
```diff
@@ -0,0 +1,296 @@
...
+      }
+    } else {
+      throw new RuntimeException("Email Request is missing Required Details");
+    }
+    return emailBuilder.buildEmail();
+  }
+
+  public void sendEmailVerification(String emailVerificationUrl, String to) throws TemplateException, IOException {
+    EmailPopulatingBuilder emailBuilder = EmailBuilder.startingBlank();
+    emailBuilder.withSubject("Email Verification[OpenMetadata]");
```
harshach, line 209: lets make that configurable

catalog-rest-service/src/main/java/org/openmetadata/catalog/email/EmailUtil.java
```diff
@@ -0,0 +1,296 @@
...
+    template.process(templatePopulator, stringWriter);
+    String mailContent = stringWriter.toString();
+    emailBuilder.withHTMLText(mailContent);
+    sendMail(emailBuilder.buildEmail());
+  }
+
+  public void sendAccountChangeEmail(String action, String actionStatus, String to)
+      throws IOException, TemplateException {
+    EmailPopulatingBuilder emailBuilder = EmailBuilder.startingBlank();
+    emailBuilder.withSubject("Account Status Change[OpenMetadata]");
```
harshach, line 254: lets make these configurable through a template

catalog-rest-service/src/main/java/org/openmetadata/catalog/resources/registration/RegistrationResource.java
```diff
@@ -0,0 +1,536 @@
...
+import org.openmetadata.catalog.util.ConfigurationHolder;
+import org.openmetadata.catalog.util.EntityUtil;
+import org.openmetadata.catalog.util.JsonUtils;
+import org.openmetadata.catalog.util.PasswordUtil;
+import org.openmetadata.catalog.util.RestUtil;
+import org.openmetadata.catalog.util.TokenUtil;
+import org.simplejavamail.api.email.Email;
+
+@Slf4j
+@Path("/v1/registration")
```
harshach, line 71: @mohitdeuex move it under /v1/user/signup

catalog-rest-service/src/main/java/org/openmetadata/catalog/security/DefaultAuthorizer.java
```diff
@@ -60,7 +65,15 @@ public void init(AuthorizerConfiguration config, Jdbi dbi) {
     PolicyCache.initialize();
     RoleCache.initialize();
     LOG.debug("Admin users: {}", adminUsers);
-    initializeUsers();
+    // For a Basic Auth , an admin can during server startup decide who all can be admin by adding admin principal
+    // when a registration is done through registration api, then we add the user as admin or normal user
+    // checking whether that username was bootstrapped into server during startup as admin
+    if (!ConfigurationHolder.getInstance()
```
harshach, line 71: @mohitdeuex any reason to change this. We can follow the same mechanism no?

mohitdeuex: @harshach the problem is , currently we create the admin accounts based on admin principals, but in basic auth we have "password" which I don't want to take in yaml, because if there are multiple adminPrinicpal we probable need to sequentially accepts all admin user password, so what this does is we still take adminPrincipals, but we don't create a account for him on server startup , when the user registers for the OM like any other user, we see that if the username is one of adminPrincipal and we create his account as admin , this ensure that the flow is intact and on similar line with other users, and we still figure out who the admins are
Here is that check :-

    if (ConfigurationHolder.getInstance()
        .getConfig(ConfigurationHolder.ConfigurationType.AUTHORIZERCONFIG, AuthorizerConfiguration.class)
        .getAdminPrincipals()
        .contains(userName)) {
      newUser.setIsAdmin(true);
    }

harshach (approved):

Issue #7407, opened 2022-09-13
Add source classpath as an optional param to load custom sources


pmbrull:
To load the source class we currently have (workflow.py)

```python
service_type: ServiceType = get_service_type_from_source_type(
    self.config.source.type
)

source_class = self.get(
    "metadata.ingestion.source.{}.{}.{}Source".format(
        service_type.name.lower(),
        self.typeClassFetch(source_type, True),
        self.typeClassFetch(source_type, False),
    )
)
```

We should allow users to load any custom class to be used as source. A possibility can be having an optional property on the source configuration `classpath`, so if `classpath` is informed, we will just run `source_class = self.get(self.config.source.classpath)`. Otherwise, we run the dynamic loading logic

harshach:
@nahuelverdugo I'll take care of the backend. Once the PR is up you can work on the ingestion side

pmbrull:
closed by https://github.com/open-metadata/OpenMetadata/pull/7419

PR #7419, opened 2022-09-13
Fix#7407: Add custom source connections for different service types

nahuelverdugo:
Part of: #7407

### Describe your changes :
These changes will allow ingesting data using custom sources. YAML example:

```yaml
source:
  type: custom_database
  serviceName: my_service
  serviceConnection:
    config:
      type: CustomDatabase
      sourcePythonClass: external.dependency.CustomSource
      config:
        param1: value1
  sourceConfig:
    config:
      type: DatabaseMetadata
sink:
  type: metadata-rest
  config: {}
workflowConfig:
  openMetadataServerConfig:
    hostPort: http://localhost:8585/api
    authProvider: no-auth
```

### Type of change :
- [x] New feature

### Checklist:

### Reviewers
@harshach


ingestion/src/metadata/ingestion/api/workflow.py
```diff
@@ -90,6 +83,16 @@ def __init__(self, config: OpenMetadataWorkflowConfig):
 
         self._retrieve_dbt_config_source_if_needed(metadata_config, service_type)
 
+        source_class = self.get(
+            self.config.source.serviceConnection.__root__.config.sourcePythonClass
+            if source_type.startswith("custom_")
```
pmbrull, line 88: do we need this check? Maybe if we just have the source path that should be enough 🤔

nahuelverdugo: We don't have the `sourcePythonClass` in all the connections, only in the custom ones. The non-custom ones must keep calculating the required python class.

pmbrull: oh you're right. Thanks

pmbrull (approved):
Thanks for preparing this, LGTM. We should also add an entry inside the Python SDK (?) to show how these would work

nahuelverdugo:
> Thanks for preparing this, LGTM. We should also add an entry inside the Python SDK (?) to show how these would work
I will create a ticket for that

nahuelverdugo:
@pmbrull created https://github.com/open-metadata/OpenMetadata/issues/7481 and https://github.com/open-metadata/OpenMetadata/issues/7480

pmbrull (approved):
Thanks for taking the time, looks great

PR #7421, opened 2022-09-13
postgres-sql-query-updated

### LINKED ISSUE #7387: Fix Postgres usage & lineage query
- In postgres usage we are filtering the same queries as in lineage. This won't work. Usage should not filter (or only filter creating users, etc)
- Revisit how we are getting the timestamp information, as looks like we are joining by database ID and losing the query start time

codingwithabhi:
### Describe your changes :

I worked on the updating sql queries for Postgres usage
fix #7387 
#
### Type of change :

- [x] Bug fix
- [x] Improvement

#
### Checklist:

#
### Reviewers
@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/postgres_query_parser.py
+++ b/ingestion/src/metadata/ingestion/source/database/postgres_query_parser.py
@@ -14,7 +14,7 @@ Postgres Query parser module
 import csv
 import traceback
 from abc import ABC
-from datetime import datetime, timedelta
+from datetime import datetime
 from typing import Iterable, Optional
 
 from sqlalchemy.engine.base import Engine
@@ -64,13 +64,11 @@ class PostgresQueryParserSource(QueryParserSource, ABC):
             )
         return cls(config, metadata_config)
 
-    def get_sql_statement(self, start_time: datetime, end_time: datetime) -> str:
+    def get_sql_statement(self) -> str:
         """
         returns sql statement to fetch query logs
         """
         return self.sql_stmt.format(
-            start_time=start_time,
-            end_time=end_time,
             result_limit=self.config.sourceConfig.config.resultLimit,
             filters=self.filters,
         )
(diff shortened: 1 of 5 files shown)


ingestion/src/metadata/ingestion/source/database/postgres_query_parser.py
```diff
@@ -146,7 +146,7 @@ def process_table_query(self) -> Optional[Iterable[TableQuery]]:
                     rows = conn.execute(
                         self.get_sql_statement(
                             start_time=self.start + timedelta(days=i),
-                            end_time=self.start + timedelta(days=i + 2),
+                            end_time=self.end + timedelta(days=i),
```
ulixius9, line 149: I believe this should be `self.start` only because on the beginning of the loop we are calculating the day difference and basically traversing from start time to end time putting `self.end` will violate the logic

ulixius9:
even though postgres now supports usage and linage option adding the same via UI is not available yet, can we add `supportsLineageExtraction` and `supportsUsageExtraction` in [postgresConnection.json](https://github.com/open-metadata/OpenMetadata/blob/main/catalog-rest-service/src/main/resources/json/schema/entity/services/connections/database/postgresConnection.json)

Screenshot 2022-09-13 at 9 32 00 PM

ingestion/src/metadata/ingestion/source/database/postgres_usage.py
```diff
@@ -26,9 +26,4 @@ class PostgresUsageSource(PostgresQueryParserSource, UsageSource):
     """
 
     sql_stmt = POSTGRES_SQL_STATEMENT
-    filters = """
-                AND (
-                     NOT s.query ILIKE 'create table %% as select %%'
-                    OR   NOT s.query ILIKE  'insert %%'
-                )
-            """
```
nahuelverdugo, line 34: What is the reason for deleting this?

codingwithabhi: these queries are required for Usage.
https://github.com/open-metadata/OpenMetadata/issues/7387#issue-1370040388

ingestion/src/metadata/utils/sql_queries.py
```diff
@@ -481,9 +481,9 @@
         JOIN pg_catalog.pg_stat_activity a ON d.datname = a.datname
```
pmbrull, line 481: Can we please double-check that the JOIN logic here is correct? Looking at the results with a community user it seemed that we were getting quite old queries listed as new. cc @ulixius9

pmbrull: Based on https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW, it seems that we are picking up `pg_stats_activity` information based on `datname`, which is the Name of the database. I believe we might need to instead JOIN here by `queryid` with any related table

ulixius9: Hey @pmbrull, got a chance to review this and found that the logic of join implementation here is incorrect and regarding joining `pg_stats_activity` with `pg_stat_statements` via `queryid` is not possible because `pg_stats_activity` does not store all the query logs otherwise we would used that table itself, I've made changes in the code to remove this invalid implementation, in my brief research I found that fetching execution start time for  `pg_stat_statements` is not possible.. so this also removes the filtering on the query based on start time

OnkarVO7 (approved):

Issue #7341, opened 2022-09-08
Improve DBT root config error handling


OnkarVO7:
**Is your feature request related to a problem? Please describe.**
If the `"dbtConfigSource":{}` in the yaml config is passed as empty, incorrect error is thrown

**Describe the solution you'd like**
Improve the error handling so that the error is clear as precise

PR #7446, opened 2022-09-14
Fix: DBT root config error handled

MilanBariya:
### Describe your changes :

I worked on the Fixing DBT root config error handling.
When `"dbtConfigSource":{}` that time it was giving tuple index of range error, and `get_dbt_details()` also execute that block even `"dbtConfigSource":{}` so that is handled . 
Fix: #7341

#
### Type of change :

- [x] Bug fix
- [x] Improvement

### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/database_service.py
+++ b/ingestion/src/metadata/ingestion/source/database/database_service.py
@@ -233,7 +233,7 @@ class DatabaseServiceSource(DBTMixin, TopologyRunnerMixin, Source, ABC):
     def __init__(self):
         if hasattr(self.source_config.dbtConfigSource, "dbtSecurityConfig"):
             if self.source_config.dbtConfigSource.dbtSecurityConfig is None:
-                logger.warning("dbtConfigSource is not configured")
+                logger.info("dbtConfigSource is not configured")
                 self.dbt_catalog = None
                 self.dbt_manifest = None
                 self.dbt_run_results = None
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/ingestion/source/database/database_service.py
```diff
@@ -231,11 +231,20 @@ class DatabaseServiceSource(DBTMixin, TopologyRunnerMixin, Source, ABC):
     context = create_source_context(topology)
 
     def __init__(self):
-        dbt_details = get_dbt_details(self.source_config.dbtConfigSource)
-        self.dbt_catalog = dbt_details[0] if dbt_details else None
-        self.dbt_manifest = dbt_details[1] if dbt_details else None
-        self.dbt_run_results = dbt_details[2] if dbt_details else None
-        self.data_models = {}
+        if hasattr(self.source_config.dbtConfigSource, "dbtSecurityConfig"):
```
OnkarVO7, line 234: @MilanBariya this condition will only be true for GCS, S3 configs. In case of http, local configs `dbtSecurityConfig` is not present. Can you double check this logic here?

OnkarVO7 (changes requested):

ingestion/src/metadata/ingestion/source/database/database_service.py
```diff
@@ -231,11 +231,20 @@ class DatabaseServiceSource(DBTMixin, TopologyRunnerMixin, Source, ABC):
...
+            if self.source_config.dbtConfigSource.dbtSecurityConfig is None:
+                logger.warning("dbtConfigSource is not configured")
```
pmbrull, line 236: nit - should this be a warning? It might be the usual case for many pipelines to not configure GCS/S3. IMO that could be just an INFO

OnkarVO7 (approved):

PR #7506, opened 2022-09-15
Fix Ingestion Time Partitioned Tables on Profiler

### LINKED ISSUE #5458: Add support for the profiler for ingestion time partition in BQ (`_PARTITIONTIME`/`_PARTITIONDATE`)
### Context
Profiler and data quality currently supports BQ partitioned table with explicit fields (i.e. fields that are present in the table), though BQ allows users to partition tables by ingestion time using pseudo column `_PARTITIONTIME` or `_PARTITIONDATE`.

### Solution
Implement support for pseudo column predicate in sampler's and runner's query

### A.C.
- Users are able to run profiler and tests against column with ingestion time partition

ayush-shah:
### Describe your changes :

fix #5458 

#
### Type of change :

- [x] Bug fix
- [ ] Documentation

#
</p>

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion


TeddyCr:
@ayush-shah it seems this was not tested locally as it is failing when trying to profile `local_bigquery.modified-leaf-330420.sdoh_cms_dual_eligible_enrollment.ingestion_partition_table`. Let me know if you need help with anything there.

ayush-shah:
Adding tests for the paritioned tables profiling workflow

ingestion/src/metadata/orm_profiler/api/workflow.py
```diff
@@ -211,34 +211,26 @@ def get_partition_details(self, entity: Table) -> Optional[TablePartitionConfig]
...
-                        "Unsupported ingestion based partition type. Skipping table"
-                    )
-
+                elif entity.tablePartition.intervalType == IntervalType.INGESTION_TIME:
+                    if entity.tablePartition.interval == "DAY":
+                        partition_field = "_PARTITIONDATE"
+                    else:
+                        partition_field = "_PARTITIONTIME"
                 return TablePartitionConfig(
                     partitionField=partition_field,
```
TeddyCr, line 228: What do you think of doing `if ... return` vs `if ... else`?

ingestion/src/metadata/orm_profiler/profiler/handle_partition.py
```diff
@@ -132,7 +132,8 @@ def build_partition_predicate(partition_details: dict, col):
             f"{partition_details['partition_field']} in "
             f"({format_partition_values(partition_details['partition_values'], col_type)})"
         )
-
+    if partition_details["partition_field"] == "_PARTITIONDATE":
+        col_type = "DATE"
```
TeddyCr, line 136: what if it's `_PARTITIONDATETIME`?

ingestion/src/metadata/orm_profiler/profiler/handle_partition.py
```diff
@@ -178,16 +179,15 @@ def handle_and_execute(_self, *args, **kwargs):
...
-                    .filter(partition_filter)
+                query_results = _self._build_query(*args, **kwargs).select_from(
+                    _self._sample if self.sampled else _self.table
                 )
-                if self.first:
-                    return query_results.first()
-                return query_results.all()
-
+                if partition_field in {"_PARTITIONDATE", "_PARTITIONTIME"}:
+                    if not self.sampled:
```
TeddyCr, line 186: Why do we have this here?

TeddyCr (changes requested):
Left a few comments

ingestion/src/metadata/orm_profiler/profiler/handle_partition.py
ayush-shah: `_PARTITIONTIME` sends a column name as partition_field

`_PARTITIONDATE` sends the column name as None, so need to add col_type as `Date`

ingestion/src/metadata/orm_profiler/profiler/handle_partition.py
ayush-shah: so {"_PARTITIONDATE", "_PARTITIONTIME"} and the other fields like column names all differ when the queries are being formed..

if `self.sampled` is `True`, the query it forms has couple of syntax issue where `Where` clause repeats and gives us a syntax error, that's why added a check that if `sampled` is `True`, we don't add a `where` as there already is one present.


Paritioned and sampled throws syntax error..that's why the conditions

ingestion/src/metadata/orm_profiler/api/workflow.py
ayush-shah: I think as the partition_field is the one being changed based on conditions, i had to keep if…elif and return based on the updated value of partition_field
if the partition_field by any chance, comes as none, it will show up as
`f"Unsupported partition type {entity.tablePartition.intervalType}. Skipping table"`

TeddyCr: You could do something like this
```python
    if entity.tablePartition.intervalType == IntervalType.TIME_UNIT:
                    return TablePartitionConfig(partitionField=entity.tablePartition.columns[0])
                if entity.tablePartition.intervalType == IntervalType.INGESTION_TIME:
                    if entity.tablePartition.interval == "DAY":
                        return TablePartitionConfig(partitionField="_PARTITIONDATE")
                    return TablePartitionConfig(partitionField="_PARTITIONTIME")
```

ingestion/src/metadata/orm_profiler/profiler/handle_partition.py
TeddyCr: 🤔 are you sure? As you can see below we are passing the column by getting it from the list of columns on the SQA Declarative Table. Since `_PARTITIONDATE`, `_PARTITIONTIME` I believe they both are not in the columns of the declarative table.
https://github.com/open-metadata/OpenMetadata/blob/c1e6e947dc08428d1ece6777e7f58ab1be75ac45/ingestion/src/metadata/orm_profiler/profiler/handle_partition.py#L171

ingestion/src/metadata/orm_profiler/profiler/handle_partition.py
TeddyCr: > if self.sampled is True, the query it forms has couple of syntax issue where Where clause repeats and gives us a syntax error, that's why added a check that if sampled is True, we don't add a where as there already is one present.

^ That probably means there is something to fix somewhere else.

ingestion/src/metadata/orm_profiler/profiler/handle_partition.py
ayush-shah: If it's `_PARTITIONTIME` for partition interval set as `DATE`
we get: 
```
col = {Column} asd.e
col_type = {DATE} DATE
partition_details = {dict: 4} {'partition_field': 'e', 'partition_start': datetime.datetime(2022, 8, 23, 0, 0), 'partition_end': datetime.datetime(2022, 9, 23, 0, 0), 'partition_values': None}
```


but if it's `_PARTITIONDATE`
we get:
```
col = {NoneType} None
col_type = {NoneType} None
partition_details = {dict: 4} {'partition_field': '_PARTITIONDATE', 'partition_start': datetime.datetime(2022, 8, 23, 0, 0), 'partition_end': datetime.datetime(2022, 9, 23, 0, 0), 'partition_values': None}
```

ayush-shah: so if it's `_PARTITIONDATE `, I set the col_type as `Date` so that the predicate it creates is just for `Date` and not `DateTime`

ingestion/src/metadata/orm_profiler/api/workflow.py
ayush-shah: ohh okay, got it.. will make that change

ingestion/src/metadata/orm_profiler/profiler/handle_partition.py
ayush-shah, line 136: _PARTITIONDATE col_type is sent as None
so in order to have that available in `YYYY-MM-DD` format and not `YYYY-MM-DD HH-MM-SS`, had to add col_type as DATE

TeddyCr (approved):

PR #7511, opened 2022-09-16
Fixes #7504 Added migration steps to delete profiler pipeline from entity relationship

### LINKED ISSUE #7504: Bootstrap migrate - Clean IngestionPipeline relationships
In v004 we are deleting profiler pipelines from the entity table and the extension, but not the relationship.

Then, when trying to load a service, we try to look for the relationship by ID and we cannot find the pipeline (as it was deleted).

We need to update the migration to properly remove the entity_relationship for the IDs of the pipelines that we delete.

TeddyCr:
### Describe your changes :
Fixes #7504 by cleaning entity_relationship
#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


nahuelverdugo (approved):
Could be any issue if someone moves from `0.12.0` to `0.12.1`
?

TeddyCr:
> Could be any issue if someone moves from `0.12.0` to `0.12.1`?

Could be yes. The problem is that we need the id of the profiler pipeline to delete them from the `entity_relationship`. If you have done the 0.11.5 -> 0.12.0 upgrade you delete those pipelines and can't delete them with the query anymore.  You'll need to go 0.11.5 -> 0.12.0
--
cc @open-metadata/backend  for feedback here

vivekratnavel:
> 

One way to tackle this problem would be to scan the entity_relationship table for "pipeline" in "fromEntity" or "toEntity" and then check if the id exists in `ingestion_pipeline_entity`. If the id does not exist, then we should delete those records. I will try to come with SQL to help you out if you like this approach

TeddyCr:
> > 
> 
> 
> 
> One way to tackle this problem would be to scan the entity_relationship table for "pipeline" in "fromEntity" or "toEntity" and then check if the id exists in `ingestion_pipeline_entity`. If the id does not exist, then we should delete those records. I will try to come with SQL to help you out if you like this approach 

That makes sense. Delete the ones that do not exist in `ingestion_pipeline_entity`

vivekratnavel:
> 



> > > 
> > 
> > 
> > One way to tackle this problem would be to scan the entity_relationship table for "pipeline" in "fromEntity" or "toEntity" and then check if the id exists in `ingestion_pipeline_entity`. If the id does not exist, then we should delete those records. I will try to come with SQL to help you out if you like this approach
> 
> @vivekratnavel i might have spoken too fast. Actually it creates relationship with users as well (we can't just do a lookup this way)

Any relationship to the pipeline stored in the `entity_relationship` table will be handled by the approach I suggested. Can you give me an example of how a user relationship is tracked?

TeddyCr:
> > > One way to tackle this problem would be to scan the entity_relationship table for "pipeline" in "fromEntity" or "toEntity" and then check if the id exists in `ingestion_pipeline_entity`. If the id does not exist, then we should delete those records. I will try to come with SQL to help you out if you like this approach
> > 
> > 
> > @vivekratnavel i might have spoken too fast. Actually it creates relationship with users as well (we can't just do a lookup this way)
> 
> Any relationship to the pipeline stored in the `entity_relationship` table will be handled by the approach I suggested. Can you give me an example of how a user relationship is tracked?

If you create a profiler pipeline, you'll see in the `entity_relationship` table that the a user entity has a relationship to the profiler pipeline.

harshach:
ingestion pipeline has a owner. IMO deleting through the APIs are safer or using java code than using sql statements.
I've spoke to @nahuelverdugo to come up with a framework to handle the data deletions through java code rather than sql

TeddyCr:
> ingestion pipeline has a owner. IMO deleting through the APIs are safer or using java code than using sql statements.
> I've spoke to @nahuelverdugo to come up with a framework to handle the data deletions through java code rather than sql

@harshach  Should we still go through with this PR though?

harshach:
Yes lets merge it in. Java based migration will take longer time and we need to handle this case. Deleting ingestion pipelines are good enough , we don't need to delete the user based relation for now. We can figure that out later

vivekratnavel:
> > > > One way to tackle this problem would be to scan the entity_relationship table for "pipeline" in "fromEntity" or "toEntity" and then check if the id exists in `ingestion_pipeline_entity`. If the id does not exist, then we should delete those records. I will try to come with SQL to help you out if you like this approach
> > > 
> > > 
> > > @vivekratnavel i might have spoken too fast. Actually it creates relationship with users as well (we can't just do a lookup this way)
> > 
> > 
> > Any relationship to the pipeline stored in the `entity_relationship` table will be handled by the approach I suggested. Can you give me an example of how a user relationship is tracked?
> 
> If you create a profiler pipeline, you'll see in the `entity_relationship` table that the a user entity has a relationship to the profiler pipeline.

If the user relation is tracked by the `entity_relationship` table, then it will be covered by the same approach I suggested since we will look for "pipeline" in both "fromEntity" and "toEntity". In this case, the owner will be tracked as 

| fromEntity      | toEntity | relation |
| ----------- | ----------- | ---------- |
| user      | pipeline       | owns

The migration sql to cover all the cases would be

```
DELETE FROM entity_relationship WHERE fromEntity='pipeline' AND fromId NOT IN (SELECT id FROM ingestion_pipeline_entity);

DELETE FROM entity_relationship WHERE toEntity='pipeline' AND toId NOT IN (SELECT id FROM ingestion_pipeline_entity);
```

harshach:
@vivekratnavel @TeddyCr lets not over complicate this. Look up ingestion pipeline and see if it exists if not delete it. we can handle the user relation later.

vivekratnavel (approved):

Issue #7272, opened 2022-09-06
Exception Management in the Topology


pmbrull:
We usually don't want to stop the workflow from running when encountering individual errors.

One possible approach would be to manage exceptions directly on the Topology.

Each source should throw two types of exceptions (please improve the names):
- RuntimeException
- WorkflowException

If the topology finds a runtime exception, it will continue. If it finds a workflow exception, it will stop everything.

This way we would be certain that sources run e2e and we have a centralized place to handle how to respond to each case

----
The primary purpose of this task must improve the user experience when displaying errors. After analyzing the current status, we notice that:
- How we handle parsing errors of the workflow configuration could be improved.
  - Pydantic errors are not clear. There is no guidance, or information, for the user when it fails due to missing/extra parameters. The same occurs with malformed YAML files.
- Workflow can have different behaviors depending if we are running them from CLI or scheduler. For example, how we print the status or set the logging level.
  - We need to abstract the workflow class
- Each topology node is divided into three main parts: producer, stages, and post-processing:
  - Producer: any errors thrown here must stop the topology execution, so we have to throw a `WorkflowException`.
  - Stages/Post-processing: any error here doesn't affect the topology execution since the main purpose is the extraction of different entities to be ingested/processed. In this case, we throw an `EntityIngestionException`
    - These exceptions must be handled and logged into a single place instead of being done multiple times in different parts of the code

We also need to take into account how the `context` reacts to an issue in the processing. For example, in the `pipeline_service`, if we have a Pipeline entity that we cannot process, we would be moving on to further stages without the right context being set. We need to make sure that on specific errors, we clean up the stage context and further nodes validate the context before moving on.

The list of tasks to be done are:
- [x] Improve errors displayed in CLI when workflow parsing fails https://github.com/open-metadata/OpenMetadata/pull/7522
- [ ] Abstract workflow class
- [ ] Handle exceptions from the topology runner
- [ ] Review exceptions for each supported service:
  - [ ] Database
  - [ ] Dashboard 
  - [ ] Pipeline
  - [ ] MlModel
  - [ ] Messaging
- [ ] Review exceptions for profiler workflow
- [ ] Review exceptions for test suite workflow
- [ ] Review how we are handling the creation of `pydantic` objects

harshach:
@pmbrull @nahuelverdugo lets begin the design/thinking process on how do we want to handle this but actual delivery should be in 0.13

pmbrull:
[...]

PR #7522, opened 2022-09-16
Fix#7272: Improve logging when initializing workflow from CLI

nahuelverdugo:
Part of: #7272
### Describe your changes :
- Set lower logging level to CLI commands
- Improve messages displayed to the user
- Remove click dependency from Workflows
- Moved workflow examples inside `src` folder

### Type of change :
- [x] Improvement

### CLI Preview (Screenshots) :
image

### Checklist:

### Reviewers


ingestion/src/metadata/cmd.py
```diff
@@ -110,20 +113,20 @@ def test(config: str) -> None:
 def profile(config: str) -> None:
     """Main command for profiling Table sources into Metadata"""
     config_file = pathlib.Path(config)
-    workflow_config = load_config_file(config_file)
-
+    workflow_config_dict = None
```
TeddyCr, line 116: Do we need it to assign it `None` or could just go for the assignment in `try:` directly?

nahuelverdugo: I am assigning `None` since `workflow_config_dict` is used in the exception catch, and the method `load_config_file` could throw an exception.

```python
workflow_config_dict = None
try:
    workflow_config_dict = load_config_file(config_file)
    logger.debug(f"Using config: {workflow_config_dict}")
    workflow = ProfilerWorkflow.create(workflow_config_dict)
except Exception as exc:
    print_init_error(exc, workflow_config_dict, WorkflowType.profile) # <-- here
    sys.exit(1)
```

TeddyCr: Make sense. Apologies for missing that 😅

TeddyCr (approved):

PR #7523, opened 2022-09-16
Fixes #7521 using package name lowercase

### LINKED ISSUE #7521: camelCase package names cause compilation failure
**Affected module**
backend

**Describe the bug**
This is probably not technically a bug but when compiling with Azul zulu 11 openjdk, the camelCase packages get compiled to all lower case, which fails the class compilation of dependent classes.

for example `package org.openmetadata.schema.services.connections.mlModel;` gets compiled to `org.openmetadata.schema.services.connections.mlmodel`

**To Reproduce**

Run `mvn install` on the parent project

result:
image

**Expected behavior**
package conventions are all lower case and this seems to be a requirement of some JVMs.  I'm using zulu 11 because I'm on an M1 mac

**Version:**
 - OS: MacOS 12.6
 - Python version: 3.9
 - OpenMetadata version: 0.12
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-spec[docker]==XYZ`]

**Additional context**

dr3s:
### Describe your changes :
Fixes #7521 using package name lowercase.   This became more complicated than I thought.  It's possible that the jsonschema change and `ServiceType` enum change are breaking.  I don't know enough about the project to be sure but data could be incompatible if it's case sensitive.  Client SDK could fail compilation if it uses the enum.

#
### Type of change :
- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

#

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
     rev: 22.3.0
     hooks:
       - id: black
-        language_version: python3.9
+        language_version: python3.8
         exclude: ingestion/src/metadata/generated
   - repo: https://github.com/timothycrosley/isort
     rev: 5.10.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
     rev: 22.3.0
     hooks:
       - id: black
-        language_version: python3.8
+        language_version: python3.9
         exclude: ingestion/src/metadata/generated
   - repo: https://github.com/timothycrosley/isort
     rev: 5.10.1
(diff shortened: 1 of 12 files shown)


harshach (approved):

dr3s:
Looks like this is going to be more complicated because there is some SPI using string building to find the package.

dr3s:
Looks like this is going to be more complicated because there is some SPI using string building to find the package.

dr3s:
If I was starting over, I would probably decouple the package name resolution from the `ServiceType` and other aspects so that SPI isn't impacted by the jsonschema.

dr3s:
I don't think this is going to work.  There are too many dependencies on the naming convention.

harshach:
@dr3s this should be straightforward, Let me take a look and get back to you

harshach:
@dr3s we pushed a fix, please try now. Closing this.

PR #7525, opened 2022-09-17
Fix #7524 Backend: Allow users to rename the Glossary Term

### LINKED ISSUE #7524: Backend: Allow users to rename the Glossary name and Glossary term names

vivekratnavel:
### Describe your changes :
Allow patching "name" field of Glossary Term

#
### Type of change :

- [x] Improvement

#
### Checklist:

#


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -225,6 +228,35 @@ protected void updateTags(String fqn, String fieldName, List<TagLabel> origTags,
...
+    private void updateChildName(EntityReference child, String parentFqn) throws IOException {
+      GlossaryTerm childTerm = get(null, child.getId(), new Fields(List.of("parent", "children")));
+      String oldFqn = childTerm.getFullyQualifiedName();
+      EntityReference parent = childTerm.getParent();
+      parent.setFullyQualifiedName(parentFqn);
+      setFullyQualifiedName(childTerm);
+      String newFqn = childTerm.getFullyQualifiedName();
+      dao.update(childTerm.getId(), JsonUtils.pojoToJson(childTerm));
+      // update all the tag usages
+      daoCollection.tagUsageDAO().updateTagFqn(oldFqn, newFqn);
```
harshach, line 240: @vivekratnavel any reason to update tags?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -225,6 +228,35 @@ protected void updateTags(String fqn, String fieldName, List<TagLabel> origTags,
...
+    }
+
+    private void updateName(GlossaryTerm origTerm, GlossaryTerm updatedTerm) throws IOException {
+      if (updatedTerm != null && !updatedTerm.getName().equals(origTerm.getName())) {
+        recordChange("name", origTerm.getName(), updatedTerm.getName());
+        List<EntityReference> children = getChildren(updatedTerm);
+        for (EntityReference child : children) {
+          updateChildName(child, updatedTerm.getFullyQualifiedName());
+        }
+        daoCollection.tagUsageDAO().updateTagFqn(origTerm.getFullyQualifiedName(), updatedTerm.getFullyQualifiedName());
```
harshach, line 256: This should generate a change event

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
vivekratnavel: When the name of a GlossaryTerm gets updated, its FQN gets updated. If we don't update tagUsage, then all the assets that were tagged with this GlossaryTerm will no longer be tagged and the records in tagUsage table become invalid and dangling pointers.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
vivekratnavel: If a root glossaryTerm has lots of children and grand children terms, updating its name will then create a change event and activity feed for all the assets that are tagged with any term in the lineage. Do we want to create change events recursively for all the affected entities while updating a glossary term?

harshach: no just the change event for that particular term where the name is being changed.

vivekratnavel: Got it. This patch will do exactly that.

harshach:
@sureshms if we do not change the name of the glossary term, it will be weird in sharing the links to the glossary term and as well as FQN of the term. There will be unwanted issues on how we are looking up in elastic search like based on fqn vs display name. My concern was , lets say there is Business Glossary.Sales.Order -> Business Glossary.Product.order there is big difference in identifying the term if we only change the display name.
We are not shipping this in 0.12.1, so we have time to discuss

Issue #7573, opened 2022-09-20
Ingest team as owner from DBT


OnkarVO7:
**Is your feature request related to a problem? Please describe.**
The owner information from the DBT catalog.json file can be a team as well. 
Along with user, teams should also be ingested from DBT

Add documentation around it.

PR #7576, opened 2022-09-20
Added ingesting teams from DBT

OnkarVO7:
### Describe your changes :

Fix https://github.com/open-metadata/OpenMetadata/issues/7573

#
### Type of change :

- [x] Improvement
- [x] New feature

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/dbt_source.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt_source.py
@@ -27,6 +27,7 @@ from metadata.generated.schema.entity.data.table import (
     ModelType,
     Table,
 )
+from metadata.generated.schema.entity.teams.team import Team
 from metadata.generated.schema.entity.teams.user import User
 from metadata.generated.schema.tests.basic import (
     TestCaseResult,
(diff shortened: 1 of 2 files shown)


ingestion/src/metadata/ingestion/source/database/dbt_source.py
```diff
@@ -96,14 +97,25 @@ def _parse_data_model(self):
...
-                        entity=User, fqn=user_fqn
-                    )
+                    dbt_owner = cnode["metadata"].get("owner")
+                    owner = None
+                    if dbt_owner:
+                        owner_name = f"*{dbt_owner}*"
+                        user_owner_fqn = fqn.build(
+                            self.metadata, entity_type=User, user_name=owner_name
+                        )
+                        if user_owner_fqn:
```
harshach, line 107: @OnkarVO7 how did this will go into else state?

ingestion/src/metadata/ingestion/source/database/dbt_source.py
```diff
@@ -96,14 +97,25 @@ def _parse_data_model(self):
...
+                                entity=User, fqn=user_owner_fqn
+                            )
+                        else:
+                            team_owner_fqn = fqn.build(
+                                self.metadata, entity_type=Team, team_name=owner_name
+                            )
+                            if team_owner_fqn:
+                                owner = self.metadata.get_entity_reference(
+                                    entity=Team, fqn=team_owner_fqn
+                                )
```
ayush-shah, line 118: Let's give them an else warning the specified user doesn't exists in OpenMetadata..
if user, else if team, else doesn't exist warning..

ingestion/src/metadata/ingestion/source/database/dbt_source.py
ayush-shah: fqn will be created everytime, let's tackle this a little better where if we don't get any entity reference, we fallback to team

OnkarVO7: @harshach so the logic is if we get the user then user will be ingested else the team will be ingested.

`fqn.build` method uses ES to search for the user. If the user is not found it will go into else

ingestion/src/metadata/ingestion/source/database/dbt_source.py
OnkarVO7: Got it I'll add the warning @ayush-shah

ingestion/src/metadata/ingestion/source/database/dbt_source.py
```diff
@@ -96,14 +97,25 @@ def _parse_data_model(self):
...
-                    user_fqn = fqn.build(
-                        self.metadata, entity_type=User, user_name=user_name
-                    )
-                    owner = self.metadata.get_entity_reference(
-                        entity=User, fqn=user_fqn
-                    )
+                    dbt_owner = cnode["metadata"].get("owner")
+                    owner = None
+                    if dbt_owner:
+                        owner_name = f"*{dbt_owner}*"
```
ayush-shah, line 103: @OnkarVO7 what will be the scenario when the owner name is `John`

and we have 3 users named 
`John`, `JohnDoe`,`JaneDoe`

will it take the first one?
if so, do we need quotes around this? like a wildcard search? or just an exact match should be fine?

ingestion/src/metadata/ingestion/source/database/dbt_source.py
harshach: why use es to look up user, lets use our apis
/api/v1/users/{name} should get you the user

OnkarVO7: @harshach the reason behind using ES was to search the users a bit better. It might happen that the user name in the DBT catalog.json file might not be the exact same as the user name in OM

ingestion/src/metadata/ingestion/source/database/dbt_source.py
OnkarVO7: @ayush-shah added the warning

ayush-shah (approved):

Issue #7362, opened 2022-09-11
Dagster integration must work with its API and enable dagster cloud 


harshach:
**Is your feature request related to a problem? Please describe.**
Currently, Dagster integration only works with MySQL and its bypassing the API from dagster 
This requires us to build an integration to all the DBs that Dagster supports and it will be harder to configure and maintain.
Let’s use the API and only enable data that comes from those APIs

NiharDoshi99:
Hi , I will work on this task.

PR #7579, opened 2022-09-20
Fixed: Dagster to work with its API

NiharDoshi99:
### Describe your changes :

I worked on the Dagster integration with its API. 
Issue #7362 

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation


ingestion/src/metadata/ingestion/source/pipeline/dagster.py
```diff
@@ -100,62 +100,106 @@ def session(self) -> Session:
...
+        host_port = self.service_connection.hostPort
+        char_to_replace = {"https://": "", "http://": ""}
+        for key, value in char_to_replace.items():
+            host_port = host_port.replace(key, value)
+
+        host, port = host_port.split(":")
+        try:
+            client = DagsterGraphQLClient(hostname=host, port_number=int(port))
+            result = client._execute(
+                """
```
ulixius9, line 112: Lets define this query as a constant

ingestion/src/metadata/ingestion/source/pipeline/dagster.py
```diff
@@ -100,62 +100,106 @@ def session(self) -> Session:
         return self._session
 
     def get_run_list(self):
+        host_port = self.service_connection.hostPort
+        char_to_replace = {"https://": "", "http://": ""}
+        for key, value in char_to_replace.items():
+            host_port = host_port.replace(key, value)
```
ulixius9, line 106: We can write this loop as one line code like 
```
host_port = host_port.replace("https://","").replace("http://","")
```

ingestion/src/metadata/ingestion/source/pipeline/dagster.py
```diff
@@ -100,62 +100,106 @@ def session(self) -> Session:
...
+                    }
+                
+                
+                    }
+                }
+            }
+        """
+            )
+        except ConnectionError:
+            return False
```
ulixius9, line 149: This exception handling can be improved because even though you are returning False it will eventually break on like `220` where you are calling this function and running loop on returned value. let's add some error and debug logs here and break the execution in case of connection error

ingestion/tests/unit/topology/pipeline/test_dagster.py
```diff
@@ -51,9 +51,9 @@
                 "type": "Dagster",
                 "hostPort": "http://localhost:3000",
                 "dbConnection": {
-                    "username": "dagster_user",
-                    "password": "dagter_pass",
-                    "databaseSchema": "dagster_db",
+                    "username": "user",
+                    "password": "password",
+                    "databaseSchema": "database_db",
```
ulixius9, line 56: is `dbConnection` field still required? can we remove this from json schema itself?

ulixius9:
Also we might need to modify the [test_connection](https://github.com/open-metadata/OpenMetadata/blob/5c0016d827ed71c5fa387fdcb226a6b03af3c0ab/ingestion/src/metadata/utils/connections.py#L863) for dagster as well as we will not be using db connection anymore

ingestion/src/metadata/ingestion/source/pipeline/dagster.py
```diff
@@ -89,73 +85,66 @@ def create(cls, config_dict, metadata_config: OpenMetadataConnection):
...
-        Return the SQLAlchemy session from the engine
-        """
-        if not self._session:
-            self._session = create_and_bind_session(self.engine)
-
-        return self._session
-
     def get_run_list(self):
+        try:
+            # client = DagsterGraphQLClient(hostname=host, port_number=int(port))
```
ulixius9, line 90: can we remove this comment

ingestion/src/metadata/utils/sql_queries.py
```diff
@@ -536,3 +536,55 @@
   from SVV_TABLE_INFO
   where diststyle not like 'AUTO%%'
 """
+
+DAGSTER_PIPELINE_DETAILS_GRAPHQL = """
```
ulixius9, line 540: Technically these are not sql queries so it should be in these file, maybe you can create a file like `graphql_queries.py` and move it there

NiharDoshi99: Ok 👍

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/pipeline/dagsterConnection.json
```diff
@@ -31,26 +31,11 @@
       "type": "integer",
       "default": "10"
     },
-    "dbConnection": {
```
ulixius9, line 34: we will need to update migration script for this change

ulixius9 (approved):

Issue #7442, opened 2022-09-14
Add metadata restore CLI


pmbrull:
to restore mysql and postgres data after the backup

pmbrull:
Also @MilanBariya let's update the docs based on this once we get it merged

MilanBariya:
> Also @MilanBariya let's update the docs based on this once we get it merged

Sure @pmbrull I will do that.

PR #7596, opened 2022-09-20
Add: Add metadata restore CLI

MilanBariya:
### Describe your changes :

I worked on the adding metadata restore CLI
Fix: #7442

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature

### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/cli/db_dump.py
+++ b/ingestion/src/metadata/cli/db_dump.py
@@ -11,7 +11,10 @@ TABLES_DUMP_ALL = {
     "entity_extension",
     "field_relationship",
     "tag_usage",
+    "openmetadata_settings",
 }
+
+EXCLUDED_TABLES = {"entity_extension_time_series": {"exclude_columns": ["timestamp"]}}
 NOT_MIGRATE = {"DATABASE_CHANGE_LOG"}
 
 STATEMENT_JSON = "SELECT json FROM {table}"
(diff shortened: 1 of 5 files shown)


ingestion/src/metadata/cli/db_dump.py
```diff
@@ -64,6 +67,40 @@ def dump_all(tables: List[str], engine: Engine, output: Path) -> None:
                 file.write(insert)
 
 
+def dump_entity_custom(engine: Engine, output: Path, inspector) -> None:
+    """
+    This function is used to dump entities with custom handling
+    """
+    with open(output, "a") as file:
+        for table, data in EXCLUDED_TABLES.items():
```
pmbrull, line 75: Instead of EXCLUDED_TABLES might we name it `custom_tables` or something? to understand a bit better how this is being used

ingestion/src/metadata/cli/restore.py
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    if not schema:
+        connection_dict["databaseSchema"] = database
+        connection = MysqlConnection(**connection_dict)
+    else:
+        connection_dict["database"] = database
+        connection = PostgresConnection(**connection_dict)
+
+    engine: Engine = get_connection(connection)
```
pmbrull, line 94: all of this logic of creating the `connection_dict`, getting the `connection` and so on is duplicated in the restore and dump commands. Maybe we can create a separated function that is used in both places

ingestion/src/metadata/cli/restore.py
```diff
@@ -0,0 +1,101 @@
...
+    PostgresConnection,
+)
+from metadata.utils.connections import get_connection
+from metadata.utils.helpers import list_to_dict
+from metadata.utils.logger import cli_logger
+
+logger = cli_logger()
+
+
+def execute_query(engine: Engine, input: Path, schema: str = None) -> None:
```
pmbrull, line 35: This is not really running a query right? It would be something like `execute_sql_file` 🤔

ingestion/src/metadata/cli/restore.py
```diff
@@ -0,0 +1,101 @@
...
+
+    with open(input, encoding="utf-8") as f:
+        for query in f.readlines():
+
+            clean_query = query.replace("%", "%%")
```
pmbrull, line 40: let's add a quick comment on why we are doing this 🙏

pmbrull (commented):
Thanks for the changes! Left a few quick comments and we'll be good to go

ingestion/src/metadata/cli/db_dump.py
MilanBariya: Yes, Sounds good, Will do that.

ingestion/src/metadata/cli/restore.py
```diff
@@ -0,0 +1,77 @@
...
+from metadata.utils.logger import cli_logger
+
+logger = cli_logger()
+
+
+def execute_sql_file(engine: Engine, input: Path, schema: str = None) -> None:
+
+    with open(input, encoding="utf-8") as f:
+        for query in f.readlines():
+            # In some sql queries '%', and it has special meaning which is trying to band parameters. % can be escaped as %%
```
pmbrull, line 31:
```suggestion
            # `%` is a reserved syntax in SQLAlchemy to bind parameters. Escaping it with `%%`
```

pmbrull (approved):

PR #7631, opened 2022-09-21
partitioned-table-filter-added

### LINKED ISSUE #7410: BQ Sharded Table are all ingested
Sharded tables (i.e. `<table_name>_[YYYYMMDD]`) are all ingested in OM.
Screenshot 2022-09-13 at 10 08 46 AM

codingwithabhi:
### Describe your changes :

I worked on the updating filter partition table.
fix #7410

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

#
### Reviewers
@open-metadata/ingestion


harshach:
@pmbrull can we merge this in?

ingestion/src/metadata/ingestion/source/database/bigquery.py
```diff
@@ -99,6 +100,22 @@ def get_columns(bq_schema):
 _types.get_columns = get_columns
 
 
+@reflection.cache
+def get_table_names(self, connection, schema=None, **kw):
+    _, project_id = auth.default()
+    query = (
+        f"SELECT table_name FROM {project_id}.{schema}.INFORMATION_SCHEMA.PARTITIONS"
```
pmbrull, line 107: Not sure if this will be giving us the full data from BigQuery. For what I can see, running this query only returns us the partitions of the schema rather than tables and views.

For ref, these are the methods that the SQAlchemy BigQuery inspector runs: https://github.com/googleapis/python-bigquery-sqlalchemy/blob/main/sqlalchemy_bigquery/base.py#L1003

https://github.com/googleapis/python-bigquery-sqlalchemy/blob/a050ccf534fe8e6d2135818474868f072212aa6f/sqlalchemy_bigquery/base.py#L833

It's possible that we are getting all the shards because by default, the lib is running with the datasets API. In any case, +1 on running directly a call to `INFORMATION_SCHEMA` but we need to make sure that all tables and views are going to be there.

we might be able to join `INFORMATION_SCHEMA.TABLES` with `INFORMATION_SCHEMA.PARTITIONS` and ROWNUM / RANK on it to keep a single copy for the sharded tables. Thanks!

pmbrull (changes requested):
Thanks for the PR @codingwithabhi 🙏 

Sorry we missed it. I believe that you are trying to tackle https://github.com/open-metadata/OpenMetadata/issues/7410 correct? Please let's add this on the PR description for some context :)

Left a comment, I am not sure the proposed fix is what we need here. Let me know what you think.

Thanks!

ulixius9:
closing this as this solution does not look a relevant solution and we can take inspiration from this pr when we handle this issue!

Issue #6770, opened 2022-08-17
Tables dissapper from UI after containers reboot


stepanof:
**Affected module**
UI

**Describe the bug**
After restarting OM containers I can't see tables in UI
![tables_dissappear](https://user-images.githubusercontent.com/68236577/185170605-2e90c48d-4c96-402e-93e0-77b9769d9990.png)

**To Reproduce**

I have changed standard docker-compose.yaml in some places:
1) Add volume for mysql container
    volumes:
      - ./db_data:/var/lib/mysql

2) Im using 
image: elasticsearch:7.17.5
instead of 
image: docker.elastic.co/elasticsearch/elasticsearch:7.10.2 
for elasticsearch container

3) Changed some ports becuase some of them are alredy locked by other application:
 - 9201:9200 for elactic
 - 8095:8585 for openmetadata-server
 - 8096:8080 for airflow

**Expected behavior**
After restarting containers to see the same what was before

**Version:**
 - OS: ubuntu 21.04
 - Python version: 3.8
 - OpenMetadata version: 11.4

pmbrull:
hi @stepanof , you might need to reindex the data on ElasticSearch with https://github.com/open-metadata/OpenMetadata/discussions/5300 after a reboot, as the volume is only on MySQL

pmbrull:
@akash-jain-10 @vijaypm lets add the volumes by default here

stepanof:
Thankx @pmbrull !!! I'll try it later. Hope this will help. You can close the issue if needed.

pmbrull:
thanks, do let us know if you need anything else. We'll keep this issue open for us to track that we are adding the volume behavior as default. Thank you for reporting

harshach:
@Vj-L any update on this to support volumes by default

harshach:
@akash-jain-10 @Vj-L update please

Vj-L:
@harshach 
We will take this on priority for this week and try to close this up

pmbrull:
@Vj-L we need to add the JWT env vars in the docker compose files as well 🙏

PR #7741, opened 2022-09-27
Added default volumes

Vj-L:
I worked on the #6770  providing the default volumes for each of the containers 

#
### Type of change :

- [x] Bug fix

### DIFF
--- a/docker/run_local_docker.sh
+++ b/docker/run_local_docker.sh
@@ -59,8 +59,17 @@ fi
 
 echo "Stopping any previous Local Docker Containers"
 docker compose  -f docker/local-metadata/docker-compose-postgres.yml down
-docker compose down
-rm -rf $PWD/docker-volume
+docker compose -f docker/local-metadata/docker-compose.yml down
+if [ -d "/docker-volume" ]
+then
+    echo "Do you want to delete the mounted directories from host?"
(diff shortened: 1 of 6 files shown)


harshach (approved):

docker/run_local_docker.sh
```diff
@@ -55,20 +55,21 @@ else
...
 
-cd docker/local-metadata || exit
+#cd docker/local-metadata || exit
 
 echo "Stopping any previous Local Docker Containers"
-docker compose -f docker-compose-postgres.yml down
+docker compose  -f docker/local-metadata/docker-compose-postgres.yml down
 docker compose down
-
+rm -rf $PWD/docker-volume
```
akash-jain-10, line 63: Add a condition to run only if the `docker-volume` directory exists!

akash-jain-10 (changes requested):
All the `docker-compose*` looks good to me. A minor condition to add in `run_local_docker.sh` script.

docker/run_local_docker.sh
Vj-L: done

harshach:
@Vj-L what is remaining here to merge in. It looks like your diff brought some other files that are not relevant to this PR?

akash-jain-10 (approved):

PR #7803, opened 2022-09-29
[Basic auth fixes] Added check to prevent brute-force attack

mohityadav766:
### Describe your changes :

Added check to prevent brute-force attack , disable user login for 10 mins.
Send access update to mail

#
### Type of change :

- [x] Improvement
- [x] New feature
#
### Checklist:

#


openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LoginAttemptCache.java
```diff
@@ -0,0 +1,57 @@
...
+public class LoginAttemptCache {
+  private final int MAX_ATTEMPT = 3;
+  private final LoadingCache<String, Integer> attemptsCache;
+
+  public LoginAttemptCache() {
+    super();
+    attemptsCache =
+        CacheBuilder.newBuilder()
+            .maximumSize(1000)
+            .expireAfterWrite(10, TimeUnit.MINUTES)
```
vivekratnavel, line 18: Can we make this 10 minutes configurable?

vivekratnavel: Once we make it configurable, we can write unit tests around it by reducing this time to probably 30 seconds in the tests.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/UserRepository.java
```diff
@@ -65,6 +65,15 @@ public UserRepository(CollectionDAO dao, SecretsManager secretsManager) {
     this.secretsManager = secretsManager;
   }
 
+  public final Fields getFieldsWithUserAuth(String fields) {
+    if (fields != null && fields.equals("*")) {
+      List<String> tempFields = getAllowedFieldsCopy();
```
harshach, line 70: @mohityadav766, why are we adding this as additional temp fields and also we shouldn't be returning this if the logged-in user is requesting for other users, even for admin

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LoginAttemptCache.java
```diff
@@ -0,0 +1,57 @@
+package org.openmetadata.service.security.auth;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+public class LoginAttemptCache {
+  private final int MAX_ATTEMPT = 3;
```
harshach, line 10: lets make this configurable as well

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LoginAttemptCache.java
mohityadav766: @vivekratnavel at constructor level or at openmetdata.yml?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/UserRepository.java
mohityadav766: @harshach the problem is whenever we are using createOrUpdate we want to make sure to have all the userData first Fetched , and then do createOrUpdate, otherwise what happens is that the fields gets overwritten , let say I don't have "authnticationMechanism" it would be null on fetch of user , and createOrUpdate will update it to null !!

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LoginAttemptCache.java
mohityadav766: done

PR #7988, opened 2022-10-06
chore (#7373)  UI : Remove momentjs and use luxon

Sachin-chaurasiya:
Closes #7373
### Describe your changes :

I worked on issue #7373 and also removed the `react-google-login` as we are no longer using it.

#
### Type of change :

- [x] Improvement

#
### Frontend Preview (Screenshots) :

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [x] I have commented on my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] All new and existing tests passed.

#
### Reviewers
@open-metadata/ui

PR #8017, opened 2022-10-07
Fixed: Flaky cypress tests

ShaileshParmar11:
### Describe your changes :

I worked on the fixing flaky cypress tests

#
### Type of change :

- [x] Improvement

#

#
### Checklist:

#

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/cypress/common/common.js
+++ b/openmetadata-ui/src/main/resources/ui/cypress/common/common.js
@@ -18,6 +18,7 @@ import { SEARCH_INDEX } from '../constants/constants';
 export const descriptionBox =
   '.toastui-editor-md-container > .toastui-editor > .ProseMirror';
 export const uuid = () => Cypress._.random(0, 1e6);
+const RETRY_TIMES = 30;
 
 const ADMIN = 'admin';
 
@@ -51,7 +52,7 @@ export const handleIngestionRetry = (
 ) => {
   const rowIndex = ingestionType === 'metadata' ? 1 : 2;
   // ingestions page
-  const retryTimes = 30;
+  
   let retryCount = count;
   const testIngestionsTab = () => {
     cy.get('[data-testid="Ingestions"]').should('be.visible');
@@ -76,7 +77,7 @@ export const handleIngestionRetry = (
         if (
           ($ingestionStatus.text() === 'Running' ||
             $ingestionStatus.text() === 'Queued') &&
-          retryCount <= retryTimes
+          retryCount <= RETRY_TIMES
         ) {
           // retry after waiting for 20 seconds
           cy.wait(20000);
@@ -927,7 +928,6 @@ export const addTeam = (TEAM_DETAILS) => {
 };
 
 export const retryIngestionRun = () => {
-  const retryTimes = 10;
   let retryCount = 0;
 
   const testIngestionsTab = () => {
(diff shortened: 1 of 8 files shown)


openmetadata-ui/src/main/resources/ui/cypress/common/common.js
```diff
@@ -13,9 +13,12 @@
 
 /// <reference types="cypress" />
 
+import { SEARCH_INDEX } from '../constants/constants';
+
 export const descriptionBox =
   '.toastui-editor-md-container > .toastui-editor > .ProseMirror';
 export const uuid = () => Cypress._.random(0, 1e6);
+const RETRY_TIMES = 30;
```
Sachin-chaurasiya, line 21: Can we move this to the constants file?

openmetadata-ui/src/main/resources/ui/cypress/common/common.js
```diff
@@ -403,12 +406,23 @@ export const searchEntity = (term, suggestionOverly = true) => {
...
-  interceptURL('GET', '/api/v1/search/*', 'explorePageSearch');
+  interceptURL(
+    'GET',
+    `/api/v1/search/query?q=*&from=*&size=*&index=${SEARCH_INDEX[entity]}`,
+    'explorePageTabSearch'
+  );
+  interceptURL(
+    'GET',
+    `/api/v1/search/suggest?q=*&index=dashboard_search_index,table_search_index,topic_search_index,pipeline_search_index,mlmodel_search_index`,
+    'searchQuery'
```
Sachin-chaurasiya, line 417: Can we use the `SEARCH_INDEX` for the index value?

openmetadata-ui/src/main/resources/ui/cypress/common/common.js
vivekratnavel, line 21: With the retry count of 30 and a waiting time of 20 seconds between each try, we are waiting a whole 10 minutes for any ingestion job to succeed. We should not be retrying these many times in my opinion.

Sachin-chaurasiya (approved):

openmetadata-ui/src/main/resources/ui/cypress/common/common.js
ShaileshParmar11: Hi @vivekratnavel, We are not waiting for a total of 10 min for every spec, as ingestion status update is entirely dependent on airflow ingestion we are not sure about when it is going to be successful, so we had to retry until it's successful but we are just providing an upper limit to it! also, it's old logic I have just moved the variable to an outside function so we can utilize this if needed!

openmetadata-ui/src/main/resources/ui/cypress/common/common.js
ShaileshParmar11: it's fine in my opinion as we need here all the index to make URL and not any particular index or dynamic index

openmetadata-ui/src/main/resources/ui/cypress/common/common.js
ShaileshParmar11: constants file contains the data which we use across all specs, whereas `RETRY_TIMES` is related only to a particular function which is in `common.js` file

ShaileshParmar11:
`Bots.spec.js` is already fixed by #7988

vivekratnavel (approved):

PR #8034, opened 2022-10-07
Database Source Lint

ulixius9:
### Describe your changes :

Database Source Lint

#
### Type of change :

- [x] Improvement

#
### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/athena.py
+++ b/ingestion/src/metadata/ingestion/source/database/athena.py
@@ -36,7 +36,7 @@ from metadata.utils.logger import ingestion_logger
 logger = ingestion_logger()
 
 
-def _get_column_type(self, type_):  # pylint: disable=too-many-branches
+def _get_column_type(self, type_):
     """
     Function overwritten from AthenaDialect
     to add custom SQA typing.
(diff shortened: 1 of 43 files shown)


ingestion/src/metadata/ingestion/source/database/athena.py
```diff
@@ -34,12 +36,12 @@
 logger = ingestion_logger()
 
 
-def _get_column_type(self, type_):
+def _get_column_type(self, type_):  # pylint: disable=too-many-branches
```
TeddyCr, line 39: I think we should not disable that. To me it means we need to maybe change our deep if/else to a look dict or soemthing similar -- example below.

```python
def bad(a):
    if a == 1: return "A"
    if a == 2: return "B"
    if a == 3: return "D"
    if a == 4: return "G"
    if a == 5: return "M"
    if a == 6: return "X"

    return "Z"

def better(a):
    return _LOOKUP.get(a, "Z")

_LOOKUP = {
    1: "A",
    2: "B",
    3: "D",
    4: "G",
    5: "M",
    6: "X",
}
```

ingestion/src/metadata/ingestion/source/database/dbt_source.py
```diff
@@ -204,21 +208,18 @@ def _parse_data_model_upstream(self, mnode):
         return upstream_nodes
 
     def _parse_data_model_columns(
-        self, model_name: str, mnode: Dict, cnode: Dict
+        self, _: str, mnode: Dict, cnode: Dict
     ) -> List[Column]:
         columns = []
         ccolumns = cnode.get("columns")
```
TeddyCr, line 214: why is that named `ccolumns`?

ulixius9: not sure why it's like that but I guess that's pointing to columns of catalogue node

ulixius9: will rename this

OnkarVO7 (approved):

nahuelverdugo (approved):

PR #8041, opened 2022-10-07
Add CLI E2E tests for MySQL

nahuelverdugo:
### Describe your changes :
- Add CLI E2E tests for MySQL that run every day at midnight. 

### Type of change :
- [x] Improvement
- [x] New feature

### Checklist:

### Reviewers


.github/workflows/py-tests.yml
```diff
@@ -76,3 +76,29 @@ jobs:
       run: |
         source env/bin/activate
         make run_python_tests
+
+      # we have to pass these args values since we are working with the 'pull_request_target' trigger
+    - name: Push Results in PR to Sonar
```
pmbrull, line 81: thanks!!!

ingestion/src/metadata/ingestion/source/database/common_db_source.py
```diff
@@ -450,6 +449,7 @@ def connection(self) -> Connection:
     def close(self):
         if self.connection is not None:
             self.connection.close()
+        self.engine.dispose()
```
pmbrull, line 452: good catch

ingestion/tests/cli_e2e/README.md
```diff
@@ -0,0 +1,95 @@
...
+- It partially implements some methods from `test_cli_db_base`.
+- `test_cli_{connector}` is the specific connector test. More tests apart the ones implemented by the `test_cli_db_base` can be run inside this class.
+
+## How to add a database connector
+
+1. Use `test_cli_mysql.py` as example. Your connector E2E CLI test must follow the name convention: `test_cli_{connector}.py` and the test 
+class must extend from `CliCommonDB.TestSuite` if the connector's source implement the `CommonDbSourceService` class, otherwise, from `CliDBBase.TestSuite`.
+
+2. Add an ingestion YAML file with the service and the credentials of it. Use when possible a Dockerized environment, otherwise, remember to use environment
+variables for sensitive information in case of external resources.
```
pmbrull, line 16: Maybe we can add a line about how the different YAMLs get generated automatically for each workflow run

ingestion/tests/cli_e2e/test_cli_db_base_common.py
```diff
@@ -0,0 +1,162 @@
...
+from sqlalchemy.engine import Engine
+
+from metadata.ingestion.api.sink import SinkStatus
+from metadata.ingestion.api.source import SourceStatus
+from metadata.ingestion.api.workflow import Workflow
+
+from .test_cli_db_base import PATH_TO_RESOURCES, CliDBBase
+
+
+class CliCommonDB(object):
```
pmbrull, line 27: let's kill the `object` inheritance here, not needed at all

pmbrull: Also maybe silly question, I see that we are creating classes as `class MysqlCliTest(CliCommonDB.TestSuite):`. What's the benefit of wrapping the class with `CliCommonDB`? To not trigger the tests?

Maybe we can then have a function returning the `TestSuite` or not use `test_` as a module name 🤔

nahuelverdugo: Using this approach we can create an abstract `TestCase` class that avoids issues with test runner discovery and let us still import the abstract test from another module. I had already used this approach in the past for the Secrets Manager tests.

pmbrull: cool thanks for explaining. Let's then just remove the `object` notation and call it a day!

pmbrull (approved):

Issue #8061, opened 2022-10-10
Add project id in config for DBT cloud


OnkarVO7:
**Feature**
For a DBT cloud account with multiple projects it is necessary to specify the project id to get the run artifacts from the required project

Specify the `project_id` as a config parameter for DBT

PR #8062, opened 2022-10-10
Added project id to DBT cloud config

OnkarVO7:
### Describe your changes :

Fix https://github.com/open-metadata/OpenMetadata/issues/8061

#
### Type of change :

- [x] Bug fix
- [x] Improvement

#
### Checklist:

#
### Reviewers

 @open-metadata/ingestion

### DIFF
--- a/openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceMetadataPipeline.json
+++ b/openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceMetadataPipeline.json
@@ -44,6 +44,11 @@
           "description": "DBT cloud account Id",
           "type": "string"
         },
+        "dbtCloudProjectId": {
+          "title": "DBT Cloud Project Id",
+          "description": "In case of multiple projects in a DBT cloud account, specify the project's id from which you want to extract the DBT run artifacts",
+          "type": "string"
+        },
         "dbtUpdateDescriptions": {
           "description": "Optional configuration to update the description from DBT or not",
           "type": "boolean",
--- a/ingestion/src/metadata/ingestion/ometa/mixins/patch_mixin.py
+++ b/ingestion/src/metadata/ingestion/ometa/mixins/patch_mixin.py
@@ -162,6 +162,9 @@ class OMetaPatchMixin(Generic[T]):
         if not table:
             return None
 
+        if not table.columns:
+            return None
+
         col_index, col = find_column_in_table_with_index(
             column_name=column_name, table=table
         )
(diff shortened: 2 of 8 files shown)


openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceMetadataPipeline.json
```diff
@@ -44,6 +44,11 @@
           "description": "DBT cloud account Id",
           "type": "string"
         },
+        "dbtCloudProjectId": {
+          "title": "DBT Cloud Project Id",
+          "description": "In case of multiple projects in a DBT cloud account, specify the project's id from which you want to extract the DBT run artifacts",
+          "type": "string"
+        },
```
nahuelverdugo, line 51: do we need to take some action to avoid retro compatibility issues? can we have a default value?

OnkarVO7: @nahuelverdugo by default if noting is passed into this is field it will store `None` on which I have added conditions. We cannot pass any other default string value since it will make the wrong API call

ayush-shah: would empty string work here as default?

OnkarVO7: @ayush-shah empty string wont work since i cannot pass an empty string into the API

nahuelverdugo: What if we read a value already stored in DB or JSON? Would it fail?

OnkarVO7: @nahuelverdugo can you elaborate the use case a bit more?

nahuelverdugo (dismissed):

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceMetadataPipeline.json
nahuelverdugo: Resolving, since it has been solved offline. It has already been tested with null vales from UI and CLI.

nahuelverdugo (approved):

ulixius9 (approved):

PR #8065, opened 2022-10-10
Prepare Ingestion Base Docker image

pmbrull:
### Describe your changes :

This PR adds a new Docker Image `openmetadata/ingestion-base`, as a python-based image only containing the necessary requirements to run `openmetadata-ingestion`. This means, no Airflow and no extras.

Why is this interesting? It would allow users to run OpenMetadata code directly from Docker, without installing any libraries in their environments. The scenario is exemplified here by using Airflow's `DockerOperator` to run the metadata ingestion, instead of relying on the `openmetadata/ingestion` locally installed packages.

More information can be found in the added README to have further iterations (if we ever need to).

Note how we added a volume to map the host `/var/run/docker.sock` to the container. This is needed for Airflow to call Docker commands and spin up an ephemeral container that would run the ingestion from the new image. This has only been added to the local containers, as on our shipped images, we are not providing the `DockerOperator` behavior.

This is also interesting because it prepares the image that we will, later on, run in the Argo Workflows.

The next step, once the image is built, is to test how to prepare a `DockerOperator` on MWAA/GCS, which would free us from any requirement incompatibility with those environments.

Thanks

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


ingestion/examples/airflow/dags/airflow_docker_operator.py
```diff
@@ -0,0 +1,69 @@
...
+with models.DAG(
+    "ingestion-docker-operator",
+    schedule_interval="@once",
+    start_date=datetime(2021, 1, 1),
+    catchup=False,
+    tags=["OpenMetadata"],
+) as dag:
+    DockerOperator(
+        command='python main.py',
+        image='openmetadata/ingestion-base:local',
```
pmbrull, line 58: To run this on your local Airflow you'll first need to run `make build-ingestion-base-local`

ingestion/operators/docker/main.py
```diff
@@ -0,0 +1,39 @@
...
+WORKFLOW_MAP = {
+    PipelineType.metadata.value: Workflow,
+    PipelineType.usage.value: Workflow,
+    PipelineType.lineage.value: Workflow,
+    PipelineType.profiler.value: ProfilerWorkflow,
+    PipelineType.TestSuite.value: TestSuiteWorkflow,
+}
+
+
+def main():
```
pmbrull, line 20: This function is just needed to prep the Airflow `DockerOperator`, as the entrypoint must be a command. In our case, that would be passed as `python main.py` with whatever env variable needed

ingestion/ingestion_dependency.sh
```diff
@@ -39,6 +39,9 @@ airflow users create \
 
 (sleep 5; airflow db upgrade)
 (sleep 5; airflow db upgrade)
+
+chmod 666 /var/run/docker.sock
```
akash-jain-10, line 43: Hello @pmbrull . This might not work with k8s as there is now `/var/run/docker.sock` file! we might want to use this explicitly with docker-compose files using command and arguments!

akash-jain-10 (changes requested):

ingestion/ingestion_dependency.sh
pmbrull: oh actually this one I needed to remove hehe thanks, missed it!

pmbrull: that was some old testing trying to prep the permissions on the volume and here it does not work. Instead, added the permission on the docker-compose itself

akash-jain-10 (approved):

Issue #5898, opened 2022-07-06
Cypress: Add a service test with DBT


vivekratnavel:
It would be good to have coverage around DBT. We can add a service with DBT configs and check the DBT tab on the entity page.

PR #8162, opened 2022-10-15
UI: Add Cypress for DTB in redshift container

ShaileshParmar11:
### Describe your changes :

I worked on adding  cypress for DBT in the redshift container
Closes [#5898](https://github.com/open-metadata/OpenMetadata/issues/5898)
#
### Type of change :

- [x] Improvement

#

#
### Checklist:

#


openmetadata-ui/src/main/resources/ui/cypress/e2e/AddNewService/redshiftWithDBT.spec.js
```diff
@@ -0,0 +1,169 @@
...
+import {
+    DBT, HTTP_CONFIG_SOURCE, LOGIN,
+    SERVICE_TYPE
+} from '../../constants/constants';
+
+const serviceType = 'Redshift';
+const serviceName = `${serviceType}-ct-test-${uuid()}`;
+const tableName = 'boolean_test';
+const DBTTable = 'customers';
+const description = `This is ${serviceName} description`;
```
chirag-madlani, line 34: Can we move them to const?
If needed split the constant file to something like: `services.constant.ts` & `common.constant.ts`

openmetadata-ui/src/main/resources/ui/cypress/e2e/AddNewService/redshiftWithDBT.spec.js
```diff
@@ -0,0 +1,169 @@
...
+
+describe('RedShift Ingestion', () => {
+  beforeEach(() => {
+    login(LOGIN.username, LOGIN.password);
+    cy.goToHomePage();
+  });
```
chirag-madlani, line 40: Can we see if we are able to restore local storage via method that @vivekratnavel suggested and save some time on tests via eliminating this code block.

openmetadata-ui/src/main/resources/ui/cypress/e2e/AddNewService/redshiftWithDBT.spec.js
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+    testServiceCreationAndIngestion(
+      serviceType,
+      connectionInput,
+      addIngestionInput,
+      serviceName,
+      'database',
+      true,
+      configureDBT
```
chirag-madlani, line 93: Not sure why we are passing it we can call it manually post this method runs successfully.
Let me know if we have any solid reason to put it as callback here

ShaileshParmar11: So DBT config step can differ from connector to connector and its data as well, so to perform dynamic logic into the common flow we are passing the function!

openmetadata-ui/src/main/resources/ui/cypress/e2e/AddNewService/redshiftWithDBT.spec.js
ShaileshParmar11: We did try this but facing an issue while reloading the page, and test is getting failed

openmetadata-ui/src/main/resources/ui/cypress/e2e/AddNewService/redshiftWithDBT.spec.js
chirag-madlani: What's challenge in calling it after this file it self?

chirag-madlani: Since it's intermediate step we need it as callback function

chirag-madlani (approved):

Issue #6157, opened 2022-07-18
Add Column Level Lineage for DBT Source


ulixius9:
Add Column Level Lineage for DBT Source based on  #5190

PR #8176, opened 2022-10-17
Added DBT col level lineage using queries

OnkarVO7:
### Describe your changes :

Fix https://github.com/open-metadata/OpenMetadata/issues/6157

#
### Type of change :

- [x] Improvement

#
image

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/dbt_source.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt_source.py
@@ -255,7 +255,9 @@ class DBTMixin:
 
         return columns
 
-    def process_dbt_lineage_and_descriptions(self) -> Iterable[AddLineageRequest]:
+    def process_dbt_lineage_and_descriptions(  # pylint: disable=too-many-locals
+        self,
+    ) -> Iterable[AddLineageRequest]:
         """
         After everything has been processed, add the lineage info
         """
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/ingestion/source/database/dbt_source.py
```diff
@@ -309,6 +312,30 @@ def process_dbt_lineage_and_descriptions(self) -> Iterable[AddLineageRequest]:
...
+            # Create Lineage from DBT Queries
+            try:
+                service_database_schema_table = fqn.split(data_model_name)
+                target_table_fqn = ".".join(service_database_schema_table[1:])
+                service_name = service_database_schema_table[0]
+                database_name = service_database_schema_table[1]
+                database_schema_name = service_database_schema_table[2]
+                create_statement = f"create table {target_table_fqn} as"
+                query = f"{create_statement} {data_model.sql.__root__}"
+                lineages = get_lineage_by_query(
```
pmbrull, line 324: You might be able to get rid of the `too-many-locals` linting by moving these assignments directly inside the `get_lineage_by_query` function. These variables are just being used once inside the function, so having them aside is not really adding much value.

Another possibility would be to split the function in smaller pieces, but it might not be needed in this case.

Thanks!

OnkarVO7: Got it will do that 👍

pmbrull (approved):

PR #8248, opened 2022-10-19
Fix test suite bug for redefined quotes

TeddyCr:
### Describe your changes :

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/interfaces/sqalchemy/sqa_test_suite_interface.py
+++ b/ingestion/src/metadata/interfaces/sqalchemy/sqa_test_suite_interface.py
@@ -16,6 +16,7 @@ supporting sqlalchemy abstraction layer
 
 from datetime import datetime, timezone
 from typing import Optional, Union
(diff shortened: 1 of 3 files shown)


pmbrull (approved):
Leaving the approval.

Could you please elaborate a bit on the issue + fix?

TeddyCr:
> 

When we convert table to orm object we use the same MetaData store by default. So if two tables in 2 different database are created as orm object then it will try to replace that existing object. It is a similar issue we had back then with the profiler. 

Let me know if that makes sense.

pmbrull:
> > 
> 
> @pmbrull When we convert table to orm object we use the same MetaData store by default. So if two tables in 2 different database are created as orm object then it will try to replace that existing object. It is a similar issue we had back then with the profiler.
> 
> Let me know if that makes sense.

ooh oki got it, thanks for explaining!

Issue #6775, opened 2022-08-17
Openmetadata secure Active Directory


AntonArgon:
**Is your feature request related to a problem? Please describe.**
Good time! Our company uses the corporate local Active Directory for authentication. The link https://docs.open-metadata.org/deployment/security did not find such an authentication method. Cloud solutions are not suitable for security requirements. 

**Describe the solution you'd like**
Are there any plans to add authentication to openmetadata via AD (ldap) config?
This issue https://github.com/open-metadata/OpenMetadata/issues/3376 does not answer.

mohityadav766:
hi @AntonArgon  , i have recently raised a PR with support for LDAP 
Can you try this out and give us any feedback if it works for you?
Let me know i can help you with setting it up

Issue #6428, opened 2022-07-29
Add LDAP Authorization


konnectr:
**Is your feature request related to a problem? Please describe.**
Can't use in production environment without authorization

**Describe the solution you'd like**

Add LDAP Authorization Capability

konnectr:
Duplicate https://github.com/open-metadata/OpenMetadata/issues/3376

mohityadav766:
hi @konnectr , i have recently raised a PR with support for LDAP 
It would really help if you give it a try, please let me know , i can help you in setting things up

stepanof:
@mohityadav766 Hello! 
If you want I can test it, but I need instrustion how to switch my docker images to images with yours changes and also I need help with settings :)

PR #8256, opened 2022-10-19
[LDAP] Added Ldap-Authentication

mohityadav766:
### Describe your changes :

Add LDAP Authorization #6428
Openmetadata secure Active Directory #6775

#
### Type of change :

- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

#


openmetadata-service/pom.xml
```diff
@@ -400,6 +400,11 @@
       <artifactId>expiringmap</artifactId>
       <version>${expiring.map.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.unboundid</groupId>
+      <artifactId>unboundid-ldapsdk</artifactId>
+      <version>6.0.6</version>
```
harshach, line 406: can we declare the version in root pom.xml and refer it here

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -1131,6 +1205,112 @@ public Response loginUserWithPassword(
     }
   }
 
+  private Response handleLdapAuthLogin(UriInfo uriInfo, LoginRequest req) {
```
harshach, line 1208: can we move all of this code out of Resource into a separate class of its own, probably AuthenticationHandlers and possibly an interface to extend from. We have username/password , SAML and Ldap. All of these existing in its own method doesn't look right to me.
We can build a interface to implement the different mechanisms

openmetadata-spec/src/main/resources/json/schema/auth/ldapConfiguration.json
```diff
@@ -0,0 +1,68 @@
...
+  "$id": "https://open-metadata.org/schema/api/auth/ldapConfiguration.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "LdapConfiguration",
+  "description": "LDAP Configuration",
+  "type": "object",
+  "javaType": "org.openmetadata.schema.auth.LdapConfiguration",
+  "properties": {
+    "host": {
+      "description": "LDAP server address without scheme(Example :- localhost)",
+      "type": "string"
```
harshach, line 11: this should be URI right?

mohityadav766: no it's the hostname

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/AuthenticatorHandler.java
```diff
@@ -0,0 +1,92 @@
...
+  default User registerUser(RegistrationRequest registrationRequest) throws IOException {
+    throw new CustomExceptionMessage(Response.Status.NOT_FOUND, NOT_IMPLEMENTED_METHOD);
+  }
+
+  default void sendEmailVerification(UriInfo uriInfo, User user) throws IOException {
+    throw new CustomExceptionMessage(Response.Status.NOT_FOUND, NOT_IMPLEMENTED_METHOD);
+  }
+
+  default void confirmEmailRegistration(UriInfo uriInfo, String emailToken) throws IOException {
+    throw new CustomExceptionMessage(Response.Status.NOT_FOUND, NOT_IMPLEMENTED_METHOD);
```
harshach, line 45: Its better to return 501 https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501 then the NOT_FOUND error code here

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+      validateEmailAlreadyExists(newRegistrationRequestEmail);
+      PasswordUtil.validatePassword(newRegistrationRequest.getPassword());
+      LOG.info("Trying to register new user [" + newRegistrationRequestEmail + "]");
+      User newUser = getUserFromRegistrationRequest(newRegistrationRequest);
+      // remove auth mechanism from the user
+      User registeredUser = userRepository.create(null, newUser);
+      registeredUser.setAuthenticationMechanism(null);
+      return registeredUser;
+    } else {
+      throw new CustomExceptionMessage(BAD_REQUEST, SELF_SIGNUP_ERROR);
```
harshach, line 107: I would use NOT supported here as well

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+    User registeredUser =
+        userRepository.get(null, emailVerificationToken.getUserId(), userRepository.getFieldsWithUserAuth("*"));
+    if (Boolean.TRUE.equals(registeredUser.getIsEmailVerified())) {
+      LOG.info("User [{}] already registered.", emailToken);
+      return;
+    }
+
+    // verify Token Expiry
+    if (emailVerificationToken.getExpiryDate().compareTo(Instant.now().toEpochMilli()) < 0) {
+      throw new RuntimeException(
```
harshach, line 126: Lets not throw RuntimeException here. Just add a TokenExpiry error to CatalogExceptionMessage and use that

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+    // deleting the entry for the token from the Database
+    tokenRepository.deleteTokenByUserAndType(registeredUser.getId().toString(), EMAIL_VERIFICATION.toString());
+  }
+
+  @Override
+  public void resendRegistrationToken(UriInfo uriInfo, String userName) throws IOException {
+    User registeredUser = userRepository.getByName(uriInfo, userName, userRepository.getFields("isEmailVerified"));
+    if (Boolean.TRUE.equals(registeredUser.getIsEmailVerified())) {
+      // no need to do anything
+      throw new CustomExceptionMessage(FORBIDDEN, "Email Already Verified For User.");
```
harshach, line 145: FORBIDDEN not a right error code right. Don't think we should even throw an error here instead return a 200 with Email Already Verified

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -177,6 +163,12 @@ public void initialize(OpenMetadataApplicationConfig config) throws IOException
     SmtpSettings smtpSettings = config.getSmtpSettings();
     this.isEmailServiceEnabled = smtpSettings != null && smtpSettings.getEnableSmtpServer();
     this.loginAttemptCache = new LoginAttemptCache(config);
+    if (config.getAuthenticationConfiguration().getProvider().equals("ldap")) {
+      authHandler = new LdapAuthenticator(dao, tokenRepository, config);
+      // initLdapConfig(config.getAuthenticationConfiguration().getLdapConfiguration());
+    } else if (config.getAuthenticationConfiguration().getProvider().equals("basic")) {
+      authHandler = new BasicAuthenticator(dao, tokenRepository, config);
+    }
```
harshach, line 171: we should add an else condition just for safety to throw an exception

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -878,17 +846,15 @@ public Response generateResetPasswordLink(@Context UriInfo uriInfo, @Valid Email
...
-        LOG.error("Error in sending mail for reset password" + ex.getMessage());
-        return Response.status(424).entity(new ErrorMessage(424, EMAIL_SENDING_ISSUE)).build();
-      }
+    try {
+      authHandler.sendPasswordResetLink(
+          uriInfo,
+          registeredUser,
+          EmailUtil.getInstance().getPasswordResetSubject(),
+          EmailUtil.PASSWORD_RESET_TEMPLATE_FILE);
+    } catch (Exception ex) {
```
harshach, line 855: we should catch specific exceptions instead of global. May be there is retires that we can do in sending the email.

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+      throw new RuntimeException("Token does not belong to the user.");
+    }
+    verifyPasswordResetTokenExpiry(passwordResetToken);
+    // passwords validity
+    if (!request.getPassword().equals(request.getConfirmPassword())) {
+      throw new RuntimeException("Password and Confirm Password should match");
+    }
+    PasswordUtil.validatePassword(request.getPassword());
+
+    String newHashedPwd = BCrypt.withDefaults().hashToString(12, request.getPassword().toCharArray());
```
harshach, line 223: what is the 12 number is about here? can we add some comments and add a final constant

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+        userRepository.getByName(
+            uriInfo, request.getUsername(), new EntityUtil.Fields(fields, String.join(",", fields)));
+    // token validity
+    if (!passwordResetToken.getUserId().equals(storedUser.getId())) {
+      throw new RuntimeException("Token does not belong to the user.");
+    }
+    verifyPasswordResetTokenExpiry(passwordResetToken);
+    // passwords validity
+    if (!request.getPassword().equals(request.getConfirmPassword())) {
+      throw new RuntimeException("Password and Confirm Password should match");
```
harshach, line 219: avoid the runtime exception, it should IllegalArgumentException

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+    }
+    loginAttemptCache.recordSuccessfulLogin(request.getUsername());
+  }
+
+  @Override
+  public void changeUserPwdWithOldPwd(UriInfo uriInfo, String userName, ChangePasswordRequest request)
+      throws IOException {
+    // passwords validity
+    if (!request.getNewPassword().equals(request.getConfirmPassword())) {
+      throw new RuntimeException("Password and Confirm Password should match");
```
harshach, line 248: same here

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+    }
+    PasswordUtil.validatePassword(request.getNewPassword());
+    List<String> fields = userRepository.getAllowedFieldsCopy();
+    fields.add(USER_PROTECTED_FIELDS);
+
+    // Fetch user
+    User storedUser =
+        userRepository.getByName(uriInfo, userName, new EntityUtil.Fields(fields, String.join(",", fields)));
```
harshach, line 256: any reason not to query using "*" for fields here

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+      // remove login/details from cache
+      loginAttemptCache.recordSuccessfulLogin(request.getUsername());
+      try {
+        sendInviteMailToUser(
+            uriInfo,
+            response.getEntity(),
+            String.format("%s: Password Update", EmailUtil.getInstance().getEmailingEntity()),
+            ADMINCREATE,
+            request.getNewPassword());
+      } catch (Exception ex) {
```
harshach, line 286: avoid catching generic Exception

harshach: any exception here should result in a valid error message back to the user so that they can retry

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+        storedBasicAuthMechanism.setPassword(newHashedPassword);
+        storedUser.getAuthenticationMechanism().setConfig(storedBasicAuthMechanism);
+        userRepository.createOrUpdate(uriInfo, storedUser);
+        // remove login/details from cache
+        loginAttemptCache.recordSuccessfulLogin(userName);
+      } else {
+        throw new CustomExceptionMessage(UNAUTHORIZED, "Old Password is not correct");
+      }
+    } else {
+      String newHashedPassword = BCrypt.withDefaults().hashToString(12, request.getNewPassword().toCharArray());
```
harshach, line 272: Aren't these if else's doing the something if so can we move to a method on changing password and check if the user is the one who is changing or admin

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+        templatePopulator.put(EmailUtil.APPLICATION_LOGIN_LINK, EmailUtil.getInstance().getOMUrl());
+        try {
+          EmailUtil.getInstance()
+              .sendMail(
+                  subject,
+                  templatePopulator,
+                  user.getEmail(),
+                  EmailUtil.EMAIL_TEMPLATE_BASEPATH,
+                  EmailUtil.INVITE_RANDOM_PWD);
+        } catch (Exception ex) {
```
harshach, line 311: same as before

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+        break;
+      default:
+        LOG.error("Invalid Password Create Type");
+    }
+  }
+
+  @Override
+  public RefreshToken createRefreshTokenForLogin(UUID currentUserId) throws JsonProcessingException {
+    // first delete the existing user mapping for the token
+    // TODO: Currently one user will be mapped to one refreshToken , so essentially each user is assigned one
```
harshach, line 326: can we file issues to track this instead of TODOs

harshach (changes requested):

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -177,6 +163,12 @@ public void initialize(OpenMetadataApplicationConfig config) throws IOException
     SmtpSettings smtpSettings = config.getSmtpSettings();
     this.isEmailServiceEnabled = smtpSettings != null && smtpSettings.getEnableSmtpServer();
     this.loginAttemptCache = new LoginAttemptCache(config);
+    if (config.getAuthenticationConfiguration().getProvider().equals("ldap")) {
```
harshach, line 166: can we do as part of OpenMetadataApplication to instantiate the implementation just like we do for DefaultAuthorizer?

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+      // no need to do anything
+      throw new CustomExceptionMessage(FORBIDDEN, "Email Already Verified For User.");
+    }
+    tokenRepository.deleteTokenByUserAndType(registeredUser.getId().toString(), EMAIL_VERIFICATION.toString());
+    sendEmailVerification(uriInfo, registeredUser);
+  }
+
+  @Override
+  public void sendEmailVerification(UriInfo uriInfo, User user) throws IOException {
+    if (isEmailServiceEnabled) {
```
harshach, line 153: can we move this check to sendEmailVerification. If the service is available it will send it otherwise it won't.
Since we are going to call sendEmail in general several places such as task notification or data insights etc.. we need to repeat this many times instead add it to a central method where the email is actually being sent

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+        LOG.error("Error in sending invite to User" + ex.getMessage());
+      }
+    }
+  }
+
+  public void sendInviteMailToUser(
+      UriInfo uriInfo, User user, String subject, CreateUser.CreatePasswordType requestType, String pwd)
+      throws IOException {
+    switch (requestType) {
+      case ADMINCREATE:
```
harshach, line 296: lets separate the words with _ , ADMIN_CREATE

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+                  subject,
+                  templatePopulator,
+                  user.getEmail(),
+                  EmailUtil.EMAIL_TEMPLATE_BASEPATH,
+                  EmailUtil.INVITE_RANDOM_PWD);
+        } catch (Exception ex) {
+          LOG.error("Failed in sending Mail to user [{}]. Reason : {}", user.getEmail(), ex.getMessage());
+        }
+        break;
+      case USERCREATE:
```
harshach, line 315: same here

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+    response.setAccessToken(jwtAuthMechanism.getJWTToken());
+    response.setRefreshToken(refreshToken.getToken().toString());
+    response.setExpiryDuration(jwtAuthMechanism.getJWTTokenExpiresAt());
+
+    return response;
+  }
+
+  public void verifyPasswordResetTokenExpiry(PasswordResetToken token) {
+    if (token.getExpiryDate().compareTo(Instant.now().toEpochMilli()) < 0) {
+      throw new RuntimeException(
```
harshach, line 360: avoid RuntimeExceptions

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+    String hashedPwd = BCrypt.withDefaults().hashToString(12, create.getPassword().toCharArray());
+
+    BasicAuthMechanism newAuthMechanism = new BasicAuthMechanism().withPassword(hashedPwd);
+    return new User()
+        .withId(UUID.randomUUID())
+        .withName(username)
+        .withFullyQualifiedName(username)
+        .withEmail(create.getEmail())
+        .withDisplayName(create.getFirstName() + create.getLastName())
+        .withIsBot(false)
```
harshach, line 399: isAdmin should be set to false as well. We should ignore or throw an error anyone coming from registration request as admin.
Only others admin can add an admin using the add user form.
Initial admins can only be added from openmetadata.yaml config

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+  public void validateEmailAlreadyExists(String email) {
+    if (userRepository.checkEmailAlreadyExists(email)) {
+      throw new RuntimeException("User with Email Already Exists");
+    }
+  }
+
+  @Override
+  public JwtResponse loginUser(LoginRequest loginRequest) throws IOException, TemplateException {
+    String userName =
+        loginRequest.getEmail().contains("@") ? loginRequest.getEmail().split("@")[0] : loginRequest.getEmail();
```
harshach, line 418: do we need to check this. JsonSchema should enforce the email must contain a email format

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -0,0 +1,477 @@
...
+    User storedUser = lookUserInProvider(userName);
+    // validate User's Password
+    validatePassword(storedUser, loginRequest.getPassword());
+    // Return a Jwt Response
+    return getJwtResponse(storedUser);
+  }
+
+  @Override
+  public void checkIfLoginBlocked(String userName) {
+    if (loginAttemptCache.isLoginBlocked(userName)) {
```
harshach, line 431: whats the time period we lock the user?

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LdapAuthenticator.java
```diff
@@ -0,0 +1,250 @@
...
+      }
+    } catch (LDAPException e) {
+      LOG.warn("[LDAP] Issue in creating a LookUp Connection");
+    }
+    return null;
+  }
+
+  @Override
+  public JwtResponse loginUser(LoginRequest loginRequest) throws IOException, TemplateException {
+    // validate Login is Not Blocked
```
harshach, line 124: same here, lets remove the comments. Can this be the default implementation as it look the same to BasicAuth

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LdapAuthenticator.java
```diff
@@ -0,0 +1,250 @@
...
+    if (loginAttemptCache.isLoginBlocked(email)) {
+      throw new AuthenticationException(MAX_FAILED_LOGIN_ATTEMPT);
+    }
+  }
+
+  @Override
+  public void recordFailedLoginAttempt(User storedUser) throws TemplateException, IOException {
+    loginAttemptCache.recordFailedLogin(storedUser.getEmail());
+    int failedLoginAttempt = loginAttemptCache.getUserFailedLoginCount(storedUser.getEmail());
+    if (failedLoginAttempt == 3 /*TODO : take it from config*/) {
```
harshach, line 157: can we fix the TODOs looks like we read it for basic auth but not for LDAP. This will be confusing for anyone during the support as we take some from config but not others

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LdapAuthenticator.java
```diff
@@ -0,0 +1,250 @@
...
+        try (LDAPConnection connection =
+            new LDAPConnection(
+                sslUtil.createSSLSocketFactory(),
+                connectionOptions,
+                ldapConfiguration.getHost(),
+                ldapConfiguration.getPort())) {
+          // Use the connection here.
+          return new LDAPConnectionPool(connection, ldapConfiguration.getMaxPoolSize());
+        } catch (GeneralSecurityException e) {
+          LOG.warn("[LDAP] Issue in creating a LookUp Connection SSL");
```
harshach, line 105: should we throw a IllegalStateException , if we only log warning and server starts users might think server is started fine but ldap is not working

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LdapAuthenticator.java
```diff
@@ -0,0 +1,250 @@
...
+        LDAPConnection conn =
+            new LDAPConnection(
+                ldapConfiguration.getHost(),
+                ldapConfiguration.getPort(),
+                ldapConfiguration.getDnAdminPrincipal(),
+                ldapConfiguration.getDnAdminPassword());
+        return new LDAPConnectionPool(conn, ldapConfiguration.getMaxPoolSize());
+      }
+    } catch (LDAPException e) {
+      LOG.warn("[LDAP] Issue in creating a LookUp Connection");
```
harshach, line 117: same here

harshach (changes requested):

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
mohityadav766: we are updating the user's password and use the method createOrUpdate , if i don't get all the fields and for example just use "authenticationMechanism" to get User , the missing fields for example isEmailVerified which i did not fetch are all updated to the current value that is null in this case, so i need to fetch all the fields for intenal calls , so that update does not cause changes to other fields

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
mohityadav766: we allow username login as well

harshach (approved):

PR #8237, opened 2022-10-19
Fixes #8236 - Initialize bots from JSON data files

sureshms:
### Describe your changes :

See #8236 for more details
#
### Type of change :

- [x] New feature

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] All new and existing tests passed.

#

PR #8281, opened 2022-10-20
Revert "Fixes #8236 - Initialize bots from JSON data files"

### LINKED ISSUE #8236: Initialize bots from JSON data files
**Feature**
#4199 

**Describe the task**
Use JSON data files to initialize the bots during start up

pmbrull:
Reverts open-metadata/OpenMetadata#8237

We need to double-check how to prepare migrations and the impact on identifying the bot type in the python side. This would kill existing deployments during an upgrade.

Thanks!


nahuelverdugo (approved):

sureshms:
> Reverts #8237
> 
> We need to double-check how to prepare migrations and the impact on identifying the bot type in the python side. This would kill existing deployments during an upgrade.
> 
> Thanks!

@pmbrull with the previous change, everything remains the same, except now `ingestion-bot`, `quality-bot`, and `profiler-bot` are reserved names of the bots. They are system bots that can't be deleted. You can look them up by the reserved names, get the corresponding user for authentication purposes.

pmbrull:
> > Reverts #8237
> > We need to double-check how to prepare migrations and the impact on identifying the bot type in the python side. This would kill existing deployments during an upgrade.
> > Thanks!
> 
> @pmbrull with the previous change, everything remains the same, except now `ingestion-bot`, `quality-bot`, and `profiler-bot` are reserved names of the bots. They are system bots that can't be deleted. You can look them up by the reserved names, get the corresponding user for authentication purposes. Why would this kill the system during the upgrade?

I might be mistaken here, but there is this [change](https://github.com/open-metadata/OpenMetadata/pull/8281/files#diff-4b0c69302fa1cab7fabe7d2dca3ed70fc5caed5aef5757b9980c9c3a0da87b37L34) where we are renaming a property from `provider` to `botType`. For existing bots with the `provider` property already stored in the database, would not they die if we read them and try to cast them as the `Bot` class, which won't have `provider` but rather `botType`?

sureshms:
@pmbrull sorry this is a change in a series of changes. I was planning to get to upgrade after this change. In the next iteration, will take care of the upgrade, python, and UI code.

pmbrull:
> @pmbrull sorry this is a change in a series of changes. I was planning to get to upgrade after this change. In the next iteration of this updated patch, will take care of the upgrade, python, and UI code.

ooh oki got it, thanks!! In this case if we're planning to tackle it in multiple iterations, would it make sense to work directly on a feature branch and bring it to main once its done and stable? We can still follow the discussions in there but we can avoid misunderstandings about missing pieces that'll come later

sureshms:
All required things are done in the PR https://github.com/open-metadata/OpenMetadata/pull/8281

Can you please review the python code changes?

PR #8421, opened 2022-10-28
Fix sonar issues

ShajithaMohammed:
### Describe your changes :
This PR addresses below code smells
- Rename constant name
- Define a constant instead of duplicating string literal 
- Use already-defined constant 

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
@@ -22,6 +22,7 @@ import static org.openmetadata.service.Entity.ADMIN_USER_NAME;
 import static org.openmetadata.service.Entity.FIELD_DELETED;
 import static org.openmetadata.service.Entity.FIELD_DESCRIPTION;
 import static org.openmetadata.service.Entity.FIELD_DISPLAY_NAME;
+import static org.openmetadata.service.Entity.FIELD_EXTENSION;
 import static org.openmetadata.service.Entity.FIELD_FOLLOWERS;
 import static org.openmetadata.service.Entity.FIELD_OWNER;
 import static org.openmetadata.service.Entity.FIELD_TAGS;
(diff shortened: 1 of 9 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -129,6 +129,7 @@
  */
 @Slf4j
 public abstract class EntityRepository<T extends EntityInterface> {
+  private static final String EXTENSION = "extension";
```
nahuelverdugo, line 132: Instead of creating a new static variable we could use `FIELD_EXTENSION` from `Entity.java`, wdyt?

nahuelverdugo, line 132: Instead of creating a new static variable we could use FIELD_EXTENSION from Entity.java, wdyt?

nahuelverdugo (commented):
@ShajithaMohammed thank you for your first contribution! I have left a minor comment, let me know if you can address it or not.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
ShajithaMohammed: Makes sense @nahuelverdugo, this is addressed now!

harshach (approved):

nahuelverdugo (approved):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
nahuelverdugo: @ShajithaMohammed awesome, thank you!

PR #8435, opened 2022-10-28
feat: add SageMaker connector

techytushar:
Signed-off-by: Tushar Mittal <chiragmittal.mittal@gmail.com>

### Describe your changes :
I worked on the adding a ingestion source for SageMaker Models

#
### Type of change :

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
</p>
Screenshot 2022-10-29 at 8 02 18 PM
Screenshot 2022-10-29 at 8 02 45 PM
Screenshot 2022-10-29 at 8 03 42 PM

#

### Checklist:

#


ingestion/src/metadata/ingestion/source/mlmodel/sagemaker.py
```diff
@@ -0,0 +1,176 @@
...
+                self.source_config.mlModelFilterPattern, mlmodel_name=model["ModelName"]
+            ):
+                self.status.filter(
+                    model["ModelName"],
+                    "MlModel name pattern not allowed",
+                )
+                continue
+
+            yield SageMakerModel(
+                **{
```
pmbrull, line 95: nit - why are we unpacking a dict here instead of preparing the class directly?

```python
yield SageMakerModel(
    name=model["ModelName"],
   ...
)
```

ingestion/src/metadata/ingestion/source/mlmodel/sagemaker.py
```diff
@@ -0,0 +1,176 @@
...
+            )
+        return cls(config, metadata_config)
+
+    def get_mlmodels(  # pylint: disable=arguments-differ
+        self,
+    ) -> Iterable[SageMakerModel]:
+        """
+        List and filters models
+        """
+        for model in self.sagemaker.list_models()["Models"]:
```
pmbrull, line 84: could we please add some exception management in case:
- there is an issue picking up the models?
- There is a pydantic ValidationError when creating the model?
The goal of a workflow is that it should be as "safe" as possible, so even if we have an issue with a single model, the pipeline itself should keep running and processing further models

ingestion/src/metadata/ingestion/source/mlmodel/sagemaker.py
```diff
@@ -0,0 +1,176 @@
...
+            logger.warning(
+                f"Wild error adding TagLabel from model tags: {model_arn} - {err}"
+            )
+        return None
+
+    def _get_hyper_params(self, *args, **kwargs) -> Optional[List[MlHyperParameter]]:
+        pass
+
+    def _get_ml_features(self, *args, **kwargs) -> Optional[List[MlFeature]]:
+        pass
```
pmbrull, line 176: Oh, is there no way to pick up the features from the API? It would be one of the most interesting pieces IMO to explore similar models and possibly reuse implementations across DS teams

pmbrull (commented):
Hi @techytushar! Many thanks for the PR; looks pretty great 🚀 

Left a couple of comments; getting the features would be pretty awesome.

Also, if you have the time it would be interesting to get a couple of tests (we have some examples of testing connectors). As the data comes from an API, playing with a mock example of a model, might be helpful.

Thanks!

ingestion/src/metadata/ingestion/source/mlmodel/sagemaker.py
techytushar: I tried to search for it but couldn't find anything. I don't think SageMaker stores info about the data features on which the model was trained on.Please let me know if you find anything, I'll be happy to add it.

techytushar:
One question I had was, the `boto3 list` functions returns 10 entries by defaults and at max they can return `100` in one call. What if the user has more than `100` models in their account ? Should we run the `list` call in a loop to get all models ?

pmbrull:
> 

Hi @techytushar, thanks for the comment. IMO what we could do here is:

1. I believe that the [boto3 list_models](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.list_models) supports pagination. We should iterate over all pages (picking them up by batches of 100 should be good enough). Then the logic would be:
    - I get a page with 100 elements. I yield all of them
    - If the  response gives me a `nextToken` element, I call the list_models again passing that token as a continuation string
    - I yield the new elements I received and repeat
2. Regarding Training Jobs vs. models. I believe that what we did on MLflow was to ingest the model itself and give it the data of the latest "Version". I assume that each training job is a new experiment being run on top of an existing model. Then, I'd say that ingesting models is the right way to go, where we pick up the freshest info based on their latest training.

In terms of getting the missing info pieces. Let me take a look at the boto3 API and see if I manage to make sense of any of it 😬 

For now I found the [list_model_metadata](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.list_model_metadata) call, which might help in getting Domain and/or Framework, which might seem that it can be used to get the ML model algorithm (?).

I'll ping here if I find anything else. In any case, many thanks for the work here. Ping me if you need anything else

pmbrull:
Oki this might not make sense, haven't had the chance to play with SageMaker for a few years. But looks like we have this Feature Store class, from where we can pick up feature groups and their details: https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_DescribeFeatureGroup.html

Maybe this can be reused here in case there is any way to link a SageMaker model to the Feature Group it uses.

Let me know what you think. Thanks!

pmbrull:
Also found the [describe_algorithm](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.describe_algorithm) call. Looks like we can get some hyper params from there if there's a way to link the algorithm to the SageMaker model itself

techytushar:
Hey @pmbrull ,
Thanks for the response.
I have added pagination to the model listing, so it will now list all the models that are available.

For the feature_store and describe_algorithm class I looked into it but there is no way we can link them with the model. When creating the model it just takes the image path and the S3 path to mount any additional file to the container and then just runs that image.

pmbrull:
> 

Thanks, sounds good. If that's a limitation on SageMaker side of things we'll need to document and live with it :)

Thanks for checking, approving!

pmbrull (approved):

PR #8544, opened 2022-11-04
WIP: Fixed amundsen ingestion

### LINKED ISSUE #8525: Automating Amundsen Ingestion WorkFlow
Creating an automatic ingestion workflow for Amundsen.

Currently for Amundsen , need to create empty service from ui , need to eliminate these.

NiharDoshi99:
### Describe your changes :

I worked on the automation amundsen ingestion workflow.
Fix #8525 

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation


openmetadata-spec/src/main/resources/json/schema/api/services/createDatabaseService.json
```diff
@@ -31,6 +31,6 @@
       "$ref": "../../type/entityReference.json"
     }
   },
-  "required": ["name", "serviceType", "connection"],
+  "required": ["name", "serviceType"],
```
ulixius9, line 34: now since we are passing the connection anyways let's not remove this from required

ingestion/examples/amundsen_data/service_type.json
```diff
@@ -0,0 +1,32 @@
...
+        "connection": {
+            "config": {
+                "awsConfig": {
+                    "awsRegion": "aws_region"
+                },
+                "type": "DynamoDB"
+            }
+        }
+    }
+}
```
ulixius9, line 32: there are many more data sources that supports (here is the [list](https://docs.open-metadata.org/connectors/metadata/amundsen#3-select-the-service-type)) we might need to add those here as well

ingestion/src/metadata/ingestion/source/metadata/amundsen.py
```diff
@@ -495,7 +503,31 @@ def get_type_primitive_type(self, data_type):
...
+                service_type: dict = json.load(file)
+
+        except FileNotFoundError:
+            raise Exception()
+
+        service = self.metadata.create_or_update(
+            CreateDatabaseServiceRequest(
+                name=service_name,
+                displayName=service_name,
+                description="this is code maede",
```
ulixius9, line 523: if we are not getting description we can skip 😄

ingestion/src/metadata/ingestion/source/metadata/amundsen.py
```diff
@@ -495,7 +503,31 @@ def get_type_primitive_type(self, data_type):
...
+                connection=service_type.get(service_name, "None")["connection"],
```
ulixius9, line 524: this fails when service_name is not found in `service_type` dict as 
`"None"["connections"]` would fail.

also in case we didn't find the service type map maybe we can create a MySQL service and ingest data over there

ingestion/src/metadata/ingestion/source/metadata/amundsen.py
```diff
@@ -495,7 +503,31 @@ def get_type_primitive_type(self, data_type):
...
+                serviceType=service_type.get(service_name, DatabaseServiceType.Mysql)[
+                    "service_name"
+                ],
```
ulixius9, line 527: also we need to handle this properly because `DatabaseServiceType.Mysql["service_name"]` would fail

ingestion/src/metadata/ingestion/source/metadata/amundsen.py
```diff
@@ -495,7 +503,31 @@ def get_type_primitive_type(self, data_type):
...
+        """
+        try:
+            file_path = (
+                Path(__file__).cwd() / "examples/amundsen_data/service_type.json"
+            )
+            with open(file_path, encoding="UTF-8") as file:
+                service_type: dict = json.load(file)
+
+        except FileNotFoundError:
+            raise Exception()
```
nahuelverdugo, line 517:
```suggestion
        except FileNotFoundError as err:
            raise Exception(f"File {file_path} not found.") from err
```

ingestion/src/metadata/ingestion/source/metadata/amundsen.py
```diff
@@ -495,7 +503,31 @@ def get_type_primitive_type(self, data_type):
         return data_type
 
     def get_database_service(self, service_name: str) -> DatabaseService:
-        service = self.metadata.get_by_name(entity=DatabaseService, fqn=service_name)
+        """
+        Method to get and create Database Service
+        """
+        try:
+            file_path = (
+                Path(__file__).cwd() / "examples/amundsen_data/service_type.json"
```
pmbrull, line 511: is this the best path to place the file? Looks like it is required, maybe we should ship it differently. We might be adding the examples directory in the CI docker build, but I am not sure we are passing this on the release docker image.

ingestion/src/metadata/ingestion/source/metadata/amundsen.py
pmbrull, line 517: If we are dying due to `FileNotFoundError`, let's raise it with proper context. E.g.,

```python
except FileNotFoundError as err:
    raise FileNotFoundError("could not get service mapping file from blablabla")
```

ingestion/src/metadata/ingestion/source/metadata/amundsen.py
pmbrull: also, if it is a static file. Instead of loading it should we just have a constant holding the contents in a dict or something?

nahuelverdugo, line 511: I think we have handled the path to the resources differently in other parts of the code like [this](https://github.com/open-metadata/OpenMetadata/blob/a7347e1873d3b23251a02d262dcd066158bfa3c5/ingestion/src/metadata/utils/workflow_output_handler.py#L47) one.

ingestion/src/metadata/ingestion/source/metadata/amundsen.py
NiharDoshi99: Yes will add it.

ingestion/src/metadata/ingestion/source/metadata/amundsen.py
NiharDoshi99: Here actually I am getting the value of `Path(__file__) .parent.parent` inside the env , so did these . Need to put like 10 parent to make to actual path.

pmbrull: cc @ulixius9

pmbrull (approved):

pmbrull (dismissed):

ulixius9 (approved):

PR #8562, opened 2022-11-07
Fix #8540 Session ID does not reset for web analytic

### LINKED ISSUE #8540: Session ID does not reset for web analytic
We notice 2 things with the sessionId in the web analytic event that is sent:
1. the same sessionId is passed for all users -> to reproduce create different users and check the session ID passed for each user in the request.
2. the session ID does not expire. what we expect -> if a user is not active for 30 minutes then the session ID should reset. Example: a user comes at 12:00 (session ID set to 1234) visits different pages, then leaves the website. Comes back at 12:15 (session ID is still 1234) does other things on the platform leaves again and comes back at 1:00PM (new session ID should be set 56789).

Currently a new session ID is set only if different browsers are being used.

Sachin-chaurasiya:
Closes #8540 
### Describe your changes :

I worked on issue #8540, which includes.

- Default we will set the expiry of 30 minutes
- If the user is inactive for 30 minutes then the session will expire and a new session will be created.
- If the user is active and interacting with pages then we will reset the expiration by extending the expiry to the next 30 minutes from the current timestamp.

Example :
Let's say the expiry is at "5:45:23 PM", and the user spent some time and then Interact with other pages in 2 minutes so the expiry time will be extended to "5:47:23 PM".
In case of inactivity expiry will remain at "5:45:23 PM".

cc: @harshach @TeddyCr 

#
### Type of change :

- [x] Improvement

#
</p>

#
### Checklist:

#
### Reviewers
@open-metadata/ui

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/package.json
+++ b/openmetadata-ui/src/main/resources/ui/package.json
@@ -12,6 +12,7 @@
     "directory": "openmetadata-ui/src/main/resources/ui"
   },
   "dependencies": {
+    "@analytics/session-utils": "^0.1.17",
     "@ant-design/icons": "^4.7.0",
     "@apidevtools/json-schema-ref-parser": "^9.0.9",
     "@auth0/auth0-react": "^1.9.0",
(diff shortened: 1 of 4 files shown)


ShaileshParmar11 (approved):
Looks good!

chirag-madlani (approved):

PR #8623, opened 2022-11-09
Fixes #8135 - Implement partitioning config for profiler

### LINKED ISSUE #8135: Allow Profiler to run based on a partition
## Context
The profiler can currently run against sample data, though in some circumstances users may want to run a against specific partition range

## Solution
On the metadata ingestion side we pick up the partitionning of table which is then passed to the table entity. We need to 
1. add logic in the profiler workflow + UI to enable profiler to run by partition (will require to also specify the partition range)
2. if the database does not support partitionning natively, allow users to define a partition field directly from the platform

TeddyCr:
### Describe your changes :
Fixes #8135

We have introduced a `partitioning` key in the profilerConfig allowing user to define a profiler config.

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

#


TeddyCr:
waiting for #8619 to be merge as bullseye is failling Python build

ingestion/src/metadata/interfaces/sqalchemy/mixins/sqa_mixin.py
```diff
@@ -93,30 +87,41 @@ def set_session_tag(self, session) -> None:
...
-                "partition_start": start,
-                "partition_end": end,
-                "partition_values": partition_config.partitionValues,
-            }
-
-        return None
+        if not partition_config:
+            return None
+
+        timedelta_mapping = {
```
pmbrull, line 103: nit - this could be a constant at the module

ingestion/src/metadata/orm_profiler/api/workflow.py
```diff
@@ -220,34 +220,60 @@ def get_profile_query(self, entity: Table) -> Optional[str]:
...
+        if (
+            hasattr(entity, "tableProfilerConfig")
+            and hasattr(entity.tableProfilerConfig, "partitioning")
+            and entity.tableProfilerConfig.partitioning
+        ):
+            return entity.tableProfilerConfig.partitioning
+
+        if (
+            hasattr(entity, "serviceType")
+            and entity.serviceType == DatabaseServiceType.BigQuery
```
pmbrull, line 243: do we have a followup issue to generalise this? Instead of handling it only for BQ cc @ulixius9

ingestion/src/metadata/test_suite/api/workflow.py
```diff
@@ -219,30 +222,57 @@ def _get_profile_query(self, entity: Table) -> Optional[str]:
 
         return None
 
-    def _get_partition_details(self, entity: Table) -> Optional[TablePartitionConfig]:
+    def _get_partition_details(
```
pmbrull, line 225: should we have a PartitionsMixin to centralise the partitioning details? I see we are repeating the logic for profiler and Test Suite

pmbrull (approved):
Thanks for the changes @TeddyCr, left some comments if you have the time. Otherwise let's open a ticket and tackle it later

ingestion/src/metadata/test_suite/api/workflow.py
TeddyCr: I am actually creating a util function. It does not even use `self`

ingestion/src/metadata/orm_profiler/api/workflow.py
TeddyCr: what do you mean? The partitioning config will be handled for all type of entities. We need that check as some BQ table will require partitioning (whether it is set or not in the config).

ingestion/src/metadata/test_suite/api/workflow.py
pmbrull: oh that's right, then we can directly move it out

ingestion/src/metadata/orm_profiler/api/workflow.py
pmbrull: sorry I think I got confused, forgot about the special BQ cols. I said nothing

pmbrull (approved):

Issue #8553, opened 2022-11-07
Parse Avro/Protobuf/JsonSchema in Ingestion


OnkarVO7:
**Feature**
Parse Avro/Protobuf/JsonSchema on ingestion side and pass the relevant data into the new APIs

PR #8654, opened 2022-11-10
Fix:#8553 Parse Avro/Protobuf/Json schemas

OnkarVO7:
### Describe your changes :

Fix https://github.com/open-metadata/OpenMetadata/issues/8553

#
### Type of change :

- [x] New feature

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion


ingestion/src/metadata/utils/avro_parser.py
```diff
@@ -0,0 +1,35 @@
+#  Copyright 2021 Collate
```
pmbrull, line 1: Instead of putting all these in utils, let's create a module `parsers`. We should try to start cleaning things out from `utils` as much as possible into some names that have an actual meaning 😬

ingestion/src/metadata/utils/avro_parser.py
```diff
@@ -0,0 +1,35 @@
...
+import traceback
+
+import avro.schema as avroschema
+
+from metadata.utils.logger import ingestion_logger
+
+logger = ingestion_logger()
+
+
+def parse_avro_schema(schema):
```
pmbrull, line 25: can you please add the signature here? Both schema type and return type. For these functions that might be tricky, as we have different packages going on, typing can help us understand what we are doing

ingestion/src/metadata/utils/protobuf_parser.py
```diff
@@ -0,0 +1,204 @@
...
+import grpc_tools.protoc
+from pydantic import BaseModel
+
+from metadata.utils.helpers import snake_to_camel
+from metadata.utils.logger import ingestion_logger
+
+logger = ingestion_logger()
+
+
+class ProtobufDataTypes(Enum):
```
pmbrull, line 35: cool!

ingestion/src/metadata/utils/protobuf_parser.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+class ProtobufParserConfig(BaseModel):
+    """
+    Protobuf Parser Config class
+    """
+
+    schema_name: str
+    schema_text: str
+    base_file_path: Optional[str] = "/tmp/protobuf_openmetadata"
```
pmbrull, line 87: let's document why this is needed and how this is used. It can be here itself in the docstring

pmbrull: otherwise magic paths can be tricky

ingestion/src/metadata/utils/protobuf_parser.py
```diff
@@ -0,0 +1,204 @@
...
+        module_path = module
+        return __import__(module_path, fromlist=[module])
+
+    def create_proto_files(self):
+        """
+        Method to generate the protobuf directory and file structure
+        """
+        try:
+            # Create a temporary directory for saving all the files if not already present
+            if not os.path.exists(self.generated_src_dir):
```
pmbrull, line 115: instead of the `os` module, let's stick to `pathlib`. We should be able to achieve the same, but that's what we use in other places of the code.

ingestion/src/metadata/utils/protobuf_parser.py
```diff
@@ -0,0 +1,204 @@
...
+            )
+
+            # import the python file
+            sys.path.append(self.generated_src_dir)
+            py_file = glob.glob(
+                os.path.join(
+                    self.generated_src_dir, f"{self.config.schema_name}_pb2.py"
+                )
+            )[0]
+            module_name = pathlib.Path(py_file).stem
```
pmbrull, line 155: aha you're already using it here hehehehe!

pmbrull (commented):
Thanks for the work! Nice exploration for protobuf. Left a few comments if you have the time.

General comment for the tests would be to try to cover as many types as possible and validate some "wrong" cases. E.g., malformed input, otherwise we are just handling generic Exception instead of having a controlled catalog of what can go wrong. For this, we can open an issue and tackle it later.

Thanks!

ingestion/src/metadata/utils/avro_parser.py
OnkarVO7: Should I create the it in the same level as the utils directory?

ingestion/src/metadata/utils/avro_parser.py
OnkarVO7: I'll add it 👍

ingestion/src/metadata/utils/protobuf_parser.py
OnkarVO7: Yes I'll add in the docstring

ingestion/src/metadata/utils/avro_parser.py
OnkarVO7: Created it in the same level as utils directory

pmbrull (approved):

OnkarVO7:
https://github.com/open-metadata/OpenMetadata/issues/8669
@pmbrull opened an issue for tests improvement

PR #8655, opened 2022-11-10
cast to bigint to avoid overflow when summing

### LINKED ISSUE #8430: SQLServer Profiler throw overflow error
The sum function will use the field dtype as the dtype of the returned value. If this value is bigger than the value field can hold an overflow error will be thrown. 

### Solution
Similar to what we implemented for BigQuery, we'll need to force the field dtype to bigint in the sum function
https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/orm_profiler/orm/functions/sum.py

sfirke:
### Describe your changes :
Fixes #8430

I'm experiencing this overflow often in my MSSQL logs, often when an ID value or other large int value is summed.  This increases the overflow threshold for `sum()` from 2 billion (2^31) to 9 quintillion (2^63).

Most database variants are affected by this (see the issue for specifics) so I made this the generic function and then created a variant-specific version that doesn't cast, for Snowflake and Vertica.  That may not be necessary if the CAST from int to bigint would be ignored but I couldn't find out and don't have a way to test.

#
### Type of change :

- [x] Bug fix

#
### Checklist:

#
### Reviewers
@TeddyCr

### DIFF
--- a/ingestion/src/metadata/orm_profiler/orm/functions/sum.py
+++ b/ingestion/src/metadata/orm_profiler/orm/functions/sum.py
@@ -42,6 +42,7 @@ def _(element, compiler, **kw):
     proc = compiler.process(element.clauses, **kw)
     return f"SUM(CAST({proc} AS NUMERIC))"
 
+
 @compiles(SumFn, Dialects.Snowflake)
 @compiles(SumFn, Dialects.Vertica)
 def _(element, compiler, **kw):
--- a/ingestion/src/metadata/orm_profiler/orm/functions/sum.py
+++ b/ingestion/src/metadata/orm_profiler/orm/functions/sum.py
@@ -31,9 +31,9 @@ class SumFn(GenericFunction):
 
 @compiles(SumFn)
 def _(element, compiler, **kw):
-    """Handle case for empty table. If empty, clickhouse returns NaN"""
+    """Cast to BIGINT to address overflow error from summing 32-bit int in most database dialects, #8430"""
     proc = compiler.process(element.clauses, **kw)
-    return f"SUM({proc})"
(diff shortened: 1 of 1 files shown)


sfirke:
I didn't see where to test but if there's a metadata db that is tested against and someone wants to add an int column that sums to > 2^31, that could be useful for testing out new databases as they get added.

pmbrull (approved):
Thanks for handling this @sfirke, nice catch and elegant solution.

pmbrull:
@sfirke If possible just run `make py_format` from the project root to clean up the python checktyle. If you have any issues with that we can commit the clean directly here and merge. Thanks!

sfirke:
If you could run that `make py_format` on this branch that would be great.  I don't have a Linux environment today, I tried installing `make` on two Windows machines and it didn't go well 😅   Thanks Pere!

pmbrull:
> 

hehe the Windows curse. Ofc let me handle this. Thanks again

ingestion/src/metadata/orm_profiler/orm/functions/sum.py
```diff
@@ -31,13 +31,20 @@ class SumFn(GenericFunction):
...
-    return f"SUM({proc})"
+    return f"SUM(CAST({proc} AS BIGINT))"
 
 
 @compiles(SumFn, Dialects.BigQuery)
 def _(element, compiler, **kw):
     """Handle case where column type is INTEGER but SUM returns a NUMBER"""
     proc = compiler.process(element.clauses, **kw)
     return f"SUM(CAST({proc} AS NUMERIC))"
+
```
pmbrull, line 44:
```suggestion


```

pmbrull (approved):

PR #8707, opened 2022-11-14
Mapping from kpi to dichart(One to one)

mohityadav766:
### Describe your changes :

Added check to allow only creation of single Api per chart

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

#

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/KpiRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/KpiRepository.java
@@ -25,6 +25,7 @@ import org.openmetadata.schema.type.FieldChange;
 import org.openmetadata.schema.type.Include;
 import org.openmetadata.schema.type.Relationship;
 import org.openmetadata.service.Entity;
+import org.openmetadata.service.exception.CustomExceptionMessage;
 import org.openmetadata.service.exception.EntityNotFoundException;
 import org.openmetadata.service.resources.kpi.KpiResource;
 import org.openmetadata.service.util.EntityUtil;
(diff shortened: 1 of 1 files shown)


openmetadata-service/src/test/java/org/openmetadata/service/resources/kpi/KpiResourceTest.java
```diff
@@ -44,6 +45,7 @@
 
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 @Slf4j
+@Disabled
```
nahuelverdugo, line 48: Add an issue for enabling it back?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/KpiRepository.java
```diff
@@ -169,6 +171,14 @@ private EntityReference getDataInsightChart(Kpi kpi) throws IOException {
     return getToEntityRef(kpi.getId(), Relationship.USES, DATA_INSIGHT_CHART, true);
   }
 
+  public void validateDataInsightChartMapping(UUID chartId) {
+    // Each Chart has one unique Kpi mapping
+    List<CollectionDAO.EntityRelationshipRecord> record = findFrom(chartId, DATA_INSIGHT_CHART, Relationship.USES, KPI);
+    if (record.size() > 0) {
```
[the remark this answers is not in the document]
mohityadav766: if particular api has any chart associated that means size > 0 , then more kpi cannot be created for that charts, but if there are no chart to kpi mapping that is in case record == 0 , we should be able to create kpis, updating the function name

openmetadata-service/src/test/java/org/openmetadata/service/resources/kpi/KpiResourceTest.java
mohityadav766: almost every create fails currently , since internally test cases extending the EntityResource calls the createEntity , and since we have one to one mapping between chart and api , this will need some handling , will fix those later

nahuelverdugo (approved):

PR #8755, opened 2022-11-15
Fix race condition between Authorizer and BotResource

nahuelverdugo:
### Describe your changes :
`BotResource` depended on `DefaultAuthorizer` for adding the roles to the bot's users, but `DefaultAuthorizer` initialization happened after `BotResource` ones. Due to this bug, the `EntityUpdater` class instantiation failed with an NPE when the updating user of the updated entity was not an `admin` since the `SubjectCache` instance was initialized by the `DefaultAuthorizer`.

### Type of change :
- [x] Bug fix

### Checklist:

### Reviewers

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/PolicyCache.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/PolicyCache.java
@@ -24,37 +24,37 @@ import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import javax.annotation.CheckForNull;
 import lombok.extern.slf4j.Slf4j;
-import org.jdbi.v3.core.Jdbi;
 import org.openmetadata.schema.entity.policies.Policy;
 import org.openmetadata.schema.entity.policies.accessControl.Rule;
+import org.openmetadata.service.Entity;
 import org.openmetadata.service.exception.EntityNotFoundException;
-import org.openmetadata.service.jdbi3.CollectionDAO;
 import org.openmetadata.service.jdbi3.EntityRepository;
-import org.openmetadata.service.jdbi3.PolicyRepository;
 import org.openmetadata.service.util.EntityUtil.Fields;
 
 /** Subject context used for Access Control Policies */
 @Slf4j
 public class PolicyCache {
-  private static final PolicyCache INSTANCE = new PolicyCache();
+  private static PolicyCache INSTANCE = null;
   private static volatile boolean INITIALIZED = false;
 
   protected static LoadingCache<UUID, List<CompiledRule>> POLICY_CACHE;
   private static EntityRepository<Policy> POLICY_REPOSITORY;
   private static Fields FIELDS;
 
-  public static PolicyCache getInstance() {
-    return INSTANCE;
-  }
-
-  /** To be called during application startup by Default Authorizer */
-  public static void initialize(Jdbi jdbi) {
+  private PolicyCache() {
     if (!INITIALIZED) {
       POLICY_CACHE = CacheBuilder.newBuilder().maximumSize(100).build(new PolicyLoader());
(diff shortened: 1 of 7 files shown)


mohityadav766 (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/PolicyCache.java
```diff
@@ -55,6 +50,13 @@ public static void initialize() {
     }
   }
 
+  public static PolicyCache getInstance() {
```
sureshms, line 53: Is the initialization done only once during system startup? If so, let's document it. Otherwise, this will have concurrency issues.

openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/RoleCache.java
```diff
@@ -31,24 +31,22 @@
...
 
-  /** To be called only once during the application start from DefaultAuthorizer */
-  public static void initialize() {
-    if (!INITIALIZED) {
-      ROLE_CACHE = CacheBuilder.newBuilder().maximumSize(100).build(new RoleLoader());
-      ROLE_REPOSITORY = Entity.getEntityRepository(Entity.ROLE);
-      FIELDS = ROLE_REPOSITORY.getFields("policies");
-      INITIALIZED = true;
+  public static RoleCache getInstance() {
+    if (INSTANCE == null) {
```
sureshms, line 46: Same comment as befoe

harshach:
@nahuelverdugo isn't DefaultAuthorizer starting ahead of all the resources? during the server start

nahuelverdugo:
@harshach no, it is initialized [after](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplication.java#L166) the resources. I think we should take out from the resources the initialization of the system objects. For example, we initialize the Organization in the TeamResource [class](https://github.com/open-metadata/OpenMetadata/blob/185872ca0868993ee70992a1fe2fdfccd78189c0/openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/TeamResource.java#L96) and we have a dependency on it on the UserRepository [here](https://github.com/open-metadata/OpenMetadata/blob/185872ca0868993ee70992a1fe2fdfccd78189c0/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/UserRepository.java#L62).

harshach:
@nahuelverdugo please look at the line 130 that's where we are creating the authorizer https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplication.java#L130 ( register authorizer)
and on line 150 registerResources https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplication.java#L153

mohityadav766:
hi @harshach 

1. It does get the constructor call before the resources init , but the initialize() method that inits the SubjectCache(), PolicyCache() etc is after the registering of resources , so before that if anywhere we use Entity.getEntityRepository(entity_name), it will always be null , the only way to use the repository before the resource registration is using jdbi object as example new UserRepository(jdbi.onDemand(CollectionDAO.class))
2. We do have a order that we follow using order in @CollectionAnnotation , but one always has to be take into account the order of initlization (PolicyRpository > RoleRepository > TeamRepository > userRepository > BotRepository)
3. user Repository is dependent organisation entry in DB , it can never be used before that entry is present in DB,  it throws exception otherwise "Entity Organisation not found"

harshach:
@mohityadav766 @nahuelverdugo if we initialize the cache instances after the resources initialize that would resolve it ?

Issue #8741, opened 2022-11-14
Teams link from the main page links to the wrong page


stevenspadotto:
**Affected module**

UI

**Describe the bug**

On the main page, Teams count is `1` but then when I click on Teams, it takes me to a page where the count is `0`.

It seems that the Teams link from the main page links to `/settings/members/teams/Organization` instead of linking to
`/settings/members/teams`.

**To Reproduce**

<img width="401" alt="Screenshot 2022-11-14 at 5 30 32 PM" src="https://user-images.githubusercontent.com/97744515/201785559-8acf074f-b529-4804-b083-981acd149465.png">

<img width="756" alt="Screenshot 2022-11-14 at 5 30 39 PM" src="https://user-images.githubusercontent.com/97744515/201785569-bb4f97d2-865d-42da-b7fa-799422832ab1.png">

**Expected behavior**

Teams link from the main page should link to `/settings/members/teams`.

**Version:**
 - OS: OS X
 - Python version:
 - OpenMetadata version: 0.12.2
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

chirag-madlani:
Hi, @stevenspadotto Thanks for reporting.
We will be having `/settings/members/teams/Organization` as our home path for the Teams page as all teams must fall under Organization
I did remove the older path and UI in this [PR](https://github.com/open-metadata/OpenMetadata/pull/8767).
Should be fixed with 0.13.0

cc: @harshach

stevenspadotto:
@chirag-madlani I see. Ok, thanks. However, on the main page, I see Teams count as `1`, but when I click on the Teams link from there, it takes me to the Organization page where I see Teams count as `0`. This discrepancy with the count can appear to be confusing.

harshach:
@chirag-madlani it looks like we are not showing the total count of all teams but instead showing the count at the top level example
<img width="1196" alt="image" src="https://user-images.githubusercontent.com/38649/201979988-101d34f8-2b68-467d-8f54-d4196313f4d6.png">
<img width="334" alt="image" src="https://user-images.githubusercontent.com/38649/201980022-16ed0ef4-2788-4dbf-baaa-185acc7fec45.png">

chirag-madlani:
[...]

PR #8767, opened 2022-11-15
fix(ui): remove teams root route and redirect it to Organization

chirag-madlani:
### Describe your changes :

I worked on the fixing #8741

#
### Type of change :

- [x] Bug fix

#
Screenshot 2022-11-15 at 4 42 18 PM

#
### Checklist:

#

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/TeamDetails/TeamDetailsV1.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/TeamDetails/TeamDetailsV1.tsx
@@ -170,7 +170,7 @@ const TeamDetailsV1 = ({
     TitleBreadcrumbProps['titleLinks']
   >([]);
   const [addAttribute, setAddAttribute] = useState<AddAttribute>();
-  const [loading, setLoading] = useState<boolean>(true);
+  const [loading, setLoading] = useState<boolean>(false);
   const [selectedEntity, setEntity] = useState<{
     attribute: 'defaultRoles' | 'policies';
     record: EntityReference;
--- a/openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
+++ b/openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
@@ -275,7 +275,7 @@
     "data-insight-chart": "Data insight chart",
     "data-assets": "Data Assets",
     "app-analytics": "App Analytics",
-    "no-team-found": "No team found with name given.",
+    "no-team-found": "No team found.",
     "quality": "Quality",
     "insight": "Insight",
     "govern": "Govern"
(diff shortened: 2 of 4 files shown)


openmetadata-ui/src/main/resources/ui/src/components/TeamDetails/TeamDetailsV1.tsx
```diff
@@ -170,7 +170,7 @@ const TeamDetailsV1 = ({
     TitleBreadcrumbProps['titleLinks']
   >([]);
   const [addAttribute, setAddAttribute] = useState<AddAttribute>();
-  const [loading, setLoading] = useState<boolean>(true);
+  const [loading, setLoading] = useState<boolean>(false);
```
Sachin-chaurasiya, line 173: @chirag-madlani, Why we are changing the default initial state?

openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
```diff
@@ -275,6 +275,7 @@
     "data-insight-chart": "Data insight chart",
     "data-assets": "Data Assets",
     "app-analytics": "App Analytics",
+    "no-team-found": "No team found with name given.",
```
Sachin-chaurasiya, line 278:
```suggestion
    "no-team-found": "No team found with the {{name}}.",
```

openmetadata-ui/src/main/resources/ui/src/components/TeamDetails/TeamDetailsV1.tsx
chirag-madlani:
```
const fetchPermissions = async () => {
    setLoading(true);
    try {
      const perms = await getEntityPermission(
        ResourceEntity.TEAM,
        currentTeam.id
      );
      setEntityPermissions(perms);
    } catch (error) {
      showErrorToast(
        error as AxiosError,
        t('message.entity-fetch-error', {
          entity: 'User Permissions',
        })
      );
    } finally {
      setLoading(false);
    }
  };

  useEffect(() => {
    !isEmpty(currentTeam) && fetchPermissions();
  }, [currentTeam]);
```

As you can see above loading will be false only when fetchPermissions is called but is currentTeam is empty it will be end up in forever loading which isn't a good UX
That's why I changes it to default false

openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
chirag-madlani: Changing it to generic as "No team found" as we might not name with us

chirag-madlani:
@Sachin-chaurasiya comments addressed

openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
```diff
@@ -275,6 +275,7 @@
     "data-insight-chart": "Data insight chart",
     "data-assets": "Data Assets",
     "app-analytics": "App Analytics",
+    "no-team-found": "No team found.",
```
Sachin-chaurasiya, line 278: @chirag-madlani , I would recommend to show team name as well.

Sachin-chaurasiya (approved):

PR #8771, opened 2022-11-15
Add secrets manager documentation back

nahuelverdugo:
### Describe your changes :
Add back secrets manager documentation after refactoring.

### Type of change :
- [x] Documentation

### Checklist:

### Reviewers

### DIFF
--- a/openmetadata-docs/content/deployment/secrets-manager/index.md
+++ b/openmetadata-docs/content/deployment/secrets-manager/index.md
@@ -1,0 +1,87 @@
+---
+title: Enable Secrets Manager
+slug: /deployment/secrets-manager
+---
+
+# Enable Secrets Manager
+
+Secret Manager integrations allow you to use your existing third-party **Key Management Store** (KMS) with OpenMetadata. 
+Your credentials and sensitive information are stored in a tool that you control, and the KMS will mediate between any 
+OpenMetadata internal requirement and sensitive information.
+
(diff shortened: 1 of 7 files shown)


openmetadata-docs/content/deployment/secrets-manager/index.md
```diff
@@ -0,0 +1,87 @@
...
+1. The migration of all the sensitive data will be done automatically after restarting the OpenMetadata server, which 
+can not be undone for the time being.
+2. Only users with permissions can edit and retrieve the service connections. The connection parameters will be hidden 
+for all other users.
+
+## How it works
+
+There are two types of secrets manager implementations.
+
+### Managed secrets manager
```
pmbrull, line 61: Should we add a line explaining how to configure a managed vs a non-managed?

pmbrull: something like, any implementation can be used as both managed and non managed, the only difference is how the config gets passed

pmbrull (approved):

openmetadata-docs/content/deployment/secrets-manager/index.md
nahuelverdugo: It is explained in each implementation section (AWS SSM, AWS secrets manager, ...), and it is almost the same. Here we only explain how it works in each case since it is common for all the implementations.

PR #8814, opened 2022-11-16
Data insight tests addition

TeddyCr:
### Describe your changes :
Added additional tests for data insight:
- check data were correctly ingested to all indexes after workflow
- Check web analytics data endpoint can be correctly read and wrote to
- mocked admin user going through the site and confirm web analytics data are correctly wrote

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

#


pmbrull (commented):
Thanks @TeddyCr for adding the tests 🙏 

Just one thought, should we implement the test in cypress instead of adding another tool? it looks like playwright follows a similar approach. It would be interesting to keep these implementations centralised

TeddyCr:
> at

Would we send a request to the API to check the data in cypress as well? I looked at Cypress but there is no Python API unfortunately, only javascript -- which is not ideal for our team.

TeddyCr:
Actually it is failing so I'll remove it for now and think of another approach (maybe cypress then 😅)

pmbrull (approved):

PR #8815, opened 2022-11-16
On init the auth provider 'openmetadata' must generate a JWT token if not present on config

nahuelverdugo:
### Describe your changes :
Setting up for the first time auth mechanism for bots could fail in case that `openmetadata` is the auth provider but there is no configuration.

### Type of change :
- [x] Bug fix

### Checklist:

### Reviewers

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/security/DefaultAuthorizer.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/security/DefaultAuthorizer.java
@@ -13,6 +13,7 @@
 
 package org.openmetadata.service.security;
 
+import static org.openmetadata.common.utils.CommonUtil.nullOrEmpty;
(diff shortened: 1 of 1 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/security/DefaultAuthorizer.java
```diff
@@ -334,6 +328,14 @@ public static User addOrUpdateBotUser(User user, OpenMetadataApplicationConfig o
     return addOrUpdateUser(user);
   }
 
+  private static JWTAuthMechanism buildJWTAuthMechanism(OpenMetadataJWTClientConfig jwtClientConfig, User user) {
+    return jwtClientConfig == null || jwtClientConfig.getJwtToken() == null
```
mohityadav766, line 332: we should add an "" check on the JWT token as well, by default we get airflowConfig as not null with empty configurations so it will initialise it with an empty token I think

mohityadav766 (approved):

PR #8851, opened 2022-11-17
Fix: ignored query with om-version and dbt

MilanBariya:
### Describe your changes :

I worked on the Fixing ignored query with om-version and dbt.   Fix `Identifier is expected, got Token[value: )] ` error while ingesting databricks lineage and usage.

#
### Type of change :

- [x] Bug fix
- [x] Improvement

#

### DIFF
--- a/ingestion/src/metadata/clients/databricks_client.py
+++ b/ingestion/src/metadata/clients/databricks_client.py
@@ -25,6 +25,8 @@ from metadata.utils.helpers import datetime_to_ts
 from metadata.utils.logger import ingestion_logger
 
 logger = ingestion_logger()
+QUERY_WITH_OM_VERSION = '/* {"app": "OpenMetadata"'
+QUERY_WITH_DBT = '/* {"app": "dbt"'
 
 
 class DatabricksClient:
@@ -113,3 +115,10 @@ class DatabricksClient:
             logger.error(exc)
 
         return query_details
+
+    def is_query_valid(self, row) -> bool:
+        query_text = row.get("query_text")
+        return not (
+            query_text.startswith(QUERY_WITH_DBT)
+            or query_text.startswith(QUERY_WITH_OM_VERSION)
+        )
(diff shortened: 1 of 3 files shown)


nahuelverdugo (commented):
I think we are repeating ourself in 2 different parts of the code. Could we move this into a common place at least so in the future is easy to change in case the comment change or we want to ignore some other queries? For example, we can put it [here](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/utils/helpers.py).

pmbrull:
> I think we are repeating ourselves in 2 different parts of the code. Could we move this into a commonplace at least so in the future is easy to change in case the comment change or we want to ignore some other queries? For example, we can put it [here](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/utils/helpers.py).

let's try to avoid `utils` and `helpers` from now on, it's becoming a free directory for random stuff. We can create a module with some meaning for databricks or add it directly in the databricks client

ingestion/src/metadata/clients/databricks_client.py
```diff
@@ -113,3 +115,10 @@ def list_query_history(self, start_date=None, end_date=None) -> List[dict]:
             logger.error(exc)
 
         return query_details
+
+    def is_query_valid(self, row) -> bool:
+        query_text = row.get("query_text")
+        return not (
+            query_text.startswith(QUERY_WITH_DBT)
+            or query_text.startswith(QUERY_WITH_OM_VERSION)
+        )
```
nahuelverdugo, line 124: If it doesn't apply only to Databricks could we have it in a utils package? What do you think? 
nit: doc on the method

nahuelverdugo (approved):

ingestion/src/metadata/clients/databricks_client.py
MilanBariya: this only applies to databricks as of now for other connectors we are filtering via qureries.

nahuelverdugo (approved):

nahuelverdugo (approved):

PR #8876, opened 2022-11-18
Added sleep between data processing ingestion and KPI processing

TeddyCr:
### Describe your changes :

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/data_insight/api/workflow.py
+++ b/ingestion/src/metadata/data_insight/api/workflow.py
@@ -19,6 +19,7 @@ Workflow definition for the ORM Profiler.
 
 from __future__ import annotations
 
+import time
 import traceback
 from datetime import datetime
 from typing import Optional, Union, cast
@@ -266,6 +267,7 @@ class DataInsightWorkflow:
         logger.info("Data processor finished running")
 
         logger.info("Sleeping for 1 second. Waiting for ES data to be indexed.")
+        time.sleep(1)
         logger.info("Starting KPI runner")
         self._execute_kpi_runner()
         logger.info("KPI runner finished running")


ingestion/src/metadata/data_insight/api/workflow.py
```diff
@@ -266,6 +267,7 @@ def execute(self):
         logger.info("Data processor finished running")
 
         logger.info("Sleeping for 1 second. Waiting for ES data to be indexed.")
+        time.sleep(1)
```
pmbrull, line 270: instead of sleeping, is there any query we can run to validate? Just thinking that this can not be as robust as "checking the timestamp we expect to have" or something similar. Then we can loop and try a couple of times or else raise an exception

TeddyCr: We are looping already, to check if we have data in ES. We saw that in some instances we will pick up partial data. We see data are in ES so we excit the while loop, though only partial data are there. 

https://github.com/open-metadata/OpenMetadata/blob/ed0d3ada3bd601a9bac5f4b17f8c98393bed6ea7/ingestion/src/metadata/data_insight/runner/kpi_runner.py#L82-L92

pmbrull: Oh I see, thanks for the explanation. I guess that we get an OK from a single record but we could be missing others from the same time 🤔 

Being part of the `execute` I guess we can keep this second here as a safety measure. Let's see if we can figure out something else down the road.

Thanks

pmbrull (approved):

PR #8896, opened 2022-11-18
Deprecate configurations for auth provider configuration on Airflow

nahuelverdugo:
### Describe your changes :
Deprecate the use of the configuration of the bot auth mechanism in the YAML configuration of the OM server. It still allows using it but does not force the user to remove it. 

### Type of change :
- [x] Improvement

### Checklist:

### Reviewers


conf/openmetadata.yaml
```diff
@@ -181,32 +181,6 @@ airflowConfiguration:
   username: ${AIRFLOW_USERNAME:-admin}
   password: ${AIRFLOW_PASSWORD:-admin}
   metadataApiEndpoint: ${SERVER_HOST_API_URL:-http://localhost:8585/api}
-  authProvider: ${AIRFLOW_AUTH_PROVIDER:-"no-auth"} # Possible values are "no-auth", "azure", "google", "okta", "auth0", "custom-oidc", "openmetadata"
```
harshach, line 184: we don't need authProvider for service accounts any more?

nahuelverdugo: No, we get it now from the Bot auth mechanism.

TeddyCr:
@nahuelverdugo, just wondering what is the status of this PR?

nahuelverdugo:
@harshach can we merge these changes? I think we can stop supporting these configs in 1.0.

harshach:
@nahuelverdugo as long as we don't need any migrations lets merge in. My understanding is if the configs are there we won't throw an error. What happens to existing SSO based service accounts will they work?

nahuelverdugo:
> @nahuelverdugo in 1.0 we can even kill any client auth that's not JWT

@pmbrull should I create a ticket for that?

> @nahuelverdugo as long as we don't need any migrations lets merge in. My understanding is if the configs are there we won't throw an error. What happens to existing SSO based service accounts will they work?

@harshach, we are still supporting them so we won't throw an error. At the start of the application, we check if bots have auth config. If they don't have it, the service accounts configured in the YAML configuration file will be used as the default auth config for bots and stored in the DB. Otherwise, we will use the JWT token config.

nahuelverdugo:
@akash-jain-10 should we remove the Global Chart Values from Helm `global.airflow.openmetadata.authConfig.*` and `global.airflow.openmetadata.authProvider` too?

harshach (approved):

Issue #8489, opened 2022-11-02
Edit display name for ingestion pipeline doesn't work


ulixius9:
**Affected module**
UI

**Describe the bug**
Edit display name for ingestion pipeline doesn't work

https://user-images.githubusercontent.com/39544459/199406571-7a4eac24-8ca5-48a6-8fe5-167ff183dbb4.mov

ulixius9:
`name` of ingestion pipeline can not be changed and editing the `displayName` might create confusion between pipeline name visible on UI and DAG name available on airflow will be different so it's better if we disable editing the name!

PR #9023, opened 2022-11-28
UI : - Disable the name field file while editing Ingestion

Ashish8689:
### Describe your changes :

- Closes : #8489
- Disable the name field file while editing Ingestion

#
### Type of change :

- [x] Bug fix

#

https://user-images.githubusercontent.com/66266464/204250200-725fdb2d-8378-4995-b264-3ab987d9f795.mov

#
### Checklist:

#
### Reviewers

@open-metadata/ui

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Ingestion/Ingestion.component.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Ingestion/Ingestion.component.tsx
@@ -407,9 +407,9 @@ const Ingestion: React.FC<IngestionProps> = ({
     () => [
       {
         title: t('label.name'),
-        dataIndex: 'name',
-        key: 'name',
-        render: (text) =>
+        dataIndex: 'displayName',
+        key: 'displayName',
+        render: (text, record) =>
           airflowEndpoint ? (
             <Tooltip
               title={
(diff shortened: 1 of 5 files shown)


openmetadata-ui/src/main/resources/ui/src/components/Ingestion/Ingestion.component.tsx
```diff
@@ -421,7 +421,7 @@ const Ingestion: React.FC<IngestionProps> = ({
                 className="tw-mr-2"
                 data-testid="airflow-tree-view"
                 disabled={!(permissions.ViewAll || permissions.ViewBasic)}
-                href={`${airflowEndpoint}/tree?dag_id=${text}`}
+                href={`${airflowEndpoint}/tree?dag_id=${record.name}`}
```
ShaileshParmar11, line 424: It might create confusion because in airflow it will be displayed with name and in OM it will be shown with displayName, and if both are the different users will be confused, we must keep the same name at both places!
cc @chirag-madlani what do you think?

Ashish8689: @ShaileshParmar11 yes, disable the `name` field while editing.

chirag-madlani (approved):

PR #9049, opened 2022-11-29
Fix #9008 - Add Airflow callback, introduce Workflow Status Mixin and remove retries

### LINKED ISSUE #9008: Killing an IngestionPipeline does not update the status
We might need to add a callback or something in the managed APIs to not lose the state on the kill. Review what happens when pipelines fail

pmbrull:
### Describe your changes :

Fix #9008 

- Pipelines won't retry anymore.
- Simplify status logic by introducing a mixin
- Reorganize managed APIs modules
- Moved success computation on the ingestion workflow out of the `stop` call, otherwise it was being marked as a success even if the pipeline failed.
- Handle failure updates in the Workflow class directly, not on the Airflow side. We want the workflows to be self-managed as much as possible (although we added the callback to be prepared for any big fire on Airflow's side that we cannot handle)

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


ingestion/src/metadata/workflow/workflow_status_mixin.py
```diff
@@ -0,0 +1,80 @@
...
+    PipelineState,
+    PipelineStatus,
+)
+from metadata.generated.schema.metadataIngestion.workflow import (
+    OpenMetadataWorkflowConfig,
+)
+from metadata.ingestion.ometa.ometa_api import OpenMetadata
+
+
+class WorkflowStatusMixin:
```
pmbrull, line 28: Using a mixin allows us to access directly the Workflow class and manage the required ingredients internally, instead of passing too many information to external methods.

ingestion/src/metadata/workflow/workflow_status_mixin.py
```diff
@@ -0,0 +1,80 @@
...
+    Helper methods to manage IngestionPipeline status
+    and workflow run ID
+    """
+
+    config: OpenMetadataWorkflowConfig
+    _run_id: Optional[str] = None
+    metadata: OpenMetadata
+
+    @property
+    def run_id(self) -> str:
```
pmbrull, line 39: It's cleaner to handle the run id setting logic once instead of computing every status call.

Also, the old `set_ingestion_pipeline_status` was returning a `str` of the pipeline run id, which was misleading

openmetadata-airflow-apis/openmetadata_managed_apis/workflows/ingestion/common.py
```diff
@@ -352,12 +285,18 @@ def build_dag(
...
         PythonOperator(
             task_id=task_name,
             python_callable=workflow_fn,
             op_kwargs={"workflow_config": workflow_config},
-            retries=ingestion_pipeline.airflowConfig.retries,
-            retry_delay=ingestion_pipeline.airflowConfig.retryDelay,
+            # There's no need to retry if we have had an error. Wait until the next schedule or manual rerun.
+            retries=0,
+            # each DAG will call its own OpenMetadataWorkflowConfig
+            on_failure_callback=partial(send_failed_status_callback, workflow_config),
```
pmbrull, line 299: partial func allows us to pass DAG-specific params such as the workflow config. Otherwise only the `context` is available in callbacks

openmetadata-airflow-apis/openmetadata_managed_apis/workflows/ingestion/test_suite.py
```diff
@@ -29,6 +26,30 @@
     Sink,
     WorkflowConfig,
 )
+from metadata.ingestion.models.encoders import show_secrets_encoder
+from metadata.test_suite.api.workflow import TestSuiteWorkflow
+from metadata.utils.logger import set_loggers_level
+
+
+def test_suite_workflow(workflow_config: OpenMetadataWorkflowConfig):
```
pmbrull, line 34: moved each of these outside of commons and inside their own module. The only common one is the ingestion, used for metadata, lineage, usage and ES reindex

openmetadata-airflow-apis/openmetadata_managed_apis/workflows/ingestion/profiler.py
```diff
@@ -36,6 +37,27 @@
...
+    This is the callable used to create the PythonOperator
+    """
+
+    set_loggers_level(workflow_config.workflowConfig.loggerLevel.value)
+
+    config = json.loads(workflow_config.json(encoder=show_secrets_encoder))
+    workflow = ProfilerWorkflow.create(config)
+
+    workflow.execute()
+    workflow.raise_from_status()
```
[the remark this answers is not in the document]
pmbrull: nice catch, completely agree. Taking a look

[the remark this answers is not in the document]
pmbrull: thanks for checking. Added a `_raise_from_status_internal` for each workflow and handled the `raise_from_status` directly from the mixin.

Thanks!

ulixius9 (approved):

harshach:
can we make sure retries are also removed for the profiler, where it can get expensive? I've seen few times one or other runtime exception triggers the whole workflow retry

pmbrull:
> can we make sure retries are also removed for the profiler, where it can get expensive? I've seen few times one or other runtime exception triggers the whole workflow retry

This removes the retries from all workflows. It's the only place where we create the Task

Issue #8010, opened 2022-10-07
Safer default Mark Deleted Tables


pmbrull:
currently if we enable that with filtered stuff, we mark everything filtered as deleted by default. The current behavior forces us to also toggle the Mark Deleted Tables from Filter Only to keep the metadata safe. We should change this behavior around. The default marking deleted behavior should be the "safe" one, only considering filter entities, and then force users to toggle a "Mark All Deleted Tables" or something if they really want to remove all the filtered data

harshach:
@pmbrull who is this going to be assigned?

PR #9065, opened 2022-11-29
Fix: Safer default Mark Deleted Tables

MilanBariya:
### Describe your changes :

I worked on the Safer default Mark Deleted Tables
Fix: #8010

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature

### Checklist:

#


ingestion/src/metadata/ingestion/source/database/database_service.py
```diff
@@ -533,7 +533,7 @@ def mark_tables_as_deleted(self):
             logger.info(
                 f"Mark Deleted Tables set to True. Processing database [{self.context.database.name.__root__}]"
             )
-            if self.source_config.markDeletedTablesFromFilterOnly:
+            if not self.source_config.markAllDeletedTables:
```
MilanBariya, line 536: `markAllDeletedTables` is the default false so this if block will run.
When the value is false that time else block will run.
(Once we ingest  database service without FilterPattern and after we use with FilterPattern, now it will not give 'Deleted' mark in other tabes )

pmbrull: Hi @MilanBariya this took me a while to understand, maybe because we are checking a `not bool` condition and there are not comments in the code.

Could we change the `if`s here so that:

```python
if self.source_config.markDeletedTables:
    ...
    if self.source_config.markAllDeletedTables:  # if enabled, delete all
        yield from self.fetch_all_schema_and_delete_tables()
    else:  # if not enabled (default) delete from schema
        schema_names_list = self.get_database_schema_names()
        ....
```

and add proper comments around the logic of each block?

Thanks

pmbrull (changes requested):

ingestion/src/metadata/ingestion/source/database/database_service.py
MilanBariya: Sure @pmbrull let me do that.
Thank you.

openmetadata-ui/src/main/resources/ui/src/components/AddIngestion/Steps/ConfigureIngestion.tsx
```diff
@@ -234,15 +234,15 @@ const ConfigureIngestion = ({
               {getSeparator('')}
             </Field>
           )}
-          {!isNil(markDeletedTablesFromFilterOnly) && (
+          {!isNil(markAllDeletedTables) && (
             <Field>
               <div className="tw-flex tw-gap-1">
                 <label>Mark Deleted Tables from Filter Only</label>
```
chirag-madlani, line 240: It should be 
`Mark All Deleted Tables`

chirag-madlani (changes requested):

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceMetadataPipeline.json
```diff
@@ -23,8 +23,8 @@
       "type": "boolean",
       "default": true
     },
-    "markDeletedTablesFromFilterOnly": {
-      "description": "Optional configuration to mark deleted tables only to the filtered schema",
+    "markAllDeletedTables": {
```
nahuelverdugo, line 26: If we rename a JSON field that we are storing in DB, we must add it to migration files.

nahuelverdugo (dismissed):

Sachin-chaurasiya (approved):

openmetadata-ui/src/main/resources/ui/src/components/AddIngestion/Steps/ConfigureIngestion.tsx
MilanBariya: Done

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceMetadataPipeline.json
MilanBariya: Done

pmbrull (approved):

Sachin-chaurasiya (approved):

Issue #6203, opened 2022-07-19
Query parser fails on Redshift COPY statements


Beetelbrox:
**Affected module**
Ingestion framework

**Describe the bug**
When we try to parse a [COPY](https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html) statement from Redshift the `source_tables` list on the parsed query object contains a `sqllineage.core.models.Path` instance with the S3 key of the copied file. Downstream code [expects this instance to be a table](https://github.com/open-metadata/OpenMetadata/blob/9308fe6df31d83895ba16e934ef6e69adba25881/ingestion/src/metadata/ingestion/processor/query_parser.py#L78-L85) and tries to access a non-existing `alias` field, crashing the parsing of the query with the error `AttributeError: 'Path' object has no attribute 'alias'`

**To Reproduce**
The error can be reproduced running the following code:
```python

from sqllineage.runner import LineageRunner
from metadata.ingestion.processor.query_parser import get_involved_tables_from_parser, get_parser_table_aliases

query = "COPY my_schema.my_table FROM 's3://bucket/path/object.csv';"

tables = get_involved_tables_from_parser(LineageRunner(query))
get_parser_table_aliases(tables)
```

**Expected behavior**
The query parser doesn't raise an error and the reference to the object is stored in some kind of entity. Not sure if we have some way of supporting this right now or if we should just filter out copy queries for now.

**Version:**
 - OS: iOS
 - Python version: 3.9.19
 - OpenMetadata version: 0.11.0
 - OpenMetadata Ingestion package version: 0.11.0

**Additional context**
This is part of the greater effort for optimizing redshift usage ingestion on #6162.

TeddyCr:
Thanks @Beetelbrox for the findings, this is very great feedback. I would say `COPY` statement could be skipped as they are primarly data manipulation statement and I feel like they provide little value for users (they could also poluate usage statistic in a sense). 

As a user, do you see any value in keeping `COPY` statements in usage?

harshach:
@nahuelverdugo shall we skip the erroring out just like MERGE INTO queries and close this out?

nahuelverdugo:
I would use the same approach we have for `MERGE INTO` and address this for `0.13.1`.

Issue #8364, opened 2022-10-26
Wrap `sqllineage` usage into a class


nahuelverdugo:
**Issue**
We should have the same behavior when calling the `LineageRunner` parser from different workflows and connectors.

For example:
We have calls directly to `LineageRunner` in [this](https://github.com/open-metadata/OpenMetadata/blob/a6822aa0942c841a6b8a811f29aea44388d220e0/ingestion/src/metadata/ingestion/source/dashboard/mode.py#L143) line and also [here](https://github.com/open-metadata/OpenMetadata/blob/a6822aa0942c841a6b8a811f29aea44388d220e0/ingestion/src/metadata/ingestion/source/dashboard/redash.py#L129). And probably in other places.

**Solution**
Wrap the usage of `sqllineage` in a class where we can align how we want to call it without coupling the use of this library to different parts of the code. 

An idea for this issue could be to create a util class:
```python
class LineageParser:
  
  @classmethod
  def create(sql_query: str) -> LineageRunner:
     return LineageRunner(clean_raw_query(query.get("raw_query")))
```
And replace calls like:
```python
table_list = LineageRunner(clean_raw_query(query.get("raw_query")))
``` 
with:
 ```python
parser = LineageParser.create(query.get("raw_query"))
```

ssandy83:
@nahuelverdugo 
Is this issue still open? Can I work on this?

nahuelverdugo:
Hi @ssandy83, thank you for your interest 🚀 ! Yes, it is still open. I will assign it to you, let me know if you have any doubts here on our [slack](https://slack.open-metadata.org/) channel.

pmbrull:
hi @ssandy83 any updates here? Thanks

ssandy83:
Hey @pmbrull 
I couldn't get to this yet. I shall be able to have an update by next week

nahuelverdugo:
Sorry @ssandy83, I hope you have not started to work on this issue, but it has been closed as part of another [issue](https://github.com/open-metadata/OpenMetadata/pull/9073) we were fixing. Feel free to reach out to us on Slack if you want to collaborate on another task :)

PR #9073, opened 2022-11-30
Fix#6203: Refactor LineageRunner use

nahuelverdugo:
Fixes: #6203 #8364

### Describe your changes :
- Refactor the use of the LineageRunner library into a class that leads with all the calls to the library. (#8364)
- Skipped `COPY table FROM path` queries when doing lineage as agreed on #6203 .

### Type of change :
- [x] Bug fix
- [x] Improvement

### Checklist:

### Reviewers


ingestion/src/metadata/ingestion/lineage/sql_lineage.py
```diff
@@ -23,62 +22,30 @@
...
+        ]  # pylint: disable=unbalanced-tuple-unpacking
         if database_schema == "<default>":
             database_schema = None
-    return {"database": database, "database_schema": database_schema, "table": table}
+        return {
+            "database": database,
+            "database_schema": database_schema,
+            "table": table,
+        }
+    return {"database": database, "database_schema": database_schema, "table": None}
```
ulixius9, line 48: I believe we are not using this function anywhere!

ingestion/src/metadata/ingestion/lineage/parser.py
```diff
@@ -37,214 +43,283 @@
...
 
-    return join_data
+        if insensitive_match(clean_query, ".*merge into .*when matched.*"):
+            clean_query = insensitive_replace(
+                raw_str=clean_query,
+                to_replace="when matched.*",  # merge into queries specific
+                replace_by="",  # remove it as LineageRunner is not able to perform the lineage
+            )
+
+        if insensitive_match(clean_query, "^COPY.*"):
```
pmbrull, line 322: is this the one for https://github.com/open-metadata/OpenMetadata/issues/6203? Could we add some info around the why in the comments itself pls?

ingestion/src/metadata/ingestion/lineage/parser.py
```diff
@@ -37,214 +43,283 @@
...
+
+    @cached_property
+    def clean_table_list(self) -> List[str]:
+        """
+        Clean the table name if it has <default>.
+        :return: clean table names
+        """
+        return [get_formatted_entity_name(str(table)) for table in self.involved_tables]
+
+    @cached_property
```
pmbrull, line 117: as discussed, this is quite beautiful but not supported on 3.7, so we'll need to manually handle the caching around here https://docs.python.org/3/library/functools.html#functools.cached_property

nahuelverdugo: Added a library to support it on Python 3.7

ingestion/src/metadata/ingestion/lineage/parser.py
nahuelverdugo: Comment added.

pmbrull (approved):

PR #9104, opened 2022-12-01
profileSample

ayush-shah:
### Describe your changes :

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceProfilerPipeline.json
```diff
@@ -35,10 +37,15 @@
       "default": true
     },
     "profileSample": {
-      "description": "Percentage of data used to execute the profiler for the whole workflow. This percentage will be applied to all tables in the workflow. Represented in the range (0, 100].",
+      "description": "Percentage of data we want to execute the profiler and tests on. Represented in the range (0, 100).",
       "type": "number",
-      "exclusiveMinimum": 0,
```
pmbrull, line 40: being a % should we keep the exclusion? I believe old types were OK

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceProfilerPipeline.json
```diff
@@ -35,10 +37,15 @@
...
-      "maximum": 100,
+      "minimum": 0,
+      "default": null
+    },
+    "profileSampleUsingRows": {
```
pmbrull, line 45: nit - maybe rename to `profileSampleRows`?

pmbrull: also, 0 should be an exclusive minimum. If people don't want sample data they can disable the config

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceProfilerPipeline.json
ayush-shah: ohh my bad, i forgot to revert the exclusive and max thingy

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceProfilerPipeline.json
ayush-shah: makes sense..
making that change

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
```diff
@@ -190,6 +193,25 @@ public RestUtil.PutResponse<?> addPipelineStatus(UriInfo uriInfo, String fqn, Pi
     return new RestUtil.PutResponse<>(Response.Status.CREATED, changeEvent, RestUtil.ENTITY_FIELDS_CHANGED);
   }
 
+  public IngestionPipeline checkProfileSampleType(IngestionPipeline ingestionPipeline) throws IOException {
+    if (Objects.equals(ingestionPipeline.getPipelineType().value(), "profiler")) {
```
harshach, line 197: we can just do string comparison

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
```diff
@@ -190,6 +193,25 @@ public RestUtil.PutResponse<?> addPipelineStatus(UriInfo uriInfo, String fqn, Pi
...
+      JSONObject origSourceConfig =
+          new JSONObject(JsonUtils.pojoToJson(ingestionPipeline.getSourceConfig().getConfig()));
+      ObjectMapper objMapper = new ObjectMapper();
```
harshach, line 200: please check the JsonUtils and other code we don't need to initialize another ObjectMapper here

TeddyCr:
@ayush-shah I think adding `profile_sample_rows` variable everywhere bloats the code and makes it a bit harder to maintain. We add this new arguments to all the objects and methods and models and we also add a new method to fetch this value. I feel like it is not a scalable approach. On top of that we add that new method on the backend side to check what kind of sample we have. 

From what I understand, where it really interests us to know if we have rows or sample % when we are going to compute the percentage itself. 

My thinking here would to keep only the `sampleProfile` and have a flag for the method (percentage or number of rows). Then we can use that flag to decide how we want to compute that sample. 

cc @harshach @pmbrull for thoughts

chirag-madlani (dismissed):

ayush-shah:
@TeddyCr, Our initial steps were to add a customType that distinguishes if it's percent or rows, as discussed with @harshach we had it modified to the current implementation using `profileSampleRows`

TeddyCr:
> @TeddyCr, Our initial steps were to add a customType that distinguishes if it's percent or rows, as discussed with @harshach we had it modified to the current implementation using `profileSampleRows`

Make sense. To me, I see this PR as adding technical debt that we'll need to manage down the road.

ayush-shah:
@TeddyCr @harshach ready for review

ingestion/src/metadata/orm_profiler/api/workflow.py
```diff
@@ -257,7 +269,7 @@ def create_profiler_interface(
                 ometa_client=create_ometa_client(self.metadata_config),
                 thread_count=self.source_config.threadCount,
                 table_entity=self._table_entity,
-                table_sample_precentage=self.get_profile_sample(self._table_entity)
+                table_sample_profile=self.get_profile_sample(self._table_entity)
```
TeddyCr, line 272: I think we should make `profile_sample` and `profile_sample_type` attributes of the object (vs passing them as dictionary). Right now it forces us to do something a bit convoluted in `core.py` ln 495-516.

We can have `profile_sample_type` default to `ProfileSampleType.PERCENTAGE`

ingestion/src/metadata/orm_profiler/profiler/sampler.py
```diff
@@ -37,11 +37,18 @@ def __init__(
         self,
         session: Optional[Session],
         table: DeclarativeMeta,
-        profile_sample: Optional[float] = None,
+        profile_sample: Optional[dict] = None,
```
TeddyCr, line 40: Same comment than in `workflow.py`. Let's pass `profile_sample` and `profile_sample_type` as attributes of the class.

ingestion/src/metadata/orm_profiler/profiler/sampler.py
```diff
@@ -51,14 +58,20 @@ def __init__(
...
+                .suffix_with(
+                    f"SAMPLE BERNOULLI ({self.profile_sample or 100})",
+                    dialect=Dialects.Snowflake,
+                )
+                .cte(f"{self.table.__tablename__}_rnd")
             )
+        return (
+            self.session.query(self.table)
+            .limit(self.profile_sample)
             .cte(f"{self.table.__tablename__}_rnd")
```
TeddyCr, line 75: I believe you should be able to remove the `.cte` here since we won't do any further operation on this table and you are creating the cte line 102 for this sampling type.

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
```diff
@@ -107,8 +107,11 @@ def setUpClass(cls) -> None:
         """
         Prepare Ingredients
         """
-        User.__table__.create(bind=cls.engine)
-        NewUser.__table__.create(bind=cls.engine)
+        try:
```
TeddyCr, line 110: what are you trying to catch here. If it is exception when the table already exists can you be more explicit?

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
```diff
@@ -151,8 +154,10 @@ def tearDownClass(cls) -> None:
             hard_delete=True,
         )
 
-        NewUser.__table__.drop(bind=cls.engine)
         User.__table__.drop(bind=cls.engine)
+        NewUser.__table__.drop(bind=cls.engine)
+        cls.session.close()
+        return super().tearDownClass()
```
TeddyCr, line 160: we can remove this line, it won't do anything. The parent class does not have any implementation https://github.com/python/cpython/blob/702a5bc4637c82dc011e98b84f0cede98eb08dda/Lib/unittest/case.py#L479-L481

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
```diff
@@ -212,6 +217,32 @@ def test_profiler_workflow(self):
 
         assert not table.tableProfilerConfig
         assert profile.profileSample == 75.0
+        assert profile.profileSampleType == ProfileSampleType.PERCENTAGE
+
+        workflow_config["processor"]["config"]["tableConfig"][0][
+            "profileSampleType"
+        ] = ProfileSampleType.ROWS
```
TeddyCr, line 224: Can you add a specific number of rows (let's 2 or 3) and check that we indeed only got 2 or 3 rows.

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceProfilerPipeline.json
```diff
@@ -35,12 +37,14 @@
       "default": true
     },
     "profileSample": {
-      "description": "Percentage of data used to execute the profiler for the whole workflow. This percentage will be applied to all tables in the workflow. Represented in the range (0, 100].",
+      "description": "Percentage of data we want to execute the profiler and tests on. Represented in the range (0, 100).",
       "type": "number",
-      "exclusiveMinimum": 0,
       "maximum": 100,
       "default": null
```
TeddyCr, line 43: You'll need to update that. The maximum should not be `100` otherwise people won't be able to sample more than 100 ROWS

TeddyCr (changes requested):
Thanks @ayush-shah for this PR. Added a few comments. In addition could you open 2 additional follow up tickets and add them to 1.0:
1. Validate input from the backend (i.e. if PERCENTAGE then 0 < value < 100
2. Add logic when ROWS to randomize the output.

ingestion/src/metadata/ingestion/ometa/mixins/server_mixin.py
```diff
@@ -67,7 +67,10 @@ def get_server_version(self) -> str:
         Run endpoint /version to check server version
         :return: Server version
         """
-        raw_version = self.client.get("/version")["version"]
+        try:
```
TeddyCr, line 70: when would that happen? Is it when the server is not started?

ayush-shah: Server had started..
but the version showed up as { }
tried the API got the same response, python was failing with key error..
did the mvn clean package, once that completed, it showed the version.

I started the server from Idea Intellij, that's when one of these edge cases hit.

ingestion/src/metadata/orm_profiler/profiler/core.py
```diff
@@ -91,6 +90,9 @@ def __init__(
         self.exclude_columns = exclude_columns
         self._metrics = metrics
         self._profile_date = profile_date
+        self.profile_sample_config = (
+            self.profiler_interface.profile_sample_config or None
```
TeddyCr, line 94: I think you can just do `self.profiler_interface.profile_sample_config`. Because in `get_profile_sample` we either return the object or None.

ayush-shah: resolving that

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
```diff
@@ -111,7 +111,7 @@ def setUpClass(cls) -> None:
             User.__table__.create(bind=cls.engine)
             NewUser.__table__.create(bind=cls.engine)
         except:
-            pass
+            print("Table Already exists")
```
TeddyCr, line 114: print statement

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
```diff
@@ -241,7 +240,7 @@ def test_profiler_workflow(self):
         ).profile
 
         assert not table.tableProfilerConfig
```
TeddyCr, line 242: Can you count the rows from the profile as well?

openmetadata-spec/src/main/resources/json/schema/entity/data/table.json
```diff
@@ -633,7 +633,6 @@
         "profileSample": {
           "description": "Percentage of data we want to execute the profiler and tests on. Represented in the range (0, 100).",
```
TeddyCr, line 634: Can you update the description, there is still range (0, 100)

ayush-shah: Resolved

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
ayush-shah: Resolved

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
ayush-shah: Resolved

ingestion/src/metadata/orm_profiler/profiler/sampler.py
TeddyCr: @ayush-shah this was not fixed

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
TeddyCr: line 246 `test_workflow_sample_profile` definition

ingestion/src/metadata/orm_profiler/profiler/sampler.py
ayush-shah: without the cte, i am not able to get `inspect(rnd).c`
it gives me an error as it's not able to find `c`

ingestion/src/metadata/ingestion/ometa/mixins/server_mixin.py
TeddyCr: I see, I don't think we should throw a version missmatch error as we don't have any version at all here.

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
ayush-shah: yep, on it.

ingestion/src/metadata/ingestion/ometa/mixins/server_mixin.py
ayush-shah: fixing it, let me create a VersionNotFound exception

ayush-shah: Modified

ingestion/tests/integration/orm_profiler/test_orm_profiler_e2e.py
ayush-shah: Done

openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
```diff
@@ -139,7 +139,7 @@
     "exclude": "Exclude",
     "enable-column-profile": "Enable column profile",
     "profile-sample-query": "Profile Sample Query",
-    "profile-sample-percentage": "Profile Sample %",
+    "profile-sample": "Profile Sample {{type}}",
```
chirag-madlani:
```suggestion
    "profile-sample-type": "Profile Sample {{type}}",
```

openmetadata-ui/src/main/resources/ui/src/components/AddIngestion/Steps/ConfigureIngestion.tsx
```diff
@@ -126,39 +137,61 @@ const ConfigureIngestion = ({
...
-          name="profileSample"
-          placeholder="75"
-          type="number"
-          value={profileSample}
-          onChange={(e) => handleProfileSampleValidation(+e.target.value)}
-        />
-        {profileSampleError && errorMsg('Value must be between 0 and 99.')}
-      </div>
+      <>
+        <Form.Item
```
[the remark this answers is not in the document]
Ashish8689: @chirag-madlani  yes, we will update to Antd

openmetadata-ui/src/main/resources/ui/src/components/SliderWithInput/SliderWithInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+ *  limitations under the License.
+ */
+
+import { Col, InputNumber, Row, Slider } from 'antd';
+import React from 'react';
+import { SliderWithInputProps } from './SliderWithInput.interface';
+
+const SliderWithInput = ({
+  value,
+  onChange,
```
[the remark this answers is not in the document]
Ashish8689: @chirag-madlani  creating a separate PR for this

Ashish8689:
@chirag-madlani updated the changes as per comments!

chirag-madlani (approved):

TeddyCr (approved):
LGTM 🚀 Thanks @ayush-shah

PR #9130, opened 2022-12-02
Fix #9123 - Update default owner and tag

### LINKED ISSUE #9123: Change default owner name for dags
When we create an ingestion process, the owner for the dags is airflow and isn't possible to change this value.
If we use an existing airflow instance that supports other applications and have some patterns for owners name, we have some problems to use airflow as owner name, because this is a generic name.

Add a parameter that we can use to change the dags owner name

pmbrull:
### Describe your changes :

Fix #9123

IMO it makes sense to clearly separate the OM-managed DAGs if the plugins are installed in existing instances. Thanks for bringing it up.

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


openmetadata-airflow-apis/openmetadata_managed_apis/workflows/ingestion/common.py
```diff
@@ -297,6 +298,8 @@ def build_dag(
             retries=0,
             # each DAG will call its own OpenMetadataWorkflowConfig
             on_failure_callback=partial(send_failed_status_callback, workflow_config),
+            # Add tag and ownership to easily identify DAGs generated by OM
+            owner="openmetadata",
```
harshach, line 302: why not pass the owner from the user who is deployed it. We already recording this in IngestionPipeline Owner

pmbrull:
will simplify this, not convinced with adding the tags tbh

PR #9154, opened 2022-12-05
UI: Added support for renaming user generated tag category

ShaileshParmar11:
### Describe your changes :

Added support for renaming user added tags category

#
### Type of change :

- [x] New feature

#

https://user-images.githubusercontent.com/71748675/205670674-3f94b1e2-a018-4fd3-90f3-1da5b74dfcfb.mov

#
### Checklist:

#


openmetadata-ui/src/main/resources/ui/src/pages/tags/index.tsx
```diff
@@ -328,14 +346,19 @@ const TagsPage = () => {
     }
   };
 
-  const UpdateCategory = async (updatedHTML: string) => {
+  const updateCategory = async (updatedCategory: TagCategory) => {
```
Sachin-chaurasiya, line 349:
```suggestion
  const handleUpdateCategory = async (updatedCategory: TagCategory) => {
```

openmetadata-ui/src/main/resources/ui/src/pages/tags/index.tsx
```diff
@@ -346,6 +369,20 @@ const TagsPage = () => {
     }
   };
 
+  const handleRenameSave = () => {
+    updateCategory({
+      name: (currentCategoryName || currentCategory?.name) ?? '',
+      description: currentCategory?.description ?? '',
```
Sachin-chaurasiya, line 375: Why we are sending the description?

openmetadata-ui/src/main/resources/ui/src/pages/tags/index.tsx
```diff
@@ -632,13 +669,85 @@ const TagsPage = () => {
...
             {currentCategory && (
-              <div
-                className="tw-flex tw-justify-between tw-items-center"
-                data-testid="header">
-                <div
-                  className="tw-text-link tw-text-base tw-py-2"
-                  data-testid="category-name">
-                  {currentCategory.displayName ?? currentCategory.name}
+              <div className="flex justify-between" data-testid="header">
+                <div className="flex items-center">
```
Sachin-chaurasiya, line 673: I would recommend using antd space here.

openmetadata-ui/src/main/resources/ui/src/pages/tags/index.tsx
```diff
@@ -632,13 +669,85 @@ const TagsPage = () => {
...
+                    <Row align="middle" gutter={8}>
+                      <Col>
+                        <Input
+                          className="input-width"
+                          data-testid="tag-category-name"
+                          name="tagCategoryName"
+                          value={currentCategoryName}
+                          onChange={(e) =>
+                            setCurrentCategoryName(e.target.value)
+                          }
```
Sachin-chaurasiya, line 684: Do not directly set the state value in the onChange callback, create a separate handler and set the state there.

Sachin-chaurasiya (commented):
@ShaileshParmar11, Here I can see we are only making changes to the edit tag category, we also need to support editing tags as well.

ShaileshParmar11:
> @ShaileshParmar11, Here I can see we are only making changes to the edit tag category, we also need to support editing tags as well.

The backend only supports renaming tags category and not tags

openmetadata-ui/src/main/resources/ui/src/pages/tags/index.tsx
ShaileshParmar11: this is the PUT call, we have to send all the required details!

openmetadata-ui/src/main/resources/ui/src/pages/tags/index.tsx
ShaileshParmar11: What difference it will make? eventually, it will also be a callback! if callback is ha more than one line of operation then it will make sense to create a separate function but in this case, I don't feel there is need to create an extra function and assign it! Please let me know if you think of another way!

openmetadata-ui/src/main/resources/ui/src/pages/tags/index.tsx
```diff
@@ -108,6 +117,8 @@ const TagsPage = () => {
   });
   const [categoryPermissions, setCategoryPermissions] =
     useState<OperationPermission>(DEFAULT_ENTITY_PERMISSION);
+  const [isNameEditing, setIsNameEditing] = useState(false);
```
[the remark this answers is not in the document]
ShaileshParmar11: even if we directly assign value, it will typecast as per value!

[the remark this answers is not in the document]
Sachin-chaurasiya: Yes, it will, in other states you are providing the generic value, so you should follow the same for all states,

openmetadata-ui/src/main/resources/ui/src/pages/tags/index.tsx
Sachin-chaurasiya: Make sense, but creating a separate handler would be better practice here.

openmetadata-ui/src/main/resources/ui/src/pages/tags/index.tsx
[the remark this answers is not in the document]
ShaileshParmar11: Yes because it was not having any default value!

Sachin-chaurasiya (approved):

PR #9159, opened 2022-12-05
Fixes  #7447 -- Add freshness metrics to profiler

### LINKED ISSUE #7447: Add Freshness Metrics for the profiler
For a select number of DBAPI (at minimum snowflake, redshift and bigquery) we should show the timeline of table insertion update (datetime). We should rely on the syslog of the DBAPI to do so

TeddyCr:
### Describe your changes :
This PR fixes #7447 by:
- implementing E2E test for datalake profiler
- refactoring profiler code bringing get_metrics function back to their class
- Added logic in ingestion to fetch system metrics from redshift, snowflake and bigquery
- Added logic on the backend to add and retrieve system metrics

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


TeddyCr:
cc @ShaileshParmar11

ulixius9 (dismissed):

ingestion/src/metadata/ingestion/ometa/mixins/table_mixin.py
```diff
@@ -215,3 +218,58 @@ def create_or_update_table_profiler_config(
...
+            after (_type_, optional): use for API pagination. Defaults to None.
+            profile_type (Union[Type[TableProfile], Type[ColumnProfile]], optional):
+                Profile type to retrieve. Defaults to TableProfile.
+
+        Raises:
+            ValueError: _description_
+            TypeError: _description_
+
+        Returns:
+            _type_: Profile object with data, after and total attributs
```
pmbrull, line 247: If we add this info as a docstring could we also add the type hint in the definition?

pmbrull: ok just saw we use a `namedtuple`. Any reason we're not defining a BaseModel that we can use and share better?

ingestion/src/metadata/orm_profiler/profiler/core.py
```diff
@@ -213,6 +216,10 @@ def custom_metrics(self) -> List[Type[CustomMetric]]:
     def query_metrics(self) -> List[Type[QueryMetric]]:
         return self._filter_metrics(QueryMetric)
 
+    @property
+    def system_metrics(self) -> List[Type[SystemtMetric]]:
```
pmbrull, line 220: Aren't system metrics just a flavour of `QueryMetric`s (e.g., the histogram)?

ingestion/src/metadata/utils/dispatch.py
```diff
@@ -54,3 +56,53 @@ def inner(fn):
...
+        try:
+            impl = registry[value]
+        except KeyError:
+            impl = registry[object]
+        return impl
+
+    def register(value, func=None) -> Callable:
+        if _is_valid_dispatch(value):
+            if func is None:
+                return lambda f: register(value, f)
```
pmbrull, line 86: Not sure I follow this 🤔

ingestion/tests/integration/orm_profiler/test_datalake_profiler_e2e.py
```diff
@@ -0,0 +1,197 @@
...
+            "securityConfig": {
+                "jwtToken": "eyJraWQiOiJHYjM4OWEtOWY3Ni1nZGpzLWE5MmotMDI0MmJrOTQzNTYiLCJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJhZG1pbiIsImlzQm90IjpmYWxzZSwiaXNzIjoib3Blbi1tZXRhZGF0YS5vcmciLCJpYXQiOjE2NjM5Mzg0NjIsImVtYWlsIjoiYWRtaW5Ab3Blbm1ldGFkYXRhLm9yZyJ9.tS8um_5DKu7HgzGBzS1VTA5uUjKWOCU0B_j08WXBiEC0mr0zNREkqVfwFDD-d24HlNEbrqioLsBuFRiwIWKc1m_ZlVQbG7P36RUxhuv2vbSp80FKyNM-Tj93FDzq91jsyNmsQhyNv_fNr3TXfzzSPjHt8Go0FMMP66weoKMgW2PbXlhVKwEuXUHyakLLzewm9UMeQaEiRzhiTMU3UkLXcKbYEJJvfNFcLwSl9W8JCO_l0Yj3ud-qt_nQYEZwqW6u5nfdQllN133iikV4fM5QZsMCnm8Rq1mvLR0y9bmJiD7fwM1tmJ791TUWqmKaTnP49U493VanKpUAfzIiOiIbhg"
+            },
+        }
+    },
+}
+
+
+@mock_s3
+class DatalakeProfilerTestE2E(TestCase):
```
pmbrull, line 78: Thanks for adding this 🙏

ingestion/src/metadata/utils/dispatch.py
```diff
@@ -54,3 +56,53 @@ def inner(fn):
 
     Register = namedtuple("Register", ["add", "registry"])
     return Register(add, registry)
+
+
+def valuedispatch(func) -> Callable:
```
pmbrull, line 61: Could you please explain the rationale behind this? I see it being used with enums and not consistently passing the value vs. the enum object

```python
@get_system_metrics_for_dialect.register(Dialects.BigQuery)
# ...
@_get_metrics.register(MetricTypes.Table.value)
```

Could you please explain why is the extra logic needed instead of just using the `enum_register`? Thanks!

ingestion/src/metadata/orm_profiler/profiler/core.py
TeddyCr: kind of, but Query metrics are querying specific columns for that table. We don't query the table but the information schema table.

ingestion/src/metadata/utils/dispatch.py
TeddyCr: It simply allows you explicitly pass a function.

ingestion/src/metadata/ingestion/ometa/mixins/table_mixin.py
TeddyCr: They are slightly different for each DB.

ingestion/src/metadata/utils/dispatch.py
TeddyCr: It always passes strings. If it does not the dispatch function will throw an error. `MetricTypes ` inherit from `enum.Enum` while Dialects inherits from `sqlalchemy.sql.sqltypes.Enum`. The first one returns an enum object so we need to access its value through the `value` method while the second one returns a `str`.

TeddyCr: https://github.com/open-metadata/OpenMetadata/blob/240babb092f2a864adc247a8c4128b8a08b9da7e/ingestion/src/metadata/utils/dispatch.py#L73-L74

pmbrull: aaaaah okok so basically we're dispatching based on strings. got it, thank you

pmbrull: sorry you were my first review of the day and coffee did not kick in 😂

ingestion/src/metadata/orm_profiler/profiler/core.py
pmbrull: oooh got it. Makes sense, We could eventually use this SystemMetric to query internal statistics etc. Thank you for explaining

pmbrull: also I just noticed there's a typo `List[Type[SystemtMetric]]:` System"t" metric

ingestion/src/metadata/ingestion/ometa/mixins/table_mixin.py
pmbrull: then could we have explicit definitions for each db and here pass the parent type? This way it might be easier to understand what we need in each scenario. We won't support all DBs out of the box, so we could just keep expanding the defs as needed

TeddyCr: Sorry i thought we were talking about something else. There is actually already something defined [here](https://github.com/open-metadata/OpenMetadata/blob/25449001ca978fd14bd7178a194cb40402bb4f6a/ingestion/src/metadata/ingestion/ometa/ometa_api.py#L131-L143) though we can't use it in the Mixins as it throws a circular import. We could move that base class but i am not sure where it actually fits. 

The other solution is to move this method out of the mixin and insidie `ometa_api`

pmbrull: in any case, I believe this is important enough to have a closed definition. It can save some headaches down the line

pmbrull: we can close this PR if you want and handle this in another iteration, don't want to hold this off just because of this though

pmbrull (approved):

PR #9121, opened 2022-12-02
Add analyzer to search in Japanese

megumiie:
### Describe your changes :

I changed OpenMetadata to be able to search in Japanese only for description fields of table and column.  
Added Japanese analyzer to table_search_index and changed it to map to description fields.

> __Warning__  
[kuromoji](https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-kuromoji.html) must be installed on Elasticsearch to enable this setting.

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Frontend Preview (Screenshots) :

image001

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] All new and existing tests passed.

#
### Reviewers

@harshach

PR #9163, opened 2022-12-06
Fix #9158: OpenSearch/ElasticSearch - Language based mappings

### LINKED ISSUE #9158: OpenSearch/ElasticSearch - Language based mappings
OpenMetadata uses ElasticSearch or OpenSearch to index all the metadata and provide search/explore options. So far, we have only provided options to index and analyze documents in the English language; however as more and more users come in and there is a strong need for the analyzer to be pluggable. We are going to address the mappings to specific for each language and allow users to configure based on their language of their choice

harshach:
### Describe your changes :

#
### Type of change :

- [x] Improvement

#
</p>

#
### Checklist:

#


pmbrull (approved):
Pretty awesome, thanks for tackling it

pmbrull:
Should we update the docs with the comment on https://github.com/open-metadata/OpenMetadata/pull/9121 that the analyzer should be previously installed on ES?

harshach:
@pmbrull yes.
also is there any way in setup.py that we can refer to these files and copy it into our python packaging.
It would be good to move away maintaining separate configs for python and java.
cc @TeddyCr

openmetadata-service/src/main/resources/elasticsearch/jp/mlmodel_index_mapping.json
```diff
@@ -0,0 +1,345 @@
...
+              }
+            }
+          }
+        }
+      },
+      "mlHyperParameters": {
+        "properties": {
+          "name": {
+            "type": "text",
+            "analyzer": "om_analyzer"
```
megumiie, line 112: Need to change this _om_analyzer_ to _om_analyzer_jp_ or 
add the definition of _om_analyzer_  to the analyzer of this index.

harshach:
Thanks @megumiie for the review. Addressed your comments

TeddyCr:
@harshach to be able to include the elasticsearch files they will need to be moved to the python package folder (e.g. `ingestion/src/metadata/resources/elasticsearch`). Other than that solution is fairly straightforward. Let me know if that would be a blocker on the Java side. 

Once I have confirmation on your end, I'll go ahead and work on the implementation.

harshach:
@TeddyCr I think it will be bit weird to move the mappings to client-side code and let the server depend on it. Other option is I can create an end point in SearchResource that will allow admin/ingest-bot to re-create the indexes and we can use ingestion to make a call

megumiie (dismissed):
@harshach Thanks for the change!
Now it works locally.

Issue #7814, opened 2022-09-30
Improve logic when ingestion sample data with for BYTES column


TeddyCr:
## Context
When ingesting sample data with bytes DType we currently convert python's primary bytes type to an hex string. While this allows us to ingest table sample where bytes columns are present information is lost for the end user.

## A.C.
Bytes type are accuratly decoded into strings and ingested

harshach:
@MilanBariya please fix this asap

MilanBariya:
Sure @harshach, once I complete a current task I will start working on this.
Thanks!.

PR #9261, opened 2022-12-13
Logic improved for BYTES column

MilanBariya:
### Describe your changes :

I worked on the Improving logic when ingestion sample data with for BYTES column
Fix: #7814

#
### Type of change :

- [x] Bug fix
- [x] Improvement

#
### Checklist:

#


ingestion/src/metadata/orm_profiler/orm/types/bytea_to_string.py
```diff
@@ -49,4 +50,11 @@ def process_result_value(self, value: str, dialect) -> Optional[str]:
         if not value:
             return None
         self.validate(value)
+
```
OnkarVO7, line 53: @MilanBariya could you please add a try/except block around the logic and fall back to default return if anything goes wrong while decoding

ingestion/src/metadata/orm_profiler/orm/types/hex_byte_string.py
```diff
@@ -51,6 +52,13 @@ def process_result_value(self, value: str, dialect) -> Optional[str]:
         if not value:
             return None
         self.validate(value)
+
```
OnkarVO7, line 55: same try/except block

OnkarVO7 (changes requested):

ingestion/src/metadata/orm_profiler/orm/types/hex_byte_string.py
```diff
@@ -51,6 +52,13 @@ def process_result_value(self, value: str, dialect) -> Optional[str]:
...
+        bytes_value = bytes(value)
```
TeddyCr, line 56: @MilanBariya why do we convert this to a `bytes` object. Normally it should already be one as at this stage it would have gone through the `self.validate()`.

Are you able to share an example of how the data is displayed in the sample data?

MilanBariya: Sure @TeddyCr , Let me share example in slack.

MilanBariya: i’m converting that in `bytes`  becaue result was in `memoryview`

here is the before after result
Screenshot 2022-12-13 at 1 13 38 PM

Screenshot 2022-12-13 at 1 13 58 PM

ingestion/src/metadata/orm_profiler/orm/types/hex_byte_string.py
MilanBariya: Done

OnkarVO7 (approved):
LGTM! Feel free to merge after @TeddyCr 's review. Thanks!

ingestion/src/metadata/orm_profiler/orm/types/bytea_to_string.py
```diff
@@ -49,4 +55,15 @@ def process_result_value(self, value: str, dialect) -> Optional[str]:
...
+        try:
+            bytes_value = bytes(value)
+            detected_encoding = chardet.detect(bytes_value).get("encoding")
+            if detected_encoding:
+                value = bytes_value.decode(encoding=detected_encoding)
+                return value
+        except Exception as exc:
+            logger.debug(traceback.format_exc())
+            logger.error(exc)
+
```
TeddyCr, line 68: I think we should add the try except only for this operation `value = bytes_value.decode(encoding=detected_encoding)`. I believe the other ones won't throw an error (if i am not mistaking).

Also in `self.validate` let's add `if not isinstance(value, (memoryview, bytes, bytearray)):`

TeddyCr (changes requested):
I think we should add the try except only for this operation value = bytes_value.decode(encoding=detected_encoding). I believe the other ones won't throw an error (if i am not mistaking).

Also in self.validate let's add if not isinstance(value, (memoryview, bytes, bytearray)):

ingestion/src/metadata/orm_profiler/orm/types/bytea_to_string.py
MilanBariya: Done

TeddyCr (approved):

OnkarVO7 (approved):

PR #9263, opened 2022-12-13
Fix #8944: Added pagination logic for PowerBi

### LINKED ISSUE #8944: PowerBI group not ingested because of limit of top=100
**Affected module**
Ingestion Framework of PowerBI

**Describe the bug**
If the number of groups in PowerBI are more than 100, some groups, and therefore its dashboards, can't be ingested. This is likely because the API call to GetGroupsAsAdmin requires a `$top` parameter, which is currently set as 100 as of the pre-release 0.13.0.

**To Reproduce**
I have created a new workspace called "Test Openmetadata" with a dashboard but OM wouldn't ingest it. When I made a GET request to `https://api.powerbi.com/v1.0/myorg/admin/groups?top=100` as seen in the debug log, the output returned only the top 100 workspaces. When I called `https://api.powerbi.com/v1.0/myorg/admin/groups?top=1500`, 'Test Openmetadata' was found in the output.

**Version:**
 - OS: Windows
 - Python version:
 - OpenMetadata version: 0.13.0
 - OpenMetadata Ingestion package version: 0.13.0

**Additional context**
From the PowerBI API output, our company currently has 1300+ PowerBI workspaces, but more than 1200+ are of type `PersonalGroup`. While our company is not interested in seeing the metadata of personal workspaces, whether to add a filter to exclude personal groups from the output depends on whether a general use case finds personal workspaces useful to track.
Another useful filter could be retrieving only worskpaces with `state=active`.

OnkarVO7:
### Describe your changes :

Fix https://github.com/open-metadata/OpenMetadata/issues/8944

#
### Type of change :

- [x] Bug fix
- [x] Improvement

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/clients/powerbi_client.py
+++ b/ingestion/src/metadata/clients/powerbi_client.py
@@ -12,6 +12,7 @@
 REST Auth & Client for PowerBi
 """
 import json
+import math
 import traceback
 from time import sleep
 from typing import List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


ingestion/src/metadata/clients/powerbi_client.py
```diff
@@ -91,18 +92,30 @@ def fetch_dashboards(self) -> Optional[dict]:
 
         return None
 
-    def fetch_all_workspaces(self) -> Optional[dict]:
+    def fetch_all_workspaces(self) -> List[dict]:
```
TeddyCr, line 95: The type Hint should be Optional[List[dict]]

OnkarVO7: Yes thanks

OnkarVO7: @TeddyCr added the `Optional`

TeddyCr (approved):

Issue #4648, opened 2022-05-03
Airflow lineage with ES instead of FQN


pmbrull:
**Is your feature request related to a problem? Please describe.**
Requiring to add the FQN can be a bit hard when informing inlets and outlets

**Describe the solution you'd like**
Use ES as a middle-step to:
1. Use antlr to parse the inlet/outlet and fetch the possible entity ID
2. Use the entity ID to build the lineage

**Describe alternatives you've considered**
Stick to FQN to have no room for operational errors or inconsistencies

**Additional context**

**Related issues**
- ...

harshach:
@pmbrull anything remaining here or can we close it

pmbrull:
We still need to pick it up. Should be a quick improvements in 0.13. Thanks

harshach:
@nahuelverdugo pick up for 0.13.1 or 1.0?

nahuelverdugo:
I will take this for `0.13.1`.

PR #9273, opened 2022-12-13
Fix#4648: Use ES call to get table entity

nahuelverdugo:
Fixes: #4648

### Describe your changes :
Use table results from ES search to process the lineage instead of using them for building the FQN and doing an extra call to the API for the same results.

### Type of change :
- [x] Improvement

### Checklist:

### Reviewers


ingestion/src/metadata/config/common.py
```diff
@@ -69,10 +69,8 @@ def load_config(self, config_fp: IO) -> dict:
         try:
             config = yaml.safe_load(config_fp)
             return config
-        except yaml.error.YAMLError:
-            raise ConfigurationError(
-                f"YAML Configuration file [{config_fp}] is not a valid YAML"
-            )
+        except yaml.error.YAMLError as exc:
+            raise ConfigurationError(f"YAML Configuration file is not valid \n {exc}")
```
nahuelverdugo, line 73: When testing it, it was not displaying the error correctly.

ingestion/src/metadata/config/common.py
```diff
@@ -84,10 +82,8 @@ def load_config(self, config_fp: IO) -> dict:
         try:
             config = json.load(config_fp)
             return config
-        except json.decoder.JSONDecodeError:
-            raise ConfigurationError(
-                f"JSON Configuration file [{config_fp}] is not a valid JSON"
-            )
+        except json.decoder.JSONDecodeError as exc:
```
nahuelverdugo, line 85: When testing it, it was not displaying the error correctly.

ingestion/src/metadata/ingestion/lineage/sql_lineage.py
```diff
@@ -124,15 +138,6 @@ def get_table_entities_from_query(
         database=database_name,
         database_schema=database_schema,
         table=table,
-    ) or (
```
nahuelverdugo, line 127: Since we are looking first at ES, we don't need to do this call.

ingestion/src/metadata/ingestion/lineage/sql_lineage.py
```diff
@@ -64,20 +65,33 @@ def search_table_entities(
     if search_tuple in search_cache:
         return search_cache.get(search_tuple)
     try:
-        table_fqns = fqn.build(
-            metadata,
+        table_entities: Optional[List[Table]] = []
+        # search on ES first
+        fqn_search_string = build_es_fqn_search_string(
```
TeddyCr, line 70: Should split the try block? I am wondering if `build_es_fqn_search_string` raises if we still want to run `get_by_name()`

ingestion/src/metadata/ingestion/lineage/sql_lineage.py
```diff
@@ -64,20 +65,33 @@ def search_table_entities(
...
+            database, database_schema, service_name, table
+        )
+        es_result_entities = metadata.es_search_from_fqn(
```
TeddyCr, line 73: Are we doing that elasticsearch call as it is less expensive than the actual OM Server?

ingestion/src/metadata/utils/fqn.py
```diff
@@ -142,19 +143,18 @@ def _(
...
-        fqn_search_string=fqn_search_string,
-    )
+    es_result = (
+        not skip_es_search
+        and metadata.es_search_from_fqn(
+            entity_type=Table,
+            fqn_search_string=fqn_search_string,
+        )
+    ) or None
+
```
TeddyCr, line 157: I like it, I am wondering if we could do 

```
es_result = metadata.es_search_from_fqn(entity_type=Table,fqn_search_string=fqn_search_string,) if not skip_es_search else None
```

I believe it gives the same results. This is very opinionated comment 😅 as it does not change the functionality itself. I just feel like it might make it more readible.

nahuelverdugo: Oh yes, you are right. It is clearer.

ingestion/src/metadata/ingestion/lineage/sql_lineage.py
nahuelverdugo: We are calling `build_es_fqn_search_string ` also from `fqn.build`, so we are not going to call `get_by_name()` anyway.

ingestion/src/metadata/ingestion/lineage/sql_lineage.py
nahuelverdugo: When calling `fqn.build`, we already get the Table entities from ES, so we are doing an unnecessary extra call to the API with the FQN. So the idea in this PR is to use the result of ES directly; otherwise, call to the API using the `get_by_name` method with the FQN.

TeddyCr (approved):

PR #9305, opened 2022-12-15
Integrated schema parsers

OnkarVO7:
### Describe your changes :

Integrated the new schema parsers for Avro/Protobuf/JsonSchema based on the new Topic model Changes
Modified tests and sample data accordingly

#
### Type of change :

- [x] Improvement
- [x] New feature

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/setup.py
+++ b/ingestion/setup.py
@@ -45,6 +45,8 @@ base_requirements = {
     "boto3~=1.19.12",
     "botocore==1.22.12",
     "avro-python3==1.10.2",
+    "grpcio-tools",
+    "protobuf",
     # compatibility requirements for 3.7
     "typing-compat~=0.1.0",
     "importlib-metadata~=4.12.0",  # From airflow constraints
--- a/ingestion/setup.py
+++ b/ingestion/setup.py
@@ -45,8 +45,8 @@ base_requirements = {
     "boto3~=1.19.12",
     "botocore==1.22.12",
     "avro-python3==1.10.2",
-    "grpcio-tools==1.51.1",
-    "protobuf==4.21.11",
+    "grpcio-tools",
+    "protobuf",
     # compatibility requirements for 3.7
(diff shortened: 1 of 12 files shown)


ingestion/src/metadata/ingestion/source/messaging/common_broker_source.py
```diff
@@ -101,19 +104,47 @@ def yield_topic(
             self.add_properties_to_topic_from_resource(topic, topic_config_resource)
 
             if topic_schema is not None:
-                topic.schemaText = topic_schema.schema_str
                 if topic_schema.schema_type.lower() == SchemaType.Avro.value.lower():
-                    topic.schemaType = SchemaType.Avro.name
```
pmbrull, line 106: These different logics based on the SchemaType could be a dispatch to keep implementations separated and here just have a single call

ingestion/setup.py
```diff
@@ -45,6 +45,8 @@ def get_long_description():
     "boto3~=1.19.12",
     "botocore==1.22.12",
     "avro-python3==1.10.2",
+    "grpcio-tools",
```
NiharDoshi99, line 48: @OnkarVO7 , can we add version for the packages.

ingestion/src/metadata/ingestion/source/messaging/common_broker_source.py
pmbrull: I believe you could use this https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/utils/dispatch.py#L26 if the `SchemaType` is an Enum

OnkarVO7: Yes

NiharDoshi99 (approved):

PR #9321, opened 2022-12-15
Extra headers for metadata API

czuares:
### Describe your changes :

- Allows for extra headers to be provided to the ometa API client
- Adds `extraHeaders` to the `workflowConfig.openMetadataServerConfig` schema
- Supports copying existing header values using string modulo format with the header dictionary

Example usage:

```yaml
  openMetadataServerConfig:
    hostPort: ${OMD_API_HOST}
    authProvider: google
    extraHeaders:
      Proxy-Authorization: "%(Authorization)s"
      Some-Env-Header: ${MY_ENV_VAR}
      Static-Header: my-value
```

The primary use case is for Google Cloud's Identity Aware Proxy (IAP) which allows secure access to private services but removes the Authorization header in the process. The `Proxy-Authorization` header gets forwarded along to the service as the `Authorization` header for the service behind IAP. In this case, the token is valid for Google SSO as well and can be passed through and enables "nested" authorization options.

Here's a [relevant StackOverflow](https://stackoverflow.com/questions/59297161/using-nested-authentication-with-google-iap) about the header.

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

#
### Reviewers
@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/ingestion/ometa/client.py
+++ b/ingestion/src/metadata/ingestion/ometa/client.py
@@ -165,10 +165,10 @@ class REST:
         # Example: "Proxy-Authorization": "%(Authorization)s"
         # This will result in the Authorization value being set for the Proxy-Authorization Extra Header
         if self.config.extra_headers:
-            extra_headers = self.config.extra_headers.get("__root__", {})
-            extra_headers = [extra_headers[eh] % headers for eh in extra_headers]
+            extra_headers: dict[str, str] = self.config.extra_headers
+            extra_headers = {k: (v % headers) for k, v in extra_headers.items()}
             logger.debug("Extra headers provided '%s'", extra_headers)
-            headers = headers | extra_headers
+            headers = {**headers, **extra_headers}
 
         opts = {
             "headers": headers,
--- a/ingestion/src/metadata/ingestion/ometa/client.py
+++ b/ingestion/src/metadata/ingestion/ometa/client.py
@@ -136,6 +136,7 @@ class REST:
         api_version: str = None,
         headers: dict = None,
     ):
+        # pylint: disable=too-many-locals
         if not headers:
             headers = {"Content-type": "application/json"}
         base_url = base_url or self._base_url
--- a/ingestion/src/metadata/ingestion/ometa/client.py
+++ b/ingestion/src/metadata/ingestion/ometa/client.py
@@ -101,6 +101,8 @@ class ClientConfig(ConfigModel):
     access_token: Optional[str] = None
     expires_in: Optional[int] = None
     auth_header: Optional[str] = None
+    extra_auth_header: Optional[str] = None
+    extra_auth_header_value: Optional[str] = None
     raw_data: Optional[bool] = False
     allow_redirects: Optional[bool] = False
     auth_token_mode: Optional[str] = "Bearer"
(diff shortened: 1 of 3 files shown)


pmbrull:
hi @czuares, thanks for the proposal. I am unsure if relying on environment variables is the best way around this. Updating the [OpenMetadataConnection](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/metadata/openMetadataConnection.json) JSON Schema would help on keeping track on the different ways to build the client.

@harshach do you think we might need something similar to `connectionOptions` here to allow key-value pairs to be passed to the headers?

Thanks

czuares:
@pmbrull Thanks for taking the time to review my PR. I would be happy to improve upon this approach. If we are considering using the JSON schema, perhaps it could look something like this:

```
workflowConfig:
  openMetadataServerConfig:
    securityConfig:
      extraApiHeaders:
      - name: SOME_HEADER
        value: ${SOMETHING_FROM_ENV}
      - name: Proxy-Authorization
        valueFrom: Authorization
```

This would allow for setting headers from the environment using the `value` key or using the `valueFrom` key to copy the value from an existing header. Thoughts?

pmbrull:
> @pmbrull Thanks for taking the time to review my PR. I would be happy to improve upon this approach. If we are considering using the JSON schema, perhaps it could look something like this:
> 
> ```
> workflowConfig:
>   openMetadataServerConfig:
>     securityConfig:
>       extraApiHeaders:
>       - name: SOME_HEADER
>         value: ${SOMETHING_FROM_ENV}
>       - name: Proxy-Authorization
>         valueFrom: Authorization
> ```
> 
> This would allow for setting headers from the environment using the `value` key or using the `valueFrom` key to copy the value from an existing header. Thoughts?

Thanks for the ping @czuares. I believe we could replicate the [connectionOptions](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/connectionBasicType.json#L7) approach we have for connections for extra headers.

Instead of adding this information inside `securityConfig` (in there we have the SSO details as external refs), we could have a new optional property with what you propose.

If you move on with the implementation, i'd also be great if you could add the relevant information in the Python SDK docs ([here](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-docs/content/sdk/python/index.md)) with your example use case.

We're now close to getting the 0.13.1 release out, so we'd be merging this PR in after getting that done to ensure we have enough testing time around it. This change would then be published in 0.13.2 in mid January.

Meanwhile, I am already marking your PR safe to test for you to get some CI checks running.

Thanks

yu-iskw:
I love the feature, because we secure our OpenMetadata with Google Cloud IAP. We would like to add extra request headers to authenticate a client.

pmbrull:
hi @czuares, let me know if you are available to get back on this or if you have any questions. Thanks!

czuares:
@pmbrull I'll try to make some time soon and reach out if/when I have some questions, thanks!

czuares:
@pmbrull Please re-review my changes when you get a chance. I have taken the approach you suggested and have a cleaner solution (in my opinion, at least 🙂)

pmbrull (approved):
Thanks for the changes! Having it in JSON schemas makes it way easier, thanks for the solution. LGTM

pmbrull:
Oh just saw the pylint failure `Too many local variables (16/15) (too-many-locals)` in the `ingestion/src/metadata/ingestion/ometa/client.py:130:4:`. Feel free to add a pylint ignore for now. We'll need to split some of those functionalities to clean it. Thanks

czuares:
@pmbrull could you please review again? I added the lint bypass and addressed a dictionary merge compatibility with python < 3.9.

The failing check appears to be unrelated.

pmbrull (approved):
Thanks! LGTM, indeed the cypress issues are not related to this. Thanks for the PR!

yu-iskw:
@czuares Thank you very much for implementing the feature. This is what I really want so that we make OpenMetadata secure with Google Cloud IAP. I look forward to that the feature will be released. 
https://cloud.google.com/iap/docs/authentication-howto

PR #9328, opened 2022-12-16
Fix: Metadata backup Azure blob issue

### LINKED ISSUE #4868: Support Azure blob for metadata backup dump
based on https://openmetadata.slack.com/archives/C02B6955S4S/p1652248191418669?thread_ts=1652243987.478259&cid=C02B6955S4S

Support Azure blob for the CLI metadata dump as well. We currently only support AWS S3.

MilanBariya:
### Describe your changes :

I worked on the ..... because ...ref [ fix #4868]

#
### Type of change :

- [x] Bug fix
- [x] Improvement

### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/cmd.py
+++ b/ingestion/src/metadata/cmd.py
@@ -210,15 +210,9 @@ def backup_args(parser: argparse.ArgumentParser):
         default=None,
     )
     parser.add_argument(
-        "--upload-s3",
+        "--upload",
         help="S3 endpoint, bucket & key to upload the backup file",
         nargs=3,
-        default=None,
-    )
-    parser.add_argument(
-        "--upload-azure",
-        help="Account-url, Container, Tenant-id, Client-id & Client-secret to upload the backup file",
-        nargs=5,
         default=None,
     )
     parser.add_argument("-o", "--options", default=None, action="append")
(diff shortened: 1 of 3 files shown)


ingestion/src/metadata/cmd.py
```diff
@@ -210,11 +210,17 @@ def backup_args(parser: argparse.ArgumentParser):
...
+        "--upload-s3",
         help="S3 endpoint, bucket & key to upload the backup file",
         nargs=3,
         default=None,
     )
+    parser.add_argument(
+        "--upload-azure",
+        help="Account-url, Container, Tenant-id, Client-id & Client-secret to upload the backup file",
+        nargs=5,
+        default=None,
```
ulixius9, line 222: when we try to upload the backup file to s3 we pickup the creds from like access key and key id from environment (basically user has to setup aws cli) can you check if a similar solution would work here instead of taking everything as argument

ulixius9 (commented):
left a comment, also can you update the docs as well

ingestion/src/metadata/cmd.py
MilanBariya: Done.

ulixius9 (approved):

PR #9383, opened 2022-12-18
DBT ingestion improvements

ulixius9:
### Describe your changes :

DBT ingestion improvements

#
### Type of change :

- [x] Bug fix
- [x] Improvement

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/dbt.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt.py
@@ -510,16 +510,18 @@ class DbtSource(DbtServiceSource):  # pylint: disable=too-many-public-methods
         logger.info(f"Processing DBT Query lineage for: {table_fqn}")
 
         try:
-            source_elements = table_fqn.split(".")
+            source_elements = fqn.split(table_fqn)
+            # remove service name from fqn to make it parseable in format db.schema.table
+            query_fqn = ".".join(source_elements[-3:])
             query = (
-                f"create table {table_fqn} as {data_model_link.datamodel.sql.__root__}"
+                f"create table {query_fqn} as {data_model_link.datamodel.sql.__root__}"
             )
             lineages = get_lineage_by_query(
                 self.metadata,
                 query=query,
-                service_name=source_elements[1],
-                database_name=source_elements[2],
-                schema_name=source_elements[3],
+                service_name=source_elements[0],
+                database_name=source_elements[1],
+                schema_name=source_elements[2],
             )
             for lineage_request in lineages or []:
                 yield lineage_request
--- a/ingestion/src/metadata/ingestion/source/database/dbt.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt.py
@@ -512,9 +512,9 @@ class DbtSource(DbtServiceSource):  # pylint: disable=too-many-public-methods
         try:
             source_elements = fqn.split(table_fqn)
             # remove service name from fqn to make it parseable in format db.schema.table
-            query_fqn = fqn._build(
+            query_fqn = fqn._build(  # pylint: disable=protected-access
                 *source_elements[-3:]
-            )  # pylint: disable=protected-access
+            )
             query = (
                 f"create table {query_fqn} as {data_model_link.datamodel.sql.__root__}"
             )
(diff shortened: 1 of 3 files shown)


ingestion/src/metadata/ingestion/source/database/dbt.py
```diff
@@ -510,16 +510,18 @@ def create_dbt_query_lineage(
         logger.info(f"Processing DBT Query lineage for: {table_fqn}")
 
         try:
-            source_elements = table_fqn.split(".")
+            source_elements = fqn.split(table_fqn)
+            # remove service name from fqn to make it parseable in format db.schema.table
+            query_fqn = ".".join(source_elements[-3:])
             query = (
-                f"create table {table_fqn} as {data_model_link.datamodel.sql.__root__}"
+                f"create table {query_fqn} as {data_model_link.datamodel.sql.__root__}"
```
ulixius9, line 517: service name should not in included while constructing this query otherwise it breaks [here](https://github.com/open-metadata/OpenMetadata/blob/1a00e81fe82c3b37627638e54800429dcce6d06d/ingestion/src/metadata/ingestion/lineage/sql_lineage.py#L131) with `too many values to extract` error

ingestion/src/metadata/ingestion/source/database/dbt_service.py
```diff
@@ -139,14 +131,9 @@ class DbtServiceSource(TopologyRunnerMixin, Source, ABC):
...
     def get_dbt_files(self) -> DbtFiles:
-        dbt_details = get_dbt_details(
+        dbt_files = get_dbt_details(
             self.source_config.dbtConfigSource  # pylint: disable=no-member
         )
-        dbt_files = DbtFiles(
-            dbt_catalog=dbt_details[0],
-            dbt_manifest=dbt_details[1],
-            dbt_run_results=dbt_details[2],
-        )
```
ulixius9, line 149: instead of expecting a tuple here directly returning the `DbtFiles` object to avoid any index out of bound error

ingestion/src/metadata/utils/dbt_config.py
```diff
@@ -80,15 +93,14 @@ def _(config: DbtLocalConfig):
...
+        return DbtFiles(
+            dbt_catalog=json.loads(dbt_catalog) if dbt_catalog else None,
+            dbt_manifest=json.loads(dbt_manifest),
+            dbt_run_results=json.loads(dbt_run_results) if dbt_run_results else None,
         )
     except Exception as exc:
         logger.debug(traceback.format_exc())
-        logger.warning(f"Error fetching dbt files from local: {exc}")
-    return None
+        raise DBTConfigException(f"Error fetching dbt files from local: {exc}")
```
ulixius9, line 103: now since its a separate workflow breaking the workflow if fetching files fails

ingestion/src/metadata/utils/dbt_config.py
```diff
@@ -117,15 +129,18 @@ def _(config: DbtHttpConfig):
             dbt_catalog = requests.get(  # pylint: disable=missing-timeout
                 config.dbtCatalogHttpPath
             )
-        return (
-            json.loads(dbt_catalog.text) if dbt_catalog else None,
-            json.loads(dbt_manifest.text),
-            json.loads(dbt_run_results.text) if dbt_run_results else None,
+        if not dbt_manifest:
+            raise DBTConfigException("Menifest file not found in file server")
```
ulixius9, line 133: also breaking the workflow if menifest file is not found in case of file server/ s3/gcs since its required to process dbt info

ingestion/src/metadata/ingestion/source/database/dbt.py
```diff
@@ -510,16 +510,18 @@ def create_dbt_query_lineage(
         logger.info(f"Processing DBT Query lineage for: {table_fqn}")
 
         try:
-            source_elements = table_fqn.split(".")
+            source_elements = fqn.split(table_fqn)
+            # remove service name from fqn to make it parseable in format db.schema.table
+            query_fqn = ".".join(source_elements[-3:])
```
pmbrull, line 515: Instead of using the `".".join(...)` we should use the methods in `fqn.py`. IMO we could rename the `_build` to `join` or something and use it in these scenarios to have proper quoting and centralising the dot logic

ingestion/src/metadata/ingestion/source/database/dbt_service.py
pmbrull: better, thank you

ingestion/src/metadata/utils/dbt_config.py
pmbrull: good catch

pmbrull (commented):
Thank you for the improvements, left a quick comment

pmbrull (approved):

PR #9403, opened 2022-12-19
UI: Data insight feedback part 6

### LINKED ISSUE #9357: Fix Tier % Computation
- [x] Percentage should only have 2 digits after the decimal point
- [x] Latest and Summary for "Total Data Assets by Tier" should not add the "NoTier" otherwise the percentage will 

Screenshot 2022-12-16 at 4 52 58 PM

ShaileshParmar11:
### Describe your changes :

UI: Data insight feedback part 6 
- Reduce the right panel width
- added more space in between the graph and right panel as well as in between the progress bar
- added % formatting in tier graph
- removed NoTier values from calculation to latest total and summary card(Closes #9357)
- updated Page views by data assets graph value title

#
### Type of change :

- [x] Improvement

#
image
image

#
### Checklist:

#


TeddyCr:
@ShaileshParmar11 on the second screenshot, is it expected to have the gauge with 22%, 14%, etc take the whole length of the gauge=? I would expect them to represent 14% of the total length of the gauge

openmetadata-ui/src/main/resources/ui/src/components/DataInsightDetail/TotalEntityInsight.tsx
```diff
@@ -129,8 +130,8 @@ const TotalEntityInsight: FC<Props> = ({ chartFilter, selectedDays }) => {
         </>
       }>
       {data.length ? (
-        <Row gutter={16}>
-          <Col span={18}>
+        <Row gutter={DI_STRUCTURE.rowContainerGutter}>
```
chirag-madlani, line 133: I can recommend another way to do so let's connect for the same.
Other then approach rest looks good to me

openmetadata-ui/src/main/resources/ui/src/utils/DataInsightUtils.tsx
```diff
@@ -486,8 +487,8 @@ export const getGraphDataByTierType = (rawData: TotalEntitiesByTier[]) => {
   });
 
   const graphData = prepareGraphData(timestamps, filteredData);
-  const latestData = getLatestCount(last(graphData));
-  const oldestData = getLatestCount(first(graphData));
+  const latestData = getLatestCount(omit(last(graphData), 'NoTier'));
```
chirag-madlani, line 490: localisation

chirag-madlani (approved):

openmetadata-ui/src/main/resources/ui/src/utils/DataInsightUtils.tsx
ShaileshParmar11: Hi @chirag-madlani we are just omitting key here

chirag-madlani (approved):

ShaileshParmar11:
> > @ShaileshParmar11 on the second screenshot, is it expected to have the gauge with 22%, 14%, etc take the whole length of the gauge=? I would expect them to represent 14% of the total length of the gauge
> 
> is the above comment addressed ^

Yes

PR #9413, opened 2022-12-20
Fix : Advanced search improvements

Sachin-chaurasiya:
### Describe your changes :

I worked on advanced search improvements
- Updated the advanced search dropdown to show the user/team display name or name.
- Updated service query field constants to show results based on the current index
- Added logic to reset the applied filter text on tab change

Test Cases:
- If the default option is present then it should be shown on click of the select box
- If no default option is present then it should not render the option menu
- Value dropdown should show related values to the current index
    - For example, if the user selects the service field and the current index is `topic_search_index`, then the value dropdown should show only service related to the topic entity.
- Applied filter text should be reset on tab change.

#
### Type of change :

- [x] Improvement

#
image

Default Values:
image

image

#
### Checklist:

#
### Reviewers
@open-metadata/ui


chirag-madlani:
@Sachin-chaurasiya can we add suggest changes in this one too? Initially it should show few options before start searching

Sachin-chaurasiya:
> @Sachin-chaurasiya can we add suggest changes in this one too? Initially it should show few options before start searching

@chirag-madlani, Sure, let me check on that part.

Sachin-chaurasiya:
> @Sachin-chaurasiya can we add suggest changes in this one too? Initially it should show few options before start searching

Done.

chirag-madlani (approved):
Can you mention test cases that you have covered?

Sachin-chaurasiya:
> Can you mention test cases that you have covered?

@chirag-madlani, I have Updated the test cases, please check the PR description.

PR #9418, opened 2022-12-20
Fix:minor changes in atlas

NiharDoshi99:
### Describe your changes :

I worked on making a minor changes in atlas

#
### Type of change :

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/ingestion/src/metadata/ingestion/source/metadata/atlas.py
+++ b/ingestion/src/metadata/ingestion/source/metadata/atlas.py
@@ -266,6 +266,13 @@ class AtlasSource(Source):
                         entity=Table, fqn=table_fqn
                     )
 
+                    tag_fqn = fqn.build(
+                        self.metadata,
+                        entity_type=Tag,
+                        tag_category_name=ATLAS_TAG_CATEGORY,
+                        tag_name=ATLAS_TABLE_TAG,
+                    )
+
                     if tbl_attrs.get("description", None) and table_object:
                         self.metadata.patch_description(
                             entity_id=table_object.id,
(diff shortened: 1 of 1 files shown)


pmbrull (dismissed):

ingestion/src/metadata/ingestion/source/metadata/atlas.py
```diff
@@ -266,6 +266,13 @@ def _parse_table_entity(self, name, entity):  # pylint: disable=too-many-locals
                         entity=Table, fqn=table_fqn
                     )
 
+                    tag_fqn = fqn.build(
+                        self.metadata,
+                        entity_type=Tag,
+                        tag_category_name=ATLAS_TAG_CATEGORY,
+                        tag_name=ATLAS_TABLE_TAG,
+                    )
+
```
ayush-shah, line 275: add a top condition checking if table object is there..
if not, we do nothing, show warning

if we have table_object,
then we check for description,
complete that,

create tag_fqn, and then patch_tags

ayush-shah (changes requested):

ingestion/src/metadata/ingestion/source/metadata/atlas.py
ayush-shah: this way, we make less calls to the ES, we check for description, if not there, we create tag fqn (call to es) and then patch_tags

NiharDoshi99: Oh ok, let me do that.

ayush-shah (approved):

PR #9506, opened 2022-12-22
Fix#9448: Add ES volumes

### LINKED ISSUE #9448: Add volumes for Elastic search service

nahuelverdugo:
Fixes #9448

### Describe your changes :
Running docker from CLI failed because of the missing parameter `--project-directory .` to use the created directory `docker-volume`. Instead, it was using the directory of the `docker-compose` file. 

We have updated the library to communicate Python and Docker into a version that supports that parameter. In addition, we have also updated our script to start docker when testing. 

### Type of change :
- [x] Improvement

### Checklist:

### Reviewers


pmbrull (approved):
Thanks for the fixes. Good point adding the context

PR #9519, opened 2022-12-23
Support for pluggable customizer for Elasticsearch mappings creation.

tyano:
### Describe your changes :

Current OpenMetadata have static Elasticsearch(ES) mappings, but the search function doesn’t work correctly if the searching word contains Japanese characters.

A similar problem is reported on  #9158, on which language-base mapping is supported, and contains mappings for Japanese language.

The patch of #9158 uses Kuromoji-analyzer for supporting Japanese language, but Analyzers used for Japanese not only Kuromoji-analyzer, Ngram-Analyzer is used often too. In my experience, Ngram is used in many web systems, because both of the analyzer have it own profit and cons. Kuromoji is useful for Dictionary-based tokenization, but Ngram is useful for free-word searching.

So choice of analyzer is depend on usecase, not on language. For example, Ngram is useful not only for Japanese, but it will useful and be matched for searching on some asian languages like Chinese or Korean.

And this problem is not able to solve only with mapping files.
If we use Ngram-analyzer, we need 2 customizations on a query:

* we need to add a `{"type": "phrase" }` option to `query_string` query of Elasticsearch for best matching
* Wildcard query doesn't work correctly on fields tokenized by ngram-tokenizer. Matching to ngram-field is always partial matching (not exact matching), so query string `*Word*` doesn't match. It must be just `Word`. So the query string sent from OM must be `(Word OR *Word*)`, not `*Word*`.

For supplying those customization onto query, we need functionality for customize QueryBuilder and query string.

I introduced a new `ElasticSearchIndexResolver` interface and made `DefaultElasticSearchIndexResolver` (which works same with current OM implementation) and `NgramElasticSearchIndexResolver` which uses Ngram-analyzer on description-fields and customize query for supporting ngram-matching. This interface is used for getting index-name, getting path of mapping file, customize query string, customize QueryBuilder and customize HighlightBuilder.

(HighlightBuilder customization might be required if new mapping files contain some new fields and is needed to add the field to highlight, but it is not used in this patch.)

Implementation-class is changeable by a setting `elasticsearch.indexResolverClassName` in openmetadata.yml.

The advantage of this patch is that users can supply their own mapping files and customization by creating a jar file which contains their own implementation of `ElasticSearchIndexResolver` and putting the jar in classpath. Users can customize query by their own usecase, not by language.

This patch is now used in our company as trial and currently works perfectly.

FYI: Before pull-request, I duscussed at Slack channel: https://openmetadata.slack.com/archives/C02DZK3QH8Q/p1671693745104339

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

When I run all tests from maven, some tests for *Resource.class which uses TestContainer show errors, but it works successfully when I run it from IDE.

#
### Reviewers
@open-metadata/backend


harshach:
Thanks @tyano for the PR. Will review and merge it in.

harshach:
@tyano do we want to maintain 2 separate index mappings. Why not merge it into one index with ngrams enabled?

harshach:
@tyano can you please fix the styling check. Will merge it in

tyano:
Fixed merge conflict with main branch and executed the stylers.

tyano:
I found some mistakes in my patch. I will fix them soon. Please don't merge until I fix it.

tyano:
This PR now conflict with #9158 , so I'm trying my code to the latest main branch.

I tried #9158 on my server and noticed that we can get search-results with the word '回答', but can not get results with '答', although '答' is a part of '回答'.
It is because #9158 uses Kuromoji analyzer for creating search-index, and kuromoji tokenized sentences by dictionary. it create a token '回答' because dictionary have the word, and never created a token '答'.

I think Ngram tokenizer will solve this problem...

harshach:
@tyano trying to merge this patch . Can you run 
```  mvn googleformatter:format@reformat-sources ``` to fix the styling check

Also I am thinking we should make ngram as the default behavior, wdyt?

**Comment by harshach** [comment]
@tyano can you fix the merge commit error that I did. I can't push changes into your branch

1. Indexed with Ngram but kept getting errors for sample data
2. 
image

Did you try testing  with sample data with ngram enabled?

**Comment by tyano** [comment]
I'll check them.

**Comment by tyano** [comment]
Found some faults.

1. We must install some plugins (analysis-icu and analysis-kuromoji) to Elasticsearch. I built my own Docker image which contains the plugins for testing. To fix this problem I will put a new Dockerfile for building the Elasticsearch.

2. The image described above must be pushed to your docker repository for making it publicly downloadable and change the image name in docker-compose.yaml and docker-compose-postgres.yaml. 
  I can not handle it, because the build process is a part of your company's release pipeline and you must supply the image name of the image which is built by your release pipeline. 
  
3. Searching english word doesn't work correctly on Ngram index. I'm now working on this problem (maybe problems on mapping files). After solving this problem, I will push new code again.

**Comment by tyano** [comment]
I pushed new fixed code.
There are some notices:

### I added a new Dockerfile: `docker/local-metadata/Dockerfile_elasticsearch`. 

This dockerfile generates an Image of Elasticsearch containing analysis-icu and analysis-kuromoji, which are required by NGRAM mapping files.

For local use, you can just use `docker/local-metadata/docker-compose.yaml` or `docker/local-metadata/docker-compose-postgres.yml`. I changed those files as building the new Dockerfile_elasticsearch, instead of downloading the image from elastic.co.

After merging this PR, you must build the new ES image and push it to your public docker registry, then must change public `docker-compose.yaml` and `docker-compose-postgres.yaml` as using the new built ES image.

Without the new ES image, users will be faced errors like screenshots you put in above posts.

### HighlightBuilder for NGRAM is changed

I noticed that searching by alphabet word didn't work correctly. When searching english word, the line like "Matches : n in Columns Description" on the result page showed wrong match count (like '70 in Column Description', despite of no such many columns exist).

It was caused by `function_score` query used from version 0.13.x. Using `function_score` query to NGRAMed fields breaks the highlights of the returned match-result.
Investigating this problem, I found this problem was easily avoidable by adding `highlight_query` onto HightlightBuilder. So I implemented `customizeHighlight` method of NgramElasticSearchIndexResolver as adding `highlight_query`.

This customization is needed only the time when `table_search_index` index is the search target. For distinguish indexes, I Added an `indexType` parameter in some part of code. 

### about Making Ngram as default

Making ngram as the default behavior seems good idea. Ngram works for both of basic space-separated language like english and no-separator-language like Chinese and Japanese. Ngram will be useful for many languages.

But I think changing index to ngram is a large change. Public testing is required.
If we use ngram as the default behavior in 0.13.x, we must describe the function is easily switchable by setting `indexResolverClassName` in `openmetadata.yaml`, so that users can change it to old default behavior if they face problems caused by the new analyzer.

**Comment by harshach** [comment]
@tyano apologies for the delay in reviewing the PR. Can you please fix the merge conflicts so we can push this for the 1.0 release. 
Thanks

**Comment by tyano** [comment]
OK, I will try to merge the latest main branch. Please wait a few days. I will do it tomorrow or next tomorrow.

**Comment by tyano** [comment]
@harshach Now I'm trying to merge the latest main branch to my PR. After merging, an error occurs when I'm deploying a Metadata Ingestion after adding a database service.

I tested on `main` branch too, for checking if the error occurs for merge errors, but same error occurred on `main` branch too.

It says: 

```
expert-db_metadata_Q1OSY4KX Failed to deploy Ingestion Pipeline due to airflow API returned Bad Request and response {"error": "Request Validation Error parsing payload. IngestionPipeline expected: 78 validation errors for IngestionPipeline
```
See an screen shot added on this post.

Could you guess what is happening? It might be an merge issue or 



スクリーンショット 2023-02-24 19 09 54

harshach:
@ulixius9 any idea about the error above?

ulixius9:
@tyano can you confirm if you have built the ingestion container after merging the main branch? from the error I can see that you are facing a json schema related issue related to processPIISensitive which we have recently introduced with this [pr](https://github.com/open-metadata/OpenMetadata/pull/10119/files#diff-7cafce473bd53e502056f9f869f2248af495e509d4a6388314a557198aed5347) it seems like the server has picked up this schema change but the ingestion container has not picked this up, If you are testing your changes on docker you need to run the `make generate` command first from root dir and then build the docker image!

tyano:
Ah, You right! I didn't run `make generate`.
I will try it.
Thank you.

tyano:
merged.

Be careful that a new ES image is introduced. 
The image will be automatically built if you use `local-metada/docker-compose.yml` or `local-metada/docker-compose.yml`   for local development. You need to build the image in you build pipeline while releasing it.

See: https://github.com/open-metadata/OpenMetadata/pull/9519#issuecomment-1396857867

tyano:
@harshach I noticed new merge conflicts appear between my branch and main branch, so I resolved the conflict and merged it again.

Be careful you need to set an environment variable `ELASTICSEARCH_INDEX_RESOLVER_CLASS_NAME` to `org.openmetadata.service.elasticsearch.indexresolver.NgramElasticSearchIndexResolver`, or set an `indexResolverClassName` config parameter in `openmetadata.yaml` for using Ngram index.

harshach:
we have introduce per language specific mappings. Closing this as resolved

Issue #5592, opened 2022-06-23
Ingest Databricks jobs


pmbrull:
Review the information we can retrieve from databricks jobs and ingest them as pipeline entities

pmbrull:
This basically means having Databricks as a Pipeline Connector as well

harshach:
@pmbrull there is a request about adding the queries from databricks sql

pmbrull:
hi @harshach I see, thanks. For query history we could use something like https://docs.databricks.com/sql/api/query-history.html# and manage it as a usage and lineage connector. I will open a separated issue for that and keep this one for the Pipeline Entities.

pmbrull:
https://github.com/open-metadata/OpenMetadata/issues/7376

pmbrull:
closed by https://github.com/open-metadata/OpenMetadata/pull/9554

PR #9554, opened 2022-12-31
Issue 5592: Add Databricks Pipeline Connector

sjrusso8:
### Describe your changes :

Fix issue #5592 

Create a new pipeline connection to interface with the [Databricks Jobs 2.1 API](https://docs.databricks.com/dev-tools/api/latest/jobs.html) and gather all the pipeline details. 

This change extends the existing databricks client to gather all the created pipelines in a workspace. The connector will then gather all available pipeline statuses for a specific pipeline id.

#
### Type of change :

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


ingestion/src/metadata/ingestion/source/database/databricks/client.py
```diff
@@ -122,3 +126,78 @@ def is_query_valid(self, row) -> bool:
             query_text.startswith(QUERY_WITH_DBT)
             or query_text.startswith(QUERY_WITH_OM_VERSION)
         )
+
+    def list_jobs(self) -> List[dict]:
```
pmbrull, line 130: Not sure if using `dict` as a typing here can have an impact on Python 3.7 🤔 I know we had some issues before with hints as `dict[...]`, so if you could please double-check that there won't be any hidden bug 🙏

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
```diff
@@ -0,0 +1,222 @@
...
+    "CANCELED": StatusType.Failed,
+    "PENDING": StatusType.Pending,
+    "RUNNING": StatusType.Pending,
+    "TERMINATING": StatusType.Pending,
+    "SKIPPED": StatusType.Failed,
+    "INTERNAL_ERROR": StatusType.Failed,
+}
+
+TASK_MAP = {
+    "notebook_task": "Notebook_Task",
```
pmbrull, line 59: Should we bring the values into the JSON Schema for Databricks pipelines to have a closed list of supported tasks?

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
```diff
@@ -0,0 +1,222 @@
...
+class DatabrickspipelineSource(PipelineServiceSource):
+    """
+    Implements the necessary methods ot extract
+    Pipeline metadata from Databricks Jobs API
+    """
+
+    def __init__(self, config: WorkflowSource, metadata_config: OpenMetadataConnection):
+        super().__init__(config, metadata_config)
+        self.task_id_mapping = {}
+        self.job_id_list = set()
```
pmbrull, line 79: Instead of setting the `self.job_id_list` here maybe we can use the `context` variable to handle the information. Might be a bit more explicit to help us know it is dependent on the execution state

pmbrull: You can directly play with the context like this https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/api/topology_runner.py#L142

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
```diff
@@ -0,0 +1,222 @@
...
+        Get Pipeline Name
+        """
+        return pipeline_details["settings"]["name"]
+
+    def yield_pipeline(self, pipeline_details: Any) -> Iterable[CreatePipelineRequest]:
+        """
+        Method to Get Pipeline Entity
+        """
+        self.job_id_list = set()
+        pipeline_ev = CreatePipelineRequest(
```
pmbrull, line 114: maybe we want to try/catch this call. I see we also have the `get_tasks` function in there without handling exceptions. This could break the workflow execution. Usually it's ok to skip an asset of being ingested, but we don't want the whole workflow to die

pmbrull (commented):
Thanks @sjrusso8 for opening the PR! Also, thanks for adapting it to the new format in the repo. We just took the chance of refactor a couple of things and I see you already amended the changes here, thank you for that.

Overall this looks great, if you could just review the failing tests we can move it forward (maven CI and pytests). We are currently fixing some cypress failures so we can disregard these for now.

Left a few quick comments if you have the time.

Thanks!

ingestion/src/metadata/ingestion/source/database/databricks/client.py
sjrusso8: I did check the [Python 3.7 typing docs](https://docs.python.org/3.7/library/typing.html#typing.Dict). It seems that you need to use `typing.Dict` when you are also annotating the arguments, so like `Dict[str, str]`. So I think just `dict` without the annotated arguments should be fine.

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
sjrusso8: Updated with a try/catch. I mirrored the approach used in the AirflowSource pipeline service.

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
sjrusso8: So I just went ahead and removed the TASK_MAP. I only really included it to make the formatting look nicer in the UI.

Instead, I updated this process just to find the key named like "*_task" in the task details for a pipeline. You can only have one task type for each task, and I am assuming that all future task types are going to have the same naming convention.

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
sjrusso8: Good call! I updated the service to use the `context` variable

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
pmbrull: thanks, then we might need to improve the process in airflow too hehe. In Airflow I feel a bit more comfortable as the data we pick up is from a db instead of an API, so we have a bit more safety once the connection is done. In any case, it should be handled

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
pmbrull: thanks for updating!

The context is a pydantic BaseModel that allows arbitrary additions whenever it is used https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/models/topology.py#L83

We use it in two ways:
1. Define the hierarchy of assets that are needed during the ingestion and keep a snapshot of "where we are" in the execution. E.g., to ingest a table, I first need to know its schema, database, and service. This information is embedded in the topology of the service itself in the `context` key https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/database/database_service.py#L95
2. We pass parameters on the fly that are needed in different steps of the execution (example https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/database/dbt/metadata.py#L290). To access/update the values you can either use it by name as your typical pydantic class `context.name` or you can jump straight into its `dict` structure if you need dynamism of some sorts `context.__dict__["name"]`. Here you could set `self.context.job_id_list = []` and then keep appending on the existing `self.context.job_id_list` during the loop. To pick it up in the status, just call `self.context.job_id_list`.

Hope this helps, this part is not really clear in the tutorials or docs as it's a rather internal detail so that's on us.

Feel free to keep asking until you have all the info you need.

Thanks!

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
pmbrull: sounds good, thanks for explaining

ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/metadata.py
```diff
@@ -0,0 +1,231 @@
...
+            depends_on = task.get("depends_on")
+            if depends_on:
+                dependent_tasks[task["task_key"]] = [v["task_key"] for v in depends_on]
+            else:
+                dependent_tasks[task["task_key"]] = None
+
+        return dependent_tasks
+
+    def yield_pipeline_status(self, pipeline_details) -> Iterable[OMetaPipelineStatus]:
+        for job_id in self.context["job_id_list"]:
```
pmbrull, line 191: Last comment 🙏 

Is this subscription correct? or should this become `self.context.job_id_list`?

sjrusso8: Updated. I was using the `self.context` variable incorrectly. Your other comment was really helpful!

pmbrull (approved):
Thanks for the implementation, looks great. Congrats on the PR 🎉

Issue #3582, opened 2022-03-22
get_column_type refactor


pmbrull:
**Affected module**
Ingestion

**Describe the bug**
It is quite hard to follow and debug what is going on in get_column_type at column_type_parser. Let's break the function down into smaller functions with meaningful names, comments, and tests. We can refactor the protected functions of the module later on, but this one gets directly called by the workflow so we should first focus on this.

It seems to me that there are 3 main steps involved. Convert each of them in a function and document the order of "checks" that we are doing.

Thanks!

ayush-shah:
Hey @pmbrull, Thanks for creating this issue.
Will take it up in 0.9.2 release

pmbrull:
Great, thanks @ayush-shah, no rush at all :) Let's just try to list down central code pieces that we could improve. Thanks 🙏

harshach:
@pmbrull @ayush-shah do we need to do anything here?

harshach:
@pmbrull @ayush-shah what is remaining here

ayush-shah:
Refactoring the code, breaking it down in functions,
Writing tests and adding upto 100% test coverage around here.
Will be part of 0.13

harshach:
@ayush-shah is this targetted for 0.13?

ayush-shah:
yes

PR #9559, opened 2023-01-02
Fix: refactor get_column_type

NiharDoshi99:
### Describe your changes :

I worked on refactoring `get_column_type`
Fix : #3582  

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/column_type_parser.py
+++ b/ingestion/src/metadata/ingestion/source/database/column_type_parser.py
@@ -213,17 +213,31 @@ class ColumnTypeParser:
 
     @staticmethod
     def get_column_type(column_type: Any) -> str:
-        if not ColumnTypeParser._COLUMN_TYPE_MAPPING.get(type(column_type)):
-            if not ColumnTypeParser._SOURCE_TYPE_TO_OM_TYPE.get(str(column_type)):
-                if not ColumnTypeParser._SOURCE_TYPE_TO_OM_TYPE.get(
-                    str(column_type).split("(", maxsplit=1)[0].split("<")[0].upper()
-                ):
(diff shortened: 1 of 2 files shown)


ingestion/src/metadata/ingestion/source/database/column_type_parser.py
```diff
@@ -213,17 +213,31 @@ class ColumnTypeParser:
...
-                ):
-                    return ColumnTypeParser._SOURCE_TYPE_TO_OM_TYPE.get("VARCHAR")
-                return ColumnTypeParser._SOURCE_TYPE_TO_OM_TYPE.get(
-                    str(column_type).split("(", maxsplit=1)[0].split("<")[0].upper()
-                )
-            return ColumnTypeParser._SOURCE_TYPE_TO_OM_TYPE.get(str(column_type))
-        return ColumnTypeParser._COLUMN_TYPE_MAPPING.get(type(column_type))
+
+        if ColumnTypeParser.get_column_type_mapping(column_type):
+            return ColumnTypeParser.get_column_type_mapping(column_type)
```
ulixius9, line 218: we are calling the same function twice here, can be avoided

ulixius9 (commented):
Thanks, looks much cleaner also let's add some test cases!

pmbrull (commented):
let's take the chance to add some unit tests around here. 

Ideally when working on functions refactor, we should create the unit test first, then refactor and validate that tests are still working as they were.

It's fine to not go back now, but at least let's increase the coverage here 🙏

pmbrull (approved):

ulixius9 (approved):

PR #9564, opened 2023-01-02
Fix #9562: Add duration in usage ingestion

### LINKED ISSUE #9562: Store execution time while usage ingestion
**Affected module**
Ingestion Framework

**Describe the bug**
Store execution time while usage ingestion

ulixius9:
### Describe your changes :

Fix #9562: Add duration in usage ingestion

#
### Type of change :

- [x] Bug fix

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion


ingestion/src/metadata/ingestion/source/database/databricks/usage.py
```diff
@@ -102,6 +102,9 @@ def process_table_query(self) -> Optional[Iterable[TableQuery]]:
                                 endTime=row.get("execution_end_time_ms"),
                                 analysisDate=datetime.now(),
                                 serviceName=self.config.serviceName,
+                                duration=row.get("duration") / 1000
+                                if row.get("duration")
+                                else None,
```
ayush-shah, line 107:
```suggestion
                                duration=row.get("duration", 0) / 1000
                                or None,
```

ayush-shah, line 107: will it be possible to edit the sqlalchemy query and get it divided by 1000 there itself

ulixius9: for databricks we are not executing a query to fetch query history instead we are hitting a rest api

ulixius9: actually I made do it this way voluntarily because it's possible for `row.get("duration")` to be 0 and `0 or None` will return None so could be a misinformation similarly if we just do `row.get("duration",0)/1000` then if duration is not available then by default it will show 0 as duration which again could be a misinformation.


WDYT?

ingestion/src/metadata/ingestion/source/database/postgres/query_parser.py
```diff
@@ -154,6 +154,7 @@ def process_table_query(self) -> Optional[Iterable[TableQuery]]:
                                 databaseName=self.get_database_name(row),
                                 serviceName=self.config.serviceName,
                                 databaseSchema=self.get_schema_name(row),
+                                duration=row.get("duration"),
```
OnkarVO7, line 157: It might make sense to typecast this into `int` since the field is an `int` field

ayush-shah: +1

OnkarVO7 (approved):

Issue #8961, opened 2022-11-23
Ingestion doesn't retrieve meta description from MSSQL objects including database, table, and column


chenteddybear:
**Affected module**
Does it impact the UI, backend or Ingestion Framework?

**Describe the bug**
Getting this warning message "WARNING - Table description error for table"  and ingestion doesn't retrieve any meta description from MSSQL , including database, table, and column

**To Reproduce**
Please see the attached screenshots

**Expected behavior**
OpenMetadata to show meta description of db objects from it mssql source

**Version:**
 - OS: ubuntu 20.04
 - Python version: 3.8.10
 - OpenMetadata version: OpenMetadata 0.12.3-Release

![image](https://user-images.githubusercontent.com/33238040/203477945-9f1b31c3-ff7c-48ff-ad89-f3b006128416.png)
![image](https://user-images.githubusercontent.com/33238040/203478113-bc95ce05-47f4-4d3d-bad4-a3188c16ffa3.png)

harshach:
@MilanBariya any update on this?

PR #9584, opened 2023-01-03
Fix: Table description error for table

MilanBariya:
### Describe your changes :

I worked on adding a Table description error for the table.
Fix: #8961

#
### Type of change :

- [x] Bug fix
- [x] Improvemeant
Example of table comments
Screenshot 2023-01-04 at 3 16 09 PM
Example of column comments
Screenshot 2023-01-04 at 3 14 36 PM

#
### Checklist:

#


ingestion/src/metadata/ingestion/source/database/mssql/metadata.py
```diff
@@ -27,10 +31,33 @@
...
+    cursor = connection.execute(
+        MSSQL_GET_TABLE_COMMENTS.format(schema_name=schema_name, table_name=table_name)
+    )
+    try:
+        for result in cursor:
+            if result[1]:
+                return {"text": result[1]}
+    except Exception:
+        return {"text": None}
+    return {"text": None}
```
pmbrull, line 55: This could be simplified to something like

```python
    try:
        for result in cursor:
            if result[1]:
                return {"text": result[1]}
    except Exception as exc:
        logger.error("something happened blabla")

    return {"text": None}
```

pmbrull: No need to have so many returns. Also, maybe instead of logger.error a logger.debug could be useful here to know if we are missing descriptions for a valid reason 🤔

MilanBariya: Sounds good @pmbrull 
i will do that.

Issue #9402, opened 2022-12-19
Profiler: Handle Rows for Datalake Profiling


ayush-shah:
Handle the rows input by user and profile that amount of rows

Issue #9400, opened 2022-12-19
Rows - Randomize output


ayush-shah:
Add logic when ROWS to randomize the output.

harshach:
@ayush-shah can you add more details

ayush-shah:
Profiler when profiles based on the given rows limit..

adding a randomization logic to fetch new random data from the source should give us a better profiling

PR #9595, opened 2023-01-04
Profile sample rows + Randomization of chunks for datalake

ayush-shah:
### Describe your changes :

fixes https://github.com/open-metadata/OpenMetadata/issues/9648
fixes https://github.com/open-metadata/OpenMetadata/issues/9402
fixes https://github.com/open-metadata/OpenMetadata/issues/9400

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


NiharDoshi99 (dismissed):

ingestion/src/metadata/orm_profiler/profiler/sampler.py
```diff
@@ -68,7 +69,9 @@ def get_sample_query(self) -> Query:
                 .cte(f"{self.table.__tablename__}_rnd")
             )
         return (
-            self.session.query(self.table)
+            self.session.query(
+                self.table, ModuloFn(RandomNumFn(), 100).label(RANDOM_LABEL)
+            )
```
TeddyCr, line 74: Are you able to give more details as to what this is solving?

ingestion/src/metadata/orm_profiler/profiler/sampler.py
```diff
@@ -118,10 +121,17 @@ def fetch_sqa_sample_data(self) -> TableData:
             .limit(self.sample_limit)
             .all()
         )
+        rows = [list(row) for row in sqa_sample]
+        for col_index, col in enumerate(sqa_columns):
+            if type(col.type) in [sqltypes.LargeBinary, sqltypes.BINARY]:
+                rows = [
+                    [f"{r}" if index == col_index else r for index, r in enumerate(row)]
+                    for row in rows
+                ]
```
TeddyCr, line 130: Instead of modifying this at runtime by iterating over the data, we should probably create a custom type and handle the conversion of the data type there -- see [here](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/orm_profiler/orm/types/bytea_to_string.py) for an example

ingestion/src/metadata/utils/helpers.py
```diff
@@ -29,6 +30,10 @@
 logger = utils_logger()
 
 
+def random_int(min_val=0, max_val=999):
```
TeddyCr, line 33: I am not sure it is a good idea to have default value. Picking 0 and 999 is arbitrary to me and it makes code less readable (as there is no real logic).

TeddyCr (changes requested):
I added a few comments and 1 question. Thanks for doing this.

ingestion/src/metadata/utils/helpers.py
ayush-shah: reverted the above change, thanks

ayush-shah:
@TeddyCr resolved the comments. Thanks

ingestion/src/metadata/orm_profiler/orm/converter.py
```diff
@@ -52,7 +52,7 @@
     DataType.CHAR: sqlalchemy.CHAR,
     DataType.VARCHAR: sqlalchemy.VARCHAR,
     DataType.BOOLEAN: sqlalchemy.BOOLEAN,
-    DataType.BINARY: sqlalchemy.LargeBinary,
+    DataType.BINARY: CustomTypes.BINARY.value,
```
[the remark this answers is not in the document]
ayush-shah: So the row has largebinary data 
and python interprets it as bytes
under Bigquery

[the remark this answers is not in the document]
ayush-shah: it's binary and largebinary datatype, from bytes_table in bigquery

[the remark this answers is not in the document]
TeddyCr: The bytes table has bytes. If it get the binary type in OM then the problem lies in the mapping. 

Screenshot 2023-01-11 at 3 37 40 PM

ingestion/src/metadata/orm_profiler/profiler/sampler.py
```diff
@@ -67,8 +67,13 @@ def get_sample_query(self) -> Query:
...
                 .cte(f"{self.table.__tablename__}_rnd")
             )
+        table_query = self.session.query(self.table)
         return (
-            self.session.query(self.table)
+            self.session.query(
+                self.table,
+                (ModuloFn(RandomNumFn(), table_query.count())).label(RANDOM_LABEL),
+            )
+            .order_by(RANDOM_LABEL)
```
TeddyCr, line 76: How would that work for row sampling?

ayush-shah: self.session.query(
                self.table,
                (ModuloFn(RandomNumFn(), table_query.count())).label(RANDOM_LABEL),
            )
            .order_by(RANDOM_LABEL)

this will help with random row sampling, 
everytime random number is generated for rows,
it get’s ordered by RANDOM_LABEL col that was generated.

it will be something like:

        row1, 4
        row2, 1
        row3, 2
        row4, 3

if we add profile sample as 3
it gets ordered like this:

        row2, 1
        row3, 2
        row4, 3
        row1, 4

then we apply limit i.e. 3

        row2,1
        row3,2
        row4,3

and post that,
we have rnd.c.random <= 3
sampled = session_query.where(rnd.c.random <= self.profile_sample).cte(
            f"{self.table.__tablename__}_sample"
        )
the sampled cte gets created with the above 3 rows for profile sampling

TeddyCr: Wouldn't it change the behavior of the sampling by percent though? Sorry if I am missing something here.

ayush-shah: Sampling with percentage is taken care of from Line 59 - 69

ingestion/src/metadata/orm_profiler/orm/converter.py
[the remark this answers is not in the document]
ayush-shah: bytes from bigquery are currently being mapped to
"BYTES": sqlalchemy.types.BINARY,
and from our source code,
 we interpret the binary as large binary
DataType.BINARY: sqlalchemy.LargeBinary,

ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
```diff
@@ -55,13 +57,29 @@
 from metadata.ingestion.source.connections import get_connection
 from metadata.ingestion.source.database.column_type_parser import create_sqlalchemy_type
 from metadata.ingestion.source.database.common_db_source import CommonDbSourceService
+from metadata.orm_profiler.orm.registry import CustomTypes
 from metadata.utils import fqn
 from metadata.utils.filters import filter_by_database
 from metadata.utils.logger import ingestion_logger
 
+
+class Bytes(_Binary, TypeEngine):
```
ayush-shah, line 66: _Binary as super class was needed here in order to as bigquery sqlalchemy treats bytes as largebinary and _Binary is part of the largebinary..

TeddyCr: @ayush-shah what I meant here is in `converter.py` we need to change the `DataType.BINARY` to `CustomTypes.BYTES.value`.

ayush-shah: that didn't work as expected..

TeddyCr: I just tested it on my end and it worked

ayush-shah: so the above change will take care of the data type issue where
Bigquery shows Bytes datatype, but it used to change it to binary on bigquery sqlalchemy during metadata ingestion in OM.
so I overrided it to create a new Data Type just for bigquery which will only change the return type here..

TeddyCr: with this approach we'll have the same issue with all the DB that are mapping bytes to binary (if that makes sense).

ayush-shah: got it, I will modify it as per above, but the column's data type will revert to binary ( which is bytes on bigquery ), 

I was thinking, it might affect other sources with datatype as binary, `DataType.BINARY: CustomTypes.BYTES.value` will affect other DBs with binary type, all the binary data types will have to go through the hextobyte conversion..

TeddyCr: that should be ok. we can tests with other databases that have a binary type

TeddyCr (approved):
LGTM

ayush-shah:
Merging as the failing checks are not related to this PR

PR #9613, opened 2023-01-04
Fix #6829: Improve usage filtering

### LINKED ISSUE #6829: Validate if we are using filterPatterns in usage/lineage
If not, we can add them if it makes sense or directly remove those

ulixius9:
### Describe your changes :

Fix #6829: Improve usage filtering

#
### Type of change :

- [x] Improvement

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/clickhouse/query_parser.py
+++ b/ingestion/src/metadata/ingestion/source/database/clickhouse/query_parser.py
@@ -15,7 +15,9 @@ Clickhouse usage module
 import ast
 from abc import ABC
 from datetime import datetime
+from typing import List
 
+from metadata.generated.schema.entity.data.database import Database
 from metadata.generated.schema.entity.services.connections.database.clickhouseConnection import (
     ClickhouseConnection,
 )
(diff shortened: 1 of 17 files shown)


ingestion/src/metadata/ingestion/source/database/redshift/queries.py
```diff
@@ -22,7 +22,6 @@
     SELECT *
       FROM pg_catalog.stl_query
      WHERE userid > 1
-          {filters}
```
pmbrull, line 25: if we remove the filter from here, could it increase the data fetched on the CTE? Now I'm not sure if we could push down the filters by adding them as soon as possible

ingestion/src/metadata/ingestion/source/database/query_parser_source.py
```diff
@@ -58,8 +61,31 @@ def __init__(self, config: WorkflowSource, metadata_config: OpenMetadataConnecti
 
     def prepare(self):
         """
-        By default, there's nothing to prepare
+        Fetch queries only from DB that is ingested in OM
```
pmbrull, line 64: nice one, thank you

ingestion/src/metadata/ingestion/source/database/query_parser_source.py
```diff
@@ -58,8 +61,31 @@ def __init__(self, config: WorkflowSource, metadata_config: OpenMetadataConnecti
...
+        if self.database_field and database_name_list:
+            self.filters += (  # pylint: disable=no-member
+                f" AND {self.database_field} IN ('"
+                + "','".join(database_name_list)
+                + "')"
+            )
+
+        if self.schema_field and schema_name_list:
+            self.filters += (  # pylint: disable=no-member
+                f" AND {self.schema_field} IN ('" + "','".join(schema_name_list) + "')"
```
ulixius9, line 87: In this approach, we are fetching all the database schema and database names  and creating a filter, that will be applied while fetching query history logs!

ingestion/src/metadata/ingestion/source/database/clickhouse/query_parser.py
```diff
@@ -62,6 +64,29 @@ def get_sql_statement(self, start_time: datetime, end_time: datetime) -> str:
...
+            if self.schema_field and database.databaseSchemas:
+                for schema in database.databaseSchemas.__root__:
+                    schema_name_list.append(schema.name)
+
+        if self.schema_field and schema_name_list:
+            self.filters += (  # pylint: disable=no-member
+                f" AND hasAny({self.schema_field}, ['"
+                + "','".join(schema_name_list)
+                + "'])"
+            )
```
ulixius9, line 92: clickhouse has this special approch where it return the array of databases instead of single database, hence handled this seperately

ingestion/src/metadata/ingestion/source/database/query_parser_source.py
ulixius9: every usage and lineage source will define the `schema_field` and `database_field` available in the query log table

pmbrull (changes requested):
Let's add a couple of unit tests around the `prepare` functionality please 🙏 

One quick way to achieve it would be to separate the `prepare` in two calls:
- fill up the db and schema list
- then build the filter based on the values.

You could then easily test the filter build method, as the ometa call should be good already

pmbrull (approved):
Thanks for adding the tests 🙏

PR #9616, opened 2023-01-05
Fix Tag Missing Relation

mohityadav766:
### Describe your changes :

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
@@ -627,6 +627,9 @@ public interface CollectionDAO {
     @RegisterRowMapper(FromRelationshipMapper.class)
     List<EntityRelationshipRecord> findFrom(@Bind("toId") String toId);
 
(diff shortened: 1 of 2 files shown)


sureshms:
@mohityadav766 why are these relationships missing?

mohityadav766:
> @mohityadav766 why are these relationships missing? Is this just an issue with changing the tag category name in relationship table to `category`?

@sureshms 
1. earlier the relationship were not there for the tagCategory -> tag or tag -> tag!, we were just operating on fullyQualifiedName, so as for the user created tags we won't be having these relationship for the tags and classification now
2. For the system tags those are never initialised again since those during startup already exist in the system, so we never insert them again , so hence the relationship does not get created. (For system tag we could handle them by just deleting them an allowing the system to create them again, but for user created tag on lower version it won't work)

akash-jain-10 (approved):

sureshms:
@mohityadav766 in that case, it would be better to go without the relationship. In fact I have a pending task to remove these relationships when we can use name prefix based relationship for the parent.

PR #9667, opened 2023-01-10
e2e: UI : User journey for permission end-2-end test

### LINKED ISSUE #7440: UI : User journey for permission end-2-end test
As part of the 0.12.0 release, we have added a new structure for roles and policies. with the new format, we have different permissions for the various entities. so based on permission users can be allowed or denied to view, and update the data.

In 0.12.1 we want to solidify this new system and have E2E tests with different user journeys. This ticket is to keep track of the user journey that we want to test E2E on UI.

Example: 
By default, the user will have the `data consumer` role with the `data consumer policy`. `data consumer policy` will have the rule `DataConsumerPolicy-EditRule`.

image

This rule states that users with this rule can view all the entities but can only edit the `description` and `tags`, so here user journey should be like this.

User logged in into the OM -> goes to the entity (table) -> should be able to view all data -> should be able to edit description and tags -> should not be able to perform any other operations like adding tests, updating tests, updating the lineage, etc.

cc : @harshach @sureshms @chirag-madlani

Sachin-chaurasiya:
Closes #7440
### Describe your changes :

I worked on issue #7440

#
### Type of change :

- [x] Improvement

#

#
### Checklist:

#
### Reviewers
@open-metadata/ui


openmetadata-ui/src/main/resources/ui/cypress/e2e/Features/DataConsumerRole.spec.js
```diff
@@ -110,7 +110,7 @@ const ID = {
   },
 };
 // Todo:- Need to fix flaky test
-describe.skip('DataConsumer Edit policy should work properly', () => {
+describe('DataConsumer Edit policy should work properly', () => {
```
chirag-madlani, line 113: Can we remove todo comment above?

chirag-madlani (approved):

PR #9678, opened 2023-01-11
Fix: added test case for atlas 

NiharDoshi99:
### Describe your changes :

I worked on adding test case for atlas and version bump for neo4j for amundsen

#
### Type of change :

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation


ingestion/tests/unit/topology/metadata/test_atlas.py
```diff
@@ -0,0 +1,612 @@
...
+    @patch.object(AtlasSource, "create_tag", mock_create_tag)
+    def test_description(self):
+        """
+        Testing description updated for database, databaseSchema, table
+        """
+        _ = list(self.atlas_source.next_record())
+        updated_database = self.metadata.get_by_name(
+            entity=Database, fqn="hive.Reporting"
+        )
+        assert updated_database.description == EXPTECTED_DATABASE.description
```
ulixius9, line 593: if we are just checking the description then it's okay just to assert it with string instead of creating the entire object and then just matching the description

ingestion/tests/unit/topology/metadata/test_atlas.py
```diff
@@ -0,0 +1,612 @@
...
+        )
+        assert (
+            updated_database_schema.description == EXPTECTED_DATABASE_SCHEMA.description
+        )
+
+        updated_table = self.metadata.get_by_name(
+            entity=Table, fqn="hive.Reporting.Reporting.sales_fact_daily_mv"
+        )
+
+        assert updated_table.description == EXPTECTED_TABLE.description
```
ulixius9, line 606: also apart from description we can also verify tags that we are updating for table

ingestion/tests/unit/topology/metadata/test_atlas.py
NiharDoshi99: OK 👍

ingestion/tests/unit/topology/metadata/test_atlas.py
```diff
@@ -0,0 +1,506 @@
...
+    return mock_data
+
+
+def mock_list_entities(self, entity_type):  # pylint: disable=unused-argument
+    return LIST_ENTITIES
+
+
+EXPECTED_DATABASE_DESCRIPTION = "THIS IS TEST_DESCRIPTION FOR DATABASE"
+
+EXPTECTED_DATABASE_SCHEMA_DESCRIPTION = "THIS IS TEST_DESCRIPTION FOR DATABASE"
```
OnkarVO7, line 115:
```suggestion
EXPTECTED_DATABASE_SCHEMA_DESCRIPTION = "THIS IS TEST_DESCRIPTION FOR DATABASE SCHEMA"
```

ingestion/tests/unit/topology/metadata/test_atlas.py
```diff
@@ -0,0 +1,506 @@
...
+    ],
+    tableConstraints=None,
+    tablePartition=None,
+    owner=None,
+    databaseSchema=EntityReference(
+        id="4cf6ee7e-9d24-4153-9318-82aa1167259b",
+        type="databaseSchema",
+        name="Reporting",
+        fullyQualifiedName="hive.Reporting.Reporting",
+        description="THIS IS TEST_DESCRIPTION FOR DATABASE",
```
OnkarVO7, line 224:
```suggestion
        description="THIS IS TEST_DESCRIPTION FOR DATABASE SCHEMA",
```

OnkarVO7 (changes requested):
Small change rest looks good

ingestion/tests/unit/topology/metadata/test_atlas.py
NiharDoshi99: So It ingest same for database and schema . So value for database and schema are same.

OnkarVO7 (approved):

Issue #7625, opened 2022-09-21
openmetadata ingestion and orm profiler needs to be fixed to extract all metadata from z/OS DB2


upenbendre:
While making ingestion work for a z/OS DB2 database, I had to make changes to the reflection.py and base.py of the ibm_db_sa, otherwise it would skip over most of the tables and views in the database. It does not look like the existing ibm_db_sa is tested thoroughly for z/OS DB2. Here are the changes I made:
reflection.py in ibm_db_sa
1.) In the declaration sys_columns = Table(“SYSCOLUMNS”, ischema, replace ‘Column(“TYPENAME”, CoerceUnicode, key=“typename”),’ with ‘Column(“COLTYPE”, CoerceUnicode, key=“typename”),’. If you refer to the query generated for get_columns, you will see why this is needed.
2.) In the def get_columns , within the for loop for r in connection.execute(query) in the first statement, add a strip().
coltype = r[1].upper().strip(). The column type from z/OS is retrieved with spaces that need to be removed.
3.) In def get_unique_constraints, replace the sysconst.c.type == ‘U’ with sysconst.c.type != ‘’ OR instead of just the check on ‘U’, check for ‘U’ or ‘P’. These type codes are tricky. I suspect different mainframe types may have different values, so to me, it doesn’t make sense to put a filter on them at this stage in the code. Having said that, the current metadata/ingestion code does not seem to call the get_unique_constraints at all. So that needs to be looked at as well.
4.) Similarly the get_indexes function also does not get invoked by metadata. If that is fixed, then I can test how the for loop ‘for r in connection.execute(query):’ inside the get_indexes function works. If this is not a metadata issue, then it needs more research.
5.) In a different post, I have also mentioned that get_incoming_foreign_keys is also invoked by metadata - and that is for z/OS DB2 as well as Linux DB2. If this is not a metadata issue, then it needs more research.
Also in the base.py of ibm_db_sa under the ischema_names dictionary, in addition to ‘TIMESTAMP’: TIMESTAMP, I also added ‘TIMESTMP’: TIMESTAMP,. (TIMESTMP is how the timestamp datatype is spelt in some mainframe installations, where the names have to be 8 characters.

**Version:**
CentOS 7
Python 3.10
OM 0.11.5
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

[...]

Issue #6967, opened 2022-08-26
DB2 Porfiler errors out


TeddyCr:
```
[2022-08-25 23:35:49] WARNING  {Profiler:core:253} - Error trying to compute column profile for `colname` - (ibm_db_dbi.ProgrammingError) ibm_db_dbi::ProgrammingError: SQLNumResultCols failed: [IBM][CLI Driver][DB2/AIX64] SQL0418N  The statement was not processed because the statement contains an invalid use of one of the following: an untyped parameter marker, the DEFAULT keyword, or a null value.  SQLSTATE=42610 SQLCODE=-418
[SQL: /* {“app”: “OpenMetadata”, “version”: “0.11.5.0"} */
SELECT AVG(DOUBLE((col))) AS mean, count(col) AS “valuesCount”, count(DISTINCT col) AS “distinctCount”, min(col) AS “min”, NULL AS anon_1, max(col) AS “max”, NULL AS anon__1, sum(CASE WHEN (col IS NULL) THEN ? ELSE ? END) AS “nullCount”, STDDEV_POP(col) AS stddev, sum(col) AS “sum”
FROM “schema“.table FETCH FIRST 1 ROWS ONLY]
[parameters: (1, 0)]
(Background on this error at: https://sqlalche.me/e/14/f405)
```

TeddyCr:
Currently working on a dev environment as M1 does not support DB2 drivers

pmbrull:
@TeddyCr one possibility for us to try out is if we can install a later version of the package that supports M1 arch

TeddyCr:
@pmbrull we are installing the latest releases of the packages. Unfortunately they do not support M1.

TeddyCr:
Should be fixed at the same time than #6967

Issue #6953, opened 2022-08-26
Fix DB2 syntax for the profiler


TeddyCr:
DB2 profiler is throwing error when computing specific metrics

```
[2022-08-25 23:35:55] WARNING  {Profiler:core:253} - Error trying to compute column profile for col - (ibm_db_dbi.ProgrammingError) ibm_db_dbi::ProgrammingError: SQLNumResultCols failed: [IBM][CLI Driver][DB2/AIX64] SQL0440N  No authorized routine named “LEN” of type “FUNCTION” having compatible arguments was found.  SQLSTATE=42884 SQLCODE=-440
[SQL: /* {“app”: “OpenMetadata”, “version”: “0.11.5.0"} */
SELECT AVG(DOUBLE((LEN(col)))) AS mean, count(col) AS “valuesCount”, count(DISTINCT col) AS “distinctCount”, NULL AS anon_1, min(LEN(col)) AS “minLength”, NULL AS anon__1, max(LEN(col)) AS “maxLength”, sum(CASE WHEN (col IS NULL) THEN ? ELSE ? END) AS “nullCount”, NULL AS anon__2, NULL AS anon__3
FROM “schema“.table FETCH FIRST 1 ROWS ONLY]
[parameters: (1, 0)]
(Background on this error at: https://sqlalche.me/e/14/f405)
```

upenbendre:
@TeddyCr here is our Slack chat regarding this issue for your reference:

[...]

PR #9679, opened 2023-01-11
Fixed DB2 Profiler

OnkarVO7:
### Describe your changes :

Fix https://github.com/open-metadata/OpenMetadata/issues/6967
Fix https://github.com/open-metadata/OpenMetadata/issues/7625
Fix https://github.com/open-metadata/OpenMetadata/issues/6953

#
### Type of change :

- [x] Bug fix

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/orm_profiler/metrics/static/column_count.py
+++ b/ingestion/src/metadata/orm_profiler/metrics/static/column_count.py
@@ -15,9 +15,32 @@ Table Column Count Metric definition
 # pylint: disable=duplicate-code
 
 from sqlalchemy import inspect, literal
+from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.orm import DeclarativeMeta
+from sqlalchemy.sql.functions import FunctionElement
 
-from metadata.orm_profiler.metrics.core import StaticMetric, _label
+from metadata.orm_profiler.metrics.core import CACHE, StaticMetric, _label
+from metadata.orm_profiler.orm.registry import Dialects
+
+
+class ColunCountFn(FunctionElement):
+    name = __qualname__
+    inherit_cache = CACHE
+
+
+@compiles(ColunCountFn)
+def _(element, compiler, **kw):
+    return compiler.process(element.clauses, **kw)
+
+
+@compiles(ColunCountFn, Dialects.IbmDbSa)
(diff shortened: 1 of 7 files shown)


ingestion/src/metadata/orm_profiler/metrics/static/column_count.py
```diff
@@ -15,9 +15,32 @@
...
+
+
+@compiles(ColunCountFn, Dialects.IbmDbSa)
+@compiles(ColunCountFn, Dialects.Db2)
+def _(element, compiler, **kw):
+    """Returns column count for db2 database and handles casting variables.
+    If casting is not provided for variables, db2 throws error.
+    """
+    proc = compiler.process(element.clauses, **kw)
+    return f"CAST({proc} AS BIGINT)"
```
TeddyCr, line 43: Is this to handle the type not matching when it is None?

OnkarVO7: @TeddyCr it's not for `None`.

The query in db2 does not fail for `NULL` datatypes as we were suspecting earlier.
The query fails when we have variables in the query i.e the `?` thing. If this is not type casted into a DB2 datatype the query fails for DB2
The logic written here and other places is to handle this use case

TeddyCr (approved):
In `ingestion/src/metadata/orm_profiler/orm/functions/sum.py` is it expected for us to CAST twice?

ingestion/src/metadata/orm_profiler/orm/functions/sum.py
```diff
@@ -58,3 +58,11 @@ def _(element, compiler, **kw):
...
     proc = compiler.process(element.clauses, **kw)
     return f"SUM({proc})"
+
+
+@compiles(SumFn, Dialects.IbmDbSa)
+@compiles(SumFn, Dialects.Db2)
+def _(element, compiler, **kw):
+    """Handle the case for DB2 where it requires to type cast the variables"""
+    proc = compiler.process(element.clauses, **kw).replace("?", "CAST(? AS INT)")
+    return f"SUM(CAST({proc} AS BIGINT))"
```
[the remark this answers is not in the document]
OnkarVO7: It is expected since the `?` requires it's own casting

PR #9687, opened 2023-01-11
Issue 9542 - Replace the default name for MySQL db

### LINKED ISSUE #9542: Replace the "default" name for a MySql database with a name specified by the user
A lot of our database instances use the same schema name: e.g. innodb.
OpenMetadata automatically assigns the database name as _default_ for a Mysql service. This causes our tables to show up under _default.innodb_ in Openmetadata even though they come from different databases. This makes it look like all the tables have been ingested from the same database.  For example in the screenshot below, table_A, table_B and table_C belong to DB1, while table_D and table_E belong to DB2, but this is not obvious in OpenMetadata:
image

We would like to see the database name in Openmetadata like in the screenshot below:
image

Since our MySQL databases are identified by their subdomain name: e.g. if DB1.domain.com and DB2.domain.com are the domain names we use to access the databases,  we use DB1 and DB2 to identify the databases. Hence it would be useful for us to: 
- either have the database name assigned automatically from the subdomain name or
- have the option to assign it manually when configuring a MySQL service

codingwithabhi:
### Describe your changes :

I worked on the adding database field for custom database name.

Fix #9542

#
### Type of change :

- [x] Improvement

#
### Checklist:

#
### Reviewers
@open-metadata/ingestion

### DIFF
--- a/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
+++ b/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
@@ -50,7 +50,7 @@
     },
     "database": {
       "title": "Database",
-      "description": "Database of the data source. This is optional parameter, if you would like to restrict the metadata reading to a single database. When left blank, OpenMetadata Ingestion attempts to scan all the databases.",
+      "description": "Database name of the data source. This is an user defined parameter, When left blank, OpenMetadata Ingestion will use default as the database name.",
       "type": "string"
     },
     "connectionOptions": {
--- a/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
+++ b/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
@@ -50,7 +50,7 @@
     },
     "databaseName": {
       "title": "Database Name",
-      "description": "Database name of the data source. This is an user defined parameter, When left blank, OpenMetadata Ingestion will use default as the database name.",
+      "description": "Optional name to give to the database in OpenMetadata. If left blank, we will use default as the database name.",
       "type": "string"
     },
     "connectionOptions": {
--- a/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
+++ b/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
@@ -32,6 +32,7 @@
       "$ref": "#/definitions/athenaScheme",
       "default": "awsathena+rest"
     },
+    
     "awsConfig": {
       "title": "AWS Credentials Configuration",
       "$ref": "../../../../security/credentials/awsCredentials.json"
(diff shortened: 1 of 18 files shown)


openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
```diff
@@ -47,6 +48,11 @@
       "description": "Athena workgroup.",
       "type": "string"
     },
+    "database": {
+      "title": "Database",
+      "description": "Database of the data source. This is optional parameter, if you would like to restrict the metadata reading to a single database. When left blank, OpenMetadata Ingestion attempts to scan all the databases.",
```
ulixius9, line 53: we might need to change this description, we will have to make it clear that this is just for display and that we will not actually ingest the database.

ulixius9 (commented):
Thanks for the pr 🙇‍♂️ a minor comment!

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/trinoConnection.json
```diff
@@ -53,6 +53,11 @@
       "description": "Catalog of the data source.",
       "type": "string"
     },
+    "database": {
```
pmbrull, line 56: Thanks for the help, just a couple of concerns here:
- For Presto and Trino we are using the Catalog field as the database in OM. I would not add the `database` field here as it can be confusing.
- For Glue, we use the catalog ID. Thus, same as above, I would not add the database field.
- Databricks same same, uses the catalog. So Let's not add the database field in there.

The idea here is to allow different names if, and only if, the user needs it. There is no concept of database as such in these connectors. I would do a couple of changes in the logic of this change:
- I'd name the field `databaseName`: We are not ingesting a different database. We'll be ingesting the schemas present in the source as databaseSchema and just change the database name. This database is just a filler to comply with OM's entity hierarchy.
- I'd have this field with a default value "default". This is the one we will use in the default implementation of [get_database_names](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/database/common_db_source.py#L131).
- Let's remove the logic change in the connection builder. This change should be only aesthetic and have no impact on how the connection gets created. Even adding the check for "supportsDatabase", having the name as "database" in the JSON Schema can be confusing.

Thanks

pmbrull (changes requested):

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
```diff
@@ -47,6 +48,11 @@
       "description": "Athena workgroup.",
       "type": "string"
     },
+    "databaseName": {
+      "title": "Database Name",
+      "description": "Database name of the data source. This is an user defined parameter, When left blank, OpenMetadata Ingestion will use default as the database name.",
```
pmbrull, line 53:
```suggestion
      "description": "Optional name to give to the database in OpenMetadata. If left blank, we will use default as the database name.",
```

pmbrull (commented):
Thanks for applying the changes. Left a suggestion on the field description, as it will appear on the UI. If you could please propagate it to the rest of the schemas we should be good to go.

Thanks!

pmbrull (dismissed):
Thanks for the changes

pmbrull:
@ulixius9 @codingwithabhi not sure if the cypress failures are expected 🤔

pmbrull (approved):

Issue #9754, opened 2023-01-17
Can't Add  Atlas Service


bigdata-spec:
Hi，
    I want to add **Atlas Service** 
    We have atlas Server and I can login **http://10.1x.x.x:21021 by admin/admin**
    when I test connect  
![image](https://user-images.githubusercontent.com/23710717/212795186-0e131187-a792-4437-80de-5540a136d5d0.png)
    Can you help me? Thank you very much.

harshach:
@bigdata-spec did you added "http://" infront of hostandport?

bigdata-spec:
> @bigdata-spec did you added "http://" infront of hostandport?

@harshach Thank you very much, **Apache Atlas work well**
![image](https://user-images.githubusercontent.com/23710717/212795599-4f6b9da2-cc04-4db6-ad9d-e8296738054a.png)

when I don't added "http://"
![image](https://user-images.githubusercontent.com/23710717/212795823-e0d536bf-1e35-4b90-aa4d-6f3ba4882fb0.png)

when I  added "http://"
![image](https://user-images.githubusercontent.com/23710717/212795912-2a278ab7-d05f-49be-ba51-7de557e491a4.png)

harshach:
@NiharDoshi99 please add clear description to the configs with an example.
cc @pmbrull

bigdata-spec:
@harshach @NiharDoshi99 Thank you very much, But I still can't run **test atlas  connection**,
I follower steps:
1.pip3 install --upgrade "openmetadata-ingestion[docker]"
2.metadata docker --start

![image](https://user-images.githubusercontent.com/23710717/215250043-9be40aa3-96bc-43ea-ac32-5a9f31ac4123.png)

harshach:
@bigdata-spec are you using 0.13.2 beta release?

bigdata-spec:
> @bigdata-spec are you using 0.13.2 beta release?

@harshach Hi,I just try **quick-start**  [](https://docs.open-metadata.org/quick-start/local-deployment)
![image](https://user-images.githubusercontent.com/23710717/215299088-5ec50484-61cf-4af9-ba70-6067fb8d6229.png)

PR #9791, opened 2023-01-18
Fix: atlas test connection

NiharDoshi99:
### Describe your changes :

I worked on fixing test connection for atlas 
Fix : #9754 

#
### Type of change :

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation


ingestion/src/metadata/ingestion/source/metadata/atlas/client.py
```diff
@@ -42,7 +42,9 @@ def __init__(self, config: AtlasConnection, raw_data: bool = False):
         self._use_raw_data = raw_data
 
     def list_entities(self, entity_type="Table") -> List[str]:
-        response = self.client.get(f"/atlas/entities?type={entity_type}")
+        response = self.client.get(
+            f"/atlas/entities?type={self.config.entity_type if self.config.entity_type else entity_type}"
```
OnkarVO7, line 46: @NiharDoshi99 will this condition be always true since the `entity_type` is a required field?

NiharDoshi99: Oh yes , let me remove this if condition 👍

OnkarVO7 (approved):

PR #9796, opened 2023-01-18
Fix #8884 & #9460 - View lineage happens in the lineage workflow

### LINKED ISSUE #8884: Move views/DBT lineage into lineage workflow
**Issue**
Currently, lineage is divided into different workflows. On the one hand, the lineage of views and DBT is done during the metadata ingestion workflow. On the other, we are extracting the rest of the lineage in the lineage workflow. 

This could cause some issues:
- Workflows are doing more than expected.
- Confusing for the user after first ingestion since they partially have lineage in views.
- It is hard to track lineage-related issues if it is part of another workflow.

**Task**
Simplify metadata ingestion workflow by moving the extraction of the lineage of views and DBT into the lineage workflow.

### LINKED ISSUE #9460: View lineage is not working properly.
**Affected module**
 Ingestion Framework

**Describe the bug**
If we ingest Mssql(Databricks etc..) without specifying the database name, the view lineage will not get ingested.
`yield_view_lineage` function is responsible for view lineage. 
Problem is that after ingesting all databases and table `yield_view_lineage` will run, now in Mssql.py we set inspector for different databases.
so now in `yield_view_lineage`, we will get the last `set_inspector`.

**To Reproduce**

Ingest Mssql service with all databases.

**Expected behavior**
View lineage is working when we give only one database name, but it should also work when we ingest with all databases

**Version:**

 - OpenMetadata version: [0.13.1]

**Additional context**

pmbrull:
### Describe your changes :

Fix #8884
Fix #9460 

Moved the View Lineage from the metadata ingestion to the lineage workflow. Reasons:
1. We centralize lineage processing in a single place. Makes it easier to debug. The lineage was a `post_processing` at the service level anyway as it required ALL the metadata to be available.
2. There was a bug where we picked up the view definition (twice) and using the latest inspector instead of the one for each database. This simplifies the logic as we just get the tables from the API.

Further work: It would be interesting to add a query parameter for `tableType` when listing tables in the API to only fetch views in these scenarios.

#8884 also listed moving dbt lineage to the lineage workflow. The reason not to is to maintain all the catalog and manifest information when processing the lineage. Whenever we jump to grabbing lineage in cases where `A -> TMP -> B` for non-materialized models we might need the catalog nodes. If, after revisiting the dbt lineage for non-materialized models, we see that we have all the required information just in the API, then we can move it to the lineage workflow, but I would not jump to that before properly analyzing the situation.

Thanks

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/common_db_source.py
+++ b/ingestion/src/metadata/ingestion/source/database/common_db_source.py
@@ -389,58 +389,6 @@ class CommonDbSourceService(
             logger.warning(f"Unexpected exception to yield table [{table_name}]: {exc}")
             self.status.failures.append(f"{self.config.serviceName}.{table_name}")
 
-    def yield_view_lineage(self) -> Optional[Iterable[AddLineageRequest]]:
-        logger.info("Processing Lineage for Views")
-        for view in self.context.table_views:
-            table_name = view.get("table_name")
-            table_type = view.get("table_type")
-            schema_name = view.get("schema_name")
-            db_name = view.get("db_name")
-            table_fqn = fqn.build(
-                self.metadata,
-                entity_type=Table,
-                service_name=self.context.database_service.name.__root__,
-                database_name=db_name,
-                schema_name=schema_name,
-                table_name=table_name,
-            )
-            table_entity = self.metadata.get_by_name(
-                entity=Table,
-                fqn=table_fqn,
-            )
-            view_definition = self.get_view_definition(
-                table_type=table_type,
-                table_name=table_name,
-                schema_name=schema_name,
-                inspector=self.inspector,
-            )
-
-            try:
-                lineage_parser = LineageParser(view_definition)
-                if lineage_parser.source_tables and lineage_parser.target_tables:
-                    yield from get_lineage_by_query(
-                        self.metadata,
-                        query=view_definition,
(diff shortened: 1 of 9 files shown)


ingestion/src/metadata/ingestion/source/database/common_db_source.py
```diff
@@ -389,58 +383,6 @@ def yield_table(
             logger.warning(f"Unexpected exception to yield table [{table_name}]: {exc}")
             self.status.failures.append(f"{self.config.serviceName}.{table_name}")
 
-    def yield_view_lineage(self) -> Optional[Iterable[AddLineageRequest]]:
-        logger.info("Processing Lineage for Views")
-        for view in self.context.table_views:
```
nahuelverdugo, line 394: Should we remove the `table_views` from the context?

nahuelverdugo (approved):

ingestion/src/metadata/ingestion/source/database/common_db_source.py
pmbrull: good catch. Thank you, removed it

nahuelverdugo (dismissed):

nahuelverdugo (approved):

PR #9800, opened 2023-01-18
Fix: Replace `sqllineage` with `openmetadata-sqllineage`

nahuelverdugo:
### Describe your changes :
I have updated the `sqllineage` used until now with `openmetadata-sqllineage`, which adds an option to extract lineage by using `sqlfluff` as the parser and considers the dialect of the query. Since it is a recent change, we will use the `sqlfluff` parser option as a fallback for the usual lineage extraction. 

In addition, the `LineageParser` now supports the dialect parameter to use `sqllineage` with `sqlfluff`.

### Type of change :
- [x] Improvement

### Checklist:

### Reviewers

### DIFF
--- a/ingestion/src/metadata/ingestion/lineage/models.py
+++ b/ingestion/src/metadata/ingestion/lineage/models.py
@@ -14,6 +14,9 @@ Models related to lineage parsing
 from enum import Enum
 from typing import Dict
 
+from metadata.generated.schema.entity.services.connections.database.athenaConnection import (
+    AthenaType,
+)
 from metadata.generated.schema.entity.services.connections.database.bigQueryConnection import (
     BigqueryType,
 )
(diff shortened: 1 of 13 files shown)


ingestion/src/metadata/ingestion/lineage/models.py
```diff
@@ -0,0 +1,91 @@
...
+    SNOWFLAKE = "snowflake"
+    SOQL = "soql"
+    SPARKSQL = "sparksql"
+    SQLITE = "sqlite"
+    TERADATA = "teradata"
+    TSQL = "tsql"
+
+
+MAP_CONNECTION_TYPE_DIALECT: Dict[str, Dialect] = {
+    str(MySQLType.Mysql.value): Dialect.ATHENA,
```
pmbrull, line 79: could we add a comment on why this mapping?

nahuelverdugo: Moved the logic into `ConnectionTypeDialectMapper` class.

pmbrull (dismissed):

akash-jain-10:
Hello @nahuelverdugo . How are we publishing the `openmetadata-sqllineage` python package ? Is that something we need to do for every release ? Can you please share some details for the same here. Thanks.

nahuelverdugo:
> Hello @nahuelverdugo . How are we publishing the `openmetadata-sqllineage` python package ? Is that something we need to do for every release ? Can you please share some details for the same here. Thanks.

We have created a repo for that https://github.com/open-metadata/openmetadata-sqllineage, and there is an action that releases when a release is created. It is a fork of the `sqllineage` library that we did to start testing a change done in the used parser for the queries. We don't need to do it on each release, and we hope the repo is temporary until we merge https://github.com/reata/sqllineage/pull/326.

akash-jain-10 (approved):

OnkarVO7 (approved):

PR #9813, opened 2023-01-19
Issue #7291 - Implements Table Rows Inserted to be Between test

### LINKED ISSUE #7291: Add Table Row Inserted to Be Between Data Quality Test
We should add a new test allowing users to tests the number of new rows added was between a certain range. We'll need 5 parameters for this test:
- min
- max
- timestampField
- timeRangeType (Optional [hourly, daily] should default to daily)
- timeRange (Optional should default to 1)

TeddyCr:
### Describe your changes :
Fixes #7291 by implementing support for table rows inserted to be between test for database source. We also refactor part of the code in this PR:
- Improved partition logic
- Split datalake and database validation unit test suite 
- Leverage pytest parametrize tests and conftest file

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/orm_profiler/profiler/runner.py
```diff
@@ -55,25 +59,76 @@ def _build_query(self, *entities, **kwargs) -> Query:
         return self._session.query(*entities, **kwargs)
 
     def _select_from_sample(self, *entities, **kwargs):
-        return self._build_query(*entities, **kwargs).select_from(self._sample)
+        """Run select statement against sample data"""
+        if kwargs.get("query_filter_"):
+            query_filter = kwargs.pop("query_filter_")
```
pmbrull, line 64: sorry random question, why do we need to `pop` the keys?

ingestion/src/metadata/orm_profiler/profiler/runner.py
```diff
@@ -55,25 +59,76 @@ def _build_query(self, *entities, **kwargs) -> Query:
...
+
+        if filter_ is not None:
+            return query.filter(filter_)
+
+        return query
 
     @partition_filter_handler()
     def select_first_from_table(self, *entities, **kwargs):
+        """Select first row from the table"""
+        if kwargs.get("query_filter_"):
```
pmbrull, line 98: I think these lines 

```python
if kwargs.get("query_filter_"):
    query_filter = kwargs.pop("query_filter_")
    filter_ = build_query_filter(**query_filter)
else:
    filter_ = None
```

are repeated a few times. Maybe we could have a function wrapping this

pmbrull (approved):
Thanks for the effort in getting all the test work in. Left a couple of questions, but feel free to bring this in. Thanks

ingestion/src/metadata/orm_profiler/profiler/runner.py
TeddyCr: We use pop as we don't want to pass that keyword to `self._build_query` 🙂

Issue #9725, opened 2023-01-13
Openmetadata Tour Page Issue table is not present for dummy data


Aashish1221:
**Affected module**
OpenMetadata home 
**Describe the bug**
hi team ..i was exploring tour option in OMD .I have noticed an issue when we enter dim_a in the search when the tour guide us & then proceed to complete steps ..table instance not found error reflects in Background for OMD ..If someone face this issue in Dummy data can it be handled so that no error releated to table come

**To Reproduce**
pfa screen cast 

**Expected behavior**
there should not be any error related to table not found

https://user-images.githubusercontent.com/42956103/212315361-6aa5b0f6-029a-4d93-9b90-dfba036fdba5.mp4

nd should appear on the UI in background 
**Version:**
 - OS: Linux
 - OpenMetadata version: 0.13.1

**Additional context**

PR #9826, opened 2023-01-19
Fixed: Openmetadata Tour Page Issue table is not present for dummy data #9725

ShaileshParmar11:
### Describe your changes :

- Fixed Tour error issue #9725 
- Fixed incorrect page API count from UI

#
### Type of change :

- [x] Bug fix

#

https://user-images.githubusercontent.com/71748675/213530816-58e17914-e9d8-4fc8-a1bd-6fea0ddbafb6.mov

#
### Checklist:

#

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/pages/teams/AddUsersModalV1.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/pages/teams/AddUsersModalV1.tsx
@@ -88,7 +88,7 @@ const AddUsersModalV1 = ({
       });
   };
 
-  const searchUsers = (text: string, page: number) => {
+  const searchUsers = (text: string, page = 1) => {
     searchData(text, page, PAGE_SIZE_MEDIUM, '', '', '', SearchIndex.USER)
       .then((res) => {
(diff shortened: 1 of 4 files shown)


openmetadata-ui/src/main/resources/ui/src/pages/teams/AddUsersModalV1.tsx
```diff
@@ -88,7 +88,7 @@ const AddUsersModalV1 = ({
       });
   };
 
-  const searchUsers = (text: string, page: number) => {
+  const searchUsers = (text: string, page = 1) => {
```
Sachin-chaurasiya, line 91: @ShaileshParmar11 , Why we are making this change?

ShaileshParmar11: As it is not picking reset value, so I have updated the logic, if we want to reset the size we will not send any number but if we want to update API call based on page number we will pass page value!

Sachin-chaurasiya (approved):

PR #9853, opened 2023-01-21
fix content header types

geoHeil:
fix docs


ulixius9 (approved):
Thanks for the fix @geoHeil 🙇

PR #9870, opened 2023-01-23
Deduplicate stl_querytext rows, fixes #8867

### LINKED ISSUE #8867: Result size exceeds LISTAGG limit
**Affected module**
Error 'Result size exceeds LISTAGG limit' in lineage ingestion logs when connects to Redshift datasets.

**Describe the bug**
OpenMetadata uses the LISTAGG to aggregate the full query text as it is divided into chunks of 200 characters in pg_catalog stl_querytext. Therefore, in this case, it fails because the size of all the chunks sums more than 65535 bytes limits.

**To Reproduce**
image

**Expected behavior**
The lineage ingestion won't blocked by this error when process long queries.

**Version:**
 - OS: OSX
 - Python version: 3.9
 - OpenMetadata version: 0.12.2
 - OpenMetadata Ingestion package version: 0.12.2

**Additional context**

osocron:
### Describe your changes :

I worked on the #8867 issue because we have the same issue in our deployment

#
### Type of change :

- [x] Bug fix

#
### Checklist:

#
### Reviewers

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/redshift/queries.py
+++ b/ingestion/src/metadata/ingestion/source/database/redshift/queries.py
@@ -31,12 +31,16 @@ REDSHIFT_SQL_STATEMENT = textwrap.dedent(
           AND starttime < '{end_time}'
           LIMIT {result_limit}
   ),
+  deduped_querytext AS ( -- In some instances, rows are duplicated, causing LISTAGG to fail due to exceeding the maximum str length of 65535.
+    SELECT DISTINCT *
+    FROM pg_catalog.stl_querytext
+  ),
   full_queries AS (
     SELECT
           query,
           LISTAGG(CASE WHEN LEN(RTRIM(text)) = 0 THEN text ELSE RTRIM(text) END, '')
             WITHIN GROUP (ORDER BY sequence) AS query_text
-      FROM pg_catalog.stl_querytext
+      FROM deduped_querytext
       WHERE sequence < 327	-- each chunk contains up to 200, RS has a maximum str length of 65535.
     GROUP BY query
   ),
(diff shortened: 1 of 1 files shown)


ayush-shah:
we will have a look and let you know if there is something that needs to be commited in this PR, Thanks @osocron 🙌 ❤️

ingestion/src/metadata/ingestion/source/database/redshift/queries.py
```diff
@@ -31,12 +31,16 @@
...
+    SELECT DISTINCT *
+    FROM pg_catalog.stl_querytext
+  ),
   full_queries AS (
     SELECT
           query,
           LISTAGG(CASE WHEN LEN(RTRIM(text)) = 0 THEN text ELSE RTRIM(text) END, '')
             WITHIN GROUP (ORDER BY sequence) AS query_text
-      FROM pg_catalog.stl_querytext
+      FROM deduped_querytext
```
ulixius9, line 43: @pmbrull need your opinion on this, I was thinking that instead of doing a `SELECT DISTINCT *
    FROM pg_catalog.stl_querytext` here IMO we should directly use `queries` here which filtered result-set  
```suggestion
      FROM queries
```

osocron: The `queries` CTE uses a different table to the `full_queries` CTE. I did a join to reduce the size of the deduplicated CTE so that the filters get applied to both queries. Let me know what you think! 🙌

pmbrull: you're right, my bad.

LGTM then, thanks for taking the time for this

pmbrull (approved):
Thanks @osocron for the contribution!

@nahuelverdugo, if you haven't started checking redshift for 0.13.2, let's see if we can test it with these changes and bring them in as well.

Thanks,

pmbrull:
@osocron you might need to do a small tweak on the file though to pass the pylint. From the logs:

```
************* Module metadata.ingestion.source.database.redshift.queries
ingestion/src/metadata/ingestion/source/database/redshift/queries.py:34:0: C0301: Line too long (141/120) (line-too-long)
```

ulixius9 (approved):

PR #9906, opened 2023-01-25
Prepare Vertica Lineage and Usage

pmbrull:
### Describe your changes :

Add Vertica lineage and usage workflows.

Also, Vertica requires a special handling of query headers so that they are stored in the query history

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/ingestion/connections/builders.py
+++ b/ingestion/src/metadata/ingestion/connections/builders.py
@@ -12,6 +12,7 @@
 """
 Get and test connection utilities
 """
+from functools import partial
 from typing import Any, Callable, Dict, Optional
 from urllib.parse import quote_plus
 
@@ -24,7 +25,10 @@ from sqlalchemy.pool import QueuePool
 from metadata.generated.schema.entity.services.connections.connectionBasicType import (
     ConnectionArguments,
 )
-from metadata.ingestion.connections.headers import inject_query_header
+from metadata.ingestion.connections.headers import (
+    inject_query_header,
+    inject_query_header_by_conn,
+)
 from metadata.ingestion.connections.secrets import connection_with_options_secrets
 
 
@@ -67,7 +71,12 @@ def create_generic_db_connection(
     )
 
     if hasattr(connection, "supportsQueryComment"):
-        listen(engine, "before_cursor_execute", inject_query_header, retval=True)
+        listen(
+            engine,
+            "before_cursor_execute",
+            partial(inject_query_header_by_conn, connection),
+            retval=True,
+        )
 
     return engine
 
--- a/ingestion/src/metadata/ingestion/connections/builders.py
+++ b/ingestion/src/metadata/ingestion/connections/builders.py
@@ -25,10 +25,7 @@ from sqlalchemy.pool import QueuePool
 from metadata.generated.schema.entity.services.connections.connectionBasicType import (
     ConnectionArguments,
 )
-from metadata.ingestion.connections.headers import (
-    inject_query_header,
-    inject_query_header_by_conn,
-)
+from metadata.ingestion.connections.headers import inject_query_header_by_conn
 from metadata.ingestion.connections.secrets import connection_with_options_secrets
 
 
--- a/ingestion/src/metadata/ingestion/source/database/lineage_source.py
+++ b/ingestion/src/metadata/ingestion/source/database/lineage_source.py
@@ -16,6 +16,8 @@ import traceback
 from abc import ABC
 from typing import Iterable, Iterator, Optional
 
+from sqlalchemy.engine import Engine
+
 from metadata.generated.schema.api.lineage.addLineage import AddLineageRequest
 from metadata.generated.schema.type.tableQuery import TableQuery
 from metadata.ingestion.lineage.models import ConnectionTypeDialectMapper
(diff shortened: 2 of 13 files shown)


ingestion/src/metadata/ingestion/connections/builders.py
```diff
@@ -67,7 +71,12 @@ def create_generic_db_connection(
     )
 
     if hasattr(connection, "supportsQueryComment"):
-        listen(engine, "before_cursor_execute", inject_query_header, retval=True)
+        listen(
+            engine,
+            "before_cursor_execute",
+            partial(inject_query_header_by_conn, connection),
```
pmbrull, line 77: I needed to pass a specific function for Vertica to send the query header in a way that is indeed stored by the query history tables

ingestion/src/metadata/ingestion/source/database/lineage_source.py
```diff
@@ -64,31 +66,38 @@ def get_table_query(self) -> Optional[Iterator[TableQuery]]:
...
-                                f"Error processing query_dict {query_dict}: {exc}"
-                            )
+                engine = get_connection(self.service_connection)
+                yield from self.yield_table_query(engine)
+
             except Exception as exc:
                 logger.debug(traceback.format_exc())
                 logger.error(f"Source usage processing error: {exc}")
 
+    def yield_table_query(self, engine: Engine) -> Iterator[TableQuery]:
```
pmbrull, line 76: Separating specific methods in lineage and usage source. This way we can directly call them in Vertica as we need to iterate over the engines (databases) to get all the data from different dbs

pmbrull:
noticed an improvement, on it

ingestion/src/metadata/ingestion/source/database/lineage_source.py
pmbrull: Simplified this logic for Vertica. Although I am keeping the separation

ingestion/src/metadata/ingestion/source/database/vertica/query_parser.py
```diff
@@ -0,0 +1,69 @@
...
+from metadata.ingestion.api.source import InvalidSourceException
+from metadata.ingestion.source.connections import get_connection
+from metadata.ingestion.source.database.query_parser_source import QueryParserSource
+from metadata.ingestion.source.database.vertica.queries import VERTICA_LIST_DATABASES
+from metadata.utils.logger import ingestion_logger
+
+logger = ingestion_logger()
+
+
+class VerticaQueryParserSource(QueryParserSource, ABC):
```
TeddyCr, line 36: Do you know why we are making the class inherit from `ABC`?

ingestion/src/metadata/ingestion/source/database/vertica/query_parser.py
```diff
@@ -0,0 +1,69 @@
...
+        if not isinstance(connection, VerticaConnection):
+            raise InvalidSourceException(
+                f"Expected VerticaConnection, but got {connection}"
+            )
+        return cls(config, metadata_config)
+
+    def get_table_query(self) -> Iterable[TableQuery]:
+        database = self.config.serviceConnection.__root__.config.database
+        if database:
+            yield from super().get_table_query()
```
TeddyCr, line 61: I believe this will call `get_table_query` from `QueryParserSource` which does not implement the method

ingestion/src/metadata/ingestion/source/database/vertica/query_parser.py
```diff
@@ -0,0 +1,69 @@
...
+        else:
+            results = self.engine.execute(VERTICA_LIST_DATABASES)
+            for res in results:
+                row = list(res)
+                logger.info(f"Ingesting from database: {row[0]}")
+                self.config.serviceConnection.__root__.config.database = row[0]
+                self.engine = get_connection(self.service_connection)
+                yield from super().get_table_query()
```
TeddyCr, line 69: Same comment than above

ingestion/src/metadata/ingestion/source/database/vertica/query_parser.py
pmbrull: The `XYZQueryParserSource` does not implement all abstract methods from `QueryParserSource`. It is then used to implement the Usage and Lineage Sources which finish the rest of the implementation

ingestion/src/metadata/ingestion/source/database/vertica/query_parser.py
pmbrull: So here we actually have some nasty inheritance working that we might want to simplify in the future because tbh, now that you said it, it has me a bit lost.

Our classes at play are:
- `VerticaLineageSource(VerticaQueryParserSource, LineageSource)`
- `VerticaQueryParserSource(QueryParserSource, ABC)` -> implements `get_table_query` and calls `super().get_table_query()`
- `LineageSource(QueryParserSource, ABC)` -> implements `get_table_query`

So when we run `VerticaLineageSource` it runs the `VerticaQueryParserSource.get_table_query` that then runs `LineageSource.get_table_query`

This works because of the inheritance order of the classes set in `VerticaLineageSource(VerticaQueryParserSource, LineageSource)`. Changing the order does not follow the right calls.

I am sure we can find a less dark way to control the flow here. IMO we can open another issue to track it separated, as this is implemented in other lineage/usage sources as well. Thanks for pointing it out

TeddyCr: I see. That makes sense. it follows the mro of `VerticaLineageSource` so it will call the `get_table_query` of `LineageSource `.

pmbrull: we're just reusing one implementation of the query parsing for lineage and usage sources. We can come up with a better way. opened https://github.com/open-metadata/OpenMetadata/issues/9936 to track this.

IMO for now we should be able to merge this one and then update in bulk

TeddyCr (approved):
LGTM. Another thing that we need to think about regarding the implementation is that it works right now since we have the specific mro. If a class only subclasses the class calling `super()....` we could have hard to debug issues.

PR #9925, opened 2023-01-25
Fix #9923: Fix usage for tables with same name

### LINKED ISSUE #9923: Incorrect queries displayed when two tables are with same name
**Affected module**
Ingestion Framework

**Describe the bug**
Incorrect queries displayed when two tables are with same name but in different schema

image

image

`public.role_new` table should not contain create query for `sales.role_new`

ulixius9:
### Describe your changes :

Fix #9923: Fix usage for tables with same name

#
### Type of change :

- [x] Bug fix

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/ingestion/bulksink/metadata_usage.py
+++ b/ingestion/src/metadata/ingestion/bulksink/metadata_usage.py
@@ -102,7 +102,7 @@ class MetadataUsageBulkSink(BulkSink):
             self.table_usage_map[table_entity.id.__root__] = {
                 "table_entity": table_entity,
                 "usage_count": table_usage.count,
-                "sql_queries": list(table_usage.sqlQueries),
+                "sql_queries": list(table_usage.sqlQueries or []),
                 "usage_date": table_usage.date,
                 "database": table_usage.databaseName,
                 "database_schema": table_usage.databaseSchema,
--- a/ingestion/src/metadata/ingestion/bulksink/metadata_usage.py
+++ b/ingestion/src/metadata/ingestion/bulksink/metadata_usage.py
@@ -102,7 +102,7 @@ class MetadataUsageBulkSink(BulkSink):
             self.table_usage_map[table_entity.id.__root__] = {
                 "table_entity": table_entity,
                 "usage_count": table_usage.count,
-                "sql_queries": table_usage.sqlQueries,
+                "sql_queries": list(table_usage.sqlQueries),
                 "usage_date": table_usage.date,
                 "database": table_usage.databaseName,
                 "database_schema": table_usage.databaseSchema,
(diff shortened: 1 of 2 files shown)


ingestion/src/metadata/ingestion/lineage/sql_lineage.py
```diff
@@ -136,8 +136,8 @@ def get_table_entities_from_query(
     table_entities = search_table_entities(
         metadata=metadata,
         service_name=service_name,
-        database=database_name,
-        database_schema=database_schema,
+        database=database_query,
+        database_schema=schema_query,
```
ulixius9, line 140: prioratise info retrieved from query

pmbrull (approved):

ingestion/src/metadata/ingestion/bulksink/metadata_usage.py
```diff
@@ -102,7 +102,7 @@ def __populate_table_usage_map(
             self.table_usage_map[table_entity.id.__root__] = {
                 "table_entity": table_entity,
                 "usage_count": table_usage.count,
-                "sql_queries": table_usage.sqlQueries,
+                "sql_queries": list(table_usage.sqlQueries),
```
nahuelverdugo, line 105: Why are we doing this if the `sqlQueries` is already a List? 
```python
    sqlQueries: Optional[List[table.SqlQuery]] = Field(
        None, description='List of SQL Queries associated with table'
    )
```

ulixius9: if you notice [here](https://github.com/open-metadata/OpenMetadata/blob/4885bc4a9ab78ca076eab549ce3db0769ad24261/ingestion/src/metadata/ingestion/bulksink/metadata_usage.py#L215) we are passing the same object of table_usage for each table_entity, and in the map both the table entities will have the same reference to the list of `sqlQueries` and when we are extending the list extend happens for all table entities which should not be the case 

just to make it more clear, below code snippet can summarise the scenario   
image


hence instead of linking the reference to the list, I'm creating a new list and linking that to map, and yes I should take care of nulls here

nahuelverdugo: Oh, I see. It makes total sense! Thank you for the explanation :)

pmbrull: let's check for nulls and bring it in. Thanks both

nahuelverdugo, line 105:
```suggestion
                "sql_queries": list(table_usage.sqlQueries or []),
```

nahuelverdugo (approved):

PR #9965, opened 2023-01-27
Bug: Change password of a user without auth mechanism set

nahuelverdugo:
### Describe your changes :
When basic auth is enabled and the user is created through the API without a password, the stored auth mechanism is null. This can cause a null pointer exception when changing the user password for the first time.

### Type of change :
- [x] Bug fix

### Checklist:

### Reviewers


TeddyCr:
Should we disallow user creation without a password instead?

nahuelverdugo:
> Should we disallow user creation without a password instead?

@TeddyCr when we create a user when the SSO is enabled, the password is not required.

TeddyCr (dismissed):

pmbrull (dismissed):

pmbrull (approved):

PR #10048, opened 2023-02-01
Feat: Avoid running Maven CI on each PR if it is not needed

nahuelverdugo:
### Describe your changes :
If the changes are unrelated to Java or UI code, this change will avoid running the required action of building with Maven on each PR/commit. 

### Type of change :
- [x] Improvement

### Checklist:

### Reviewers


.github/workflows/maven-build-skip.yml
```diff
@@ -0,0 +1,29 @@
+#  Copyright 2021 Collate
```
akash-jain-10, line 1: We don't need another Workflow just to specify "Step is not required". Github Actions will not run on those PR's which does not have changes in the path's mentioned in Maven CI.

nahuelverdugo: As far as I understand it, we are running it even for PRs related to documentation because it is a required action for all of them. It will be skipped when actions are run in the main branch. This is the suggested way to do it by the GitHub [documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks).

akash-jain-10 (approved):

harshach (approved):

Issue #10046, opened 2023-02-01
Support service account impersonation of Google Cloud


yu-iskw:
**Is your feature request related to a problem? Please describe.**
We don't have a serious problem at the moment. But, we would like to secure service account credentials of Google Cloud a little bit more. It would be awesome to support service account impersonation of Google Cloud on the web UI and CLI of OpenMetadata.

**Describe the solution you'd like**
We are able to authenticate Google Cloud using `GCPCredentials`. But, we have to directly grant required permissions to service accounts. So, if the service account credentials is breached, attackers can refer to resources on Google Cloud. But, if we grant only permissions to impersonate a service account which in reality has required permissions, we can a bit mitigate security risk. 

Aside from that, regarding the CLI of OpenMetadata, we don't have to directly grant required IAMs or permissions to execut CLI commands. Let's assume if a data engineer have to manually operate a OpenMetadata CLI command due to some reasons. We don't have to grant the IAMs or permissions to the data engineer. Instead, only we have to do is to grant a permission to impersonate the impersonated service account to the data engineer. By doing so, it would make it easier to manage IAMs and permissions about who can ingest what souces.

```
ource:
  type: bigquery
  serviceName: "<service name>"
  serviceConnection:
    config:
      type: BigQuery
      credentials:
        gcsConfig:
          type: service_account
          # Specify an impersonate service account
          impersonate_service_account: "impersonated-service-account@sample-project.iam.gserviceaccount.com"
          projectId: project ID # ["project-id-1", "project-id-2"]
          privateKeyId: us-east-2
          privateKey: |
            -----BEGIN PRIVATE KEY-----
            Super secret key
            -----END PRIVATE KEY-----
          clientEmail: client@mail.com
          clientId: 1234
          # authUri: https://accounts.google.com/o/oauth2/auth (default)
          # tokenUri: https://oauth2.googleapis.com/token (default)
          # authProviderX509CertUrl: https://www.googleapis.com/oauth2/v1/certs (default)
          clientX509CertUrl: https://cert.url
```

**Describe alternatives you've considered**
N/A

**Additional context**

- https://cloud.google.com/iam/docs/impersonating-service-accounts

PR #10127, opened 2023-02-07
Support to impersonate a service account in the Bigquery source

yu-iskw:
### Describe your changes :
I worked on the new feature to support to impersonate a service account in the BigQuery metadata ingestion, because this allows us to reap the advantages of using federated identity for developers (via ADC) without needing to grant individual access to read and write data directly, and without needing to create separate service account and keys for each user. It also allows you to completely eliminate the need for service account keys in CI as long as your CI is running on GCP.

fix https://github.com/open-metadata/OpenMetadata/issues/10046

### Type of change :

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
N/A

#
### Checklist:

#


ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
```diff
@@ -206,7 +207,16 @@ def get_column_tag_labels(
...
+        quota_project_id = self.service_connection.connectionOptions.get("quotaProjectId", None)
+        location = self.service_connection.connectionOptions.get("location", None)
+        impersonate_service_account = self.service_connection.connectionOptions.get("impersonateServiceAccount", None)
+        lifetime = self.service_connection.connectionOptions.get("lifetime", None)
+        self.client = get_bigquery_client(
+            quota_project_id=quota_project_id,
+            project_id=database_name,
+            impersonate_service_account=impersonate_service_account,
+            location=location,
+            lifetime=lifetime)
```
yu-iskw, line 219: I would like to discuss if we should implement the input entity or not, instead of the connection options.

ingestion/src/metadata/utils/credentials.py
```diff
@@ -137,3 +145,39 @@ def generate_http_basic_token(username, password):
...
+        lifetime = 3600
+
+    source_credentials, _ = auth.default()
+    if quoted_project_id is not None:
+        source_credentials, quoted_project_id = auth.default(quota_project_id=quoted_project_id)
+    target_credentials = impersonated_credentials.Credentials(source_credentials=source_credentials,
+                                                              target_principal=impersonate_service_account,
+                                                              target_scopes=scopes,
+                                                              lifetime=lifetime)
+    return target_credentials
```
yu-iskw, line 183: We can use the functions for other Google Cloud products in the future. For instance, I plan to implement a Google Cloud Spanner source. The function would be useful to support service account impersonation even for Spanner.

pmbrull:
Thanks for working on this. @ayush-shah could you please help with the review?

ingestion/src/metadata/utils/credentials.py
```diff
@@ -137,3 +145,39 @@ def generate_http_basic_token(username, password):
     """
     token = base64.b64encode(f"{username}:{password}".encode("utf-8")).decode("utf-8")
     return token
+
+
+def get_default_credentials(quota_project_id: Optional[str] = None,
```
pmbrull, line 150: nit - It might be helpful to set the function name with `gcp` or `google`, as the credentials module is not tied to GCP. Something like `get_gcp_default_credentials`. Same with `get_gcp_impersonate_credentials`. Thanks

ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
pmbrull: It indeed might be helpful to have these parameters defined in the JSON Schema so that we can give proper descriptions and guidelines to users

pmbrull (commented):
Thanks for working on this. Left a nit improvement we can do on naming, but other than that LGTM. @ayush-shah let us know if there's anything we'd need to care of here. Thanks!

ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
yu-iskw: Alright. I will give it a try to change the JSON schema.

ingestion/src/metadata/utils/credentials.py
yu-iskw: I got it. I will rename the function so that it gets more intuitive.

openmetadata-spec/src/main/resources/json/schema/security/credentials/gcsCredentials.json
```diff
@@ -110,6 +128,19 @@
...
+      "description": "we enable the authenticated service account to impersonate another service account",
+      "oneOf": [
+        {
+          "$ref": "#/definitions/GCPImpersonateServiceAccountValues",
+          "default": {}
+        },
+        {
+          "type": "null"
+        }
+      ]
```
yu-iskw, line 143: I have added the new optional object in addition to the credentials values, because service account impersonation has to work based on the credentials.

ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
```diff
@@ -206,7 +206,19 @@ def get_column_tag_labels(
...
+            self.service_connection.credentials.gcpImpersonateServiceAccount.impersonateServiceAccount
+        )
+        lifetime = (
+            self.service_connection.credentials.gcpImpersonateServiceAccount.lifetime
+        )
+        self.client = get_bigquery_client(
+            project_id=database_name,
+            impersonate_service_account=impersonate_service_account,
+            lifetime=lifetime,
+        )
```
yu-iskw, line 221: I have improved this by getting properties through the object. 

I plan to support other properties in another pull request. One is `location` to specify a location of a GCP client. The other is to specify OAuth 2.0 scopes. For instance, if we have BigQuery external tables whose sources are located in Google Drive, we have to pass `https://www.googleapis.com/auth/drive` in addition to `https://www.googleapis.com/auth/cloud-platform`. It would be good to pass such optional information to no only the GCP credentials but also the target impersonate service account.

yu-iskw:
@pmbrull I found it difficult to implement service account impersonation on top of `python-bigquery-sqlalchemy`. We may have to give up the issue at the moment, if I am correct. 

https://github.com/googleapis/python-bigquery-sqlalchemy/issues/42

yu-iskw:
According to the documentation of `python-bigquery-sqlalchemy`, we can pass a BigQuery client to the function to create an engine. I will give it a try to create a BigQuery client with service account impersonation and then create an engine locally. 

https://github.com/googleapis/python-bigquery-sqlalchemy#supplying-your-own-bigquery-client

harshach (approved):
@ayush-shah review and merge

ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
```diff
@@ -206,7 +206,19 @@ def get_column_tag_labels(
...
     def set_inspector(self, database_name: str):
-        self.client = Client(project=database_name)
+        # TODO support location property in JSON Schema
+        # TODO support OAuth 2.0 scopes
+        impersonate_service_account = (
+            self.service_connection.credentials.gcpImpersonateServiceAccount.impersonateServiceAccount
+        )
+        lifetime = (
+            self.service_connection.credentials.gcpImpersonateServiceAccount.lifetime
+        )
```
ayush-shah, line 216: can we add a check here to check if the connectiontype is selected as `gcpImpersonateServiceAccount` and then get the attributes from that instead. or else it would be throw an error and ingestion would fail if the user selects another type

ayush-shah (dismissed):

ayush-shah:
will take a look as part of 1.2

harshach:
@ayush-shah need a conclusion on this?

ayush-shah:
> @ayush-shah need a conclusion on this?

@harshach this PR might need modifications as part of the conflicts it's facing + as added in the comments, there needs a check over connectionType

ayush-shah (approved):

PR #10132, opened 2023-02-07
Added dbt fixes for source nodes and es parent nodes

OnkarVO7:
### Describe your changes :

Added dbt fixes for source nodes and es parent nodes

#
### Type of change :

- [x] Bug fix

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/src/metadata/utils/dbt_config.py
+++ b/ingestion/src/metadata/utils/dbt_config.py
@@ -182,18 +182,30 @@ def _(config: DbtCloudConfig):  # pylint: disable=too-many-locals
         runs_data = response.get("data")
         if runs_data:
             run_id = runs_data[0]["id"]
-            logger.debug("Requesting [dbt_catalog]")
-            dbt_catalog = client.get(
-                f"/accounts/{account_id}/runs/{run_id}/artifacts/{DBT_CATALOG_FILE_NAME}"
-            )
+            try:
+                logger.debug("Requesting [dbt_catalog]")
+                dbt_catalog = client.get(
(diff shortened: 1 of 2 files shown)


pmbrull (changes requested):
LGTM, but please let's add tests around the changes

ingestion/src/metadata/utils/dbt_config.py
```diff
@@ -182,18 +182,30 @@ def _(config: DbtCloudConfig):  # pylint: disable=too-many-locals
...
-            dbt_catalog = client.get(
-                f"/accounts/{account_id}/runs/{run_id}/artifacts/{DBT_CATALOG_FILE_NAME}"
-            )
+            try:
+                logger.debug("Requesting [dbt_catalog]")
+                dbt_catalog = client.get(
+                    f"/accounts/{account_id}/runs/{run_id}/artifacts/{DBT_CATALOG_FILE_NAME}"
+                )
+            except Exception as exc:
+                logger.info(
```
NiharDoshi99, line 191: I think this should be `warning` instead of `info`

OnkarVO7: @NiharDoshi99 it should be info, since it is a optional thing we only want to inform users that catalog file is not passed

NiharDoshi99: Oh Ok 👍

pmbrull (approved):
we'll get the tests in another PR that is already open

PR #10169, opened 2023-02-09
Relax data asset name restrictions to allow the special characters ex…

sureshms:
…cept "::"

### Describe your changes :

Many databases allow special characters in the object names. With this change, the name restriction is relaxed except names containing "::" which is used as EntityLink separator.

For the entities that are created within OpenMetadata such as Glossary, Tags, TestCase etc. the name restrictions to allow only selected special characters still apply.

#
### Type of change :

- [x] Bug fix
- [x] Improvement

#
### Checklist:


harshach (approved):

sureshms:
@OnkarVO7 and @chirag-madlani, not sure what these test failures are related to. But should not be from this PR.

PR #10176, opened 2023-02-10
update columnValueMaxToBeBetween json schema.

JINZHANG2017:
correct columnValueMaxToBeBetween.json parameter "Min" definitions.


pmbrull (approved):
Thank you, good catch!

harshach (approved):

PR #10207, opened 2023-02-14
Support postgres multiple DB ingestion for user with no access

ulixius9:
### Describe your changes :

Support postgres multiple DB ingestion for user with no access

#
### Type of change :

- [x] Bug fix

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion


openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/postgresConnection.json
```diff
@@ -59,6 +59,12 @@
       "type": "string",
       "default": "PostgresPolicyTags"
     },
+    "ingestAllDatabases": {
+      "title": "Ingest All Databases",
+      "description": "Ingest data from all databases in Redshift. You can use databaseFilterPattern on top of this.",
```
OnkarVO7, line 64: @ulixius9 please change `Redshift`->`Postgres`

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/postgresConnection.json
```diff
@@ -94,5 +100,5 @@
     }
   },
   "additionalProperties": false,
-  "required": ["hostPort", "username"]
+  "required": ["hostPort", "username", "database"]
```
OnkarVO7, line 103: @ulixius9 will the `database` be a required field even in case when `ingestAllDatabases` is set to true?

OnkarVO7 (changes requested):

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/postgresConnection.json
ulixius9: Yes it will be required, when this is similar to what we do for redshift even if we want to ingest all databases we will have to pass at least one database name

OnkarVO7 (dismissed):

NiharDoshi99 (approved):

Issue #10264, opened 2023-02-21
Ingest Database, Schema & Table Owners: Snowflake 


ulixius9:
**Affected module**
Ingestion Framework

**Describe the bug**
Ingest Database, Schema & Table Owners: Snowflake

NiharDoshi99:
Currently we are not getting any owner details from snowflake 


![Image](https://user-images.githubusercontent.com/51595473/224019256-2bffcceb-e82d-4d19-83bd-6468b7a99bac.png)

PR #10276, opened 2023-02-22
Add Database & Schema Description: Snowflake

ulixius9:
### Describe your changes :

Add Database & Schema Description: Snowflake

#
### Type of change :

- [x] Improvement

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion


ingestion/src/metadata/ingestion/source/database/snowflake/metadata.py
```diff
@@ -179,12 +183,39 @@ def set_partition_details(self) -> None:
                     f"{row.TABLE_SCHEMA}.{row.TABLE_NAME}"
                 ] = row.CLUSTERING_KEY
 
+    def set_schema_description_map(self) -> None:
+        results = self.engine.execute(SNOWFLAKE_GET_SCHEMA_COMMENTS).all()
+        for row in results:
+            self.schema_desc_map[(row.DATABASE_NAME, row.SCHEMA_NAME)] = row.COMMENT
+
+    def set_database_description_map(self) -> None:
```
pmbrull, line 191: Random thought: would it make sense to move this to the sqlalchemy_utils? The overall structure reminds me of the methods we have here https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/utils/sqlalchemy_utils.py#L51

ulixius9: @pmbrull  Ummm, in this particular case for snowflake I would prefer not to move this to sqlalchemy_utils because we will have to modify the same query to fetch the owner as well for issue #10264 and we can populate both the owner and comment by executing one query within same function which might not be the same case for other sources!

also, I have not included the owner in this particular pr because currently we just get the username of the owner need to explore ways if we can fetch the email of the owner

pmbrull (approved):

PR #10282, opened 2023-02-22
Added dbt oracle fix

OnkarVO7:
### Describe your changes :

Manifest generated by dbt-oracle gives `database: "None"` in string format instead of standard `null` datatype of json. 
The fix here will handle this usecase.
ref: https://openmetadata.slack.com/archives/C02B6955S4S/p1677047690848149?thread_ts=1675932001.274849&cid=C02B6955S4S

#
### Type of change :

- [x] Bug fix
- [x] Improvement

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion


ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
```diff
@@ -456,7 +456,9 @@ def yield_data_models(self, dbt_objects: DbtObjects) -> Iterable[DataModelLink]:
                             self.metadata,
                             entity_type=Table,
                             service_name=self.config.serviceName,
-                            database_name=manifest_node.database,
+                            database_name=None
+                            if manifest_node.database == "None"
```
pmbrull, line 460: can this only happen for the database? Also it might be better to have a controlled list of elements that we consider as "missing". `None` can be one, not sure if users reported similar strings like `null`, `none`, etc. In any case, we can have a separated constant to track this out.

pmbrull (changes requested):
Let's reflect this changes on the tests please 🙏

NiharDoshi99 (dismissed):

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
OnkarVO7: makes sense will add a constant list and include these keywords

geoHeil:
Referencing the reason https://github.com/oracle/dbt-oracle/issues/74 why we have this PR

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
OnkarVO7: @pmbrull have made the changes can you please take a look now

pmbrull (commented):
thanks for the changes. Let's add this in the unit tests. Thanks

pmbrull (approved):

PR #10299, opened 2023-02-23
chore(ui): add check for notification is present or not in the browser

Sachin-chaurasiya:
### Describe your changes :

I worked on adding a check for whether a notification is present in the window because some browser does not support the notification API.

#
### Type of change :

- [x] Bug fix

#
I have tested on a simulator and the alert shown is only for testing purposes.

https://user-images.githubusercontent.com/59080942/220907380-c42d4e08-ef1d-417d-8dfb-cd20136f17d3.mov

#
### Checklist:

#
### Reviewers
@open-metadata/ui

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/utils/BrowserNotificationUtils.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/BrowserNotificationUtils.ts
@@ -11,11 +11,11 @@
  *  limitations under the License.
  */
 
-export const isNotificationAPIAvailable = () => 'Notification' in window;
+export const isNotificationAPIAvailable = 'Notification' in window;
 
 export const shouldRequestPermission = () =>
-  isNotificationAPIAvailable() &&
+  isNotificationAPIAvailable &&
   !['granted', 'denied'].includes(Notification.permission);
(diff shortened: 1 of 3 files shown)


openmetadata-ui/src/main/resources/ui/src/utils/BrowserNotificationUtils.ts
```diff
@@ -11,8 +11,11 @@
  *  limitations under the License.
  */
 
+export const isNotificationAPIAvailable = () => 'Notification' in window;
```
chirag-madlani, line 14: Can we store it as variable and not as function?

Sachin-chaurasiya: We can do that, but I think storing it in function will be more appropriate if we want to check it multiple times throughout our code.

Sachin-chaurasiya: I will change it to variable since we do not want to check it multiple times.

chirag-madlani (approved):

PR #10313, opened 2023-02-24
Add Webp instead of png

ayush-shah:
Optimize all png's to webp ( better compression + file size reduction )


ayush-shah:
Closing this as current PNG optimized have lesser image size compared to new webp here

PR #10337, opened 2023-02-26
Store test connection output

pmbrull:
### Describe your changes :

So that we can pick it up in argo templates

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#

### DIFF
--- a/ingestion/operators/docker/test_connection.py
+++ b/ingestion/operators/docker/test_connection.py
@@ -12,13 +12,16 @@
 Entrypoint to test the connection to a source
 """
 import os
+from typing import Optional
 
 import yaml
 
 from metadata.generated.schema.api.services.ingestionPipelines.testServiceConnection import (
     TestServiceConnectionRequest,
 )
+from metadata.ingestion.connections.test_connections import TestConnectionResult
 from metadata.ingestion.source.connections import get_connection, get_test_connection_fn
+from metadata.utils.constants import UTF_8
 from metadata.utils.secrets.secrets_manager_factory import SecretsManagerFactory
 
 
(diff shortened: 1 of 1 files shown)


OnkarVO7 (approved):

ingestion/operators/docker/test_connection.py
```diff
@@ -54,7 +57,10 @@ def main():
     test_connection_fn = get_test_connection_fn(
         test_service_connection.connection.config
     )
-    test_connection_fn(connection)
+    results: Optional[TestConnectionResult] = test_connection_fn(connection)
+    if results:
+        with open("/tmp/test_connection_results.json", "w", encoding=UTF_8) as file:
```
harshach, line 62: if there are multiple test connections going on , wouldn't this override the file with different connection results.
lets say there are 2 users logged in and each with their own sessions and simultaneously adding services can get unwanted results.
Also the case where we write a test_connection_results.json from previous test connection, current test connection fails , the polling of this file can return a stale result from previous file
IMO, storing these results in a file doesn't look like a good idea .
Might as well expand the connection on the server side to have test connection and write the results back there instead in a file

harshach (changes requested):
Lets look at alternate options instead of writing to a file

ingestion/operators/docker/test_connection.py
pmbrull: these will happen running on an ephemeral pod. When we call the test_connection from Argo, the workflow will execute on a pod, write locally, and then expose the results via the output param read from the file: https://argoproj.github.io/argo-workflows/walk-through/output-parameters/

pmbrull:
Closing this one. When we add the status as part of the service, we won't need this to send the data out of the pod. We will communicate entirely through API: Pod -> OM API <- UI

pmbrull:
Closing this one. When we add the status as part of the service, we won't need this to send the data out of the pod. We will communicate entirely through API: Pod -> OM API <- UI

PR #10350, opened 2023-02-27
Fixes  #7758 - Added Column value and Integer Range Partitionning

### LINKED ISSUE #7758: Add support for Value Range Partition
**Context**
Value based partition should ne supported for the profiler

TeddyCr:
### Describe your changes :
Fixes  #7758. This PR:
- rename `orm_profiler` module to `profiler`
- introduce `ProfilerProtocol.create` method to instantiate the correct profiler interface based on the service/connection type. This should help extend the profiler more easily to orther forms of entities in OpenMetadata
- Add support for partitionning to datalake profiler
- Add support for values based and column based partitioning on the ingestion side

#
### Type of change :

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/connections/session.py
```diff
@@ -26,7 +26,7 @@ def create_and_bind_session(engine: Engine) -> Session:
     return session()
 
 
-def create_and_bind_thread_safe_session(engine: Engine) -> Session:
+def create_and_bind_thread_safe_session(engine: Engine) -> scoped_session:
```
pmbrull, line 29: Thanks for catching these hints

ingestion/src/metadata/profiler/api/workflow.py
```diff
@@ -375,6 +273,15 @@ def get_database_entities(self):
             if self.filter_databases(database)
         ]
 
+        if not databases:
```
pmbrull, line 276: nice moving it here

ingestion/src/metadata/profiler/api/workflow.py
```diff
@@ -428,68 +335,81 @@ def copy_service_config(self, database) -> DatabaseService.__config__:
...
+                ),
+                entity,
+                self.get_config_for_entity(entity),
+                self.source_config,
+                copied_service_config,
+                create_ometa_client(self.metadata_config),
+                sqa_metadata=sqa_metadata,
             )
+            self.create_profiler(entity, profiler_interface)
+            self.profiler = cast(Profiler, self.profiler)  # satisfy type checker
```
pmbrull, line 361: are you already running mypy on the profiler sources? We should add this everywhere

ingestion/src/metadata/profiler/api/workflow.py
```diff
@@ -428,68 +335,81 @@ def copy_service_config(self, database) -> DatabaseService.__config__:
...
         Run the profiling and tests
         """
 
         try:
-            self.run_profiler_workflow()
+            for database in self.get_database_entities():
+                copied_service_config = self.copy_service_config(database)
+                sqa_metadata = (
+                    MetaData()
+                    if not isinstance(copied_service_config, DatalakeConnection)
```
pmbrull, line 398: nit - if we start adding instance validations, should we keep it a tad more generic to allow room for growth? For example, for when we add Spark, or any other source to profile

ingestion/src/metadata/profiler/profiler/interface/profiler_protocol.py
```diff
@@ -0,0 +1,239 @@
...
+
+class ProfilerProtocol(ABC):
+    """Protocol interface for the profiler processor"""
+
+    _profiler_type: Optional[str] = None
+    subclasses = {}
+
+    def __init_subclass__(cls, *args, **kwargs) -> None:
+        """Hook to map subclass objects to profiler type"""
+        super().__init_subclass__(*args, **kwargs)
```
pmbrull, line 46: I might need an example here to follow this 🤔 Maybe worth it to have it as part of the docstring itself

ingestion/src/metadata/profiler/profiler/sampler.py
```diff
@@ -165,14 +168,52 @@ def _partitioned_table(self) -> Query:
             self.table.__table__.c,
         )
 
-        if not self._partition_details.partitionValues:
-            sample = (
+        if (
+            self._partition_details.partitionIntervalType
+            == PartitionIntervalType.COLUMN_VALUE
```
pmbrull, line 173: sorry broad question - does this Sampler handle the pandas too?

TeddyCr: it does not. Right now pandas sampler will be in `datalake_sampler.py`.

ingestion/src/metadata/profiler/api/workflow.py
TeddyCr: I have it running in VSCode. We might want to do it chunk by chunk, wdyt?

ingestion/src/metadata/profiler/api/workflow.py
TeddyCr: Do you have an example in mind?

ingestion/src/metadata/profiler/profiler/interface/profiler_protocol.py
TeddyCr: Basically when a class inherit from another class `__init_subclass__` is called on the parent. Our interfaces inherit from the protocol. These subclass will be registered in the subclass dict with the key as the connection type and the value as the interface class itself.

When we call `.create()` we will instantiate the class based on the connection type of the service

```python
        return cls.subclasses[_profiler_type](
            ometa_client=ometa_client,
            thread_count=thread_count,
            entity=entity,
            profile_sample_config=profile_sample_config,
            sample_query=sample_query,
            table_partition_config=table_partition_config,
            **kwargs,
        )
```

If we want to add a dashboard service profiler, we should only have to create a new interface class that inherits from the protocol.

ingestion/src/metadata/profiler/api/workflow.py
pmbrull: Instead of `isinstance(copied_service_config, DatalakeConnection)`, we could have `isinstace(copied_service_config,  SUPPORTED_NON_SQA_CONNECTIONS` or something like that, where `SUPPORTED_NON_SQA_CONNECTIONS = (DatalakeConnection, DeltalakeConnection,....)`.

I believe if we pass a tuple of types this should work https://stackoverflow.com/questions/33311258/python-check-if-variable-isinstance-of-any-type-in-list

pmbrull: then whenever we add a new profiler, we just need to prep the implementation of the metrics etc, and add the class in `SUPPORTED_NON_SQA_CONNECTIONS` without having to change any other code

ingestion/src/metadata/profiler/api/workflow.py
pmbrull: sounds great. If you have it on your end already, IMO after this is closed let's add the `mypy` recipe in the Makefile and put it in the CI only for this. We can then keep adding other directories

ingestion/src/metadata/profiler/api/workflow.py
TeddyCr: Makes sense. I'll update this.

ingestion/src/metadata/profiler/api/workflow.py
TeddyCr: https://github.com/open-metadata/OpenMetadata/issues/10358

pmbrull (approved):

PR #10378, opened 2023-03-01
Fix #10292: Remove DB & Schema Usage Filters

### LINKED ISSUE #10292: Remove database and schema query filters for usage
**Affected module**
Ingestion Framework

**Describe the bug**
We are apply the database and schema filter to fetch query logs of the databases and schema that are already ingested into OMD, but it it possible that the query was executed on a different database but refer to some other database by using fully qualified name of table or it is also possible that the query was executed without configuring the DB, in that case the database field in query history table might be null to avoid such scenarios we should remove the db and schema filter from usage & lineage!

ulixius9:
### Describe your changes :

Fix #10292: Remove DB & Schema Usage Filters

#
### Type of change :

- [x] Bug fix
- [x] Improvement

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion

### DIFF
--- a/ingestion/tests/unit/test_usage_filter.py
+++ b/ingestion/tests/unit/test_usage_filter.py
@@ -30,8 +30,6 @@ from metadata.generated.schema.type.entityReference import (
 )
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
 from metadata.ingestion.source.database.clickhouse.usage import ClickhouseUsageSource
-from metadata.ingestion.source.database.mssql.usage import MssqlUsageSource
-from metadata.ingestion.source.database.redshift.usage import RedshiftUsageSource
 
 T = TypeVar("T", bound=BaseModel)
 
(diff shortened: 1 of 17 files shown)


ingestion/tests/unit/test_usage_filter.py
```diff
@@ -210,26 +148,6 @@ class UsageQueryFilterTests(TestCase):
...
-        config = OpenMetadataWorkflowConfig.parse_obj(mock_redshift_config)
-        redshift_source = RedshiftUsageSource.create(
-            mock_redshift_config["source"],
-            config.workflowConfig.openMetadataServerConfig,
-        )
-        redshift_source.prepare()
-        assert redshift_source.db_filters == EXPECTED_REDSHIFT_FILTER
-
     @patch.object(OpenMetadata, "list_all_entities", mock_list_entities)
     def test_prepare_clickhouse(self):
```
ulixius9, line 152: clickhouse stores the list of dbs involved in the query but rest of them does not so when a query contains table from different dbs like `db.schema.table` rest of the sources only stores the db on which the query was executed hence it filter outs the queries which might be relevant!

pmbrull:
will this have any impact on the queries that we can actually ingest?

ulixius9:
@pmbrull except clickhouse no other db provides this info so without parsing we cannot know if it will be "ingestable"

OnkarVO7 (approved):

PR #10383, opened 2023-03-01
Fix #10382 - Test Connection Definitions, Results and Resource

### LINKED ISSUE #10382: Centralise Test Connection definitions and results
## Context

We want to able to thoroughly test the different steps required to fully extract the metadata of a service. Previously, we only tested the networking and basic authentication (e.g., we can connect to the database and run a simple query). Now, we want to be able to cover all the required functionalities: Listing schemas, databases, tables, tags, and lineage,...

These additional steps will also need to be reflected in the UI, and we want to give the user early feedback on everything that is being tested, instead of just returning a batch count of results.

## Proposal

The idea will then be to:
1. Centralise the Test Connection Definitions in the server, similar to how we create test definitions for DQ.
2. Each service will have its own Test Connection Definition. This will be stored as JSON data and created when the server starts.
3. We will have a new JSON Schema defining Test Connection Results, as the outcome of computing the Test Connection Definition for a given service.
4. Each `service` will have a property of the Last Test Connection Results computed on it.
5. When testing a connection, the Python function will:
    1. GET the Test Connection Definition for the service type
    2. Check the steps coming from the server
    3. Look for the method implementing each step and execute it
    4. Update the Test Connection Results for the service after each step is computed
6. The UI will poll the test connection status of the service to keep getting the results. We'll have a `RUNNING` status which will only be updated after the last step in the Python side has been computed. This will tell the UI to keep polling.

## Implementation Details
1. We'll create a Test Connection Definition JSON Schema & Resource
2. During the server start, the definitions will be created from JSON data
3. The Test Connection Result is a JSON Schema and will be used as a simple property of a `service`

pmbrull:
### Describe your changes :

Fix #10382

- Create Test Connection Definition Resource & Repository
- Add sample JSON data files that instantiate some tests
- New table created in the migration files to handle the test defs

#
### Type of change :

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestDefinitionResource.java
```diff
@@ -233,7 +233,6 @@ public TestDefinition get(
       })
   public TestDefinition getByName(
       @Context UriInfo uriInfo,
-      @Parameter(description = "Id of the test definition", schema = @Schema(type = "UUID")) @PathParam("id") UUID id,
```
pmbrull, line 236: This was throwing a warning on IntelliJ, I believe it's not needed

nahuelverdugo (approved):

PR #10448, opened 2023-03-06
Don't store the OM connection in the Ingestion Pipeline or Workflow

pmbrull:
### Describe your changes :

Instead, we'll set it dynamically in the resource for the run and deploy methods.

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


nahuelverdugo (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
```diff
@@ -85,19 +85,25 @@ public void storeEntity(IngestionPipeline ingestionPipeline, boolean update) thr
...
       secretsManager.encryptOrDecryptIngestionPipeline(ingestionPipeline, true);
+      // We store the OM sensitive values in SM separately
+      openmetadataConnection =
+          secretsManager.encryptOrDecryptOpenMetadataConnection(openmetadataConnection, true, true);
     }
 
     // Don't store owner. Build it on the fly based on relationships
-    ingestionPipeline.withOwner(null).withService(null).withHref(null);
+    // We don't want to store the OM connection.
+    ingestionPipeline.withOwner(null).withService(null).withHref(null).withOpenMetadataServerConnection(null);
```
ayush-shah, line 101: Would we be able to use decryptorNullify here?
sorry for asking the above question as not sure what is the main func of that, but based on name, i think it might be able to set children objects to null

nahuelverdugo: `decryptorNullify` is a method we use in the resources before sending the responses from the API to decrypt the values stored in DB, which are encrypted. However, in the repository, we set those values to null since the relationship with those Objects is built on the fly based on the relationships.

nahuelverdugo (approved):

PR #10455, opened 2023-03-07
Fix #10454: Improve Search Relevancy, by adding functional scoring an…

### LINKED ISSUE #10452: Enable Table and Column search by BOTH name and displayName
For example, some systems like SAP have Field Name and Field Label. And Also Table Name and Table label: https://www.leanx.eu/en/sap/table/ekko.html. Both are widely used. So one would expect to find either EKKO or Purchasing Document Header equally when searching for the table. Technical name is EKKO (what is visible in integrations, but users know it by Purchasing Document Header). 

To include BOTH name and displayName in the search. Meaning, as per above example. I could search for EKKO or for Purchasing Document Header and it should take me to the same table/return table in both cases in the search result. IT users and Business users search for tables / columns differently - IT/ Data users use technical names and Business Users - Labels. And I was under impression this is the reason why we have Name and Display Name in the first place.

Providing name in the description, so it is searchable, but then I need to override application metadata.

**Additional context**

### LINKED ISSUE #10454: Improve Search Relevancy, by adding functional scoring and add ngram analyzer
**Feature**
1. Better weightage to the entity names and displayname
2. Similarly matched results will be boosted based on their weekly usage count (based on query logs)
3. Adding Ngram for partial word matches

**Describe the task**

harshach:
…d add ngram analyzer; Fix #10452: Enable Table and Column search by BOTH name and displayName

### Describe your changes :

#
### Type of change :

- [x] Improvement

- [ ] Documentation

#
</p>

#
### Checklist:

#

### DIFF
--- a/ingestion/examples/sample_data/datasets/tables.json
+++ b/ingestion/examples/sample_data/datasets/tables.json
@@ -1,7 +1,7 @@
 {
       "tables": [
           {
-              "name": "dim_address",
+              "name": "bl_vertical_lists",
               "description": "This dimension table contains the billing and shipping addresses of customers. You can join this table with the sales table to generate lists of the billing and shipping addresses. Customers can enter their addresses more than once, so the same address can appear in more than one row in this table. This table contains one row per customer address.",
               "version": 0.1,
               "updatedAt": 1638354087391,
@@ -135,15 +135,15 @@
               "tags": [],
               "usageSummary": {
                   "dailyStats": {
-                      "count": 100,
+                      "count": 73,
                       "percentileRank": 45
                   },
                   "weeklyStats": {
-                      "count": 100,
+                      "count": 73,
                       "percentileRank": 45
                   },
                   "monthlyStats": {
-                      "count": 100,
+                      "count": 67,
                       "percentileRank": 45
                   },
                   "date": "2021-12-01"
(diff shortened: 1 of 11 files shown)


nahuelverdugo (commented):
I have set back the name of the example tables since it was breaking some tests.
Apart of that I notice we stop showing why it is matching in the results like this:
image
Now it does not appear anything. Is it expected? do we want to handle this change in a different PR/issue?

nahuelverdugo (dismissed):

nahuelverdugo (dismissed):

harshach:
> I have set back the name of the example tables since it was breaking some tests. Apart of that I notice we stop showing why it is matching in the results like this: image Now it does not appear anything. Is it expected? do we want to handle this change in a different PR/issue?

@nahuelverdugo not understanding the use cases?

harshach:
@nahuelverdugo searched for address_id and we do show it matches a column name 
image

nahuelverdugo:
@harshach, if you create a table with name and display_name that are different when you search for the display_name, it does not show why the search matches the table.

chirag-madlani:
@harshach Have verified the Cypress Failures, It's flaky and I will make them consistent over the time

ShaileshParmar11 (approved):
UI side looks good

nahuelverdugo (approved):

Issue #10420, opened 2023-03-03
Some Tables are not getting ingested in Databricks


MilanBariya:
**Affected module**
Ingestion Framework

**Describe the bug**
When the column type is `ARRAY<MAP<STRING, FLOAT>>, ARRAY<STRING>, MAP<STRING, STRING>`
that tables are not getting ingested.

**To Reproduce**
Create a table in Databricks that has column type `ARRAY<MAP<STRING, FLOAT>>, ARRAY<STRING>, MAP<STRING, STRING>`

**Expected behavior**
All tables should be ingested.

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8]
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**
Ref: https://openmetadata.slack.com/archives/C02B6955S4S/p1677771316973059

PR #10469, opened 2023-03-08
Fix: Databricks complex column type issue

MilanBariya:
### Describe your changes :

When the column type is ARRAY<MAP<STRING, FLOAT>>, ARRAY<STRING>, MAP<STRING, STRING>
that tables are also getting ingested.
Fix: #10420

Screenshot 2023-03-08 at 2 42 34 PM
Screenshot 2023-03-08 at 2 42 06 PM

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
```diff
@@ -127,12 +127,14 @@ def get_columns(self, connection, table_name, schema=None, **kw):
             if db_name is not None:
                 rows = dict(
                     connection.execute(
-                        f"DESCRIBE {db_name}.{table_name} {col_name}"
+                        f"DESCRIBE {db_name}.{schema}.{table_name} {col_name}"
```
OnkarVO7, line 130: @MilanBariya  can the `schema` be `None` in any case?

MilanBariya: @OnkarVO7 `schema` will not be `None`,  either it will come from config, or it will take from `database`.

ulixius9: I would still add the handling for none just in case we have missed something

MilanBariya: Sure, let me handle that.

pmbrull: nit - might be better to keep just `if db_name and schema` instead of passing the `schema is not None`. Here the DESCRIBE would fail for empty strings for schemas. Better to check for the false value of the whole string

pmbrull (dismissed):

NiharDoshi99 (approved):

PR #10470, opened 2023-03-08
Pyimpala fix colnames, comments and dialect sql compilation

pmbrull:
### Describe your changes :

- Colnames were loaded as `table.col`. Fixed to be just `col`.
- We did not retrieve table comments. Ammended that.
- We were missing SQL compilation for Impala which was breaking some metrics.

We still need to figure out some SQL errors on the profiler. I am wondering if my setup is just not correct. Will dig further.

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/hive/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/hive/metadata.py
@@ -205,7 +205,9 @@ def get_impala_columns(self, connection, table_name, schema=None, **kwargs):
         column_info.append(
             {
                 "name": remove_table_from_column_name(table_name, col[0]),
-                "type": _impala_type_to_sqlalchemy_type[col[1]],
+                # Using Hive's map instead of Impala's, as we are pointing to a Hive Server
+                # Passing the lower as Hive's map is based on lower strings.
+                "type": _type_map[col[1].lower()],
                 "nullable": True,
                 "autoincrement": False,
             }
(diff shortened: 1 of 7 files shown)


ingestion/src/metadata/ingestion/source/database/hive/metadata.py
```diff
@@ -182,9 +185,40 @@ def get_table_comment(  # pylint: disable=unused-argument
...
+    schema = cursor.cursor.description
+    # We need to fetch the empty results otherwise these queries remain in
+    # flight
+    cursor.fetchall()
+    column_info = []
+    for col in schema:
+        column_info.append(
+            {
+                "name": remove_table_from_column_name(table_name, col[0]),
+                "type": _impala_type_to_sqlalchemy_type[col[1]],
```
ulixius9, line 208: I'm not 100% sure if this would always work because there are many types which are supported by hive but not by the impala for example `binary` in that case it might fail unless the query returns something else for a binary `datatype` 🤔

pmbrull: thanks for the catch. Updated using Hive's type map

ulixius9 (dismissed):

ulixius9 (approved):

Issue #10463, opened 2023-03-07
Ingestion will fail to delete table if testCase is attached to the entity


TeddyCr:
**Context**
When running the ingestion workflow, if "Mark Table as Deleted" is toggled and a table does not exists in the DB but has a testCase assigned in OM, the ingestion will fail to delete this table from OM throwing the below error

```
[2023-03-07 09:11:02,122] {metadata_rest.py:350} DEBUG - Traceback (most recent call last):
  File "/opt/airflow/venv/lib/python3.8/site-packages/metadata/ingestion/ometa/client.py", line 214, in _one_request
    resp.raise_for_status()
  File "/opt/airflow/venv/lib/python3.8/site-packages/requests/models.py", line 1022, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://localhost:8585/api/v1/tables/575f8cfc-5339-41e0-bde7-b67517528111?recursive=false&hardDelete=false

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/airflow/venv/lib/python3.8/site-packages/metadata/ingestion/sink/metadata_rest.py", line 345, in delete_table
    self.metadata.delete(entity=Table, entity_id=record.table.id)
  File "/opt/airflow/venv/lib/python3.8/site-packages/metadata/ingestion/ometa/ometa_api.py", line 725, in delete
    self.client.delete(url)
  File "/opt/airflow/venv/lib/python3.8/site-packages/metadata/ingestion/ometa/client.py", line 308, in delete
    return self._request("DELETE", path, data)
  File "/opt/airflow/venv/lib/python3.8/site-packages/metadata/ingestion/ometa/client.py", line 192, in _request
    return self._one_request(method, url, opts, retry)
  File "/opt/airflow/venv/lib/python3.8/site-packages/metadata/ingestion/ometa/client.py", line 222, in _one_request
    raise APIError(error, http_error) from http_error
metadata.ingestion.ometa.client.APIError: table is not empty

[2023-03-07 09:11:02,122] {metadata_rest.py:351} ERROR - Unexpected error deleting table [__root__='<TABLE NAME>']: table is not empty
```

**How to reproduce it**
1. ingest metadata w/ mark table as deleted toggled
2. create test case for a table
3. delete the table from DB
4. run ingestion

**Expected behavior**
The table should be deleted

PR #10476, opened 2023-03-08
fix: mark table deleted dependencies

NiharDoshi99:
### Describe your changes :

I worked on soft deleting table if testCase is attached
Fix: #10463 

#
### Type of change :

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation


openmetadata-service/src/main/java/org/openmetadata/service/resources/databases/TableResource.java
```diff
@@ -383,7 +383,7 @@ public Response delete(
           boolean hardDelete,
       @Parameter(description = "Id of the table", schema = @Schema(type = "UUID")) @PathParam("id") UUID id)
       throws IOException {
-    return delete(uriInfo, securityContext, id, false, hardDelete);
+    return delete(uriInfo, securityContext, id, true, hardDelete);
```
pmbrull, line 386: Please don't change this in the default value of the API... We'll need to pass this as a flag in the `ometa_api` delete call

pmbrull (changes requested):

openmetadata-service/src/main/java/org/openmetadata/service/resources/databases/TableResource.java
NiharDoshi99: Yes

nahuelverdugo (approved):

PR #10486, opened 2023-03-08
Fixes #10480 Glossary rename results in rename of Classification with…

### LINKED ISSUE #10480: Glossary terms and tags overlapping in tag_usage table
## Bug

We can end up with entries in `tag_usage` table pointing to entities that do not exist after a rename.

This breaks the UI for any entity being tagged with the "phantom" tag/glossary term.

## To reproduce

1. Create a glossary with a name `My Name` and a glossary term
image

2. Create a Classification with the same name `My Name` and some tags
image

3. Go to a table and tag it with the glossary term
image

This is how the tag_usage looks like now

image

4. Rename your classification to `New Name`. This updates the `tagFQN` entry in `tag_usage` to something that does not exist:

image

5. Try to go to the table. It will blow up as it cannot retrieve the tag set in `tag_usage`.
image

sureshms:
… the same name

### Describe your changes :

See #10480 for more details
#
### Type of change :

- [x] Bug fix

#
### Checklist:


harshach (dismissed):

pmbrull (approved):

sureshms:
@chirag-madlani can the UI team review this? The main change is, the enum from`TagSource` called `Tag` has been renamed to `Classification`.

ShaileshParmar11 (approved):
UI side looks good

chirag-madlani:
@ShaileshParmar11 Let's use enum if we can instead hardcoded string for easy tracking

PR #10515, opened 2023-03-10
Fixes #9064 - Ingestion for S3 bucket containers

### LINKED ISSUE #9064: Storage Service connector
Thanks to everyone participating in the discussion https://github.com/open-metadata/OpenMetadata/discussions/8124.

Let's start writing down the initial ideas. Note that this is a WIP, and we'll keep curating the points as the design discussions happen.

## Context

What is happening today:

- Ingesting individual files as separated `Table`s. This does not properly represent partitioned data. We are not applying any difference between "structured" files (representing a table) vs. unstructured files (any unorganized bucket/path).
- The path is used as the `Table` name, which can sometimes be longer than the current `name` limitations https://github.com/open-metadata/OpenMetadata/issues/6889
- We are not surfacing `Location` entities. Currently, only the Glue connector uses `Location` data as the EXTERNAL TABLE path information. 

## Initial Goals & Thoughts

- We ingest Buckets as `Location` entities and add the Paths inside them. EXTERNAL TABLEs are then linked to a Path within a Bucket.
- We do not ingest all tables' files as separate Entities. While configuring an input sample size, we check if the files within a directory all follow the same schema. If so, the full directory will be represented as a single `Table`.
    - We will build a set of examples of supported inputs/outputs: what happens if all files follow the same schema, if all files minus one have the same schema, etc.
- Figure out how to scope the ingestion of structured and unstructured prefixes. Should we allow a list of paths?
- For data without implicit schema (JSON, CSV, TSV), define rules for either fetching the schema file or requiring an input field for pointing to the schema location.
- Develop a "Structured Datalake" connector that will group and analyze schemas to ingest Tables + a `StorageService` connector that will ingest each individual file as files.

### Figuring out the JSON Schema

The first solution was to have a top-level entity `Bucket` or `Folder` containing the internal paths as the second Entity. The issue here is how to easy the discovery at each different level (`level1/level2/level3`). Also, this gives special treatment to buckets, which might be useful for some scenarios, but what happens if we have a generic `datalake` bucket and the important pieces we want to explore are the layers (e.g., RAW, L1, L2,..., or `DEV`, `SIT`, `PROD`,...)?

The second (and final) approach has been to prepare a more generic structure with a single `Container` Entity, that can have other Containers as children. This allows to give the same importance regardless of the level of the path. You can find the definition of the JSON Schema [here](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/entity/data/container.json).

- NOTE: Why the name `Container`? Because `Bucket` can seem cloud-specific and only refer to the top-level asset. We need a term that can refer to a real AWS Bucket, an internal blob in there, and even a Dropbox path when/if we include that in the future.

### Structured vs. Unstructured

The second difficulty is figuring out how to handle structured paths i.e., those that uniquely identify to a real "table", with a specific schema, format, etc. vs. the unstructured paths, where we just have free format files such as mp4, jpg, pkl,...

To handle this, our JSON Schema has an optional property `dataModel`, where the schema will be informed in the cases where it applies.

### Building the Connector -> Crawling solution

Once the APIs are in place and the JSON Schema is defined, the difficulty comes when automating the metadata discovery. Every request has an associated cost and we cannot easily know what to expect when ingesting a bucket. Is the data going to be structured? How many files should we sample for us to flag a directory as structured or not? How many paths should we try to discover? Are all of them going to be important for the user? And many more.

Given the complexity and the cost of false positives during the crawling, or just by the fact of ingesting metadata that might not even be useful, we've made the decision to direct the ingestion based on a `manifest.json` file to be loaded when configuring the service.

### Manifest JSON

In the manifest, we'll specify in which paths we have the data we want to surface, if it's partitioned or not, and the format. This will help the ingestion workflow go directly where the bread and butter is.

By default, we'll try to ingest all the buckets (top-level containers), and keep high-level information about them:
- number of objects
- size of the bucket
But we're not going to jump into its contents to avoid any errors and edge cases.

---

## Roadmap

We'll be taking this in two steps:
1. Prepare the Object Storage APIs, Connector and UI
2. Deprecate the current Storage Service for the new Object Storage. We'll keep the name `Service Storage`. We will also deprecate the current `Location` Entity. Instead, we'll add a `type` to the properties of the containers which will specify if it's an `object`, `path`, etc. depending on the source.

cristiancalugaru:
### Describe your changes :

Fixes #9064

Initial support for S3 bucket containers. This includes:
Support for ingesting containers where a metadata template file exists in the root path of the bucket, containing information about the path and structure of files which should be considered. Support includes csv, tsv, json, parquet and avro (the same as the datalake connector). Sample metadata file:

`{
    "entries":[
        {
            "dataPath": "transactions",
            "isStructured": true,
            "structureFormat": "csv",
            "isPartitioned": false
        }
    ]
}
`
Support for ingesting a bucket without any structure defined in a metadata file - this is restricted to creating a container entity without any data model and fileTypes, but with details about total nr of files and total size in bytes of all the files(for files in the STANDARD storage type class only).

Also, changed numberOfObjects and size schema types for a container to Number (and consequently to float in python) to support larger values.

#
### Type of change :

- [x] New feature

#
### Reviewers
@pmbrull


ingestion/src/metadata/utils/s3_utils.py
```diff
@@ -115,3 +115,11 @@ def read_avro_from_s3(client: Any, key: str, bucket_name: str):
...
         get_file_text(client=client, key=key, bucket_name=bucket_name)
     )
+
+
+def prefix_exits(client: Any, bucket_name: str, prefix: str):
+    """
+    Checks if a given prefix exists in a bucket
+    """
+    res = client.list_objects_v2(Bucket=bucket_name, Prefix=prefix, MaxKeys=1)
+    return 'Contents' in res
```
pmbrull, line 125: Should this be a constant? I see it use it in a few places

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
```diff
@@ -0,0 +1,287 @@
...
+            fileFormats=[],  # TODO should we fetch some random files by extension here? Would it be valuable info?
+            dataModel=None
+        )
+
+    def _get_sample_file_path(self, bucket_name: str, metadata_entry: MetadataEntry) -> Optional[str]:
+        """
+        Given a bucket and a metadata entry, returns the full path key to a file which can then be used to infer schema
+        or None in the case of a non-structured metadata entry, or if no such keys can be found
+        """
+        result = f"{metadata_entry.dataPath.strip('/')}"
```
pmbrull, line 268: '/' might be a separator constant too

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
```diff
@@ -0,0 +1,287 @@
...
+        if not metadata_entry.isStructured:
+            return None
+        if metadata_entry.isPartitioned and metadata_entry.partitionColumn:
+            result = f"{result}/{metadata_entry.partitionColumn.strip('/')}"
+
+        response = self.s3_client.list_objects_v2(Bucket=bucket_name, Prefix=result)
+        # no objects found in the data path
+        if 'Contents' not in response:
```
pmbrull, line 276: Sould this be a call to `self. prefix_exits`?

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
```diff
@@ -0,0 +1,287 @@
...
+            return None
+        # this will look only in the first 1000 files under that path (default for list_objects_v2).
+        # We'd rather not do pagination here as it would incur unwanted costs
+        candidate_keys = [
+            entry['Key'] for entry in response['Contents'] if
+            entry and entry['Key'] and entry['Key'].endswith(metadata_entry.structureFormat)
```
pmbrull, line 282: if `Key` is not a key in `entry` this will blow up. Maybe check `entry.get("Key")`

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
```diff
@@ -0,0 +1,287 @@
...
+            )
+        return cls(config, metadata_config)
+
+    def get_containers(self) -> Iterable[S3ContainerDetails]:
+        """
+        List and filter containers
+        """
+        buckets = []
+        try:
+            buckets = self.s3_client.list_buckets()['Buckets']
```
pmbrull, line 117: Should we `.get("Buckets")` and handle `None` in buckets?

pmbrull: rather than allowing this to blow up in a KeyError

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
```diff
@@ -0,0 +1,287 @@
...
+class S3BucketResponse(BaseModel):
+    """
+    Class modelling a response received from s3_client.list_buckets operation
+    """
+
+    class Config:
+        extra = Extra.forbid
+
+    Name: str = Field(..., description="Bucket name", title="Bucket Name")
+    CreationDate: datetime = Field(
```
pmbrull, line 68: We could stick to Python naming conventions, such as `name` or `creation_date`

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
```diff
@@ -0,0 +1,287 @@
...
+        try:
+            buckets = self.s3_client.list_buckets()['Buckets']
+        except Exception as err:
+            logger.debug(traceback.format_exc())
+            logger.error(f"Failed to fetch buckets list - {err}")
+
+        # No pagination required, as there is a hard 1000 limit on nr of buckets per aws account
+        for bucket in buckets:
+            bucket_response: S3BucketResponse = S3BucketResponse.parse_obj(bucket)
+            bucket_name = bucket_response.Name
```
pmbrull, line 125: not sure we need this assignation since we have the model already prepared

pmbrull (commented):
Thanks for this, left some quick styling comments, but overall lgtm

ingestion/src/metadata/utils/s3_utils.py
```diff
@@ -115,3 +115,11 @@ def read_avro_from_s3(client: Any, key: str, bucket_name: str):
     return read_from_avro(
         get_file_text(client=client, key=key, bucket_name=bucket_name)
     )
+
+
+def prefix_exits(client: Any, bucket_name: str, prefix: str):
```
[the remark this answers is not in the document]
cristiancalugaru: Added type for Client and return type.

ingestion/src/metadata/utils/s3_utils.py
cristiancalugaru: extracted to constant.

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
cristiancalugaru: extracted to constant.

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
cristiancalugaru: yes indeed, used the existing call.

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
```diff
@@ -0,0 +1,287 @@
...
+
+OPENMETADATA_TEMPLATE_FILE_NAME = "openmetadata.json"
+
+
+class S3Metric(Enum):
+    NUMBER_OF_OBJECTS = "NumberOfObjects"
+    BUCKET_SIZE_BYTES = "BucketSizeBytes"
+
+
+class S3BucketResponse(BaseModel):
```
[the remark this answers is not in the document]
cristiancalugaru: Much more readable that way

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
cristiancalugaru: Used get and extra check for None before iterating the list of buckets.

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
cristiancalugaru: Updated naming to snake case for all BaseModels

ingestion/src/metadata/ingestion/source/objectstore/s3/metadata.py
cristiancalugaru: Em, well this was used in 3 places and I just thought **bucket_name** is shorter and clearer instead of using **bucket_reponse.Name** everywhere.

rae89:
@cristiancalugaru If I am understanding this correctly, a user will upload a metadata template file to any s3 that they want to ingest into the openmetadata platform?  Is there a reason this design implementation was chosen?  As as user I find this more cumbersome to upload an additional file to s3 to tag which buckets/paths need to be ingested.  Why not configure the s3 connector to accept the bucket and path to ingest from?

ingestion/src/metadata/ingestion/source/objectstore/s3/connection.py
```diff
@@ -0,0 +1,54 @@
...
+        cloudwatch_client=aws_client.get_client(service_name="cloudwatch"),
+    )
+
+
+def test_connection(client: S3ObjectStoreClient) -> None:
+    """
+    Test connection to both s3 and cloudwatch
+    """
+    try:
+        client.s3_client.list_buckets()
```
pmbrull, line 47: could you do a small change here to handle the new `TestConnectionSteps` structure? https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/dashboard/tableau/connection.py#L88

pmbrull: same logic as you have, just set in two "steps"

ingestion/src/metadata/ingestion/source/objectstore/s3/connection.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def get_connection(connection: S3StoreConnection) -> S3ObjectStoreClient:
+    """
+    Returns 2 clients - the s3 client and the cloudwatch client needed for total nr of objects and total size
+    """
+    aws_client = AWSClient(connection.awsConfig)
+    return S3ObjectStoreClient(
+        s3_client=aws_client.get_client(service_name="s3"),
+        cloudwatch_client=aws_client.get_client(service_name="cloudwatch"),
```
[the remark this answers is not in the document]
cristiancalugaru: Added.

ingestion/src/metadata/ingestion/source/objectstore/s3/connection.py
cristiancalugaru: Updated to the TestConnection approach.

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/objectstore/containerMetadataConfig.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$id": "https://open-metadata.org/schema/metadataIngestion/containerMetadataConfig.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "ObjectStoreContainerConfig",
+  "description": "Definition of the properties contained by an object store container template config file",
```
sureshms, line 5: Should we use the more general term storage service instead of object store?

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/objectstore/containerMetadataConfig.json
```diff
@@ -0,0 +1,60 @@
...
+          "type": "string"
+        },
+        "isStructured": {
+          "title": "Is Structured",
+          "description": "Whether the container defines a supported structure like avro, parquet, csv.",
+          "type": "boolean",
+          "default": false
+        },
+        "structureFormat": {
+          "title": "Schema format",
```
sureshms, line 25: if structureFormat is null, then isStructured is false?

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/objectstoreServiceMetadataPipeline.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$id": "https://open-metadata.org/schema/metadataIngestion/objectStoreServiceMetadataPipeline.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
sureshms, line 3: Again storageService maybe a more generic name.

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/objectstore/containerMetadataConfig.json
pmbrull: I believe that the current `storageService` is the one we want to deprecate. Kind of like the `object` approach to represent blob-like services such as S3 or ADLS. True that the structure will also allow us to eventually ingest "directory" structures, but I appreciate the separation of objects vs any other free storage service

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/objectstore/containerMetadataConfig.json
pmbrull: nice catch. This one we can remove

pmbrull (approved):
Thanks for the changes, LGTM

PR #10524, opened 2023-03-11
ui: explore page filters refactoring

aniketkatkar97:
### Describe your changes :

I worked on:

- [x] Changing the behavior of quick dropdown filters to be the same as left panel filters
  - AND condition across different fields.
  - OR condition inside the same field filters.

https://user-images.githubusercontent.com/51777795/225558150-1f2a9580-8ce5-4087-a9c0-5913e374e048.mov

- [x] Changed `facet filters` updation logic where only the counts of the entities will be updated after filters are applied instead of updating the actual filters list.

https://user-images.githubusercontent.com/51777795/226531346-fc4c997c-d94d-47d8-bf35-315aa5782fb4.mov

- [x] Separated the URL parameter variables to store `Facet filter`, `Quick dropdown filter`, and `Advanced search filter` to be able to determine respectively applied filters from the URL.

https://user-images.githubusercontent.com/51777795/225558689-72726cb1-2260-4588-b0a8-5475d0891b67.mov

- [x] Added logic to retain the applied `Owner` and `Tags` quick dropdown filters after tabs change.

https://user-images.githubusercontent.com/51777795/225559187-6bbf5aaf-4735-4521-936c-893e3bdc9f9c.mov

#
### Type of change :

- [x] Bug fix
- [x] Improvement

#

#
### Checklist:

#
### Reviewers

Frontend: @open-metadata/ui


openmetadata-ui/src/main/resources/ui/src/constants/explore.constants.ts
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2022 Collate.
+ *  Copyright 2023 Collate.
```
chirag-madlani, line 2: There's no need to update this.

chirag-madlani (dismissed):
Looks good to me. @harshach Verify when you have some time

harshach (approved):

harshach:
lets merge this in @aniketkatkar97

aniketkatkar97:
@harshach Looking into a bug I found while testing on this PR. I will push the fix in some time.

chirag-madlani:
@aniketkatkar97 check Maven-CI & [3 Code Smells](https://sonarcloud.io/project/issues?id=open-metadata-ui&pullRequest=10524&resolved=false&types=CODE_SMELL)

openmetadata-ui/src/main/resources/ui/src/components/Explore/Explore.utils.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (!acc[dataCategory.label]) {
+      acc[dataCategory.label] = [];
+    }
+    acc[dataCategory?.label].push({
+      key: value,
+      label: value,
+    });
+
+    return acc;
+  }, {} as Record<string, SearchDropdownOption[]>);
```
chirag-madlani, line 53: I feel you want to just group data from array keys right?
Can we try this instead of this logic? https://lodash.com/docs/4.17.15#groupBy

openmetadata-ui/src/main/resources/ui/src/components/Explore/Explore.utils.ts
```diff
@@ -0,0 +1,89 @@
...
+    queryFilter.query.bool.must
+  ) {
+    const filters: Array<QueryFieldValueInterface> = [];
+
+    queryFilter.query.bool.must.forEach((item) => {
+      const filterValues = item?.bool?.should;
+
+      if (filterValues) {
+        filters.push(...filterValues);
+      }
```
chirag-madlani, line 80: Better to add some dafe guards in the code like: 
```suggestion
    const conditions = get(queryFilter, query.bool.must, [])
    conditions.forEach((item) => {
      const filterValues = item?.bool?.should;

      if (filterValues) {
        filters.push(...filterValues);
      }
```

openmetadata-ui/src/main/resources/ui/src/pages/explore/ExplorePage.component.tsx
```diff
@@ -87,41 +104,97 @@ const ExplorePage: FunctionComponent = () => {
...
+
+    if (
+      noAdvancedSearchQuickFilters ||
+      !has(advancesSearchQuickFilters, 'query.bool.must') ||
+      isUndefined(advancesSearchQuickFilters?.query.bool.must)
+    ) {
+      return undefined;
+    } else {
+      // Getting the filters that can be common for all the Entities
+      const must = advancesSearchQuickFilters?.query.bool.must.filter(
```
chirag-madlani, line 138: same

openmetadata-ui/src/main/resources/ui/src/pages/explore/ExplorePage.component.tsx
```diff
@@ -240,44 +350,77 @@ const ExplorePage: FunctionComponent = () => {
...
+      const newAggregates: Aggregations = {};
+
+      if (
+        !isEmpty(withFilterAggregations) &&
+        !isEmpty(withoutFilterAggregations) &&
+        !isUndefined(withoutFilterAggregations) &&
+        !isUndefined(withFilterAggregations)
+      ) {
+        Object.keys(withoutFilterAggregations).forEach((filterKey) => {
+          const aggregateFieldValue = getUpdatedAggregateFieldValue(
```
chirag-madlani, line 388: Is this a API call? If yes we need to rewrite this in a batter way

aniketkatkar97: No its a utility function.

openmetadata-ui/src/main/resources/ui/src/components/Explore/Explore.utils.ts
aniketkatkar97: I've added this logic because here we're not just grouping the data from a single array key. We are using 2 different arrays here and refactoring the data from them.

openmetadata-ui/src/main/resources/ui/src/components/Explore/Explore.utils.ts
aniketkatkar97: I have added a safeguard here by checking if the object has the intended fields or not using the `has` method of lodash. But the `get` method suggested by you makes more sense here as it will also increase the readability of the code. I'll update the same.

chirag-madlani (approved):

Issue #10468, opened 2023-03-08
Apache Druid database integration does not work


ArtyomyuS:
**Affected module**
Does it impact the UI, backend or Ingestion Framework? - No

**Describe the bug**
While doing a simple integration with everything running locally Open Meta is no able to fetch meta information for the server for Apache Druid.

**To Reproduce**

https://user-images.githubusercontent.com/2368617/223665806-410bb3fc-d832-487f-aa10-ff1c799f3cbd.mov

**Expected behavior**
As in the example the database has a datasource(table) wikipedia, expected to see meta information in the Open Meta 

**Version:**
Running it with docker compose. So all setup associated with 0.13.2 version

**Additional context**
N/A

PR #10553, opened 2023-03-13
fix druid ingestion

NiharDoshi99:
### Describe your changes :

I worked fixing druid ingestion
Fix: #10468 

#
### Type of change :

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation


ingestion/src/metadata/ingestion/source/database/common_db_source.py
```diff
@@ -129,11 +129,16 @@ def get_database_names(self) -> Iterable[str]:
         Sources with multiple databases should overwrite this and
         apply the necessary filters.
         """
-        custom_database_name = self.service_connection.__dict__.get("databaseName")
-
-        database_name = self.service_connection.__dict__.get(
-            "database", custom_database_name or "default"
+        custom_database_name = self.service_connection.__dict__.get(
+            "databaseName", None
```
ulixius9, line 133: why do we need to return `None` in case of not found by default get returns `None` if key was not found

ingestion/src/metadata/ingestion/source/database/common_db_source.py
```diff
@@ -129,11 +129,16 @@ def get_database_names(self) -> Iterable[str]:
...
         )
+
+        database_name = self.service_connection.__dict__.get("database", "default")
```
ulixius9, line 136: then what is the difference between "database" and "databaseName" 🤔 IMO both are serving the same purpose if you want we can rename the `database` field to `databaseName`

ingestion/src/metadata/ingestion/source/database/common_db_source.py
NiharDoshi99: Yes 👍

ingestion/src/metadata/ingestion/source/database/common_db_source.py
```diff
@@ -131,9 +131,12 @@ def get_database_names(self) -> Iterable[str]:
         """
         custom_database_name = self.service_connection.__dict__.get("databaseName")
 
-        database_name = self.service_connection.__dict__.get(
-            "database", custom_database_name or "default"
-        )
+        database_name = self.service_connection.__dict__.get("database")
```
ulixius9, line 134: this will always return None as we have removed it from schema itself

NiharDoshi99: these is in `common_db_source.py` so other database might have database in their jsonSchema

ulixius9: oh okay I thought we are doing it for druid/metadata but yes in case of  `common_db_source` it makes sense but I believe the previous implementation should work then right ?


```
database_name = self.service_connection.__dict__.get("database", custom_database_name or "default")
```

NiharDoshi99: yes

ulixius9 (approved):

PR #10563, opened 2023-03-14
Fix #10489: Handle unknown data types & store raw data type

### LINKED ISSUE #10489: Metadata ingestion for source table RAW datatype vs. OM data type
## Context

We currently map source table data types to OpenMetadata types. This is becoming hard to scale, and issues arise when we don't properly map a source type to OM type. These issues arise on the Profiler, where we use the OM DataType to prepare the ORM objects. If the types do not properly match the source types, our queries can be wrong.

## Profiler changes

In the profiler ORM converter ([link](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/profiler/orm/converter.py#L32)), we can use sqlalchemy dialects to map the RAW types to sqlalchemy types: `RAW -> SQLAlchemy` instead of `RAW -> OM -> SQLAlchemy`.

## Discussion

Let's discuss possible changes here and their PROs / CONs:
1.  Update `table.json#/datatype` to `string` -> This would have many major changes when running the metadata ingestion
2. Add a new `table.json#/rawDataType` as `string` and use that in the ORM converter. We would need to show this one in the UI -> Question, how do we then depracate the `dataType` change?

ulixius9:
### Describe your changes :

Fix #10489: Handle unknown data types & store raw data type

#
### Type of change :

- [x] Improvement

#
### Checklist:

#
### Reviewers

@open-metadata/ingestion


ingestion/src/metadata/ingestion/source/database/athena/metadata.py
```diff
@@ -129,7 +121,8 @@ def get_columns(self, connection, table_name, schema=None, **kw):
             "default": None,
             "autoincrement": False,
             "comment": c.comment,
-            "raw_data_type": c.type if is_complex(c.type) else None,
+            "raw_data_type": c.type,
```
harshach, line 124: lets call this as system_data_type

ingestion/src/metadata/ingestion/source/database/azuresql/metadata.py
```diff
@@ -20,7 +22,43 @@
...
+    get_view_definition,
+)
+from metadata.utils.sqlalchemy_utils import (
+    get_all_table_comments,
+    get_all_view_definitions,
+)
+
+ischema_names.update(
+    {
+        "nvarchar": create_sqlalchemy_type("NVARCHAR"),
```
harshach, line 39: @ulixius9 can we make sure in our tests we create tables with all of these types for each source?

ingestion/src/metadata/ingestion/source/database/athena/metadata.py
harshach: @ulixius9 can you take care of this

ingestion/src/metadata/ingestion/source/database/athena/metadata.py
```diff
@@ -142,16 +136,30 @@ def get_columns(self, connection, table_name, schema=None, **kw):
...
 
+@reflection.cache
+def get_view_definition(self, connection, view_name, schema=None, **kw):
+    """
+    Gets the view definition
+    """
+    full_view_name = view_name if not schema else f"{schema}.{view_name}"
+    res = connection.execute(f"SHOW CREATE VIEW {full_view_name}").fetchall()
+    if res:
+        return "\n".join(i[0] for i in res)
```
harshach, line 156: is this a list of lists?  and are we sure I[0] always available?

ulixius9: yes i[0] is always available

basically if your view definition is like 
```
create view 
as 
select * from table
```

it will return as [(create view ,), (as,),(select * from table,]

OnkarVO7 (approved):

ingestion/src/metadata/ingestion/source/database/athena/metadata.py
ulixius9: done

ingestion/src/metadata/ingestion/source/database/azuresql/metadata.py
ulixius9: many types are already covered, will open a separate pr to cover the remaining types!

OnkarVO7 (approved):

OnkarVO7 (approved):

mohityadav766 (approved):

PR #10603, opened 2023-03-15
Fixes 10351: Fixes Metrics Computation, Samping, test suites and partioning

### LINKED ISSUE #8874: Antlr issue for strings with double quotes
**Affected module**
Ingestion

**Describe the bug**
A string of format `TABLEAU.\"PUBLIC\".SC_0700_HSE_CONSUMPTIONS` when passed to the `fqn.split` method throws an exception
```
line 1:8 no viable alternative at input '"'
  File "/Users/onkarravgan/Desktop/project/newest_proj/OpenMetadata/ingestion/src/metadata/generated/antlr/FqnParser.py", line 206, in name
    raise NoViableAltException(self)
antlr4.error.Errors.NoViableAltException: None

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/Users/onkarravgan/Desktop/project/newest_proj/OpenMetadata/ingestion/src/metadata/utils/fqn.py", line 414, in split_table_name
    details: List[str] = split(table_name)
  File "/Users/onkarravgan/Desktop/project/newest_proj/OpenMetadata/ingestion/src/metadata/utils/fqn.py", line 66, in split
    tree = parser.fqn()
  File "/Users/onkarravgan/Desktop/project/newest_proj/OpenMetadata/ingestion/src/metadata/generated/antlr/FqnParser.py", line 115, in fqn
    self.name()
  File "/Users/onkarravgan/Desktop/project/newest_proj/OpenMetadata/ingestion/src/metadata/generated/antlr/FqnParser.py", line 211, in name
    self._errHandler.recover(self, re)
  File "/Users/onkarravgan/Desktop/project/newest_proj/OpenMetadata/env/lib/python3.8/site-packages/antlr4/error/ErrorStrategy.py", line 697, in recover
    raise ParseCancellationException(e)
antlr4.error.Errors.ParseCancellationException: None
```

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.12.2]

### LINKED ISSUE #10351: Fix datalake profiler/test suite implementation
**Context**
A few elements of the implementations have not been implemented correctly and need to addressed:
- [ ] Metrics are not computed correctly. As the code snippet below shows, we are doing some retrieval based on index which 1) does not compute the actual metric if multiple chunks are present, and 2) does not guarantee an index will match the correct metric. -> #10603 
```python
            row = []
            for metric in metrics:
                for df in dfs:
                    row.append(
                        metric().df_fn(df.astype(object).where(pd.notnull(df), None))
                    )
            if row:
                if isinstance(row, list):
                    row_dict = {}
                    for index, table_metric in enumerate(metrics):
                        row_dict[table_metric.name()] = row[index]
                    return row_dict
                return dict(row)
```

- [ ] Partitionning for DQ was not implemented in the ingestion workflow. While user could set up the partitionning config, in the platform for a datalake entity, this one will never be reflected on the user side
- [ ] Data Quality tests will only be ran against the first chunk. If multiple chunk are present the DQ will never be ran against those beyond the first one -> #10603 
```python
        return DataLakeTestSuiteInterface(
            ometa_client=self.client,
            df=ometa_to_dataframe(
                service_connection_config.configSource,
                get_connection(service_connection_config).client,
                table_entity,
            )[0],
            table_entity=table_entity,
        )
```
- [ ] implement sampling -> #10603 
---
cc @harshach @pmbrull @ayush-shah

ayush-shah:
### Describe your changes :

Fixes #10351
Fixes #8874

Test suite related:
sampling fix, add partitioning

Profiler fix:
metrics, sampling fix
#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
</p>

#
### Checklist:

#


TeddyCr:
@ayush-shah I fixed a few logic for certain metrics. You have a few tests failing in`ingestion/tests/unit/test_suite/test_validations_datalake.py`. Can you go over the failures. 

Can you also add `NaN` values in the CSV file so we have something a bit closer to reality.

ingestion/src/metadata/ingestion/source/database/datalake/metadata.py
```diff
@@ -565,6 +621,25 @@ def get_s3_files(client, key, bucket_name, client_kwargs=None):
...
+                        return DataType.JSON.value
+                    if isinstance(data_frame[column_name].iloc[0], str):
+                        return DataType.STRING.value
+            except Exception as err:
+                logger.warning(
+                    f"Failed to disinguish data type for column {column_name}, Falling back to {data_type}, exc: {err}"
+                )
+                logger.debug(traceback.format_exc())
+        return data_type
+
```
TeddyCr, line 642: Could you add some test around this and add some comment explaining what we are doing?

ingestion/src/metadata/profiler/metrics/static/distinct_count.py
```diff
@@ -43,13 +42,13 @@ def metric_type(self):
...
-        df = cast(DataFrame, df)
-
+    def df_fn(self, dfs=None):
         try:
-            return len(set(df[self.col.name].values.tolist()))
+            import pandas as pd  # pylint: disable=import-outside-toplevel
+
+            return len(
+                pd.concat((df[self.col.name] for df in dfs)).dropna().drop_duplicates()
+            )
```
TeddyCr, line 51: A few things here:

1. let's handle memory error
```python
try:
    df = pd.concat([df[self.col.name] for df in dfs])
except MemoryError:
    ...
```
3. let's use `len(df["self.col.name"].value_counts().index)` -> this will handle `nan` and will be probably more efficient than `dropna` and then `dropduplicates` https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.value_counts.html

ingestion/src/metadata/profiler/metrics/static/max_length.py
```diff
@@ -58,19 +57,26 @@ def fn(self):
...
+                map(
+                    lambda df: length_vectorize_func(
+                        df.astype(object).where(
+                            notnull(df),
+                            "",
+                        )[self.col.name]
+                    ).max(),
+                    dfs,
+                )
+            )
```
TeddyCr, line 78: I am assuming we perform `df.astype(object)` as we'll fall back to a string type correct?

if so we pobably need to do something like:
```python
max(length_vectorize_func(df["a"].fillna("").astype(str)).max() for df in df)
```

ingestion/src/metadata/profiler/metrics/static/mean.py
```diff
@@ -74,21 +74,35 @@ def fn(self):
...
+                if not pd.isnull(mean):
+                    means.append(mean)
+                    weights.append(df[self.col.name].count())
 
         if is_concatenable(self.col.type):
-            length_vector_fn = vectorize(len)
-            return length_vector_fn(df[self.col.name]).mean()
+            length_vectorize_func = vectorize(len)
+            for df in dfs:
+                mean = length_vectorize_func(df[self.col.name].dropna()).mean()
```
TeddyCr, line 97: should we add `.astype(str)` here as a safe guard?

ingestion/src/metadata/profiler/metrics/static/mean.py
```diff
@@ -74,21 +74,35 @@ def fn(self):
...
+                if not pd.isnull(mean):
+                    means.append(mean)
+                    weights.append(df[self.col.name].count())
```
TeddyCr, line 100:
```suggestion
                    weights.append(df[self.col.name].dropna().count())
```

ingestion/src/metadata/profiler/metrics/static/min_length.py
TeddyCr, line 1: The same logic than `max_length` would apply here I believe. https://github.com/open-metadata/OpenMetadata/pull/10603/files#r1157063050

ingestion/src/metadata/profiler/metrics/static/null_count.py
```diff
@@ -50,9 +49,6 @@ def fn(self):
...
         return SumFn(case([(column(self.col.name).is_(None), 1)], else_=0))
 
-    def df_fn(self, df=None):
+    def df_fn(self, dfs=None):
         """pandas function"""
-        from pandas import DataFrame  # pylint: disable=import-outside-toplevel
-
-        df = cast(DataFrame, df)
-        return df[self.col.name].isnull().sum()
+        return sum(map(lambda df: df[self.col.name].isnull().sum(), dfs))
```
TeddyCr, line 54: Generally `map` in Python has little use case. Most common is when you have already defined function
e.g.
```python
map(sum, [1,2,3,4])
```
But here we end up doing something a bit more convulated. It is more pythonic to do something like

```python
return sum(df[self.col.name].isnull().sum() for df in dfs)
```

ingestion/src/metadata/profiler/metrics/static/regexp_match_count.py
TeddyCr, line 1: Same comment than in https://github.com/open-metadata/OpenMetadata/pull/10603/files#r1157076331

ingestion/src/metadata/profiler/metrics/static/stddev.py
```diff
@@ -90,17 +90,13 @@ def fn(self):
...
         import pandas as pd  # pylint: disable=import-outside-toplevel
 
-        df = cast(pd.DataFrame, df)
-
         if is_quantifiable(self.col.type):
-            stddev = df[self.col.name].std()
-            if pd.isnull(stddev):
-                return None
-            return stddev
+            return pd.concat(df[self.col.name] for df in dfs).std()
```
TeddyCr, line 98: let's handle potential memory error

ingestion/src/metadata/profiler/metrics/static/unique_count.py
```diff
@@ -65,16 +65,12 @@ def query(
...
         """
         Build the Unique Count metric
         """
-        from pandas import DataFrame  # pylint: disable=import-outside-toplevel
-
-        df = cast(DataFrame, df)
-
         try:
-            return df[self.col.name].nunique()
+            return sum(len(df[self.col.name].dropna().unique()) for df in dfs)
```
TeddyCr, line 73: Here we need to compute the count of values that are unique (occure only once) across all the dataframes.

we'll need 2 things:
1. keep a tally of value count
2. add to the tally count when we visit each dataframe

we'll need to do something like that (please check the implementation in details, don't just copy/past)

```python
from collections impot Counter

c = Counter()

for df in dfs:
    c.update(df[self.col.name].value_counts().to_dict(()

unique_count = len([k for k,v in c.items() if c == 1])
...
```

ingestion/src/metadata/profiler/profiler/datalake_sampler.py
```diff
@@ -50,53 +47,28 @@ def __init__(
...
-            else None,
-            frac=self.profile_sample
-            if self.profile_sample_type == ProfileSampleType.PERCENTAGE
-            else None,
-            random_state=random.randint(0, 100),
-            replace=True,
-        )
         return (
-            sampled_data_frame.astype(object)
+            data_frame.astype(object)
```
TeddyCr, line 51: why do we convert it to `object` here?

ingestion/src/metadata/profiler/profiler/datalake_sampler.py
```diff
@@ -50,53 +47,28 @@ def __init__(
...
             .where(
-                notnull(sampled_data_frame),
+                notnull(data_frame),
                 None,
             )
-            .values.tolist()
+            .values.tolist()[:100]
```
TeddyCr, line 56: shouldn't this be `self.sample_limit` vs 100?

ingestion/src/metadata/profiler/profiler/datalake_sampler.py
```diff
@@ -50,53 +47,28 @@ def __init__(
...
-                self._fetch_rows(data_frame),
-            )
-        chunk_limit = math.ceil(self.profile_sample / CHUNKSIZE)
-        cols = data_frame[0].columns.tolist()
         rows = []
-        for index, chunk in enumerate(data_frame):
-            if index >= chunk_limit:
+        cols = data_frame[0].columns.tolist()
+        for chunk in data_frame:
+            rows.extend(self._fetch_rows(chunk)[: 100 - len(rows)])
```
TeddyCr, line 67: same here, shouldn't it be `self.sample_limit` vs 100

ingestion/src/metadata/profiler/profiler/interface/pandas/pandas_profiler_interface.py
```diff
@@ -118,10 +126,14 @@ def _(
...
-                for df in dfs:
-                    row.append(
-                        metric().df_fn(df.astype(object).where(pd.notnull(df), None))
+                row.append(
+                    metric().df_fn(
+                        [
+                            df.astype(object).where(pd.notnull(df), None)
+                            for df in self.dfs
+                        ]
                     )
```
TeddyCr, line 135: why are we converting all the dataframe to objects?

ingestion/src/metadata/profiler/profiler/interface/pandas/pandas_profiler_interface.py
```diff
@@ -161,12 +172,8 @@ def _(
...
             for metric in metrics:
-                for df in dfs:
-                    row.append(
-                        metric(column).df_fn(
-                            df.astype(object).where(pd.notnull(df), None)
-                        )
-                    )
+                metric_resp = metric(column).df_fn(self.dfs)
+                row.append(None if pd.isnull(metric_resp) else metric_resp)
             row_dict = {}
```
TeddyCr, line 177: I think you should populate that dictionnary in your loop above (vs working with index). We should probably rename the variable to `metrics`

ingestion/src/metadata/profiler/profiler/interface/pandas/pandas_profiler_interface.py
```diff
@@ -118,10 +126,14 @@ def _(
...
+                    metric().df_fn(
+                        [
+                            df.astype(object).where(pd.notnull(df), None)
+                            for df in self.dfs
+                        ]
                     )
+                )
             if row:
                 if isinstance(row, list):
                     row_dict = {}
```
TeddyCr, line 139: same comment than https://github.com/open-metadata/OpenMetadata/pull/10603/files#r1157196386

ingestion/src/metadata/profiler/profiler/interface/profiler_protocol.py
```diff
@@ -128,17 +128,13 @@ def get_profile_sample_config(
...
-                    profile_sample=entity.tableProfilerConfig.profileSample,
-                    profile_sample_type=entity.tableProfilerConfig.profileSampleType,
-                )
+        if hasattr(entity.tableProfilerConfig, "profileSample") and hasattr(
+            entity.tableProfilerConfig, "profileSampleType"
+        ):
+            return ProfileSampleConfig(
+                profile_sample=entity.tableProfilerConfig.profileSample,
+                profile_sample_type=entity.tableProfilerConfig.profileSampleType,
+            )
```
TeddyCr, line 137: can we keep the `try/except/else` block, I find it more readable. It also has logic to check if `entity.tableProfilerConfig.profileSample` is None

ingestion/src/metadata/test_suite/api/workflow.py
```diff
@@ -286,12 +287,14 @@ def _create_runner_interface(self, entity_fqn: str):
         return DataLakeTestSuiteInterface(
             service_connection_config=service_connection_config,
             ometa_client=self.client,
-            df=ometa_to_dataframe(
-                service_connection_config.configSource,
+            dfs=return_ometa_dataframes(
```
TeddyCr, line 290: Any advantage of having that function `return_ometa_dataframes`? i see it only return `ometa_to_dataframe` function and adds no additional logic.

ingestion/src/metadata/test_suite/validations/table/pandas/tableCustomSQLQuery.py
```diff
@@ -28,4 +28,4 @@ class TableCustomSQLQueryValidator(
 
     def _run_results(self, sql_expression: str):
         """compute result of the test case"""
-        return self.runner.query(sql_expression)
+        return self.runner[0].query(sql_expression)
```
TeddyCr, line 31: You should run the `sql_expression` against every chunk and aggregate the rows return by the expression and return those rows.

ingestion/src/metadata/utils/pandas_utils.py
```diff
@@ -0,0 +1,26 @@
...
+def return_ometa_dataframes(
+    service_connection_config, client, table, profile_sample_config
+):
+
+    return ometa_to_dataframe(
+        config_source=service_connection_config.configSource,
+        client=client,
+        table=table,
+        profile_sample_config=profile_sample_config,
+    )
```
TeddyCr, line 26: I think this can be removed, we don't do anything more in this function and just call `ometa_to_dataframe` directly

TeddyCr (changes requested):
Added a few comments

ingestion/src/metadata/profiler/metrics/static/stddev.py
ayush-shah: will add that, thanks

ingestion/src/metadata/profiler/metrics/static/null_count.py
ayush-shah: yes, I was thinking of a faster memory efficient approach, but yes, generators work as well.

ingestion/src/metadata/profiler/metrics/static/column_names.py
```diff
@@ -84,9 +84,9 @@ def fn(self):
...
         return ColunNameFn(literal(col_names, type_=sqlalchemy.types.String))
 
-    def df_fn(self, df=None):
+    def df_fn(self, dfs=None):
         from pandas import DataFrame  # pylint: disable=import-outside-toplevel
 
-        df = cast(DataFrame, df)
+        df = cast(DataFrame, dfs[0])
 
         return df.columns.values.tolist()
```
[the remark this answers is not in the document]
ayush-shah: I tried the same, it's just DataFrame and not list of dataframes.

ingestion/src/metadata/interfaces/datalake/datalake_test_suite_interface.py
```diff
@@ -44,12 +44,17 @@ def __init__(
         ometa_client: OpenMetadata = None,
         service_connection_config: DatalakeConnection = None,
         table_entity=None,
-        df=None,
+        dfs=None,
```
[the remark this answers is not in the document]
ayush-shah: I will remove the default fallback.

ingestion/src/metadata/ingestion/source/database/datalake/metadata.py
```diff
@@ -99,11 +103,63 @@ def ometa_to_dataframe(config_source, client, table):
...
+                ]
+        else:
+            # randomize the samples
+            data = [
+                df.sample(
+                    frac=1,
+                    random_state=random.randint(0, 100),
+                )
+                for df in data
+            ]
```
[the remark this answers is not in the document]
ayush-shah: I have moved this to pandas mixin, so it's available for both, profielr and test suites

ingestion/src/metadata/ingestion/source/database/datalake/metadata.py
ayush-shah: yes, let me add that

ingestion/src/metadata/profiler/metrics/static/distinct_count.py
ayush-shah: added memory error.
and modified it to use value_counts

ingestion/src/metadata/profiler/metrics/static/max_length.py
ayush-shah: we were using astype object before to remove the nan using `where(notnull(df), None)`

ingestion/src/metadata/profiler/metrics/static/mean.py
ayush-shah: yes, that would resolve some edge cases i was facing

ingestion/src/metadata/profiler/metrics/static/min_length.py
ayush-shah: made it similar, thanks

ingestion/src/metadata/profiler/metrics/static/not_regexp_match_count.py
```diff
@@ -55,13 +55,14 @@ def fn(self):
...
         """pandas function"""
         if not hasattr(self, "expression"):
             raise AttributeError(
                 "Regex Count requires an expression to be set: add_props(expression=...)(Metrics.REGEX_COUNT)"
             )
 
-        return df[self.col.name][
-            df[self.col.name].str.contains(self.expression)
-        ].count()
+        return sum(
```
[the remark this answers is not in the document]
ayush-shah: not sure if i understand, what check?

ingestion/src/metadata/profiler/metrics/static/unique_count.py
ayush-shah: I think the unique count needs to be calculated with the whole and not the chunks, correct me if i am wrong..because if we are checking with chunks, we might miss some scenarios where the data might be similar in different chunks

ingestion/src/metadata/profiler/profiler/datalake_sampler.py
ayush-shah: we were modifying it to object as it was a basic fallback to Nan, modified it.

ingestion/src/metadata/profiler/profiler/datalake_sampler.py
ayush-shah: I have modified the logic, and made it more readable

ingestion/src/metadata/profiler/profiler/datalake_sampler.py
ayush-shah: same, I have modified the logic, and made it more readable, added a comment as well

ingestion/src/metadata/profiler/profiler/interface/pandas/pandas_profiler_interface.py
ayush-shah: same as here: https://github.com/open-metadata/OpenMetadata/pull/10603#discussion_r1158047450
modified it

ingestion/src/metadata/test_suite/api/workflow.py
ayush-shah: its's for the tests to mock, i had to add a method that would return mocked df, i have modified it to add sampling in it as well

ingestion/src/metadata/test_suite/validations/table/pandas/tableCustomSQLQuery.py
ayush-shah: yep, fixed it..

I was running tests on parquet, it brought me whole list of dataframe
recently found out that parquet is not being sent as chunks, will fix it in another PR

ingestion/src/metadata/test_suite/validations/table/pandas/tableRowInsertedCountToBeBetween.py
```diff
@@ -72,4 +72,4 @@ def _run_results(self, column_name: str, range_type: str, range_interval: int):
         """
         threshold_date = self._get_threshold_date(range_type, range_interval)
 
-        return len(self.runner.query(f"{column_name} >= {threshold_date}"))
+        return len(self.runner[0].query(f"{column_name} >= {threshold_date}"))
```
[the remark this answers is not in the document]
ayush-shah: yep, making that change, makes sense..

ingestion/src/metadata/utils/pandas_utils.py
ayush-shah: added sampling to this method, we are using it for mocking and returning list[df] for tests

ingestion/src/metadata/profiler/profiler/interface/profiler_protocol.py
ayush-shah: resolved

ingestion/src/metadata/profiler/metrics/static/column_names.py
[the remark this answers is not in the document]
TeddyCr: 🤔 aren't you passing a list of dataframes to the metrics?

ingestion/src/metadata/profiler/metrics/static/not_regexp_match_count.py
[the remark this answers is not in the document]
TeddyCr: Here we have `.str.contains` we are expecting a "string" type column to run this

ingestion/src/metadata/utils/pandas_utils.py
TeddyCr: Same comment than here, https://github.com/open-metadata/OpenMetadata/pull/10603#discussion_r1158587065

ingestion/src/metadata/profiler/metrics/static/column_names.py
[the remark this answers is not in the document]
ayush-shah: no, it's just dataframe..
we just need column names and we will be able to fetch it from the first df

ingestion/src/metadata/test_suite/api/workflow.py
ayush-shah: it now accomodates sampling as well..it will call the ometa method, run sampling logic and return teh sampled dataframe

ingestion/src/metadata/data_quality/validations/table/pandas/tableCustomSQLQuery.py
```diff
@@ -28,4 +31,13 @@ class TableCustomSQLQueryValidator(
...
+        import pandas as pd  # pylint: disable=import-outside-toplevel
+
+        try:
+            return pd.concat(runner.query(sql_expression) for runner in self.runner)
+        except MemoryError:
+            logger.error(
+                "Unable to compute due to memory constraints."
+                "We recommend using a smaller sample size or partitionning for the query."
+            )
+            return []
```
TeddyCr, line 43: You don't need to concatenante it here. https://github.com/open-metadata/OpenMetadata/pull/10603#discussion_r1157217665

ingestion/src/metadata/data_quality/validations/table/pandas/tableRowInsertedCountToBeBetween.py
```diff
@@ -70,6 +73,19 @@ def _run_results(self, column_name: str, range_type: str, range_interval: int):
...
+                    runner.query(f"{column_name} >= {threshold_date}")
+                    for runner in self.runner
+                )
+            )
+        except MemoryError:
+            logger.error(
+                "Unable to compute due to memory constraints."
+                "We recommend using a smaller sample size or partitionning for the query."
+            )
+            return 0
```
TeddyCr, line 91: Same thing, don't need to aggregate it https://github.com/open-metadata/OpenMetadata/pull/10603#discussion_r1157219495

ingestion/src/metadata/mixins/pandas/pandas_mixin.py
```diff
@@ -76,3 +79,56 @@ def get_partitioned_df(self, dfs):
...
+            else:
+                # randomize the samples
+                return [
+                    df.sample(
+                        frac=1,
+                        random_state=random.randint(0, 100),
+                    )
+                    for df in data
+                ]
+            return data
```
TeddyCr, line 134: Shouldn't we just return the data here if there no sampling config?

ingestion/src/metadata/profiler/interface/pandas/pandas_profiler_interface.py
```diff
@@ -124,25 +118,12 @@ def _(
...
             for metric in metrics:
-                row.append(
-                    metric().df_fn(
-                        [
-                            df.astype(object).where(pd.notnull(df), None)
-                            for df in self.dfs
-                        ]
-                    )
+                row_dict[metric.name()] = metric().df_fn(
+                    [df.where(pd.notnull(df), None) for df in self.dfs]
```
TeddyCr, line 124: Shouldn't we do that once before looping? if we have `n` metrics we'll be doing the same operation n times from what I understand.

ingestion/src/metadata/profiler/interface/profiler_protocol.py
```diff
@@ -128,21 +128,19 @@ def get_profile_sample_config(
...
-
-        return None
+        try:
+            return ProfileSampleConfig(
+                profile_sample=entity.tableProfilerConfig.profileSample,
+                profile_sample_type=entity.tableProfilerConfig.profileSampleType,
+            )
+        except AttributeError:
+            # Exception Triggered if profileSample attribute not available in tableProfilerConfig
+            return None
```
TeddyCr, line 143: what is that? I believed we had it this way in `main`

```python
    @staticmethod
    def get_profile_sample_config(
        entity: Table,
        entity_config: Optional[TableConfig],
        source_config: DatabaseServiceProfilerPipeline,
    ) -> Optional[ProfileSampleConfig]:
        """_summary_

        Args:
            entity (Table): table entity object
            entity_config (Optional[TableConfig]): table config object from yaml/json file
        Returns:
            Optional[dict]: dict
        """
        if entity_config:
            return ProfileSampleConfig(
                profile_sample=entity_config.profileSample,
                profile_sample_type=entity_config.profileSampleType,
            )

        try:
            profile_sample = entity.tableProfilerConfig.profileSample
        except AttributeError:
            pass
        else:
            if profile_sample:
                return ProfileSampleConfig(
                    profile_sample=entity.tableProfilerConfig.profileSample,
                    profile_sample_type=entity.tableProfilerConfig.profileSampleType,
                )

        if source_config.profileSample:
            return ProfileSampleConfig(
                profile_sample=source_config.profileSample,
                profile_sample_type=source_config.profileSampleType,
            )

        return None
```

ingestion/src/metadata/profiler/metrics/static/distinct_count.py
```diff
@@ -43,15 +43,16 @@ def fn(self):
...
+        from collections import Counter  # pylint: disable=import-outside-toplevel
 
-            return len(
-                pd.concat((df[self.col.name] for df in dfs)).dropna().drop_duplicates()
-            )
+        try:
+            counter = Counter()
+            for df in dfs:
+                counter.update(df[self.col.name].dropna().to_list())
+            return len(counter.keys())
```
TeddyCr: From what I am seeing `.value_counts().to_dict()` is about 20% faster (https://github.com/open-metadata/OpenMetadata/pull/10603#discussion_r1157101575)

ingestion/src/metadata/profiler/metrics/static/stddev.py
```diff
@@ -95,7 +95,14 @@ def df_fn(self, dfs=None):
...
         if is_quantifiable(self.col.type):
-            return pd.concat(df[self.col.name] for df in dfs).std()
+            try:
+                return pd.concat(df[self.col.name] for df in dfs).std()
+            except MemoryError:
+                logger.error(
+                    f"Unable to compute distinctCount for {self.col.name} due to memory constraints."
+                    f"We recommend using a smaller sample size or partitionning."
+                )
+                return 0
```
TeddyCr, line 105: let's return None otherwise we will be setting the median value to 0 when we did not compute it at all.

ingestion/src/metadata/profiler/processor/datalake_sampler.py
```diff
@@ -63,8 +54,9 @@ def get_col_row(self, data_frame):
         cols = []
         rows = []
         cols = data_frame[0].columns.tolist()
+        # Sample Data should not exceed sample limit
         for chunk in data_frame:
-            rows.extend(self._fetch_rows(chunk)[: 100 - len(rows)])
+            rows.extend(self._fetch_rows(chunk)[: self.sample_limit])
```
TeddyCr, line 59: Don't you need to keep the `- len(row)` here so you know how many rows you need to add?

ingestion/tests/unit/test_suite/test_validations_datalake.py
```diff
@@ -183,7 +183,7 @@
             "test_case_column_values_to_be_unique",
             "columnValuesToBeUnique",
             "COLUMN",
-            (TestCaseResult, "20", "2", TestCaseStatus.Failed),
+            (TestCaseResult, "20", "0", TestCaseStatus.Failed),
```
TeddyCr, line 186: Should this be 1? This nickname `NickName` should be unique right?

ingestion/src/metadata/profiler/processor/datalake_sampler.py
ayush-shah: as the chunk in itself is 200000 big, we won't need to do that..

ingestion/tests/unit/test_suite/test_validations_datalake.py
ayush-shah: I went through the test case data..
the unique value were 0 because
it was repeating..

ingestion/src/metadata/profiler/metrics/static/stddev.py
ayush-shah: resolved

ingestion/src/metadata/profiler/metrics/static/distinct_count.py
```diff
@@ -43,15 +43,16 @@ def fn(self):
...
         except Exception as err:
             logger.debug(
-                f"Don't know how to process type {self.col.type} "
-                f"when computing Distinct Count.\n Error: {err}"
+                f"Don't know how to process type {self.col.type}"
+                f"when computing Unique Count.\n Error: {err}"
```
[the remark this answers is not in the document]
ayush-shah: resolved

ingestion/src/metadata/data_quality/validations/table/pandas/tableCustomSQLQuery.py
ayush-shah: the query would return dataframe with or without rows here based on the query

ayush-shah: if the dataframes are empty, we concat, and the len of that dataframe being empty will result in test success

TeddyCr: Imagine a query that returns all the data. That will throw an error. We just need to sum the Len and not concatenate. Does it make sense?

ayush-shah: no, i tested this out, it will return dataframes with data, if the dataframe has data, we concat, and return the same, it will check the len of the rows.

but agree on the sum of length, that would reduce load on the memory, instead of concat, let me return the sum of len.

ingestion/src/metadata/data_quality/validations/table/pandas/tableRowInsertedCountToBeBetween.py
ayush-shah: This will work as expected, given that concat can come with memory error, added that exception as well..
to improve performance, let me sum the length..

ingestion/src/metadata/mixins/pandas/pandas_mixin.py
ayush-shah: not sure if i remember correctly, but randomizing the sample was something that we discussed, if i am not wrong.

ingestion/src/metadata/profiler/interface/profiler_protocol.py
ayush-shah: ohh, if it's already taken care of, let me revert that back

ingestion/src/metadata/profiler/metrics/static/unique_count.py
```diff
@@ -69,11 +69,16 @@ def df_fn(self, dfs=None):
...
         Build the Unique Count metric
         """
+        from collections import Counter  # pylint: disable=import-outside-toplevel
+
         try:
-            return sum(len(df[self.col.name].dropna().unique()) for df in dfs)
+            counter = Counter()
+            for df in dfs:
+                counter.update(df[self.col.name].dropna().to_list())
+            return len([i for i in counter if counter[i] == 1])
```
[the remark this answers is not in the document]
ayush-shah: the current approach works as expected..
to make it more readable, let me add your suggested changes.

ingestion/src/metadata/mixins/pandas/pandas_mixin.py
```diff
@@ -80,20 +80,23 @@ def get_partitioned_df(self, dfs):
...
         data = ometa_to_dataframe(
             config_source=service_connection_config.configSource,
             client=client,
             table=table,
         )
+        if isinstance(data, DataFrame):
+            data: List[DataFrame] = [data]
 
-        if data:
+        if data and isinstance(data, list):
```
TeddyCr, line 99: why would `data` not be a list? Shouldn't `ometa_to_dataframe` always return a list of dataframes?

ayush-shah: Yes, this is just a safeguard statement (maybe an overkill for a safeguard) so that things don't fail.

TeddyCr: 🤔 I think here we would not be handling things properly right and make it a bit harder to debug. If we don't have a list of dataframes, if we want to keep that `if` logic, then let's raise an error and not return an empty list.

ayush-shah: we are handling the return of list of dataframes right for now. We have a ticket open to improve the logic for all the connections to be taken care of in 1.1.
let me add a warning for dataframe names which are not returned as list if there is an edge case by any chance and raise the error.

ingestion/src/metadata/mixins/pandas/pandas_mixin.py
```diff
@@ -80,50 +81,57 @@ def get_partitioned_df(self, dfs):
...
+        try:
+            data = ometa_to_dataframe(
+                config_source=service_connection_config.configSource,
+                client=client,
+                table=table,
+            )
+            if isinstance(data, DataFrame):
+                data: List[DataFrame] = [data]
+            if not isinstance(data, list):
+                raise TypeError(f"Couldn't fetch {table.name.__root__}")
```
TeddyCr, line 101: why is this in the if statement and not?

```python
if isinstance(<>):
    ...
if data and isinstance(<>):
    ...
raise TypeError(f"Couldn't fetch {table.name.__root__}")
```

You also have add `# pylint: disable=inconsistent-return-statements` that must tip you of that something is not concistent with the logic right?

ayush-shah: 1st if: it's not a list of dataframes but just a dataframe, we convert it into list of dataframes just in case.
2nd if: if the data is not an empty list and check if it's of type list

inconsistent-return-statements is because we have not added a return to the exception

      except TypeError as err:
                  logger.error(err)

TeddyCr: That make sense. A few things:
1. My orginal comment simply asked to replace `return []` by `raise Something`. There is no point returning an empty list. This just makes things harder to debug. At this point if, for some reasons, we don't have a list of DataFrame then let's raise the error
2. I am not what the try/except is trying to catch here. From what I understand. If there is a failure here, it will be caught at the workflow level and raise there -- wouldn't it? Our try except here, would keep the program execution going until it fails on the next step because the variable calling this function would be `None` when we would expect it to be `list[df]` making the program slightly more difficult to debug.

Does it make sense?

ayush-shah: that makes it more clear.

TeddyCr (approved):
LGTM

Issue #10584, opened 2023-03-14
Add support to Data Model entity for Dashboards


nahuelverdugo:
**Is your feature request related to a problem? Please describe.**
We need support to get the Data Models on what the Dashboards are based into a separate Entity. For this first iteration, we are going to implement it for Tableau.

For Tableau, we plan to get each sheet object model and display it as an entity of the dashboards. We will use this new entity to get the lineage between database services databases and dashboards.

**Tasks**
- [x] Create JSON schema, repository, and endpoint for the data model entity. @nahuelverdugo  https://github.com/open-metadata/OpenMetadata/pull/10636
- [x] Display Data Model Entity in UI. @Ashish8689 
- [x] Add DataModel Filter under Dashboard Entity in explore page in UI. @aniketkatkar97 
- [x] Ingest Tableau Data Model for dashboards. https://github.com/open-metadata/OpenMetadata/pull/10872
- [x] Ingest Lineage for Tableau dashboards considering data model entity. https://github.com/open-metadata/OpenMetadata/pull/10923
- [x] Update ES indexes @nahuelverdugo https://github.com/open-metadata/OpenMetadata/pull/10851
- [x] The Metadata Ingestion Pipeline should support the following:
    - [x] Toggle to ingest or not the data models
    - [x] Data model filter pattern
- [x] Add support for column level lineage between dashboadDataModel <> Table. This is required in the backend  https://github.com/open-metadata/OpenMetadata/pull/11489

pmbrull:
We'll do the col-level lineage in 1.1, as it requires some changes in the backend

PR #10636, opened 2023-03-16
Fix#10584: Add Data Model as an entity

nahuelverdugo:
Part of #10584 

### Describe your changes :
Add models and endpoints for Data Model as an entity.

**Pending:**
- Example data
- Improve testing
- Add/Remove data model to a dashboard

### Type of change :
- [x] New feature

### Checklist:

### Reviewers


Daandamhuis:
@nahuelverdugo: Could this also be implemented for Power BI Data Models?

Daandamhuis:
Done :) 

https://github.com/open-metadata/OpenMetadata/issues/10689

openmetadata-spec/src/main/resources/json/schema/api/data/createDataModel.json
```diff
@@ -0,0 +1,61 @@
+{
+  "$id": "https://open-metadata.org/schema/api/data/createDataModel.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "CreateDataModelRequest",
+  "description": "Create Data Model entity request",
```
mohityadav766, line 5: nit: "." missing

mohityadav766: please check other places as well, missing in a couple of places

openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java
```diff
@@ -486,4 +487,21 @@ public static void sortByTagHierarchy(List<Tag> tags) {
     // Sort tags by tag hierarchy. Tags with parents null come first, followed by tags with
     tags.sort(Comparator.comparing(Tag::getFullyQualifiedName));
   }
+
+  /**
+   * This method is used to populate the entity with all details of EntityReference Users/Tools can send minimum details
+   * required to set relationship as id, type are the only required fields in entity reference, whereas we need to send
+   * fully populated object such that ElasticSearch index has all the details.
+   */
+  public static List<EntityReference> getNonDeletedEntityReferences(List<EntityReference> entities) throws IOException {
```
mohityadav766, line 496: can we just update the existing method 
  public static List<EntityReference> getEntityReferences(List<EntityRelationshipRecord> list) throws IOException {
and accept one more param "Include"

nahuelverdugo: About the second comment, I have just extracted this method from `DashboardRepositoy` that was called from the `prepare` method:

```java
private List<EntityReference> getCharts(List<EntityReference> charts) throws IOException {
    if (nullOrEmpty(charts)) {
      return Collections.emptyList();
    }
    List<EntityReference> chartRefs = new ArrayList<>();
    for (EntityReference chart : charts) {
      EntityReference chartRef = Entity.getEntityReference(chart, Include.NON_DELETED);
      chartRefs.add(chartRef);
    }
    return chartRefs.isEmpty() ? null : chartRefs;
  }
```

So what we do here is validate these entities exist in DB.

openmetadata-spec/src/main/resources/json/schema/api/data/createDataModel.json
```diff
@@ -0,0 +1,61 @@
...
+      "description": "Owner of this data model.",
+      "$ref": "../../type/entityReference.json"
+    },
+    "service": {
+      "description": "Link to the data model service where this data model is hosted in.",
+      "$ref": "../../type/basic.json#/definitions/fullyQualifiedEntityName"
+    },
+    "serviceType": {
+      "description": "Service type where this data model is hosted in.",
+      "$ref": "../../entity/services/dashboardService.json#/definitions/dashboardServiceType"
```
sureshms, line 40: Are the data models coming from regular databases or do dashboard services create there own datasets to power the dashboard?

openmetadata-spec/src/main/resources/json/schema/entity/data/dataModel.json
```diff
@@ -0,0 +1,130 @@
+{
+  "$id": "https://open-metadata.org/schema/entity/data/dataModel.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "DataModel",
+  "$comment": "@om-entity-type",
+  "description": "Data Model entity definition. Data models are the schemas used to build dashboards, charts, or other data models.",
+  "type": "object",
+  "javaType": "org.openmetadata.schema.entity.data.DataModel",
+  "javaInterfaces": ["org.openmetadata.schema.EntityInterface"],
```
sureshms, line 9: There are multiple DataModels today. Should we call this type as DashboardDataModel? This will avoid conflict of multiple java classes named as DataModel.

openmetadata-spec/src/main/resources/json/schema/entity/data/dataModel.json
```diff
@@ -0,0 +1,130 @@
+{
+  "$id": "https://open-metadata.org/schema/entity/data/dataModel.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
sureshms, line 3: This type can be called as dashboardDataModel.json

openmetadata-spec/src/main/resources/json/schema/entity/data/dataModel.json
```diff
@@ -0,0 +1,130 @@
...
+      },
+      "default": null
+    },
+    "dataModels": {
+      "description": "List of data models used by this data model.",
+      "$ref": "../../type/entityReference.json#/definitions/entityReferenceList"
+    }
+  },
+  "required": [
+    "modelType"
```
sureshms, line 127: name is also required, right?

openmetadata-service/src/main/java/org/openmetadata/service/resources/datamodels/DataModelResource.java
```diff
@@ -0,0 +1,443 @@
...
+import org.openmetadata.service.resources.Collection;
+import org.openmetadata.service.resources.EntityResource;
+import org.openmetadata.service.resources.databases.DatabaseUtil;
+import org.openmetadata.service.security.Authorizer;
+import org.openmetadata.service.util.EntityUtil;
+import org.openmetadata.service.util.RestUtil;
+import org.openmetadata.service.util.ResultList;
+
+@Path("/v1/datamodels")
+@Api(value = "Data Model data asset collection", tags = "Data Model data asset collection")
```
sureshms, line 65: If this is specific to the dashboard, we should consider `v1/dashboards/datamodels` as the path. Btw, would charts also have `dataModel` references?

openmetadata-service/src/main/java/org/openmetadata/service/resources/datamodels/DataModelResource.java
```diff
@@ -0,0 +1,443 @@
...
+    DataModelList() {
+      // Empty constructor needed for deserialization
+    }
+  }
+
+  @GET
+  @Operation(
+      operationId = "listDataModels",
+      summary = "List dataModels",
+      tags = "dataModels",
```
sureshms, line 97: Instead of a separate API section for dataModels, setting tags to `dashboards` will group this under dashboards section in the API.

openmetadata-service/src/main/java/org/openmetadata/service/resources/datamodels/DataModelResource.java
```diff
@@ -0,0 +1,443 @@
...
+      @Context UriInfo uriInfo,
+      @Context SecurityContext securityContext,
+      @Parameter(description = "Id of the data model", schema = @Schema(type = "UUID")) @PathParam("id") UUID id,
+      @Parameter(description = "Id of the user to be added as follower", schema = @Schema(type = "UUID")) UUID userId)
+      throws IOException {
+    return dao.addFollower(securityContext.getUserPrincipal().getName(), id, userId).toResponse();
+  }
+
+  @DELETE
+  @Path("/{id}/followers/{userId}")
```
sureshms, line 348: Are we going to allow separate followers for dataModel or would we just allow following dashboards? cc @harshach

openmetadata-spec/src/main/resources/json/schema/api/data/createDataModel.json
harshach: they are coming from databases. Its more like a collection, when they create a view it is part of DB https://powerbi.microsoft.com/en-us/what-is-data-modeling/

openmetadata-service/src/main/java/org/openmetadata/service/resources/datamodels/DataModelResource.java
```diff
@@ -0,0 +1,443 @@
...
+      @Parameter(description = "Fully qualified name of the data model", schema = @Schema(type = "string"))
+          @PathParam("fqn")
+          String fqn)
+      throws IOException {
+    return deleteByName(uriInfo, securityContext, fqn, false, hardDelete);
+  }
+
+  @PUT
+  @Path("/restore")
+  @Operation(
```
sureshms, line 415: Do we need soft deleted and restore functionality for dataModels? cc @harshach

openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java
```diff
@@ -486,4 +487,22 @@ public static void sortByTagHierarchy(List<Tag> tags) {
...
+   * fully populated object such that ElasticSearch index has all the details.
+   */
+  public static List<EntityReference> getEntityReferences(List<EntityReference> entities, Include include)
+      throws IOException {
+    if (nullOrEmpty(entities)) {
+      return Collections.emptyList();
+    }
+    List<EntityReference> refs = new ArrayList<>();
+    for (EntityReference entityReference : entities) {
+      EntityReference chartRef = Entity.getEntityReference(entityReference, include);
```
sureshms, line 503: Call chartRef as `entityRef` since this method is generic.

openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java
```diff
@@ -486,4 +487,22 @@ public static void sortByTagHierarchy(List<Tag> tags) {
...
+      refs.add(chartRef);
+    }
+    return refs.isEmpty() ? null : refs;
```
sureshms, line 506: refs will always have members because the check on line 498 already checked the list is not empty.

openmetadata-spec/src/main/resources/json/schema/entity/data/dataModel.json
Daandamhuis: Sorry, maybe I'm overstepping, but a data model is not only for building dashboards. I would at least including "Reports and Dashboards". In the "Power BI" world a datamodel is used in several different places, but mainly Interative- and Static reports and dashboard .

openmetadata-service/src/main/java/org/openmetadata/service/resources/datamodels/DataModelResource.java
nahuelverdugo: Does it mean we should do the same for the charts since Charts are related to Dashboards?

openmetadata-service/src/main/java/org/openmetadata/service/resources/datamodels/DataModelResource.java
nahuelverdugo: We are allowing separate followers for Charts, and they are part of Dashboards, I followed the same approach.

openmetadata-service/src/main/java/org/openmetadata/service/resources/datamodels/DataModelResource.java
nahuelverdugo: Same here, I followed the same approach that we have for Charts.

openmetadata-spec/src/main/resources/json/schema/entity/data/dataModel.json
nahuelverdugo: I think it is fine to call it `DashboardDataModel`, I didn't like the idea of repeated class names. @Daandamhuis when we refer here to `DashboardDataModel` we said that the DataModel is part of a `DashboardService`. This is a first iteration where we want to extract the `DataModels` and create a lineage from `Data Source` > `Data Model` > `Dashboard`. We can keep iterating later to make it more precise in case the `Data Model` can be associated to a `Dashboard` or a `Report`, etc.

mohityadav766 (approved):
Approved

mohityadav766 (approved):

PR #10650, opened 2023-03-17
fix e2e test for mysql

NiharDoshi99:
### Describe your changes :

I worked on fixing e2e cli for mysql

#
### Type of change :

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation


ingestion/tests/cli_e2e/test_cli_mysql.py
```diff
@@ -92,12 +92,12 @@ def expected_filtered_schema_excludes() -> int:
 
     @staticmethod
     def expected_filtered_table_includes() -> int:
-        return 49
+        return 50
```
NiharDoshi99, line 95: increasing count, because we have added a new table `query_entity` which is filtered

TeddyCr (approved):

PR #10659, opened 2023-03-19
Move common functionality to EntityRepository store method

sureshms:
### Describe your changes :

Setting the fields `owner`, `tags`, `href` before writing entity JSON to the backend database and restoring them after the write can be moved to `EntityRepository.store()` method instead of doing it in every class the implements the `storeEntity` method.

#
### Type of change :

- [x] Improvement
#
### Checklist:


harshach (approved):

sureshms:
@chirag-madlani cypress tests are failing. This change should not be causing it.

Issue #9344, opened 2022-12-16
Admin User Gets Created again and triggers mail every time


mohityadav766:
**Affected module**
Does it impact the UI, backend or Ingestion Framework?
Backend

**Describe the bug**
When an admin owns a particular entity, then during the server, since user are initialised before any other resource, then the line
    user.setOwns(fields.contains("owns") ? getOwns(user) : null);
    will cause issues

**To Reproduce**

Steps to reproduce the issue ->

Create an admin
Login with the password sent over mail, and create a Data Insight pipeline
Restart the server, now when we try to fetch User will all the fields since UserRepository is initialised before the IngestionPipeline resource it will fail with EntityNotFoundException and again try to create a user and send mail to the user.

Screenshots or steps to reproduce

**Expected behavior**

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8]
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

mohityadav766:
fixed

PR #10697, opened 2023-03-21
Fix Email sent everytime a admin owns an Entity during startup

mohityadav766:
### Describe your changes :

Fix Email Issue #9344

Steps to reproduce the issue ->
1. Create an admin 
2. Login with the password sent over mail, and create a Data Insight pipeline
3. Restart the server, now when we try to fetch User will all the fields since UserRepository is initialised before the IngestionPipeline resource it will fail with EntityNotFoundException and again try to create a user and send mail to the user.

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

#


openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -179,8 +179,9 @@ public void initialize(OpenMetadataApplicationConfig config) {
     this.authenticationConfiguration = config.getAuthenticationConfiguration();
     SmtpSettings smtpSettings = config.getSmtpSettings();
     this.isEmailServiceEnabled = smtpSettings != null && smtpSettings.getEnableSmtpServer();
-    this.dao.initializeUsers(config);
+    // Keep this before initializeUsers, else getUpdater() will fail
     SubjectCache.initialize();
+    this.dao.initializeUsers(config);
```
harshach, line 184: this can run into other issues
isn't subjectCache requires user repository to be available
and how does subject cache will have the admins pre-initialized?

mohityadav766: when we have user migrating from any other SSO to basic auth, during initlization we check populate authMechanism which was null earlier with a randomly generated password, in that case SubjectContext.getInstance will be null

harshach (approved):

PR #10723, opened 2023-03-23
[WIP] - Remove SecretsManagerCredentials from IngestionPipeline

pmbrull:
### Describe your changes:

The `secretsManagerCredentials` property in the ingestion pipeline was not being used anywhere (not being set at any point in time in the backend) and it was misleading.

We are now removing it, and updating the migration scripts just in case. As we already removed the full OM connection from the Ingestion Pipeline Entity table, we are only updating the OM service.

---

UPDATE: Ok so this is a bit trickier than anticipated, but we need to make our intent clear on how we are defining the classes.

Current approach:
1. The OM connection class has a `secretsManagerCredentials` that we build on the fly, and then pass it to the `ometa_api` object.
2. This is confusing in two ways:
    1. We share this object with server <> client and it is actually only needed in the client
    2. We make it seem that is the server the one passing the keys to the client to connect to the SM

Preferred approach:
1. The OM connection class does not have the property `secretsManagerCredentials`. This is something that relies completely on the client side
2. The `SecretManagerFactory` accepts the Secret Manager Provider + a registry of functions (this is an abstract class we'll need to define).
3. Basically, depending on the environment, we'll load SM creds in a way or another (e.g., from `airflow.cfg`, env vars, reading a file,...). The Function Registry should be selected by the client, depending on the environment, and allow the factory to know how to read the creds.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-clients/openmetadata-java-client/pom.xml
+++ b/openmetadata-clients/openmetadata-java-client/pom.xml
@@ -95,14 +95,10 @@
         </dependency>
         <!-- TEST -->
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.13.2</version>
-        </dependency>
-        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>
             <version>5.9.2</version>
+            <scope>test</scope>
         </dependency>
 
         <!--avoid security issue https://security.snyk.io/vuln/SNYK-JAVA-ORGMOZILLA-1314295-->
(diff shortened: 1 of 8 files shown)


openmetadata-clients/openmetadata-java-client/pom.xml
```diff
@@ -99,6 +99,11 @@
             <artifactId>junit</artifactId>
             <version>4.13.2</version>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>5.9.2</version>
+        </dependency>
```
nahuelverdugo, line 106: <3 

Should we remove `junit`? Are we using that for anything?

nahuelverdugo (approved):
Minor comment in case we can remove an unused dependency from the `openmetadata-clients` POM.

nahuelverdugo (approved):

pmbrull:
this requires more work, we'll handle it as part of a #11548 with a new branch

PR #10756, opened 2023-03-24
Fix Sample data ingestion which have enum type columns

### LINKED ISSUE #9462: Profiler ingestion can not import data from MySQL tables which have enum-type columns
**Affected module**
Ingestion

**Describe the bug**
I tried to import sample data from my MySQL database through profiler ingestion. Using OpenMetadata 0.12.3. MySQL version is ‘8.0.31’.
But it doesn’t import from some tables with warning message like:

```
WARNING - Error fetching sample data: 'f2f' is not among the defined enum values. Enum name: None. Possible values: None
```

Ingestion process finished as `SUCCESS`, but some tables are not imported.
The table, on which this warning message is shown, has a column with enum type like following:

```
`method_type` enum('f2f','tel','other')
```

The warning message says ‘f2f’ is not a valid enum value, but it is right value.

**To Reproduce**

The table definition, on which this problem occurred, is following. But I think this problem will occur on any tables which have enum-type columns.

```
CREATE TABLE `projects` (
  `id` int unsigned NOT NULL AUTO_INCREMENT COMMENT 'ID',
  `name` varchar(255) CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci NOT NULL COMMENT '案件名',
  `summary` text CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci COMMENT '案件概要',
  `accepted_at` date DEFAULT NULL COMMENT '案件受領日',
  `method_type` enum('f2f','tel','other') CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci DEFAULT NULL COMMENT '実施方法（f2f => 対面、tel => 電話、other => その他）',
  `venue` varchar(255) CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci DEFAULT NULL COMMENT '実施場所',
  `keywords` text CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci COMMENT 'キーワード',
  `note` text CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci COMMENT '備考',
  `created_at` timestamp NULL DEFAULT NULL COMMENT '作成日時',
  `updated_at` timestamp NULL DEFAULT NULL COMMENT '更新日時',
  `deleted_at` timestamp NULL DEFAULT NULL COMMENT '削除日時',
  `status` enum('registered','searching','recommended','decided','lost') CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci NOT NULL DEFAULT 'registered' COMMENT 'ステータス（registered => ''---''、searching => サーチ中、recommended => 推薦済、decided => 決定、lost => 失注）',
  `type` enum('spot','other') CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci DEFAULT NULL COMMENT '案件タイプ（sport => スポット、other => その他）',
  `specialized_area_id` int unsigned NOT NULL COMMENT '専門領域ID',
  `salesforce_project_id` varchar(140) CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci DEFAULT NULL COMMENT 'Salesforce案件ID',
  `interview_id` varchar(256) CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci DEFAULT NULL COMMENT 'SPEEDAのインタビューID',
  PRIMARY KEY (`id`),
  KEY `projects_specialized_area_id_foreign` (`specialized_area_id`),
  KEY `projects_idx01` (`id`,`deleted_at`),
  KEY `projects_idx02` (`interview_id`,`deleted_at`),
  CONSTRAINT `projects_specialized_area_id_foreign` FOREIGN KEY (`specialized_area_id`) REFERENCES `specialized_areas` (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=12208 DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_unicode_ci COMMENT='案件'
```

**Expected behavior**
Tables which have enum-type columns must be imported. 

**Version:**
 - OS: 
 
> Distributor ID:	Debian
> Description:	Debian GNU/Linux 11 (bullseye)
> Release:	11
> Codename:	bullseye

 - Python version: 3.9.13
 - OpenMetadata version: 0.12.3
 - OpenMetadata Ingestion package version: openmetadata-ingestion[all]==0.12.3

MilanBariya:
### Describe your changes:

Fixes <issue-number>

I worked on Fixing Sample data ingestion which have enum type columns.
Fix #9462 
Screenshot 2023-03-24 at 9 40 01 PM

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/profiler/profiler/sampler.py
```diff
@@ -40,6 +42,27 @@
 RANDOM_LABEL = "random"
 
 
+def _object_value_for_elem(self, elem):
```
MilanBariya, line 45: Override this function, Becouse in Enum it was giving KeyError from here `return self._object_lookup[elem]`.
now make this to` return self._object_lookup.get(elem, elem)`.

ingestion/src/metadata/profiler/profiler/sampler.py
```diff
@@ -12,11 +12,13 @@
 Helper module to handle data sampling
 for the profiler
 """
+# pylint: disable=consider-using-f-string,disable=inconsistent-return-statements, disable=protected-access
```
pmbrull, line 15: We don't want to disable this for the full file. Whatever can be fixed in the change code, let's fix. Otherwise, just disable it in the specific line where this is raised

pmbrull (changes requested):

ingestion/src/metadata/profiler/profiler/sampler.py
MilanBariya: sure @pmbrull

ingestion/src/metadata/profiler/profiler/sampler.py
```diff
@@ -40,6 +42,27 @@
...
+                % (
+                    elem,
+                    self.name,
+                    util.langhelpers.repr_tuple_names(self.enums),
+                )
+            ),
+            replace_context=err,
+        )
+
+
```
ulixius9, line 62: If we are doing a get in dict there is no possibility of getting a key error so no need sof this exception handling!

MilanBariya: Yes, we can remove that.

ingestion/src/metadata/profiler/profiler/sampler.py
pmbrull: Not sure if returning `elem` if this fails when getting the values from the dict is the right approach or we would be altering the reality. Thinking if returning `None` might be better suited for this.

@TeddyCr what are your thoughts on this?

ulixius9: @pmbrull so [here](https://github.com/open-metadata/OpenMetadata/blob/f54efc649b4bcbf154ba78c53ca23787b9aee671/ingestion/src/metadata/profiler/orm/converter.py#L66) we have mapped enum type to sqlalchemy.Enum and we are supposed to pass the enum associated with this column (checkout the first code snippet) now we do not store this enum map which is required 

image


So in that case, if we map by default return None, we will always get None because there is no enum map to lookup, so what we are doing here is basically trusting the database that it will be storing the correct map key and showing directly that on the UI, but even in this approach we will be only able to display what database has stored (i.e the key) and not the actual value of the same! 

image

PR #10806, opened 2023-03-28
Fix#10542: Add sample data mask functionality

### LINKED ISSUE #10542: Mask PII sample data and don't return column metrics
If a column is flagged as `PII.Sensitive` and the user does not have permissions, we should not show the sample data for that column (maybe show ***) and we should not return the column metrics as well.

Otherwise, a sensitive column could be semi-inferred just by seeing the min/max values, for example.

cc @harshach @TeddyCr

nahuelverdugo:
### Describe your changes:

Fixes #10542

I have added a Rule Operator called `ViewPIISampleData` associated with the `Table` resource that allows us to create rules to mask in the Sample Data the columns tagged with `PII.Sensitive`. In case of columns with children, if any of the children is tagged with `PII.Sensitive` the parent column display in the Sample Data will be masked.

For this schema:
image

This is how it looks like when the user is the owner:
image

When the user is not the owner or it is not specified:
image

### Type of change:
- [x] New feature

### Checklist:
- [x] The issue properly describes why the new feature is needed, what's the goal, and how we are building it. Any discussion or decision-making process is reflected in the issue.

### DIFF
--- a/openmetadata-service/src/test/java/org/openmetadata/service/resources/permissions/PermissionsResourceTest.java
+++ b/openmetadata-service/src/test/java/org/openmetadata/service/resources/permissions/PermissionsResourceTest.java
@@ -22,8 +22,10 @@ import static org.openmetadata.schema.type.MetadataOperation.EDIT_DISPLAY_NAME;
 import static org.openmetadata.schema.type.MetadataOperation.EDIT_LINEAGE;
 import static org.openmetadata.schema.type.MetadataOperation.EDIT_OWNER;
 import static org.openmetadata.schema.type.MetadataOperation.EDIT_TAGS;
+import static org.openmetadata.schema.type.MetadataOperation.VIEW_PII_SAMPLE_DATA;
 import static org.openmetadata.schema.type.Permission.Access.ALLOW;
 import static org.openmetadata.schema.type.Permission.Access.CONDITIONAL_ALLOW;
+import static org.openmetadata.schema.type.Permission.Access.DENY;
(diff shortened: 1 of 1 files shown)


harshach:
@nahuelverdugo this is bit more nuanced.
Examples users may want to mask the PII.Sensitive data only in such cases only columns which are tagged as PII should be masked
Second use case would be , a user who has permission to view PII data i.e his role / policy has the permissions to look at PII data. In this case even though the column is PII tagged we should show the data

nahuelverdugo:
> @nahuelverdugo this is bit more nuanced. Examples users may want to mask the PII.Sensitive data only in such cases only columns which are tagged as PII should be masked Second use case would be , a user who has permission to view PII data i.e his role / policy has the permissions to look at PII data. In this case even though the column is PII tagged we should show the data

@harshach in this PR we give the possibility to create rules for policies. 

I have added this one for the Organization:

```json
  {
      "name": "OrganizationPolicy-ViewPIISampleData-Rule",
      "description" : "Deny seeing column sample data for table columns tagged with PII.Sensitive.",
      "resources" : ["table"],
      "operations": ["ViewPIISampleData"],
      "effect": "deny",
      "condition": "!isOwner()"
    }
```

If the user is the owner (or part of the owner team), it will be able to see the columns tagged with `PII.Sensitive`. Otherwise, the column value will be shown masked.

Am I missing something?

nahuelverdugo:
I am closing this PR since the approach must be different.

Issue #9185, opened 2022-12-07
SearchResource endpoint to re-create ES indexes


pmbrull:
~~We're doing something similar on the Airflow APIS https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-airflow-apis/setup.py#L71~~

~~We now want to package the ES index definitions files inside the ingestion package so that we can define them in a single place and use them to recreate the indexes during the `metadata_to_es` workflow.~~

~~Let's use the files and structure added in this PR https://github.com/open-metadata/OpenMetadata/pull/9163~~

Based on the discussion in the PR, the goal of this task is to create an endpoint in `SearchResource` to allow admins and bots to re-create indexes from an API call.

The goal is to not have to maintain two sets of index definitions in the backend and ingestion side.

pmbrull:
Based on the discussion in the PR, the goal of this task is to create an endpoint in `SearchResource` to allow admins and bots to re-create indexes from an API call.

The ingestion framework can then call this endpoint in `metadata_to_es`.

PR #10852, opened 2023-03-30
Update ingestion ES mappings & restructure sink

pmbrull:
### Describe your changes:

1. There were many mappings not updates since the last changes in the backend. We're going to tackle this problem completely in https://github.com/open-metadata/OpenMetadata/issues/9185
2. The ES sink was not properly structured and it was quite messy. Now the last cleaning piece pending will be done with the issue above.
3. We also added the missing Container and Query entities indexing capabilities in the framework

In this PR we centralized a ton of the processing to make it easier to follow and evolve.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/sink/elasticsearch.py
+++ b/ingestion/src/metadata/ingestion/sink/elasticsearch.py
@@ -18,6 +18,7 @@ We disable unexpected-keyword-arg as we get a false positive for request_timeout
 import json
 import ssl
 import traceback
+from functools import singledispatch
 from typing import List, Optional
 
 import boto3
@@ -32,6 +33,7 @@ from metadata.generated.schema.entity.classification.classification import (
     Classification,
 )
 from metadata.generated.schema.entity.classification.tag import Tag
+from metadata.generated.schema.entity.data.container import Container
 from metadata.generated.schema.entity.data.dashboard import Dashboard
 from metadata.generated.schema.entity.data.database import Database
 from metadata.generated.schema.entity.data.databaseSchema import DatabaseSchema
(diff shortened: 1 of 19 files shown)


ingestion/src/metadata/ingestion/sink/elasticsearch.py
```diff
@@ -223,70 +209,89 @@ def __init__(
...
                 WEB_ANALYTIC_ENTITY_VIEW_REPORT_DATA_INDEX_MAPPING,
             )
 
+        if self.config.index_containers:
+            self._check_or_create_index(
+                ES_INDEX_MAP[Container.__name__],
+                CONTAINER_ELASTICSEARCH_INDEX_MAPPING,
+            )
+
+        if self.config.index_queries:
```
pmbrull, line 283: all these if-else will disappear in https://github.com/open-metadata/OpenMetadata/issues/9185

ingestion/src/metadata/ingestion/sink/elasticsearch.py
```diff
@@ -223,70 +207,89 @@ def __init__(
...
+                ES_INDEX_MAP[Query.__name__],
+                QUERY_ELASTICSEARCH_INDEX_MAPPING,
+            )
+
+        # Prepare write record dispatching
+        self._write_record = singledispatch(self._write_record)
+        self._write_record.register(Classification, self._write_classification)
+        self._write_record.register(ReportData, self._write_report_data)
+        self._write_record.register(Policy, self._write_policy)
+
```
TeddyCr, line 292: Should we move these into their own class with a `write()` method, instantite that object `here` dynamically and have `def write_record()` call `self.<object>.write(client,  record)`?

pmbrull: it would be the same right? The only difference on handling the process in a class or as functions. I'd rather go the function approach here as there's no other methods we'd need to add in the class

TeddyCr (approved):

PR #10876, opened 2023-04-01
fixes #10755: Changed description change method for service connections from 'Put' to 'PATCH'

### LINKED ISSUE #10755: UI: Use PATCH API to update service description
**Affected module**
UI

**Describe the bug**
When a user with Data Steward role tries to edit the service description `connection must not be null` error is thrown, this is because the User does not have access to the connection and PUT request requires the connection object, if UI makes use of PATCH api like this error can be resolved! also this makes the behaviour consistent as we are using patch api everywhere else to update the description! 

image

aniketkatkar97:
### Describe your changes:

Fixes #10755 

I worked on changing description change method for service connections from 'Put' to 'PATCH'

#
### Type of change:

- [x] Bug fix
- [x] Improvement

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/pages/service/index.tsx
```diff
@@ -729,20 +729,15 @@ const ServicePage: FunctionComponent = () => {
...
+      const jsonPatch = compare(serviceDetails, updatedData);
 
       try {
-        const response = await updateService(
-          serviceName,
-          id,
-          updatedServiceDetails
-        );
+        const response = await updateOwnerService(serviceName, id, jsonPatch);
         setDescription(updatedHTML);
```
Sachin-chaurasiya, line 741: We should set description from the response here.

harshach (dismissed):

Sachin-chaurasiya (approved):

chirag-madlani (approved):

ShaileshParmar11 (approved):

PR #10902, opened 2023-04-03
Fix ES aligning suggests with Java

pmbrull:
### Describe your changes:

A bunch of the suggests had the wrong numbers. Cause:
- Wrong generalization when doing the ES cleanup
- Some weights on older ingestion versions were also not aligned with the ones on `<Entity>Index.java`. This PR aligns this
- We introduce the right pydantic type to handle suggests, which also were different in Java and Python

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/models/es_documents.py
+++ b/ingestion/src/metadata/ingestion/models/es_documents.py
@@ -31,6 +31,11 @@ from metadata.generated.schema.type.tagLabel import TagLabel
 from metadata.generated.schema.type.usageDetails import UsageDetails
 
 
+class ESSuggest(BaseModel):
+    input: str
+    weight: int
+
+
 class TableESDocument(BaseModel):
     """ElasticSearch Mapping doc"""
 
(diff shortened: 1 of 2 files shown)


harshach:
@pmbrull instead of maintaining the indexes on the both the side. We should make a call to OpenMetadata server to create the indexes cc @mohityadav766 
1. Since we have recreateIndexes api that we are using to do that from the UI we should use that from ingestion as well.
2. Only allow bot or admin to re-create the indexes
3. If the ingestion pipeline is not configured with re-create indexes to true (default should always be false) we shouldn't need to make this call

ulixius9 (dismissed):

pmbrull:
> @pmbrull instead of maintaining the indexes on the both the side. We should make a call to OpenMetadata server to create the indexes cc @mohityadav766
> 
> 1. Since we have recreateIndexes api that we are using to do that from the UI we should use that from ingestion as well.
> 2. Only allow bot or admin to re-create the indexes
> 3. If the ingestion pipeline is not configured with re-create indexes to true (default should always be false) we shouldn't need to make this call

Issue open https://github.com/open-metadata/OpenMetadata/issues/9185

ulixius9 (approved):

pmbrull:
changes already in main cc @chirag-madlani

PR #10914, opened 2023-04-04
Ingestion pipeline by service type task

07Himank:
I worked on ... because ... added filter for fetching ingestion pipeline by pipelineType.


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java
```diff
@@ -126,6 +133,18 @@ private String getWebhookTypePrefixCondition(String tableName, String typePrefix
         : String.format("%s.webhookType LIKE '%s%%'", tableName, typePrefix);
   }
 
+  private String getPipelineTypePrefixCondition(String tableName, String pipelineType) {
+    pipelineType = escape(pipelineType);
+    if (DatasourceConfig.getInstance().getDriverClass().startsWith("com.mysql.cj.jdbc")) {
```
nahuelverdugo, line 138: We are hiding some logic here, can we move this into a method of `DatasourceConfig` called `isMySQLConnector()` or something like this?

nahuelverdugo (commented):
The PR description is empty, it is hard to know what are we reviewing or if it is part of some issue.

07Himank:
> The PR description is empty, it is hard to know what are we reviewing or if it is part of some issue.



> The PR description is empty, it is hard to know what are we reviewing or if it is part of some issue.

In this Pr we had added the filter to fetch the ingestion pipeline by pipelineType for eg:-pipelineType=dataInsight

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java
07Himank: Thank you for your suggestion. I have updated the code.

nahuelverdugo (approved):

PR #10917, opened 2023-04-04
fixes: Improve search functionality for teams page #10653

### LINKED ISSUE #10653: Improve search functionality for glossary term and teams page
Glossary term and teams page has the tree structure view, and the search is limited to only terms/teams available on UI side (which are already fetched and stored in the state), but the problem with this approach is, I will not get terms/team which does not yet fetch.

Solution:
The solution would be to fetch team/term from ES at the time of searching

ShaileshParmar11:
### Describe your changes:

Fixes #10653 

https://user-images.githubusercontent.com/71748675/229782466-50c78f18-511f-4aed-8515-7e41b2e892db.mov

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/components/TeamDetails/TeamDetailsV1.tsx
```diff
@@ -347,6 +350,19 @@ const TeamDetailsV1 = ({
         ]),
   ];
 
+  const searchTeams = async (text: string) => {
+    try {
+      const res = await getSuggestions(text, SearchIndex.TEAM);
+      const data = (res.data as RawSuggestResponse<SearchIndex.TEAM>).suggest[
+        'metadata-suggest'
+      ][0].options.map((value) => value._source);
```
chirag-madlani, line 358:
```
const getSuggestResults = async (value: string) => {
    try {
      const data = await getSuggestions<ExploreSearchIndex>(
        value,
        SearchIndex[
          entityType as keyof typeof SearchIndex
        ] as ExploreSearchIndex
      );
      setData(
        formatDataResponse(data.data.suggest['metadata-suggest'][0].options)
      );
    } catch (error) {
      showErrorToast(
        error as AxiosError,
        t('server.entity-fetch-error', {
          entity: t('label.suggestion-lowercase-plural'),
        })
      );
    }
  };

```

For better typing @ShaileshParmar11

openmetadata-ui/src/main/resources/ui/src/components/TeamDetails/TeamDetailsV1.utils.tsx
```diff
@@ -69,21 +68,3 @@ export const getTabs = (
...
-    if (matched) {
-      results = [...results, { ...team, children: undefined }];
-    }
-    if (hasChildren) {
-      results = [...results, ...searchTeam(team.children as Team[], value)];
-    }
-  }
-
-  return results;
-};
```
chirag-madlani, line 89: Let's have this method around we need it in future

chirag-madlani (approved):

openmetadata-ui/src/main/resources/ui/src/components/TeamDetails/TeamDetailsV1.utils.tsx
ShaileshParmar11: sure!

PR #10944, opened 2023-04-05
Fixed Issue #10943: Impala query engine metadata ingestion and median function profiler

### LINKED ISSUE #10943: Metadata does not ingest properly when connected to an Impala query engine
**Affected module**
Ingestion Framework

**Describe the bug**

- Tables and views are not recognized as separate entities.
- Descriptions are not being captured
- "Show Views" SQL query is not valid in Impala dialect
- The SQL function "percentile" used in the profiler for computing Median, First and Third quartile is not valid in Impala.

**To Reproduce**

**Expected behavior**

**Version:**
 - Python version:

**Additional context**

ksirmons:
### Describe your changes:

Fixes #10943 

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement

#
### Checklist:


ingestion/src/metadata/profiler/orm/converter.py
```diff
@@ -104,6 +104,19 @@ def check_snowflake_case_sensitive(table_service_type, table_or_col) -> Optional
...
+    
+    Args:
+        table_service_type: the main sql engine to determine if we should always quote.
+    Return: True or False
+    """
+    
+    if table_service_type == databaseService.DatabaseServiceType.Hive:
+        return True
+    
+    return None
```
TeddyCr, line 119: @ksirmons do you think we should add this logic to `check_snowflake_case_sensitive` and renamed `check_snowflake_case_sensitive` to `quote_names`?

ingestion/src/metadata/ingestion/source/database/hive/metadata.py
```diff
@@ -183,6 +183,59 @@ def get_table_comment(  # pylint: disable=unused-argument
...
 
+def get_impala_table_or_view_names(connection, schema=None, targetType="table"):
+    query = f"show tables"
+    if schema:
+        query += " IN " + schema
+        
+    cursor = connection.execute(query)
+    results = cursor.fetchall()
+    tables_and_views = [result[0] for result in results]
+    #print(tables_and_views)
```
TeddyCr, line 194: Could we remove this? 🙂

ingestion/src/metadata/ingestion/source/database/hive/metadata.py
```diff
@@ -251,13 +323,20 @@ def prepare(self):
...
         only after hive 2.2.0 version
         """
-        result = dict(self.engine.execute("SELECT VERSION()").fetchone())
-        version = result.get("_c0", "").split()
-        if version and self._parse_version(version[0]) >= self._parse_version(
-            HIVE_VERSION_WITH_VIEW_SUPPORT
-        ):
-            HiveDialect.get_table_names = get_table_names
-            HiveDialect.get_view_names = get_view_names
+        if self.engine.driver == "impala":
```
TeddyCr, line 326:
```python
        if self.engine.driver == Dialects.Impala:
```

We'll need to import `from metadata.profiler.orm.registry import Dialects` at the top

ingestion/src/metadata/profiler/orm/converter.py
ksirmons: Actually I think we should always quote the column names no matter the db engine,  unless there is a place to check the column name against a reserve word list.

TeddyCr: @ksirmons sqlalchemy will automatically quote the names based on the engine backend implementation if this argument is set to `None` -- which is the default value. In certain scenarios (i.e. snowflake case sensitive column names) we had to override it and set it to True.

ingestion/src/metadata/profiler/orm/functions/median.py
```diff
@@ -82,6 +81,39 @@ def _(elements, compiler, **kwargs):
...
+        group by quarter
+        )
+        select max(case when quarter = 1 then quartile_value end) as first_q
+        , max(case when quarter = 2 then quartile_value end) as second_q
+        , max(case when quarter = 3 then quartile_value end) as third_q
+        , max(case when quarter = 4 then quartile_value end) as fourth_q
+        from quarters
+        group by grp
+        ;
+    """
```
TeddyCr, line 113: Could we move this to the function docstring?

TeddyCr (approved):
LGTM 🚀 Thank you @ksirmons for the contribution

PR #10962, opened 2023-04-06
Fixes #6340: Implement Median support for MySQL

### LINKED ISSUE #6340: Implement Median support for MySQL
There is no straightfroward support for median computation with MySQL currently. Solutions often involve few subqueries. Currently we are returning `NULL` for MySQL; we need to implement the correct computation to support this metric.

```
@compiles(MedianFn, Dialects.MySQL)
def _(elemenst, compiler, **kwargs):
    """Median computation for MySQL currently not supported
    Needs to be tackled in https://github.com/open-metadata/OpenMetadata/issues/6340
    """
    return "NULL"
```

schlameel:
### Describe your changes:

Fixes #6340 - Implement Median support for MySQL

I added SQL to calculate the median (percentile) value.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Improvement

**IMPORTANT NOTE** - this is untested in OM [per Teddy](https://openmetadata.slack.com/archives/C02B6955S4S/p1680761094410389?thread_ts=1680213775.207729&cid=C02B6955S4S). It has been tested in MySQL and python. I performance tested this and other solutions and this was fastest in a limited, non-real world environment.


ingestion/src/metadata/profiler/orm/functions/median.py
```diff
@@ -116,10 +116,28 @@ def _(elements, compiler, **kwargs):
...
+    return """
+    (SELECT
+        {col}
+    FROM (
+        SELECT
+            t.{col}, 
+            @row_num := @row_num + 1 AS row_num
+        FROM 
+            {table} t,
+            (SELECT @row_num:=0) counter
```
TeddyCr, line 133: @schlameel just wondering why we are doing this cross join here?

ingestion/src/metadata/profiler/orm/functions/median.py
```diff
@@ -116,10 +116,28 @@ def _(elements, compiler, **kwargs):
...
+    table = elements.clauses.clauses[1].value
+    percentile = elements.clauses.clauses[2].value
+
+    return """
+    (SELECT
+        {col}
+    FROM (
+        SELECT
+            t.{col}, 
+            @row_num := @row_num + 1 AS row_num
```
TeddyCr, line 130: I think we'll need to replace `@row_num := @row_num + 1 AS row_num` with `ROW_NUMBER() OVER () AS row_num`. I get the following warning on my end

```
[Warning: #1287 Setting user variables within expressions is deprecated and will be removed in a future release](https://stackoverflow.com/questions/67256687/mysql-warning-1287-setting-user-variables-within-expressions-is-deprecated-a)
```

TeddyCr (changes requested):
@schlameel thank you for opening this PR. I added a few comments

ingestion/src/metadata/profiler/orm/functions/median.py
schlameel: @TeddyCr what are you using to generate that warning? I'm trying to find the best solution and want to see if my solution is still throwing that warning.  I can use the ROW_NUMBER() replacement, but I needed the final value of `row_num` to calculate row number that matches the percentile.  Thanks.

ingestion/src/metadata/profiler/orm/functions/median.py
schlameel: It is really just setting an initial value for `@row_num. Changed.

ingestion/src/metadata/profiler/orm/functions/median.py
TeddyCr: I am using MySQL 8.0.

I am just wondering what is the advantage betweenusing a the variable and a cross join vs the below

```sql
    (
    SELECT
        {col}
     FROM (
         SELECT
             t.{col}, 
             ROW_NUMBER() OVER () AS row_num,
             COUNT(*) OVER() AS counter
         FROM 
            {table} t
         ORDER BY installed_rank
         ) temp
     WHERE
	 	temp.row_num = ROUND({percentile} * counter)
     )
```

schlameel: For testing, I am using the om MySQL docker image with a table of 1 million rows.  The query:
```
    (SELECT
        {col}
    FROM (
        SELECT
            t.{col}, 
            ROW_NUMBER() OVER () AS row_num
        FROM 
            {table} t,
            (SELECT @counter := COUNT(*) FROM {table}) t_count 
        ORDER BY {col}
        ) temp
    WHERE temp.row_num = ROUND({percentile} * @counter)
    )
```
is five to seven times faster with an index and 4 times faster without an index than your query above.

Not knowing the potential impacts on production systems, I tried to find the fastest query I could. Granted my single table is probably an oversimplification.

TeddyCr: Make sense. That was my assumption, just wanted to confirm. Thanks @schlameel

TeddyCr (approved):
LGTM 🚀

PR #10967, opened 2023-04-07
fix(ui): increase worker size for unit tests

chirag-madlani:
Increase workers from 3 to 4 so that unit tests CI can run faster

I worked on increasing worker size for unit tests to speed up CI


ShaileshParmar11:
Meven & sonar both failing

chirag-madlani:
Not optimising in time hence closing this for now

Issue #10916, opened 2023-04-04
UI: configure add new query flow for table entity


ShaileshParmar11:
**Feature**
Add feature issue reference
need to configure add a new query flow for the table entity

**Describe the task**

need mock for the flow @devyani-kaushik

PR #10978, opened 2023-04-10
ui: integrated add query flow

ShaileshParmar11:
### Describe your changes:

Fixes [#10916](https://github.com/open-metadata/OpenMetadata/issues/10916)

I worked on Integrating new add query flow, and feedback for query feature

image

image

image

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/components/TableQueries/QueryCard.tsx
```diff
@@ -128,23 +156,44 @@ const QueryCard: FC<QueryCardProp> = ({
           title={
             <Space className="font-normal p-y-xs" size={8}>
               <Text>{queryDate}</Text>
-              <Text>{SINGLE_DOT}</Text>
-              <Text>{`${t('label.by-lowercase')} ${query.updatedBy}`}</Text>
+              <Text className="text-gray-400">{PIPE_SYMBOL}</Text>
```
chirag-madlani, line 159: Did we use same Pipe symbol as EntityInfo

openmetadata-ui/src/main/resources/ui/src/components/TableQueries/table-queries.style.less
```diff
@@ -40,3 +38,23 @@
 .query-card-container.selected {
   border: 1px solid @info-color;
 }
+
+.query-right-panel {
+  position: sticky;
+  height: calc(100vh - @summary-panel-offset);
+  z-index: 9;
```
chirag-madlani, line 45: We should not modify zIndex, Is it required?

ShaileshParmar11:
> @ShaileshParmar11 when users adding query against a table we should add that table to queryUsedIn by default.

currently, we are showing the selected option in the dropdown, do we not show that and add the table directly in the query used in and if users select more tables we also add that?

openmetadata-ui/src/main/resources/ui/src/components/TableQueries/QueryCard.tsx
ShaileshParmar11: the symbol is the same, but in entity info, we have added it from localization. and we are avoiding that, right? to add symbol in localized file so I have added const for symbol

chirag-madlani: Let's make a note and remove it in next release

openmetadata-ui/src/main/resources/ui/src/components/TableQueries/table-queries.style.less
ShaileshParmar11: Thanks for pointing this out, and yes we don't need z-index here!

chirag-madlani (approved):

harshach (approved):

PR #10995, opened 2023-04-11
Use metabase models in client

ulixius9:
### Describe your changes:

 We have defined some models but we are not using it these models in metabase client, in this pr I've converted all the expected json reponse and mapped them to their respective models,

note: in the models I've marked most of the fields as optional as we are not using most of then and we do not want to break the processing of dashbord if some fields were not found 

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/metabase/client.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/metabase/client.py
@@ -17,6 +17,12 @@ from typing import List, Optional
 import requests
 
 from metadata.ingestion.connections.test_connections import SourceConnectionException
+from metadata.ingestion.source.dashboard.metabase.models import (
+    MetabaseDashboardDetails,
+    MetabaseDashboardList,
+    MetabaseDatabase,
+    MetabaseTable,
+)
 from metadata.utils.logger import ingestion_logger
 
 logger = ingestion_logger()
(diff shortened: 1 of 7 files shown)


ingestion/src/metadata/ingestion/source/dashboard/metabase/client.py
```diff
@@ -72,29 +78,28 @@ def get_dashboards_list(self) -> Optional[List[dict]]:
...
         if resp_dashboards.status_code == 200:
-            return resp_dashboards.json()
-        return []
-
-    def get_dashboard_name(self, dashboard: dict) -> str:
-        """
-        Get Dashboard Name
-        """
-        return dashboard["name"]
+            return MetabaseDashboardList(dashboards=resp_dashboards.json())
```
OnkarVO7, line 81: Can you also add the relevant return type to the method

ingestion/src/metadata/ingestion/source/dashboard/metabase/client.py
```diff
@@ -72,29 +78,28 @@ def get_dashboards_list(self) -> Optional[List[dict]]:
...
+        logger.warning(f"Failed to fetch the dashboards: {resp_dashboards.text}")
+        return MetabaseDashboardList(dashboards=[])
```
OnkarVO7, line 83: I believe instead of returning this return `None` and then handle the same wherever this is called

ingestion/src/metadata/ingestion/source/dashboard/metabase/metadata.py
```diff
@@ -64,40 +64,40 @@ def create(cls, config_dict, metadata_config: OpenMetadataConnection):
...
+        self, dashboard_details: MetabaseDashboard
     ) -> Iterable[CreateDashboardRequest]:
         """
         Method to Get Dashboard Entity
         """
         dashboard_url = (
-            f"{clean_uri(self.service_connection.hostPort)}/dashboard/"
-            f"{dashboard_details['id']}-"
-            f"{replace_special_with(raw=dashboard_details['name'].lower(), replacement='-')}"
+            f"{clean_uri(self.service_connection.hostPort)}/dashboard/{dashboard_details.id}-"
```
OnkarVO7, line 93: Adding the logic to build the urls in a separate method might be cleaner

ingestion/src/metadata/ingestion/source/dashboard/metabase/models.py
OnkarVO7, line 1: @ulixius9 no need to add all the fields to the models. You can only define whatever is required and we use

OnkarVO7 (changes requested):

ingestion/src/metadata/ingestion/source/dashboard/metabase/client.py
ayush-shah: if we are looping it in the next topology, i believe keeping an empty list should work as expected

ayush-shah: if we are looping it in the next topology, i believe keeping an empty list should work as expected

OnkarVO7 (approved):
LGTM!

PR #11019, opened 2023-04-12
Looker - Update GitHub creds for UI form

pmbrull:
### Describe your changes:

The `GitHubCredentials` are optional. However, if we do not pass the `oneOf`, the UI renders it directly and as its internal properties are required, it forces us to specify the data there, even if we don't want to set this up.

Adding a dummy `oneOf` so that we can choose not to pass anything.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


TeddyCr (approved):
LGTM

PR #11023, opened 2023-04-12
Update to support 3.10 and 3.11

ayush-shah:


ayush-shah:
Closing this as 3.11 is not supported as of now, fails for multiple packages.

PR #11044, opened 2023-04-13
Fix for issue leading to vulnerability flags + extras

mohityadav766:
### Describe your changes:

Fixes <issue-number>

1. Fixed Vulnerability Issues
2. Created pipelines on startup for Reindexing and Insight
3. LdapUtil removed unnecessary connection-option initialisation
4. Fixed dependencies for WeSocket and removed deprecated method configure()

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/resources/services/metadata/MetadataServiceResource.java
```diff
@@ -93,23 +94,20 @@ private void registerMetadataServices(OpenMetadataApplicationConfig config) {
...
-              }
-            });
+        if (servicesList.size() == 1) {
+          MetadataService service = servicesList.get(0);
+          service.setFullyQualifiedName(service.getName());
+          service.setConnection(metadataConnection);
+          service.setAllowServiceCreation(false);
+          dao.initializeEntity(service);
+        } else {
+          throw new RuntimeException("Only one Openmetadata Service can be initialized from the Data.");
```
harshach, line 104: @mohityadav766 bit confused by this. 
We should've system flag for the OpenMetadata service that indicates that this config/service refers to the live system.
in upcoming releases we should've ability to add other OpenMetadata service to import data or export

openmetadata-service/src/main/java/org/openmetadata/service/util/IngestionPipelineUtils.java
```diff
@@ -0,0 +1,32 @@
...
+public class IngestionPipelineUtils {
+  public static AirflowConfig getDefaultAirflowConfig() {
+    return new AirflowConfig()
+        .withPausePipeline(false)
+        .withConcurrency(1)
+        .withPipelineTimezone("UTC")
+        .withRetries(3)
+        .withRetryDelay(300)
+        .withPipelineCatchup(false)
+        .withScheduleInterval("0 0 * * *")
```
harshach, line 27: are we hard-coding the schedule interval here?

mohityadav766: @harshach this is the default airflow config that we are using to create a insights and reindexing pipeline on startup , user can later on will be able to edit these if needed

openmetadata-service/src/main/java/org/openmetadata/service/resources/services/metadata/MetadataServiceResource.java
mohityadav766: @harshach this is the OpenMetadata Service creation , currently we just have one service available , if we want we can add more in the bootstrap later and intilialize as per the list , but for now we just have one system service , so this check is to ensure only one is there

PR #11051, opened 2023-04-13
K8 step update

dhruvinmaniar123:
Updated the docs that mention the install command to install openmetadata helm charts 

#
### Type of change:
- [x] Documentation

#
### Checklist:


harshach:
@dhruvinmaniar123 we should also document about the volumes. Its not clear that they need to create it in the default space.

akash-jain-10:
> @dhruvinmaniar123 we should also document about the volumes. Its not clear that they need to create it in the default space.

Hello @harshach . Volumes will be created in the same namespace where OpenMetadata Dependencies will be deployed for Airflow. The code snippet [here](https://docs.open-metadata.org/deployment/kubernetes/eks#provision-efs-backed-pvs-pvcs-for-airflow-dags-and-airflow-logs) includes the namespace where these volumes will be created.

akash-jain-10 (approved):

Issue #11068, opened 2023-04-14
Adding databases to OM using IAM role-based authentication


mhornst:
**Is your feature request related to a problem? Please describe.**
We would like to onboard a lot of databases / data stores to OpenMetadata (incl. S3 buckets, RDS, Redshift, DynamoDB etc.) during this quarter and would prefer to have IAM-role based authentication when adding connections for these. Currently, we would need to create a username & pwd in each.

**Describe the solution you'd like**
Have the possibility to rely on AWS IAM role assumption when connecting databases. This way, we could grant the right permissions to OpenMetadata's IAM role and would only need to open up network connectivity. In short, we would not need to rely on creating additional users and connecting databases, but manage policies for sources on AWS side using an IAM role for OpenMetadata.

**Describe alternatives you've considered**
Create individual users manually / with a script and connect databases one-by-one (not scalable as we want to onboard many dbs)

**Additional context**
–

PR #11078, opened 2023-04-15
ISSUE-11068: Add IAM authentication for MySQL ingestion

tutte:
### Describe your changes:

Fixes [ISSUE-11068](https://github.com/open-metadata/OpenMetadata/issues/11068)

I worked on MySQL IAM Authentication because users need to ingest from databases at scale without having to manage static credentials.

#
### Type of change:
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

- [x] The issue properly describes why the new feature is needed, what's the goal, and how we are building it. Any discussion
- [x] I have updated the documentation.

### DIFF
--- a/ingestion/src/metadata/clients/aws_client.py
+++ b/ingestion/src/metadata/clients/aws_client.py
@@ -34,6 +34,7 @@ class AWSServices(Enum):
     KINESIS = "kinesis"
     QUICKSIGHT = "quicksight"
     ATHENA = "athena"
+    RDS = "rds"
 
 
 class AWSAssumeRoleException(Exception):
@@ -181,3 +182,6 @@ class AWSClient:
 
     def get_athena_client(self):
         return self.get_client(AWSServices.ATHENA.value)
+
+    def get_rds_client(self):
+        return self.get_client(AWSServices.RDS.value)
--- a/ingestion/src/metadata/ingestion/source/database/mysql/connection.py
+++ b/ingestion/src/metadata/ingestion/source/database/mysql/connection.py
@@ -16,12 +16,18 @@ from typing import Optional
 
 from sqlalchemy.engine import Engine
 
+from metadata.clients.aws_client import AWSClient
+
 from metadata.generated.schema.entity.automations.workflow import (
     Workflow as AutomationWorkflow,
 )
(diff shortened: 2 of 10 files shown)


pmbrull:
Thanks for popping this up @tutte. Answered you via mail, but just wanted to add here the relevant context with my thoughts on the change:

In terms of how we could shape the feature, I imagine something similar to how we handle Airflow [connections](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/pipeline/airflowConnection.json#L35). We could have a top-level `connection` property which is then referenced either by a `[Basic Auth](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/security/credentials/basicAuth.json)` with the typical user/pwd, or `awsConfig`. Then, based on the class types we receive from `connection`, we can create the connection URL one way or another. The good thing about the `oneOf` setting in the JSON Schemas, is that users can just select which type of auth they want, and they don't see all the details from other refs.

This change would then require a couple of things:
- Migration scripts update: Basically, we store the service entities as JSON in the database. If we change the shape of the connection, we need to update the stored data to reflect these changes. We use flyway for this. [Example](https://github.com/open-metadata/OpenMetadata/blob/main/bootstrap/sql/com.mysql.cj.jdbc.Driver/v009__create_db_connection_info.sql).
- Validate that the credentials are properly masked on the database. Otherwise, we might need to patch some things in the backend.

harshach:
@tutte is this ready to be reviewed. Let us know since the PR is in draft mode. Thanks

ayush-shah:
Hello @tutte, there are json schema changes that will be needed here..let me know if you are taking care of the same..
we have an issue https://github.com/open-metadata/OpenMetadata/issues/11068, I am starting with postgres related changes.

ayush-shah:
cc: @open-metadata/ingestion

ayush-shah:
Closing this PR as no response from @tutte.
we can find the similar PR that fixes #11068 here: https://github.com/open-metadata/OpenMetadata/pull/11937

PR #11169, opened 2023-04-20
Issue 10805 redshift e2e

TeddyCr:
E2E Tests for redshift.


pmbrull (dismissed):

nahuelverdugo (changes requested):
We have to update https://github.com/open-metadata/OpenMetadata/blob/main/.github/workflows/py-cli-e2e-tests.yml. We need to add the env vars and the test we want to run.

akash-jain-10 (approved):

nahuelverdugo (approved):

Issue #10764, opened 2023-03-25
Include `serviceType` in list Ingestion Pipeline API 


ulixius9:
**Affected module**
backend

**Describe the bug**
I've one db service and one messaging service with same name the `test`  the pipeline `test_metadata_uRv0fTRY`  belongs to redpanda and `test_metadata_uTxJXnzz`  belongs to mysql but still i can see both of these pipelines in both of the services! 

![image](https://user-images.githubusercontent.com/39544459/227712918-cc338bc6-9b40-4dce-a5e5-183e7347fa3f.png)
![image](https://user-images.githubusercontent.com/39544459/227712924-0584a5c2-bbd8-41e3-9dc4-b58fa19defaf.png)

The API call to get these Ingestion pipeline looks like this `http://localhost:8585/api/v1/services/ingestionPipelines?fields=owner,pipelineStatuses&service=test` if it had a option to pass service type like `messageingService` or `databaseService` this can be resolved

PR #11295, opened 2023-04-26
Ingestion pipeline by service type

07Himank:
I worked on ... because ...Fetching the ingestion pipeline by service Type.

Fix : #10764

- [ x] Bug fix


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -1827,6 +1828,102 @@ default Class<IngestionPipeline> getEntityClass() {
     default String getNameColumn() {
       return "fullyQualifiedName";
     }
+
+    @Override
+    default int listCount(ListFilter filter) {
```
mohityadav766, line 1833: we can move this to CollectionDAO, listCount , list after can have a different behaviour based on the query param, check for queryEntitiy what we did with "entityId"

mohityadav766 (approved):

PR #11363, opened 2023-04-28
Centralize SSL Schemas & Add SSL for Airflow REST

pmbrull:
### Describe your changes:

- The OpenMetadata Connection had a "custom" SSL JSON Schema. Removed that in favor of the SSL config used for the ingestion as well. This requires changes in the JSON Schemas + ometa_client
- We did not have any way to set SSL for the Airflow REST client. Reused the same approach as with ES
- Checked the setup with Airflow locally and added docs

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [x] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/util/ElasticSearchClientUtils.java
```diff
@@ -72,25 +61,14 @@ public static RestHighLevelClient createElasticSearchClient(ElasticSearchConfigu
...
-      } catch (IOException
-          | NoSuchAlgorithmException
-          | CertificateException
-          | KeyStoreException
-          | KeyManagementException e) {
-        throw new RuntimeException("Failed to crete SSLContext to for ElasticSearch Client", e);
-      }
+    if (elasticSearchConfiguration.getScheme().equals("https")) {
+      return SSLUtil.createSSLContext(
+          elasticSearchConfiguration.getTruststorePath(),
```
harshach, line 69: @pmbrull you can use https without providing the trust store. If they do this will fail. Please revert to the older code.

harshach (approved):

Issue #11142, opened 2023-04-19
Table PUT API fails when child column contains `.` in name


ulixius9:
**Affected module**
Backend

**Describe the bug**
When creating a table with put api with example request as follows 

```
{
  "columns": [
    {
      "children": [
        {
            "dataType": "NUMBER",
            "name": "child.test.hello"
        }
      ],
      "dataType": "NUMBER",
      "name": "parent"
    }
  ],
  "databaseSchema": "local_mysql.default.openmetadata_db",
  "name": "test_ss_s",
  "tableType": "Regular"
}
``` 

if table did not exist it gets created successfully but when trying to hit the same api again (basically trying to update the table)

```
{
  "code": 400,
  "message": "Invalid name parent.\"child.test.hello\""
}
```

harshach:
@ulixius9 you need to escape the quotes just like we are doing it for the table names

ulixius9:
@harshach I'm not sure if I follow that, I'm not passing quotes in the name instead passing `.` in the name 

and this only happens when there is a `.` in name of child column and works if there is a `.` in the name of parent or top column 


https://user-images.githubusercontent.com/39544459/233543183-4e4d2f79-a185-4308-aff9-3e3c443a119b.mov

PR #11392, opened 2023-05-02
fixed issue for column children name

07Himank:
I worked on ... because ...fix issue with children column name containing "." .

Fix : #11142
<!-- Bug fix
- [ x] I have added a test that covers the exact scenario we are fixing. For complex issues, comment the issue number in the test for future reference.


harshach (approved):

openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java
```diff
@@ -347,9 +347,9 @@ public static String getFieldName(String... strings) {
   /** Return column field name of format "columns".columnName.columnFieldName */
   public static <T extends EntityInterface> String getColumnField(
       T entityWithColumns, Column column, String columnField) {
+
     // Remove table FQN from column FQN to get the local name
-    String localColumnName =
-        EntityUtil.getLocalColumnName(entityWithColumns.getFullyQualifiedName(), column.getFullyQualifiedName());
+    String localColumnName = column.getName();
```
07Himank, line 352: Hey @sureshms ,
I have a small doubt here why are we using the getLocalColumnName function to get the column name instead we can directly do by column.getName().
as when we are using that function for column children it return the name as parentName."childerName" which breaks further for children.

PR #11394, opened 2023-05-02
Fix geometry type for postgres

ulixius9:
### Describe your changes:

Fix geometry type for postgres

Postgres fails for type geometry with error `GEOMETRY() takes no arguments`, which is an error related to library itself when creating table with geometry column like this `geometry(Geometry,4326),` it passed the `Geometry,4326` params in the args of `GEOMETRY` which is unexpected hence the error, with this pr we are overriding the `_get_column_info` method which is responsible for determining the type of column and explicitly setting the args to empty to avoid such errors

slack support thread: https://openmetadata.slack.com/archives/C02B6955S4S/p1682698118936489

#
### Type of change:

- [x] Bug fix

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/postgres/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/postgres/metadata.py
@@ -17,9 +17,7 @@ from typing import Iterable, Tuple
 
 from sqlalchemy import sql
 from sqlalchemy.dialects.postgresql.base import PGDialect, ischema_names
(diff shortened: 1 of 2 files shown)


TeddyCr (approved):
LGTM. @ulixius9 before merging, can you take a look at the security hotspot and review it (you'll need to be logged into Sonar). If you can add a comment on the review that would be awesome.

Screenshot 2023-05-03 at 2 22 43 PM

PR #11463, opened 2023-05-05
Bootstrap users fix + Ldap Fix error message

mohityadav766:
### Describe your changes:

- Reorganize Bootstrap for clear admin User startup
- Ldap Message fix in case invalid binding 
- Added admin user cleanup when migrating between SSO and basic auth

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/util/UserUtil.java
```diff
@@ -254,4 +260,16 @@ public static List<EntityReference> getRoleForBot(String botName) {
...
     return listOf(RoleResource.getRole(botRole));
   }
+
+  public static void cleanUpBasicAuth() {
+    // Delete Admin UserName
+    UserRepository userRepository = (UserRepository) Entity.getEntityRepository(Entity.USER);
+    try {
+      // Fetch Original User, is available
+      User originalUser = userRepository.getByName(null, ADMIN_USER_NAME, new EntityUtil.Fields(List.of("id,name")));
+      userRepository.delete(ADMIN_USER_NAME, originalUser.getId(), true, true);
```
harshach, line 270: this is problematic if the admin user created a service. We need to clean up these relationships as well. Same goes when we delete the user we should just clean up the relations and keep the entities such user owns and follows should go away if the user is getting hard deleted.

mohityadav766: fixed it ...instead of deletion we use an empty auth mechanism for non SSO and basic auth will have valid populated auth mechanism

harshach (approved):

PR #11473, opened 2023-05-07
Fix Dynamodb connection

ayush-shah:
### Describe your changes:

Fixes Dynamodb listtables check
With valid Session token
image

with invalid session token
image

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


OnkarVO7 (dismissed):

ingestion/src/metadata/ingestion/source/database/dynamodb/connection.py
```diff
@@ -32,6 +33,14 @@ def get_connection(connection: DynamoDBConnection):
     return AWSClient(connection.awsConfig).get_dynamo_client()
 
 
+def check_list_tables(client):
+    """
+    Test ListTables under dynamodb
+    """
+    for _ in client.tables.all():
+        return True
```
nahuelverdugo, line 41: What are we trying to test here, if that call fails or not? Maybe something like this:
```suggestion
def check_list_tables(client) -> bool:
    """
    Test ListTables under dynamodb
    """
    try:
        client.tables.all()
    except Exception:
        return False
    return True
```

openmetadata-service/src/main/resources/json/data/testConnections/database/dynamodb.json
```diff
@@ -1,15 +1,15 @@
...
-    ]
-  }
\ No newline at end of file
+  "name": "DynamoDB",
+  "fullyQualifiedName": "DynamoDB",
+  "displayName": "DynamoDB Test Connection",
+  "description": "This Test Connection validates the access against the database and basic metadata extraction of schemas and tables.",
+  "steps": [
+    {
+      "name": "ListTables",
```
nahuelverdugo, line 8: Not 100% sure, do we need to migrate this or not?

ingestion/src/metadata/ingestion/source/database/dynamodb/connection.py
ayush-shah: so client.tables.all() passes even with wrong token, but when we run it as an iterator, it fails giving us a proper error.

openmetadata-service/src/main/resources/json/data/testConnections/database/dynamodb.json
ayush-shah: we will need it

nahuelverdugo (dismissed):

OnkarVO7 (approved):

OnkarVO7 (approved):

ingestion/src/metadata/ingestion/source/database/dynamodb/connection.py
pmbrull: @ayush-shah then IMO we should have this as a comment in the method

ingestion/src/metadata/ingestion/source/database/dynamodb/connection.py
```diff
@@ -32,6 +33,14 @@ def get_connection(connection: DynamoDBConnection):
     return AWSClient(connection.awsConfig).get_dynamo_client()
 
 
+def check_list_tables(client):
+    """
+    Test ListTables under dynamodb
```
pmbrull, line 38:
```suggestion
    Test ListTables under dynamodb
    
    `client.tables.all()` passes even with the wrong token, but when we run it as an iterator,
    it fails and gives us a proper error.
```

pmbrull (approved):

PR #11481, opened 2023-05-08
Fixed#11368: Activity feeds are not recorded for actions against creation of sub-terms in glossary.

### LINKED ISSUE #11368: Activity feeds are not recorded for actions against creation of sub-terms in glossary
**Affected module**
UI

**Describe the bug**
Activity feeds should show all the creation, maintenance,  deletion activities in OpenMetadata as performed by the user in the UI.

In the below user case, DG policy term was created , along with a few subterms. 

Only creation of parent term is visible. Creation of sub-terms is not appearing in the activity feed.
Later, the desc of one of the sub-term was edited and same is appearing in the activity feed i.e. apparently maintenance activity is recorded but not the creation of sub-terms.

image

image

use case 2: Policy is created, but same is not appearing in activity feed.

image

**
image

use case 3 : role updation against user is not being captured in activity feed.

image
image

**
version 
13.3

parthp2107:
### Describe your changes:

Fixes #11368 

#
### Type of change:

- [x] Improvement
- [x] New feature

#
### Checklist:

#
### Reviewers
@harshach @mohityadav766


openmetadata-service/src/main/java/org/openmetadata/service/events/ChangeEventHandler.java
```diff
@@ -243,6 +245,13 @@ private List<Thread> getThreads(ContainerResponseContext responseContext, String
     }
 
     if (entityInterface.getChangeDescription() == null) {
+      if (entity instanceof GlossaryTerm || entity instanceof Glossary) {
```
mohityadav766, line 248: @parthp2107 are we not getting any change event for this with ENTITY_CREATED ??

parthp2107: Tried:
```
String changeType = responseContext.getHeaderString(RestUtil.CHANGE_CUSTOM_HEADER);
if (RestUtil.ENTITY_CREATED.equals(changeType))
```
Getting changeType as null.

parthp2107: Screenshot 2023-05-08 at 6 14 01 PM

harshach: @parthp2107 lets check why we are not getting entity created change events thats the root issue

parthp2107: Addressed in the latest commit.

mohityadav766 (approved):

openmetadata-service/src/main/java/org/openmetadata/service/events/ChangeEventHandler.java
```diff
@@ -242,6 +242,17 @@ private List<Thread> getThreads(ContainerResponseContext responseContext, String
       return List.of(thread);
     }
 
+    int responseCode = responseContext.getStatus();
```
harshach, line 245: 1. there are too many if conditions which can easily turned into switch or if else conditions
2. Wouldn't this add too many entity created , especially incase of ingestion, we are filtering on ingestion-bot so hopefully this shouldn't be an issue cc @mohityadav766 
3.  can't we keep the if condition simple like this RestUtil.ENTITY_SOFT_DELETED.equals(changeType)? instead of adding responseCode and not sure why we are checking  !responseContext.getEntity().getClass().equals(Thread.class)) . This should be filtered above itself no?

PR #11504, opened 2023-05-09
fix(ui) : restrict global effect css of manage dropdown button

Ashish8689:
### Describe your changes:

- I worked on restricting global css effect of manage dropdown button

image
image

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/common/entityPageInfo/ManageButton/ManageButton.less
+++ b/openmetadata-ui/src/main/resources/ui/src/components/common/entityPageInfo/ManageButton/ManageButton.less
@@ -51,13 +51,15 @@
   background-color: @manage-button-bg-primary;
 }
 
-.ant-dropdown-menu {
-  padding: 8px 0px;
-  background: @white;
-  border: 1px solid #dde3ea;
-  box-shadow: 1px 1px 3px rgb(0 0 0 / 12%);
-  border-radius: 4px;
-  .ant-dropdown-menu-item {
-    padding: 5px 16px;
+.manage-dropdown-list-container {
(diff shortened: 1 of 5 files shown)


openmetadata-ui/src/main/resources/ui/src/components/common/entityPageInfo/ManageButton/ManageButton.less
```diff
@@ -51,13 +51,15 @@
...
+    padding: 8px 0px;
+    background: @white;
+    border: 1px solid #dde3ea;
+    box-shadow: 1px 1px 3px rgb(0 0 0 / 12%);
+    border-radius: 4px;
+    .ant-dropdown-menu-item {
+      padding: 5px 16px;
+    }
   }
 }
```
ShaileshParmar11, line 65: instead of repeating style, create one dropdown.style.less in style/component folder and add this CSS in that file with some meaning full name like custom-dropdown-border and use that class everywhere, so in that way if in future we need same style to other place we can have it with single class.

ShaileshParmar11 (commented):
one comment

openmetadata-ui/src/main/resources/ui/src/components/common/entityPageInfo/ManageButton/ManageButton.less
Ashish8689: This is currently intended only for the manage button, we will going to apply this  across the app in the further release.

chirag-madlani (approved):

PR #11508, opened 2023-05-09
Fix #7712: Align AuthProvider values in the JSON schema definition

### LINKED ISSUE #7712: Align AuthProvider values in the JSON schema definition
**Issue**
The auth provider value is currently defined with different types in the JSON schema definitions:
-  `openmetadata-spec/src/main/resources/json/schema/entity/services/connections/metadata/openMetadataConnection.json` as enum
- `openmetadata-spec/src/main/resources/json/schema/configuration/airflowConfiguration.json` as string
- `openmetadata-spec/src/main/resources/json/schema/configuration/authenticationConfiguration.json` as string

Therefore, we are adding unnecessary transformations throughout the code. Like this:

```java
OpenMetadataServerConnection.AuthProvider.fromValue(
            openMetadataApplicationConfig.getAuthenticationConfiguration().getProvider())
```

**Solution**
Define an enum for all of them so they can be reused.

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


zhenyu:
@pmbrull would you please tell me when will you merge this? Thanks

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/metadata/openMetadataConnection.json
```diff
@@ -18,6 +18,24 @@
       "additionalProperties": {
         "type": "string"
       }
+    },
+    "authProvider": {
```
mohityadav766, line 22: we should also use this reference in ssoServiceType in ssoAuth.json , because that list is same as this with only difference of "openmetadata"

zhenyu:
@harshach , any estimate day for this PR to be merged? Anything I can help to boost the merge?

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/metadata/openMetadataConnection.json
harshach: @mohityadav766 our goal is to remove all the authentication providers for OM 
UI can ssoType for enabling authentication to SSO. But OM API auth will only support the JWT token auth. So we can keep these separate and remove all the references to other providers except for "openmetadata" ideally we should rename to JWTToken.

PR #11527, opened 2023-05-10
Fixes #11526 - Updated Entity should be stored for the case when enti…

### LINKED ISSUE #11526: Updated Entity should be stored for the case when entity has changed without version change
**Describe the bug**
When there is small changes or automatic updates to entity (like change in size of entity), the updated content should be stored without changing the version.

pdeeparao:
…ty has changed without version change

### Describe your changes:

Fixes #11526

I worked on ... because ...
When there is small changes or automatic updates to entity (like change in size of entity) that does not cause version update, the entity is not stored. 

#
### Type of change:

- [ x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
@@ -1370,6 +1370,7 @@ public abstract class EntityRepository<T extends EntityInterface> {
     protected boolean majorVersionChange = false;
     protected final User updatingUser;
     private boolean entityRestored = false;
+    private boolean entityChanged = false;
 
     public EntityUpdater(T original, T updated, Operation operation) {
       this.original = original;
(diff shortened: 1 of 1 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -1551,34 +1552,58 @@ public boolean isEntityRestored() {
...
     }
 
     public final <K> boolean recordChange(
         String field, K orig, K updated, boolean jsonValue, BiPredicate<K, K> typeMatch)
         throws JsonProcessingException {
+      return recordChangeInternal(field, orig, updated, jsonValue, typeMatch, true);
+    }
+
+    private <K> boolean recordChangeInternal(
+        String field, K orig, K updated, boolean jsonValue, BiPredicate<K, K> typeMatch, boolean addToVersion)
```
sureshms, line 1577: Instead of addToVersion, is updateVersion a better name?

sureshms:
This is no longer required

PR #11602, opened 2023-05-15
Fix #11548 - Secrets Managers comms with OMeta

### LINKED ISSUE #11548: Secrets Managers comms with OMeta
## Currently Happening
- The `openMetadataServerConnection` property in the Ingestion Pipeline (built dynamically) has the `secretsManagerCredentials` property.
- This makes it look like the server is sending the SM keys to the client
- However, in Airflow, the client uses the `build_secrets_manager_credentials` method to update the Ingestion Pipeline
- Then, when the OMeta client is instantiated, it picks up this info to build the SecretsManagerFactory

## Issues
- This approach is misleading since it makes it look like the server sends the SM creds to the client. However, that info is instantiated only in the client
- We are “patching” a class we receive from the server in order to create the SecretsManagerFactory

Basically, we are misusing classes and patching information when how the client gets SM creds should be isolated.

## New Approach
- The `openMetadataServerConnection` should not have a `secretsManagerCredentials` at all
- In the client, the SecretsManagerFactory should be the one knowing how to load the credentials from the environment
    - One approach for Airflow
    - Another for Argo
    - Whatever else is needed
- The IngestionPipeline should say to the client (within the `openMetadataServerConnection`):
    - The SM provider being used, which is already there as `secretsManagerProvider`
    - The environment in which it is running in order for it to load the creds properly. We can create a new `secretsManagerEnvironment`, which can be `Airflow`, if we read creds from `airflow.cfg`, `Env`, if we read from env vars, etc.

This will require to clean up the `workflowConfig` of the YAMLs (and docs https://docs.open-metadata.org/v1.0.0/deployment/secrets-manager/supported-implementations/aws-ssm-parameter-store#cli)

We’ll need to issue a notice when this gets released if people are using the SM when building their own DAGs,

pmbrull:
### Describe your changes:

Fixes #11548

Updating the docs for 1.1.0 here https://github.com/open-metadata/OpenMetadata/pull/11604

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


TeddyCr (approved):
LGTM

ayush-shah (approved):

PR #11650, opened 2023-05-18
Logged the Email Messages which are not sent as smpt settings is not …

07Himank:
I worked on ... because ...

Fix : #11017
#
### Type of change:

- [ x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/util/EmailUtil.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/util/EmailUtil.java
@@ -146,6 +146,10 @@ public class EmailUtil {
           user.getEmail(),
           EMAIL_TEMPLATE_BASEPATH,
           ACCOUNT_STATUS_TEMPLATE_FILE);
+    } else {
+      LOG.warn(
+          String.format(
+              "action:- %s was not sent to user:- %s as SMTP setting is not enabled", action, user.getName()));
     }
   }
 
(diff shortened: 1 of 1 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/util/EmailUtil.java
```diff
@@ -146,6 +146,10 @@ public void sendAccountStatus(User user, String action, String status) throws IO
           user.getEmail(),
           EMAIL_TEMPLATE_BASEPATH,
           ACCOUNT_STATUS_TEMPLATE_FILE);
+    } else {
+      LOG.warn(
+          String.format(
+              "action:- %s was not sent to user:- %s as SMTP setting is not enabled", action, user.getName()));
```
mohityadav766, line 152: Create a constant and use that

mohityadav766 (changes requested):
use a static String

openmetadata-service/src/main/java/org/openmetadata/service/util/EmailUtil.java
07Himank: handled in latest commit

07Himank:
> use a static String

handled in latest commit

mohityadav766 (approved):

PR #11653, opened 2023-05-18
Clean NER Scanner imports

pmbrull:
### Describe your changes:

Which was forcing the presidio package import out of scope. I broke this when trying to have the `NERScanner` with the `scan` as a classmethod. However, in this case it makes sense to properly handle the class state and initialize the analyzer only once in the init.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/pii/ner_scanner.py
+++ b/ingestion/src/metadata/pii/ner_scanner.py
@@ -48,20 +48,23 @@ class NERScanner:
     Based on https://microsoft.github.io/presidio/
     """
 
-    import spacy
-    from presidio_analyzer import AnalyzerEngine
-    from presidio_analyzer.nlp_engine.spacy_nlp_engine import SpacyNlpEngine
+    def __init__(self):
+        import spacy
+        from presidio_analyzer import AnalyzerEngine
+        from presidio_analyzer.nlp_engine.spacy_nlp_engine import SpacyNlpEngine
 
-    try:
-        spacy.load(SPACY_EN_MODEL)
-    except OSError:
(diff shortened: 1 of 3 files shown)


ayush-shah (approved):

ingestion/src/metadata/pii/ner_scanner.py
```diff
@@ -48,20 +48,23 @@ class NERScanner:
...
-        from spacy.cli import download
+        try:
+            spacy.load(SPACY_EN_MODEL)
+        except OSError:
+            logger.warning("Downloading en_core_web_md language model for the spaCy")
+            from spacy.cli import download
 
-        download(SPACY_EN_MODEL)
-        spacy.load(SPACY_EN_MODEL)
+            download(SPACY_EN_MODEL)
```
OnkarVO7, line 62: Here if the download fails for some reason, the entire workflow will fail. Might be better to handle the exception

OnkarVO7 (changes requested):

ingestion/src/metadata/pii/ner_scanner.py
pmbrull: I think I am ok on making such errors super evident. Otherwise it might be confusing to have silent download failures and just skip the auto PII. If the errors persists or cannot be fixed for whatever reason, users can disable the tagging and move on.

wdyt?

OnkarVO7: Ok it should be fine then if the user can disable it if it fails 👍

OnkarVO7 (approved):

PR #11702, opened 2023-05-22
Fix redshift view lineage parsing issue

ulixius9:
### Describe your changes:

Fix redshift view lineage parsing issue

similar to https://github.com/open-metadata/OpenMetadata/pull/11689

part of fix for #7427

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/database/redshift/metadata.py
```diff
@@ -50,131 +39,36 @@
...
-    with connection.connect() as cnct:
-        result = cnct.execute(
-            REDSHIFT_GET_SCHEMA_COLUMN_INFO.format(schema_clause=schema_clause)
-        )
-        for col in result:
-            key = RelationKey(col.table_name, col.schema, connection)
-            all_columns[key].append(col)
-    return dict(all_columns)
+STANDARD_TABLE_TYPES = {
+    "r": TableType.Local,
```
TeddyCr, line 68: nit. should that be an `l` (for Local)?

ingestion/src/metadata/ingestion/source/database/redshift/queries.py
```diff
@@ -229,3 +229,48 @@
...
+        null AS "diststyle",
+        s.esowner AS "owner_id",
+        u.usename AS "owner_name",
+        null AS "view_definition",
+        null AS "privileges"
+    FROM
+        svv_external_tables t
+        JOIN svv_external_schemas s ON s.schemaname = t.schemaname
+        JOIN pg_catalog.pg_user u ON u.usesysid = s.esowner
+    where 1 {schema_clause} {table_clause}
```
TeddyCr, line 274: We should avoid formatting queries this way - this is vulnerable to SQL injection. Instead, we should parametrize the queries -- see the screenshot for an example. There might be a low probability a user will actually be able to inject any kind of SQL we should still have this air tight.

https://github.com/open-metadata/OpenMetadata/assets/13626425/41be5ec4-2858-4b91-a372-a295678a4875

ulixius9: @TeddyCr regarding this we actually want to inject the sql in the query. [this](https://github.com/open-metadata/OpenMetadata/blob/e8c588e1f79d738fe3e5669ec8f54ead8bd25112/ingestion/src/metadata/ingestion/source/database/redshift/utils.py#L380) is where the query is being called. we are adding the condition of table and schema only when we have the information available otherwise we fetch all the tables.

ulixius9: so I actually tried the same thing and it gets translated  to this `where 1 'schema = ''public''' ''` where the condition is within quotes and throws a syntax error

TeddyCr: Oh my bad, did not see we were passing an identifier. Make sense. We cannot parametrize identifier from what I understand.

TeddyCr (approved):
LGTM. Could you just review the hotspot and add a comment in Sonar if you manually resolve it

PR #11762, opened 2023-05-25
Maven GitHub

Anuj359:
Updating the Github actions for maven publish! 
This should be triggered in the following scenarios -

- pull request merge on main branch only
- Release publish
- On demand workflow dispatch from any branch


harshach (dismissed):

.github/workflows/publish-maven-package.yml
```diff
@@ -17,12 +17,18 @@ on:
   pull_request_target:
     types:
       - closed
+    branches:
```
akash-jain-10, line 20: Instead of this, wouldn't `push` event on main branch make more sense ?

ref - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

akash-jain-10 (approved):

PR #11837, opened 2023-05-31
Fix Looker explore git link & Add BitBucket reader

pmbrull:
### Describe your changes:

- We were using the `source_file` property from the explore, which does not bring the right info. We are now picking up the `lookml_link` to get the path from the git repo.
- We added the BitBucket reader to pick up source files from BitBucket
- We can have multiple projects in looker, each of them pointing to a different github repo. We are now taking care of this by building one parser for each repo and picking up the parser dynamically based on the project git remote url
- Add version validation as a looker test step
- Add some more unit tests
- Also, the Spline connector was missing the `init` file. Without it, pylint does not pick it up. After adding it there were some pending linting issues fixed here

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/dashboard/looker/connection.py
```diff
@@ -61,8 +61,17 @@ def list_datamodels_test():
         """
         assert client.all_lookml_models(limit=1)
 
+    def validate_api_version():
+        """
+        Make sure we get a True
+        """
+        assert "4.0" in (
```
TeddyCr, line 68: does it have to be 4.0 specifically or it could be 4.1, etc.

TeddyCr (approved):

ingestion/src/metadata/ingestion/source/dashboard/looker/connection.py
pmbrull: we are using the sdk for 4.0 so feels safe to stick with this. If they update it we might need to change as well. https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/dashboard/looker/connection.py#L44

Issue #11758, opened 2023-05-25
[ESUpgrade] Upgrade ES and OpenSearch Version Support


mohityadav766:
**Is your feature request related to a problem? Please describe.**

Currently, we have to use the 7.0 version of ES Client since that is only supported by both ES and OpenSearch.
The following is the issue with the version:-
1.  ES Client has some Vulnerability issues, and upgrading it to 7.14 break the support for the higher Open search version.
2. Also since there are incompatibilities b/w the ES and OS clients and cross-compatibility is not supported we need to find a way to support both from OM.

**Describe the solution you'd like**
- The idea is to find the list of functionalities from the OM and provide a singular interface within OM, with two implementations one for OS and one for ES.
- Provide a configuration in YAML to allow users to configure OS or ES

**Describe alternatives you've considered**

**Additional context**

**Related issues**
- ...

PR #11879, opened 2023-06-05
Open search elastic search client task

07Himank:
I worked on ... because ...as we update Elastic search client above 7.13, open search client support is not compatibly,
added two diffrent clients for open search and elastic search.

- [ x] Bug fix
- [x ] Improvement
- [ x] New feature

fix : #11758


openmetadata-service/src/main/java/org/openmetadata/service/search/open/OpenSearchDataInsightProcessor.java
```diff
@@ -0,0 +1,85 @@
+package org.openmetadata.service.search.open;
```
harshach, line 1: lets call Open -> OpenSearch

07Himank: OKay will take care of it

PR #11919, opened 2023-06-07
Fix #10542 & #10830 - Mask Sample Data PII & Column Metrics

### LINKED ISSUE #10542: Mask PII sample data and don't return column metrics

### LINKED ISSUE #10830: Mask test cases and profiling data of columns tagged with 'PII.Sensitive'
**Description**
This issue is an enhancement of #10542 to mask other Entities related to a column tagged with `PII.Sensitive`.

**Requirements**
Open to discussion before setting them. Questions raised when a column is tagged with `PII.Sensitive`:
- Should we hide test cases of it?
- Should we hide profiling data of it? If yes, would it depend on the metric? 

About the last question, if profiling data could help infer a column's value, it should not be an issue. For example, in a salary column field, if we get the max and min values from it, we cannot associate the value with any other field.

pmbrull:
### Describe your changes:

Fix #10542 & fix #10830

We are now masking:
- Sample Data (GET)
- Column Profile (GET)
- Test Cases (LIST)

We allow admins and owners and mask the results for the rest of users.

Allowed view:
image

Masked view:
image

Allowed view:
image

Masked view:
image

Allowed view:
image

Masked view (we will set to null test results, test parameters and test description)
image

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/security/DefaultAuthorizer.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/security/DefaultAuthorizer.java
@@ -25,6 +25,7 @@ import org.openmetadata.schema.type.ResourcePermission;
 import org.openmetadata.service.OpenMetadataApplicationConfig;
 import org.openmetadata.service.security.policyevaluator.OperationContext;
 import org.openmetadata.service.security.policyevaluator.PolicyEvaluator;
+import org.openmetadata.service.security.policyevaluator.ResourceContext;
 import org.openmetadata.service.security.policyevaluator.ResourceContextInterface;
 import org.openmetadata.service.security.policyevaluator.SubjectCache;
 import org.openmetadata.service.security.policyevaluator.SubjectContext;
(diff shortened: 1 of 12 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/security/DefaultAuthorizer.java
```diff
@@ -97,6 +98,13 @@ public boolean shouldMaskPasswords(SecurityContext securityContext) {
     return !subjectContext.isBot();
   }
 
+  /** In 1.2, evaluate policies here instead of just checking the subject */
+  @Override
+  public boolean authorizePII(SecurityContext securityContext, EntityReference owner) {
+    SubjectContext subjectContext = getSubjectContext(securityContext);
+    return subjectContext.isAdmin() || subjectContext.isBot() || subjectContext.isOwner(owner);
```
mohityadav766, line 105: are we only allowing admin /bots /owner to view the Sample Data?

pmbrull: Everyone can view Sample Data unless there are columns flagged as PII.Sensitive. In that case, only admins/bots or the table owner(s) can view those specific columns.

On 1.2, we'll improve this by adding policy management

harshach (approved):

PR #11931, opened 2023-06-08
[1.0.4 port] - Update Test Connection Definition FQN #11861

pmbrull:
### Describe your changes:

1.0.4 release specific port of  #11861

This PR has 4 less files modified than the original PR since on `main` we have two more connectors: SAP Hana & Spline, which count for 4 files for their `connection.py` and test connection def JSON.

Also adding a version fix for the `deltalake` that is part of another big PR on `main` and cannot be cherry-picked

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

PR #11932, opened 2023-06-08
Align bootstrap after 1.0.4 patch

pmbrull:
### Describe your changes:
This PR https://github.com/open-metadata/OpenMetadata/pull/11931 on 1.0.4 adds a required query as `v012`. In order to keep the final 1.1 release synced with these migration files, we need to have the same `v012` script as on 1.0.4 and move the rest of the queries to `v013`

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach:
@pmbrull why are we create v013? we have v011 till 1.0.3, so the new v012 should take care of it https://github.com/open-metadata/OpenMetadata/tree/1.0.3/bootstrap/sql/com.mysql.cj.jdbc.Driver/

pmbrull:
> @pmbrull why are we create v013? we have v011 till 1.0.3, so the new v012 should take care of it https://github.com/open-metadata/OpenMetadata/tree/1.0.3/bootstrap/sql/com.mysql.cj.jdbc.Driver/

@harshach for the 1.0.4 tag fix we needed to move one query from v012 to the 1.0.4 release, but we don't want all the other queries. That's why we are moving a single query to v012 (which is what we already have on 1.0.4 branch) and creating a v013 for 1.1

harshach:
but that query will already applied right? why are we even moving it

ayush-shah (dismissed):

harshach (approved):

akash-jain-10 (approved):

PR #11948, opened 2023-06-09
dbt test parameter values fixed

OnkarVO7:
### Describe your changes:

dbt test parameter values fixed because of invalid datatype

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ayush-shah (approved):

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
```diff
@@ -917,7 +917,7 @@ def create_test_case_parameter_values(self, dbt_test):
         values = manifest_node.test_metadata.kwargs.get("values")
```
ayush-shah, line 917: maybe adding a typing List[str] be a better idea? wdyt? it will help clarify things in the below code

OnkarVO7: This `values` list can be of any type not necessarily str. So adding List[str] would not be the right thing here

PR #11964, opened 2023-06-12
Use python for Slack Link Monitoring

pmbrull:
### Describe your changes:

Since the current approach with `jq` is giving us some failures which look like false positives. Trying to set a more comprehensive approach with Python which will give us some more control and logging over what is happening.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/scripts/slack-link-monitor.py
+++ b/scripts/slack-link-monitor.py
@@ -1,0 +1,26 @@
+#  Copyright 2021 Collate
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  http://www.apache.org/licenses/LICENSE-2.0
+#  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


scripts/slack-link-monitor.py
```diff
@@ -0,0 +1,26 @@
...
+        "https://www.linkmonitor.dev/api/v1/validate",
+        headers={"Content-Type": "application/json"},
+        json={"url": "https://slack.open-metadata.org"},
+    )
+    if res.json().get("status") != "active":
+        raise RuntimeError("Expired status!")
+except RuntimeError as err:
+    raise err
+except Exception as err:
+    logging.error(f"Something went wrong fetching the data - [{err}]")
```
pmbrull, line 26: We will now only fail if the status is indeed different than `active`. Any other exception will get logged for us to analyze instead of directly getting a failure in Slack

akash-jain-10 (approved):

Issue #11979, opened 2023-06-13
matchAnyTag() is case insensitive


TeddyCr:
**Affected module**
User permission

**Describe the bug**
When we set a rule to matchTag it looks like our tag validation is case insensitive (e.g. `.PERSONAL` tag does not exist) but the permission is case sensitive (i.e. my user can't update the table with tag `.Personal`).

**To Reproduce**
Create a rule with the a `matchAnyTag()` condition with an upper case tag.

Screenshots or steps to reproduce

![Screenshot 2023-06-13 at 6 36 53 PM](https://github.com/open-metadata/OpenMetadata/assets/13626425/fdac8213-3430-4a96-8d46-09905fedee21)
![Screenshot 2023-06-13 at 6 35 00 PM](https://github.com/open-metadata/OpenMetadata/assets/13626425/060efc87-4c82-47f3-9256-035001d12e9f)

**Version:**
 - OpenMetadata version: `main`

07Himank:
Closing this ticket as it is already fixed fixed.
<img width="1137" alt="Screenshot 2023-08-07 at 6 49 06 PM" src="https://github.com/open-metadata/OpenMetadata/assets/112613760/7b2380d0-24b6-416f-b5ff-69ec39576b83">
<img width="1114" alt="Screenshot 2023-08-07 at 6 48 56 PM" src="https://github.com/open-metadata/OpenMetadata/assets/112613760/97f33e34-7c28-48b2-be76-b7d09424d749">

Issue #7098, opened 2022-08-31
API responses have consistent response independently of the DB used


nahuelverdugo:
**Issue** 
We support MySQL and PostgreSQL in the backend, but MySQL's `String` comparisons are case-insensitive.
When running a request like:

`http://localhost:8585/api/v1/testDefinition/name/COLUMNVALUEMAXTOBEBETWEEN` 

The OM server responds differently depending on the database.

With `PostgreSQL` as backend DB:
![image](https://user-images.githubusercontent.com/11074908/187745663-302ef7ad-8e0d-4b8a-9671-da1e76e622cb.png)

With `MySQL` as backend DB: 
![image](https://user-images.githubusercontent.com/11074908/187745735-aa98f7eb-1c92-4319-8354-cf6c0268581d.png)

We must ensure that queries in the backend return the same result. One example of a query that has this issue is the following one in the `EntityDAO` class:

```java
@SqlQuery("SELECT json FROM <table> WHERE <nameColumn> = :name <cond>")
String findByName(
    @Define("table") String table,
    @Define("nameColumn") String nameColumn,
    @Bind("name") String name,
    @Define("cond") String cond);
```

PR #12001, opened 2023-06-16
Mysql query case sensitive issue

07Himank:
I worked on ... because ...
I made the findByName mysql query case sensitive by using BINARY keyword.
which fixes the issues :- #11979 and #7098

### Type of change:

- [x ] Bug fix


harshach:
@07Himank this is lot more involved. Lets not merge it

07Himank:
> 



> @07Himank this is lot more involved. Lets not merge it

@harshach okay will not merge it.

PR #12038, opened 2023-06-19
GX up to 0.17+

VolkovGeoPhy:
### Describe your changes:
Update ingestion requirement great expectation to actual version
Fixes <issue-number>

I worked on ... because of keeping actual library versions

#
### Type of change:

- [!] Improvement

#
### Checklist:

- [!] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.


ingestion/setup.py
```diff
@@ -33,7 +33,7 @@ def get_long_description():
     "boto3": "boto3>=1.20,<2.0",  # No need to add botocore separately. It's a dep from boto3
     "geoalchemy2": "GeoAlchemy2~=0.12",
     "google-cloud-storage": "google-cloud-storage==1.43.0",
-    "great-expectations": "great-expectations~=0.16.0",
+    "great-expectations": "great-expectations>=0.17.0",
```
pmbrull, line 36: @TeddyCr should we keep this as `~=0.17`?

TeddyCr: Correct. cc @VolkovGeoPhy

TeddyCr (changes requested):
We should have it as `~=0.17`

ingestion/setup.py
VolkovGeoPhy: done

pmbrull (approved):

pmbrull:
oh actually it looks like python 3.7 cannot go further than 0.6.13:

```
python version: 0.16.14 Requires-Python >=3.8;
```

Since we are not ready to deprecate python 3.7 we should hold off on this change. Created the 3.7 issue and added your PR there so that we can move on when the deprecation is done https://github.com/open-metadata/OpenMetadata/issues/12082

Thanks

harshach:
@pmbrull since we deprecated 3.7, can we merge this in?

PR #12046, opened 2023-06-20
Add Pipeline Service Status Job to publish to Prometheus

pmbrull:
### Describe your changes:

Every 5 minutes we'll have a background Job running the `pipelineServiceClient.getStatus()` method and alerting us if we get any `unhealthy` status.

The Prometheus metric has the cluster name for us to create further alarms:

```
pipelineServiceClientStatus_counter_total{clusterName="openmetadata",status="unhealthy",} 23.0
```

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/clients/pipeline/airflow/AirflowRESTClient.java
```diff
@@ -264,7 +264,10 @@ public Response getServiceStatus() {
       }
 
     } catch (Exception e) {
-      throw PipelineServiceClientException.byMessage("Failed to get REST status.", e.getMessage());
+      Map<String, String> status =
+          buildUnhealthyStatus(String.format("Failed to get REST status due to [%s].", e.getMessage()));
+
+      return Response.ok(status, MediaType.APPLICATION_JSON_TYPE).build();
```
pmbrull, line 270: we were missing building the status response if Airflow was not available

mohityadav766 (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/events/scheduled/PipelineServiceStatusJob.java
```diff
@@ -0,0 +1,58 @@
...
+        (PipelineServiceClient)
+            jobExecutionContext.getJobDetail().getJobDataMap().get(JOB_CONTEXT_PIPELINE_SERVICE_CLIENT);
+    PrometheusMeterRegistry meterRegistry =
+        (PrometheusMeterRegistry) jobExecutionContext.getJobDetail().getJobDataMap().get(JOB_CONTEXT_METER_REGISTRY);
+    String clusterName = (String) jobExecutionContext.getJobDetail().getJobDataMap().get(JOB_CONTEXT_CLUSTER_NAME);
+    try {
+      registerStatusMetric(pipelineServiceClient, meterRegistry, clusterName);
+    } catch (Exception e) {
+      LOG.error("[Pipeline Service Status Job] Failed in sending metric due to", e);
+      throw new PipelineServiceStatusJobException(e);
```
harshach, line 39: do we want to throw a specific exception here? We should just log and forget about it.
Alos in catch we should register the metric saying the scheduler is not healthy

openmetadata-service/src/main/java/org/openmetadata/service/events/scheduled/PipelineServiceStatusJobHandler.java
```diff
@@ -0,0 +1,86 @@
...
+            .withIdentity(PIPELINE_SERVICE_STATUS_JOB, STATUS_GROUP)
+            .usingJobData(dataMap);
+
+    return jobBuilder.build();
+  }
+
+  private Trigger getTrigger() {
+    return TriggerBuilder.newTrigger()
+        .withIdentity(STATUS_CRON_TRIGGER, STATUS_GROUP)
+        .withSchedule(SimpleScheduleBuilder.simpleSchedule().withIntervalInSeconds(5 * 60).repeatForever())
```
harshach, line 73: configuring interval would be good instead of hardcoding.

openmetadata-service/src/main/java/org/openmetadata/service/events/scheduled/PipelineServiceStatusJob.java
```diff
@@ -0,0 +1,57 @@
...
+        (PipelineServiceClient)
+            jobExecutionContext.getJobDetail().getJobDataMap().get(JOB_CONTEXT_PIPELINE_SERVICE_CLIENT);
+    PrometheusMeterRegistry meterRegistry =
+        (PrometheusMeterRegistry) jobExecutionContext.getJobDetail().getJobDataMap().get(JOB_CONTEXT_METER_REGISTRY);
+    String clusterName = (String) jobExecutionContext.getJobDetail().getJobDataMap().get(JOB_CONTEXT_CLUSTER_NAME);
+    try {
+      registerStatusMetric(pipelineServiceClient, meterRegistry, clusterName);
+    } catch (Exception e) {
+      LOG.error("[Pipeline Service Status Job] Failed in sending metric due to", e);
+      publishUnhealthyCounter(meterRegistry, clusterName);
```
harshach, line 38: this may trigger false negatives. We should do an exponential backoff with 3 retries before calling it unhealthy.
Can be handled in another PR cc @pmbrull

harshach (approved):

openmetadata-service/src/main/java/org/openmetadata/service/events/scheduled/PipelineServiceStatusJob.java
pmbrull: thanks, will take care of this https://github.com/open-metadata/OpenMetadata/issues/12083

PR #12072, opened 2023-06-21
Fixes #11743 - Remove SQLParse dependency for System Metrics

### LINKED ISSUE #11743: Revisit snowflake system metrics performance
If we have HUGE `INSERT INTO` queries in the account history table for Snowflake, the profiler performance gets impacted and we end up with OOM.

The issue is we are using the vanilla `sqlparse.parse` for the query, instead of wrapping up the behavior into our own `LineageParser`.

Note that we are using sqlparse here instead of sqllineage since we need to handle DELETE statements, which are not flagged in sqllineage yet.

We should:
- For a first quick win, add a `@timeout` decorator to the sqlparse parsing process.
- For a long-term improvements, see if we can detect the DELETE statements in the sqllineage as well.

TeddyCr:
### Describe your changes:

Fixes #11743

- As SQLParse generated performance issues, we removed the existing dependency the system metrics had on it.
- Added integration tests for the system metrics (skipped by default) to be able to test changes in a consistent way

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


pmbrull (approved):
Damn I see you went through the regex rabbit hole 😅 Thanks for this, def faster than parsing for info that we don't need

pmbrull (approved):

PR #11960, opened 2023-06-12
Adding MD5 hash to fullyQualifiedName and names to simplify DB indexes for lookups and increasing the size of FQN

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

PR #12092, opened 2023-06-22
[1.1.1] - Bump size for FQN

### LINKED ISSUE #7680: Column name max length is too low
**Affected module**
Ingestion Framework

**Describe the bug**
While ingesting tables from Snowflake, we got a lot of warnings saying that the column name was too long. The maximum in Open Metadata seems to be 128 characters, but Snowflake column names [can be up to 255 characters long](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html#identifier-requirements).

Example error message (long column name replaced with `xyz`):
```
[2022-09-22 11:06:52] WARNING  {Ingestion:sql_column_handler:246} - Unexpected exception processing column [{'name': 'xyz', 'type': DECIMAL(precision=38, scale=0), 'nullable': True, 'default': None, 'autoincrement': False, 'comment': None, 'primary_key': False}]: 1 validation error for Column
name -> __root__
  ensure this value has at most 128 characters (type=value_error.any_str.max_length; limit_value=128)
```

**To Reproduce**
Ingest a Snowflake table with a column name longer than 128 characters.

**Expected behavior**
I expect Open Metadata to be able to ingest tables with long column names.

**Version:**
 - OS: MacOS
 - Python version: 3.9.9
 - OpenMetadata version: 0.12
 - OpenMetadata Ingestion package version: `openmetadata-ingestion[snowflake]==0.12.0.2`

**Additional context**

pmbrull:
### Describe your changes:

Since we landed the FQN hash PR #11960, we can now support longer FQNs in the API.

## Background
The process here between API <> Storage is as follows:
1. New Entity is created via the API
2. We set the FQN of the Entity based on its hierarchy (e.g., `service.db.schema.table`, or `parent1.parent2.container`).
3. When going to store the Entity:
    1. The FQN gets split, e.g., `service.db.schema.table` -> `service`, `db`, `schema`, `table`.
    2. Each unit is hashed
    3. The result is joined together and stored in its hashed form in the db, e.g., `hash1.hash2.hash3.hash4`.
4. When reading, we pick up the full FQN hash from the db, restore to its original name by pieces and join in the final FQN string.

## Changes
Where we have variable / long sizes in FQNs is:

- Tables (due to the Datalake connector, as the name is the path). Therefore this involves the FQNs of
    - Columns, built as `tableFQN.ColumnName`
    - Test Suites, built as `tableFQN.TestSuite`
- Containers, since it gets built by concatenating `parent.children` together.

### Container

For containers, we are setting the FQN size to 1024. This is why:
- We are using MD5 as the hashing algorithm. Each MD5 transformation of a string gives us a 32 characters string.
- MySQL max index size is 256 characters
- This means we can store up to 256 / 32=8 levels of hashes in the db
- EntityName is sized as 128 by default
- The final FQN max size that the API can return is then 8 * 128 = 1024

We will then just support up to 8 parent <> children levels.

### Table

We can follow a similar logic for table names. However, here it also makes sense to bump the `EntityName` since it will give us more room to ingest Datalake paths:

- Setting EntityName to 516
- Bumping FQN to 896 (bumped table name + schema + db + service)

Column FQN will be 1024 -> table FQN + column name (128)

### Test Suite

Test Suites FQN will be 1024 -> Table FQN + `.testSuite` and to keep round numbers

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-spec/src/main/resources/json/schema/entity/data/table.json
```diff
@@ -14,9 +14,21 @@
       "description": "Name of a table. Expected to be unique within a database.",
       "type": "string",
       "minLength": 1,
-      "maxLength": 128,
+      "maxLength": 512,
```
harshach, line 17: This is too big no?, max is 256 by snowflake right?

openmetadata-spec/src/main/resources/json/schema/entity/data/table.json
```diff
@@ -14,9 +14,21 @@
...
       "pattern": "^((?!::).)*$"
     },
+    "fullyQualifiedEntityName": {
+      "description": "A unique name that identifies a table. Example for table 'DatabaseService.Database.Schema.Table'.",
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 896
```
harshach, line 24: we should refer to common fullyQualifiedName in basic.json

openmetadata-spec/src/main/resources/json/schema/entity/data/table.json
pmbrull: did a check and 516 is super long and it would require us to allocate more space on the `name` field in the db. 256 looks like a good compromise and is aligned with the rest of the entities

openmetadata-spec/src/main/resources/json/schema/entity/data/table.json
pmbrull: as discussed, using the common FQN and removing the max limit size in there

openmetadata-spec/src/main/resources/json/schema/type/basic.json
```diff
@@ -103,10 +103,9 @@
       "pattern": "^(?U)[\\w'\\- .&()]+$"
     },
     "fullyQualifiedEntityName": {
-      "description": "A unique name that identifies an entity. Example for table 'DatabaseService:Database:Table'.",
+      "description": "A unique name that identifies an entity. Example for table 'DatabaseService.Database.Schema.Table'.",
       "type": "string",
-      "minLength": 1,
-      "maxLength": 256
+      "minLength": 1
     },
```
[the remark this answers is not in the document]
pmbrull: yes. I believe we discussed since we are hashing on the db we don't need to have a limit on the FQN. Just limiting the `entityName` should be enough

sureshms:
@pmbrull, can you also change the Column Name length from 128 to 256 and address #7680 also in this patch?

pmbrull:
> @pmbrull, can you also change the Column Name length from 128 to 256 and address #7680 also in this patch?

done, thanks

openmetadata-spec/src/main/resources/json/schema/type/basic.json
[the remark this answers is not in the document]
harshach: @pmbrull entityLink will run into issues here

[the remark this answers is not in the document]
pmbrull: should we then keep 1024?

PR #12105, opened 2023-06-22
Fix #12083 - Backoff retry when polling for pipeline service client status

### LINKED ISSUE #12083: Add exponential backoff in PipelineServiceStatusJob
if we get an exception when running `registerStatusMetric` to avoid false negatives

pmbrull:
### Describe your changes:

Fixes #12083

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/events/scheduled/PipelineServiceStatusJob.java
```diff
@@ -39,11 +46,35 @@ public void execute(JobExecutionContext jobExecutionContext) {
     }
   }
 
+  private String getServiceStatus(PipelineServiceClient pipelineServiceClient) {
+    RetryConfig retryConfig =
+        RetryConfig.<String>custom()
+            .maxAttempts(MAX_ATTEMPTS)
+            .waitDuration(Duration.ofMillis(BACKOFF_TIME_SECONDS * 1_000L))
+            .retryOnResult(response -> !HEALTHY_STATUS.equals(response))
+            .failAfterMaxAttempts(false)
```
pmbrull, line 55: We check 3 times if the status is not unhealthy. If it still is, we return it to be registered in prometheus

harshach (approved):

PR #12139, opened 2023-06-24
Mask queries, Table sample data and Topic sample data

pmbrull:
### Describe your changes:

If a query is flagged as `PII.Sensitive`, we won't show it anymore to non-owners/admins/bots on the LIST API.

image

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/resources/query/QueryResource.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/resources/query/QueryResource.java
@@ -13,7 +13,6 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import java.io.IOException;
 import java.util.List;
-import java.util.Objects;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import javax.json.JsonPatch;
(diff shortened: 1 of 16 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/query/QueryResource.java
```diff
@@ -143,7 +145,19 @@ public ResultList<Query> listQueries(
     if (!CommonUtil.nullOrEmpty(entityId)) {
       filter.addQueryParam("entityId", entityId.toString());
     }
-    return super.listInternal(uriInfo, securityContext, fieldsParam, filter, limitParam, before, after);
+    ResultList<Query> queries =
+        super.listInternal(uriInfo, securityContext, fieldsParam, filter, limitParam, before, after);
+    List<Query> maskedQueries =
```
harshach, line 150: we shouldn't be filtering out queries after fetching them. This will cause issues with pagination. Above pagination call says 10 results returned and we filter out and UI shows as 10 results instead we display only 8 results

openmetadata-service/src/main/java/org/openmetadata/service/resources/topics/TopicResource.java
```diff
@@ -347,6 +349,32 @@ public Topic addSampleData(
     return addHref(uriInfo, topic);
   }
 
+  @GET
+  @Path("/{id}/sampleData")
```
pmbrull, line 353: added the GET sample data endpoint to follow the same approach as with tables. This helps us validate the permissions for masking sample data

OnkarVO7 (approved):

mohityadav766 (approved):

PR #12148, opened 2023-06-26
Refrain from sending an empty string value for the email field when creating a team to avoid a thrown exception

EsmaeilChitgar:
…ST request for creating a team.

Update behavior to disallow submitting an empty string value for the email field when creating a team. If the email input is left blank without entering any characters, no problem is encountered. However, if the email input is filled and then emptied, an exception with the message "[email must match "^\S+@\S+.\S+$", email size must be between 6 and 127]" was previously thrown because the email pattern did not accept empty strings. Rather than altering the email pattern to accept empty strings, the change ensures that empty string values are not submitted.

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


EsmaeilChitgar:
To view the error message, enter some characters in the Email input field and then leave it blank.

email exception

chirag-madlani (approved):

PR #12161, opened 2023-06-26
added composition aggs to aggregate API

07Himank:
I worked on ... because ... added composite aggregation to aggregate API.


mohityadav766 (approved):

mohityadav766:
@aniketkatkar97 can you make changes as per this on the quick-dropdown filter?

PR #12184, opened 2023-06-27
Test suite index task(addded test_suite_search_index)

07Himank:
I worked on ... because ... added test_suite_search as a different index.

#
### Type of change:

- [ x] Improvement


openmetadata-service/src/main/resources/elasticsearch/en/test_suite_index_mapping.json
```diff
@@ -0,0 +1,86 @@
+{
```
harshach, line 1: @07Himank we need to add this index mapping in zh and jp languages as well. Make sure you use appropriate stemmers and settings there

07Himank: yeah sure will do it

07Himank: handle in latest commit

harshach (approved):

PR #12187, opened 2023-06-27
Added pagination Tableau data sources graphql

OnkarVO7:
### Describe your changes:

Added pagination for tableau graphql datasources query and api

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/tableau/client.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/tableau/client.py
@@ -11,6 +11,7 @@
 """
 Wrapper module of TableauServerConnection client
 """
+import math
 import traceback
 from typing import Any, Callable, Dict, List, Optional
 
@@ -23,9 +24,11 @@ from metadata.ingestion.source.dashboard.tableau import (
     TABLEAU_GET_WORKBOOKS_PARAM_DICT,
 )
 from metadata.ingestion.source.dashboard.tableau.models import (
+    DataSource,
     TableauChart,
     TableauDashboard,
     TableauDatasources,
+    TableauDatasourcesConnection,
     TableauOwner,
 )
 from metadata.ingestion.source.dashboard.tableau.queries import (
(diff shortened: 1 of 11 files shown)


ingestion/src/metadata/ingestion/source/dashboard/tableau/client.py
```diff
@@ -124,7 +144,32 @@ def get_datasources(self):
...
+        """
+        try:
+            # Query the graphql endpoint once to get total count of data sources
+            tableau_datasource = self._query_datasources(entities_per_page=1, offset=1)
+            entities_per_page = min(50, self.pagination_limit)
+            indexes = math.ceil(tableau_datasource.totalCount / entities_per_page)
+
+            # Paginate the results
+            data_sources = []
+            for index in range(indexes):
```
harshach, line 161: are we waiting till all the data_sources are fetched?
Why not get a page emit downstream and fetch in next loop?

OnkarVO7: @harshach so we need to prepare the data first so that the datasources get linked to the dashboards [here](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/dashboard/tableau/metadata.py#L114) and then they are used to create the lineage

Also could you please elaborate what do you mean by ->`Why not get a page emit downstream and fetch in next loop?`

pmbrull: Since data models are required multiple times during the processing, the approach now is to keep them in memory to reuse them better. This is something we can further explore in 1.2 to clean things up. For example, I don't see the value of having a `dataModels` field in the Dashboard if we are already bringing out the lineage relationship.

I believe we can find some simplification here that would also streamline the processing. IMO let's keep it as-is now since this is the battle-tested approach, and improve it in 1.2

pmbrull (approved):

PR #12217, opened 2023-06-29
Fix #10181: Dot character in service names cause issues when running the ingestion

### LINKED ISSUE #10181: Dot character in service names cause issues when running the ingestion
**Affected module**
UI and backend

**Describe the bug**
OM API lets us build service names with dots on them and our code treats all names as `fqn`. When a service is created with name `sevice.name` our code will translate it into its `fqn` which is `"sevice.name"` which causes errors when searching entities by name. 

**To Reproduce**

- Create a service with a dot on it using the API.
- Try to ingest from the service.

**Expected behavior**
We should ingest without any error, but we get:

```
metadata.ingestion.api.topology_runner.MissingExpectedEntityAckException: Missing ack back from [DatabaseService: __root__='local_my.sql'] - Possible causes are changes in the server Fernet key or mismatched JSON Schemas for the service connection.
```

**Version:**
 - Python version:

**Additional context**

**Approaches to fix it:**

1. Use `fqn` for the entity name

This option would imply changes in the DB schema and many parts of the code.

2. Stop using `fqn` to refer to service names

We must check when we use the `fqn` to refer to the service names and change it.

3. Stop using `.` char in the service entity names

We have to check current service names in DB and change them.

CC @sureshms

harshach:
### Type of change:

- [x] Bug fix

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java
@@ -8,6 +8,7 @@ import java.util.Map;
 import lombok.Getter;
 import org.openmetadata.schema.type.Include;
 import org.openmetadata.schema.type.Relationship;
+import org.openmetadata.schema.utils.EntityInterfaceUtil;
 import org.openmetadata.service.Entity;
 import org.openmetadata.service.resources.databases.DatasourceConfig;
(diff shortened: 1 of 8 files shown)


harshach:
@pmbrull @mohityadav766 need your review here. We can merge this for 1.1.1

pmbrull (approved):

PR #11840, opened 2023-05-31
worked on the quicktype definitions

phalix:
There are several issues with the quicktype generation right now.

First emailVerificationToken defines the following:
"required": [
"token",
"user",
"tokenType",
"tokenStatus",
"expiryDate"
],

Though there is no type user, but a type userid.

In addition the $id of several classes do not match the path they are actually in. E.g. createWebAnalyticEvent.json defines the $id as https://open-metadata.org/schema/api/tests/createWebAnalyticEvent.json, though the actual path is /api/analytics/createWebAnalyticEvent.json.

Lastly, I ran into problems due to the usage of https://open-metadata.org/ in the id, which caused the quicktype generator to constantly do http request, which also lead to blocked requests.

If you want to have a detailed log of the generation, you can always --debug all on the quicktype command on json2ts.sh:
./node_modules/.bin/quicktype -s schema 
{schema_directory}$1" -o $PWD"/"$om_ui_directory$2 --just-types --debug all > output.log 2>&1
Also I put the output to a file.

### Describe your changes:

Fixes #10811

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

PR #12332, opened 2023-07-07
fix(ui): json2ts error on build

aniketkatkar97:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

- quicktype version bump
- remove unwanted required property from jsons

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-spec/src/main/resources/json/schema/analytics/basic.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$id": "https://open-metadata.org/schema/analytics/basic.json",
+    "$id": "/analytics/basic.json",
     "$schema": "http://json-schema.org/draft-07/schema#",
```
sureshms, line 3: @aniketkatkar97 why are you making the URL change?

aniketkatkar97: the `json2ts` script is not working properly, which is causing the build failure in the main. The cause was found to be that while resolving the refs, errors were thrown as `Error: Key definitions not in schema object at {path}`. We were suspecting the that it is not getting the json file properly through the URL. So trying an approach to give relative path instead. Wich seems to have solved the issue.
cc: @chirag-madlani @akash-jain-10

chirag-madlani: @sureshms @harshach  can we merge this changes?

sureshms: @chirag-madlani and @aniketkatkar97, why did this problem start in the first place? We need a full URL so that JSON can be resolved by the external tools.

harshach:
@chirag-madlani @aniketkatkar97 can we make sure QuickType doesn't make a URL call instead relies on local path.
This is an issue several users are facing and also cause for the QuickType take longer duration to go through the generation process 
On option is to remove $id but that doesn't look good from spec perspective https://github.com/open-metadata/OpenMetadata/pull/11840

chirag-madlani:
> @chirag-madlani @aniketkatkar97 can we make sure QuickType doesn't make a URL call instead relies on local path.
> This is an issue several users are facing and also cause for the QuickType take longer duration to go through the generation process 
> On option is to remove $id but that doesn't look good from spec perspective https://github.com/open-metadata/OpenMetadata/pull/11840
> 

@harshach this changes will block network calls and resolve path locally

chirag-madlani:
> @chirag-madlani @aniketkatkar97 can we make sure QuickType doesn't make a URL call instead relies on local path.
> This is an issue several users are facing and also cause for the QuickType take longer duration to go through the generation process 
> On option is to remove $id but that doesn't look good from spec perspective https://github.com/open-metadata/OpenMetadata/pull/11840
> 

@harshach this changes will block network calls and resolve path locally

openmetadata-spec/src/main/resources/json/schema/analytics/basic.json
chirag-madlani: @sureshms we have verified the full build with changes, can you point me to where exactly full URL needed?

harshach: @chirag-madlani changing this is not an option "$id": "https://open-metadata.org/schema/analytics/basic.json"
that's the reason we said no to the previous PR
My input was to block the URL fetch from QuickType to remove the https://open-metadata.org/schema/ from the $id before passing it to the QuickType or fix the QuickType to ignore it and make a local resolution

sureshms:
@chirag-madlani and @harshach, in the case of jsonschema2pojo that converts JSON to java, because our `$ref` references are relative paths, the resolution happens locally. Isn't that done in QuickType? They also have local file resolution.

harshach:
> @chirag-madlani and @harshach, in the case of jsonschema2pojo that converts JSON to java, because our `$ref` references are relative paths, the resolution happens locally. Isn't that done in QuickType? They also have local file resolution.

@sureshms that's an issue with QuickType it looks at $id and tries to resolve through https instead of local which is what causing the slowness in the build and also some users face issue because they might be building it from a host which blocks external http calls

PR #12378, opened 2023-07-12
Update Dockerfile.ci

ayush-shah:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/Dockerfile.ci
+++ b/ingestion/Dockerfile.ci
@@ -82,7 +82,7 @@ WORKDIR /home/airflow/openmetadata-airflow-apis
 RUN pip install "."
 
 WORKDIR /home/airflow/ingestion
-ARG INGESTION_DEPENDENCY="all"
+ARG INGESTION_DEPENDENCY
 RUN pip install ".[${INGESTION_DEPENDENCY}]"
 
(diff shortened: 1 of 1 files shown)


ingestion/Dockerfile.ci
```diff
@@ -82,7 +82,7 @@ WORKDIR /home/airflow/openmetadata-airflow-apis
 RUN pip install "."
 
 WORKDIR /home/airflow/ingestion
-ARG INGESTION_DEPENDENCY="all"
+ARG INGESTION_DEPENDENCY
```
akash-jain-10, line 85: @ayush-shah Why do we want to remove the default value of argument here ?

ayush-shah: so if i modify env `INGESTION_DEPENDENCY` in the docker-compose file under ingestion container service:
it doesn't take that into account

akash-jain-10: This is the argument / environment variable you need to pass during Docker build phase and not during the running the docker compose!

pmbrull: I think that param is working as expected there. We should close this

PR #12571, opened 2023-07-24
[Backend] Added Quick Links and Article as Knowledge Resources 

mohityadav766:
…to Team Entity

### Describe your changes:

Added Quick Links and Article as Knowledge Resources as an Extension to Team Entity

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -559,6 +559,27 @@ void insert(
             + "ORDER BY extension")
     List<ExtensionRecord> getExtensions(@Bind("id") String id, @Bind("extensionPrefix") String extensionPrefix);
 
+    @RegisterRowMapper(ExtensionMapper.class)
+    @SqlQuery("SELECT extension, json FROM entity_extension WHERE id = :id AND extension LIKE :extension")
```
harshach, line 563: @mohityadav766 any reason to add these, why not just pass it as field?

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/TeamResource.java
```diff
@@ -510,13 +513,133 @@ public CsvImportResult importCsv(
     return importCsvInternal(securityContext, name, csv, dryRun);
   }
 
+  @GET
```
harshach, line 516: Let's not add this as another API end point. we will do this as part of another APIs

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
mohityadav766: need it in direct list apis, existing method has prefix match only

Issue #12011, opened 2023-06-16
Updating tag name will not be reflect in matchAnyTag('') permission


TeddyCr:
If a tag is renamed (scenario below) the permission will not be updated to reflect the tag permission
Scenario
1. a tag rule is set (e.g. matchAnyTag('foo.bar')
2. this tag is renamed from foo.bar to foo.baz
3. the user who had the matchAnyTag('foo.bar') looses permission against asset that have the (now updated tag) named foo.baz?

harshach:
@aji-aju take a look at this issue cc @mohityadav766

PR #12593, opened 2023-07-25
solved issue of matchAnyTag and matchAllTags condition not getting up…

07Himank:
Fixes : #12011

I worked on ... because ... 

- solved issue of matchAnyTag and matchAllTags condition not updating. when updating tag and classification name.
- added more two methods matchAnyGlossary and matchAllGlossaryTerms.

 

#
### Type of change:

- [ x] Bug fix
- [ x] Improvement


07Himank:
@harshach in MatchAnyTag and MatchAllTags() condition do we can want to allow glossaryTerm too ??

mohityadav766 (approved):

PR #12636, opened 2023-07-28
Updating the docker env

Anuj359:
1. Updating work_mem for postgres containers
2. Removing "--sort_buffer_size=10M" from the server section in docker-compose-openmetadata
3. Adding the missing env variables in docker/docker-compose-openmetadata/env-postgres and docker/docker-compose-openmetadata/env-mysql
4. Updating docker/openmetadata.yaml with conf/openmetadata.yaml


docker/openmetadata.yaml
akash-jain-10, line 1: This needs to be a symlink! Check [this](https://stackoverflow.com/questions/954560/how-does-git-handle-symbolic-links) out and make the updates in the PR.

akash-jain-10 (changes requested):

akash-jain-10 (approved):

PR #12660, opened 2023-07-31
no batch in migration updates

mohityadav766:
### Describe your changes:

Remove the batch call and used per request commit

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/migration/MigrationUtil.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/migration/MigrationUtil.java
@@ -14,7 +14,6 @@ import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.tuple.Pair;
 import org.jdbi.v3.core.Handle;
-import org.jdbi.v3.core.statement.PreparedBatch;
 import org.openmetadata.common.utils.CommonUtil;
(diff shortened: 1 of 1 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/migration/MigrationUtil.java
```diff
@@ -164,13 +162,19 @@ public static <T extends EntityInterface> void readAndProcessEntity(
...
                   : FullyQualifiedName.buildHash(entity.getFullyQualifiedName());
-          upsertBatch.bind("nameHashColumnValue", hash).bind("id", entity.getId().toString()).add();
+          int result =
+              handle
+                  .createUpdate(updateSql)
+                  .bind("nameHashColumnValue", hash)
+                  .bind("id", entity.getId().toString())
+                  .execute();
+          if (result <= 0) {
+            LOG.error("No Rows Affected for Updating Hash with Entity Name : {}", entity.getFullyQualifiedName());
```
harshach, line 172: if result is zero should we retry? if there is no exception

mohityadav766: it will again come in further reads with ->  select * from <entity_table> where nameHash = '' or nameHash is null;

PR #12712, opened 2023-08-02
Able to get the instance of resource classes everywhere

EsmaeilChitgar:
### Describe your changes:
This pull request offers a solution for obtaining an instance of a resource class within another class, such as in situations where multiple tasks need to be performed with a single UI click. The specific use case addressed is retrieving user information in the DatabaseServiceResource class. To meet this requirement, two ways for obtaining an instance of a resource class are suggested:

UserResource userResource = ResourceFactory.getResource(UserResource.class);
var userResource = (UserResource)ResourceFactory.getResource(UserResource.class);

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/resources/ResourceFactory.java
```diff
@@ -0,0 +1,29 @@
+package org.openmetadata.service.resources;
+
+import java.lang.reflect.Field;
+
+public class ResourceFactory {
```
harshach, line 5: @EsmaeilChitgar can you please add in the description what is the problem scope and why we are adding this code. May be there is an easier option todo this.

EsmaeilChitgar: Sure. Let's say you need to createOrUpdate multiple entities simultaneously. For instance, you have a single click request in front side, and upon receiving that request, you need to createOrUpdate a DatabaseService, a Database, a DatabaseSchema, and several tables. In this scenario, you would typically need to access the databaseResource, DatabaseSchemaResource, and TableResource from the DatabaseServiceResource.

Is there a simpler approach to achieve this?

sureshms: @EsmaeilChitgar, couple of things.
1. From the UI a REST API call is made to the server. Currently, it is a request for one Resource at a time for doing one operation. If we need an operation that does multiple things, we should build a Resource or a method call for that.
2. Second, you don't want to be using Resources to build such operations. A resource takes a request and provides a response. Instead what we should use is Repository classes. And a functionality you are building is already supported by Entity class to getEntityRepository method.

This change appears to be not necessary.

EsmaeilChitgar: I wasn't aware of the getEntityRepository method, and it appears that its usage would hinder my ability to utilize the useful methods of the EntityResource class, including the copy method. To overcome this limitation, I will need to develop an alternative solution, such as creating my own copy method or exploring other possibilities.

sureshms: Assuming you want to build a bulk API, this is not the right approach. You need to build a resource that can take a request (that has information about multiple entities) and user Repository classes to update multiple entities at the same time. In this case copy method would involve more than one type of entity, and as it exists is not useful.

Can you describe what your use case is and what functionality you are building?

EsmaeilChitgar: suppose there is a request to createOrUpdate a table. I would call the createOrUpdate method of the DatabaseServiceResource to create it if does not exist. The same approach would be followed for the database and databaseSchema. The info containig name for database service, database, and database schema would be obtained from the input parameters.

sureshms: Sorry it doesn't make sense to me. Current createOrUpdate for table has only information relevant to table + a databaseSchema reference. In order to create entities up the chain, the createOrUpdate request to table must have information about databaseSchema, database, and databaseService.  I am -1 on this change.

EsmaeilChitgar: Yes. Your' right. I understood.

Issue #11178, opened 2023-04-20
Add URL as default entity field schema definition


stephanesol:
**Is your feature request related to a problem? Please describe.**

The dashboard entity in OMD has a [dashboardUrl](https://github.com/open-metadata/OpenMetadata/blob/9acf45cd1b3e9c055121feee3fcd82a716502bd3/openmetadata-spec/src/main/resources/json/schema/entity/data/dashboard.json#L43) .  If populated this URL is exposed to end users via the UI and enables them to continue their data exploration journey.   I think this url field should be made available many more if not all entity types.   

For my use case, i'd like to be able to populate a URL for snowflake tables so users can go from discovering data in OpenMetadata to directly querying/engaging with the data source in another tool.

The UI for the dashboard entity does this well and i think it could be extended to other services like databases.

**Describe the solution you'd like**

Allow some kind of url on the database table/view entity that could be populated via ingestion or api.
Allow this to be exposed, if available, in the UI similar to how it is done for dashboards.

**Describe alternatives you've considered**

Right now the alternative i'm using is using custom properties, which i understand can be populated with anything i want 🙇  but i think it can be buried in the UI.  This requested change, I believe,  would provide users with a clearer call to action to further explore data.

**Additional context**

cc @harshach from community office hours 🙇

harshach:
@ulixius9 sourceURL is added as part of this PR https://github.com/open-metadata/OpenMetadata/commit/c3cec54be926dd511087af689798ddc3fc91d0ba
but we missed adding it for Topic, Database, DatabaseSchema, MLModel, Container. Let's make sure we add it everywhere.

@07Himank can you take this up

PR #12773, opened 2023-08-07
Added source url to search indexes and missing entites

07Himank:
Fixes : #11178

I worked on ... because ... added SourceUrl to the missing entites as well as search index mapping files

#
### Type of change:

- [ x] Bug fix
- [ x] Improvement


openmetadata-spec/src/main/resources/json/schema/api/data/createTopic.json
```diff
@@ -80,6 +80,10 @@
       "description": "Entity extension data with custom attributes added to the entity.",
       "$ref": "../../type/basic.json#/definitions/entityExtension"
     },
+    "sourceUrl": {
+      "description": "Source URL of topic.",
+      "$ref": "../../type/basic.json#/definitions/sourceUrl"
```
harshach, line 85: @07Himank we already have URL type, take a look at that

harshach:
@07Himank lets create a separate api in search resource to be able to look up by just passing URL without adding any other ranking etc..

07Himank:
> @07Himank lets create a separate api in search resource to be able to look up by just passing URL without adding any other ranking etc..

@harshach , Thats what i wanted to confirm we need to add a separate API to get entities from sourceURL (with exact match ) right ? as we discussed

openmetadata-spec/src/main/resources/json/schema/api/data/createTopic.json
07Himank: @harshach I took reference from dashboard.json , where sourceURL was already there do we need to update that ?

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticSearch/ElasticSearchClientImpl.java
```diff
@@ -363,6 +363,29 @@ public Response search(ElasticSearchRequest request) throws IOException {
...
 
+  /**
+   * @param sourceUrl
+   * @return
+   */
+  @Override
+  public Response searchBySourceUrl(String sourceUrl) throws IOException {
+    String[] indexes =
+        new String[] {
+          ElasticSearchIndexDefinition.ElasticSearchIndexType.TABLE_SEARCH_INDEX.indexName,
```
harshach, line 374: we should create alias for the indexes and search across all the indexes at once using the alias

07Himank: thanks for review, will push the changes..

mohityadav766 (approved):

mohityadav766 (approved):

PR #12790, opened 2023-08-08
fixed issue table_search_index deleted field not getting updated when…

07Himank:
Fixes : #12792

I worked on ... because ... when databaseSchema is deleted,
table_search_index deleted field should get updated as true or false respectively for softDelete and restore.

#
### Type of change:

- [ x] Bug fix
- [ x] Improvement


openmetadata-service/src/main/java/org/openmetadata/service/search/elasticSearch/ElasticSearchClientImpl.java
```diff
@@ -1087,6 +1089,46 @@ public void updateDatabase(ChangeEvent event) throws IOException {
...
+      String[] indexes =
+          new String[] {ElasticSearchIndexDefinition.ElasticSearchIndexType.TABLE_SEARCH_INDEX.indexName};
+      BulkRequest request = new BulkRequest();
+      int batchSize = 50;
+      int totalHits;
+      int currentHits = 0;
+      SearchRequest searchRequest;
+      SearchResponse response;
+      UpdateRequest updateRequest;
+      do {
```
harshach, line 1103: why are we doing this loop @07Himank

Issue #11494, opened 2023-05-08
Azure blob storage connection


LPontes:
I'm looking to implement openmetadata in the company where I work, but we use azure's data lake gen 2.
I would like to see Azure blob storage connector implelmented on OM.
Meanwhile I'm trying to use Databricks connector however It relies on a dedicated cluster

tomtom215:
We'd love this as well as we are only using Azure Blob and Azure File without the option of using S3 or MinIO

pmbrull:
hi @LPontes, @tomtom215, you should be able to already use the Datalake connector [docs](https://docs.open-metadata.org/v1.1.0/connectors/database/datalake) to ingest individual files as tables from ADLS.

We'll use this ticket to implement ADLS as a Storage Service [docs](https://docs.open-metadata.org/v1.1.0/connectors/storage)

Thanks

PR #12810, opened 2023-08-09
Storage connection naming, ADLS service and namespace packaging

pmbrull:
### Describe your changes:

As a prerequisite for  #11494

There is no connector yet so we can rename freely.

Moreover, since we need to use namespace packages to expand the functionalities for Collate, we are removing most of the `__init__.py` files to be able to freely extend the code as needed. From the [PEP](https://peps.python.org/pep-0420/):
> A namespace package is not fundamentally different from a regular package. It is just a different way of creating packages. Once a namespace package is created, there is no functional difference between it and a regular package.

Also:
- Not sure why we had the `examples` inside the `metadata` module. Moving them directly in `ingestion`

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach (dismissed):

ingestion/src/metadata/ingestion/source/database/hive/metastore_dialects/mysql/__init__.py
```diff
@@ -1,23 +0,0 @@
...
-from sqlalchemy.dialects import registry
-
-from .dialect import HiveMysqlMetaStoreDialect
-
-__version__ = "0.1.0"
-registry.register(
-    "hive.mysql",
-    "metadata.ingestion.source.database.hive.metastore_dialects.mysql.dialect",
-    "HiveMysqlMetaStoreDialect",
-)
```
ulixius9, line 23: @pmbrull  these files would be required to load the dialect

ingestion/src/metadata/ingestion/source/database/hive/metastore_dialects/postgres/__init__.py
```diff
@@ -1,23 +0,0 @@
...
-from sqlalchemy.dialects import registry
-
-from .dialect import HivePostgresMetaStoreDialect
-
-__version__ = "0.1.0"
-registry.register(
-    "hive.postgres",
-    "metadata.ingestion.source.database.hive.metastore_dialects.postgres.dialect",
-    "HivePostgresMetaStoreDialect",
-)
```
ulixius9, line 23: this one as well

ulixius9 (approved):

ingestion/src/metadata/ingestion/source/database/hive/metastore_dialects/mysql/__init__.py
pmbrull: thanks! Any reason for them to be inside `__init__`?

pmbrull: since they are used when getting the connection maybe we can move this logic to `connection` instead of doing the import?

pmbrull: actually I'll add the `__init__.py` back under these directories

ulixius9 (approved):

PR #12912, opened 2023-08-17
issue-12903 DELETE endpoint for profiler data

### LINKED ISSUE #12903: Data profiler deletes tableProfile and columnProfile, but not systemProfile
**Affected module**
This is backend and API. 

**Describe the bug**
Using [deletDataProfiler](https://docs.open-metadata.org/swagger.html#operation/deleteDataProfiler) you can delete table profile data and column profile data, but not system profile data. Choosing entityType="system" should also be a possibility. 

**To Reproduce**
Use  API provided in the link above and try to remove systemProfile.

**Expected behavior**
I expect to have a possibility to remove systemProfiles without deleting the whole Table Profile Data. 

**Version:**
 - OpenMetadata version: 1.1.1

**Additional context**
-

TeddyCr:
### Describe your changes:

Fixes #12903 
Fixes delete endpoint for profiler data

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -412,16 +412,21 @@ public Table addTableProfileData(UUID tableId, CreateTableProfile createTablePro
   public void deleteTableProfile(String fqn, String entityType, Long timestamp) {
     // Validate the request content
     String extension;
+    Class classMapper;
```
TeddyCr, line 415: @open-metadata/backend let me know if there is a better approach here as Sonar throws the following

> Generic types shouldn’t be used raw (without type parameters) in variable declarations or return values. Doing so bypasses generic type checking, and defers the catch of unsafe code to runtime.

harshach (approved):

PR #12967, opened 2023-08-22
Fix #12475: Case sensitivity of user entity name

### LINKED ISSUE #12475: Case sensitivity of user entity name in Azure SSO
**Affected module**
Azure SSO

**Describe the bug**
After logging in through Azure SSO a new user_entity is created with (basically) my E-Mail address. Because of the ActiveDirectory Toolset which is used to administer the Azure AD, a lot of these E-Mail addesses are written with capital letters (like 'Juergen.Zornig' in my case). In the database the user_entity object is created exactly like this, so the user exists now. But In the App OM tells me 'user instance for "juergen.zornig" not found', therefore I don't have any rights or privileges at all.

**To Reproduce**

1. Login via Azure SSO with a new (not logged in before) user
2. User gets created, Userid is taken from email (cannot be modified)
grafik

3. Logged in, but without User Profile
grafik

**Expected behavior**
There is a fitting user profile after a (Azure SSO) user is created by OM

**Version:**
 - OS: Docker Host is Ubuntu Server 22.04.1
 - OpenMetadata version: Docker Deployment for OM 1.1
 - OpenMetadata Ingestion package version: Docker Deployment for OM 1.1

**Additional context**
We are using docker compose to directly execute the containers using the downloaded docker-compose-postgres.yml file with the additional --env-file flag as noted in the instructions (so Postgres Backend on Docker) 

After Logout and logging in again, I get back to the "create user" modal which tells me, that the user already exists. So no chance to login again without deleting my userid in the database backend.

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -2797,12 +2798,21 @@ List<String> listAfter(
...
-    @ConnectionAwareSqlQuery(value = "SELECT count(*) FROM user_entity WHERE email = :email", connectionType = POSTGRES)
+    @SqlQuery("SELECT COUNT(*) FROM user_entity WHERE LOWER(email) = LOWER(:email)")
     int checkEmailExists(@Bind("email") String email);
 
-    @SqlQuery("SELECT json FROM user_entity WHERE email = :email")
+    @SqlQuery("SELECT json FROM user_entity WHERE LOWER(email) = LOWER(:email)")
     String findUserByEmail(@Bind("email") String email);
+
+    @Override
+    default User findEntityByName(String fqn, Include include) {
```
sureshms, line 2808: This override is not required, right? It all boils down to the second method you have overridden?

sureshms (dismissed):

mohityadav766:
closing this as handles by https://github.com/open-metadata/OpenMetadata/pull/12942

PR #13029, opened 2023-08-29
Part of #7272 - Centralize Workflows, Status, and Exception Management

pmbrull:
### Describe your changes:

Part of #7272

Pending:
- Profiler, Data Insights and Data Quality workflows should be created from `BaseWorkflow`
- Status handling needs to be updated for the workflows above after the update
- Clean the legacy Sink, Processor, and BulkSink classes in favor of using `steps.*`
- Re-implement the `test_workflow_output_handler` with the new status structure

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/api/topology_runner.py
```diff
@@ -182,17 +187,21 @@ def sink_request(self, stage: NodeStage, entity_request: C) -> Iterable[Entity]:
         """
 
         # Either use the received request or the acknowledged Entity
-        entity = entity_request
+        entity = entity_request.right
 
         if stage.nullable and entity is None:
```
TeddyCr, line 192: wouldn't this be `if not stage.nullable` if something can be nullable (ie `stage.nullable is True`) then we should not throw a ValueError right?

ingestion/src/metadata/workflow/base.py
```diff
@@ -0,0 +1,245 @@
...
+        by each ingestion workflow.
+
+        Pass each record from the source down the pipeline:
+        Source -> (Processor) -> Sink
+        or Source -> (Processor) -> Stage -> BulkSink
+
+        Note how the Source class needs to be an Iterator. Specifically,
+        we are defining Sources as Generators.
+        """
+        for record in self.source.run():
```
TeddyCr, line 128: where do we enforce that `Source` should have a run()

pmbrull: [here](https://github.com/open-metadata/OpenMetadata/pull/13029/files#diff-0299422e9540c993001cd05eea601d12500c7f3de9b589d596625458e0346260R144). Might add the `run` as an abstractmethod to the Step class itself

TeddyCr: the `Source` class inherit from the `StageStep` class?

pmbrull:
> @pmbrull, could you add a UML diagram of the class relationship somewhere, it would help get a better overview of the structure and the different relationships between objects

I'll take care of it. I'll publish it in another PR

TeddyCr (approved):

PR #13032, opened 2023-08-29
Handle relationship test case + entity not found error throwing

TeddyCr:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -122,25 +122,25 @@ public void prepare(TestCase test) {
...
+    throw new EntityNotFoundException(
+        String.format("Error occurred when retrieving executable test suite for testCase %s. ", test.getName())
+            + "No executable test suite was found.");
   }
 
   private List<TestSuite> getTestSuites(TestCase test) {
     // `testSuites` field returns all the `testSuite` (executable and logical) linked to that testCase
     List<CollectionDAO.EntityRelationshipRecord> records =
         findFromRecords(test.getId(), entityType, Relationship.CONTAINS, TEST_SUITE);
-    ensureSingleRelationship(entityType, test.getId(), records, Relationship.CONTAINS.value(), true);
```
harshach, line 143: we should ensureSingleRelationship, for testcase -> testSuite that is executable no?

TeddyCr: Wouldn't we fall in the same case though as we'll list all all the relationships to the testSuites in the method?

PR #13051, opened 2023-08-31
exact match on sourceUrl

07Himank:
I worked on ... because ...exact match on sourceUrl

#
### Type of change:

- [ x] Improvement


mohityadav766 (commented):
@07Himank  let's fix the serviceType.keyword issue also happening in ES in this

PR #13056, opened 2023-08-31
Bootstrap script update

Anuj359:
Updating bootstrap script to include force and debug as getopts.

The script now can be run as:
`bootstrap.sh create` or 
`bootstrap.sh create -d` or 
`bootstrap.sh create -f -d` or
`bootstrap.sh create -d -f`

-d and -f are set to use debug and force, which are optional and by default are false.

### DIFF
--- a/bootstrap/bootstrap_storage.sh
+++ b/bootstrap/bootstrap_storage.sh
@@ -10,6 +10,30 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+printUsage() {
+  cat <<-EOF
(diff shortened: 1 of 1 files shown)


bootstrap/bootstrap_storage.sh
```diff
@@ -58,60 +103,36 @@ else
 fi
 
 execute() {
-  if  [ ${debug} ] ; then
+  local command="$1"
+  local debug="$2"
+  local force="$3"
```
akash-jain-10, line 108: Why use local when you have global variables access inside function `execute` ?

mohityadav766 (approved):

Issue #12954, opened 2023-08-22
UI - restrict roles and policy in settings for admins only


chirag-madlani:
**Affected module**
Does it impact the UI, backend or Ingestion Framework?
UI

**Describe the bug**
- users can access all the roles and policies

**To Reproduce**

Screenshots or steps to reproduce

- Navigate to settings page
- User can see the Role & Policy option and can access the pages

**Expected behavior**

Only admins can view or modify Roles & Policy sections

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8] -- 1.1.2
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

Sachin-chaurasiya:
Fixed with #13073

PR #13073, opened 2023-09-04
fix(#12954): restrict roles and policy in settings for admins only

Sachin-chaurasiya:
### Describe your changes:

Fixes 12954

I worked on issue #12954 

https://github.com/open-metadata/OpenMetadata/assets/59080942/e9e3acd1-4b28-491d-82a1-cd8e2d64c8c0

#
### Type of change:

- [x] Bug fix

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/pages/PoliciesPage/PoliciesDetailPage/PoliciesDetailPage.tsx
```diff
@@ -256,7 +232,7 @@ const PoliciesDetailPage = () => {
     (rule: Rule) => {
       return (
         <Dropdown
-          disabled={!policyPermission.EditAll}
+          disabled={!isAdminUser}
```
karanh37, line 235: If non admin users doesn't have access to view the page, which we restricted from GlobalSettingsRouter.
Why do we need disabled conditions ?
Also are we not following the approach where we don't have access we are the hiding the action buttons only ?

Sachin-chaurasiya: Thanks @karanh37 , I will remove the disabled part and as only the admin user will have the page access so will not hide anything on condition.

chirag-madlani (approved):

PR #13075, opened 2023-09-04
Revert "fix: Align Docker Base Image to use airflow 2.5.3"

jan-pieter:
Reverts open-metadata/OpenMetadata#12737

The reason for downgrading the airflow base image was that the helm chart did not support airflow 2.6.x. Since then the dependency has been updated as part of https://github.com/open-metadata/openmetadata-helm-charts/pull/168 and airflow 2.6.3 is used. The database upgrade now fails because the new revision ID is not known in the old airflow version resulting in the following error: `alembic.script.revision.ResolutionError: No such revision or branch 'c804e5c76e3e'`. See https://airflow.apache.org/docs/apache-airflow/stable/migrations-ref.html

### DIFF
--- a/ingestion/Dockerfile
+++ b/ingestion/Dockerfile
@@ -1,4 +1,4 @@
-FROM apache/airflow:2.5.3-python3.9
+FROM apache/airflow:2.6.3-python3.9
 USER root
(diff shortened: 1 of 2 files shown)


jan-pieter:
Done in #13151

Issue #13067, opened 2023-09-04
[Looker] Error when run ingest metadata from Looker


locnh2012:
**Affected module**
Ingestion Framework

**Describe the bug**
```
[2023-08-29, 06:07:13 UTC] {taskinstance.py:1824} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/operators/python.py", line 181, in execute
    return_value = self.execute_callable()
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/operators/python.py", line 198, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/home/airflow/.local/lib/python3.9/site-packages/openmetadata_managed_apis/workflows/ingestion/common.py", line 221, in metadata_ingestion_workflow
    workflow.execute()
  File "/home/airflow/.local/lib/python3.9/site-packages/metadata/ingestion/api/workflow.py", line 215, in execute
    raise err
  File "/home/airflow/.local/lib/python3.9/site-packages/metadata/ingestion/api/workflow.py", line 196, in execute
    for record in self.source.next_record():
  File "/home/airflow/.local/lib/python3.9/site-packages/metadata/ingestion/api/topology_runner.py", line 132, in next_record
    yield from self.process_nodes(get_topology_root(self.topology))
  File "/home/airflow/.local/lib/python3.9/site-packages/metadata/ingestion/api/topology_runner.py", line 98, in process_nodes
    yield from self.process_nodes(child_nodes)
  File "/home/airflow/.local/lib/python3.9/site-packages/metadata/ingestion/api/topology_runner.py", line 83, in process_nodes
    yield from self.sink_request(
  File "/home/airflow/.local/lib/python3.9/site-packages/metadata/ingestion/api/topology_runner.py", line 194, in sink_request
    entity_fqn = self.fqn_from_context(
  File "/home/airflow/.local/lib/python3.9/site-packages/metadata/ingestion/source/dashboard/dashboard_service.py", line 491, in fqn_from_context
    *context_names, entity_request.name.__root__
AttributeError: 'AddLineageRequest' object has no attribute 'name'
```

**To Reproduce**

Error when ingesting lineage metadata from looker which specify LookML repo

**Expected behavior**
The pipeline run successfully

**Version:**
 - Python version:
 - OpenMetadata version: 1.1.1
[...]

PR #13140, opened 2023-09-11
Support LookML multi repos

locnh2012:
### Describe your changes:

To Support Looker Multi repos and Lineage from the tables to LookMLView.

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


harshach:
@locnh2012 is the PR ready to review?

locnh2012:
> @locnh2012 is the PR ready to review?

Yes, I think it is ready to review, please.

ingestion/src/metadata/ingestion/api/models.py
```diff
@@ -39,3 +44,4 @@ class Either(BaseModel, Generic[T]):
 
     left: Optional[StackTraceError]
     right: Optional[T]
+    middle: Optional[Nothing]
```
pmbrull, line 47: Let's remove the `middle` change in the `Either` spec. If the ingestion fails we might be missing some validation in what we're yielding

ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
```diff
@@ -127,6 +127,7 @@ class DashboardServiceTopology(ServiceTopology):
                 context="dataModel",
                 processor="yield_bulk_datamodel",
                 consumer=["dashboard_service"],
+                ack_sink=False,
```
pmbrull, line 130: We removed the `ack_sink` from the Topology model to simplify this. Since here we don't want to run the GET in the topology_runner, we can just remove the `context` and manage the context in the `source` itself

ingestion/src/metadata/ingestion/api/topology_runner.py
```diff
@@ -189,8 +189,9 @@ def sink_request(
         # Either use the received request or the acknowledged Entity
         entity = entity_request.right
 
-        if not stage.nullable and entity is None:
-            raise ValueError("Value unexpectedly None")
+        # if not stage.nullable and entity is None:
+        #     logger.error(f"TopologyRunnerMixin::sink_request: {entity_request.left}")
+        #     raise ValueError("Value unexpectedly None")
```
pmbrull, line 194: this we should uncomment

ingestion/src/metadata/readers/file/local.py
```diff
@@ -38,6 +39,9 @@ def read(self, path: str, **kwargs) -> Union[str, bytes]:
         to let the client use this data as needed.
         """
         try:
+            logger.info(f"LocalReader::read: path = {path}")
+            if path == "data_engineering/data_engineering.explore.lkml":
+                logger.info(f"LocalReader::read: start debug")
```
pmbrull, line 44: this looks like some testing thing

ingestion/src/metadata/ingestion/source/dashboard/looker/metadata.py
```diff
@@ -190,12 +274,31 @@ def parser(self, all_lookml_models: Sequence[LookmlModel]) -> None:
...
-                        )
-                    )
+                # project_name: LkmlParser(
+                #     # reader=self.reader(
+                #     #     credentials=self.get_lookml_project_credentials(
+                #     #         project_name=project_name
+                #     #     )
+                #     # )
+                #     reader=self.reader(Path(self._main_lookml_repo.path)),
+                #     remote_readers={
```
pmbrull, line 284: if this code is not relevant anymore we can delete the comment block

ingestion/src/metadata/readers/file/local.py
locnh2012: removed

ingestion/src/metadata/ingestion/api/models.py
locnh2012: In some scenarios, it only skip to process but not an errors. 
For example, if lineage already added so it can be skip to call api to openmetadata server to reduce numbers of api call. So `middle` can help us in some cases like this, wdyt? @pmbrull

ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
locnh2012: AFAIK, the default value of `ack_sink=True`, so if we remove `ack_sink` from the Topology model, it can cause previous error which have been describe in here: https://github.com/open-metadata/OpenMetadata/issues/13067

ingestion/src/metadata/ingestion/source/dashboard/looker/metadata.py
locnh2012: I forgot to ingtegrate with bitbucket, do you think that it can work with both github & bitbucket?

ingestion/src/metadata/ingestion/api/models.py
pmbrull: but in those cases we are not `yield`ing anything, so should be fine since nothing gets to the runner right

ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
pmbrull: just remove the context as well and it won't be added. You are controlling the context on the source itself

ingestion/src/metadata/ingestion/source/dashboard/looker/metadata.py
pmbrull: if the lib you are using to clone handles both then it should be fine, but we should have support for both sources yes

ingestion/src/metadata/ingestion/api/models.py
locnh2012: Removed `Middle` and `Nothing`

ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
locnh2012: Removed `context` and `ack_sink` params

ingestion/src/metadata/ingestion/source/dashboard/looker/bulk_parser.py
```diff
@@ -0,0 +1,148 @@
...
+from ingestion.src.metadata.readers.file.api_reader import ApiReader
+from ingestion.src.metadata.readers.file.local import LocalReader
+
+logger = ingestion_logger()
+
+EXTENSIONS = (".lkml", ".lookml")
+IMPORTED_PROJECTS_DIR = "imported_projects"
+
+
+class SingletonMeta(type):
```
pmbrull, line 37: we already have a `Singleton` class you can use https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/utils/singleton.py

ingestion/src/metadata/ingestion/source/dashboard/looker/bulk_parser.py
```diff
@@ -0,0 +1,148 @@
...
+    LkmlFile,
+    LookMlView,
+    ViewName,
+)
+from metadata.readers.file.base import ReadException
+from metadata.utils.logger import ingestion_logger
+
+
+from ingestion.src.metadata.readers.file.api_reader import ApiReader
+from ingestion.src.metadata.readers.file.local import LocalReader
```
pmbrull, line 29: these imports are going to fail once the package is installed. Instead of `ingestion.src.metadata` it should directly be `metadata....`

pmbrull: i've seen them in a few places

ingestion/src/metadata/ingestion/source/dashboard/looker/bulk_parser.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        Check if view is cached, and return it.
+        Otherwise, return None
+        """
+        if view_name in self._views_cache:
+            return self._views_cache[view_name]
+
+        return None
+
+    def find_view(self, view_name: ViewName, path: Includes) -> Optional[LookMlView]:
```
pmbrull, line 125: I don't think this method is being called, and the `path` argument is unused

ingestion/src/metadata/ingestion/source/dashboard/looker/metadata.py
```diff
@@ -88,18 +93,36 @@
...
+from ingestion.src.metadata.ingestion.lineage.parser import LineageParser
+from ingestion.src.metadata.ingestion.source.dashboard.looker.bulk_parser import (
+    BulkLkmlParser,
+)
+from ingestion.src.metadata.ingestion.source.dashboard.looker.models import (
+    LookMLRepo,
+    LookMLManifest,
+)
+from ingestion.src.metadata.ingestion.source.dashboard.looker.utils import _clone_repo
+from ingestion.src.metadata.readers.file.local import LocalReader
```
pmbrull, line 119: same with the imports

pmbrull (commented):
Hi @locnh2012 thanks for the help here. Aside from the comments, make sure to run `make py_format` to get imports etc. properly organized

Thanks!

ingestion/src/metadata/ingestion/source/dashboard/looker/bulk_parser.py
locnh2012: changed to `Singleton`

ingestion/src/metadata/ingestion/source/dashboard/looker/bulk_parser.py
locnh2012: Import: Updated

ingestion/src/metadata/ingestion/source/dashboard/looker/metadata.py
locnh2012: Import Updated

ingestion/src/metadata/ingestion/source/dashboard/looker/bulk_parser.py
locnh2012: @pmbrull I've remove the `path` argument, btw, this method is still called in `_process_view()` function
https://github.com/open-metadata/OpenMetadata/blob/60633dbfc0dc9ad319f2cddff4a4496ddcb5d2f1/ingestion/src/metadata/ingestion/source/dashboard/looker/metadata.py#L526

PR #13143, opened 2023-09-11
Issue-12305 - Add logic to sort test cases result by latest execution

### LINKED ISSUE #12305: Fix test Case ordering on DQ Page

TeddyCr:
### Describe your changes:

Fixes #12305
- Added logic to store latest execution state with the test case entity
- Added logic to sort test case entities by latest execution date of a result

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -256,23 +267,84 @@ private void setTestSuiteSummary(TestCase testCase, Long timestamp, TestCaseStat
     for (TestSuite testSuite : testSuites) {
       testSuite.setSummary(null); // we don't want to store the summary in the database
       List<ResultSummary> resultSummaries = listOrEmpty(testSuite.getTestCaseResultSummary());
-      if (resultSummaries.isEmpty()) {
-        resultSummaries.add(resultSummary);
-      } else {
-        // We'll remove the existing summary for this test case and add the new one
+      if ((isDeleted) && (resultSummaries.isEmpty())) {
```
harshach, line 270: this should be OR condition? we will not add if the testCase is deleted or there are not results?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -256,23 +267,84 @@ private void setTestSuiteSummary(TestCase testCase, Long timestamp, TestCaseStat
...
+
+      if (!shouldUpdateResultSummary(storedResultSummary, timestamp, isDeleted)) {
+        continue; // if the state should not be updated then nothing to do
+      }
+
+      if (storedResultSummary != null) {
         resultSummaries.removeIf(summary -> summary.getTestCaseName().equals(resultSummary.getTestCaseName()));
+      }
+
+      if (!isDeleted) {
```
harshach, line 284: do we need this condition , given we already checked it above?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -233,6 +241,8 @@ public RestUtil.PutResponse<?> deleteTestCaseResult(String updatedBy, String fqn
       testCase.setTestCaseResult(storedTestCaseResult);
       ChangeDescription change = deleteTestCaseChangeDescription(testCase.getVersion(), storedTestCaseResult);
       ChangeEvent changeEvent = getChangeEvent(updatedBy, testCase, change, entityType, testCase.getVersion());
+      setTestSuiteSummary(testCase, timestamp, storedTestCaseResult.getTestCaseStatus(), true);
```
harshach, line 244: @TeddyCr this code is bit confusing, why are we passing deleted as true?
are we looking at the testCase is deleted or not, if so shouldn't we check testCase.deleted instead of passing as another argument

TeddyCr: @harshach this is a testCaseResult -- not a testCase entity. If a testCase entity is deleted we handle it in the `postDelete` to remove its state from the testSuite.

As far as a testCaseResult goes it does not have a `deleted` field. We need to update the state of the testSuite entity and the testCase entity if we have a testCaseResult that is deleted and it is the latest one.

I think this is the downside of adding state to the entities. We have manage these cases.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
TeddyCr: This handles cases where you would try to delete a testCaseResult and the resultSummaries (which holds the state) is already empty. In this scenario there is nothing to do. Let me know if that makes sense.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
TeddyCr: correct, it should be removed in the method

TeddyCr:
@harshach here is the list of scenario we handle when it comes to `testCaseResult` 
Screenshot 2023-09-13 at 9 58 23 AM

For testCase (the entity) deletion, it is more straightforward as we just need to remove it from the testSuite entity result summary.

harshach (approved):

PR #13170, opened 2023-09-13
fix: implement percentile computation logic for SingleStore

TeddyCr:
### Describe your changes:
- remove the dispatch logic in the profiler interface to allow override of metric computation methods
- override at runtime percentile metrics for SingleStore. We can't add a `compiles` logic as the dialect will be flagged as `mysql`. We go around this by implementing a SingleStore profiler interface and overwrite the window metrics method and ingesting SingleStore metrics at runtime

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
```diff
@@ -305,6 +286,9 @@ def _(
         Returns:
             dictionnary of results
         """
+        session = kwargs.get("session")
+        column = kwargs.get("column")
```
pmbrull, line 290: if we need both the session and column can we add them as explicit parameters?

TeddyCr: @pmbrull we can. I was trying to comply with the parent method signature -- as mypy was complaining. Let me know what you think

pmbrull: then we could add them to the parent as well right? I imagine that any metric computation would always need the `session` and `columns` so we could define it as such in the abs

TeddyCr: Not necessarily. data lake does not need a session object and `column` is table-specific.

pmbrull (dismissed):

PR #13283, opened 2023-09-20
Update ingestion logic to use PATCH API for lifeCycle info

OnkarVO7:
### Describe your changes:

Update ingestion logic to use PATCH API for lifeCycle info

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/bulksink/metadata_usage.py
```diff
@@ -348,9 +350,43 @@ def _get_table_life_cycle_data(
         the query being processed.
         """
         try:
-            # Temporary clean up this method to be implemented following the new schema
-            life_cycle = (  # pylint: disable=unused-variable
-                init_empty_life_cycle_properties()
+            life_cycle = LifeCycle()
```
ayush-shah, line 353: if this is not something we want repeated, we can modify it to be used in init

ayush-shah, line 353: if this is not something we want repeated, we can modify it to be used in init

ingestion/src/metadata/ingestion/bulksink/metadata_usage.py
```diff
@@ -348,9 +350,43 @@ def _get_table_life_cycle_data(
...
+                query_type = get_query_type(create_query=create_query)
+                access_details = AccessDetails(
+                    timestamp=create_query.queryDate.__root__,
+                    accessedBy=user,
+                    accessedByAProcess=process_user,
+                )
+                if query_type == "Created" and (
+                    not life_cycle.created
+                    or life_cycle.created.timestamp.__root__
+                    < access_details.timestamp.__root__
```
ayush-shah, line 372: `or` part is repeated on all the ifs, maybe we could modify it to have it used ones

ingestion/src/metadata/utils/life_cycle_utils.py
```diff
@@ -80,3 +82,23 @@ def get_query_type(create_query) -> Optional[str]:
...
+    """
+    try:
+        destination = entity.copy(deep=True)
+        destination.lifeCycle = life_cycle
+        metadata.patch(entity=type(entity), source=entity, destination=destination)
+    except Exception as exc:
+        logger.debug(traceback.format_exc())
+        logger.warning(
+            f"Error trying to Patch life cycle data for {entity.fullyQualifiedName.__root__}: {exc}"
+        )
```
ayush-shah, line 104: Patching part would be a better fit in mixins rather than utils

ayush-shah: let's use utils for helper functions

ayush-shah (changes requested):

ingestion/src/metadata/ingestion/bulksink/metadata_usage.py
OnkarVO7: It should get initialized to empty for each table so cannot be moved to init

ingestion/src/metadata/ingestion/bulksink/metadata_usage.py
OnkarVO7: variables after the or are different

ingestion/src/metadata/utils/life_cycle_utils.py
OnkarVO7: Wanted to put it in mixins, but in mixins we use the `REST` client. Here I want to directly call the `metadata.patch` method from mixins itself.
This is brought here only so that I can do the deep copy for the patch API

ayush-shah: we have similar thing done for `patch_test_case_definition` under patch_mixins

OnkarVO7: makes sense I'll move it under patch_mixins then

ayush-shah (approved):

PR #13313, opened 2023-09-22
Fix: Bigquery query log not picked up for multiproject

ulixius9:
### Describe your changes:

Fix: Bigquery query log not picked up for multiproject

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
+++ b/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
@@ -146,9 +146,13 @@ def get_table_entities_from_query(
     split_table = table_name.split(".")
     empty_list: List[Any] = [None]  # Otherwise, there's a typing error in the concat
 
-    database_query, schema_query, table = (
-        empty_list * (3 - len(split_table))
-    ) + split_table
+    if len(split_table) > 3:
+        table = split_table[-1]
+        database_query, schema_query = None, None
+    else:
+        database_query, schema_query, table = (
+            empty_list * (3 - len(split_table))
+        ) + split_table
 
     table_entities = search_table_entities(
         metadata=metadata,
--- a/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
+++ b/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
@@ -12,7 +12,7 @@
 Helper functions to handle SQL lineage operations
 """
 import traceback
-from typing import Any, Iterable, List, Optional
+from typing import Any, Iterable, List, Optional, Tuple
 
 from metadata.generated.schema.api.lineage.addLineage import AddLineageRequest
 from metadata.generated.schema.entity.data.table import Table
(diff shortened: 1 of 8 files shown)


ingestion/src/metadata/ingestion/lineage/sql_lineage.py
```diff
@@ -146,9 +146,13 @@ def get_table_entities_from_query(
     split_table = table_name.split(".")
     empty_list: List[Any] = [None]  # Otherwise, there's a typing error in the concat
 
-    database_query, schema_query, table = (
-        empty_list * (3 - len(split_table))
-    ) + split_table
+    if len(split_table) > 3:
```
pmbrull, line 149: what's the idea behind this?

ingestion/src/metadata/ingestion/source/database/bigquery/helper.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def get_inspector_details(
+    database_name: str, service_connection: BigQueryConnection
+) -> InspectorWrapper:
+    """
+    Method to get the bigquery inspector details
+    """
+    # TODO support location property in JSON Schema
+    # TODO support OAuth 2.0 scopes
```
pmbrull, line 45: shall we create a couple of tickets for this?

ingestion/src/metadata/ingestion/source/database/query_parser_source.py
```diff
@@ -108,6 +108,9 @@ def get_filters(self) -> str:
             return f"{self.filters} AND {self.source_config.filterCondition}"
         return self.filters
 
+    def get_engine(self):
+        yield self.engine
```
pmbrull, line 112: we would need the same logic for the Stored Procedures right? in there we are also listing the queries in the `stored_procedures_mixin`

ingestion/src/metadata/ingestion/lineage/sql_lineage.py
ulixius9: so with bigquery sometimes the table names are like this `<project-id>.<dataset-id>.infromation_schema.<table-name>` we also do the same query [here](https://github.com/open-metadata/OpenMetadata/blob/7524b673d0c4338dca49d6ebfabd9b989bf117d9/ingestion/src/metadata/ingestion/source/database/bigquery/queries.py#L48) so to avoid a value error (too many values to unpack). in case when we have more than 3 values to unpack then we just say that the last one is the table name and move ahead

ingestion/src/metadata/ingestion/source/database/bigquery/helper.py
ulixius9: Umm sure we can, also not sure if this is being asked by the community, it was there in the code I just moved it here 😅

ingestion/src/metadata/ingestion/source/database/query_parser_source.py
ulixius9: makes sense, we can do it in a separate pr as we will have to cherrypick this one to 1.1.5

ingestion/src/metadata/ingestion/lineage/sql_lineage.py
pmbrull: ah I see.. Does this only happen with `INFORMATION_SCHEMA`? or is this something more generic where we can identify db/schema/table info?

Could we please add a comment explaining the scenario and maybe a unit test?

ingestion/src/metadata/ingestion/source/database/bigquery/helper.py
pmbrull: ah ok, then @ayush-shah maybe you can chip in on what we are actually missing

ingestion/src/metadata/ingestion/source/database/query_parser_source.py
ulixius9: sure!

ingestion/src/metadata/ingestion/lineage/sql_lineage.py
ulixius9: It only happens with the `information_schema`, otherwise the dataset would contain the table directly. 

and sure I can add some comments.

ingestion/src/metadata/ingestion/source/database/bigquery/helper.py
ayush-shah: @ulixius9 @pmbrull location and oauth2 scopes to be added in bigqueryConnection if someone wants to modify those in order to connect to external resource
ref: https://github.com/open-metadata/OpenMetadata/pull/10127/files#r1100868327
cc: @yu-iskw

pmbrull (approved):

PR #13379, opened 2023-09-28
Cost Analysis - Data Insights Reports and Aggregation

OnkarVO7:
### Describe your changes:

Added reports and indexes for Cost Analysis in Data Insights workflow

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/data_insight/processor/cost_analysis_report_data_processor.py
```diff
@@ -0,0 +1,234 @@
...
+                    yield ReportData(
+                        timestamp=self.timestamp,
+                        reportDataType=ReportDataType.RawCostAnalysisReportData.value,
+                        data=cost_analysis_data,
+                    )
+
+                self.processor_status.scanned(entity.name.__root__)
+            except Exception as err:
+                logger.debug(traceback.format_exc())
+                logger.error(f"Error trying fetch cost analysis data -- {err}")
```
pmbrull, line 83: maybe we should use a `StackTraceError` here to keep track of any errors

ingestion/src/metadata/data_insight/processor/cost_analysis_report_data_processor.py
```diff
@@ -0,0 +1,234 @@
...
+
+class RawCostAnalysisReportDataProcessor(DataProcessor):
+    """Processor class used as a bridge to refine the data"""
+
+    _data_processor_type = "RawCostAnalysisReportData"
+
+    def fetch_data(self) -> Iterable[table.Table]:
+        try:
+            yield from self.metadata.list_all_entities(
+                table.Table, limit=1000, fields=["*"]
```
pmbrull, line 45: do we really need `*` in the fields? Maybe we can just ask for the minimum necessary ones to move less data, since we might already be pulling a bunch of tables

ingestion/src/metadata/data_insight/processor/cost_analysis_report_data_processor.py
```diff
@@ -0,0 +1,234 @@
...
+                logger.error(f"Error trying fetch cost analysis data -- {err}")
+
+    def process(self) -> Iterable[ReportData]:
+        yield from self.refine()
+
+    def get_status(self):
+        return self.processor_status
+
+
+class AggregatedCostAnalysisReportDataProcessor(DataProcessor):
```
pmbrull, line 92: both classes could come from the same parent, since at least the `fetch_data` is always going to be the same

ingestion/src/metadata/data_insight/processor/cost_analysis_report_data_processor.py
pmbrull: Actually we can only do this after refactoring the DI to the BaseWorkflow

openmetadata-service/src/main/resources/elasticsearch/aggregated_cost_analysis_report_data_index.json
```diff
@@ -0,0 +1,72 @@
...
+          "serviceName": {
+            "type": "keyword"
+          },
+          "serviceType": {
+            "type": "keyword"
+          },
+          "serviceOwner": {
+            "type": "keyword"
+          },
+          "unUsedDataAssets": {
```
pmbrull, line 24: should be `unusedDataAssets` since it's a single word

pmbrull:
pushed some of the changes and cleaned it up

TeddyCr (approved):

PR #13384, opened 2023-09-29
Issue 12679 - Handle Entity Object Instantiation Error + Refactor Workflow

### LINKED ISSUE #12679: Data insight report not considering all tables
**Affected module**
UI/Data insight report workflow

**Describe the bug**

Running in a daily basis the default data insight report when looking the report page results only appears or are considered a %/few tables, not all the tables (100%) of the repository). 

The rest of object stats seem correct (nº of databases, schemas and topics). 

**To Reproduce**

https://openmetadata.slack.com/files/U03PNMQDXRR/F05KHANN1C3/image.png

https://openmetadata.slack.com/files/U03PNMQDXRR/F05KHDYUVDY/image.png

https://openmetadata.slack.com/files/U03PNMQDXRR/F05L753V48Y/image.png

**Expected behavior**
100% entities considered in the report

**Version:**
 - OS: rhel
 - Python version: 3.8
 - OpenMetadata version: 1.1.0

**Additional context**

TeddyCr:
### Describe your changes:

Fixes #12679 
- refactor DI to inherit from BaseWorkflow
- Added logic to skip failure when instantiating object from client response

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/ometa/ometa_api.py
```diff
@@ -389,7 +395,17 @@ def list_entities(
...
+            entities = []
+            for elmt in resp["data"]:
+                try:
+                    entities.append(entity(**elmt))
+                except Exception as exc:
+                    logger.error(f"Error creating entity. Failed with exception {exc}")
+                    continue
+        else:
+            entities = [entity(**elmt) for elmt in resp["data"]]
+
```
TeddyCr, line 408: if skip failures we will continue instantiating other entities and skip over the ones that may not conform to the expected object shape. This will allow the DI workflow to continue processing the entity type.

IMPORTANT: this does not solve the underlying issue that is faced [here](https://github.com/open-metadata/OpenMetadata/issues/12679) for example where a tag entity has not been migrated properly to new constraints -- this is out of the DI scope.

```
Error trying to fetch entity -- 1 validation error for Table
tags -> 2 -> tagFQN -> __root__
  ensure this value has at most 128 characters (type=value_error.any_str.max_length; limit_value=128)
```

ingestion/src/metadata/data_insight/producer/entity_producer.py
```diff
@@ -0,0 +1,69 @@
...
+        databaseSchema.DatabaseSchema,
+        mlmodel.MlModel,
+        pipeline.Pipeline,
+        table.Table,
+        topic.Topic,
+        container.Container,
+    ]
+
+    # pylint: disable=protected-access,dangerous-default-value
+    def fetch_data(self, limit=100, fields=["*"]) -> Iterable:
```
pmbrull, line 51: to not have the `dangerous-default-value` we could do a `fields: Optional[List[str]] = None`, and then:

```python
yield from self.metadata.list_all_entities(
    entity, limit=limit, fields=fields or ["*"]
)
```

ingestion/src/metadata/ingestion/ometa/ometa_api.py
TeddyCr: If we need it in other places then makes sense. I had put it here as a "hidden" functionality as I thought it was an edge cases.

ingestion/src/metadata/data_insight/producer/entity_producer.py
TeddyCr: It is kind of a false flag here. We don't do operations on the we just pass it as a parameter to the endpoint. The problem with what you suggest is that you can't determine if you pass None, to have no fields (maybe you don't care about the fields or None so that you return all the fields.

pmbrull (approved):

Issue #13401, opened 2023-10-02
Refactor DI aggregate endpoint to make it more generic


TeddyCr:

harshach:
@ulixius9 @IceS2 can you check what we need to accomplish as part of this task.

ulixius9:
cloding by https://github.com/open-metadata/OpenMetadata/pull/16532

PR #13408, opened 2023-10-02
Cost analysis agg

TeddyCr:
### Describe your changes:
- Implemented aggregation logic for unused and aggunused assets reports
- added sample data

⚠️ the `aggregation` endpoint of the DI will be refactor in 1.3 to be more generic. https://github.com/open-metadata/OpenMetadata/issues/13401. We should also think of a more centralized approach for ES/OS as everything is duplicated https://github.com/open-metadata/OpenMetadata/issues/13406

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/dataInsight/DataInsightAggregatorInterface.java
```diff
@@ -16,13 +17,17 @@ public abstract class DataInsightAggregatorInterface {
...
+  protected final org.opensearch.search.SearchHits hitsOs;
+  protected final SearchHits hitsEs;
   protected final DataInsightChartResult.DataInsightChartType dataInsightChartType;
 
   protected DataInsightAggregatorInterface(
       Aggregations aggregations, DataInsightChartResult.DataInsightChartType dataInsightChartType) {
     this.aggregationsEs = aggregations;
     this.aggregationsOs = null;
     this.dataInsightChartType = dataInsightChartType;
+    this.hitsOs = null;
```
harshach, line 29: can we expand on this variable name. Its hard to understand what its meant for by looking at the name

TeddyCr: @harshach it is the hits from the ES response. You get the response like this, `hits` just contains hits. Can you give me more details as what you would expect to see as far as variable name?

```
"hits": {
    "total": {
      "value": 5,
      "relation": "eq"
    },
    "max_score": 1.0,
    "hits": [...]
  },
  "aggregations": {
    "my-agg-name": {                           
      "doc_count_error_upper_bound": 0,
      "sum_other_doc_count": 0,
      "buckets": []
    }
  }
```

harshach: @TeddyCr whats the difference between hitOs and hitEs?
what Os and Es here

TeddyCr: Oh my bad. that is OpenSearch and ElasticSearch -- just followed the naming convention that was there 🤷. I opened this issue https://github.com/open-metadata/OpenMetadata/issues/13406 to refactor that part of the code to have a more maintainable approach as right now you need to write the same code twice for OS and ES.

harshach: do we need to separate the variable. Can't we read that into the hits?

harshach: my suggestion here is to create Lombok class
public class SearchHits {
@Getter @Setter hits;
}
any other member you need in here and expect. the searchClient to return that.
I refactored bunch of code yet there is lot to go  https://github.com/open-metadata/OpenMetadata/pull/13397

TeddyCr: if I understand correctly you want to pass the `searchClient` as an attribute of `DataInsightAggregatorInterface` and have a `hits` attribute to the `searchClient` and retrieve it as `searchClient.hits` from the `DataInsightAggregatorInterface` class instance?

Apologies for the question, I am not sure I understand completely

pmbrull (approved):
we can go over the styling improvements in another PR to unblock the UI

PR #13436, opened 2023-10-04
Support migration extensions

pmbrull:
### Describe your changes:

Collate specific migrations need to be in the form of x.y.z-collate. This will help the TablesInitializer to properly sort the migration files to execute OpenMetadata migrations first.
The version will appear as usual in the SERVER_CHANGE_LOG table.

image

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/migration/api/MigrationWorkflow.java
```diff
@@ -3,6 +3,7 @@
 import java.io.File;
 import java.util.*;
```
harshach, line 4: can you remove this wild-card import

openmetadata-service/src/main/java/org/openmetadata/service/migration/api/MigrationWorkflow.java
```diff
@@ -20,16 +21,17 @@ public class MigrationWorkflow {
...
+      String nativeSQLScriptRootPath,
+      ConnectionType connectionType,
+      String extensionSQLScriptRootPath,
+      boolean forceMigrations) {
     this.jdbi = jdbi;
     this.migrationDAO = jdbi.onDemand(MigrationDAO.class);
     this.forceMigrations = forceMigrations;
     // Sort Migration on the basis of version
     List<MigrationFile> availableMigrations =
-        Arrays.stream(Objects.requireNonNull(new File(nativeSQLScriptRootPath).listFiles(File::isDirectory)))
```
harshach, line 29: we should add a unit test around this.
To test different combinations. This one is critical piece of the migrations and any issue there will blow up our migrations

harshach: intent of test is to use
1. OM bootstrap files does it do ordering correct
2. OM + Collate bootstrap files with OM-1.2.0 , COLLATE-1.2.0, COLLATE-1.2.1 another combo OM-1.2.0, OM-1.2.1 , COLLATE-1.2.2 etc.. different combinations to make sure our ordering is apply OM first, if there is equivalent COLLATE apply after. If there is no OM version but there is COLLATE version we should apply that
3. For each migration we store version that applied we need to make sure Collate version are stored as well

pmbrull: added test & screenshot. The result will be as
```
"1.1.0", "1.1.0-collate", "1.2.0", "1.2.1", "1.2.2-collate"
```

openmetadata-service/src/main/java/org/openmetadata/service/migration/api/MigrationWorkflow.java
```diff
@@ -12,24 +18,34 @@
...
+      boolean forceMigrations) {
     this.jdbi = jdbi;
     this.migrationDAO = jdbi.onDemand(MigrationDAO.class);
     this.forceMigrations = forceMigrations;
+    this.nativeSQLScriptRootPath = nativeSQLScriptRootPath;
+    this.connectionType = connectionType;
+    this.extensionSQLScriptRootPath = extensionSQLScriptRootPath;
+  }
+
+  public void loadMigrations() {
```
pmbrull, line 45: added a `loadMigrations` to be run separately instead of inside the constructor to ease the testing

openmetadata-service/src/test/java/org/openmetadata/service/OpenMetadataApplicationTest.java
```diff
@@ -54,6 +54,7 @@ public abstract class OpenMetadataApplicationTest {
   public static DropwizardAppExtension<OpenMetadataApplicationConfig> APP;
   protected static final WebhookCallbackResource webhookCallbackResource = new WebhookCallbackResource();
   public static final String FERNET_KEY_1 = "ihZpp5gmmDvVsgoOG6OVivKWwC9vd5JQ";
+  public static Jdbi jdbi;
```
pmbrull, line 57: To init the MigrationWorkflow in the tests we need the `jdbi`. LMK if there's a better way to pass this without changing the var here

harshach (approved):

PR #13447, opened 2023-10-05
Fixes #13267 - Remove maxLen and minLen from profiler default metrics

### LINKED ISSUE #13267: Min/max length calculated twice for every concatenable column during profiling
**Affected module**
Ingestion – profiler

**Describe the bug**
In the docs, `Min` and `Min Length` are described as two separate metrics. They are both on by default. `Min` is for numeric columns. `Min Length` is for string columns.

However, the implementation of `Min` and `Max` are such that they _also_ calculate the Length of string columns. 

Docs: https://docs.open-metadata.org/v1.1.x/connectors/ingestion/workflows/profiler/metrics
```
Min
Only for numerical values. Returns the minimum.

Max
Only for numerical values. Returns the maximum.

Min Length
Only for concatenable values. Returns the minimum length of the values in a column.

Max Length
Only for concatenable values. Returns the maximum length of the values in a column.
```

Implementation:
https://github.com/open-metadata/OpenMetadata/blob/e25c5968f370fd609f5517bec2369ff199a7ca3e/ingestion/src/metadata/profiler/metrics/static/min.py#L68-L69

This has two consequences

1) The default set of metrics calculates min/max length twice for every column
2) If only the `Min` and `Max` of numeric are desired, the min/max length for string columns will _also_ be calculated. There is no way to turn that off except by checking the type of every column and configuring it independently.

**To Reproduce**

Run profiler workflow on a Snowflake table that includes varchar columns

image

**Expected behavior**
`MIN` and `MAX` should _only_ operate on numeric columns, as specified in the docs. `MIN_LENGTH` and `MAX_LENGTH` should operate on concatenatble columns.

Another possibility, though less desirable imo, is to update the docs to reflect the current behavior.

**Version:**
 - Python version: 3.10
 - OpenMetadata version: 1.1.5
 - OpenMetadata Ingestion package version: `openmetadata-ingestion[snowflake,postgres]==1.1.5`

TeddyCr:
### Describe your changes:

Fixes #13267 
- max/min len of the string is already computed in the Max metrics
```python
    @_label
    def fn(self):
        """sqlalchemy function"""
            return MaxFn(LenFn(column(self.col.name, self.col.type)))
        if (not is_quantifiable(self.col.type)) and (not is_date_time(self.col.type)):
            return None
        return MaxFn(column(self.col.name, self.col.type))

```


pmbrull (approved):

pmbrull:
merging, checkstyle fixed in main

PR #13454, opened 2023-10-05
DQ base workflow

pmbrull:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/data_quality/processor/test_case_runner.py
```diff
@@ -0,0 +1,307 @@
...
+                self.config.processor.dict().get("config")
+            )
+        )
+
+    def _run(self, record: TableAndTests) -> Either:
+        # First, create the executable test suite if it does not exist yet
+        # This could happen if the process is executed from YAML and not the UI
+        if record.executable_test_suite:
+            # We pass the test suite request to the sink
+            return Either(right=record.executable_test_suite)
```
TeddyCr, line 70: Wouldn't this be handle in `source/test_suite.py` ln 172? If we return this in this step wouldn't we be just passing the testSuite to the sink?

TeddyCr (commented):
Left one comment. 😊

ingestion/src/metadata/data_quality/processor/test_case_runner.py
pmbrull: yes exactly that. In the source we return 2 possibilities:
1. First, if needed, the executable test suite to be created. In that case, the processor does not need to do anything, just pass it to the sink to get it created
2. Then, pick up the test cases (if any) and return them with the table. Now it's where the processor needs to get the cases from the CLI, execute them etc.

pmbrull: so in the case where we have a table without executable test suite, the source will first return the suite to be created, and then the rest of the logic will happen

Issue #7058, opened 2022-08-30
Enable search filter by custom property


TeddyCr:
In the search screen enable user to filter by custom properties

![image](https://user-images.githubusercontent.com/13626425/187445340-d23389e1-1415-4495-84fd-151ff34f1869.png)

PR #13496, opened 2023-10-09
Custom Property Backend Support and list approved glossary terms

07Himank:
I worked on ... because ...

FIx : #7058 

- custom property backend support.(sample query for refrence in video)
- list only approved glossary terms.

https://github.com/open-metadata/OpenMetadata/assets/112613760/c3950086-214a-4a91-a3f8-a9bb31df76c0

https://github.com/open-metadata/OpenMetadata/assets/112613760/e6b21b6d-a4b2-4534-9390-e9ef8fe88474

Refrence Video..

#
### Type of change:

- [ x] Bug fix
- [x ] Improvement


openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -354,6 +354,10 @@ public Response search(SearchRequest request) throws IOException {
       searchSourceBuilder.sort(request.getSortFieldParam(), SortOrder.fromString(request.getSortOrder()));
     }
 
+    if (request.getIndex().equalsIgnoreCase("glossary_term_search_index")) {
```
harshach, line 357: why not add this to GlossaryTermIndex.java instead of hard-coding like this?

07Himank: we need to add this query at the time when query api is called to fetch list .. thats why i thought and added it here..

harshach: better to do it is to push down to Index classes no? pass the request and each class will add additonal things to request if required if not returns the same.
You can add default implementatoin in searchIndex to return the passed in request and override where its needed

Issue #13007, opened 2023-08-25
ES Indexing for very large S3 Storage Service buckets fails


cristiancalugaru:
Affected module
UI & Ingestion framework

Describe the bug
Containers with a very large size or number of objects  (e.g. 1,037,864,944,958,702 for size in bytes and 21.474.83.647 for no. Objects) are not properly indexed into elasticsearch. Sample exception:

Failed while updating ES. Reason[Elasticsearch exception [type=mapper_parsing_exception, reason=failed to parse field [numberOfObjects] of type [integer] in document with id '4fb4835b-88f4-4a26-aca7-cb6e430690ca'. Preview of field's value: '1.7661858297E10']], Cause[ElasticsearchException[Elasticsearch exception [type=input_coercion_exception, reason=Numeric value (1.7661858297E10) out of range of int (-2147483648 - 2147483647)
at [Source: (ByteArrayInputStream); line: 1, column: 638]]]], Stack [ElasticsearchStatusException[Elasticsearch exception [type=mapper_parsing_exception, reason=failed to parse field [numberOfObjects] of type [integer] in document with id '4fb4835b-88f4-4a26-aca7-cb6e430690ca'. Preview of field's value: '1.7661858297E10']]; nested: ElasticsearchException[Elasticsearch exception [type=input_coercion_exception, reason=Numeric value (1.7661858297E10) out of range of int (-2147483648 - 2147483647)
at [Source: (ByteArrayInputStream); line: 1, column: 638]]];

To Reproduce

Ingest a bucket with a very large number of objects. When indexing, it will fail because of the elasticsearch mapping type for the numberOfObjects field

Expected behavior
The bucket information should be correctly indexes.

OpenMetadata version: 1.1.2

cristiancalugaru:
@harshach any idea if this can make it to 1.2?

harshach:
yes will make it into 1.2, let me add a sample data and see where we are missing

harshach:
@07Himank this doesn't look like fixed, can we just make it as text type in es index

07Himank:
https://github.com/open-metadata/OpenMetadata/pull/13507 @harshach we fixed this here ...

07Himank:
It works fine in latest main the error is older when the type was integer ..see if its makes sense or i can just change the type.

harshach:
@cristiancalugaru need exact number you are running into error with for us to reproduce

PR #13507, opened 2023-10-10
fixed ES Indexing for very large S3 Storage Service buckets fails

07Himank:
Fixes : #13007

- Changed numberOfObjects type from integer to double in index mapping file. as its the same in java file.
- Added a sample value in sample data and tested it . it failed first after changin it to double it passed.
- Ref :- look at the screenshots.

Screenshot 2023-10-10 at 3 14 35 PM
Screenshot 2023-10-10 at 3 27 03 PM

Screenshot 2023-10-10 at 3 10 34 PM

I worked on ... because ...fixed ES Indexing for very large S3 Storage Service buckets fails

#
### Type of change:

- [ x] Bug fix
- [ x] Improvement


ayush-shah (dismissed):

ingestion/examples/sample_data/storage/containers.json
```diff
@@ -36,7 +36,7 @@
         }
       ]
     },
-    "numberOfObjects": "50",
+    "numberOfObjects": "2147483648",
```
harshach, line 39: can we put the exact number that we have it in the ticket

07Himank: @harshach, Done works fine reindexing also works fine .. pushed in latest commit.

harshach (approved):

ayush-shah (approved):

Issue #12399, opened 2023-07-12
Assets not visible for "Child Columns" (1.1)


DovileKr:
**Affected module**
Glossary

**Describe the bug**
1. I have a table. A column of STRUCT type has 2 child columns
2.   I have tagged the table, the primary column and the child columns with Glossary terms

ONLY the Glossary Terms tagging the Table and Primary Column, display Asset Info on the Glossary Term page.

**To Reproduce**

1. Create a table with child columns. 
2.  Tag with Glossary Terms the table, the Parent and the Children
3. Check  which terms display Asset

`        {
            "name": "Parent",
            "dataType": "STRUCT",
            "dataLength": 10,
            "dataTypeDisplay": "struct",
            "description": "Customer Number. Explicit search help attachment to field ",
            "fullyQualifiedName": "testcc.ECC.LO.KNA1.Parent",
            "constraint": "NOT_NULL",
            "ordinalPosition": 3,
            "children": [
                {
                    "name": "Child",
                    "dataType": "CHAR",
                    "dataLength": 10,
                    "description": "child",
                    "fullyQualifiedName": "testcc.ECC.LO.KNA1.Parent.Child",
                    "constraint": "NOT_NULL",
                    "ordinalPosition": 1
                },
                {
                    "name": "GrandChild",
                    "dataType": "CHAR",
                    "dataLength": 10,
                    "description": "grandchild",
                    "fullyQualifiedName": "testcc.ECC.LO.KNA1.Parent.GrandChild",
                    "constraint": "NOT_NULL",
                    "ordinalPosition": 2
                }
            ]
        }`

**Expected behavior**
All terms used for tagging columns / tables should show asset

**Version:**
 - Python version:
 - OpenMetadata version: [e.g.1.0]
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

DovileKr:
Parent term tagged to column = shows asset
GrandParent term tagged to child column, assets 0

![image](https://github.com/open-metadata/OpenMetadata/assets/26750060/33ffe600-0fcb-4e3d-8b8e-8cb0cb82c66a)

DovileKr:
still a problem for 1.1

DovileKr:
Hi, do we have a roadmap fix for this bug?

DovileKr:
@harshach still not there for 1.1.3, yes?

[...]

PR #13539, opened 2023-10-12
Nested columns search fix

### LINKED ISSUE #6810: Tables with columns with children are not indexed properly
**Affected module**
UI and backend
**Describe the bug**
When tables have columns with children, the child columns (especially their descriptions) are not indexed.

**To Reproduce**
See sample json. 
1. Searching for the keywords in the child columns do not yield results
2. Advanced column search will suggest "parent.child" only sometimes and never "parent.child1.child2"
3. Advanced column search will not actually return any table hits when the "parent.child" suggestion is selected. 

**Expected behavior**
All descriptions and column names must be searchable.
Advanced column search must suggest even when the entered query is a child column (e.g., child2 should suggest parent.child1.child2)
**Version:**
 - OS: [e.g. iOS]:n/a
 - Python version: 3.9
 - OpenMetadata version: 0.11.4

**Additional context**

07Himank:
Fixes #6810 , #12399

I worked on ... because ...Nested columns search fix

#
### Type of change:

- [ x] Bug fix
- [ x] Improvement


openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -526,7 +526,9 @@ private static SearchSourceBuilder buildTopicSearchBuilder(String query, int fro
     hb.field(new HighlightBuilder.Field("messageSchema.schemaFields.description").highlighterType(UNIFIED));
     hb.field(new HighlightBuilder.Field("messageSchema.schemaFields.children.name").highlighterType(UNIFIED));
     SearchSourceBuilder searchSourceBuilder = searchBuilder(queryBuilder, hb, from, size);
-    searchSourceBuilder.aggregation(AggregationBuilders.terms(ES_MESSAGE_SCHEMA_FIELD).field(ES_MESSAGE_SCHEMA_FIELD));
+    searchSourceBuilder
+        .aggregation(AggregationBuilders.terms(ES_MESSAGE_SCHEMA_FIELD).field(ES_MESSAGE_SCHEMA_FIELD))
```
harshach, line 530: lets call the names as 
1. fieldNames
2. columnNames

openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TableIndex.java
```diff
@@ -46,22 +49,32 @@ public Map<String, Object> buildESDoc() {
     List<SearchSuggest> schemaSuggest = new ArrayList<>();
     List<SearchSuggest> databaseSuggest = new ArrayList<>();
     List<SearchSuggest> serviceSuggest = new ArrayList<>();
+    Set<List<TagLabel>> tagsWithChildren = new HashSet<>();
```
harshach, line 52: it should be a single set not set of Lists

07Himank: @harshach, we are getting List<TagLable> from functions so its done that way other wise will need loops to add it.. we are flattening it before adding to doc.

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
07Himank: @harshach ,Done pushed in latest commit

openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TableIndex.java
```diff
@@ -46,22 +49,32 @@ public Map<String, Object> buildESDoc() {
...
       }
+      doc.put("columnNames", columnsWithChildrenName);
     }
     parseTableSuggest(suggest);
     serviceSuggest.add(SearchSuggest.builder().input(table.getService().getName()).weight(5).build());
     databaseSuggest.add(SearchSuggest.builder().input(table.getDatabase().getName()).weight(5).build());
     schemaSuggest.add(SearchSuggest.builder().input(table.getDatabaseSchema().getName()).weight(5).build());
     ParseTags parseTags = new ParseTags(Entity.getEntityTags(Entity.TABLE, table));
+    tagsWithChildren.add(parseTags.getTags());
+    List<TagLabel> flattenedTagList =
```
harshach, line 74: this should be a set. Otherwise we will get duplicates

07Himank: above we are having it as a set of List<TagLabel> so it will be unique only right .. i have tested it adding duplicate values it works fine..

harshach (approved):

PR #13549, opened 2023-10-12
Add DataInsightsApp

mohityadav766:
### Describe your changes:

Add DataInsightApplication

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach (approved):

openmetadata-service/src/main/java/org/openmetadata/service/Entity.java
```diff
@@ -255,11 +255,14 @@ public static void initializeRepositories(Jdbi jdbi) {
           clz.getDeclaredConstructor().newInstance();
         } catch (Exception e) {
           try {
-            clz.getDeclaredConstructor(Jdbi.class).newInstance(jdbi);
+            clz.getDeclaredConstructor(OpenMetadataApplicationConfig.class).newInstance(config);
```
harshach, line 258: why are we changing this? why do we need to pass Config to repositories?

mohityadav766: Hey @harsha , need it to build OpenMetadataServerConnectionBuilder for insights pipeline , we could take it as input for App init () method , but then not every app needs OpenmetadataApplicationConfig, let me know if you need me to update it to pass through init method of app

harshach: ok got it, we need do something better here like switch case statement, but we can do it later

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public static AirflowConfig getDefaultAirflowConfig() {
+    return new AirflowConfig()
+        .withPausePipeline(false)
+        .withConcurrency(1)
+        .withPipelineTimezone("UTC")
+        .withRetries(3)
+        .withRetryDelay(300)
+        .withPipelineCatchup(false)
+        .withScheduleInterval("0 0 * * *")
```
harshach, line 113: why are we hardcoding this?

openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
```diff
@@ -517,7 +517,11 @@ public Response scheduleApplication(
           app,
           JdbiUnitOfWorkProvider.getInstance().getHandle().getJdbi().onDemand(CollectionDAO.class),
           searchRepository);
-      Response.status(Response.Status.OK).entity("App Scheduled to Scheduler successfully.");
+      if (app.getAppType().equals(AppType.Internal)) {
+        Response.status(Response.Status.OK).entity("App Scheduled to Scheduler successfully.");
+      } else {
+        Response.status(Response.Status.OK).entity("App is External, Use Ingestion for scheduling App pipeline.");
```
harshach, line 523: why are we returning this response here? We need to handle this by creating a ingesiton pipeline to deploy the pipeline to

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java
mohityadav766: this is default airflow config shipped with the app, like we used to do earlier as well, user can later on update this

openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
mohityadav766: ingestion pipeline creation should be done with the ingestionPipelineResource no ? although UI takes decision on the basis of appType as external to use IngestionPipelineResource , we are just handling here in case someone hits this for external Apps

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java
harshach: @mohityadav766 this should come from manifest rather than hard-coding in app

openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
harshach: when you install the app it should be deploy the pipeline with the default or user given schedule. User won't manage it differently

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java
mohityadav766: fixed

openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
mohityadav766: fixed

PR #13596, opened 2023-10-16
Fixes #13595 Consolidate changes by a user in a single session to a s…

### LINKED ISSUE #13595: Consolidate changes by a user in a single session to a single change
Currently, when users edit an entity in the UI, such as changing the description, adding new tags, etc,  each change is tracked separately, resulting in multiple version changes. This issue proposes storing the changes made by the user as he/she makes them incrementally but consolidating multiple changes made by the user in a session to a single versioned change to reduce the number of spurious version changes to an entity.

**This is especially required for knowledge articles where every time the user types and the changes are autosaved, a new version of the entity is created, resulting in 100s of different versions for minor edits.**

cc @harshach @pmbrull @mohityadav766 @TeddyCr 

### The functionality works as follows
1. When a PATCH request is received to update an entity if the previous change was made by the same user and within the session timeout (currently 10 minutes), the new updates are consolidated with the previous updates into a single update.
2. To do this, `EntityUpdater` reverts the entity from the current version to the previous version, then updates the entity from the previous version to the new updated version from the patch.
3. The changes are not consolidated if the new PATCH update comes after the session timeout or from a different user.

### Example
1. A user X updates the description of an entity. The entity goes from version v1 to v2.
2. The same user updates the display name within the session timeout. Entity updater gets an update call with `original` at v2 and `updated` with the changes to the display name.
3. EntityUpdater reverts the entity from v2 to v1. This undoes all the changes made in v2, including changes to relationships, such as `owner`, `tags`, etc. 
4. Then, the entity updater updates from the `previous` version v1 to the new `updated` version to consolidate all the changes. **Note that the entity version after consolidation remains v2.**

### Why consolidate only for PATCH?
PUT requests updates have many semantics, such as not overwriting `owner` and `description`, etc, with empty values to ensure it works well for ingestion. These semantics cause problems during the revert operation. PATCH always provides a complete entity to update, making it easy to consolidate changes.

### Some side effects of this change
When a user makes multiple changes that are consolidated, each change made results in a change event. Each time, the change consolidates the previous change. This will result in multiple Activity Feeds, Slack, and other notifications.

sureshms:
…ingle change

### Describe your changes:

Fixes #13595 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
@@ -633,10 +633,6 @@ public interface CollectionDAO {
             + "ORDER BY extension")
     List<ExtensionRecord> getExtensions(@BindUUID("id") UUID id, @Bind("extensionPrefix") String extensionPrefix);
 
-    @RegisterRowMapper(ExtensionMapper.class)
-    @SqlQuery("SELECT json FROM entity_extension WHERE id = :id AND extension = :extension " + "ORDER BY extension")
-    List<String> getExtensionJsons(@BindUUID("id") UUID id, @Bind("extension") String extension);
-
     @SqlUpdate("DELETE FROM entity_extension WHERE id = :id AND extension = :extension")
     void delete(@BindUUID("id") UUID id, @Bind("extension") String extension);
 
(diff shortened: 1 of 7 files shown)


harshach (approved):

pmbrull:
reverting this since there are many pytest failures. We'll reopen the PR so that we can validate things in there

pmbrull:
reopened the PR here https://github.com/open-metadata/OpenMetadata/pull/13610

PR #13632, opened 2023-10-18
Fix #13603: Configurable Sample Data Rows

### LINKED ISSUE #13603: Configurable Sample Data Rows & Sample Data Stored in S3
1. [ ] OpenMetadata profile upgrade to capture a maximum of 10,000 rows based on configuration
1. [ ] Allow Admin to configure S3 bucket location to write data from sampling to S3 buckets
1. [ ] UI Updates to add Sampling and S3 bucket location

-----
## Proposal

- We can configure the default number of rows to be picked up for sample data in the profiler workflow
  - We can also have a specific number of sample data rows in the table properties (same that we do with the metric sampling % / count) -> [tableProfilerConfig](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/entity/data/table.json#L711C6-L711C25)
- We are going to store the data in the object storage as a blob with `service/database/schema/table/sample.parquet`
  - We can store the data as `sample_001.parquet`, `sample_002.parquet`,... so that we can read the sample data paginated directly. Think about tradeoff of calls vs. one single read and paginate internally
- We can configure the S3 location at the service level to ensure that all profiler pipelines are pointing to the same storage location.
  - We can have that in the Advanced Configuration -> UI will need to update this
- We'll need to update the TableResource to pick up the sample data from S3 using the creds from the service, which should be stored using the Secrets Manager.
- Do we need to paginate the sample data rows if we get +100 rows? (paginate [this](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/resources/databases/TableResource.java#L511) endpoint) -> UI will need to update this
- [P2] Getting the sample data in that case can become more expensive, do we want to cache (?)

ulixius9:
### Describe your changes:

Fix #13603: Configurable Sample Data Rows

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/util/S3Util.java
```diff
@@ -0,0 +1,109 @@
...
+    return getOrDefault(s3Config.getPrefix(), "")
+        + "/"
+        + table.getService().getName()
+        + "/"
+        + table.getDatabase().getName()
+        + "/"
+        + table.getDatabaseSchema().getName()
+        + "/"
+        + table.getName()
+        + "/sample_data.json";
```
pmbrull, line 75: this should be a parquet

pmbrull: and maybe we can `"/".join(...)` with the Java equivalent. Also, not sure if we have the `/` defined as a separator anywhere, otherwise we could

openmetadata-service/src/main/java/org/openmetadata/service/util/S3Util.java
```diff
@@ -0,0 +1,109 @@
...
+      this.client.putObject(putOb, RequestBody.fromBytes(bytesToWrite));
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
+    } catch (S3Exception e) {
+      System.err.println(e.getMessage());
+    }
+  }
+
+  public TableData getSampleData(SampleDataS3Config s3Config, Table table) {
+    this.initClient(s3Config);
```
pmbrull, line 94: not sure if this is a bit dangerous. Imagine we have 2 profilers running at the same time, each pointing to a different bucket. We'd be overriding the internal client of the class.

One approach could be to define an internal `Map<S3Config, Client>` so that we pick up the right internal client in each case

openmetadata-service/src/main/java/org/openmetadata/service/util/S3Util.java
ulixius9: Sure, can do that.

ayush-shah (changes requested):

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/connectionBasicType.json
```diff
@@ -64,6 +64,29 @@
       "description": "Supports DBT Extraction.",
       "type": "boolean",
       "default": true
+    },
+    "sampleDataS3Config": {
```
pmbrull, line 68:
```json
sampleDataConfig: {
  oneOf: ["openmetadata", "S3", "ADLS']
}

```

PR #13677, opened 2023-10-22
Fixes #7858: Moved ingestion related make commands into Makefile in ingestion directory

### LINKED ISSUE #7858: Prep ingestion Makefile
Instead of putting all ingestion related work in the root project Makefile

We can then import the ingestion specific make commands into the project Makefile

TomBushell:
### Describe your changes:

Targets #7858.

I moved ingestion specific make commands in the root project Makefile into a new Makefile in the ingestion/ directory. These recipes can be run in either directory now. To do this I also modified some file paths in scripts/datamodel_code_generator.py.

#
### Type of change:

- [x] Improvement

#
### Checklist:


scripts/datamodel_generation.py
```diff
@@ -17,24 +17,30 @@
...
 )
 
 from datamodel_code_generator.__main__ import main
 
-args = "--input openmetadata-spec/src/main/resources/json/schema --input-file-type jsonschema --output ingestion/src/metadata/generated/schema --set-default-enum-member".split(
-    " "
-)
+current_directory = os.getcwd()
+ingestion_path = "./" if current_directory.endswith("OpenMetadata/ingestion") else "ingestion/"
+directory_root = "../" if current_directory.endswith("OpenMetadata/ingestion") else "./"
```
ayush-shah, line 32: @TomBushell chances are the directory would be named as something else for different users, we could keep `endswith("/ingestion")` to determine the ingestion directory, what do you think about that?

ayush-shah (changes requested):
@TomBushell also, the PR has formatting issues, if you can run the commands given in the comment above and fix it, Thanks

scripts/datamodel_generation.py
TomBushell: Yep, this will work.

ingestion/src/metadata/antlr/split_listener.py
```diff
@@ -21,10 +21,10 @@ class FqnSplitListener(FqnListener):
     def __init__(self):
         self._list = []
 
-    def enterQuotedName(self, ctx: FqnParser.QuotedNameContext):
+    def enter_quoted_name(self, ctx: FqnParser.QuotedNameContext):
```
pmbrull, line 24: we'll need to double-check why the server is unable to start. I am not sure if we might be overlooking anything by updating this method names since those are related to ANTLR.

I'd suggest to revert this changes and see if the py-tests pass properly, since the major topic in the github issue was about cleaning up the Makefile (which is beautifully done here already).

Thanks

pmbrull:
hi @TomBushell have you been able to identify why the sample data ingestion is not working? Looks like the "Sample Data DAG is not reachable".

Let us know if you have any questions on how to identify this.

Running `sh docker/run_local_docker.sh` would redo the same process that the CI is trying here.

Thanks

TomBushell:
> hi @TomBushell have you been able to identify why the sample data ingestion is not working? Looks like the "Sample Data DAG is not reachable".
> 
> Let us know if you have any questions on how to identify this.
> 
> Running `sh docker/run_local_docker.sh` would redo the same process that the CI is trying here.
> 
> Thanks

I'm not sure yet why this is happening. I have run `sh docker/run_local_docker.sh` and there are no problems.

pmbrull (approved):

pmbrull:
Ci looks good after merging `main`. Bringing this in, thanks~

PR #13690, opened 2023-10-23
upgrade: airflow base image from 2.6.3 to 2.7.2

viniciusdsmello:
### Describe your changes:

Upgrades Airflow version from 2.6.3 to 2.7.2

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/common/src/main/java/org/openmetadata/common/utils/CommonUtil.java
+++ b/common/src/main/java/org/openmetadata/common/utils/CommonUtil.java
@@ -50,6 +50,8 @@ import org.apache.commons.io.IOUtils;
 @Slf4j
 public final class CommonUtil {
 
+  private static final List<String> jarNameFilter = List.of("openmetadata", "collate");
+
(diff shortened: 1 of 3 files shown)


pmbrull:
hi, thanks. However not sure we want to bump the image to 2.7.2 just yet. We might need to give 2.7.2 some more room in case a 2.7.3 ships. Also sometimes there are issues with Airflow migrations so I am not a fan of bumping the ingestion image base too regularly. We don't use any special Airflow feature that is needed from new releases.

What I'd do:
- Wait at least a couple of more weeks before deciding if we need the new version. We mostly only bump because of security. Also we can only bump when the community helm charts support the new versions as well.
- The important part is validating if there is any code change required for 2.7.X. If so, we can add this to support a broader version range

Thanks

harshach:
@pmbrull what do we want to do here?

PR #13753, opened 2023-10-28
Fixes #13382: Replace the user ID with user name in Glossary Term

### LINKED ISSUE #13382: The UI should always display username instead of userid
In the history of a Glossary Term, the OpenMetadata UI shows the UserId instead of showing the User Name which would be much more useful.  See the instance from bellow screen capture, where we should have a user Name like "Joe Kido" instead of his user ID . 

**Additional context**
history

NormanZhang-zzt:
### Describe your changes:

Fixes  #13382

I worked on replacing the UserId with the User Name because displaying user name could be more intuitive and useful  

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


chirag-madlani:
Hi @NormanZhang-zzt Thanks for your contribution. Can you catch displayName using existing method which will use already fetched data for user details if not it will query via API
https://github.com/open-metadata/OpenMetadata/blob/b14563676af80b305483224a0efea50bbd070d09/openmetadata-ui/src/main/resources/ui/src/AppState.ts#L220

chirag-madlani:
Also can you attach screenshot of your changes will be helpful to review your PR much faster Thanks.

chirag-madlani (dismissed):

PR #13770, opened 2023-10-30
Enhancement in search in explore and internal

07Himank:
I worked on ... because ... Enhancement in search in explore and internal. more appropriate matching 

#
### Type of change:

- [ x] Improvement


harshach:
@07Himank are we merging this? why is it in draft mode?

07Himank:
> @07Himank are we merging this? why is it in draft mode?

some things were remaining for testing .. thats why it was in draft mode

PR #13807, opened 2023-11-01
Fix #13603: Configurable Sample Data Rows for Profiler

### LINKED ISSUE #13603: Configurable Sample Data Rows & Sample Data Stored in S3

ulixius9:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DatabaseRepository.java
```diff
@@ -116,6 +130,40 @@ private void populateService(Database database) {
...
+    Database database = dao.findEntityById(databaseId);
+
+    if (databaseProfilerConfig.getProfileSampleType() != null && databaseProfilerConfig.getProfileSample() != null) {
+      EntityUtil.validateProfileSample(
+          databaseProfilerConfig.getProfileSampleType().toString(), databaseProfilerConfig.getProfileSample());
+    }
+
+    daoCollection
+        .entityExtensionDAO()
+        .insert(
```
pmbrull, line 144: maybe silly question, is this actually an UPSERT?

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
```diff
@@ -85,6 +85,10 @@
       "description": "Supports Lineage Extraction.",
       "type": "boolean",
       "default": true
+    },
+    "sampleDataStorageConfig": {
+      "title": "S3 Config for Sample Data",
+      "$ref": "../connectionBasicType.json#/definitions/sampleDataStorageConfig"
```
pmbrull, line 91: should the services have only the sampleDataStorageConfig or the overall ProfilerConfig?

ingestion/src/metadata/ingestion/ometa/mixins/es_mixin.py
```diff
@@ -54,7 +54,7 @@ def _search_es_entity(
         self,
         entity_type: Type[T],
         query_string: str,
-        fields: Optional[List[str]] = None,
+        fields: Optional[str] = None,
```
pmbrull, line 57: Why we don't want to us a list? Easier to work with right?

ingestion/src/metadata/profiler/api/models.py
```diff
@@ -40,16 +43,40 @@ class ColumnConfig(ConfigModel):
...
+
+class TableConfig(BaseProfileConfig):
+    """table profile config"""
+
     profileQuery: Optional[str] = None
     partitionConfig: Optional[PartitionProfilerConfig]
     columnConfig: Optional[ColumnConfig]
 
+    @classmethod
+    def from_database_and_schema_config(
```
pmbrull, line 63: nice! Let's add one unit test around this to ensure we are not missing any weird case pls

ingestion/src/metadata/profiler/interface/profiler_interface.py
```diff
@@ -204,6 +350,35 @@ def get_profile_sample_config(
...
+        source_config: DatabaseServiceProfilerPipeline,
+    ) -> Optional[int]:
+        """_summary_
+        Args:
+            entity_config (Optional[TableConfig]): table config object from yaml/json file
+            source_config DatabaseServiceProfilerPipeline: profiler pipeline details
+        Returns:
+            Optional[int]: int
+        """
+        if entity_config and entity_config.sampleDataCount:
```
pmbrull, line 368: nit, instead of having 4 if-else, we can do a list of objects and iterate over them? Might be easier to understand what we try to achieve

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DatabaseRepository.java
ulixius9: ummm not sure on this, replicated the logic that we already had for table profiler config [here](https://github.com/open-metadata/OpenMetadata/blob/bdfc9226d5d31db18cbb846c04fe4bea48021264/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java#L320) 🤔

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
ulixius9: Ummm I think we only discussed to add the storage config, do we want to add the profiler config here?

ingestion/src/metadata/ingestion/ometa/mixins/es_mixin.py
ulixius9: Yes lists are easier to work with but since we are using LRU cache it fails to store the list in the LRU cache hashmap

ingestion/src/metadata/profiler/api/models.py
ulixius9: oh sure

ingestion/src/metadata/ingestion/ometa/mixins/es_mixin.py
pmbrull: ah damn

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/athenaConnection.json
pmbrull: ah ok then I might be remembering it wrong. If that's what we agreed then we can move on cc @TeddyCr

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DatabaseRepository.java
pmbrull: ah cool yes it takes care of it https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java#L614 sorry should've looked at it before

pmbrull (approved):

chirag-madlani (approved):

PR #13763, opened 2023-10-30
fix(ui): domain page and Asset modal issues

Ashish8689:
### Describe your changes:

- fix domain label issue in navbar dropdown list after renaming domain display name to empty
- added all aliases in a search query for count
- remove domain filter item from domain asset page
- fix unwanted api calls in domain page

### Asset Modal

- Added `All` item selector in Asset option selector
- fix the fetch call after adding any asset, the result will be based on if the filter is there or not on the fetch

Issue : 

https://github.com/open-metadata/OpenMetadata/assets/66266464/026e88b2-3857-44ec-864f-2695cc06c22f

https://github.com/open-metadata/OpenMetadata/assets/66266464/77375feb-1362-431a-bad9-eec75a48fa17

https://github.com/open-metadata/OpenMetadata/assets/66266464/1318ea29-4d6e-4747-a43f-abf0eff4418d

https://github.com/open-metadata/OpenMetadata/assets/66266464/86bce0b7-8e4a-4764-a832-7401e2be5017

Resolved : 

https://github.com/open-metadata/OpenMetadata/assets/66266464/d6a2b0c1-255b-4716-92a1-6300ff6f4ac0

image

https://github.com/open-metadata/OpenMetadata/assets/66266464/17d408a6-36d9-426d-bdb2-553aff5fcfe8

https://github.com/open-metadata/OpenMetadata/assets/66266464/143e697b-19a4-4d1a-bc09-b6ef568a31c6

https://github.com/open-metadata/OpenMetadata/assets/66266464/48efbe66-f086-45aa-bbcb-1d847cfdc3a8

https://github.com/open-metadata/OpenMetadata/assets/66266464/a20cf480-7c1a-42a4-8e3e-9c3476b010d9

image

https://github.com/open-metadata/OpenMetadata/assets/66266464/78352650-c538-49fb-8956-5200f0616c30

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

PR #13827, opened 2023-11-02
added data asset alias and a aggs to get count on index for asset count

07Himank:
I worked on ... because ... added data asset alias and a aggs to get count on index for asset count

#
### Type of change:

- [ x] Improvement


07Himank:
added changes in https://github.com/open-metadata/OpenMetadata/pull/13763 this pr

PR #13843, opened 2023-11-03
update respective index with the assets

07Himank:
I worked on ... because ... update respective index with the assets

#
### Type of change:

- [ x] Bug fix
- [ x] Improvement

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
@@ -99,8 +99,8 @@ public class DataProductRepository extends EntityRepository<DataProduct> {
   protected void postUpdate(DataProduct original, DataProduct updated) {
     super.postUpdate(original, updated);
     Map<String, EntityReference> assetsMap = new HashMap<>();
-    original.getAssets().forEach(asset -> assetsMap.put(asset.getId().toString(), asset));
-    updated.getAssets().forEach(asset -> assetsMap.put(asset.getId().toString(), asset));
+    listOrEmpty(original.getAssets()).forEach(asset -> assetsMap.put(asset.getId().toString(), asset));
+    listOrEmpty(updated.getAssets()).forEach(asset -> assetsMap.put(asset.getId().toString(), asset));
     for (EntityReference assetRef : assetsMap.values()) {
       EntityInterface asset = Entity.getEntity(assetRef, "*", Include.ALL);
       searchRepository.updateEntity(asset);
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
@@ -19,7 +19,9 @@ import static org.openmetadata.service.Entity.FIELD_ASSETS;
 import static org.openmetadata.service.util.EntityUtil.entityReferenceMatch;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
```diff
@@ -114,10 +116,14 @@ private void updateAssets() {
       // Remove assets that were deleted
       for (EntityReference asset : deleted) {
         deleteRelationship(original.getId(), DATA_PRODUCT, asset.getId(), asset.getType(), Relationship.HAS);
+        EntityInterface entityInterface = Entity.getEntity(asset, "*", Include.ALL);
+        searchRepository.updateEntity(entityInterface);
```
harshach, line 120: why are you calling this inline, why can't we do this in post updates

07Himank: Pushed the change, @karanh37 can you please verify from you end is it working as expected ?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
```diff
@@ -91,6 +95,18 @@ public void restorePatchAttributes(DataProduct original, DataProduct updated) {
     updated.withDomain(original.getDomain()); // Domain can't be changed
   }
 
+  @Override
+  protected void postUpdate(DataProduct original, DataProduct updated) {
+    super.postUpdate(original, updated);
+    Map<String, EntityReference> assetsMap = new HashMap<>();
+    original.getAssets().forEach(asset -> assetsMap.put(asset.getId().toString(), asset));
```
mohityadav766, line 102: should add a null check here or use listOrEmpty

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
karanh37: Verified on UI, changes are working fine for Data Product.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
```diff
@@ -91,6 +95,18 @@ public void restorePatchAttributes(DataProduct original, DataProduct updated) {
...
   }
 
+  @Override
+  protected void postUpdate(DataProduct original, DataProduct updated) {
+    super.postUpdate(original, updated);
+    Map<String, EntityReference> assetsMap = new HashMap<>();
+    listOrEmpty(original.getAssets()).forEach(asset -> assetsMap.put(asset.getId().toString(), asset));
+    listOrEmpty(updated.getAssets()).forEach(asset -> assetsMap.put(asset.getId().toString(), asset));
+    for (EntityReference assetRef : assetsMap.values()) {
+      EntityInterface asset = Entity.getEntity(assetRef, "*", Include.ALL);
```
harshach, line 105: why are we making this call?

harshach: This is going to penalize the post/put operations on data product as we are making additional calls. Also it looks like this will be triggered on any update to the data product such as description updates , ownership cc @mohityadav766

07Himank: @harshach , beacuse of this i was thinking and added it to the only in entitySpicificUpdates while doing updateAssets to avoid unnecessary calls. in this as we need to update the indexes of assets which are added, removed or update.

PR #13868, opened 2023-11-07
Remove glossary_mixin and change tests use standard patch

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


pmbrull (approved):

pmbrull:
btw even if we rely on the correct `patch` method (I think we added this mixin before we migrated and I forgot to translate to the new logic) it might be interesting to keep the existing methods as helpers.

I'll review why tests are failing too

harshach:
@pmbrull really don't want to give offhanded methods like this. Some of those pretty bad on how its removing a reviewer without proper contract as a method. We need to delete them and not have any of these kind off hand methods.

pmbrull:
> @pmbrull really don't want to give offhanded methods like this. Some of those pretty bad on how its removing a reviewer without proper contract as a method. We need to delete them and not have any of these kind off hand methods.

I'll add a deprecation notice on these methods for 1.2.1. We can bring this in with https://github.com/open-metadata/OpenMetadata/pull/13617

Issue #13896, opened 2023-11-08
bug: Filtering schema in database view doesn't work


mgorsk1:
**Affected module**
Search

**Describe the bug**
In database view we see a list of schemas with a possibility to search/filter them. The bug is that this list is not properly filtered and always return empty view.

**To Reproduce**

1. go into service > database view
![image](https://github.com/open-metadata/OpenMetadata/assets/5680655/c20ad3a5-9f4e-4ef7-b1ab-3645fe935620)
2. use search field to limit schemas shown
![image](https://github.com/open-metadata/OpenMetadata/assets/5680655/8501b4c7-741c-4ef5-80ba-ee2bf32f9404)

3. result is empty (bug)

**Expected behavior**
A list is filtered out properly and shows matching schemas.

**Version:**
 - Python version:
 - OpenMetadata version: 1.2.0
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

My investigation shows that:

- query to elasticsearch looks like this:

```
"query":{
      "bool":{
         "must":[
            {
               "bool":{
                  "must":[
                     {
                        "query_string":{
                           "query":"athena",
                           "fields":[
                           ],
                           "lenient":true
                        }
                     }
                  ],
                  "filter":[
                     {
                        "bool":{
                           "must":[
                              {
                                 "term":{
                                    "database.fullyQualifiedName":{
                                       "value":"service.database"
                                    }
                                 }
                              }
                           ],
                           "boost":1.0
                        }
                     }
                  ],
                  "boost":1.0
               }
            },
            {
               "term":{
                  "deleted":{
                     "value":false
                  }
               }
            }
         ],
         "boost":1.0
      }
   },
```
- **database_schema_search_index** has `database.fullyQualifiedName` field as `text` type https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/resources/elasticsearch/en/database_schema_index_mapping.json#L308
- as you see filter contains term query that operates on text field which is a bad practice in elasticsearch, according to elasticsearch docs `Avoid using the term query for [text](https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html) fields.`

I think the bug is this field should be of 'keyword' type as we search for exact matches (database schema wise).

mgorsk1:
[...]

PR #13898, opened 2023-11-08
Fixes 13896: fix fullyQualifiedName field mapping for database schema search

mgorsk1:
### Describe your changes:

Fixes [13896](https://github.com/open-metadata/OpenMetadata/issues/13896)

I worked on elasticsearch mapping fix because current one breaks functionality of searching through schemas in database view.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


mgorsk1:
btw. more thorough analysis of elasticsearch index mappings shows that fullyQualifiedName is sometimes of text type, sometimes as keyword - unless this is a deliberate choice I would suggest refactoring to store always as keyword. I didn;t want to do this in this PR as I don't know full scope of dependencies so I think this PR should be as minimal as possible to unblock the feature I've noticed broken by this.

openmetadata-service/src/main/resources/elasticsearch/en/database_schema_index_mapping.json
```diff
@@ -306,7 +306,7 @@
             }
           },
           "fullyQualifiedName": {
-            "type": "text"
+            "type": "keyword"
```
harshach, line 309: @mgorsk1 Thanks for the contribution. Can you follow the same pattern for fullyQualifiedName similar to other indexes 
```"fullyQualifiedName": {
        "type": "keyword",
        "normalizer": "lowercase_normalizer"
      },```
and also this change needed for other languages as well

07Himank (approved):

PR #13918, opened 2023-11-09
feat: limit fields used for database schema search in database view

mgorsk1:
### Describe your changes:

Fixes <issue-number>

I worked on limiting fields on which schema search is conducted within database view because currently search is done on all fields and I think it's excessive. As a user I would expect this list is filtered mostly on schema name (and perhaps description).

#### Additional Context

One example why I think it's excessive - Imagine having 'med' schema (amongst tens of different schemas in the database) - since search is done on all fields of **database_schema_search_index** that would also include tags.state field, which can take 'Confirmed' value. In such case user would type in 'med' in search box and receive tens of schemas (every schema that has at least one Confirmed tag).

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach:
@chirag-madlani @karanh37 can one of you review this.

chirag-madlani:
Hi @mgorsk1 Can you attach recording of working functionality?

chirag-madlani (approved):

PR #14098, opened 2023-11-24
Fix 14040: Part 1 Remove get_by_name calls from topology

OnkarVO7:
### Describe your changes:

Remove get_by_name calls from topology

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/api/topology_runner.py
+++ b/ingestion/src/metadata/ingestion/api/topology_runner.py
@@ -179,6 +179,9 @@ class TopologyRunnerMixin(Generic[C]):
 
     def update_context(self, stage: NodeStage, entity_name: str):
         """Append or update context"""
+        # We'll store the entity_name in the topology context instead of the entity_fqn
+        # and build the fqn on the fly wherever required.
+        # This is mainly because we need the entity_name in other places
         if stage.context and not stage.cache_all:
             self._replace_context(key=stage.context, value=entity_name)
         if stage.context and stage.cache_all:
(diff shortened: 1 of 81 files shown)


ingestion/src/metadata/ingestion/api/topology_runner.py
```diff
@@ -169,19 +170,19 @@ def fqn_from_context(self, stage: NodeStage, entity_request: C) -> str:
...
-            self.context.__dict__[dependency].name.__root__
+            self.context.__dict__[dependency]
             for dependency in stage.consumer or []  # root nodes do not have consumers
         ]
         return fqn._build(  # pylint: disable=protected-access
             *context_names, entity_request.name.__root__
         )
 
-    def update_context(self, stage: NodeStage, entity: Entity):
+    def update_context(self, stage: NodeStage, entity_name: str):
```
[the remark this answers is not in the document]
pmbrull: sounds good! We can add a comment about this in the code itself

ingestion/src/metadata/ingestion/api/topology_runner.py
```diff
@@ -197,39 +198,19 @@ def yield_and_update_context(
...
-                    fqn=entity_fqn,
-                    fields=["*"],  # Get all the available data from the Entity
-                )
-                tries -= 1
+        entity_name = model_str(right.name)
+        yield entity_request
 
         # We have ack the sink waiting for a response, but got nothing back
-        if stage.must_return and entity is None:
+        if stage.must_return and entity_name is None:
```
pmbrull, line 205: does it make sense to have the `must_return` anymore? We are not running any GET so we might as well skip this validation I guess.

Another approach which I might like a bit better is to keep a `must_return` only for important pieces, such as the Services. Those we want to be able to GET all the time otherwise there's no point continuing the ingestion

ingestion/src/metadata/ingestion/api/topology_runner.py
```diff
@@ -283,7 +264,7 @@ def _(
             )
 
         # We want to keep the full payload in the context
-        self.update_context(stage=stage, entity=right)
+        self.update_context(stage=stage, entity_name=right)
```
pmbrull, line 267: in this method we are missing removing the GET I believe

ingestion/src/metadata/ingestion/api/topology_runner.py
```diff
@@ -245,7 +226,7 @@ def _(
         lineage has been properly drawn. We'll skip the process for now.
         """
         yield entity_request
-        self.update_context(stage=stage, entity=right)
+        self.update_context(stage=stage, entity_name=right)
```
pmbrull, line 229: Maybe we can update this call. Not sure why we are passing the full `right` to the `entity_name`

ingestion/src/metadata/ingestion/api/topology_runner.py
pmbrull: and same as with the lineage, not sure why we are passing the full `right` to the entity_name

ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
```diff
@@ -281,8 +281,27 @@ def yield_datamodel_dashboard_lineage(
         if hasattr(self.context, "dataModels") and self.context.dataModels:
             for datamodel in self.context.dataModels:
                 try:
+                    datamodel_fqn = fqn.build(
+                        metadata=self.metadata,
+                        entity_type=DashboardDataModel,
+                        service_name=self.context.dashboard_service,
+                        data_model_name=datamodel,
+                    )
+                    datamodel_entity = self.metadata.get_by_name(
```
pmbrull, line 290: why are we running a GET here?

ingestion/src/metadata/ingestion/api/topology_runner.py
OnkarVO7: Added a get_by_name call only if the `must_return` attribute is kept true

ingestion/src/metadata/ingestion/api/topology_runner.py
OnkarVO7: Modified the call

pmbrull (approved):
We are still adding some GETs required for the lineage. We will address this in a followup PR caching OM's state

PR #14106, opened 2023-11-24
#14099: fix bigquery test connection & SP query

### LINKED ISSUE #14099: Bigquery optimise queries for `JOBS_BY_PROJECT` table.
**Affected module**
Ingestion Framework

**Describe the bug**
When querying the table `JOBS_BY_PROJECT` to fetch the queries for analyzing usage or lineage we are scanning the entire table which can be costly. 

The table is partitioned by `creation_time` and clustered by project id, when querying this table we should include these filters [ref](https://cloud.google.com/bigquery/docs/information-schema-jobs#schema).

queries that need to be updated.
* https://github.com/open-metadata/OpenMetadata/blob/1540329e1eceb13e4c97079a24f233b15eab5ba7/ingestion/src/metadata/ingestion/source/database/bigquery/queries.py#L41
* https://github.com/open-metadata/OpenMetadata/blob/1540329e1eceb13e4c97079a24f233b15eab5ba7/ingestion/src/metadata/ingestion/source/database/bigquery/queries.py#L73

ulixius9:
### Describe your changes:

Fix #14099: fix bigquery test connection & SP query

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/bigquery/queries.py
+++ b/ingestion/src/metadata/ingestion/source/database/bigquery/queries.py
@@ -38,7 +38,8 @@ WHERE creation_time BETWEEN "{start_time}" AND "{end_time}"
 )
 
 BIGQUERY_TEST_STATEMENT = textwrap.dedent(
-    """SELECT query FROM `region-{region}`.INFORMATION_SCHEMA.JOBS_BY_PROJECT limit 1"""
+    """SELECT query FROM `region-{region}`.INFORMATION_SCHEMA.JOBS_BY_PROJECT 
+    where creation_time > '{creation_date}' limit 1"""
 )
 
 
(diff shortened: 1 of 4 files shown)


ingestion/src/metadata/ingestion/source/database/bigquery/queries.py
```diff
@@ -81,7 +82,7 @@
     user_email as user_name
   FROM `region-{region}`.INFORMATION_SCHEMA.JOBS_BY_PROJECT
   WHERE statement_type = 'SCRIPT'
-    AND start_time >= '{start_date}'
+    AND creation_time >= '{start_date}'
```
pmbrull, line 85: is this change transparent? Do we get the same results when fetching for stored procedures?

ulixius9: `start_time field represents the time when the job transitions from the PENDING state to either RUNNING or DONE`

so IMO there should not be a huge difference between the creation_time and start_time and also this means that start_time would always be > creation_time so it would not make sense it we apply filter on both the start_time

pmbrull: sounds good, thanks. Let's keep it as this

pmbrull (approved):

PR #14175, opened 2023-11-29
ISSUE-14095: added overloaded method for method initClient to accept Feign.Builder as one of the method's parameter.

FreeDevz:
Fixes ISSUE-14095

- What changes did you make?
I added overloaded method for method initClient in org.openmetadata.client.gateway.OpenMetadata class to accept Feign.Builder as one of the method's parameter.
- Why did you make them?
To allow users to be able to configure the ApiClient, eg. set timeouts, add custom error handlers, use other Client besides OkHttpClient.
I made sure mvn compile passes on openmetadata-clients/openmetadata-java-client/pom.xml

### DIFF
--- a/openmetadata-clients/openmetadata-java-client/src/main/java/org/openmetadata/client/gateway/OpenMetadata.java
+++ b/openmetadata-clients/openmetadata-java-client/src/main/java/org/openmetadata/client/gateway/OpenMetadata.java
@@ -51,13 +51,17 @@ public class OpenMetadata {
   }
 
   public void initClient(OpenMetadataConnection config) {
-    apiClient = new ApiClient();
     Feign.Builder builder =
         Feign.builder()
             .encoder(new FormEncoder(new JacksonEncoder(apiClient.getObjectMapper())))
             .decoder(new JacksonDecoder(apiClient.getObjectMapper()))
             .logger(new Slf4jLogger())
             .client(new OkHttpClient());
+    initClient(config, builder);
+  }
+
+  public void initClient(OpenMetadataConnection config, Feign.Builder builder) {
+    apiClient = new ApiClient();
     apiClient.setFeignBuilder(builder);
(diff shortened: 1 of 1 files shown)


tinhnvgg:
Hi there,

I’m working with the 1.5.0-rc1 version of the [OpenMetadata](https://github.com/open-metadata/OpenMetadata/blob/1.5.0-rc1/openmetadata-clients/openmetadata-java-client/src/main/java/org/openmetadata/client/gateway/OpenMetadata.java) gateway client.

### Proposal for a New Constructor
Currently, there is no **no-args constructor**, and when creating an instance of `OpenMetadata`, the `initClient()` method is automatically called. However, if we want to use a custom Feign.Builder, we need to manually call another `initClient()` with both `OpenMetadataConnection` and `Feign.Builder` instances.

The method `initClient(OpenMetadataConnection config, Feign.Builder builder)` performs the same tasks as `initClient(OpenMetadataConnection config)` but with the custom `Feign.Builder`. This leads to the unnecessary initialization of apiClient twice-once with the default `Feign.Builder` and then again when overriding it with the custom one.

```java
OpenMetadataConnection connection = new OpenMetadataConnection();
// Configure the connection...
OpenMetadata openMetadata = new OpenMetadata(connection); // Creates default apiClient and Feign.Builder
Feign.Builder feignBuilder  = .... // Configure the Feign.Builder
openMetadata.initClient(connection, feignBuilder); // Reinitializes apiClient and overrides the default Feign.Builder with the custom one
```
### Suggested Improvement
To streamline this process, could we add a constructor that accepts both OpenMetadataConnection and Feign.Builder as parameters?
```java
public OpenMetadata(OpenMetadataConnection config, Feign.Builder builder) {
    apiClient = new ApiClient();
    initClient(config, builder); // Calls the existing initClient method
    validateVersion();
}
```
Optionally, we could also have another constructor with a `boolean validateVersion` parameter if needed.

Thank you for considering this enhancement.

PR #14185, opened 2023-11-30
fix: CI Workflows concurrency group names

preetsshah:
### Describe your changes:

I worked on changing the concurrency group names for a couple of workflow files - maven-postgres-tests-build-skip.yml and maven-postgres-tests-build.yml. Since these workflows had the same group name as maven-build.yml workflow. So at any given point of time, only 1 workflow can execute among the 3 (which will be the one with latest commit) and it will cancel any in-progress workflow executions (as we have set concurrency.cancel-in-progress to true). With this change, this conflict will not happen.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/.github/workflows/maven-postgres-tests-build-skip.yml
+++ b/.github/workflows/maven-postgres-tests-build-skip.yml
@@ -31,7 +31,7 @@ permissions:
   checks: write
 
 concurrency: 
-  group: maven-build-${{ github.head_ref || github.run_id }}
+  group: maven-postgres-tests-skip-build-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 jobs:
     maven-postgresql-ci:
--- a/.github/workflows/maven-postgres-tests-build-skip.yml
+++ b/.github/workflows/maven-postgres-tests-build-skip.yml
@@ -30,9 +30,6 @@ permissions:
   contents: read
   checks: write
 
-concurrency: 
-  group: maven-postgres-tests-skip-build-${{ github.head_ref || github.run_id }}
-  cancel-in-progress: true
 jobs:
     maven-postgresql-ci:
         runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/maven-postgres-tests-build-skip.yml
```diff
@@ -31,7 +31,7 @@ permissions:
   checks: write
 
 concurrency: 
-  group: maven-build-${{ github.head_ref || github.run_id }}
+  group: maven-postgres-tests-skip-build-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
```
akash-jain-10, line 35: Let's remove the concurrency from the skip tests!

.github/workflows/maven-postgres-tests-build.yml
```diff
@@ -46,7 +46,7 @@ permissions:
   checks: write
 
 concurrency: 
-  group: maven-build-${{ github.head_ref || github.run_id }}
+  group: maven-postgres-tests-build-${{ github.head_ref || github.run_id }}
```
akash-jain-10, line 49:
```suggestion
  group: maven-build-postgres-${{ github.head_ref || github.run_id }}
```

akash-jain-10 (changes requested):

.github/workflows/maven-postgres-tests-build-skip.yml
preetsshah: Removed concurrency from maven-postgres-tests-build-skip.yml

.github/workflows/maven-postgres-tests-build.yml
preetsshah: Updated this one as well.

PR #14197, opened 2023-12-01
FEAT Fix Issues in Glosary

mohityadav766:
### Describe your changes:

- [x] Add bulk api for adding assets to Glossary Term
- [x] Adds a validation api and add glossaryTags api , to update the tags attached to glossaryTerm and also attaching them to Assets
- [x] Removes Relationship storing for Derived Tags and build on READ
- [ ] ML Model has tags issue as it does not create any relationship and stores in field
- [ ] Should we be also looking at column tags during for attaching ? cc @harshach 

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/csv/EntityCsv.java
```diff
@@ -435,11 +435,11 @@ protected void importFailure(CSVPrinter printer, String failedReason, CSVRecord
   }
 
   private void setFinalStatus() {
-    Status status = Status.FAILURE;
+    ApiStatus status = ApiStatus.FAILURE;
```
harshach, line 438: @mohityadav766 any reason to change this?

PR #14204, opened 2023-12-01
FIXES #14049 - Split test case resolution status from test case result

### LINKED ISSUE #14049: Separate status from test case result
### Context
Failure status is currently stored alongside a `testCaseResult`. For each failure a new failure status is created and stored with the `testCaseResult` -- see below.
https://github.com/open-metadata/OpenMetadata/blob/d55100592d250f15a200410ceea91bb91ab13913/openmetadata-spec/src/main/resources/json/schema/tests/basic.json#L89-L150

To prevent re writing the test case failure status to a testCaseResult, we'll be separating failure status as their own entities and link it to a testCaseResult (if status of testCaseResult if different than resolved).

### A.C.
- Separate failure status into its own extension. Should be a time series extension linking failureStatus [Many] <-> [One] testCase
- Create `assign` and `review` statuses
- Enable listing of failure status by:
     - status
     - test case
     - assignee -> this should be an entity reference
     - reviewer -> this should be an entity reference

TeddyCr:
### Describe your changes:

Fixes #14049 
 
**Context**
Test case resolution has been linked to the test case result itself. In this PR we split the test case resolution from the test case result into its own timeseries entity -- test case results keep a reference to an open test case resolution on failure.

- Implement CRUD operations for test case resolution
- Implement logic to index and retrieve test case resolution entities
- Implement logic to link test case failure with a test case resolution
- Added sample test case resolution data

**IMPORTANT**
The UI changed have been made to keep the UI functioning though this is a temporary change. Currently the test case resolution is not implemented from a UI point of view cc @ShaileshParmar11.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Improvement

- [x] I have added the tag `Backward-Incompatible-Change`.


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -233,7 +229,33 @@ public RestUtil.PutResponse<?> addTestCaseResult(
     return new RestUtil.PutResponse<>(Response.Status.CREATED, changeEvent, RestUtil.ENTITY_FIELDS_CHANGED);
   }
 
-  public RestUtil.PutResponse<?> deleteTestCaseResult(String updatedBy, String fqn, Long timestamp) {
+  private TestCaseResolutionStatus setTestCaseFailureStatus(TestCase testCase, String updatedBy) {
+    String json =
+        daoCollection
+            .testCaseResolutionStatusTimeSeriesDao()
+            .getLatestExtension(testCase.getFullyQualifiedName(), TESTCASE_RESULT_EXTENSION);
```
harshach, line 236: We can improve this query by looking for non-resolved test case resolution status only. 
I can imagine if the test case going through its life cycle there may be multiple statuses here and we may pull all of them in this query?

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResolutionStatusResource.java
```diff
@@ -0,0 +1,292 @@
...
+    filter.addQueryParam("testCaseResolutionStatusType", testCaseResolutionStatusType);
+    filter.addQueryParam("assignee", assignee);
+    filter.addQueryParam("reviewer", reviewer);
+    filter.addQueryParam("entityFQNHash", testCaseFQN);
+
+    return repository.list(offset, startTs, endTs, limitParam, filter, latest);
+  }
+
+  @GET
+  @Path("/sequenceId/{sequenceId}")
```
harshach, line 149: Can we use id or stateId( as in the state machine step) . SequenceID has a different understanding of it.

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResolutionStatusResource.java
```diff
@@ -0,0 +1,292 @@
...
+            content =
+                @Content(
+                    mediaType = "application/json",
+                    schema = @Schema(implementation = CreateTestCaseResolutionStatus.class)))
+      })
+  public Response create(
+      @Context UriInfo uriInfo,
+      @Context SecurityContext securityContext,
+      @Valid CreateTestCaseResolutionStatus createTestCaseFailureStatus) {
+    OperationContext operationContext = new OperationContext(Entity.TEST_CASE, MetadataOperation.EDIT_TESTS);
```
harshach, line 218: We should also check the ownership of the test case.
Also what happens if another user wants to open a ticket on a test case failure. 
They may not have test case edit option

openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TestCaseResolutionStatusIndex.java
```diff
@@ -0,0 +1,33 @@
...
+
+  @Override
+  public Map<String, Object> buildESDoc() {
+    Map<String, Object> doc = JsonUtils.getMap(testCaseResolutionStatus);
+    return doc;
+  }
+
+  public static Map<String, Float> getFields() {
+    Map<String, Float> fields = new HashMap<>();
+    fields.put("testCaseResolutionStatusType", 15.0f);
```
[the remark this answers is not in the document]
TeddyCr: Ah yes, I was thinking of it as a filter here, make sense.

harshach (approved):

PR #14224, opened 2023-12-04
Fix 14040: Part 2 Add Patch Request when updating the entities from create request

OnkarVO7:
### Describe your changes:

Add Patch Request when updating the entities from create request
Add sourceHash to entities and compare it while updating the entites

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/utils/source_hash_utils.py
+++ b/ingestion/src/metadata/utils/source_hash_utils.py
@@ -31,4 +31,4 @@ def generate_source_hash(create_request: C, exclude_fields: Optional[Dict]) -> s
     create_request_json = create_request.json(exclude=exclude_fields)
 
     json_bytes = create_request_json.encode("utf-8")
-    return hashlib.sha256(json_bytes).hexdigest()
+    return hashlib.md5(json_bytes).hexdigest()
--- a/ingestion/src/metadata/utils/source_hash_utils.py
+++ b/ingestion/src/metadata/utils/source_hash_utils.py
@@ -1,0 +1,34 @@
+#  Copyright 2021 Collate
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  http://www.apache.org/licenses/LICENSE-2.0
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""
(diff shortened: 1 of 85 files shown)


openmetadata-spec/src/main/resources/json/schema/entity/data/dashboard.json
```diff
@@ -141,6 +141,12 @@
     "lifeCycle": {
       "description": "Life Cycle properties of the entity",
       "$ref": "../../type/lifeCycle.json"
+    },
+    "sourceHash": {
+      "description": "Source hash of the entity",
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 64
```
harshach, line 149: 16 bytes might be good enough?

openmetadata-spec/src/main/resources/json/schema/entity/services/pipelineService.json
```diff
@@ -188,6 +188,12 @@
     "domain" : {
       "description": "Domain the Pipeline service belongs to.",
       "$ref": "../../type/entityReference.json"
+    },
+    "sourceHash": {
+      "description": "Source hash of the entity",
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 64
```
harshach, line 196: do we need this for services? all we need is actual data that is coming from the sources such as tables/ topics/ dashboards etc..

openmetadata-spec/src/main/resources/json/schema/entity/data/dashboard.json
OnkarVO7: @harshach I'm currently using sha256 to generate the hash of 64 char length. I'll use md5 instead which uses 32 chars. Let me know if that sounds ok

openmetadata-spec/src/main/resources/json/schema/entity/services/pipelineService.json
OnkarVO7: Makes sense I'll remove the field from service entities

ingestion/src/metadata/utils/source_hash_utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def generate_source_hash(create_request: C, exclude_fields: Optional[Dict]) -> str:
+    """
+    Given a create_request model convert it to json string and generate a hash value
+    """
+
+    create_request_json = create_request.json(exclude=exclude_fields)
+
+    json_bytes = create_request_json.encode("utf-8")
```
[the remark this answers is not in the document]
OnkarVO7: Yes it will work. If a new field is introduced, the hash won't match and a patch request for the new field will be created.

ulixius9 (approved):

PR #14345, opened 2023-12-12
#13445 Profiler Graph Should not Show a Point if value is None

### LINKED ISSUE #13445: Profiler Graph Should not Show a Point if value is None
**Context**
When no data is returned by the profiler we currently set the point to 0, we should instead don't show a point - similar to what is done in the Data Insight.

Screenshot 2023-10-04 at 5 27 26 PM

ShaileshParmar11:
### Describe your changes:

Fixes #13445

- refactored the logic to generate profiler metrics
- updated logic to not show points in the graph if data is not available.
- added unit test for changes

https://github.com/open-metadata/OpenMetadata/assets/71748675/c09f2498-aff0-483d-be76-9ea649570a97

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/utils/TableProfilerUtils.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/TableProfilerUtils.ts
@@ -11,12 +11,17 @@
  *  limitations under the License.
  */
 
-import { findLast, isUndefined, sortBy } from 'lodash';
+import { findLast, isUndefined, last, sortBy } from 'lodash';
 import { MetricChartType } from '../components/ProfilerDashboard/profilerDashboard.interface';
 import { SystemProfile } from '../generated/api/data/createTableProfile';
 import { Table, TableProfile } from '../generated/entity/data/table';
 import { CustomMetric } from '../generated/tests/customMetric';
 import { customFormatDateTime } from './date-time/DateTimeUtils';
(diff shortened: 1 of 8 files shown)


openmetadata-ui/src/main/resources/ui/src/utils/TableProfilerUtils.ts
```diff
@@ -76,7 +81,7 @@ export const calculateSystemMetrics = (
     return {
       ...item,
       stackId: stackId,
-      latestValue: operation?.rowsAffected ?? 0,
+      latestValue: operation?.rowsAffected,
```
Sachin-chaurasiya, line 84: What if the latest value is undefined? IMO we should have the default value?

Sachin-chaurasiya (commented):
Left a few comments, thanks

openmetadata-ui/src/main/resources/ui/src/utils/TableProfilerUtils.ts
ShaileshParmar11: That's the main change we wanted as per ticket, if api doesn't have data, we will not show graph/point in graph.

Sachin-chaurasiya (approved):

chirag-madlani (approved):

PR #14386, opened 2023-12-14
Fixes #14331 - Add Test Case Incident Manager Migration

### LINKED ISSUE #14331: Migrate existing incident status to the new API.
We'll take the latest failure to create the new one (for open ones i.e. non-resolve) and will migrate the resolved ones

TeddyCr:
### Describe your changes:

Fixes #14331 

- Take the latest test case failure status and ingest it as a test case incident.
- For test case result where we previously attached the status to the result we will only keep the latest status and discard the other ones as we introduced the concept of stateId with incident manager.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


bootstrap/sql/migrations/native/1.3.0/mysql/postDataMigrationSQLScript.sql
```diff
@@ -5,4 +5,45 @@ SET json = REPLACE(json, '"customMetricsProfile"', '"customMetrics"');
 -- Delete customMetricsProfile from entity_extension
 -- This was not supported on the processing side before 1.3.
 DELETE FROM openmetadata_db.entity_extension ee   
-where extension  like '%customMetrics';
\ No newline at end of file
+where extension  like '%customMetrics';
+
+-- BEGIN: Incident Manager Migration
+-- STEP 1: Update test case testCaseResult.testCaseFailureStatus field
+UPDATE test_case tc
```
harshach, line 12: should we just delete any existing references to resolution instead of migrating?

TeddyCr: We can, if we do so should then not migrate anything and just deleted existing reference?

harshach: thats what I am thinking, so that we don't do any big migration. IMO not sure how many users may be using it. If you not heard from any user in community support or customers. My suggestion is to drop everything related to resolution and bring the new functionality

TeddyCr: I have not. Will do

harshach (approved):

Issue #13823, opened 2023-11-02
Ingestion Pipeline pydantic validations w/ airflow apis


pmbrull:
If we have misalignment in the ingestion pipeline JSON file that gets generated for the ingestion from OM UI and the pydantic models, the errors are hard to understand:

```
 IngestionPipeline expected: 170 validation errors for IngestionPipeline
sourceConfig -> config -> viewParsingTimeoutLimit
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> type
  value is not a valid enumeration member; permitted: 'DatabaseUsage' (type=type_error.enum; enum_values=[<DatabaseUsageConfigType.DatabaseUsage: 'DatabaseUsage'>])
sourceConfig -> config -> databaseFilterPattern
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> includeStoredProcedures
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> includeTables
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> includeTags
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> includeViews
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> markDeletedTables
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> queryParsingTimeoutLimit
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> schemaFilterPattern
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> tableFilterPattern
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> useFqnForFiltering
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> viewParsingTimeoutLimit
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> type
  value is not a valid enumeration member; permitted: 'DatabaseLineage' (type=type_error.enum; enum_values=[<DatabaseLineageConfigType.DatabaseLineage: 'DatabaseLineage'>])
sourceConfig -> config -> includeStoredProcedures
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> includeTables
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> includeTags
  extra fields not permitted (type=value_error.extra)
sourceConfig -> config -> includeViews
  extra fields not permitted (type=value_error.extra)
[...]

PR #14461, opened 2023-12-20
Fix 13823: Validate and Parse gracefully IngestionPipeline

OnkarVO7:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/13823

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/api/parser.py
```diff
@@ -434,3 +437,38 @@ def parse_test_connection_request_gracefully(
...
+
+    :param config_dict: JSON ingestion pipeline config
+    :return:Ingestion Pipeline config or scoped error
+    """
+
+    try:
+        ingestion_pipeline = IngestionPipeline.parse_obj(config_dict)
+        return ingestion_pipeline
+
+    except ValidationError:
```
pmbrull, line 458: should we take care of the `serviceConnection`? Not only the `sourceConfig`

pmbrull (changes requested):

ingestion/src/metadata/ingestion/api/parser.py
pmbrull: I believe that's where we get also a lot of noise, for example when we configure an URI without the scheme or something

pmbrull: as discussed, that was the wrong comment. We'll be adding it to the run_automations endpoint though

ingestion/src/metadata/ingestion/api/parser.py
```diff
@@ -400,37 +406,84 @@ def parse_workflow_config_gracefully(
     raise ParsingConfigurationError("Uncaught error when parsing the workflow!")
 
 
-def parse_test_connection_request_gracefully(
```
OnkarVO7, line 403: Removing the method since it wasn't being used anywhere except for tests.

pmbrull (approved):

PR #14507, opened 2023-12-27
MINOR - generic profiler optimization for sampling and BQ

### LINKED ISSUE #14433: High Query Cost using Bigquery TABLE_STORAGE
We have identified the issue related to Bigquery profiling over table storage for fetching the system level tags like no. of rows, sizeInBytes, creation time, col names and col count.

Instead of using table storage from information schema which can be huge, we can use `__TABLES__` to get metadata for that table.

ref: https://openmetadata.slack.com/archives/C02B6955S4S/p1702021485941039

TeddyCr:
### Describe your changes:
- Fixes #14433 - we'll keep the current method as a fall back in case table metadata cannot be access from `__TABLES__`
- Limit sampling to only the column being profiled (should improve execution and limit cost for columnar database)
- Default to 1 DAY for BQ partitioned tables

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/utils/partition.py
```diff
@@ -47,8 +48,10 @@ def get_partition_details(entity: Table) -> Optional[PartitionProfilerConfig]:
                 return PartitionProfilerConfig(
                     enablePartitioning=True,
                     partitionColumnName=entity.tablePartition.columns[0],
-                    partitionIntervalUnit=entity.tablePartition.interval,
-                    partitionInterval=30,
+                    partitionIntervalUnit=PartitionIntervalUnit.DAY
+                    if entity.tablePartition.interval != "HOUR"
+                    else entity.tablePartition.interval,
+                    partitionInterval=1,
```
TeddyCr, line 54: Initially we use whatever is defined in BQ table. Though, we will now force it to DAY by default, but if it is set to HOUR on the table side we should not increase the partition interval.

ingestion/src/metadata/profiler/interface/sqlalchemy/bigquery/profiler_interface.py
```diff
@@ -32,6 +32,9 @@ def _get_struct_columns(self, columns: dict, parent: str):
         for key, value in columns.items():
             if not isinstance(value, STRUCT):
                 col = Column(f"{parent}.{key}", value)
+                # pylint: disable=protected-access
+                col._set_parent(self.table.__table__)
```
pmbrull, line 36:
```suggestion
                col._set_parent(self.table.__table__)  # pylint: disable=protected-access
```

Would this be the same? or does it get too long

ingestion/tests/unit/topology/database/test_deltalake.py
```diff
@@ -1,226 +1,226 @@
...
-            tableType=TableType.Regular,
-            description="testing around",
-            columns=expected_columns,
-            tableConstraints=None,
-            databaseSchema=MOCK_DATABASE_SCHEMA.fullyQualifiedName,
-            viewDefinition=None,
-        )
-
-        self.assertEqual(table_request, expected_table_request)
+# #  Copyright 2021 Collate
```
pmbrull, line 1: why do we need to comment this test?

ingestion/src/metadata/profiler/interface/sqlalchemy/bigquery/profiler_interface.py
TeddyCr: It gets too long

ingestion/tests/unit/topology/database/test_deltalake.py
TeddyCr: oh shit, that is my bad.

Issue #11799, opened 2023-05-29
Add owners for pipeline tasks


gnomolio:
**Is your feature request related to a problem? Please describe.**
We need to account Airflow tasks, but it's possible to define owners only for DAGs (ingested as pipelines to OM)

**Describe the solution you'd like**
Add ownership to pipeline tasks.

**Describe alternatives you've considered**
None.

pmbrull:
@harshach for this we'll need to think how we want to show that info on the UI, since tasks for now are just a list view in the pipeline details

harshach:
we will show it as another column under the list view

pmbrull:
UI pending @chirag-madlani

PR #14510, opened 2023-12-27
#11799 - Fix Airfow ownership & add pipeline tasks

pmbrull:
### Describe your changes:
Related to #11799 

Adding ownership to tasks and using `field_relationship` table to store this info. We're deleting the relationship if a task gets deleted or if the full pipeline is deleted

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/ometa/mixins/user_mixin.py
+++ b/ingestion/src/metadata/ingestion/ometa/mixins/user_mixin.py
@@ -39,6 +39,28 @@ class OMetaUserMixin:
         + ES_INDEX_MAP[User.__name__]
     )
 
+    # Allow for more flexible name lookup
+    name_search = (
+        "/search/query?q={name}&from={from_}&size={size}&index="
+        + ES_INDEX_MAP[User.__name__]
+    )
+
+    def _get_user_by_es(self, query_string: str, fields: Optional[list] = None) -> Optional[User]:
+        """Fetch user information via ES"""
+
+        try:
+            entity_list = self._search_es_entity(
+                entity_type=User, query_string=query_string, fields=fields
+            )
+            for user in entity_list or []:
+                return user
+        except Exception as err:
+            logger.debug(traceback.format_exc())
+            logger.warning(
+                f"Could not get user info from ES due to {err}"
+            )
+
+
     @lru_cache(maxsize=None)
     def get_user_by_email(
         self,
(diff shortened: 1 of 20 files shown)


ingestion/src/metadata/ingestion/ometa/mixins/user_mixin.py
```diff
@@ -39,6 +39,29 @@ class OMetaUserMixin:
         + ES_INDEX_MAP[User.__name__]
     )
 
+    # Allow for more flexible name lookup
+    name_search = (
```
harshach, line 43: @pmbrull there is fieldQuery api as well, where you can pass fieldName and fieldValue in this case name=User.name will get you exact match result

ingestion/src/metadata/ingestion/ometa/mixins/user_mixin.py
```diff
@@ -39,6 +39,29 @@ class OMetaUserMixin:
...
+        "/search/query?q={name}&from={from_}&size={size}&index="
+        + ES_INDEX_MAP[User.__name__]
```
harshach, line 45: we should look up both user and team right?

ingestion/src/metadata/ingestion/ometa/mixins/user_mixin.py
```diff
@@ -39,6 +39,29 @@ class OMetaUserMixin:
...
+    )
+
+    def _get_user_by_es(
+        self, query_string: str, fields: Optional[list] = None
+    ) -> Optional[User]:
+        """Fetch user information via ES"""
+
+        try:
```
harshach, line 53: this can be replaced by fieldQuery as well

ingestion/src/metadata/ingestion/ometa/mixins/user_mixin.py
pmbrull: re the `name=User.name` wanted to keep it same as the UI is doing when searching for users. E.g., if the user name is random.user but it is defined in Airflow as Random User , if I used the exact match I think it was not giving me back the record

ingestion/src/metadata/ingestion/ometa/mixins/user_mixin.py
pmbrull: adding it

harshach (approved):

PR #14511, opened 2023-12-27
FIXES #14396: Exclude empty test suites from quality page

### LINKED ISSUE #14396: Data Quality page displays tables without tests
The issue started with ver 1.2.0 and persists after updating to 1.2.3

We have several tables on our Data Quality page that don't and won't have tests on them.
The table may have had tests on them in the past but were deleted

image

TeddyCr:
### Describe your changes:

Fixes #14396 

Add parameter to exclude test suite without test cases from listing

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Improvement

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java
@@ -58,6 +58,7 @@ public class ListFilter {
     condition = addCondition(condition, getTestCaseCondition());
     condition = addCondition(condition, getTestSuiteTypeCondition());
     condition = addCondition(condition, getTestSuiteFQNCondition());
+    condition = addCondition(condition, getEmptyTestSuiteCondition());
     condition = addCondition(condition, getDomainCondition());
     condition = addCondition(condition, getEntityFQNHashCondition());
     condition = addCondition(condition, getTestCaseResolutionStatusType());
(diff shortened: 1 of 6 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestSuiteResource.java
```diff
@@ -125,6 +125,12 @@ public ResultList<TestSuite> list(
               schema = @Schema(type = "string", example = "executable"))
           @QueryParam("testSuiteType")
           String testSuiteType,
+      @Parameter(
+              description = "Include empty test suite in the response.",
```
mohityadav766, line 129: any specific reason @TeddyCr , we should always list the testSuites empty or not if the satisfy other conditions .. no ?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java
```diff
@@ -227,6 +228,20 @@ private String getTestSuiteTypeCondition() {
...
   }
 
+  private String getEmptyTestSuiteCondition() {
+    String includeEmptyTestSuites = getQueryParam("includeEmptyTestSuites");
+    if (includeEmptyTestSuites == null || Boolean.parseBoolean(includeEmptyTestSuites)) {
+      // if we want to include empty test suites, then we don't need to add a condition
+      return "";
+    }
+
+    if (Boolean.TRUE.equals(DatasourceConfig.getInstance().isMySQL())) {
```
harshach, line 238: @TeddyCr why are we looking at test results to filter out. The goal a test suite has 3 test cases on day 1 , later point someone deleted all those tests cases , now we have empty test suite. Instead of listing it in the DQ portal, we will not show them.
Currently if I create a test suite with a test case and not run it or Yet to run. I.e if I schedule it for next day , it won't get listed in DQ page

harshach: and we have different definition of what includeEmptyTestSuite means in case of TableList https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java#L1897

Issue #14369, opened 2023-12-13
Custom order in custom properties


AntonioChen0206:
**Is your feature request related to a problem? Please describe.**
We add a lot of custom properties and the user might want the list sorted in alphabetical order, Ascending or Descending?

**Describe the solution you'd like**
Design a button beside the title "NAME" to toggle Ascending or Descending?

**Describe alternatives you've considered**
NA

**Additional context**
NA

harshach:
@chirag-madlani can we do this on the UI side?

chirag-madlani:
@harshach Yes we can have it for table column

karanh37:
Fixed with #14537

PR #14537, opened 2024-01-02
fix(#14369): added sort option in custom property table

Abhishek332:
### **Before** 
https://github.com/open-metadata/OpenMetadata/assets/80886271/e65a8850-2a42-48db-b6fc-bc8a800c92cf

### **After** 
https://github.com/open-metadata/OpenMetadata/assets/80886271/5f556462-53d2-4756-8841-5b2177f563a9

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/CustomEntityDetail/CustomPropertyTable.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/CustomEntityDetail/CustomPropertyTable.tsx
@@ -23,7 +23,7 @@ import { CUSTOM_PROPERTIES_DOCS } from '../../constants/docs.constants';
 import { NO_PERMISSION_FOR_ACTION } from '../../constants/HelperTextUtil';
 import { ERROR_PLACEHOLDER_TYPE, OPERATION } from '../../enums/common.enum';
 import { CustomProperty } from '../../generated/entity/type';
-import { getEntityName } from '../../utils/EntityUtils';
+import { columnSorter, getEntityName } from '../../utils/EntityUtils';
 import RichTextEditorPreviewer from '../common/RichTextEditor/RichTextEditorPreviewer';
 import ConfirmationModal from '../Modals/ConfirmationModal/ConfirmationModal';
 import { ModalWithMarkdownEditor } from '../Modals/ModalWithMarkdownEditor/ModalWithMarkdownEditor';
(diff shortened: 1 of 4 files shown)


openmetadata-ui/src/main/resources/ui/src/components/CustomEntityDetail/CustomPropertyTable.tsx
```diff
@@ -89,6 +89,7 @@ export const CustomPropertyTable: FC<CustomPropertyTableProp> = ({
         dataIndex: 'name',
         key: 'name',
         render: (_, record) => getEntityName(record),
+        sorter: (col1, col2) => col1.name.localeCompare(col2.name),
```
chirag-madlani, line 92: We are displaying `displayName ?? name` property from data and sorting over name only, I feel it will be confusing for user

We can use something like below
```js
const name1 = getEntityName(col1);
const name2 = getEntityName(col2);
return name1.localeCompare(name2);
```
Also we have internationalisation support in the application, Can you check if this will work well with other langs as well?

openmetadata-ui/src/main/resources/ui/src/components/common/CustomPropertyTable/CustomPropertyTable.tsx
```diff
@@ -172,6 +172,7 @@ export const CustomPropertyTable = <T extends ExtentionEntitiesKeys>({
         key: 'name',
         width: 200,
         render: (_, record) => getEntityName(record),
+        sorter: (col1, col2) => col1.name.localeCompare(col2.name),
```
chirag-madlani, line 175: same

openmetadata-ui/src/main/resources/ui/src/components/CustomEntityDetail/CustomPropertyTable.tsx
```diff
@@ -89,7 +89,12 @@ export const CustomPropertyTable: FC<CustomPropertyTableProp> = ({
         dataIndex: 'name',
         key: 'name',
         render: (_, record) => getEntityName(record),
-        sorter: (col1, col2) => col1.name.localeCompare(col2.name),
+        sorter: (col1, col2) => {
+          const name1 = getEntityName(col1);
+          const name2 = getEntityName(col2);
+
+          return name1.localeCompare(name2);
+        },
```
chirag-madlani, line 97: Let's create one util sorter and pass it here as callback function to avoid duplicate code

Abhishek332: done

chirag-madlani (approved):

PR #14541, opened 2024-01-02
MINOR - Create Test Case Resolution ts entry & delete resolution when…

pmbrull:
… Test Case is deleted

### Describe your changes:

- When a test failed, we added the `testCaseResolutionStatusReference` field, but it did not add the entry on `test_case_resolution_status_time_series` time series table
- When deleting a test case, we were not cleaning up the `test_case_resolution_status_time_series` entries

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
@@ -3592,6 +3592,18 @@ public interface CollectionDAO {
     default String getTimeSeriesTableName() {
       return "data_quality_data_time_series";
     }
+
+    @ConnectionAwareSqlUpdate(
+        value =
+            "SELECT json FROM data_quality_data_time_series where entityFQNHash = :entityFQNHash "
+                + "AND JSON_EXTRACT(json, '$.incidentId') IS NOT NULL",
+        connectionType = MYSQL)
+    @ConnectionAwareSqlUpdate(
+        value =
+            "SELECT json FROM data_quality_data_time_series where entityFQNHash = :entityFQNHash "
+                + "AND json ->> 'incidentId' IS NOT NULL",
+        connectionType = POSTGRES)
+    List<String> getResultsWithIncidents(@Bind("entityFQNHash") String entityFQNHash);
   }
 
   interface TestCaseResolutionStatusTimeSeriesDAO extends EntityTimeSeriesDAO {
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
@@ -1074,6 +1074,7 @@ public interface CollectionDAO {
         @BindList("teamIds") List<String> teamIds,
         @Define("condition") String condition);
 
+    // TODO: UPDATE ME SHOULD BE THE STATE_ID somewhere here
     @ConnectionAwareSqlQuery(
         value =
             "SELECT json FROM thread_entity "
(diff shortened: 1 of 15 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -673,6 +709,7 @@ public FeedRepository.TaskWorkflow getTaskWorkflow(FeedRepository.ThreadContext
     return super.getTaskWorkflow(threadContext);
   }
 
+  // TODO: Is this needed? cc @Harsha
   public static class TestCaseFailureResolutionTaskWorkflow extends FeedRepository.TaskWorkflow {
```
pmbrull, line 713: @harshach since we are handling the task status when adding new status in the Incident Manager, not sure if these methods here are necessary anymore. Don't really know how these workflows come into play

pmbrull: but the task flow happens when updating the Incident Manager status right? On my mind:
- When a test fails, we create a `NEW` incident and open a task to the test owner about "New Incident"
- The owner can `ACK` the incident
- The owner can move the Incident to `Assigned`. This will close the initial task and open a task for the assignee about "You should fix this"
- When the assignee/owner move the Incident to Resolved, the final task gets closed

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -3592,6 +3592,32 @@ interface DataQualityDataTimeSeriesDAO extends EntityTimeSeriesDAO {
     default String getTimeSeriesTableName() {
       return "data_quality_data_time_series";
     }
+
+    @ConnectionAwareSqlQuery(
+        value =
+            "SELECT json FROM data_quality_data_time_series where entityFQNHash = :testCaseFQNHash "
```
harshach, line 3598: its worth promoting incidentId as field at top level and add a index insted of doing JSON_EXTRACT

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseResolutionStatusRepository.java
```diff
@@ -96,121 +97,205 @@ private void validatePatchFields(
...
+    return incident != null
+        && incident.getTestCaseResolutionStatusType() != TestCaseResolutionStatusTypes.Resolved;
+  }
+
+  private Thread getIncidentTask(TestCaseResolutionStatus incident) {
+    // Fetch the latest task (which comes from the NEW state) and close it
+    String jsonThread =
+        Entity.getCollectionDAO()
+            .feedDAO()
+            .fetchThreadByTestCaseResolutionStatusId(incident.getStateId());
```
harshach, line 110: I hope we are encoding this in entityLink such that we can retrieve all the incidents(tasks) that belong to a test case along with a particular incident

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -487,6 +507,26 @@ public ResultList<TestCaseResult> getTestCaseResults(String fqn, Long startTs, L
         testCaseResults, String.valueOf(startTs), String.valueOf(endTs), testCaseResults.size());
   }
 
+  /**
+   * Check all the test case results that have an ongoing incident and get the stateId of the incident
+   */
+  private List<UUID> getIncidentIds(TestCase test) {
```
harshach, line 513: this is going to be an expensive call

harshach (dismissed):

OnkarVO7 (approved):

Issue #14539, opened 2024-01-02
Use consistent font colours in service connection form 


ulixius9:
**Affected module**
UI

**Describe the bug**

The input username and hostport seems like placeholders rather than the actual values, whereas the password fields has a different font color
![image](https://github.com/open-metadata/OpenMetadata/assets/39544459/6852dcb5-1884-442d-8690-6ff4e1b1d076)

harsh-vador:
I will take this up

Abhishek332:
@ulixius9 can you assign it to me.

karanh37:
Fixed with #14551

PR #14551, opened 2024-01-03
fix(#14539): make the input font-color consistent

Abhishek332:
**### Before fix**
Screenshot 2024-01-03 at 11 23 40 AM

**### After fix**
Screenshot 2024-01-03 at 4 25 01 PM


openmetadata-ui/src/main/resources/ui/src/styles/app.less
```diff
@@ -769,3 +769,8 @@ a[href].link-text-grey,
   width: 20px;
   height: 20px;
 }
+
+.ant-input,
+.ant-input-password-icon.anticon {
+  color: #757575;
+}
```
karanh37, line 776: i don't think we should be overriding the entire ant-input.
Can you check the documentation of Antd Input. There should be a variable based on which color is passed to the input. We can override that variable in variables.less file.
Also directly providing color code 757575 is not a good idea, we should be using variables.

karanh37 (approved):

harshach:
image
we should provide a link to here https://docs.open-metadata.org/v1.2.x/deployment/ingestion/openmetadata#troubleshooting

chirag-madlani:
@Abhishek332 can you add a link for Airflow troubleshooting to your next PR? Thanks

PR #14641, opened 2024-01-09
#14630 added oracle stored procedures

### LINKED ISSUE #14630: Support lineage ingestion for Oracle Stored Procedures
We're using stored procedures heavily on our Oracle Data Warehouse for our ETL processes. Current Open Metadata version only support ingesting lineage from Oracle's View, so our logic/lineage in the Stored Procedure are not ingested and displayed. Therefore, if Open Metadata can support this feature, it will be really helpful to us (and other user working in Finance, Banking, Insurance industry, where Oracle Stored Procedure are used mainly).

Open Metadata to ingest Stored Procedures on Oracle database and show a lineage between tables (or even column level) based on the Stored Procedure logic.

NiharDoshi99:
### Describe your changes:

Fixes #14630 

I worked on including oracle stored procedure

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation


ingestion/src/metadata/ingestion/source/database/oracle/metadata.py
```diff
@@ -131,3 +158,107 @@ def get_view_definition(
...
+        """
+        Returns owner's entity reference
+        """
+        owner = None
+        user_owner_fqn = fqn.build(
+            self.metadata, entity_type=User, user_name=owner_name
+        )
+        if user_owner_fqn:
+            owner = self.metadata.get_entity_reference(entity=User, fqn=user_owner_fqn)
+        else:
```
ayush-shah, line 202: as we are returning an owner at the end and as it is under if else block, to avoid two conditions, let's return the owner here itself.
same goes for team owner fqn, if we don't find any fqn, we show warning

ayush-shah, line 202: as we are returning an owner at the end and as it is under if else block, to avoid two conditions, let's return the owner here itself.
same goes for team owner fqn, if we don't find any fqn, we show warning

OnkarVO7: Use the method [here](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/dashboard/tableau/metadata.py#L124) to get the owner entity reference

OnkarVO7 (changes requested):

ingestion/src/metadata/ingestion/source/database/oracle/metadata.py
NiharDoshi99: Ok 👍

OnkarVO7 (approved):

ayush-shah (approved):

PR #14733, opened 2024-01-16
Fixes #13556: Support for Salesforce table description ingestion

### LINKED ISSUE #13556: Support for Salesforce table description ingestion
For Salesforce, which is a Database service, there is no function to extract the description assigned to an object (table).
Even if the object (table) description is written in Salesforce, it must be manually added to OpenMetadata.Even if the description is included in Salesforce, it must be manually added to OpenMetadata.

We would like to extract object (table) descriptions in the Salesforce Metadata ingestion and automatically import them into OpenMetadata. This feature eliminates the need for double management of table descriptions in Salesforce and OpenMetadata. In addition, the automatic inclusion of decsription will improve the rate at which descriptions are assigned to tables and increase the organization's understanding of the data.

**Additional context**
In Salesforce, it is possible to optionally assign a description to an object, as shown in the attached image.

salesforce_customobject_description_20231013

kwgdaig:
### Describe your changes:

Fixes 13556

Fixes [#13556](https://github.com/open-metadata/OpenMetadata/issues/13556) Support for Salesforce table description ingestion

I have confirmed that we can ingest descriptions of salesforce object.
If the query to get the description fails due to insufficient privileges or other reasons, the process will continue with a warning log output.
image

#
### Type of change:

- [x] Improvement

#
### Checklist:


ingestion/src/metadata/ingestion/source/database/salesforce/metadata.py
```diff
@@ -191,6 +193,26 @@ def get_tables_name_and_type(self) -> Optional[Iterable[Tuple[str, str]]]:
...
             )
 
+    def set_table_desc_map(self) -> None:
+        """
+        Method to set table description obtained gotten Tooling API to map
+        """
+        try:
+            results = self.client.toolingexecute(
+                "query/?q=SELECT+QualifiedApiName,Description+FROM+EntityDefinition"
+            )["records"]
```
TeddyCr, line 203: is `records` key expected to here in the response?

ingestion/src/metadata/ingestion/source/database/salesforce/metadata.py
```diff
@@ -79,6 +79,8 @@ def __init__(self, config, metadata: OpenMetadata):
         self.client = get_connection(self.service_connection)
         self.table_constraints = None
         self.database_source_state = set()
+        self.table_desc_map = {}
```
TeddyCr, line 82: I think we don't need to store the data as part of the object. From what I am seeing it is  used in `yield_table`. You can probably just return the data from `get_table_description` move the logic of `set_table_description` in `get_table_description` and get rid of `set_table_description`.

TeddyCr (changes requested):
Thank you for the PR @kwgdaig I left a few comment regarding changes. 😊

ingestion/src/metadata/ingestion/source/database/salesforce/metadata.py
kwgdaig: @TeddyCr Yes, if the query is successful, we will get a response like [this](https://github.com/open-metadata/OpenMetadata/issues/13556#issuecomment-1769791063).

ingestion/src/metadata/ingestion/source/database/salesforce/metadata.py
kwgdaig: @TeddyCr 
We get table descriptions for all tables in salesforce at once, not for each table. If we moved the logic of `set_table_description` into `get_table_description`, I think that it would be a redundant process as `self.client.toolingexecute(...)` would be executed as many times as the number of tables to get the descriptions of all the tables each time. Is there a better method to avoid executing `self.client.toolingexecute(...)` many times?

kwgdaig:
> Thank you for the PR @kwgdaig I left a few comment regarding changes. 😊
@TeddyCr  Thank you for checking the code!  I have replied to your comment.

ingestion/src/metadata/ingestion/source/database/salesforce/metadata.py
kwgdaig: @TeddyCr 
As discussed in Slack, we have modified it to get the description for each table.
The reason for the fix is as follows.
- We want to split the results to consider the possibility that there is a limit on the number of results to be retrieved by the API.
- It seems that using the LIMIT and OFFSET clauses would not return correct results due to [SOQL Operation Limitations](https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/reference_objects_soql_limits.htm).

ingestion/src/metadata/ingestion/source/database/salesforce/metadata.py
kwgdaig: I have handled the case of no key by using except block.

TeddyCr (approved):

hareesh-anjane:
Hi Team, 

Is this fix live already? When we can expect this fix to be released?

 One of my client is looking to bring the description of an object in Salesforce into OpenMetadata but they are seeing the value of Salesforce "Field Label" instead of "Description"

kwgdaig:
Hi @hareesh-anjane,
I am sorry for the delay in getting back to you.
I have added in this PR only the feature of extracting the description of a Salesforce object (table).
For the field (column) description, the specification was to extract the value of "Field Label" in Salesforce as the description in OpenMetadata from before.
If you want to extract the value of Salesforce "Field Label",  I think you probably need to create a separate issue to discuss what is appropriate to extract as field (column) description for Salesforce object.

PR #14736, opened 2024-01-16
MINOR - Update developers documentation on loading sample data

IceS2:
### Describe your changes:

Small update on the developers documentation to fix outdated information and refactored the sections to make following it a bit easier:

- Updated the outdated information on loading sample data
- Changed the order of the sections to avoid having to jump down and up the page a couple times

#
### Type of change:

- [x] Documentation

#
### Checklist:


IceS2:
There's one extra thing that I was thinking about and it is about swapping the order of the documentation sections.

If whe check the following image we can see that within `Prerequisites` we are:

- Bootstrapping MySQL with tables: This requires the package to be built.
- Loading sample data into MySQL: This requires the package to be built and the service running.

Due to this, we are linking to the bottom of the page within this section but the flow feels a bit weird (Specially since it's close together and my screen doesn't move so I wasn't sure where to look)

I was thinking that we could have the sections:

1. Prerequisites: docker-compose commands
2. Building
3. Create a distribution
4. Bootstrap MySQL
5. Run Instance (I think we could also add here how to run the service without using IntelliJ)
6. Load Data

Thoughts?

image

pmbrull (dismissed):

PR #14739, opened 2024-01-16
#9269 - Add MSSQL Stored Procedure Support

### LINKED ISSUE #9269: Add ingestion of MsSQL procedures to OpenMetaData
We use procedures on MsSQL servers in our ETL processes. If the procedure was changed so our ETL can fail and we won't know the reason. So it will be really helpfull if OM could send alert if procedures code was changed. 

OM can ingest not only tables metadata but also metadata of procedures on MsSQL servers. Also procedures may be ingested as pipelines.

pmbrull:
### Describe your changes:

Fixes #9269

Also:
- Added `supportsQueryComment` to MSSQL. Tested it with our instance and it's all good there
- Cleaned up some unnecessary fields from the Stored Procedures code for snowflake, BQ, redshift and oracle

left a comment in the issue with the SP lineage query we are using. It's not great since I did not find a way to link SP <> Query via session ID or anything similar, so we might end up having some false positives.

Let's see if someone in the community is more versed in MSSQL than I am and can chip in

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


bootstrap/sql/migrations/native/1.3.0/mysql/schemaChanges.sql
```diff
@@ -159,3 +159,8 @@ CREATE TABLE IF NOT EXISTS consumers_dlq (
     timestamp BIGINT UNSIGNED GENERATED ALWAYS AS (json ->> '$.timestamp') NOT NULL,
     UNIQUE(id, extension)
 );
+
+-- Add supportsQueryComment to MSSQL
+update dbservice_entity
+set json = JSON_SET(json, '$.connection.config.supportsQueryComment', true)
+where serviceType = 'Mssql';
```
ayush-shah, line 166: @pmbrull  as supportsQueryComment by default is set to true, if it doesn't find it, it will be set to true, do we need a migration in these cases?

pmbrull: Without it the server did not start. I don't think the problem is the value but the server expecting the `key` there I guess

PR #14791, opened 2024-01-19
Fix #14764: Profiler make compute metric configurable

### LINKED ISSUE #14764: Profiler: Support only sample data ingestion.
**Affected module**
Ingestion Framework

**Describe the bug**
Currently the profiler workflow tries to compute the metric in all the scenarios even if the sample size is 0. We should introduce a flag or feature to skip the metric calculation and only ingest sample data.

ulixius9:
### Describe your changes:

Fix #14764: Profiler make compute metric configurable

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/profiler/processor/core.py
+++ b/ingestion/src/metadata/profiler/processor/core.py
@@ -92,6 +92,7 @@ class Profiler(Generic[TMetric]):
         """
 
         self.profiler_interface = profiler_interface
+        self.source_config = self.profiler_interface.source_config
         self.include_columns = include_columns
         self.exclude_columns = exclude_columns
         self._metrics = metrics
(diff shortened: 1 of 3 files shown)


ayush-shah (approved):

TeddyCr (commented):
@ulixius9 what do you think of passing the `computeMetric` to the `get_profiler_runner` from `ingestion/src/metadata/profiler/processor/processor.py` either passing `self.source_config` as an argument or `self.source_config.computeMetric`. 

This way in `ingest/src/metadata/profiler/source/base/profiler_source.py` in `get_profiler_runner()` you can pass the computerMetric to the profiler runner instance and init it with computeMetric true/false. You'll need to add the necessary attributes to `DefaultProfiler` and `Profiler`.

This way in `Process` we don't modify the signature and we can just reference `self.should_compute_metrics` -- we should probably do something similar with `generate_sample_data` -- in which case it might just make sense to pass the `self.source_config`.

TeddyCr (approved):

PR #14792, opened 2024-01-19
MINOR - Better handling of Ingestion Pipeline Status

pmbrull:
### Describe your changes:

Setting the wrong service name in the `ingestionPipelineFQN` values in the YAML hangeg the workflow, since it failed before stopping the timer.

Adding a couple of things here:
1. Since we are creating the Ingestion Pipeline already, using its FQN. This means that even if users set the wrong serviceName in the `ingestionPipelineFQN`, it will still work. I am not changing this to only pipeline name or anything since it would require to patch the pipeline service client logic.
2. Adding exception handling so that if something explodes, the workflow gets stopped properly

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/workflow/application.py
+++ b/ingestion/src/metadata/workflow/application.py
@@ -145,7 +145,7 @@ class ApplicationWorkflow(BaseWorkflow, ABC):
             and self.calculate_success() < SUCCESS_THRESHOLD_VALUE
         ):
             raise WorkflowExecutionError(
-                f"{self.runner.name} reported errors: {Summary.from_step(self.source)}"
+                f"{self.runner.name} reported errors: {Summary.from_step(self.runner)}"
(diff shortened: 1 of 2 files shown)


IceS2 (approved):
LGTM.

Since I don't have enough context yet I'm unsure but I guess the `Source` import was due to some copy paste since this Workflow seems a bit special and without the need of a source at all, right?

Thanks (=

pmbrull:
> LGTM.
> 
> Since I don't have enough context yet I'm unsure but I guess the `Source` import was due to some copy paste since this Workflow seems a bit special and without the need of a source at all, right?
> 
> Thanks (=

it was indeed some old copy paste :)

PR #14825, opened 2024-01-23
Fixes  #5448: Implement initial Iceberg Connector using PyIceberg

### LINKED ISSUE #5448: Create Generic Iceberg Source Connector
### Context
Ingestion of iceberg table is currently supported through the Glue source. As Iceberg tables requires the use of a metastore giving the ability to the user to define in the connector configuration which metastore they are using will allow us to support all implementation of Iceberg tables

### Solution
Leveraging the [Python API](https://py.iceberg.apache.org/) for Iceberg we would be able to leverage existing logic to fetch table metadata based.

### A.C.
- Iceberg source is available as a connector in OM

IceS2:
### Describe your changes:

Fixes #5448 

image

image

image

Initial implementation on the Iceberg Connector, leveraging PyIceberg to connect to supported catalogs and extract Iceberg table Metadata.

**Forcing some requirements to make it work with the rest of the dependencies. It's important that someone double checks it :eyes:**

**Notes**
- Uses PyIceberg 0.4.0 due to being the last version that uses Pydantic v1

**Covers**
- Overall Support for **Hive**, **REST**, **Glue** and **DynamoDB** as Catalogs
- Overall Support for **Local**, **AWS S3** and **Azure Blob Storage** as FileStorage

**Does not (yet) Support**
- SQL Catalog (Not supported by PyIceberg 0.4.0
- **Account Key**, **Connection String**, **SaS Token** on Azure Blob Storage
- **Signer**, **Proxy Uri**, **profileName**, **assumeRoleArn**, **assumeRoleSessionName**, **assumeRoleSourceIdentity** on AWS S3
- **Google Storage**
- **HDFS**
- **Partition Transformations** are not tracked (But source columns are)

#
### Type of change:

- [x] New feature

#
### Checklist:

- [x] The issue properly describes why the new feature is needed, what's the goal, and how we are building it. Any discussion


ingestion/setup.py
```diff
@@ -20,16 +20,19 @@
 # Add here versions required for multiple plugins
 VERSIONS = {
     "airflow": "apache-airflow==2.7.3",
+    "adlfs": "adlfs==2022.11.2",  # Python 3.7 does only support up to 2022.2.0
```
pmbrull, line 23: we don't support python 3.7 anymore so I guess we can just remove the comment

ingestion/setup.py
```diff
@@ -20,16 +20,19 @@
...
     "geoalchemy2": "GeoAlchemy2~=0.12",
     "google-cloud-storage": "google-cloud-storage==1.43.0",
+    "gcsfs": "gcsfs==2022.11.0",
     "great-expectations": "great-expectations~=0.18.0",
     "grpc-tools": "grpcio-tools>=1.47.2",
     "msal": "msal~=1.2",
     "neo4j": "neo4j~=5.3.0",
     "pandas": "pandas<=2,<3",
     "pyarrow": "pyarrow~=14.0",
+    "pydantic": "pydantic==1.10.13",
```
pmbrull, line 35: should we keep the `~=`? in case any other minor patches are released there

ingestion/setup.py
pmbrull: also might be better to keep things a bit more generic with `~=` in case people install this on their instances

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/icebergConnection.json
```diff
@@ -0,0 +1,199 @@
...
+    },
+    "restCatalogConnection": {
+      "title": "REST",
+      "description": "Configuration for using a REST catalog.",
+      "type": "object",
+      "properties": {
+        "uri": {
+          "title": "Uri",
+          "description": "Uri to the REST catalog. Example: 'http://rest-catalog/ws/'",
+          "type": "string"
```
pmbrull, line 41: should it be type `uri`?

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/icebergConnection.json
```diff
@@ -0,0 +1,199 @@
...
+        },
+        "credential": {
+          "title": "OAuth2 Credential",
+          "description": "OAuth2 credential to use when initializing the catalog. Example: 't-1234:secret', where 't-1234' is the Client Id and 'secret' the Client Secret.",
+          "type": "string"
+        },
+        "token": {
+          "title": "Token",
+          "description": "Berarer token to use for the 'Authorization' header.",
+          "type": "string"
```
pmbrull, line 51: type should be `password` for any piece of the config that we want to mask

pmbrull: + all the definitions should have additional properties as false just in case

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/icebergConnection.json
```diff
@@ -0,0 +1,199 @@
...
+            },
+            {
+              "title": "Azure",
+              "$ref": "../../../../security/credentials/azureCredentials.json"
+            }
+          ]
+        },
+        "warehouseLocation": {
+          "title": "Warehouse Location",
+          "description": "Warehouse Location. Used to specify a custom warehouse location.",
```
pmbrull, line 174: maybe put an example for each value

ingestion/tests/unit/topology/database/test_iceberg.py
```diff
@@ -0,0 +1,697 @@
...
+
+        # Happy Path scenario
+        class LoadTableMock:
+            data = iter(MOCK_TABLE_LIST)
+
+            def name(self):
+                return next(self.data)
+
+
+        with \
```
pmbrull, line 549: I think the formatter might cry due to this `\`

ingestion/src/metadata/ingestion/source/database/iceberg/catalog/__init__.py
TeddyCr, line 1: What do you think of making `IcebergCatalogFactory` either an instantiable class or just breaking it down with `catalog_type_map` being a global and `from_connection` being a simple function? I understand the intent but we don't really need that object if we won't use it as an object.

Do you see any advantage in that context of defining this piece of code directly in `_init__.py`?

IceS2: Hey, to be honest I don't have a really strong opinion... Been messing around with both ways actually xD

I believe that trying to wrap things with specific classes tends to make the code a bit more readable because it gives you context straight away, but if we avoid doing that in this project I wouldn't mind having just a `from_connection` function.

About defining it on __init__.py, for me it made sense as the `IcebergCatalogFactory` is basically the entrypoint for the sub module.

Again if it's against our common practices I wouldn't mind defining it on another file and to make the import statement more directy we can "re-export" it on the __init__.py.

pmbrull (approved):
Thank you, @IceS2, for the effort here. Not an easy first connector to tackle.

Congratulations on your first contribution

PR #14852, opened 2024-01-24
Fix 13970: Fix Consecutive Array issue in Athena

### LINKED ISSUE #13970: Metadata ingestion failed for nested arrays column in Athena
**Affected module**
Ingestion Framework

**Describe the bug**
When ingesting Athena table with column having more than one level of consecutive array, it will fail and unable to ingest the metadata to OMD.
image (3)

image (2)

Note: It is fine if there is a struct between two arrays
image

**Expected behavior**
Unnest the array normally in the ingestion framework and UI.

**Version:**
 - OpenMetadata version: 1.2.0

ayush-shah:
### Describe your changes:

Fixes #13970

The issue described was being caused because Array(Array) failed as the it needs item_type as argument and it was sent empty. As we don't cater to nested arrays for their children and their types, dataType will be Array, arrayDataType will be Array, dataTypeDisplay would be array<array<.....>>

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/athena/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/athena/metadata.py
@@ -93,20 +93,14 @@ def _get_column_type(self, type_):
         "struct": sqa_types.SQAStruct,
         "row": sqa_types.SQAStruct,
         "map": sqa_types.SQAMap,
+        "decimal": types.DECIMAL,
+        "varchar": types.VARCHAR,
+        "char": types.CHAR,
     }
-    if name in ["decimal"]:
-        col_type = types.DECIMAL
+    if name in ["decimal", "char", "varchar"]:
+        col_type = col_map[name]
         if length:
-            precision, scale = length.split(",")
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/ingestion/source/database/athena/metadata.py
```diff
@@ -93,28 +93,28 @@ def _get_column_type(self, type_):
         "struct": sqa_types.SQAStruct,
         "row": sqa_types.SQAStruct,
         "map": sqa_types.SQAMap,
+        "decimal": types.DECIMAL,
+        "varchar": types.VARCHAR,
+        "char": types.CHAR,
     }
-    if name in ["decimal"]:
-        col_type = types.DECIMAL
+    if name in ["decimal", "char", "varchar"]:
```
TeddyCr, line 100: No need to change it, but you can use a `set` here for a more efficient lookup.

PR #14914, opened 2024-01-29
by default make isBot : false in team index

07Himank:
I worked on ... because ... by default make isBot : false in team index

#
### Type of change:

- [x] Bug fix
- [x] Improvement


harshach (approved):

chirag-madlani:
@07Himank @harshach Do we need migration here? or we are doing `re-create` all at the time of the upgrade itself?

07Himank:
> @07Himank @harshach Do we need migration here? or we are doing `re-create` all at the time of the upgrade itself?

Recreate should work fine

PR #14938, opened 2024-01-30
remove wildcards from search

karanh37:
### Describe your changes:

Explore issues

1. Remove wildcards from global search dropdown and search performed on enter.
2. Using counts from the respective index calls. Previously we are making two calls, one for fetching count and one for actual data based on the selected index. The count received was used for left menu items and for pagination of content. We have replaced that, we will rely on respective index calls for the count. So we will be using `hits.total.value` for pagination.
3. Increased height of code block for Store Procedure

I worked on ... because ...

https://github.com/open-metadata/OpenMetadata/assets/33024356/eb22868e-b5cc-4f75-8704-cf0b66529850

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/rest/miscAPI.ts
```diff
@@ -36,7 +36,8 @@ export const searchData = <SI extends SearchIndex>(
   sortOrder: string,
   searchIndex: SI | SI[],
   onlyDeleted = false,
-  trackTotalHits = false
+  trackTotalHits = false,
+  wildcard = true
```
chirag-madlani, line 40: since we are removing wildcard, do we need this new param?

chirag-madlani (approved):

PR #14950, opened 2024-01-30
Added Index Factory Config

07Himank:
I worked on ... because ... Added Index Factory Config

https://github.com/open-metadata/OpenMetadata/assets/112613760/56a43fb7-6275-441a-9a52-ff8f0ad25507

#
### Type of change:

- [x] Bug fix
- [x] Improvement

### DIFF
--- a/conf/openmetadata.yaml
+++ b/conf/openmetadata.yaml
@@ -256,7 +256,8 @@ elasticsearch:
   socketTimeoutSecs: ${ELASTICSEARCH_SOCKET_TIMEOUT_SECS:-60}
   keepAliveTimeoutSecs: ${ELASTICSEARCH_KEEP_ALIVE_TIMEOUT_SECS:-600}
   batchSize: ${ELASTICSEARCH_BATCH_SIZE:-10}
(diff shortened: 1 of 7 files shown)


conf/openmetadata.yaml
```diff
@@ -256,7 +256,8 @@ elasticsearch:
   socketTimeoutSecs: ${ELASTICSEARCH_SOCKET_TIMEOUT_SECS:-60}
   keepAliveTimeoutSecs: ${ELASTICSEARCH_KEEP_ALIVE_TIMEOUT_SECS:-600}
   batchSize: ${ELASTICSEARCH_BATCH_SIZE:-10}
-  searchIndexMappingLanguage: ${ELASTICSEARCH_INDEX_MAPPING_LANG:-EN}
+  searchIndexMappingLanguage : ${ELASTICSEARCH_INDEX_MAPPING_LANG:-EN}
+  searchIndexFactoryClassName : org.openmetadata.service.search.SearchIndexFactory
```
harshach, line 260: any reason we need this as config?

07Himank: @harshach we have two index factory one is SearchIndexFactory and one is CollateIndexFactory so we are making that object from classname.

PR #14994, opened 2024-02-01
MINOR: Modified snowflake view definition query

OnkarVO7:
### Describe your changes:

`SHOW VIEWS` does not incorporate the changes of a view in case it has been modified using `ALTER VIEW` statement.
We'll use the [GET_DDL](https://docs.snowflake.com/en/sql-reference/functions/get_ddl) method to get the correct view definitions.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/snowflake/utils.py
+++ b/ingestion/src/metadata/ingestion/source/database/snowflake/utils.py
@@ -121,13 +121,10 @@ def get_view_definition(  # pylint: disable=unused-argument
     schema = schema or self.default_schema_name
     if schema:
         cursor = connection.execute(
-            "SHOW /* sqlalchemy:get_view_definition */ VIEWS "
-            f"LIKE '{view_name}' IN {schema}"
+            f"SELECT GET_DDL('VIEW','{schema}.{view_name}') AS \"text\""
         )
     else:
-        cursor = connection.execute(
-            "SHOW /* sqlalchemy:get_view_definition */ VIEWS " f"LIKE '{view_name}'"
-        )
+        cursor = connection.execute(f"SELECT GET_DDL('VIEW','{view_name}') AS \"text\"")
     n2i = self.__class__._map_name_to_idx(cursor)  # pylint: disable=protected-access
     try:
         ret = cursor.fetchone()


ingestion/src/metadata/ingestion/source/database/snowflake/utils.py
```diff
@@ -121,13 +121,10 @@ def get_view_definition(  # pylint: disable=unused-argument
     schema = schema or self.default_schema_name
     if schema:
         cursor = connection.execute(
-            "SHOW /* sqlalchemy:get_view_definition */ VIEWS "
-            f"LIKE '{view_name}' IN {schema}"
+            f"SELECT GET_DDL('VIEW','{schema}.{view_name}') AS \"text\""
```
ayush-shah, line 124: should we have something like this

/* {{"app": "OpenMetadata", %%}} */

OnkarVO7: Added it

ingestion/src/metadata/ingestion/source/database/snowflake/utils.py
```diff
@@ -121,12 +121,11 @@ def get_view_definition(  # pylint: disable=unused-argument
...
     if schema:
         cursor = connection.execute(
-            "SHOW /* sqlalchemy:get_view_definition */ VIEWS "
-            f"LIKE '{view_name}' IN {schema}"
+            f'SELECT /* "app": "OpenMetadata" */ GET_DDL(\'VIEW\',\'{schema}.{view_name}\') AS "text"'
         )
     else:
         cursor = connection.execute(
-            "SHOW /* sqlalchemy:get_view_definition */ VIEWS " f"LIKE '{view_name}'"
+            f'SELECT /* "app": "OpenMetadata" */ GET_DDL(\'VIEW\',\'{view_name}\') AS "text"'
```
pmbrull, line 128: aren't we including the comments directly on the engine?

pmbrull: I think that if we use the connection engine, it'll be handled by this https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/connections/headers.py#L35 so no need to add it explicitely. Note how we don't add it anywhere

OnkarVO7: Ah makes sense thanks for reviewing it. Let me revert it

ayush-shah (approved):

PR #14996, opened 2024-02-01
adding unit tests

Abhishek332:
this PR contains unit test for -

- [x] components/Alerts/AlertsDetails
- [x] components/AppTour
- [x] components/Modals/TourEndModal
- [x] components/Applications/AppDetails
- [x] components/Applications/AppInstallVerifyCard
- [x] components/Applications/AppLogsViewer
- [x] components/Applications/AppRunsHistory
- [x] components/Applications/AppSchedule
- [x] components/Applications/MarketPlaceAppDetails

Screenshot 2024-02-01 at 3 04 59 PM
Screenshot 2024-02-01 at 5 35 27 PM
Screenshot 2024-02-01 at 6 04 12 PM  
Screenshot 2024-02-02 at 4 48 02 PM
Screenshot 2024-02-02 at 9 10 41 PM
Screenshot 2024-02-02 at 9 09 46 PM
Screenshot 2024-02-05 at 3 10 40 PM
Screenshot 2024-02-05 at 6 19 52 PM
Screenshot 2024-02-06 at 9 07 54 AM


openmetadata-ui/src/main/resources/ui/src/components/Alerts/AlertsDetails/AlertDetails.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+    render(
+      <Router history={history}>
+        <AlertDetailsComponent {...mockProps1} />
+      </Router>
+    );
+
+    const editButton = screen.getByRole('button', { name: 'label.edit' });
+    const deleteButton = screen.getByRole('button', { name: 'label.delete' });
+
+    userEvent.click(deleteButton);
```
Sachin-chaurasiya, line 131: Can we wrap this within `act`? please take care of this in all the places.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/AlertsDetails/AlertDetails.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+    expect(deleteButton).not.toBeInTheDocument();
+  });
+
+  it('should call action handlers on click of buttons', () => {
+    const history = createMemoryHistory();
+
+    history.push = mockPush;
+
+    render(
+      <Router history={history}>
```
Sachin-chaurasiya, line 123: Do we need this? I think we can use memoryRouter here.

Sachin-chaurasiya (commented):
Left a few comments, thanks

openmetadata-ui/src/main/resources/ui/src/components/Alerts/AlertsDetails/AlertDetails.test.tsx
Abhishek332: can you eleborate more why we want to wrap the userEvent.click in act at every place. It should be wrapped in act only if click event is doing state update.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/AlertsDetails/AlertDetails.test.tsx
Abhishek332: We want to check here Link is redirecting onclick or not, MemoryRouter not accept history prop.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/AlertsDetails/AlertDetails.test.tsx
Sachin-chaurasiya: Yes, we should wrap the event only if we are updating the state.

By this, I meant we should check all the places and if we are updating the state on some event then it should be wrapped in `act`.
> Please take care of this in all the places.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/AlertsDetails/AlertDetails.test.tsx
Sachin-chaurasiya: Then in that case can we mock the `useHistory` hook push method?

Abhishek332: We are not using userHistory hook in this component, I am testing there <Link> redirect path that will be create based on props passed in AlertDetails component.

To verify it, path creating currectly, clicking on link and checking if push method of Router getting called with the value of "to" prop in Link.

Sachin-chaurasiya: Okay, got it, thanks for the explanation @Abhishek332

Sachin-chaurasiya (approved):

PR #15012, opened 2024-02-02
add what's new content for 1.3.0

Sachin-chaurasiya:
https://github.com/open-metadata/OpenMetadata/assets/59080942/bdfba034-afd6-475a-ae86-d2a68bcf82cd


harshach:
@Sachin-chaurasiya @chirag-madlani we need to add videos right why are we doing this way?

Sachin-chaurasiya:
> @Sachin-chaurasiya @chirag-madlani we need to add videos right why are we doing this way?

Hello @harshach, currently only contents are available. Once videos become available, we will add them and merge this PR. cc: @ShilpaVernekar

chirag-madlani (approved):

PR #15013, opened 2024-02-02
FIXES 15215: Add execution time tracker for ingestion

IceS2:
### Describe your changes:

Start implementation to keep track of the Execution time of any code section we want to keep. The main idea is to be able to understand the accumulated taken on different parts such as when extracting data from a source system or actually updating the OpenMetadata server.

I Implemented the `ExecutionTimeTracker` as a Singleton in order to be able to keep the state globally.
In order to avoid some overhead when uneeded we can use the `enabled` attribute that is currently tied to the LogLevel.

It also has the possibility to just log instead of keeping state of a certain part of the code by using the `store` attribute.

We provide the two `calculate_execution_time` and `calculate_execution_time_generator` decorators, substituting the ones we had defined previously while maintaining the same behaviour where they were used.

By the end we can provide a summary such as:

```
[2024-02-02 18:27:19] DEBUG    {metadata.Ingestion:base:194} - Time Consumed
----------------------------------------
GET: 0.04s

Sink: 0.8s
----------------------------------------
Sink.GET: 0.12s

Source: 1.0m 44.98s
----------------------------------------
Source.GET: 45.72s
```

#
### Type of change:

- [x] Improvement

#
### Checklist:


ingestion/src/metadata/workflow/base.py
```diff
@@ -92,6 +93,7 @@ def __init__(
         self._timer: Optional[RepeatedTimer] = None
         self._ingestion_pipeline: Optional[IngestionPipeline] = None
         self._start_ts = datetime_to_ts(datetime.now())
+        self._execution_time_tracker = ExecutionTimeTracker(log_level.value == "DEBUG")
```
pmbrull, line 96: We can use LogLevels.DEBUG

ingestion/src/metadata/workflow/base.py
```diff
@@ -189,6 +191,7 @@ def execute(self) -> None:
 
         # Force resource closing. Required for killing the threading
         finally:
+            logger.debug("%s", self._execution_time_tracker.print_summary())
```
pmbrull, line 194: Maybe we can add the print here https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/workflow/output_handler.py#L107

ingestion/src/metadata/utils/execution_time_tracker.py
```diff
@@ -0,0 +1,216 @@
...
+            )
+
+            if context.stored:
+                self._save(context, elapsed)
+
+    def _save(self, context: ExecutionTimeTrackerContext, elapsed: float):
+        """Small utility to save the new measure to the global accumulator."""
+        self.state[context.name] = self.state.get(context.name, 0) + elapsed
+
+    def print_summary(self):
```
pmbrull, line 135: to move here https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/workflow/output_handler.py#L107

pmbrull (commented):
LGTM, a couple of structure changes

pmbrull (approved):

PR #15130, opened 2024-02-09
FIXES 15215: Implement initial Multithreading approach for the Metadata Ingestion on Databases

IceS2:
### Describe your changes:

Initial implementation in order to be able to improve performance for the Metadata Ingestion by using Multithread.
It is implemented on the TopologyRunner so it could be used by any Source but the initial support is only for the DatabaseSource at either the Table or Schema node (Depending on performance tests).

The Ingestion Workflows work as a chain of generators and when trying to implement Multiprocessing, as soon as the first yield in the thread was reached the thread was marked as done and the whole generator was returned. Not helpful.

To circunvent this, I created a Queue system so that instead of yielding from the threads, I add the result of a given stage to the queue and wait for the main thread to process it.
This allow us to concurrently send requests to the Source system while maintaining the needed sequencial processing, making more use of the idle I/O time (just on the Source side).

### Flow Comparisson
---

This difference on the flow can be seen comparing the following images:

Old Flow
image

New Flow
image

**Please note that the Old flow still exists for every Node that is not threaded.**

### Performance Tests
---

BigQuery

4 Schemas, 200 Tables Each:

Workflow finished in time: 58.0m 3.24s- 0 Threads - 1604 Records

Workflow finished in time: 32.0m 26.21s - 2 Threads - 1604 Records

Workflow finished in time: 14.0m 51.35s - 4 Threads - 1604 Records

Redshift

4 Schemas, 1000 Tables Each

Workflow finished in time: 2.0m 56.09s - 0 Threads - 3930 Records

Workflow finished in time: 2.0m 26.51s - 2 Threads - 3930 Records

Workflow finished in time: 2.0m 20.7s  - 4 Threads - 3930 Records

Snowflake

4 Schemas, 1000 Tables Each

Workflow finished in time: 4.0m 32.41s - 0 Threads (7815 Records)

Workflow finished in time: 3.0m 15.71s - 2 Threads (7815 Records)

Workflow finished in time: 3.0m 13.64s - 4 Threads (7815 Records)

### A note on the TopologyContext object
---
In order to be able to mantain the correct Context between multiple threads a new class was created, the TopologyContextManager that is responsible for actually retrieving the correct context based on the thread where the code is executed.

**Currently we don't have any mechanism to tell a given Stage/Node that they should save the Context on the Global context as well for future usage**

I did it by manually getting the `global context` on the only place I saw it was needed.

#
### Type of change:

- [x] Improvement

#
### Checklist:


pmbrull (dismissed):
Thank you @IceS2! Fantastic job, really excited to see this in! Nothing much to comment tbh. Maybe the only piece is that it'd be sad to lose the diagram you added in the README. Having that as part of the actual module README, or explained thoroughly in the method docstring would help keep us aligned in the future

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/databaseServiceMetadataPipeline.json
```diff
@@ -93,6 +93,12 @@
       "description": "Regex to only fetch databases that matches the pattern.",
       "$ref": "../type/filterPattern.json#/definitions/filterPattern",
       "title": "Database Filter Pattern"
+    },
+    "threads": {
+      "description": "Number of Threads to use in order to paralellize Table ingestion.",
```
sushi30, line 98: suggestion: document how this translates to actual load on the data source (even a loose formula like 1 thread ~ 1 connection)

pmbrull (approved):

ingestion/README.md
```diff
@@ -14,3 +14,17 @@ OpenMetadata Ingestion is a simple framework to build connectors and ingest meta
...
+
+### TopologyRunner
+
+All the Ingestion Workflows run through the TopologyRunner.
+
+The flow is depicted in the images below.
+
+**TopologyRunner Standard Flow**
+
+![image](https://github.com/IceS2/OpenMetadata/assets/4912399/c253af53-c11a-4b91-b101-892fa8169c81)
```
pmbrull, line 26: when working on the docs, we could add the images directly in the repo to make sure they stick

pmbrull (approved):

PR #15136, opened 2024-02-11
Support Min/Max Threshold for Custom SQL Tests #14983

### LINKED ISSUE #14983: Support Min/Max Threshold for Custom SQL Tests
Currently, custom SQL quality tests only supports row count <= threshold. It will be great if we can set max/min value instead of a single threshold.

Change Threshold to Max Value
Add another filed for Min Value

Without this feature, if I hope to set a test which should return at least 10 rows, I can not set the threshold to 10. Insted, I have to write a case when expression to return 0 if it meets the expectation and 1 if it doesn't. 

**Additional context**
N/A

Anixton:
### Describe your changes:

Fixes #14983

I worked on Support Min/Max Threshold for Custom SQL Tests because it's good first issue.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


pmbrull:
Thanks for the PR @Anixton! Let us take a look here and bring this in! Appreciate it! cc @TeddyCr

TeddyCr:
@Anixton thanks for the PR. The changes you made are not in the correct file.  You will need to modify things in:
- https://github.com/open-metadata/OpenMetadata/blob/37a7cf99b66fd39aaedbf520d6c5d82837936477/openmetadata-service/src/main/resources/json/data/tests/tableCustomSQLQuery.json
- https://github.com/open-metadata/OpenMetadata/blob/d0cffdcd66542834662e9714117cd5ec03294aeb/ingestion/src/metadata/data_quality/validations/table/base/tableCustomSQLQuery.py

Basically replacing the threshold with `minThreshold` and `maxThreshold` and implementing the logic so that we can enforce the behavior either when both are set or just one is set.

Anixton:
@TeddyCr 
Hey I did some changes. Can you check if my modifications are correct?

TeddyCr:
> @TeddyCr Hey I did some changes. Can you check if my modifications are correct?

thanks for the update. It is partially correct. You also need to add the logic on the python side of things to handle the result processing: 
- https://github.com/open-metadata/OpenMetadata/blob/d0cffdcd66542834662e9714117cd5ec03294aeb/ingestion/src/metadata/data_quality/validations/table/base/tableCustomSQLQuery.py

Anixton:
> > @TeddyCr Hey I did some changes. Can you check if my modifications are correct?
> 
> thanks for the update. It is partially correct. You also need to add the logic on the python side of things to handle the result processing:
> 
> * https://github.com/open-metadata/OpenMetadata/blob/d0cffdcd66542834662e9714117cd5ec03294aeb/ingestion/src/metadata/data_quality/validations/table/base/tableCustomSQLQuery.py

I added the logic on python side.

PR #15153, opened 2024-02-12
MINOR: Fix superset query to fetch columns

ulixius9:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/superset/db_source.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/superset/db_source.py
@@ -15,7 +15,6 @@ Superset source module
 import traceback
 from typing import Iterable, Optional
 
-from sqlalchemy import sql
 from sqlalchemy.engine import Engine
 from sqlalchemy.engine.url import make_url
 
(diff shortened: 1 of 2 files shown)


ingestion/src/metadata/ingestion/source/dashboard/superset/db_source.py
```diff
@@ -87,7 +86,7 @@ def get_column_list(self, table_name: str) -> Iterable[FetchChart]:
         try:
             if table_name:
                 col_list = self.engine.execute(
-                    sql.text(FETCH_COLUMN), table_name=table_name.lower()
```
TeddyCr, line 90: @ulixius9 have you tried with `text` using `from sqlalchemy import text`?

ulixius9: Hey @TeddyCr, I just tested it out and still it does not work, it throws the syntax error as it converts the `%` into `%%(table_name)s` to escape it and then the text binding does not work.

Hence I had to remove the text wrapping and pass the string as it is

pmbrull (approved):

PR #15192, opened 2024-02-15
#13985 - Azure KV Secrets Manager

### LINKED ISSUE #13985: Please Add Azure Key Vault as a supported Secrets Manager Implementation

We are running in Azure and would like to implement using a Secrets Manager, however Azure Key Vault is not a supported implementation.

The ability to use Azure Key Vault as a Secrets Manager similar to the AWS solutions.

We are using Azure Key Vault in the deployment pipeline to keep the secrets hidden until the actual deployment on the target system.

**Additional context**

Related conversation: [https://openmetadata.slack.com/archives/C02B6955S4S/p1697746755061959](https://openmetadata.slack.com/archives/C02B6955S4S/p1697746755061959)

pmbrull:
### Describe your changes:

Fixes #13985

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


akash-jain-10 (dismissed):

ingestion/src/metadata/utils/secrets/azure_kv_secrets_manager.py
```diff
@@ -0,0 +1,72 @@
...
+class AzureKVSecretsManager(ExternalSecretsManager, ABC):
+    """
+    Azure Key Vault Secrets Manager class
+    """
+
+    def __init__(
+        self,
+        credentials: Optional["AzureCredentials"],
+    ):
+        super().__init__(SecretsManagerProvider.azure_kv)
```
TeddyCr, line 42: I think we can just do 

```
self.provider = SecretsManagerProvider.azure_kv
```

To make it more explicit than call the parent `__init__` since we don't do anything else in the parent `__init__`. WDYT?

ingestion/src/metadata/utils/secrets/client/loader.py
```diff
@@ -37,7 +40,9 @@ def _(_: SecretsManagerProvider) -> None:
 
 
 @secrets_manager_client_loader.add(SecretsManagerClientLoader.airflow.value)
-def _(provider: SecretsManagerProvider) -> Optional[AWSCredentials]:
+def _(
+    provider: SecretsManagerProvider,
+) -> Optional[Union[AWSCredentials, AzureCredentials]]:
```
TeddyCr, line 45: nit, I think we can define the type as `ExternalSecretsManager`

openmetadata-service/src/main/java/org/openmetadata/service/secrets/SecretsManager.java
```diff
@@ -51,11 +51,16 @@ public abstract class SecretsManager {
...
 
+  protected record SecretsIdConfig(
+      String separator,
+      Boolean needsStartingSeparator,
+      String cleanSecretReplacer,
+      Pattern secretIdPattern) {}
+
   @Getter private final SecretsConfig secretsConfig;
   @Getter private final SecretsManagerProvider secretsManagerProvider;
+  @Getter private final SecretsIdConfig secretsIdConfig;
```
TeddyCr, line 62: Can this and the buld method be in the Azure class since it seems to be Azure related ?

TeddyCr (commented):
Left a few commenst let me know if it does not make sense.

openmetadata-service/src/main/java/org/openmetadata/service/secrets/SecretsManager.java
pmbrull: kept it in the general one as as default with the idea to give room for any other SM to override it. Right now it's only Azure KV, but I am guessing other solutions might have their own specific conventions. The first implementation was a bit too AWS-centric

ingestion/src/metadata/utils/secrets/azure_kv_secrets_manager.py
pmbrull: fair

ingestion/src/metadata/utils/secrets/client/loader.py
pmbrull: good point, thank you

ingestion/src/metadata/utils/secrets/azure_kv_secrets_manager.py
pmbrull: mh actually it might be easier to miss if we don't have it here and just have it as a line we need to remember to add. The IDE / lint would still point us that the class is not calling the super().__init__()

TeddyCr (approved):

PR #15280, opened 2024-02-21
fix issue 14802: disable self-sigup interface when using ldap authent…

### LINKED ISSUE #14802: SELF_SIGNUP setting doesn't work for LDAP authentication provider
**Affected module**
impact the UI

**Describe the bug**
The authenticate page shows the option "Create Account" when using ldap provider. 
For the setting AUTHENTICATION_PROVIDER , when using basic value, this works fine. But if it's ldap, it's doesn't work as expected

**To Reproduce**

1.  change the AUTHENTICATION_PROVIDER config to ldap
2. change the AUTHENTICATION_ENABLE_SELF_SIGNUP to false
3.  open the main page (like image attachment)

**Expected behavior**
When setting the AUTHENTICATION_ENABLE_SELF_SIGNUP to false, it's expect that the user can't sign in creating new user because the account must exist in LDAP provider.

**Version:**
 - OS: AlmaLinux 9.2 using docker image openmetadata/server:1.2.4
 - Python version:
 - OpenMetadata version: 1.2.4
 - OpenMetadata Ingestion package version: not used

image

thiagopirex:
### Describe your changes:

Fixes #14802 

I worked on this because my enviroment uses Active Directory to authenticate users. So when using ldap provider authentication in OpenMetadata, I don't want to enable self-signup for users. This change enables and fix the configuration  of self-signup and ldap provider, disabling the  "Forgot Password" and "Create Account" options on the login page interface.

In my environment, the "Forgot Password" and "Create Account" options are provided by another application.

image

#
### Type of change:

#
### Checklist:


harshach:
@thiagopirex Thanks for the PR. Did you try disabling self-signup via config https://github.com/open-metadata/OpenMetadata/blob/6c0f85cc98a965c65156dade396fd0a832105062/conf/openmetadata.yaml#L176
We don't want to introduce SSO based code to make decisions in the UI. This should be controlled via config above.

thiagopirex:
> @thiagopirex Thanks for the PR. Did you try disabling self-signup via config
> 
> https://github.com/open-metadata/OpenMetadata/blob/6c0f85cc98a965c65156dade396fd0a832105062/conf/openmetadata.yaml#L176
> 
> 
> We don't want to introduce SSO based code to make decisions in the UI. This should be controlled via config above.

Yes, I put "false" for the AUTHENTICATION_ENABLE_SELF_SIGNUP setting. But when using ldap provider authentication, the login page shows both otpions, as reported in issue #14802

openmetadata-ui/src/main/resources/ui/src/pages/LoginPage/SignInPage.tsx
```diff
@@ -214,7 +214,7 @@ const SignInPage = () => {
                     data-testid="email"
                     label={
                       isAuthProviderLDAP
-                        ? t('label.email')
+                        ? t('label.username')
```
chirag-madlani, line 217: Hi @ thiagopirex, I checked with @mohityadav766 who is working on the backend side of things and currently, OM doesn't support login via Username for LDAP so we need to revert this label change
```suggestion
                        ? t('label.email')
```

openmetadata-ui/src/main/resources/ui/src/pages/LoginPage/SignInPage.tsx
```diff
@@ -224,7 +224,7 @@ const SignInPage = () => {
                       autoFocus
                       placeholder={
                         isAuthProviderLDAP
-                          ? t('label.email')
+                          ? t('label.username')
```
chirag-madlani, line 227: same as above

chirag-madlani (changes requested):
Thanks for the contribution 🚀 
Can you address my comments? I will merge it in afterwards

openmetadata-ui/src/main/resources/ui/src/pages/LoginPage/SignInPage.tsx
chirag-madlani: Yes, but I cannot have it for others as well right? It's just the label so we want it to be generic for everyone.

thiagopirex: Ok.

metalshanked: email seems specific. even we use "username".  Some generic perhaps instead of user or email or even Username/Email ?

chirag-madlani (approved):

openmetadata-ui/src/main/resources/ui/src/pages/LoginPage/SignInPage.tsx
```diff
@@ -268,13 +268,14 @@ const SignInPage = () => {
...
-
-                {(authConfig?.enableSelfSignup || isAuthProviderLDAP) && (
+                {(!isAuthProviderLDAP) && (
+                  <div className="mt-8" onClick={onClickForgotPassword}>
+                    <Typography.Link underline data-testid="forgot-password">
+                      {t('label.forgot-password')}
+                    </Typography.Link>
+                  </div>
+                )}
+                {(authConfig?.enableSelfSignup) && (!isAuthProviderLDAP) && (
```
harshach, line 278: @chirag-madlani lets change based on one config. Why are we making it allowable for if its LDAP

PR #15371, opened 2024-02-27
[Fix] Don't update mutually exclusive on update calls

mohityadav766:
### Describe your changes:

- Don't allow updation of mutually Exclusive
- Only during creation update we can set mutually exclusive

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ClassificationRepository.java
```diff
@@ -121,9 +121,8 @@ public ClassificationUpdater(
     @Transaction
     @Override
     public void entitySpecificUpdate() {
-      // TODO mutuallyExclusive from false to true?
       recordChange(
```
harshach, line 124: @mohityadav766 lets remove the recordChange here, which should ignore any changes via patcha and we should also block the PUT calls to update it

mohityadav766: @harshach making it recordCHange original to original will keep this as it is in PUT and PATCH both

mohityadav766:
closing taken care of here #16068

PR #15394, opened 2024-02-28
MINOR - Add `/status` endpoint

pmbrull:
### Describe your changes:

To help us easily check database access, ES health, Pipeline Service Client, jwks configuration and migrations.

It will require changes in Collate to define the MigrationValidationClient there to handle Collate-only migrations

```bash
❯ http GET http://localhost:8585/api/v1/system/healthcheck "Authorization: Bearer ${JWT_TOKEN}"
HTTP/1.1 200 OK
Content-Encoding: gzip
Content-Length: 354
Content-Type: application/json
Date: Wed, 28 Feb 2024 13:48:01 GMT
Vary: Accept-Encoding

{
    "database": {
        "description": "Validate that we can properly run a query against the configured database.",
        "passed": true
    },
    "jwks": {
        "description": "Validate that the ingestion-bot JWT token can be properly decoded.",
        "passed": true
    },
    "migrations": {
        "description": "Validate that all the necessary migrations have been properly executed.",
        "message": "Found the versions [[1.1.0, 1.1.1, 1.1.2, 1.1.5, 1.1.6, 1.1.7, 1.2.0, 1.2.1, 1.2.3, 1.2.4, 1.3.0, 1.2.0-collate, 1.2.3-collate, 1.3.0-collate, 1.3.1, 1.4.0]], but expected [[1.1.0, 1.1.1, 1.1.2, 1.1.5, 1.1.6, 1.1.7, 1.2.0, 1.2.1, 1.2.3, 1.2.4, 1.3.0, 1.3.1, 1.4.0]]",
        "passed": false
    },
    "pipelineServiceClient": {
        "description": "Validate that the pipeline service client is available.",
        "passed": true
    },
    "searchInstance": {
        "description": "Validate that the search client is available.",
        "message": "Search client is not available",
        "passed": false
    }
}

```

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


conf/openmetadata.yaml
```diff
@@ -151,6 +151,7 @@ migrationConfiguration:
   flywayPath: "./bootstrap/sql/migrations/flyway"
   nativePath: "./bootstrap/sql/migrations/native"
   extensionPath: ""
+  migrationValidationClass: ${MIGRATION_VALIDATION_CLASS:-"org.openmetadata.service.migration.OpenMetadataMigrationValidation"}
```
harshach, line 154: do we need to make this configurable?

pmbrull: In Collate the migrations will be different (since we'll have 1.2.1-collate, etc.)

sushi30:
Some thoughts:
1. `validation` in the context of web servies is usually for ser/de (validating a json).
2. This is more of a health check. As such makes more sense to add it under something like `/healthcheck?full=true`.
3. This is operational information. probably makes some sense to collect it periodically and expose it via Prometheus metrics. (cc: @tutte)

tutte:
> Some thoughts:
> 
> 1. `validation` in the context of web servies is usually for ser/de (validating a json).
> 2. This is more of a health check. As such makes more sense to add it under something like `/healthcheck?full=true`.
> 3. This is operational information. probably makes some sense to collect it periodically and expose it via Prometheus metrics. (cc: @tutte)

Agree. Although is a _healthcheck-like_, this shouldn't replace the health endpoint and the healthcheck should be a `return 200`.
To me a good route would be something like `/status` if qs is not an option.
In any case, prometheus metrics should expose this information

pmbrull:
> > Some thoughts:
> > 
> > 1. `validation` in the context of web servies is usually for ser/de (validating a json).
> > 2. This is more of a health check. As such makes more sense to add it under something like `/healthcheck?full=true`.
> > 3. This is operational information. probably makes some sense to collect it periodically and expose it via Prometheus metrics. (cc: @tutte)
> 
> Agree. Although is a _healthcheck-like_, this shouldn't replace the health endpoint and the healthcheck should be a `return 200`. To me a good route would be something like `/status` if qs is not an option. In any case, prometheus metrics should expose this information

I like he `/status`. Updated. Also, this is not replacing the existing healthcheck. Ingestion Pipeline health is already a prom metric. I can do followups for db/ES

harshach (approved):

conf/openmetadata.yaml
harshach: shouldn't the migration validation should be same no?
why hardcode this into the code, instead of reading from file system of bootstrap directory and list out the versions? @pmbrull 
otherwise everyone needs to remember to add this to this part of the code again

tutte (approved):

conf/openmetadata.yaml
pmbrull: make sense, I'll update this

tutte (approved):

PR #15440, opened 2024-03-04
DOC - Updated custom sql documentation

TeddyCr:


openmetadata-docs/content/v1.2.x/connectors/ingestion/workflows/data-quality/tests.md
```diff
@@ -270,19 +270,22 @@ parameterValues:
 ```
 
 ### Table Custom SQL Test
-Write you own SQL test. The test will pass if the following condition is met:
-- The query result return 0 row
+Write you own SQL test. When writting your query you can use 2 strategies:
+- `ROWS` (default): expects the query to be written as `SELECT <field>, <field> FROM <foo> WHERE <condition>`. **Note** if your query returns a large amount of rows it might cause an "Out Of Memeory" error. In this case we recoomend you to use the `COUNT` startegy.
+- `COUNT`: expects the query to be written as `SELECT COUNT(<field>) FROM <foo> WHERE <condition>`.
```
harshach, line 275: add a section for threshold and on how to think abou writing a test.
There is docs below, but looks everyone overlooks those. So to be make it clear,lets add it here

openmetadata-docs/content/v1.2.x/connectors/ingestion/workflows/data-quality/tests.md
```diff
@@ -270,19 +270,22 @@ parameterValues:
...
 
 **Properties**
 
 * `sqlExpression`: SQL expression
+* `strategy`: one of `ROWS` or `COUNT`
+* `threshold`: and integerer defining the threshold above which the test should fail (default to 0 if not specified)
```
ayush-shah, line 281:
```suggestion
* `threshold`: and integer defining the threshold above which the test should fail (default to 0 if not specified)
```

openmetadata-docs/content/v1.3.x/connectors/ingestion/workflows/data-quality/tests.md
```diff
@@ -292,19 +292,22 @@ Validate a list of table column name matches an expected set of columns
...
-- The query result return 0 row
+Write you own SQL test. When writting your query you can use 2 strategies:
+- `ROWS` (default): expects the query to be written as `SELECT <field>, <field> FROM <foo> WHERE <condition>`. **Note** if your query returns a large amount of rows it might cause an "Out Of Memeory" error. In this case we recoomend you to use the `COUNT` startegy.
+- `COUNT`: expects the query to be written as `SELECT COUNT(<field>) FROM <foo> WHERE <condition>`.
 
 **Properties**
 
 * `sqlExpression`: SQL expression
+* `strategy`: one of `ROWS` or `COUNT`
+* `threshold`: and integerer defining the threshold above which the test should fail (default to 0 if not specified)
```
ayush-shah, line 303:
```suggestion
* `threshold`: and integer defining the threshold above which the test should fail (default to 0 if not specified)
```

ayush-shah (approved):

PR #15473, opened 2024-03-06
MINOR: add the systemApp property

sushi30:
### Describe your changes:

Add the option for an app to be defined as a system app so it cannot be uninstalled.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
@@ -68,6 +68,7 @@ import org.openmetadata.service.OpenMetadataApplicationConfig;
 import org.openmetadata.service.apps.ApplicationHandler;
 import org.openmetadata.service.apps.scheduler.AppScheduler;
 import org.openmetadata.service.clients.pipeline.PipelineServiceClientFactory;
+import org.openmetadata.service.exception.CatalogExceptionMessage;
 import org.openmetadata.service.exception.EntityNotFoundException;
 import org.openmetadata.service.jdbi3.AppRepository;
 import org.openmetadata.service.jdbi3.CollectionDAO;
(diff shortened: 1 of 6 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
```diff
@@ -686,6 +687,9 @@ public Response delete(
           @PathParam("name")
           String name) {
     App app = repository.getByName(null, name, repository.getFields("bot,pipelines"));
+    if (app.getSystemApp()) {
+      throw new AuthorizationException("System App cannot be uninstalled");
```
harshach, line 691: can we not use AuthorizationException here or introduce a new exception.
AuthorizationException should be used specifically for permissions not for anything else.

openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
```diff
@@ -711,6 +716,9 @@ public Response delete(
       @Parameter(description = "Id of the App", schema = @Schema(type = "UUID")) @PathParam("id")
           UUID id) {
     App app = repository.get(null, id, repository.getFields("bot,pipelines"));
+    if (app.getSystemApp()) {
+      throw new AuthorizationException("System App cannot be uninstalled");
```
harshach, line 720: same here

openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
sushi30: changed to BadRequest

openmetadata-service/src/main/java/org/openmetadata/service/resources/apps/AppResource.java
sushi30: same

harshach (approved):

PR #15493, opened 2024-03-08
fix(ui): glossary details path with tabs

chirag-madlani:
### Describe your changes:

Fixes: glossary sub path was not correctly fetched

I worked on ... because ...

https://github.com/open-metadata/OpenMetadata/assets/12962843/78fce077-5474-4dab-bec4-7afe4d8b64e5

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ShaileshParmar11 (approved):

openmetadata-ui/src/main/resources/ui/src/utils/EntityUtilClassBase.test.ts
```diff
@@ -109,7 +110,11 @@ describe('EntityUtilClassBase', () => {
     const fqn = 'testingGlossary';
     entityUtil.getEntityLink(EntityType.GLOSSARY, fqn);
 
-    expect(getGlossaryPath).toHaveBeenCalledWith(fqn);
+    expect(getGlossaryTermDetailsPath).toHaveBeenCalledWith(
+      fqn,
+      undefined,
+      undefined
+    );
   });
```
Ashish8689, line 118: Aren't we need to test the return value of it ? With the tab, subTab and only with the fqn.

Ashish8689 (approved):

openmetadata-ui/src/main/resources/ui/src/utils/EntityUtilClassBase.test.ts
chirag-madlani: Return values will be verified with test of that functions, Here we just need to verify the correct function call  with params based on entityType

PR #15498, opened 2024-03-08
- Prevent Job Init again and again

mohityadav766:
### Describe your changes:

- Store Job Instances in Factory
- Prevent Job init again and again

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach:
@mohityadav766 since we are changing the interface we need to fix it for other apps.
cc @pmbrull @sushi30 please take a look and see the changes you need to make

pmbrull:
> @mohityadav766 since we are changing the interface we need to fix it for other apps. cc @pmbrull @sushi30 please take a look and see the changes you need to make

We can merge this. We'll fix later, but at least we should be able to validate the improvements on sandbox-beta. Thanks!

openmetadata-service/src/main/java/org/openmetadata/service/apps/scheduler/AppJobFactory.java
```diff
@@ -0,0 +1,81 @@
...
+import org.quartz.JobDataMap;
+import org.quartz.JobDetail;
+import org.quartz.Scheduler;
+import org.quartz.SchedulerException;
+import org.quartz.simpl.SimpleJobFactory;
+import org.quartz.spi.TriggerFiredBundle;
+
+@Slf4j
+public class AppJobFactory extends SimpleJobFactory {
+  private static final ConcurrentHashMap<String, Job> jobMap = new ConcurrentHashMap<>();
```
harshach, line 23: always keep Map<String, Job> jobMap as the type

openmetadata-service/src/main/java/org/openmetadata/service/apps/scheduler/AppJobFactory.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Override
+  public Job newJob(TriggerFiredBundle bundle, Scheduler scheduler) throws SchedulerException {
+    JobDetail jobDetail = bundle.getJobDetail();
+    JobDataMap jobDataMap = bundle.getJobDetail().getJobDataMap();
+    App jobApp = JsonUtils.readOrConvertValue(jobDataMap.get(APP_INFO_KEY), App.class);
+    Class<? extends Job> jobClass = jobDetail.getJobClass();
+
+    try {
+      if (jobMap.containsKey(jobApp.getClassName())) {
```
harshach, line 41: This is not necessary, it can be done as singleton instead of using map to store the object

mohityadav766:
this change isn't needed

PR #15510, opened 2024-03-11
MINOR - Prepare AI chat feed

pmbrull:
### Describe your changes:

Required for https://github.com/open-metadata/openmetadata-collate/pull/373

We're adding the `ai` property to the threads so that we can pass AI generated queries information when chatting with AI

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplication.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplication.java
@@ -61,6 +61,7 @@ import org.jdbi.v3.sqlobject.SqlObjects;
 import org.openmetadata.schema.api.security.AuthenticationConfiguration;
 import org.openmetadata.schema.api.security.AuthorizerConfiguration;
 import org.openmetadata.schema.services.connections.metadata.AuthProvider;
+import org.openmetadata.service.apps.ApplicationHandler;
 import org.openmetadata.service.apps.scheduler.AppScheduler;
 import org.openmetadata.service.config.OMWebBundle;
 import org.openmetadata.service.config.OMWebConfiguration;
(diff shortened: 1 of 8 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplication.java
```diff
@@ -206,6 +205,7 @@ public void run(OpenMetadataApplicationConfig catalogConfig, Environment environ
     // start event hub before registering publishers
     EventPubSub.start();
 
+    ApplicationHandler.initialize(catalogConfig);
```
pmbrull, line 208: Initialize the handler early, since we'll use it in Collate MetaPilot Resource

openmetadata-service/src/main/java/org/openmetadata/service/apps/ApplicationHandler.java
```diff
@@ -28,6 +28,9 @@ private ApplicationHandler(OpenMetadataApplicationConfig config) {
   }
 
   public static void initialize(OpenMetadataApplicationConfig config) {
+    if (instance != null) {
```
pmbrull, line 31: Don't init the handler twice

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/FeedRepository.java
```diff
@@ -935,6 +935,10 @@ private boolean fieldsChanged(Thread original, Thread updated) {
                 || !Objects.equals(
                     original.getAnnouncement().getEndTime(),
                     updated.getAnnouncement().getEndTime())))
+        || (original.getAi() == null && updated.getAi() != null)
```
harshach, line 938: why is this specific Ai() is necessary

pmbrull: here I'm storing the last updated query from the conversation with the MetaPilot to easily pass the history to the API

pmbrull: should also help users go over the thread history but have a quick look at what was the outcome

harshach: if the thread belongs to a bot in this case it would be Metapilot, do we need this specific condition of getAi() type? and also it should be getAI() as its abbreviation

pmbrull:
change from ai to chatbot

harshach (approved):

PR #15577, opened 2024-03-14
Fix #15563: Fixed incorrect col ordering after patch request from ingestion

### LINKED ISSUE #15563: Metadata PATCH unexpected behaviour
**Context**
Metadata ingestion performs a PATCH request when updating information about entities. When patching arrays (e.g. columns for table entity) if the order of the array is different from the existing array for the entity, we'll be patching the wrong element.

**Example**
Let's say my table column names are ordered as follows 
```
['jurisdiction name', 'count participants', 'count female', 'percent female', 'count male', 'percent male', 'partition_0']
```
If the next time I am processing metadata the order of the column I retrieve from the system change let's say as shown below[...]
```
['partition_0', 'jurisdiction name', 'count participants', 'count female', 'percent female', 'count male', 'percent male',]
```
[...] my patch request will look like this
```
[{'op': 'add', 'path': '/tablePartition', 'value': {'columns': [...], 'intervalType': 'COLUMN-VALUE'}}
{'op': 'replace', 'path': '/tableType', 'value': 'Partitioned'}
{'op': 'replace', 'path': '/columns/0/name', 'value': 'partition_0'}
{'op': 'replace', 'path': '/columns/0/dataTypeDisplay', 'value': 'string'}
{'op': 'replace', 'path': '/columns/0/dataType', 'value': 'STRING'}
{'op': 'replace', 'path': '/columns/1/name', 'value': 'jurisdiction name'}
{'op': 'replace', 'path': '/columns/2/name', 'value': 'count participants'}
{'op': 'replace', 'path': '/columns/3/name', 'value': 'count female'}
{'op': 'replace', 'path': '/columns/3/dataTypeDisplay', 'value': 'bigint'}
{'op': 'replace', 'path': '/columns/3/dataType', 'value': 'BIGINT'}]
```
Where I will be patching column name at index 0 with value `partition_0` effectively changing the name of the column resulting in a mis match description.

**How to reproduce it**
Spin up OM 1.2.5 with an Athena service (ingest `om_test_glue`). Add description to columns. Upgrade to 1.3.1 (we added support for table partition in athena in 1.3.1 and the partitioned column is at index 0 in the list of columns). You will now see shifted description -- see screenshots.

Screenshot 2024-03-14 at 10 13 09 AM
Screenshot 2024-03-14 at 9 26 46 AM

OnkarVO7:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/15563

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/models/patch_request.py
+++ b/ingestion/src/metadata/ingestion/models/patch_request.py
@@ -12,7 +12,6 @@
 Pydantic definition for storing entities for patching 
 """
 import json
-import logging
 from typing import Dict, List, Optional
 
 import jsonpatch
@@ -21,8 +20,6 @@ from pydantic import BaseModel
 from metadata.ingestion.api.models import Entity, T
 from metadata.ingestion.ometa.mixins.patch_mixin_utils import PatchOperation
 from metadata.ingestion.ometa.utils import model_str
-
-logger = logging.getLogger("metadata")
 
 
 class PatchRequest(BaseModel):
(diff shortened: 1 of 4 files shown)


ingestion/src/metadata/ingestion/models/patch_request.py
```diff
@@ -192,20 +203,61 @@ def build_patch(
...
 
+def _sort_array_entity_fields(
+    source: T,
+    destination: T,
+    array_entity_fields: Optional[List] = None,
+):
+    """
+    Sort the array entity fields to make sure the order is consistent
+    """
+    for field in array_entity_fields:
```
harshach, line 230: @OnkarVO7 how are we sorting it, basically what element are we  choosing to sort on?

OnkarVO7: We are sorting in the same order how the columns are stored on OpenMetadata side

harshach: @OnkarVO7 any reason to apply this sorting , is the response not coming with that sort order?

OnkarVO7: @harshach yes the response for columns from the source for a few connectors is coming in random order as described in https://github.com/open-metadata/OpenMetadata/issues/15563. Due to this the patch request for the columns list does not get formed correctly

Hence we are sorting the response from source according to the columns order stored on OM side

TeddyCr (approved):

ingestion/src/metadata/ingestion/models/patch_request.py
```diff
@@ -192,20 +206,64 @@ def build_patch(
...
+                        # Remove the updated attribute from the destination dictionary
+                        del destination_dict[model_str(source_attr.name)]
+                    else:
+                        updated_attributes.append(None)
+
+                # Combine the updated attributes with the remaining destination attributes
+                final_attributes = updated_attributes + list(destination_dict.values())
+                setattr(destination, field, final_attributes)
+    except Exception as exc:
+        logger.warning(f"Unable to sort fields due to: {exc}")
```
TeddyCr, line 261: What happens if we fall into this flow, do we patch with un sorted fields?

TeddyCr (commented):
Added a comment

ingestion/src/metadata/ingestion/models/patch_request.py
OnkarVO7: Yes it will patch with unsorted fields.

TeddyCr (approved):

pmbrull (approved):

PR #15589, opened 2024-03-15
#367C - Suggestions Accept / Reject All 

pmbrull:
### Describe your changes:

https://github.com/open-metadata/openmetadata-collate/issues/367

- Add accept/reject all to suggestions
- Add bot displayName (from App displayName)

Still need a solution for the logo cc @karanh37 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.Collectors;
 import javax.json.JsonPatch;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
@@ -154,30 +155,29 @@ public class SuggestionRepository {
 
   @Getter
   public static class SuggestionWorkflow {
-    protected final Suggestion suggestion;
-    protected final MessageParser.EntityLink entityLink;
+    protected final EntityRepository<?> repository;
+    // The workflow is applied to a specific entity at a time
+    protected final EntityInterface entity;
 
-    SuggestionWorkflow(Suggestion suggestion) {
-      this.suggestion = suggestion;
-      this.entityLink = MessageParser.EntityLink.parse(suggestion.getEntityLink());
+    SuggestionWorkflow(EntityRepository<?> repository, EntityInterface entity) {
+      this.repository = repository;
+      this.entity = entity;
     }
 
-    public EntityInterface acceptSuggestions(
-        EntityRepository<?> repository, EntityInterface entityInterface) {
+    public EntityInterface acceptSuggestions(Suggestion suggestion, EntityInterface entity) {
+      MessageParser.EntityLink entityLink = MessageParser.EntityLink.parse(suggestion.getEntityLink());
       if (entityLink.getFieldName() != null) {
-        entityInterface =
-            repository.applySuggestion(
-                entityInterface, entityLink.getFullyQualifiedFieldValue(), suggestion);
-        return entityInterface;
+        return repository.applySuggestion(
+              entity, entityLink.getFullyQualifiedFieldValue(), suggestion);
       } else {
         if (suggestion.getType().equals(SuggestionType.SuggestTagLabel)) {
-          List<TagLabel> tags = new ArrayList<>(entityInterface.getTags());
+          List<TagLabel> tags = new ArrayList<>(entity.getTags());
           tags.addAll(suggestion.getTagLabels());
-          entityInterface.setTags(tags);
-          return entityInterface;
+          entity.setTags(tags);
+          return entity;
         } else if (suggestion.getType().equals(SuggestionType.SuggestDescription)) {
-          entityInterface.setDescription(suggestion.getDescription());
-          return entityInterface;
+          entity.setDescription(suggestion.getDescription());
+          return entity;
         } else {
           throw new WebApplicationException("Invalid suggestion Type");
         }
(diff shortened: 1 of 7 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
```diff
@@ -154,32 +155,32 @@ public void deleteSuggestionInternalForAnEntity(EntityInterface entity) {
...
+    protected final EntityRepository<?> repository;
+    // The workflow is applied to a specific entity at a time
+    protected final EntityInterface entity;
 
-    SuggestionWorkflow(Suggestion suggestion) {
-      this.suggestion = suggestion;
-      this.entityLink = MessageParser.EntityLink.parse(suggestion.getEntityLink());
+    SuggestionWorkflow(EntityRepository<?> repository, EntityInterface entity) {
+      this.repository = repository;
+      this.entity = entity;
```
pmbrull, line 164: @harshach made the workflow dependant on a repository & entity, rather than at the suggestion level

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
```diff
@@ -333,4 +418,9 @@ private List<Suggestion> getSuggestionList(List<String> jsons) {
     }
     return suggestions;
   }
+
+  public final List<Suggestion> listAll(SuggestionFilter filter) {
+    ResultList<Suggestion> suggestionList = listAfter(filter, Integer.MAX_VALUE - 1, "");
```
pmbrull, line 423: need to -1 since the `listAfter` does `limit + 1`

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -751,7 +751,7 @@ public Table applySuggestion(EntityInterface entity, String columnFQN, Suggestio
         } else if (suggestion.getType().equals(SuggestionType.SuggestDescription)) {
           col.setDescription(suggestion.getDescription());
         } else {
-          throw new WebApplicationException("Invalid suggestion Type");
+          throw new SuggestionException("Invalid suggestion Type");
```
pmbrull, line 754: Update exceptions to have suggestion-specific exception handling

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
harshach: do we need to pass repostory? since its a singleton we can access it anywhere @pmbrull

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
```diff
@@ -190,12 +191,24 @@ public RestUtil.PutResponse<Suggestion> acceptSuggestion(
...
+  public RestUtil.PutResponse<List<Suggestion>> acceptSuggestionList(
+      UriInfo uriInfo,
+      List<Suggestion> suggestions,
+      SecurityContext securityContext,
+      Authorizer authorizer) {
+    acceptSuggestionList(suggestions, securityContext, authorizer);
+    List<Suggestion> updatedHref =
+        suggestions.stream()
+            .map(suggestion -> SuggestionsResource.addHref(uriInfo, suggestion))
+            .collect(Collectors.toList());
```
harshach, line 208: use toList() instead of Collectors.toList()

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
```diff
@@ -220,6 +238,70 @@ protected void acceptSuggestion(
     update(suggestion, user);
   }
 
+  protected void acceptSuggestionList(
+      List<Suggestion> suggestions, SecurityContext securityContext, Authorizer authorizer) {
+    String user = securityContext.getUserPrincipal().getName();
```
harshach, line 243: we need to validate if this user has permissions to edit descripton or edit tags

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
```diff
@@ -220,6 +238,70 @@ protected void acceptSuggestion(
...
+
+    // Entity being updated
+    EntityInterface entity = null;
+    EntityRepository<?> repository = null;
+    String origJson = null;
+    SuggestionWorkflow suggestionWorkflow = null;
+    SuggestionType suggestionType = getTypeFromSuggestions(suggestions);
```
harshach, line 250: this should be a single type, what if user has suggestions for description and tags. This will merge them both

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
```diff
@@ -220,6 +238,70 @@ protected void acceptSuggestion(
...
+      }
+      // update entity with the suggestion
+      entity = suggestionWorkflow.acceptSuggestion(suggestion, entity);
+    }
+
+    // Patch the entity with the updated suggestions
+    String updatedEntityJson = JsonUtils.pojoToJson(entity);
+    JsonPatch patch = JsonUtils.getJsonPatch(origJson, updatedEntityJson);
+
+    OperationContext operationContext = new OperationContext(repository.getEntityType(), patch);
```
harshach, line 276: would be better to this do early in the method rather than going through parsing of suggestions.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
```diff
@@ -220,6 +238,70 @@ protected void acceptSuggestion(
     update(suggestion, user);
   }
```
harshach, line 240: lets add @transaction mark on this method

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SuggestionRepository.java
pmbrull: I might be missing something here. But the idea is that we go over all the suggestions and build a single patch from them all (instead of patching 1 by 1). So we can only build this once all suggestions have been reviewed no?

harshach (approved):

PR #15612, opened 2024-03-19
Fixes #15614: remove data from quartz

### LINKED ISSUE #15614: Implement app context for scheduled apps
https://github.com/open-metadata/OpenMetadata/issues/15613

implement the **app context** part.

sushi30:
## Changes

1. Remove data from quartz application. Store only app id.
2. Added exception handling when trying to trigger an app.
3. Save bot info in the app data in DB.
4. Added runtime properties in `AbstractNativeApplication.execute`.
5. Added migration in runtime to the new context. This cannot be done with the migration framework because it requires de-serializing java byte-code from the `QUARTS_JOB_DETAILS` table.

 https://github.com/open-metadata/OpenMetadata/issues/15614

## Testing

The migration was tested in the following form:

1. Deploy clean db.
2. Checkout main branch.
3. Run the app.
4. Run the re-indexing app. **Validate success**
5. Stop the app.
6. Checkout `remove-app-context-quartz` branch.
7. Run the app.
8. Run the re-indexing app. **Validate success**.

### Details

You can go into the db and make sure that the expected changes are observed in `QUARTZ_JOB_DETAILS` and `installed_apps`

#### QUARTZ_JOB_DETAILS

Removed application data

**BEFORE**
Screenshot 2024-03-19 at 16 59 05

**AFTER**
Screenshot 2024-03-19 at 16 59 12

#### installed_apps

Added bot information

**BEFORE**

Screenshot 2024-03-19 at 17 03 46

**AFTER**
Screenshot 2024-03-19 at 17 00 44

#
### Type of change:

- [x] Improvement

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/apps/scheduler/AppScheduler.java
```diff
@@ -177,7 +176,7 @@ public void deleteScheduledApplication(App app) throws SchedulerException {
 
   private JobDetail jobBuilder(App app, String jobIdentity) throws ClassNotFoundException {
     JobDataMap dataMap = new JobDataMap();
-    dataMap.put(APP_INFO_KEY, JsonUtils.pojoToJson(app));
+    dataMap.put(APP_ID_KEY, app.getId());
```
mohityadav766, line 179: @sushi30 we should work with AppName rather than appId I think

sushi30: @mohityadav766 by using appName we introduce 2 isues:
1. Risk of shading the appName being shaded by another app (for external applications).
2. We remove the capability to deploy multiple instances of the same app. (example: multiple re-indexing jobs for different languages)

mohityadav766: @sushi30 
we need to anyway have only a single application instance anyway, also user will never be able to register apps with same name

sushi30: @mohityadav766
ID is the way we link entities throughout the application. Using the `name` for the application will break this convention. I am yet to be persuaded regarding the reason.

1. An app id should not change for the lifecycle of the installation.
2. While we currently only support one instance of the applications, I think we should leave the door open for enabling more complex usages.
3. An app name can change in the future (not supported at the moment, but might be...). An ID is supposed to stay constant.

mohityadav766 (approved):

PR #15618, opened 2024-03-20
Fix(#15296): Issue with the same service name

### LINKED ISSUE #15296: Issue with the same service name
Hi,

**Describe the bug**
If you use the same name for the same service in Snowflake and Tableau, for example: BIDA. The ingestions between services are crossed, i.e. Tableau ingestions appear in the Snowflake service and vice versa.

This example shows the Tableau service showing snowflake ingestion:
image

We believe it is case sensitive, i.e. if we had called it Bida this would not have happened, for other services that we called the same but in upper case in tableau and lower case in snowflake we did not have this problem.
image

This problem, we have not seen that the ingestions have not had any problems, it is just that ingestions appeared where they did not belong.

For us it is not a problem, because we have destroyed the tableau service, and called it something else, but we consider it a bug and have reported it as such.

**Version:**
 - OpenMetadata version: 1.3
 - OpenMetadata Ingestion package version: 1.3.0

thanks,
Carlos
Hotelbeds

Sachin-chaurasiya:
### Describe your changes:

Fixes #15296 

I worked on issue #15296 

https://github.com/open-metadata/OpenMetadata/assets/59080942/b6136685-f7e2-41b2-9572-0e0ed414d944

#
### Type of change:

- [x] Bug fix

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/cypress/constants/constants.ts
+++ b/openmetadata-ui/src/main/resources/ui/cypress/constants/constants.ts
@@ -465,6 +465,7 @@ export const service = {
   description: 'This is a Glue service',
   newDescription: 'This is updated Glue service description',
   Owner: 'Aaron Johnson',
+  serviceType: 'databaseService',
 };
 
 export const SERVICE_TYPE = {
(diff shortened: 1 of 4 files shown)


chirag-madlani (dismissed):

chirag-madlani (commented):
@Sachin-chaurasiya can you also check around tests for the serviceType?

pmbrull (dismissed):

Sachin-chaurasiya:
> @Sachin-chaurasiya can you also check around tests for the serviceType?

Thanks @chirag-madlani , I have pushed the changes.

PR #15680, opened 2024-03-25
MINOR: Fix jsonpatch operation order

IceS2:
## Overview

We encounter some errors when applying Patches that look like `'["lorem ipsum"]' contains no element for index 1 and for '1'`.

This happens when a [JSON Patch](https://jsonpatch.com/) is badly formatted and tries to make certain operations on an. Array index that does not exist.

I saw that we were ordering the received Patch in the backend to fix some similar issues we faces in the past but while reading through the RFC I understood that this is not an issue from the JSON library we are using, as the RFC states the following:

> Operations are applied sequentially in the order they appear in the array.

This means that the one that builds the JSON Patch is responsible for actually building a correct JSON Patch to apply and when we reorder the operations on the backend to fix a certain case we potentially break some other cases (In the specific case I was working this was happening. The JSON Patch sent was valid but reordering the operations was making it invalid)

## Changes

- **JsonUtils stops reordering the JSONPatch operations and the responsibility of being given a proper patch is on the client**

    If we modify the order of the JSONPatch received we potentially make it invalid. Since we don't know have the context on how the patch was created we should just apply the operations sequentially as the RFC states.

- **Changed the library we use to create a JSONPatch based on JSONDiff. We were using the [johnzon](https://johnzon.apache.org/) implentation and we are now using the [json-patch](https://github.com/java-json-tools/json-patch) library**

    Tests are using the JSONDiff methods to create a JSONPatch and a few were failing after this change due to johnzon implementing the `remove` operation on array wrongly.
    We can see [here](https://github.com/apache/johnzon/blob/v1.2.21/johnzon-core/src/main/java/org/apache/johnzon/core/JsonPatchDiff.java#L66) that they are removing the elements from the lower index to the highest index. It should either be done backwards or removing always the same index.
    json-patch actually opts for the second option and we can compare the generated patches below (I Just added the relevant part):

    **johnzon patch**
  ```json
  [
    ...
    {
      "op": "remove",
      "path": "/columns/2/tags/0"
    },
    {
      "op": "remove",
      "path": "/columns/2/tags/1"
    },
    {
      "op": "remove",
      "path": "/columns/2/tags/2"
    },
    ...
  ]
  ```

  **json-patch patch**
    ```json
    [
      ...
      {
        "op": "remove",
        "path": "/columns/2/tags/0"
      },
      {
        "op": "remove",
        "path": "/columns/2/tags/0"
      },
      {
        "op": "remove",
        "path": "/columns/2/tags/0"
      }
    ]
    ```

- **Updated the Ingestion Framework to keep the JSONPatch `replace` operation with a `None` value generated by the [jsonpatch library](https://github.com/stefankoegl/python-json-patch). Instead of changing it for a `remove` operation we are appending the `remove` operations at the end**

    For some reason the Python JSONPatch library we use decided to send `replace` operations to a `None` value instead of a `remove` operations in arrays (Maybe to avoid the index issue we see with johnzon). We don't want to save arrays with `None` items (and actually this breaks in the backend). 
    In order to fix this I decided to append one `remove` operation for each `replace` to a `None` value while taking into account the index drift that might happen on arrays

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/util/JsonUtils.java
```diff
@@ -193,40 +192,15 @@ public static <T> T convertValue(Object object, TypeReference<T> toValueTypeRef)
...
-    // -----------------------------------------------------------
-    // JSON patch modification 1 - Reorder the operations
-    // -----------------------------------------------------------
-    // JsonPatch array operations are not handled correctly by johnzon libraries. Example, the
-    // following operation:
-    // {"op":"replace","path":"/tags/0/tagFQN","value":"User.BankAccount"}
-    // {"op":"replace","path":"/tags/0/labelType","value":"MANUAL"}
-    // {"op":"remove","path":"/tags/1"}
-    // {"op":"remove","path":"/tags/2"}
-    // Removes second array element in a 3 array field /tags/1
```
sushi30, line 206: I suggest adding a test case for this scenario. This seems like a contentious issue that we should track for regressions if we decide to handle it differently in the future.

IceS2: Hey, do you mean testing that we are raising an error fi the operation comes on the wrong order? Didn't get it.

Thanks!

mohityadav766 (dismissed):

pmbrull (changes requested):
@IceS2 as discussed, let's add a couple of unit tests on the `JsonPatchUpdater`

pmbrull (dismissed):

pmbrull (approved):

PR #15702, opened 2024-03-26
Fix 15576 - Eval Data Type issue fix

### LINKED ISSUE #15576: Upgrade caused column type change resulting in profiler error
It looks like the OM version upgrade (from 1.0.5 to 1.3.1) changed the data type of some existing columns named Version. They were all defined as string before, but were changed if the value looked like an int or float. So 0.0.1 stayed as string, 0.1 became float, and 1 became int. The old Version column was deleted and the new Version column was added with the new data type. The sourceHash column (internal use only) was also added in all cases at the same time.

I've attached two json files.
	• analysis.json is the schema file
	• Example Analysis.json is an example json file with the same format as the ones causing a problem.

We would like to know how to change the data type for the version column back to string without having to drop and recreate the entities.  I've rerun the Analyses ingestion and the data type did not change back to string, though perhaps a new file timestamp is not enough to cause an update.  We could not find a way in the UI or the API to update the data type of a column.

We discovered the data type change because the Analyses profiler run after the upgrade caused an error that seems to trace back to the data type change.  Below is part of the log showing the error.
[2024-03-12T23:14:42.063+0000] {profiler_interface.py:298} ERROR - SQALikeColumn(name='Version', type=<DataType.FLOAT: 'FLOAT'>) metric_type.value: unsupported operand type(s) for +: 'int' and 'str'
[2024-03-12T23:14:42.064+0000] {status.py:76} WARNING - SQALikeColumn(name='Version', type=<DataType.FLOAT: 'FLOAT'>) metric_type.value: unsupported operand type(s) for +: 'int' and 'str'
[2024-03-12T23:14:42.064+0000] {status.py:77} DEBUG - Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.10/site-packages/metadata/profiler/interface/pandas/profiler_interface.py", line 177, in _compute_static_metrics
    metric_resp = metric(column).df_fn(runner)
  File "/home/airflow/.local/lib/python3.10/site-packages/metadata/profiler/metrics/static/sum.py", line 53, in df_fn
    return sum(df[self.col.name].sum() for df in dfs)
TypeError: unsupported operand type(s) for +: 'int' and 'str'
During handling of the above exception, another exception occurred:
  File "/home/airflow/.local/lib/python3.10/site-packages/metadata/profiler/interface/pandas/profiler_interface.py", line 290, in compute_metrics
    row = self._get_metric_fn[metric_func.metric_type.value](
  File "/home/airflow/.local/lib/python3.10/site-packages/metadata/profiler/interface/pandas/profiler_interface.py", line 186, in _compute_static_metrics
    raise RuntimeError(exc)
RuntimeError: unsupported operand type(s) for +: 'int' and 'str'

[analysis.json](https://github.com/open-metadata/OpenMetadata/files/14604955/analysis.json)
[Example Analysis.json](https://github.com/open-metadata/OpenMetadata/files/14604957/Example.Analysis.json)

ayush-shah:
### Describe your changes:

Fixes #15576 

.Modify Parsers to get proper column types

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/profiler/interface/pandas/profiler_interface.py
```diff
@@ -171,19 +190,23 @@ def _compute_static_metrics(
         """
         import pandas as pd  # pylint: disable=import-outside-toplevel
 
+        row_dict = {}
         try:
-            row_dict = {}
+            dataframes = [df.dropna() for df in runner]
```
TeddyCr, line 195: I think if we gonna drop rows with `na` we should state very clearly in the documentation and make it clear in the migration doc as this will change the count of certain metrics.

ingestion/src/metadata/profiler/metrics/static/mean.py
```diff
@@ -121,7 +121,22 @@ def df_fn(self, dfs=None):
...
+                except Exception as err:
+                    if str(err).startswith("Could not convert") and str(err).endswith(
+                        "to numeric"
+                    ):
+                        [
+                            float(item)
+                            for item in df[self.col.name].values
+                            if item.isdigit()
+                        ]
+                        mean = None
```
TeddyCr, line 135: what are we doing here? we convert all the values to `float` and then set the mean to None? I think if we can't compute the mean better not computing it at all. Otherwise it is going to be super costly with very large data set to iterate and convert every value

ingestion/src/metadata/profiler/metrics/static/min.py
```diff
@@ -98,7 +98,9 @@ def df_fn(self, dfs=None):
         if is_quantifiable(self.col.type):
             return min((df[self.col.name].min() for df in dfs))
         if is_date_time(self.col.type):
```
TeddyCr, line 100: should we add the type to `is_date_time`. I am assuming if the type is already a datetime type then we should not have to convert it to a pd.datetime at runtime right?

ingestion/src/metadata/profiler/metrics/static/min_length.py
```diff
@@ -67,7 +67,7 @@ def df_fn(self, dfs=None):
             min_length_list = []
 
             for df in dfs:
-                if any(df[self.col.name]):
+                if any(df[self.col.name].dropna()):
```
TeddyCr, line 70: same comment here if we drop `na` we need to state it clearly

ingestion/src/metadata/profiler/processor/core.py
```diff
@@ -451,7 +454,7 @@ def _prepare_column_metrics(self) -> List:
                 column=column,
                 table=self.table,
             )
-            for column in columns
+            for column in self._columns
```
TeddyCr, line 457: you should call the property attribute not the private one `self.columns` ln. 142

ingestion/src/metadata/profiler/processor/sampler/pandas/sampler.py
```diff
@@ -136,7 +136,7 @@ def get_col_row(self, data_frame, columns: Optional[List[SQALikeColumn]] = None)
                 break
         return cols, rows
 
-    def random_sample(self):
+    def random_sample(self, is_sampled: bool = False):
```
TeddyCr, line 139: what is that argument for?

TeddyCr (changes requested):
left a few comments

ingestion/src/metadata/profiler/interface/pandas/profiler_interface.py
ayush-shah: Yes, will make that clear

ingestion/src/metadata/profiler/metrics/static/mean.py
ayush-shah: removed the logic of assigning type as float here

ayush-shah: reverted it back, the code was the before the discussion around evaluating the literal datatypes..

ingestion/src/metadata/profiler/metrics/static/min.py
ayush-shah: min_ = min((pd.to_datetime(df[self.col.name]).min() for df in dfs))
      if pd.isna(min_):
          return None
      return int(min_.timestamp() * 1000)
      
as the data is not parsed as datetime always, instead it sometimes is parsed as string
`YYYY-MM-DD HH-MM-SS` pandas considers it datetime, whereas when we try to parse `YYYY-MM-DD` using pandas, it evals it as an object / string..
one of the reasons why we need to convert it to datetime

ingestion/src/metadata/profiler/metrics/static/min_length.py
ayush-shah: will do the same 👍

ingestion/src/metadata/profiler/processor/sampler/pandas/sampler.py
ayush-shah: as we fetch the data frames ( already sampled ) this is to make sure that they don't get sampled again

TeddyCr: Do you mean when we parse the rows to evaluate the types?

ingestion/src/metadata/profiler/processor/core.py
ayush-shah: Property attribute whenever called, executes multiple times, i think it is needed only ones per table, i have modified it to be used once, and store it in init and call it then

ingestion/src/metadata/profiler/processor/sampler/pandas/sampler.py
ayush-shah: as we call `return_ometa_dataframes_sampled the dataframe comes sampled, having to do a random sample over it again won't be needed that's why the extra arg

ingestion/src/metadata/profiler/metrics/composed/null_ratio.py
```diff
@@ -48,10 +48,12 @@ def fn(self, res: Dict[str, Any]) -> Optional[float]:
...
         """
+        import pandas as pd
+
         res_count = res.get(Count.name())
         res_null = res.get(NullCount.name())
 
         if res_count is not None and res_null is not None:
-            return res_null / (res_null + res_count)
-
+            result = res_null / (res_null + res_count)
```
TeddyCr, line 57: any potential division by 0 error here?

ingestion/src/metadata/profiler/processor/core.py
```diff
@@ -145,31 +145,31 @@ def columns(self) -> List[Column]:
         by skipping the columns to ignore.
         """
 
-        if self._columns:
-            return self._columns
+        if self.column_list:
+            return self.column_list
```
TeddyCr, line 149: why do you modify the interface? If you need to set it define a setter for the property

TeddyCr (changes requested):
Left a few comments

ingestion/src/metadata/profiler/processor/core.py
ayush-shah: Yes, Let me revert that change

ingestion/src/metadata/profiler/metrics/composed/null_ratio.py
ayush-shah: Had come across NaN related issue in certain scenarios, that's why the check around pd.isna

ingestion/src/metadata/profiler/interface/pandas/profiler_interface.py
```diff
@@ -113,7 +113,7 @@ def complex_df(self):
                             if col.dataType == value
                         ][0]
                     except IndexError:
-                        pass
+                        coltype = []
```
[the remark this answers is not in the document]
ayush-shah: agree to that, that would be a better way to handle things. will modify the same, thanks

ingestion/src/metadata/profiler/processor/core.py
ayush-shah: modified, Thanks

TeddyCr (approved):

Issue #15561, opened 2024-03-14
UI: Not able to copy query from query code editor


ulixius9:
**Affected module**
UI

**Describe the bug**
Not able to copy query from query code editor, checkout the below video 

https://github.com/open-metadata/OpenMetadata/assets/39544459/4d5d111f-ae6c-4861-bf72-b226b8842d92

Sachin-chaurasiya:
@Ashish8689 , let's add the copy button like we have in other places.

PR #15731, opened 2024-03-27
#15561: supported copy button in query editor

Ashish8689:
### Describe your changes:

Fixes [#15561](https://github.com/open-metadata/OpenMetadata/issues/15561)

- supported `copy button in query editor` and also direct copy from the editor
- also supported copy from the editor for `storedProcedure` and `dashboardDataModel` in the `sql and code tab`.

image
image

https://github.com/open-metadata/OpenMetadata/assets/66266464/e04bffaa-e0ba-4a42-954b-3150039a797c

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/pages/AddQueryPage/AddQueryPage.component.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/pages/AddQueryPage/AddQueryPage.component.tsx
@@ -227,9 +227,7 @@ const AddQueryPage = () => {
                   <SchemaEditor
                     className="custom-query-editor query-editor-h-200 custom-code-mirror-theme"
                     mode={{ name: CSMode.SQL }}
-                    options={{
-                      readOnly: false,
-                    }}
+                    showCopyButton={false}
                   />
                 </Form.Item>
                 <Form.Item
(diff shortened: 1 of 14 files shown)


Sachin-chaurasiya:
@Ashish8689, I noticed that you removed this prop from several places. Could you please let me know the reason behind it?

```js
options={{
            readOnly: false,
          }}
```

Ashish8689:
yes, so we have this props already in the defaultOptions  for SchemaEditor. so don't need to pass it!
image

Sachin-chaurasiya:
> @Sachin-chaurasiya ,  so we have this props already in the defaultOptions  for SchemaEditor. so don't need to pass it!
> 
> image
> 
> 

Got it, thanks.

openmetadata-ui/src/main/resources/ui/src/pages/AddQueryPage/AddQueryPage.component.tsx
```diff
@@ -227,9 +227,7 @@ const AddQueryPage = () => {
                   <SchemaEditor
                     className="custom-query-editor query-editor-h-200 custom-code-mirror-theme"
                     mode={{ name: CSMode.SQL }}
-                    options={{
-                      readOnly: false,
-                    }}
+                    showCopyButton={false}
```
Sachin-chaurasiya, line 230: @Ashish8689 , I think we can have default option as a false in the editor, so we only have to pass this prop if we want to make it true, wdyt?

Ashish8689: so, the idea was to have the copy options as always where ever we are using editor and only some places we don't want that copy button. That the reason we have set by default as true and based on your requirement u can set it to false.

chirag-madlani (approved):

PR #15734, opened 2024-03-27
Integrate rules into elasticsearch result

HesamoddinMonfared:
I worked on applying rules to search results.
The current implementation is for elasticsearch only.

The implementation has been tested with different rules, policies and roles and the results show the correctness.

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/resources/search/ApplyOMDRulesForElasticSearch.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/resources/search/ApplyOMDRulesForElasticSearch.java
@@ -1,0 +1,234 @@
+package org.openmetadata.service.resources.search;
+
+import static org.openmetadata.service.Entity.*;
+import static org.openmetadata.service.security.DefaultAuthorizer.getSubjectContext;
+
+import es.org.elasticsearch.index.query.BoolQueryBuilder;
+import es.org.elasticsearch.index.query.QueryBuilders;
+import es.org.elasticsearch.search.builder.SearchSourceBuilder;
+import java.io.IOException;
+import java.util.*;
+import javax.ws.rs.core.SecurityContext;
+import javax.ws.rs.core.UriInfo;
+import org.openmetadata.schema.entity.classification.Classification;
+import org.openmetadata.schema.entity.policies.accessControl.Rule;
+import org.openmetadata.schema.entity.teams.User;
+import org.openmetadata.schema.type.MetadataOperation;
+import org.openmetadata.service.jdbi3.ListFilter;
+import org.openmetadata.service.resources.CollectionRegistry;
+import org.openmetadata.service.resources.tags.ClassificationResource;
+import org.openmetadata.service.resources.teams.UserResource;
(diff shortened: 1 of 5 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/search/ApplyOMDRulesForElasticSearch.java
```diff
@@ -0,0 +1,234 @@
...
+          } // End of for
+        } // End of while
+    */
+
+    // Combine searchSourceBuilder.query() with new Query that enforce policy rules
+    searchSourceBuilder.query(QueryBuilders.boolQuery().must(searchSourceBuilder.query()).must(baseQueryBuilder));
+    return searchSourceBuilder;
+  }
+
+  private BoolQueryBuilder createQueryBuilderBasedOnRule(SpelNode ast) throws IOException {
```
harshach, line 131: IMO it would be better if we codify for each expression/rule to return the search condition for it. 
Example if the condition says isOwner() only allow, then we should return the condition.getSearchCondition(SecurityContext) -> return 'owner.name=%s'.format(SecurityContext.getUsername() or owner.name=(getTeams(SecurityContext.getUsername())

openmetadata-service/src/main/java/org/openmetadata/service/resources/search/ApplyOMDRulesForElasticSearch.java
```diff
@@ -0,0 +1,234 @@
...
+    try {
+      BoolQueryBuilder boolQueryBuilder = QueryBuilders.boolQuery();
+      var keyword = ast2keyword(ast);
+
+      if (ast.getChildCount() == 0) {
```
harshach, line 136: The complexity involved enforcing these on the search side is
1. Hierarchy of Rule processing
2. We process Deny rules first before the allow rules
3. Enforcing the Resource side attribute conditions example Allow Edit only if the table contains certain tags

harshach:
@HesamoddinMonfared Thanks for taking this up. Lets visit the goals of this integration
1. Users who doesn't have view access to certain data assets shouldn't be able to see it as part of the search results
2. We are not going to worry about the Edit Access conditions as they will be taken care by the APIs itself
3. Complexity of View conditions Processing
     4. A team can set a condition that only their team members should be able to view all of it assets 
     5. A team can set a condition that only their team members should be able to view all of it assets except if it contains certain tags
     6. Admin can set a condition at Org level, any data in a domain should only viewable by users with-in that domain except for assets that are part of Data Products

who will build the conditions, the delegation should be part of the rule itself . i.e each rule should have method to equivalent to returning true or false . Example isOwner() takes the security context and returns true or false if the user who logged in is a owner of the asset or part of the team that is owner of the asset.
Similarly it should return a condition that can be applied to search query instead of this logic outside. This way any new conditions added can be implemented methodically rather adding this logic in another place which can easily introduce gaps


Condition evaluation
1. Go through hierarchy of the conditions similar to what we are doing in authorizer. 
2. Process the Deny rule first
3. All of these rules should have an operator with AND, if there are multiple rules at the same state we should look at the condition specified for that rule

PR #15758, opened 2024-03-29
[MINOR] fix antlr parser definition for entity link

TeddyCr:
### Describe your changes:
- update the regex pattern for antlr fqn or name

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Bug fix


openmetadata-spec/src/main/antlr4/org/openmetadata/schema/EntityLink.g4
```diff
@@ -132,5 +132,5 @@ ENTITY_FIELD
     
 
 NAME_OR_FQN
-    : ~(':')+ ('>')*? ~(':'|'>')+
```
TeddyCr, line 135: To me it looks like we should not match the `>` character in the name or fqn, but any character but the reserved ones (i.e. seperator and end character). Let me know if I am missing something

ayush-shah: yes, that's the goal of that pattern🙌

ayush-shah: it should not start with one or more `:`
it can have `>` in the middle but not at the end
it should not end with `:` or `>`

ayush-shah: the reason to put that there was that some entities like dashboards and pipelines that had `>` in the middle of the name which caused some issues if i remember correctly

ayush-shah (approved):

PR #15770, opened 2024-04-02
FIX #13553 Added option to exclude drafts: superset ingestion

### LINKED ISSUE #13553: Superset ingest: option to exclude drafts
We are using OpenMetadata to ingest Superset dashboards, and we have many dashboards that are in draft stage, and currently, they pollute the catalog and it is hard to find the real published dashboards.

At the ingest sourceConfig, an `includeDrafts` option. When set to `False` drafts are not ingested.

An alternate solution might be to include "draft" vs "published" as a new metadata field, and allow filtering on that field during search. That could be a nice addition, but seems like simple than not ingesting. Drafts are inherently cruft for our use case.

SumanMaharana:
### Describe your changes:

Fixes #13553

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/superset/queries.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/superset/queries.py
@@ -18,8 +18,8 @@ select
 	s.id,
 	s.slice_name,
 	s.description,
-    s.datasource_id,
-    s.viz_type,
+    	s.datasource_id,
+    	s.viz_type,
 	t.table_name,
 	t.schema,
 	db.database_name,
@@ -37,7 +37,7 @@ select
 	d.id, 
 	d.dashboard_title, 
 	d.position_json,
-    d.published,
+    	d.published,
 	au.email 
 from 
 	dashboards d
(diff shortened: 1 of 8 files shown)


ingestion/src/metadata/ingestion/source/dashboard/superset/queries.py
```diff
@@ -46,6 +47,22 @@
...
 """
 
+FETCH_PUBLISHED_DASHBOARDS = """
+select
+	d.id, 
+	d.dashboard_title, 
+	d.position_json,
+    d.published,
+	au.email 
+from
```
pmbrull, line 57: nit - let's format the sql a bit better to have everything properly aligned and with only the necessary sapces

SumanMaharana: Formatted the queries.

- I had 2 tab indentations setup on my editor so it was looking fine there :)

OnkarVO7 (approved):

PR #15794, opened 2024-04-03
FIXES #14772: Display Assets without Owner/Description

### LINKED ISSUE #14772: Display assets with no description/owner
**Context**
In the data insight report, we show the percentage coverage of ownership and description. Currently as a user, if I want to see all the assets that do not have description and ownership coverage I am unable to do it easily. 

- As a user, I want to quickly find the assets that do not have ownership and description coverage

**A.C**
- [ ] Add a view to list tables with no description or ownership.

IceS2:
### Describe your changes:

Fixes #14772

In order to display assets without Owner/Description we have the following approach:

From the Data Insights page a user will be able, by the click of a button, to land on the Explore page with Advance Filters already applied (Either to get the assets with incomplete description or without owner).
This filters will be persistent across Asset Types and they will be able to navigate freely.

In order to achieve that:
- We already have the filter available for the Owner attribute
- For description we are creating a new pre-computed attribute `descriptionStatus` that will either resolve to 'INCOMPLETE' or 'COMPLETE'.
- A new Advanced Filter will be added for this attribute
- UI modified accordingly

#
### Type of change:

- [x] Improvement

#
### Checklist:


TeddyCr (approved):

openmetadata-service/src/main/resources/elasticsearch/en/table_index_mapping.json
```diff
@@ -99,6 +99,7 @@
         "normalizer": "lowercase_normalizer"
       },
       "columns": {
+        "type": "nested",
```
harshach, line 102: can we check the column name aggregate

IceS2: Hey @harshach , great catch. With this change it actually breaks the column search.

I was searching for alternatives and if we also add the `include_in_parent = true` attribute it seems to fix the search. https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html

I'm a bit worried that basically this means we're saving this information twice. Once on its own index and the other one as it currently is.

Was also thinking about other alternatives that we could use instead of changing the mapping type:

- [Script Query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-script-query.html): It's said to have a bad performance and I couldn't access the fields that I actually needed to access since they are analyzed text fields. We could technically change that but it [doesn't seem to be a good way](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-fields.html#modules-scripting-doc-vals)

- Having extra meta-metadata fields that we could calculate internally and use it for filtering.

What do you think?

Thanks

TeddyCr: Should we add a test around this use case @IceS2?

IceS2: While talking with @ShaileshParmar11 on this task, he mentioned that we've got a really nice aggregate we use to fetch the filter list and that it also shows the amount of documents with that specific filter.

With the proposed solution I saw that it weasn't really possible to do that since we're doing some hacky wildcard search on text fields and I was trying to change it.

I learnt a bit more about ElasticSearch and thought about adding some runtime fields and we could have something like this:

```    
"runtime":  {
      "descriptionStatus": {
        "type": "keyword",
        "script": {
          "source": "int columnQty = 0; int columnDescriptionQty = 0; if (doc.containsKey('columnNames')) { columnQty = doc['columnNames'].length; } if (params['_source'].containsKey('columns')) { for (column in params['_source']['columns']) { if (column.containsKey('description') && column['description'] != '') { columnDescriptionQty += 1; } }} if (columnDescriptionQty == columnQty && params['_source'].containsKey('description') &&params['_source']['description'] != '') {  emit('COMPLETE')} else {  emit('INCOMPLETE')}"
        }
      },
      "ownerKeywordWithNoOwner": {
        "type": "keyword",
        "script": {
          "source": "if (doc.containsKey('owner.name.keyword') && doc['owner.name.keyword'].size() > 0) { emit(doc['owner.name.keyword'].value) } else { emit('No Owner') }"
        }
      }
},
```

I believe we could define the fields on the mapping itself for each asset instead of relying on a runtime field at query time since each asset might have different conditions.

What do you think?

harshach: @IceS2 any reason to have these as runtime fields , why not pre-compute them in the index docs TableIndex.java etc..

IceS2: Only reason is me not being familiar with the code yet ;P

That's probably the best solution (And fyi, Opensearch does not support runtime mappings at the time being)

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -5,21 +5,7 @@
 import static org.openmetadata.service.Entity.FIELD_DESCRIPTION;
 import static org.openmetadata.service.Entity.FIELD_DISPLAY_NAME;
 import static org.openmetadata.service.Entity.FIELD_NAME;
-import static org.openmetadata.service.search.EntityBuilderConstant.COLUMNS_NAME_KEYWORD;
```
TeddyCr, line 8: you need to set your intellij to not do the import `*` https://stackoverflow.com/questions/3587071/disable-intellij-starred-package-imports

openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/ContainerIndex.java
```diff
@@ -78,6 +78,9 @@ public Map<String, Object> buildESDoc() {
             ? 0
             : container.getVotes().getUpVotes() - container.getVotes().getDownVotes());
     doc.put("domain", getEntityWithDisplayName(container.getDomain()));
+    doc.put(
+        "descriptionStatus",
+        getDescriptionStatus(container.getDescription(), container.getDataModel().getColumns()));
```
TeddyCr, line 83: Should we move this to the parent class to avoid duplicating the code for all EntityInterface object and just call that method from here? On the SearchIndex side we could do something like

```
try {
     Method method = entity.getClass().getMethod("getColumns");
     List<Column> cols = ((List<Column>) method.invoke(entity);
     ...
catch (NoSuchMethodException, ... e) {
    ...
}
```

openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OpenSearchClient.java
```diff
@@ -5,21 +5,7 @@
 import static org.openmetadata.service.Entity.FIELD_DESCRIPTION;
 import static org.openmetadata.service.Entity.FIELD_DISPLAY_NAME;
 import static org.openmetadata.service.Entity.FIELD_NAME;
-import static org.openmetadata.service.search.EntityBuilderConstant.COLUMNS_NAME_KEYWORD;
```
TeddyCr, line 8: same comment regarding intellij

openmetadata-service/src/test/java/org/openmetadata/service/resources/databases/TableResourceTest.java
```diff
@@ -66,29 +66,22 @@
 import static org.openmetadata.service.util.TestUtils.UpdateType.NO_CHANGE;
 import static org.openmetadata.service.util.TestUtils.UpdateType.REVERT;
 
+import es.org.elasticsearch.client.Request;
+import es.org.elasticsearch.client.Response;
+import es.org.elasticsearch.client.RestClient;
 import java.io.IOException;
 import java.text.ParseException;
-import java.util.ArrayList;
```
TeddyCr, line 71: same comment here regarding intellij setting

openmetadata-service/src/test/java/org/openmetadata/service/resources/databases/TableResourceTest.java
```diff
@@ -66,29 +66,22 @@
...
 import lombok.extern.slf4j.Slf4j;
 import org.apache.http.client.HttpResponseException;
-import org.junit.jupiter.api.MethodOrderer;
-import org.junit.jupiter.api.Order;
-import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.TestInfo;
-import org.junit.jupiter.api.TestInstance;
-import org.junit.jupiter.api.TestMethodOrder;
+import org.apache.http.util.EntityUtils;
+import org.junit.jupiter.api.*;
```
TeddyCr, line 82: here as well

TeddyCr (commented):
Thanks for the PR. Left a commennt for discussion 😊

TeddyCr (changes requested):
@IceS2 you'll need to revert the import *

openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/ContainerIndex.java
TeddyCr: cc @mohityadav766 any thought if this makes sense ^

mohityadav766: @IceS2 @TeddyCr yup it does make sense , we can move this along with other attribute that are available across Assets implementing EntityInterface to have default behaviour in base class, only entity specific fields can be updated in here

mohityadav766: @IceS2 we can have buildESDoc as a default methods, and at the end call addEntitySpecificDataToDoc , where we can add other attribute that are specfic to entityType

IceS2: Just pushed some changes that go into that direction!

mohityadav766 (approved):
Looks good @IceS2 , nice work getting things centralised for Indexes :)

TeddyCr (approved):

TeddyCr:
💡 Maven MySQL failure related to ES test flakiness

PR #15814, opened 2024-04-04
Fix  #10354 - Implement Test Suite filter from Search

### LINKED ISSUE #10354: Filter test suites by owner / name
Hi, 
We have many test suites in openmetadata quality section, and would like to filter these test suites by owner. 
We would like to have a drop down menu with owners list (teams list) so that each team can see their tests among the hundreads of test suites that we have. 

Alternatively, a search bar to look for suites with a specific name will also be useful. If you will implement this option, we will add the owner name as a part of the test suite name. 

Thanks a lot in advance!

TeddyCr:
### Describe your changes:
- Implement list from search for test suite entities

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestSuiteRepository.java
```diff
@@ -172,31 +137,36 @@ public TestSummary getTestSummary(UUID testSuiteId) throws IOException {
...
+        JsonObject testCaseResultSummary =
+            searchRepository.aggregate(query, TEST_SUITE, aggregationJson);
+        testSummary = getTestCasesExecutionSummary(testCaseResultSummary);
+      }
+      return testSummary;
+    } catch (IOException e) {
+      LOG.error("Error reading aggregation query", e);
     }
-    return testSummary;
+    return null;
```
harshach, line 169: should this be an empty testSummary instead of null?

harshach (approved):

PR #15849, opened 2024-04-10
FIX #13475 - Add Parameter Validation for Min/Max tests

### LINKED ISSUE #13475: ui/backend: test case validation for min max value
currently, users can add a higher number in max value compared to the min value, need to add validation for the same.
below are some use cases:
Need to add validation at the UI as well as in the Backend
- min value must be lower then the max value
- only min value can be added
- only max value can be added

TeddyCr:
### Describe your changes:

Fixes #13475 
- Add parameter validation for min/max tests.

⚠️ if the parameter value cannot be parsed to a double, we'll skip the validation entirely.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Improvement

### DIFF
--- a/bootstrap/sql/migrations/native/1.4.0/mysql/postDataMigrationSQLScript.sql
+++ b/bootstrap/sql/migrations/native/1.4.0/mysql/postDataMigrationSQLScript.sql
@@ -1,0 +1,364 @@
+-- Start of Test Definition Parameter Definition Validation Migration
+UPDATE test_definition
+set json = JSON_INSERT(
+	JSON_REMOVE(json, '$.parameterDefinition'),
+	'$.parameterDefinition',
+	JSON_ARRAY(
+		JSON_OBJECT(
+		  'name', 'minValue',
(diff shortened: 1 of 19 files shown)


bootstrap/sql/migrations/native/1.4.0/mysql/postDataMigrationSQLScript.sql
```diff
@@ -0,0 +1,364 @@
+-- Start of Test Definition Parameter Definition Validation Migration
+UPDATE test_definition
```
harshach, line 2: since these are test_definitions would it be easier to it via code then the sql?

TeddyCr: I was hesitating with this but I though SQL query would be less prone to error. Happy to move them to java.

harshach (approved):

PR #15878, opened 2024-04-12
MINOR: added a DELETE method for failedRowsSample

sushi30:
https://github.com/open-metadata/openmetadata-collate/issues/388

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
@@ -961,7 +961,7 @@ public class TestCaseResource extends EntityResource<TestCase, TestCaseRepositor
           UUID id,
       @Valid TableData tableData) {
     OperationContext operationContext =
-        new OperationContext(entityType, MetadataOperation.EDIT_SAMPLE_DATA);
+        new OperationContext(entityType, MetadataOperation.EDIT_TEST_CASE_FAILED_ROWS_SAMPLE);
     authorizer.authorize(securityContext, operationContext, getResourceContextById(id));
     TestCase testCase = repository.find(id, Include.NON_DELETED);
     if (testCase.getTestCaseResult() == null
@@ -992,7 +992,7 @@ public class TestCaseResource extends EntityResource<TestCase, TestCaseRepositor
       @Parameter(description = "Id of the table", schema = @Schema(type = "UUID")) @PathParam("id")
           UUID id) {
     OperationContext operationContext =
-        new OperationContext(entityType, MetadataOperation.VIEW_SAMPLE_DATA);
+        new OperationContext(entityType, MetadataOperation.VIEW_TEST_CASE_FAILED_ROWS_SAMPLE);
     ResourceContext<?> resourceContext = getResourceContextById(id);
     TestCase testCase = repository.find(id, Include.NON_DELETED);
     authorizer.authorize(securityContext, operationContext, resourceContext);
--- a/openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
@@ -1000,6 +1000,31 @@ public class TestCaseResource extends EntityResource<TestCase, TestCaseRepositor
     return repository.getSampleData(testCase, authorizePII);
   }
 
+  @DELETE
+  @Path("/{id}/failedRowsSample")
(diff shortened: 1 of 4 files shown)


TeddyCr (approved):

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
```diff
@@ -1000,6 +1000,31 @@ public TableData getFailedRowsData(
...
+            responseCode = "404",
+            description = "Failed rows sample data for test case {id} is not found.")
+      })
+  public Response deleteFailedRowsData(
+      @Context UriInfo uriInfo,
+      @Context SecurityContext securityContext,
+      @Parameter(description = "Id of the table", schema = @Schema(type = "UUID")) @PathParam("id")
+          UUID id) {
+    OperationContext operationContext =
+        new OperationContext(entityType, MetadataOperation.VIEW_SAMPLE_DATA);
```
harshach, line 1021: @sushi30 we should check DELETE_SAMPLE_DATA operation here. or Owner of the test or Owner of the test that this table belongs to

harshach: Also we should introduce a separate permission here.
1. Who should be able to view sample data for a failed test 
  2.  This should only be for Admins, Owner of table, owner of test
3. How we grant sample data permission
   4.  We shouldn't mix view sample data of table to sample failed rows of test case. Both serve different purpose

TeddyCr: > Also we should introduce a separate permission here.

@harshach IMO the principle here is the same in both cases - though use case are different. I can't think of a use case where one would be allowed to view sample data for a table but not the sample data of the failed rows for the same table. 

Do you have any use case in mind, I might be missing something?

sushi30: @harshach while they serve different purposes, the permissions are essentially for the same data (i.e: viewing data from the underlying table). the connection between the data and the test is only from OM perspective.

in terms of governance policy, if a user has certain permissions to view the data in the context of a profiler result, is there a reason they should have different access in the context of a test case result?

sushi30:
@harshach I used the `EDIT_SAMPLE_DATA` as the required permissions for deleting failed sample data LMK if this works.

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
```diff
@@ -1000,6 +1000,31 @@ public TableData getFailedRowsData(
...
+            responseCode = "404",
+            description = "Failed rows sample data for test case {id} is not found.")
+      })
+  public Response deleteFailedRowsData(
+      @Context UriInfo uriInfo,
+      @Context SecurityContext securityContext,
+      @Parameter(description = "Id of the table", schema = @Schema(type = "UUID")) @PathParam("id")
+          UUID id) {
+    OperationContext operationContext =
+        new OperationContext(entityType, MetadataOperation.DELETE_SAMPLE_DATA);
```
harshach, line 1021: @sushi30 my suggestion here is to define a additional MetadataOperation, it can be DELETE_FAILED_ROWS_OF_A_TEST_CASE

sushi30: I used `DELETE_TEST_CASE_FAILED_ROWS_SAMPLE` so its in form of:
`<OP>_<RESOURCE>_<SUB_RESOURCE>`

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
```diff
@@ -961,7 +961,7 @@ public TestCase addFailedRowsData(
           UUID id,
       @Valid TableData tableData) {
     OperationContext operationContext =
-        new OperationContext(entityType, MetadataOperation.EDIT_SAMPLE_DATA);
+        new OperationContext(entityType, MetadataOperation.EDIT_TESTS);
```
TeddyCr, line 964: I think that might be worth an `EDIT_TEST_CASE_FAILED_ROWS_SAMPLE`

harshach (approved):

PR #15966, opened 2024-04-19
#6420 - Allow better bot role binding

### LINKED ISSUE #6420: Ingestion, Profiler and Support Bot (JWT)
We should have:
- Ingestion Bot
- Profiler Bot
- Support Bot -> 1h expiry for users to give us access and debug on their instance

Then, each workflow will have to point to the required bot, and automatically pick up the credentials from SM.

This way we will have different "service accounts" for different applications. It will also allow us to control better and grow the number of bots/applications, e.g., for data insights.

pmbrull:
### Describe your changes:

Fixes #6420

In Collate I'll create the support bot

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
@@ -91,6 +91,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import javax.json.Json;
+import javax.json.JsonObject;
+import javax.json.JsonObjectBuilder;
 import javax.json.JsonPatch;
 import javax.validation.constraints.NotNull;
 import javax.ws.rs.core.Response.Status;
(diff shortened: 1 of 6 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -1439,8 +1439,6 @@ && userHasRelationshipWithAnyBot(original, bot)) {
               create.getAuthenticationMechanism(),
               original.getAuthenticationMechanism());
       user.setRoles(original.getRoles());
-    } else if (bot != null && ProviderType.SYSTEM.equals(bot.getProvider())) {
-      user.setRoles(UserUtil.getRoleForBot(botName));
```
pmbrull, line 1443: I could not figure out a way to get inside this path. In any case, train of thought is that if there is already a user, we won't re-apply the bot roles

openmetadata-service/src/main/resources/json/data/bot/roles/roleBinding.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ingestion-bot": "IngestionBotRole"
```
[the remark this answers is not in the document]
mohityadav766: and if the role is already there , we should create user bindings with that role

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -449,6 +452,26 @@ public static <U> List<U> getEntitiesFromSeedData(String entityType, String path
     return entities;
   }
 
+  /**
+   * Merge the contents from the json files in the path into a single JsonObject
+   */
+  public static JsonObject getBindingsFromSeedData(String path) throws IOException {
+    JsonObjectBuilder jsonObjectBuilder = Json.createObjectBuilder();
```
harshach, line 459: @pmbrull what is the need for this

pmbrull: I'll remove it and use another approach suggested by mohit

mohityadav766 (approved):

PR #15975, opened 2024-04-19
Fix #15611 Parse PowerBI Dax files for lineage

### LINKED ISSUE #15611: Parse PowerBI DAX files
to create datasets out of them

OnkarVO7:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/15611

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/dashboard/powerbi/connection.py
```diff
@@ -22,28 +22,34 @@
...
 
 
 def get_connection(connection: PowerBIConnection) -> PowerBiApiClient:
     """
     Create connection
     """
-    return PowerBiApiClient(connection)
+    client = PowerBiClient(api_client=PowerBiApiClient(connection))
+    if connection.pbitFilesSource:
+        client.file_client = PowerBiFileClient(connection)
```
IceS2, line 38: Doing it this way defeats a little bit the purpose of Pydantic since you'll have no validation whatsoever when assigning it directly.

(I think it's fine on this case but wanted to give you a heads-up

ingestion/src/metadata/ingestion/source/dashboard/powerbi/metadata.py
```diff
@@ -539,6 +552,90 @@ def create_datamodel_report_lineage(
...
+                entity_type=Table,
+                service_name=db_service_name,
+                database_name=None,
+                schema_name=None,
+                table_name=table.name,
+            )
+            table_entity = self.metadata.get_by_name(
+                entity=Table,
+                fqn=table_fqn,
+            )
```
IceS2, line 576: Is there any way to defer this API call to the Sink? We are avoiding having API calls on the Source due to how Multithreading works.

It should be ok if this is ran as a post_process step of the root node but it'd be better if we could just defer the call to the Sink like we did with the OMetaLifecycle

IceS2 (approved):

ingestion/src/metadata/ingestion/source/dashboard/powerbi/connection.py
OnkarVO7: Makes sense. I just wanted to wrap the clients in a class so that they can be accessed wherever easily

IceS2: I think the wrapping is fine to be honest, maybe shouldn't be a pydantic class or you could do something like:
```
file_client = None
if connection.pbitFilesSource:
  file_client = PowerBiFileClient(connection)
# Instantiate pydantic class with both clients (or one None)
```

pmbrull (commented):
I think this change might need some tests. Since we're talking about new readers of static files, adding coverage at least for the file parsing would be worth it

pmbrull (changes requested):
I think this change might need some tests. Since we're talking about new readers of static files, adding coverage at least for the file parsing would be worth it

ingestion/src/metadata/ingestion/source/dashboard/powerbi/connection.py
OnkarVO7: Got it will make the change

OnkarVO7: This is updated

IceS2 (approved):

PR #15985, opened 2024-04-22
[MINOR] Add childrenCount field for glossaryTerm

sonika-shah:
### Describe your changes:

Added the childrenCount field to the GlossaryTerm API to support expand/collapse functionality for terms -  based on their count on glossary term listing page.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/resources/glossary/GlossaryTermResource.java
```diff
@@ -244,6 +244,7 @@ public ResultList<GlossaryTerm> list(
     } else { // Forward paging or first page
       terms = repository.listAfter(uriInfo, fields, filter, limitParam, after);
     }
+    repository.addChildrenCount(terms);
```
mohityadav766, line 247: we should do it in setFields of glossaryTerms I feel, this should not be be handled in the listing api

PR #15988, opened 2024-04-22
issue-15858: reduce es call in ingestion for foreign key processing

### LINKED ISSUE #15858: Review Foreign Key Processing during ingestion.
While ingesting foreign key constraint we do ES lookup to find the referred table's FQN https://github.com/open-metadata/OpenMetadata/blob/b2ce491ff1c2e473222b006a9e1c1915ed537e87/ingestion/src/metadata/ingestion/source/database/common_db_source.py#L539

because we do not get the database name of the FK from sqlalchemy dialect. This may lead into multiple issues, it may produce incorrect versioning in case the ES did not populate, also the FK information only gets reflected after the 2nd run.

To solve this we can modify/patch the SQA dialects, and grab the information about the database as well, for example snowflake stores the fk db name, once we have all the ingredients to generate FQN we will not have to make a call to ES, and FK can reflect from the first run itself.

image

harshsoni2024:
### Describe your changes:

Fixes #15858 

I worked on to reduce es call while ingestion in foreign key processing

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/postgres/utils.py
+++ b/ingestion/src/metadata/ingestion/source/database/postgres/utils.py
@@ -63,6 +63,120 @@ def get_etable_owner(
 
 
 @reflection.cache
+def get_foreign_keys(
+    self, connection, table_name, schema=None, postgresql_ignore_search_path=False, **kw
+):
+    preparer = self.identifier_preparer
+    table_oid = self.get_table_oid(
+        connection, table_name, schema, info_cache=kw.get("info_cache")
+    )
+
+    FK_SQL = """
+        SELECT r.conname,
+            pg_catalog.pg_get_constraintdef(r.oid, true) as condef,
+            n.nspname as conschema,
+            d.datname AS con_db_name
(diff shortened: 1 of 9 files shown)


ingestion/src/metadata/ingestion/source/database/postgres/utils.py
```diff
@@ -62,6 +62,120 @@ def get_etable_owner(
...
+            d.datname AS con_db_name
+        FROM  pg_catalog.pg_constraint r,
+            pg_namespace n,
+            pg_class c
+        JOIN pg_database d ON d.datname = current_database()
+        WHERE r.conrelid = :table AND
+            r.contype = 'f' AND
+            c.oid = confrelid AND
+            n.oid = c.relnamespace
+        ORDER BY 1
```
ulixius9, line 87: lets put this query in quries.py

ulixius9 (approved):

PR #15992, opened 2024-04-22
FIX #15593 - test case exc summary endpoint improvements

### LINKED ISSUE #15593: Backend: Api end point for test case states count (pass/failed/aborted.)
currently, for showing states for all the columns and table we need to fetch all the test cases of that table and column and show the count for pass failed and aborted,
With the help of a new endpoint, we can avoid heavy calls from the UI side and get only the required information.
API should contains the count details of table and column test cases

### LINKED ISSUE #15968: Result of total test case on Data quality tab not correct.
I found about test case not correct when compare between test case vs total test case.
I have try to re-index but result still incorrect.
image

Version:
OS: CentOS Linux release 7.9.2009 (Core)
OpenMetadata version: 1.3.2
OpenMetadata Ingestion package version: ingestion:1.3.2

TeddyCr:
### Describe your changes:

Fixes #15593 & Fixes #15968

- enabled search tests for test cases resource
- move test case execution summary logic to test case index
- added `columnTestCaseSummary` when retrieving test case execution summary for test suite ID

**Implementation details**
The GET `/v1/dataQuality/testSuites/executionSummary` accepts an optional `testSuiteId=...` query parameter. When retrieving a test case execution summary for a specific test suite ID. the summary of column level (represented by entityLink) will be returned.

The general response is of the form as shown below where `columnTestSummary` will be null if the request is sent without a `testSuiteId` and the `success`, `failed`, etc. keys at the root level of the document will represent the aggregate of the values in columnTestSummary + any test case defined at the table level (which will not show in the `columnTestSummary` detail.

**Response example**
```json
{
        "success": 5
        "failed": 3
        "aborted": 1
        "queued": 0
        "total": 9
        "columnTestSummary": [
              {
                      "success": 2
                      "failed": 1
                      "aborted": 1
                      "queued": 0
                       "total": 4
                       "entityLink":  "<#E::table::something::columns::name>"
              },
              ....
        ]

}
```

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Bug fix

Improvement


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -660,9 +660,10 @@ public RestUtil.DeleteResponse<TestCase> deleteTestCaseFromLogicalTestSuite(
     deleteRelationship(testSuiteId, TEST_SUITE, testCaseId, TEST_CASE, Relationship.CONTAINS);
     // remove test case from logical test suite summary and update test suite
     removeTestCaseFromTestSuiteResultSummary(testSuiteId, testCase.getFullyQualifiedName());
-    EntityReference entityReference =
-        Entity.getEntityReferenceById(TEST_SUITE, testSuiteId, Include.ALL);
-    testCase.setTestSuite(entityReference);
+    TestCase updatedTestCase = Entity.getEntity(Entity.TEST_CASE, testCaseId, "*", Include.ALL);
```
harshach, line 663: do we need to all of these on the delete path. Wouldn't this run into an issue if there is permanent deletion?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestSuiteRepository.java
```diff
@@ -238,6 +307,47 @@ public RestUtil.DeleteResponse<TestSuite> deleteLogicalTestSuite(
     return new RestUtil.DeleteResponse<>(updated, changeType);
   }
 
+  private void updateTestSummaryFromBucket(JsonObject bucket, TestSummary testSummary) {
+    String key = bucket.getString("key");
+    Integer count = bucket.getJsonNumber("doc_count").intValue();
+    switch (key) {
+      case "success":
```
harshach, line 314: we must have enum in json schema for this?

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -877,6 +878,21 @@ public static List<AggregationBuilder> buildAggregation(JsonObject aggregations)
             }
             aggregationBuilders.add(nestedAggregationBuilder);
             break;
+          case "aggs":
+            // Sub aggregation logic
+            if (keySet.contains("nested")) {
```
harshach, line 883: can we inverse this condition to remove return from here

openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OpenSearchClient.java
```diff
@@ -896,6 +897,21 @@ public static List<AggregationBuilder> buildAggregation(JsonObject aggregations)
             }
             aggregationBuilders.add(nestedAggregationBuilder);
             break;
+          case "aggs":
+            // Sub aggregation logic
+            if (keySet.contains("nested")) {
```
harshach, line 902: same as above

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
TeddyCr: Here we are not deleting the test case, just removing it from a logical test suite.

TeddyCr: LMK if that makes sense

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestSuiteRepository.java
TeddyCr: From my understanding case value in the switch statement need to be constant known at compile time. The enum value returns a TestCaseStatus type while what we get from ES is a string. Calling `Success.value()` on the case statement throws a private access error.

There is also the lower part the Enum is `Success` while what we get from the bucket is `success` - though setting a nonNormalized field could remove this issue.

LMK if that makes sense or if you have suggestions I am not aware of.

harshach: @TeddyCr I was expecting Status status = Status.valueOf(bucket.getString("key")) 
use that do the switch case

harshach (approved):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestSuiteRepository.java
TeddyCr: Make sense. I tried to not cast it to keep it generic across aggregation

PR #15998, opened 2024-04-23
Minor Remove classifier config

TeddyCr:


pmbrull (approved):

pmbrull:
@TeddyCr approved it, thought it might be worth it to add this blurb in the docs for breaking change and maybe explain if/how we'll get this going in the future

PR #16002, opened 2024-04-23
issue-15607(3-rd try)

RabochiyYan:
Short blurb explaining:
- What changes did you make?
 I rewrote the query that gets stored procedures from ms sql and its definitions.
- Why did you make them?
Because i found out that openmetadata 1.3.0 cant get more than 4000 symbols of stored procedure definition
I have replaced one file in our docker image and ran ingestion. There were not any changes except stored procedure definition being fully loaded.

#
### Type of change:
- [1] Bug fix
#
### Checklist:

- [1] I have added a test that covers the exact scenario we are fixing. For complex issues, comment the issue number in the test for future reference.

At first I thought that I had broken the metadata ingestion because it took quite a long time to complete, but then I realized that this was obviously due to the increased amount of information for parsing

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/mssql/queries.py
+++ b/ingestion/src/metadata/ingestion/source/database/mssql/queries.py
@@ -191,10 +191,12 @@ MSSQL_GET_STORED_PROCEDURES = textwrap.dedent(
     """
 SELECT
   ROUTINE_NAME AS name,
-  NULL AS owner,
+  NULL AS owner,            
   ROUTINE_BODY AS language,
-  ROUTINE_DEFINITION AS definition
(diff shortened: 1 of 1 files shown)


pmbrull (approved):
thanks for the fix

pmbrull:
@RabochiyYan do we need any updated permissions to access those sys tables?

RabochiyYan:
@pmbrull 
I think not, but im not sure.
Your documentation says that user have to be granted VIEW SERVER STATE permission, because OM ingestion need some another views in schema "sys". Updated query accesses the same table. So it depends on if VIEW SERVER STATE give you ability to make select to all views in the schema.

Issue #16001, opened 2024-04-23
User search is not handling special chars


Alexandr-Nedomets:
**Describe the bug**
Incorrect symbols search '+', '&'
Get error for symbols search '[', ']' 

**To Reproduce**
Enter '+', '&', '[', ']' into main search 
or use the user search https://sandbox.open-metadata.org/settings/members/users?user=%5B
![image](https://github.com/open-metadata/OpenMetadata/assets/55582159/ed9cede2-eedf-4ffd-913e-26f90a52daa3)

**Expected behavior**
Correct symbols search '+', '&'
Fix error for symbols search '[', ']' 

**Version:**
 - OpenMetadata version: 1.2.5 and current, maybe in another versions too

Alexandr-Nedomets:
I'm ready to do this issue
if all ok

harshach:
Thanks @Alexandr-Nedomets please provide a PR we will review it.

PR #16025, opened 2024-04-24
fix character escaping for elasticsearch to work correctly

Alexandr-Nedomets:
### Describe your changes:
Fix character escaping for elasticsearch to work correctly
in more detail in [ISSUE-16001](https://github.com/open-metadata/OpenMetadata/issues/16001)
Fixes ISSUE-16001
#
### Type of change:

- [x ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/utils/StringsUtils.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/StringsUtils.ts
@@ -188,7 +188,9 @@ export const customServiceComparator = (a: string, b: string): number => {
 export const replacePlus = (fqn: string) => fqn.replaceAll('+', ' ');
 
 export const ES_RESERVED_CHARACTERS: Record<string, string> = {
-  '+': '\\+',
+  '&': '%26',
+  '#': '%23',
+  '+': '%2B',
   '-': '\\-',
   '=': '\\=',
   '&&': '\\&&',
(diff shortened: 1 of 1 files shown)


harshach (approved):

harshach:
@karanh37 can you review and merge

openmetadata-ui/src/main/resources/ui/src/utils/StringsUtils.ts
```diff
@@ -188,7 +188,9 @@ export const customServiceComparator = (a: string, b: string): number => {
 export const replacePlus = (fqn: string) => fqn.replaceAll('+', ' ');
 
 export const ES_RESERVED_CHARACTERS: Record<string, string> = {
-  '+': '\\+',
```
karanh37, line 191: Hi this is breaking at other places.
The idea behind this method is just to escape special characters.
We should not be encoding characters here.

karanh37: Refer Glossary Assets page.
Glossary terms with special names are not able to fetch because we are encoding here.

karanh37 (changes requested):

openmetadata-ui/src/main/resources/ui/src/utils/StringsUtils.ts
Alexandr-Nedomets: Hi!
The error may occur because when querying the api /api/v1/search/query, the searchValue must be converted using the escapeESReservedCharacters method.
If you are wrong, describe the problem in more detail

openmetadata-ui/src/main/resources/ui/src/utils/StringsUtils.ts
```diff
@@ -188,7 +188,9 @@ export const customServiceComparator = (a: string, b: string): number => {
...
+  '&': '%26',
+  '#': '%23',
```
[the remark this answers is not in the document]
Alexandr-Nedomets: Hi, @chirag-madlani
If you fix it like this, the search will not work correctly.
This fix will not fix the problems @karanh37 wrote about.

PR #16035, opened 2024-04-25
Minor fix test case sorting in search client

TeddyCr:
### Describe your changes:
- fix parsing of null timestamp field

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Bug fix
-->


sushi30 (dismissed):

Sachin-chaurasiya (approved):

chirag-madlani (dismissed):

Sachin-chaurasiya (approved):

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java
```diff
@@ -548,6 +549,14 @@ public void softDeleteOrRestoreEntity(EntityInterface entity, boolean delete) {
   public void deleteOrUpdateChildren(EntityInterface entity, IndexMapping indexMapping) {
     String docId = entity.getId().toString();
     String entityType = entity.getEntityReference().getType();
+    List<String> testRelatedAliases =
+        indexMapping.getChildAliases(clusterAlias).stream()
```
harshach, line 553: any reason to have this filter? it seems getChildAliases not working seamlessly . Can we drive this through config rather than filtering out here

harshach: @TeddyCr lets add relatedAliases in the indexMapping.json
https://github.com/open-metadata/OpenMetadata/blob/b7c54c9ab556ccd03f62ca7b0ac90ff7d26f19c4/openmetadata-service/src/main/resources/elasticsearch/indexMapping.json#L7

if the table is getting deleted the entity that belongs to this table in relatedAliases should get deleted as well. This can bake into delete path in indexing. It will be cleaner instead of making a specific fix for test cases

harshach (approved):

harshach (approved):

PR #16042, opened 2024-04-25
MINOR: ci: retries

sushi30:
retry tests up to 3 times in main branch. this is in order to reduce false positives after pushing to main. handle exists like this:
image

Next step would be to let the `mvn` command return 0 even when tests fail (not on error).

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/.github/workflows/maven-build.yml
+++ b/.github/workflows/maven-build.yml
@@ -101,11 +101,12 @@ jobs:
           sudo make install_antlr_cli
 
       - name: Build with Maven
+        uses: nick-fields/retry@v3.0.0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: mvn clean package -am
         with:
           max_attempts: ${{ github.ref == 'refs/heads/main' && 3 || 1 }}
+          command: mvn clean package -am
 
       - name: Clean Up
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/maven-build.yml
```diff
@@ -104,6 +104,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: mvn clean package -am
+        with:
+          max_attempts: ${{ github.ref == 'refs/heads/main' && 3 || 1 }}
```
akash-jain-10, line 108: This will not work! Please refer to the [github actions workflow syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions)!

Github does not provide out of box way to rerun a step :(  You might want to look for [github actions from marketplace](https://github.com/marketplace?type=actions)!

akash-jain-10 (changes requested):

.github/workflows/maven-build.yml
sushi30: fixed

akash-jain-10: Thanks @sushi30

akash-jain-10 (approved):
LGTM 🚀

Issue #15377, opened 2024-02-27
[GlossaryTerm] Issue in Glossary Term


mohityadav766:
**Affected module**
Does it impact the UI, backend or Ingestion Framework?
Backend

**Describe the bug**
 - Mutually Exclusive should only be allowed during creation of tags
 - Description updates should not invoke the validation flow
 - Aggregation ES result in mutually exclusive tags to appear together
 

**To Reproduce**

Screenshots or steps to reproduce

**Expected behavior**

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8]
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

mohityadav766:
Closing this , this should handled now

PR #16068, opened 2024-04-29
[GlossaryTerm] Fix Glossary Term Issue

mohityadav766:
### Describe your changes:

- Do not update mutuallyExclusive (#15377 first)
- Do not update the tags if original and updated are equal (#15377 second , decription update invokes updateTags flow) 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -699,6 +714,11 @@ protected void updateTags(
         return; // Nothing to update
       }
 
+      // If equal return
+      if (validateIfTagsAreEqual(origTags, updatedTags)) {
```
harshach, line 718: can we add inverse check here and not have returns in between a method

mohityadav766: Good Point , updated @harshach thanks

harshach (approved):

harshach (approved):

Issue #15973, opened 2024-04-19
Ingestion dbt : handle multiple GCP projects / database while ingesting manifest


aballiet:
**Is your feature request related to a problem? Please describe.**

Currently lineage is incomplete if dbt use source tables located in an other GCP project / database.

When using several GCP projects as sources, meaning several serviceName, we end up with error :  `MY_OTHER_GCP_PROJECT.DATASET.TABLE' not found in OpenMetadata Please check if the table exists and is ingested in OpenMetadataAlso name, database, schema of the manifest node matches with the table present in OpenMetadata` whereas I have this table ingested in OMD.

**Describe the solution you'd like**

Ingestion should map the GCP_PROJECT_ID or database_name while creating the lineage instead of solely looking among entities from `serviceName`.

**Describe alternatives you've considered**

No real alternative

**Additional context**

PR #16079, opened 2024-04-29
feat: compute dbt model lineage across databases

aballiet:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/15973

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
```diff
@@ -116,6 +117,16 @@ def __init__(self, config: WorkflowSource, metadata: OpenMetadata):
             else "dbtTags"
         )
 
+        # Create a service map to get dbt model lineage across different databases
+        all_databases = self.metadata.es_search_from_fqn(
+            entity_type=Database,
+            fqn_search_string="*",
+        )
+        self.database_services_map = {
```
OnkarVO7, line 125: This map might fail when you will have databases with same names under different services.

OnkarVO7 (commented):
@aballiet thanks for the PR. Left a comment where the map might not work correctly.

The approach I would think to get here is, where we build the parent_fqn we can search the elasticsearch using `*.db.schema.table` this way we can search through all the dbs for the table from manifest.json
Let me know your thoughts

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
aballiet: True ! I did not know it was allowed in backend

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
```diff
@@ -948,7 +948,7 @@ def add_dbt_test_result(self, dbt_test: dict):
                     test_case_fqn = fqn.build(
                         self.metadata,
                         entity_type=TestCase,
-                        service_name=self.config.serviceName,
+                        service_name="*",
```
OnkarVO7, line 951: Here let's use the `source_elements[0]` instead of `*`

OnkarVO7 (changes requested):

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
aballiet: Ok

aballiet: Should we remove `self.config.serviceName` param as it is not needed anymore

OnkarVO7: Yes if it is not used anywhere we can remove it

aballiet: Done 👍

OnkarVO7 (approved):

PR #16096, opened 2024-04-30
MINOR: S3 Add support for single bucket ingestion

ulixius9:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/storage/s3/connection.py
+++ b/ingestion/src/metadata/ingestion/source/storage/s3/connection.py
@@ -59,8 +59,13 @@ def test_connection(
     of a metadata workflow or during an Automation Workflow
     """
 
+    def test_buckets(connection: S3Connection, client: S3ObjectStoreClient):
+        if connection.bucketName:
+            return partial(client.s3_client.list_objects, Bucket=connection.bucketName)
+        return client.s3_client.list_buckets
+
     test_fn = {
-        "ListBuckets": client.s3_client.list_buckets,
+        "ListBuckets": test_buckets(service_connection, client),
         "GetMetrics": partial(
             client.cloudwatch_client.list_metrics, Namespace="AWS/S3"
         ),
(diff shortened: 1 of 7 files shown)


openmetadata-spec/src/main/resources/json/schema/entity/services/connections/storage/s3Connection.json
```diff
@@ -24,6 +24,12 @@
       "title": "AWS Credentials Configuration",
       "$ref": "../../../../security/credentials/awsCredentials.json"
     },
+    "bucketName": {
```
TeddyCr, line 27: Would it make sense to support `n` number of buckets?

ulixius9: technically that is supported if you leave this field empty and put filters this is what we follow even for databases right 🤔

TeddyCr: I see; what would be the difference with setting a bucketName vs adding a bucket name to the filter?

ulixius9: so when you want to ingest multiple buckets you need `ListBuckets` permission, this pr is to cover the scenario when user do not want to give permission for `ListBuckets` due to some security reasons and only wants to ingest a single bucket.


so when using a filter u need `ListBuckets` permissions, when using `bucketName` you do not need that permission

ingestion/src/metadata/ingestion/source/storage/s3/connection.py
```diff
@@ -59,8 +59,15 @@ def test_connection(
     of a metadata workflow or during an Automation Workflow
     """
 
+    def test_buckets(connection: S3Connection, client: S3ObjectStoreClient):
+        if connection.bucketNames:
+            return partial(
+                client.s3_client.list_objects, Bucket=connection.bucketNames[0]
```
TeddyCr, line 65: does it mean we'll only test the connection for the 1st bucket in the list?

ulixius9: yes

pmbrull (approved):

PR #16196, opened 2024-05-09
feat: use native trino client authentication classes

mgorsk1:
### Describe your changes:

Fixes <issue-number>

Since Trino client used in trino integration [provides classes for different types of authentication (and includes both basic and jwt) classes](https://github.com/trinodb/trino-python-client/blob/master/README.md#authentication-mechanisms) I worked on refactoring trino/connection.py class to use them directly instead of workarounds with adjusting the connection url because it enables us to streamline inclusion of different authentication mechanisms and also enables us to use `Oauth2Authentication` class in a same way, which can be handy for some use cases. It will also make it super easy to extend in the future with other supported authentication mechanisms such as certificate or kerberos authentication.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/common/azureConfig.json
```diff
@@ -1,15 +0,0 @@
...
-    "type": "object",
-    "javaType": "org.openmetadata.schema.services.connections.database.common.AzureConfig",
-    "properties": {
-        "azureConfig": {
-            "title": "Azure Credentials Configuration",
-            "$ref": "../../../../../security/credentials/azureCredentials.json"
-        }
-    },
-    "additionalProperties": false
-}
```
ulixius9, line 15: @mgorsk1 we can not remove this file as this is being referred by other services as well for example mysql: https://github.com/open-metadata/OpenMetadata/blob/fa66b5580654e2187aee31bc5611574ed2949b4c/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/mysqlConnection.json#L55

@ayush-shah maybe can you explain the purpose of creating this new file instead of referring the `security/credentials/azureCredentials.json` directly?

ingestion/src/metadata/ingestion/source/database/trino/connection.py
```diff
@@ -77,14 +76,51 @@ def get_connection_url(connection: TrinoConnection) -> str:
...
+        )
+        connection.connectionArguments.__root__["http_scheme"] = "https"
+
+    elif (
+        connection.authType
+        == noConfigAuthenticationTypes.NoConfigAuthenticationTypes.OAuth2
+    ):
+        connection.connectionArguments.__root__["auth"] = OAuth2Authentication()
+        connection.connectionArguments.__root__["http_scheme"] = "https"
+
```
ulixius9, line 123: @mgorsk1 correct me if I'm wrong, but this method would only work if you are running a trino metadata job on your local machine via CLI mode where you have access to a browser, as `OAuth2Authentication` would print a link or redirect you to a the same link where you can complete the authentication

mgorsk1: that's correct, I added id because that's the scenario we needed. The example scenario is:
- trino configured with oauth2
- users run profiling jobs using OM SDK from their private environments
- profiling jobs takes time so we need a mechanism that will collect access_token but also refresh it

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/common/azureConfig.json
mgorsk1: I can revert, the reason i refactored was AzureCredentials did not show as an option in the ui when selecting authentication mechanism.

harshach:
@ulixius9 whats the status of this PR?

ulixius9 (approved):

Issue #16199, opened 2024-05-09
Error omiting all results on explore and search for any term.


heitorrbarros:
**Affected module**
The bug was impacting the explore menu on UI.

**Describe the bug**
When we try to search for a table or something else the error is raised.

**To Reproduce**

1. Go to explore menu 
![image](https://github.com/open-metadata/OpenMetadata/assets/3456754/9029a300-86c7-4100-b712-90d09a6b55fa)
2. Search for any table, for exemple `orders`
![image](https://github.com/open-metadata/OpenMetadata/assets/3456754/1a04ccc2-9c0a-4cfb-9d30-32957b8530ed)

**Expected behavior**
We expect the results that matches with the search terms.

**Version:**
 - OS: OpenMetadata container, docker.getcollate.io/openmetadata/server:1.3.4
 - OpenMetadata version: 1.3.4

**Additional context**
Stack trace:
```
ERROR [2024-05-09 12:36:10,416] [dw-1845 - GET /api/v1/search/query?q=orders&from=0&size=15&index=table_search_index] o.o.s.r.s.SearchResource - Got exception: [OpenSearchStatusException] / message [OpenSearch exception [type=search_phase_execution_exception, reason=all shards failed]] / related resource location: [org.openmetadata.service.resources.search.SearchResource.search](SearchResource.java:173)
os.org.opensearch.OpenSearchStatusException: OpenSearch exception [type=search_phase_execution_exception, reason=all shards failed]
 at os.org.opensearch.rest.BytesRestResponse.errorFromXContent(BytesRestResponse.java:207)
 at os.org.opensearch.client.RestHighLevelClient.parseEntity(RestHighLevelClient.java:2228)
 at os.org.opensearch.client.RestHighLevelClient.parseResponseException(RestHighLevelClient.java:2205)
 at os.org.opensearch.client.RestHighLevelClient.internalPerformRequest(RestHighLevelClient.java:1924)
 at os.org.opensearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:1877)
 at os.org.opensearch.client.RestHighLevelClient.performRequestAndParseEntity(RestHighLevelClient.java:1845)
 at os.org.opensearch.client.RestHighLevelClient.search(RestHighLevelClient.java:1088)
 at org.openmetadata.service.search.opensearch.OpenSearchClient.search(OpenSearchClient.java:450)
 at org.openmetadata.service.search.SearchRepository.search(SearchRepository.java:643)
[...]

PR #16209, opened 2024-05-09
Fixes 16199: Check the totaVotes values for OpenSearch

heitorrbarros:
### Describe your changes:

Fixes [16199](https://github.com/open-metadata/OpenMetadata/issues/16199)

I worked on the fix for OpenSearch 2.9 because I'm using on my organization environment. The fixes of this bug is important for us to improve the User Experience by avoiding a empty screen when the user search for somenthing.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach:
@heitorrbarros I am not sure if the fix is right, totalVotes doesn't have any size to it, its just a value like this 
``` "totalVotes": 27,```

heitorrbarros:
> @heitorrbarros I am not sure if the fix is right, totalVotes doesn't have any size to it, its just a value like this ` "totalVotes": 27,`

Yeah! I

PR #16210, opened 2024-05-09
- Fix for reindex job not waiting for completion

mohityadav766:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/AppRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/AppRepository.java
@@ -216,6 +216,14 @@ public class AppRepository extends EntityRepository<App> {
     return JsonUtils.readValue(json, AppRunRecord.class);
   }
 
+  public AppRunRecord getLatestAppRunsAfterStartTime(UUID appId, long startTime) {
+    String json = daoCollection.appExtensionTimeSeriesDao().getLatestAppRun(appId, startTime);
+    if (json == null) {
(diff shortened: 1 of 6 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/AppRepository.java
```diff
@@ -216,6 +216,14 @@ public AppRunRecord getLatestAppRuns(UUID appId) {
     return JsonUtils.readValue(json, AppRunRecord.class);
   }
 
+  public AppRunRecord getLatestAppRunsAfterStartTime(UUID appId, long startTime) {
+    String json = daoCollection.appExtensionTimeSeriesDao().getLatestAppRun(appId, startTime);
+    if (json == null) {
+      throw new UnhandledServerException("No Available Application Run Records.");
```
harshach, line 222: can we throw a better error than unhandled server exception
this give 503 error back essentially causing the UI to throw not a clear back to user.
What we want UI to behave here.

harshach: also we introduced ERROR_TYPE so that UI can handle those errors better. Lets use that specific error

mohityadav766: updated @harshach

harshach (approved):

PR #16215, opened 2024-05-10
minor: retry logged in user after token refresh

Sachin-chaurasiya:
### Describe your changes:

- The user initially logged into the application. However, if they return after some time and their token is nearing expiration, attempting to access the `'/users/loggedInUser'` API will trigger a 401 error. To resolve this, we'll attempt to refresh the token. Once the token is successfully refreshed, we'll retry the `'/users/loggedInUser'` API call that previously failed.

#
### Type of change:

- [x] Improvement

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Auth/AuthProviders/AuthProvider.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Auth/AuthProviders/AuthProvider.tsx
@@ -21,6 +21,7 @@ import {
 import { MsalProvider } from '@azure/msal-react';
 import {
   AxiosError,
+  AxiosRequestConfig,
   AxiosRequestHeaders,
   InternalAxiosRequestConfig,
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/src/components/Auth/AuthProviders/AuthProvider.tsx
```diff
@@ -510,6 +519,10 @@ export const AuthProvider = ({
         if (error.response) {
           const { status } = error.response;
           if (status === ClientErrors.UNAUTHORIZED) {
+            // store the failed request for retry after successful silent signIn
+            if (error.config.url === '/users/loggedInUser') {
+              failedLoggedInUserRequest = error.config;
```
chirag-madlani, line 524: Instead storing whole config, can we just store it as flag?

chirag-madlani (approved):

PR #16222, opened 2024-05-10
cypress: AUT cypress failure part 1

ShaileshParmar11:
Fixed cypress failure of AUT

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/cypress/common/Entities/EntityClass.ts
```diff
@@ -101,10 +101,11 @@ export const glossaryDetails2 = {
   tags: [],
   mutuallyExclusive: false,
 };
+const glossaryTermDetails1Name = `CypressBankNumber-${uuid()}`;
 
 export const glossaryTermDetails1 = {
-  name: `CypressBankNumber-${uuid()}`,
-  displayName: 'Cypress BankNumber',
+  name: glossaryTermDetails1Name,
+  displayName: glossaryTermDetails1Name,
```
chirag-madlani, line 108: I have kept it different intentionally, wanted to search using displayName and it should gave correct results.
Can we try to test that functionality here?

openmetadata-ui/src/main/resources/ui/cypress/common/Utils/Tags.ts
```diff
@@ -69,19 +81,25 @@ export const removeTags = (
...
       .scrollIntoView()
       .click();
+    verifyResponseStatusCode('@searchTags', 200);
 
     // Remove all added tags
-    cy.get(`[data-testid="selected-tag-${tag}"] [data-testid="remove-tags"]`)
-      .scrollIntoView()
-      .click();
+    cy.get(
+      `[data-testid="selected-tag-${tag}"] [data-testid="remove-tags"]`
```
chirag-madlani, line 99: I have seen some failure around here, Any reason we are removing scrollIntoView here?

openmetadata-ui/src/main/resources/ui/cypress/e2e/Pages/DataInsightSettings.spec.ts
```diff
@@ -124,9 +124,10 @@ describe(
 
       // Adding a manual wait to allow some time between deploying the pipeline and triggering it
       // eslint-disable-next-line cypress/no-unnecessary-waiting
-      cy.wait(1000);
+      cy.wait(2000);
```
chirag-madlani, line 127: @ShaileshParmar11 can we think something else here instead of Manuel wait?

chirag-madlani (commented):
Thanks @ShaileshParmar11 
Some minor comment if you can check

openmetadata-ui/src/main/resources/ui/cypress/common/Entities/EntityClass.ts
ShaileshParmar11: Thanks for the insights, I will surely have look!

openmetadata-ui/src/main/resources/ui/cypress/common/Utils/Tags.ts
ShaileshParmar11: I believe adding scrollIntoView might be shifting the position of the Update button, which could be the cause of the issue. However, I'm still investigating to pinpoint the root cause, as it's only failing in CI and not locally.

openmetadata-ui/src/main/resources/ui/cypress/e2e/Pages/DataInsightSettings.spec.ts
ShaileshParmar11: After clicking the deploy button, there's a background process to deploy into Airflow, which isn't under our control. The backend handles this process, and the UI doesn't have a way to determine if it's deployed. That's why I'm relying on a manual waiting approach. If you have any ideas, please feel free to share them.

Sachin-chaurasiya (approved):

PR #16226, opened 2024-05-10
MINOR: add debug tool in ingestion Dockerfile

HuanjieGuo:
### Describe your changes:

add vim, PS, telnet in ingestion image.

* vim: change the source code or json file sometimes.
* procps: use it to check the workflow process.
* telnet: check the network.

I worked on enhance the ingestion Dockerfile because I deployed it in my env and found that it lacks a lot of necessary commands.

#
### Type of change:

- [x] Improvement

#
### Checklist:

### DIFF
--- a/ingestion/Dockerfile
+++ b/ingestion/Dockerfile
@@ -37,6 +37,9 @@ RUN apt-get -qq update \
     unzip \
     git \
     wget --no-install-recommends \
+    vim \
+    procps \
+    telnet \
     # Accept MSSQL ODBC License
     && ACCEPT_EULA=Y apt-get install -y msodbcsql18 \
     && rm -rf /var/lib/apt/lists/*


TeddyCr (commented):
Thanks @HuanjieGuo for the PR. The best approach here would be for you to extend the Docker image for your use case. We will not be adding additional dependencies that are not required to run the ingestion.

cc @pmbrull for second opinion.

pmbrull:
> Thanks @HuanjieGuo for the PR. The best approach here would be for you to extend the Docker image for your use case. We will not be adding additional dependencies that are not required to run the ingestion.
> 
> cc @pmbrull for second opinion.

Hi, agree with @TeddyCr. We need to keep the final image as slim as possible.

If you need debugging capabilities for development, we can discuss to have them added to the `Dockerfile.ci` one.

Appreciate you openning the PR, but I am afraid it's not something we can't merge.

Thanks

PR #16264, opened 2024-05-15
Fixes #16263 : Mode Ingestion not ingesting any records bug fix

### LINKED ISSUE #16263: [WITH PR] Mode Ingestion not ingesting any records
**Affected module**
This impacts the ingestion framework, particularly Mode Dashboard Ingestion

**Describe the bug**
I will be creating a PR for this issue, this is due to missing params in the Mode API call.

When Mode Ingestion pipeline is run, zero records will be ingested to OpenMetadata even though the pipeline run is marked as success in Airflow & UI.

**To Reproduce**

1. Set up a Mode ingestion pipeline with any configurations
2. Run the newly created pipeline
3. Zero records will be ingested regardless if any dashboards exist in your Mode instance.
4. The pipeline run will be marked as SUCCESS both in UI & Airflow instance

**Expected behavior**
Records will be ingested normally to Open Metadata.

**Version:**
 - OS: macOS 14.2
 - Python version: 3.10.13
 - OpenMetadata version: 1.3.4
 - OpenMetadata Ingestion package version: openmetadata-ingestion[docker]==1.3.4

**Additional context**

I will be creating a PR on this issue

dixon2678:
### Describe your changes:

Fixes [#16263](https://github.com/open-metadata/OpenMetadata/issues/16263)

I worked on changing the collections API get collection method by adding a filter=custom param because no records will be returned if this param is not included.

[Reference Mode API docs](https://mode.com/developer/api-reference/management/collections/#listCollections) 

#
### Type of change:

- [x] Bug fix
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) -> I am unable to confirm whether the current users that uses existing deprecated Personal Token authentication method will be impacted.

#
### Checklist:


ingestion/src/metadata/ingestion/source/dashboard/mode/client.py
```diff
@@ -75,7 +75,7 @@ def fetch_all_reports(self, workspace_name: str) -> Optional[list]:
             dict
         """
         all_reports = []
-        response_collections = self.client.get(f"/{workspace_name}/{COLLECTIONS}")
+        response_collections = self.client.get(f"/{workspace_name}/{COLLECTIONS}?filter=all")
```
dixon2678, line 78: filter=all param is added to get records from collections API

ulixius9: @dixon2678 this looks good to me does this need any special permissions?

ingestion/src/metadata/ingestion/source/dashboard/mode/client.py
```diff
@@ -75,7 +75,7 @@ def fetch_all_reports(self, workspace_name: str) -> Optional[list]:
             dict
         """
         all_reports = []
-        response_collections = self.client.get(f"/{workspace_name}/{COLLECTIONS}?filter=all")
+        response_collections = self.client.get(f"/{workspace_name}/{COLLECTIONS}?filter=custom")
```
dixon2678, line 78: Hi @ulixius9 I have made this change from `all` to `custom`

To answer your inquiry, all API Workspace tokens in Mode will have admin privileges by default (also only admins are able to create them) and according to the [docs](https://mode.com/help/articles/workspace-api-tokens), it looks like granular permission control are not yet supported.

Personal tokens that support granular permssion control to Mode resources appears to be deprecated according to this [docs](https://mode.com/developer/api-reference/authentication/), and no new personal tokens are able to be created as of now. This will be fully deprecated by August.

Several points when using the new Workspace Token:
- At least in my case here, not specifying the filter parameter returns no collections. I have tried it with both Restricted and Unrestricted (default access) collections.
- Specifying `custom` as a filter returns all collections in the Workspace, except Personal collections -> I think this is the most appropriate approach.
- Specifying 'all' returns all collections including Personal collections.

However, there is no way that I could confirm whether the existing users that are still using the deprecated Personal tokens (with limited access, as opposed to the new Workspace token method that grants default admin access) are able to use the `filter=custom` params without any errors, as I am also unable to create a new Personal token in Mode.

Perhaps should we hold off this PR until the usage of Personal Tokens are fully deprecated in August 2024?  @ulixius9

ingestion/src/metadata/ingestion/source/dashboard/mode/client.py
dixon2678: @ulixius9 I have added my reply on the latest change 🙇‍♂️

ingestion/src/metadata/ingestion/source/dashboard/mode/client.py
ulixius9: Thanks @dixon2678 for this details explanation, IMO we can approach it this way -> we introduce a new field in the service connection form so that user can configure this filter i.e wether they want an `all`, `custom` or no filter at all


this way it is possible for users to ingest or skip personal collections and also possible for them to use the deprecated version if they still decide to use is, what do you think about this approach?

ingestion/src/metadata/ingestion/source/dashboard/mode/client.py
```diff
@@ -67,15 +67,22 @@ def __init__(self, config):
         )
         self.client = REST(client_config)
 
-    def fetch_all_reports(self, workspace_name: str) -> Optional[list]:
+    def fetch_all_reports(self, workspace_name: str, filter: Optional[str] = None) -> Optional[list]:
```
dixon2678, line 70: I have added filter as an argument with the default value of None (i.e. no filter). What do you think? @ulixius9

ayush-shah: Hey @dixon2678 
Adding `all` instead of None as default will bring in all the data
```suggestion
    def fetch_all_reports(self, workspace_name: str, filter: Optional[str] = "all") -> Optional[list]:
```

Issue #16414, opened 2024-05-24
[ingestion] dbt 1.8.0 manifest error due to non enum values


aballiet:
**Affected module**
Does it impact the UI, backend or Ingestion Framework?

DBT ingestion pipeline

**Describe the bug**

`Unexpected exception parsing DBT node due to 'str' object has no attribute 'value'`

**To Reproduce**

Screenshots or steps to reproduce

![image](https://github.com/open-metadata/OpenMetadata/assets/25565600/4b0173a9-d171-4c77-8659-58219f98dfcb)

**Expected behavior**

**Version:**
 - OS: [e.g. iOS] : K8s
 - Python version:
 - OpenMetadata version: [e.g. 0.8] : 1.4.0
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

I am willing to offer a PR

PR #16415, opened 2024-05-24
fix: resource_type is not an enum

aballiet:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/16414

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
@@ -342,7 +342,7 @@ class DbtSource(DbtServiceSource):
                     # If the run_results file is passed then only DBT tests will be processed
                     if (
                         dbt_objects.dbt_run_results
-                        and manifest_node.resource_type.value
+                        and manifest_node.resource_type
                         == SkipResourceTypeEnum.TEST.value
                     ):
                         # Test nodes will be processed further in the topology
@@ -360,7 +360,7 @@ class DbtSource(DbtServiceSource):
                         continue
 
                     # Skip the analysis and test nodes
-                    if manifest_node.resource_type.value in [
+                    if manifest_node.resource_type in [
                         item.value for item in SkipResourceTypeEnum
                     ]:
                         logger.debug(f"Skipping DBT node: {key}.")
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
```diff
@@ -339,10 +339,11 @@ def yield_data_models(
                 )
             for key, manifest_node in manifest_entities.items():
                 try:
+                    resource_type = getattr(manifest_node.resource_type, "value", manifest_node.resource_type)
                     # If the run_results file is passed then only DBT tests will be processed
                     if (
                         dbt_objects.dbt_run_results
-                        and manifest_node.resource_type.value
+                        and resource_type
```
TeddyCr, line 346: What are the cases where `resource_type` would be `None` since we have `resource_type = getattr(manifest_node.resource_type, "value", manifest_node.resource_type)` on ln 342?

TeddyCr:
@aballiet thanks for the PR. Letf a comment. I think you'll also need to run `make py_format_check` to verify why the checkstyle is failing. You can also run `make py_format` to do an initial formatting of the code.

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
OnkarVO7: @TeddyCr the `resource_type` is a required attribute for the dbt nodes so it'll always be there

TeddyCr: We should be ok removing this check then right @OnkarVO7. cc @aballiet

aballiet: Yes, will do it 👍

aballiet: No in fact it's an equality check on `resource_type`

TeddyCr (approved):

PR #15974, opened 2024-04-19
FIXES 15626:  Fix issue with not url model store

IceS2:
### Describe your changes:

On certain cases the MlModelStore is empty because OpenMetadata fails to parse URIs for which Pydantic doesn't have support for [Example issue](https://github.com/pydantic/pydantic/issues/2157).

Given this, we changed the type from `uri` to `String`.

Fixes #15626

#
### Type of change:

- [x] Bug fix

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [x] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

PR #16480, opened 2024-05-30
#15243 - Pydantic V2 & Airflow 2.9

### LINKED ISSUE #15243: Pydantic v2 and Airflow 2.8
**Affected module**

Ingestion framework

**Describe the bug**

Since the launch of Airflow 2.8 they dropped support for Pydantic v1. Because of that, openmetadata ingestion cannot be installed alongside Airflow which is a common scenario considering the installation of the lineage backend.

**To Reproduce**

image

_Airflow 2.8 release notes_

image

_Openmetadata setup.py_

**Expected behavior**

Make openmetadata ingestion to accept both v1 and v2 versions. Pydantic offers backwards compatibility by using the `pydantic.v1` module, so it will be a matter of using the correct syntax for v1 and v2 through the use of if elses on the `__version__` variable. 

For most models there will be no difference, but for the custom type in the pydantic openmetadata custom types module there may be a big change.

**Version:**
 - OS: [e.g. iOS] Ubuntu 22.04
 - Python version: 3.11.7
 - OpenMetadata version: [e.g. 0.8] 1.2.5
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`] 1.2.5

**Additional context**

I know airflow 2.8 is still incompatible with openmetadata ingestion, but it would be important to migrate or to make it compatible with both v1 and v2 in order to follow project and lib changes. Airflow is just one problem, but projects that use openmetadata may get incompatibility issues due to this outdated dep. 

An example of compatiblity with v1 and v2 at the same time is [bentoml](https://github.com/bentoml/BentoML).

pmbrull:
### Describe your changes:

Fixes #15243

This PR does all the necessary changes in the code and dependencies to move to Pydantic V2. This includes bumping Airflow to the latest 2.9.1.

It'd been great to handle Airflow bump in a separate PR, but Airflow 2.7.3 does not support Pydantic V2, so we're doing both changes at once.

One option could've been to stick to `from pydantic.v1 import ...`, but we'd still need to plan the full pydantic v2 migration.

## Backward Incompatible Change

This change will require users who rely on the SDK to work with Pydantic V2 as well.

## Changelog

- ⚠️ Pydantic V2 uses a new Rust regex parser. It does not have the same feature set as the Python one. This pattern `^((?!::).)*$` which we use as a constraint for `EntityName` is not supported. We updated the dynamic class generation to ignore the regex pattern until Rust is OK with it. Until then, the validation will happen directly at the server instead of the client. This should not have any further impact.
- `json(encoder=...)` is deprecated. Instead, we need to create our own `CustomSecretStr` class and manage the secrets encoding directly there. We have updated the behavior of the serializer to run `get_secret_value()` by default when doing a model/json dump. We can no longer control the serialization at dump time, but rather we need to move this control at the type level.
- `json` and `parse_obj` are deprecated and renamed to `model_dump` and `model_dump_json`
- ⚠️ `Optional` fields do not have `None` as a default anymore. They need to be added manually. For the dynamic codegen, adding the `description` on the fields already takes care of passing the `None`.
- We can't add extra fields on `BaseModel.__init__`. We are using the `default_factory` instead for the `Status` class, which is indeed cleaner.
- Bumped requirements to ensure pydantic v2 compat.
- The new `RootModel` renames `__root__` to `root`.
- `DateTime` should have time zero
- ⚠️ v2 complains about types in tests which before were wrongly passing. E.g., fields that are `str` cannot be set as an `int` while expecting the typing to be handled dynamically. This makes us be more strict with typing.
- ⚠️  While we create a model, V2 will know how to convert times to their `RootModel`. For example, we can set `Table(name="ABC")` as well as `Table(name=EntityName("ABC")`. However, when patching information, e.g., `table.name = "ABC"`, the model will not automatically wrap the `EntityName` there and the serialization fails. We need to be very strict on doing `table.name = EntityName("ABC")` or we'll get `AttributeError: 'str' object has no attribute 'root'`.
- `ModelMetaClass` is now internal. We'll type as `Type[BaseModel]` instead for schema parsers when defining the function types.
- ⚠️ `model_json_dump` does not transform strings to unicode when needed, e.g., `"магазин"` to `"\u043c\u0430\u0433\u0430\u0437\u0438\u043d"`. This forces us to define manually the dumping process. ([ref](https://github.com/pydantic/pydantic/issues/8825#issuecomment-1946206415)). The OM server does not understand non-ascii characters.
- With the new Airflow base image, we updated the Dockerfiles and removed some snyk security fixes that were hard to maintain and only related to the base debian image. This should not have any impact other than helping us with maintainability. Still debian has other issues that are not patched. We'll rely on debian base images.
- `AnyUrl` value is now based on `Url` not `str`. Also, the parsed URL always ends with `/`. We need to handle that now and properly convert it to string when passed as an argument or env var.
- We had to update the `GcpExternalAccount` JSON Schema to disable the extra properties, otherwise it was wrongly picked up by pydantic parsing from dicts.
- YAML runs of GCP sources (BigQuery, BigTable), if using credentials path, we don’t need to pass the root key anymore, directly specify: `gcpConfig: credentials.json` cc @ayush-shah 
- We added the `str` as an extra type for `min/max` metrics in `table.json`. This gives us a safe parsing fallback for dates that are not automatically converted into the right datetime. This should be transparent cc @TeddyCr to confirm.
- The values of `parameterValues` in test cases are defined as string. We need to properly pass those as strings (not ints) when defining tests. We'll need to update the docs clarifying this when talking about YAML ingestions.
- The type of the workflow config for `Processor`, `Sink`, etc. is `config: Optional[basic.ComponentConfig] = None`. In V1, the model could eat directly the Processor Profiler class (for example). Now, however, it complains when we define the class config directly during testing (e.g., [here](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/tests/integration/postgres/test_data_quality.py#L58)). While reading the YAML for the execution and inferring the right steps config at runtime works, if we define a workflow class in Python by hand, we need to stick to the `ComponentConfig` until we apply better classes in the processor and we use `oneOf` similar to the sources.

## Further Actions
- `profiler/processor/models.py`: `@validator` is deprecated with `each_item`. We need to update that cc @TeddyCr 
- Update queryParserData.json to be a Timestamp, not a string! IT is used by the UI as a timestamp. cc @OnkarVO7 This needs to be validated on usage workflows to ensure we are properly passing dates, not strings
- The values of `parameterValues` in test cases are defined as string. We need to properly pass those as strings (not ints) when defining tests. We'll need to update the docs clarifying this when talking about YAML ingestions. cc @TeddyCr 
- discuss if we need to define closed classes for processor, sink, bulksink, etc. configurations instead of relying on a generic `ComponentConfig`.
- Properly validate all usage workflows to ensure there's no errors when parsing dates from different sources

## Notes

- This change will update the `sourceHash` of every asset. It should be a transparent change, other than running a PUT instead of a PATCH in the next ingestion.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/clients/domo_client.py
```diff
@@ -103,14 +103,10 @@ def __init__(
...
-        self.config.instanceDomain = (
-            self.config.instanceDomain[:-1]
-            if self.config.instanceDomain.endswith("/")
-            else self.config.instanceDomain
-        )
         HEADERS.update({"X-DOMO-Developer-Token": self.config.accessToken})
         client_config: ClientConfig = ClientConfig(
-            base_url=self.config.instanceDomain,
+            # AnyUrl string ends with / and the domo API does not respond properly if it has 2 // at the end
+            base_url=str(self.config.instanceDomain)[:-1],
```
ulixius9, line 109: IMO the previous code was correct here 🤔 in case the user voluntarily or by mistakenly puts `/` in the end then we are trying to remove that but if user does not put the `/` end we will be modifying an actual url right

for example `http://localhost:8585/` ->  `http://localhost:8585`
but `http://localhost:8585`  -> `http://localhost:858`

we can use this function here
https://github.com/open-metadata/OpenMetadata/blob/e9ec4367ee42ea00c1345505c22d7d2c7247ba9f/ingestion/src/metadata/utils/helpers.py#L354

pmbrull: I added a "test" about this here https://github.com/open-metadata/OpenMetadata/pull/16480/files#diff-ea5a94c37687536bba98a30863374a14b14107ba16069c681e8610eba3c72521R87 (pasting in case the files load slow)

```python
def test_any_url():
    """It always ends with /"""
    assert str(AnyUrl("https://example.com")) == "https://example.com/"
    assert str(AnyUrl("https://example.com/")) == "https://example.com/"
```

ingestion/src/metadata/ingestion/source/database/domodatabase/models.py
```diff
@@ -51,4 +51,4 @@ class OutputDataset(DomoDatabaseBaseModel):
     columns: int
     schemas: Optional[Schema] = Field(alias="schema")
```
IceS2, line 52: Are we missing a default value here?

IceS2 (commented):
Took a look at the files withing `metadata/ingestion/source/database`.
They look good.

Loved that now we have to be timezone aware. It's a good practice due to some corner cases.

Out of curiosity, why did you decide to go with Annotated for the Topologies?

Thanks

pmbrull:
> Took a look at the files withing `metadata/ingestion/source/database`. They look good.
> 
> Loved that now we have to be timezone aware. It's a good practice due to some corner cases.
> 
> Out of curiosity, why did you decide to go with Annotated for the Topologies?
> 
> Thanks

https://polar.sh/frankie567/posts/typing-annotated-the-new-python-cool seems like the correct way to go with pydantic to be more strict with typing

And thanks for finding the missing defaults @IceS2

IceS2:
> > Took a look at the files withing `metadata/ingestion/source/database`. They look good.
> > Loved that now we have to be timezone aware. It's a good practice due to some corner cases.
> > Out of curiosity, why did you decide to go with Annotated for the Topologies?
> > Thanks
> 
> https://polar.sh/frankie567/posts/typing-annotated-the-new-python-cool seems like the correct way to go with pydantic to be more strict with typing
> 
> And thanks for finding the missing defaults @IceS2

From the article I saw that it's added on Python 3.9:

>have missed the new [Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated) construct, which was introduced in Python 3.9

I think we are still supporting Python 3.8, no? I had to add some workarounds for testcontainers [here](https://github.com/open-metadata/OpenMetadata/pull/15974/files#diff-51c0c123b825b1c331fadd475a97dd0783282959ae7822a4b3993fe7dd09ab8dL311) because of it.

Are we going to stop supporting 3.8?

pmbrull:
> > > Took a look at the files withing `metadata/ingestion/source/database`. They look good.
> > > Loved that now we have to be timezone aware. It's a good practice due to some corner cases.
> > > Out of curiosity, why did you decide to go with Annotated for the Topologies?
> > > Thanks
> > 
> > 
> > https://polar.sh/frankie567/posts/typing-annotated-the-new-python-cool seems like the correct way to go with pydantic to be more strict with typing
> > And thanks for finding the missing defaults @IceS2
> 
> From the article I saw that it's added on Python 3.9:
> 
> > have missed the new [Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated) construct, which was introduced in Python 3.9
> 
> I think we are still supporting Python 3.8, no? I had to add some workarounds for testcontainers [here](https://github.com/open-metadata/OpenMetadata/pull/15974/files#diff-51c0c123b825b1c331fadd475a97dd0783282959ae7822a4b3993fe7dd09ab8dL311) because of it.
> 
> Are we going to stop supporting 3.8?

Since we were doing the v2 migration, figured we could get started "modernizing" the codebase. We'll still maintain 3.8 for now, and until EOL. We're importing `Annotation` from `typing_extensions`.

The testcontainers limit in 3.8 I think is related to some internal networking issues, but not to this PR

ingestion/src/metadata/ingestion/models/custom_properties.py
```diff
@@ -54,12 +54,12 @@ class CustomPropertyType(BaseModel):
 
     id: basic.Uuid
     name: basic.EntityName
-    displayName: Optional[str]
-    fullyQualifiedName: Optional[basic.FullyQualifiedEntityName]
+    displayName: Optional[str] = None
+    fullyQualifiedName: Optional[basic.FullyQualifiedEntityName] = None
     description: Optional[basic.Markdown]
```
IceS2, line 59: I think we missed the default value here

ingestion/src/metadata/ingestion/source/dashboard/qlikcloud/models.py
```diff
@@ -20,14 +20,14 @@
 class QlikApp(BaseModel):
     """QlikCloud App model"""
 
-    description: Optional[str]
-    name: Optional[str]
+    description: Optional[str] = None
+    name: Optional[str] = None
     id: str
     app_id: Optional[str] = Field(alias="resourceId", default=None)
```
IceS2, line 26: Nitpick, maybe we could keep it as we have on the other places: `Field(None, alias="resourceId")`

Thought this one was missing for a moment xD

ingestion/src/metadata/ingestion/source/pipeline/spline/models.py
```diff
@@ -47,7 +47,7 @@ class Extra(BaseModel):
 
 class ExecutionPlan(BaseModel):
     id: Optional[str] = Field(..., alias="_id")
```
IceS2, line 49: Missing defualt

ingestion/src/metadata/ingestion/source/dashboard/qlikcloud/models.py
pmbrull: ye not sure why its different

IceS2 (approved):

ingestion/src/metadata/clients/domo_client.py
ulixius9: ohhh nice

DataOps7:
Which release will include these changes?
I'm using the OM ingestion python SDK  (version `1.4.3.3`) and it still uses Pydantic V1.

PR #16484, opened 2024-05-30
Fix  #16229 - Tag and Service filters for test cases

### LINKED ISSUE #16229: Add tier and service filters for test cases
Currently, we do not have a way to filter test cases based on tier and services

What support do we need?
Backend support to add filters in API,
UI support to add filter dropdown in ui

image

TeddyCr:
### Describe your changes:

Fixes #16229 
- Inherit tags from tables. Column test case will inherit from table + column tags
- Add filters to list test cases by tags and service

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Improvement

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
@@ -87,7 +87,7 @@ import org.openmetadata.service.util.ResultList;
 public class TestCaseResource extends EntityResource<TestCase, TestCaseRepository> {
   public static final String COLLECTION_PATH = "/v1/dataQuality/testCases";
 
-  static final String FIELDS = "owner,testSuite,testDefinition,testSuites,incidentId,domain";
+  static final String FIELDS = "owner,testSuite,testDefinition,testSuites,incidentId,domain,tags";
   static final String SEARCH_FIELDS_EXCLUDE =
       "testPlatforms,table,database,databaseSchema,service,testSuite";
 
@@ -352,6 +352,11 @@ public class TestCaseResource extends EntityResource<TestCase, TestCaseRepositor
           @QueryParam("sortType")
           @DefaultValue("desc")
           String sortType,
+      @Parameter(
+              description = "Return only required fields in the response",
+              schema = @Schema(type = "string"))
+          @QueryParam("includeFields")
+          String includeFields,
       @Parameter(description = "domain filter to use in list", schema = @Schema(type = "string"))
           @QueryParam("domain")
           String domain,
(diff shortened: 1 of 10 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
```diff
@@ -378,7 +389,10 @@ public ResultList<TestCase> listFromSearch(
     searchListFilter.addQueryParam("testPlatforms", testPlatforms);
     searchListFilter.addQueryParam("q", q);
     searchListFilter.addQueryParam("excludeFields", SEARCH_FIELDS_EXCLUDE);
+    searchListFilter.addQueryParam("includeFields", includeFields);
     searchListFilter.addQueryParam("domain", domain);
+    searchListFilter.addQueryParam("tags", tags);
```
harshach, line 394: do we want to expose tags field vs tier. I do see tags might be helpful, but tier should be separate filter.

openmetadata-service/src/main/resources/elasticsearch/en/test_case_index_mapping.json
```diff
@@ -406,6 +406,27 @@
           }
         }
       },
+      "tags": {
```
harshach, line 409: we are mapping the tags from the parent table right? I think for the starters we should just use Tier. Make it easy to maintain the propagation here. Tier changes are less frequent

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
TeddyCr: We can, it will be the same but can be better semantically for users?

openmetadata-service/src/main/resources/elasticsearch/en/test_case_index_mapping.json
TeddyCr: Yep, we'll have that as an inherited field

>  Make it easy to maintain the propagation here
what do you mean by this? We added this test [here](https://github.com/open-metadata/OpenMetadata/blob/ae7eb4944cf98dddcc27717af51291f413ccee93/openmetadata-service/src/test/java/org/openmetadata/service/resources/dqtests/TestCaseResourceTest.java#L853) to make sure the updates are propagated correctly.

harshach: usually column tags can be a quite a bit so was thinking out loud. Another use-case could be propagate glossary terms as well. Users are asking for managing tests via glossary terms, where this may be helpful. If we are propagating tags field it should do the glossary terms as part of it. So this is good to have

TeddyCr: I'll open a different PR for Glossary terms if that is OK so we don't block the UI to ship that in 1.4.2 if that is ok.

harshach (approved):

PR #16486, opened 2024-05-30
Fix #16370 - Added view/DDL definition rule

### LINKED ISSUE #16370: Adding view/DDL definition rule.
Hi OMD Team! 
We at the organization believe that view definition code is sensitive information. We tried to deny its viewing at the privilege level, but this only works with the Deny ViewAll rule for Table.

I see the following solution:
Add a rule to view the view definition and DDL definition in the access settings.

Alternatively, you can add a switch to Metadata Ingestion. Like a "Include DDL" option.

SumanMaharana:
### Describe your changes:

Fixes #16370 

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
@@ -615,27 +615,26 @@ class BigquerySource(
         Get the DDL statement or View Definition for a table
         """
         try:
-            if self.source_config.includeDDL:
-                if table_type == TableType.View:
-                    view_definition = inspector.get_view_definition(
-                        fqn._build(self.context.get().database, schema_name, table_name)
-                    )
-                    view_definition = (
-                        f"CREATE VIEW {schema_name}.{table_name} AS {str(view_definition)}"
-                        if view_definition is not None
-                        else None
-                    )
-                    return view_definition
-
-                schema_definition = inspector.get_table_ddl(
-                    self.connection, table_name, schema_name
+            if table_type == TableType.View:
+                view_definition = inspector.get_view_definition(
+                    fqn._build(self.context.get().database, schema_name, table_name)
                 )
-                schema_definition = (
-                    str(schema_definition).strip()
-                    if schema_definition is not None
+                view_definition = (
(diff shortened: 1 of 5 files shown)


ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
```diff
@@ -615,26 +615,27 @@ def get_schema_definition(
         Get the DDL statement or View Definition for a table
         """
         try:
-            if table_type == TableType.View:
-                view_definition = inspector.get_view_definition(
-                    fqn._build(self.context.get().database, schema_name, table_name)
+            if self.source_config.includeDDL:
```
TeddyCr, line 618: We should do that check where we call `get_schema_definition` so `get_schema_definition` actually get only the schema definition

ingestion/src/metadata/ingestion/source/database/clickhouse/metadata.py
```diff
@@ -153,19 +153,20 @@ def get_schema_definition(
         Get the DDL statement or View Definition for a table
         """
         try:
-            if table_type in {TableType.View, TableType.MaterializedView}:
-                definition_fn = inspector.get_view_definition
-                schema_definition = definition_fn(table_name, schema_name)
-            else:
-                schema_definition = inspector.get_table_ddl(
-                    self.connection, table_name, schema_name
+            if self.source_config.includeDDL:
```
TeddyCr, line 156: same question here regarding the check

ingestion/src/metadata/ingestion/source/database/clickhouse/metadata.py
```diff
@@ -153,19 +153,20 @@ def get_schema_definition(
...
+                if table_type in {TableType.View, TableType.MaterializedView}:
```
TeddyCr, line 157: let's keep the code style constant if we do `if;return;return` let's do this everywhere and not switch between `if;return;return` and `if;else;return`

SumanMaharana: will change the code style into `if;return;return`

TeddyCr (approved):

PR #16504, opened 2024-06-01
playwright: test setup and improvement

ShaileshParmar11:
problem statement:
current entity spec is running with a single browser object and not fully utilizing the power of the playwright
image

Because entity spec uses a single browser object, if one test block fails, it will attempt to re-run that test. However, due to the single browser object, it ends up re-running all the tests instead of just the failed one.
image

Resolution:
This PR takes full advantage of Playwright by giving each test block its own page object. Now, if a test block fails, only that test block will be re-run.

Additional improvement:
- setup the role-based authentication, refer to docs more details -> https://playwright.dev/docs/auth#multiple-signed-in-roles

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/package.json
```diff
@@ -53,6 +53,7 @@
     "@azure/msal-react": "^2.0.12",
     "@dagrejs/dagre": "^1.1.2",
     "@deuex-solutions/react-tour": "^1.2.6",
+    "@estruyf/github-actions-reporter": "^1.7.0",
```
chirag-madlani, line 56: Can we have this as devDeps

openmetadata-ui/src/main/resources/ui/playwright.config.ts
```diff
@@ -32,11 +32,18 @@ export default defineConfig({
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 2 : undefined,
```
chirag-madlani, line 35: What should be the ideal value here?

ShaileshParmar11: For CI, the number of workers can vary between 2 and 4. Currently, I have started testing with 2 workers. It's important to ensure that the number of workers does not exceed the number of CPU cores available on the CI machine. for now, 2 workers are running smoothly in CI.

chirag-madlani (approved):

PR #16576, opened 2024-06-08
FIX - e2e test for pydantic v2

pmbrull:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/ometa/client.py
+++ b/ingestion/src/metadata/ingestion/ometa/client.py
@@ -142,7 +142,7 @@ class REST:
             headers = {"Content-type": "application/json"}
         base_url = base_url or self._base_url
         version = api_version if api_version else self._api_version
-        url: URL = URL(base_url + version + path)
+        url: URL = URL(base_url + "/" + version + path)
         if (
             self.config.expires_in
             and datetime.datetime.utcnow().timestamp() >= self.config.expires_in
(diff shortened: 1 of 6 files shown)


ingestion/src/metadata/ingestion/ometa/client.py
```diff
@@ -142,7 +142,7 @@ def _request(
             headers = {"Content-type": "application/json"}
         base_url = base_url or self._base_url
         version = api_version if api_version else self._api_version
-        url: URL = URL(base_url + "/" + version + path)
+        url: URL = URL(base_url + version + path)
```
pmbrull, line 145: trying to fix metabase with this, since we'd be adding double `//` from `AnyUrl` strings. Not sure if it's the safest way

pmbrull: UPDATE: we'll need to fix this directly in the metabase client

ingestion/src/metadata/ingestion/source/dashboard/metabase/client.py
```diff
@@ -77,7 +77,7 @@ def __init__(
         self.config = config
         session_token = self._get_metabase_session()
         client_config: ClientConfig = ClientConfig(
-            base_url=str(self.config.hostPort),
+            base_url=str(self.config.hostPort)[:-1],  # remove trailing slash
```
IceS2, line 80: Small nitpick (Probably for future mini refactor):

We could have a function called `remove_trailing_slash` or something like that since we are doing it in many places and it'd be self describing vs commenting everywhere

IceS2 (approved):

ingestion/src/metadata/ingestion/source/dashboard/metabase/client.py
pmbrull: ye agree

PR #16608, opened 2024-06-11
Refactor OpenLineage Integration to handle additive OpenLineage events and add official OpenLineage API endpoint

dechoma:
This pull request refactors the OpenLineage integration.
Co-authored with @mgorsk1 

The main objectives are to enable support for additive OpenLineage events (by introducing state management), add an official endpoint for the OpenLineage client, and introduce a new type of service connection that allows consuming OpenLineage events posted to the official endpoint

#### Additive OpenLineage Event Handling: 
Redesigned the connector to handle additive OpenLineage events by introducing state management

#### New Endpoint:
Added an official OpenLineage-compatible endpoint, **/api/v1/openlineage/lineage**, to receive lineage events from the official OpenLineage client. This enables official OpenLineage agents (e.g., from Trino, Spark, Flink) to deliver lineage information directly to OpenMetadata

#### New OpenLineage service connection type: 
Introduced a new type of OpenLineage connection that processes lineage information delivered by the **/api/v1/openlineage/lineage** endpoint.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/bootstrap/sql/migrations/native/1.5.0/mysql/schemaChanges.sql
+++ b/bootstrap/sql/migrations/native/1.5.0/mysql/schemaChanges.sql
@@ -1,0 +1,28 @@
+UPDATE pipeline_service_entity
+SET json = JSON_SET(json, '$.connection.config.openlineageConnection', JSON_EXTRACT(json, '$.connection.config'))
+WHERE servicetype = 'OpenLineage';
+
+UPDATE pipeline_service_entity
+SET json = JSON_REMOVE(json,
+                       '$.connection.config.topicName',
+                       '$.connection.config.brokersUrl',
+                       '$.connection.config.poolTimeout',
+                       '$.connection.config.SSLCALocation',
+                       '$.connection.config.SSLKeyLocation',
+                       '$.connection.config.sessionTimeout',
+                       '$.connection.config.consumerOffsets',
+                       '$.connection.config.securityProtocol',
+                       '$.connection.config.consumerGroupName',
+                       '$.connection.config.SSLCertificateLocation',
+                       '$.connection.config.openlineageConnection.cleanFinalizedRuns')
+WHERE servicetype = 'OpenLineage';
+
+
+CREATE TABLE `openlineage_events` (
+  `id` varchar(36) GENERATED ALWAYS AS (json_unquote(json_extract(`json`,_utf8mb4'$.id'))) VIRTUAL NOT NULL,
+  `eventtype` varchar(36) GENERATED ALWAYS AS (json_unquote(json_extract(`json`,_utf8mb4'$.eventtype'))) VIRTUAL NOT NULL,
+  `runid` varchar(36) GENERATED ALWAYS AS (json_unquote(json_extract(`json`,_utf8mb4'$.runid'))) VIRTUAL NOT NULL,
+  `recieved_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `processed_at` timestamp ,
+  `json` json NOT NULL
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
--- a/bootstrap/sql/migrations/native/1.5.0/postgres/schemaChanges.sql
+++ b/bootstrap/sql/migrations/native/1.5.0/postgres/schemaChanges.sql
@@ -1,0 +1,29 @@
+-- Update openlineageConnection field
+UPDATE pipeline_service_entity
+SET json = jsonb_set(json, '{connection,config,"openlineageConnection"}', (json->'connection'->'config')::jsonb, true)
+WHERE servicetype = 'OpenLineage';
+
+-- Remove old config entries
+UPDATE pipeline_service_entity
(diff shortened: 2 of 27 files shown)


harshach:
@dechoma Thanks for the PR. 
did a cursory look on the PR.  Few things
1. Instead of adding OpenLineage end point, this could be modeled as a sub endpoint under /api/v1/lineage
2. This end point can take the payload of OpenLineage and convert that into Lineage request of OpenMetadata
3. OpenMetadata Lineage specification considers much larger picture and we do not want to create whole another API just for OpenLineage which will confuse the users and also clearly causes confusion between users of OpenMetadata 

Can you please let us know why we need another API endpoint instead of using the OpenLineage payload to convert to OM Lineage request.

mgorsk1:
The reason for specific endpoint is OpenLineage events are accumulative, meaning that single payload won't contain complete lineage information. Consider spark application with openlineage integration that emits 3 events for one ETL job:

- event 1 (type: START) runid=1, inputs=[table a, table b], outputs=[]
- event 2 (type: RUNNING) runid=1, inputs=[], outputs=[]
- event 3 (type: COMPLETE) runid=1, inputs=[], outputs=[table c]

so you have same run (runid=1) with 3 separate events emitted. This is a real scenario in OpenLineage, which by design doesn't have to maintain state across events (particularly tricky in spark). To transform this into meaningful lineage information in OpenMetadata we need to merge these 3 events (once we receive complete event). Otherwise it's not possible to connect column lineage information (table-level would be possible having pipeline object in the middle).

dechoma:
Hi, @harshach could you let me know if @mgorsk1 's explanation is sufficient? The main reason behind the new API was to manage the state in an organized way as @mgorsk1  mentioned

pmbrull:
> Hi, @harshach could you let me know if @mgorsk1's explanation is sufficient? Due to fact that OpenLineage events are generated in additive way(described in comment above) we decided to introduce the state, and the main reason behind the new API was to manage that state in a organized way.

hi @dechoma @mgorsk1, thanks for the explanation.

I am not sure if adding connector-specific endpoints to the API is the way to go. The goal of OM is to bring a generic API layer and schemas that we can then use to ingest any source.

Here we are talking that the issue is not the current API structure or payload definition, but rather how OpenLineage emits the events. This looks to me too specific about OpenLineage to solve it on the server rather than trying to keep the logic on the connector itself.

Maybe it's a too naive approach, but could we dump 24h of events in a file, iterate over ir and aggregate the lineage info for each runId, and then clean the file? Since we're talking about ETL lineage - processes that most likely will be repeated over time - the results should eventually cover all the right lineage, even if in the first run, the event dump is missing a few events for a few runIds.

I am not saying we need to follow this approach, just wondering if you considered any other alternative without having to rely on the server itself?

Thanks

mgorsk1:
Hey @pmbrull, thanks for replying. Indeed adding new API endpoints was not our first choice, but we have not found other way to interact with OpenMetadata storage (database or search engine) and this is where I think OpenLineage event state should be managed (this or Elasticsearch index). Naively dumping events to files near OM is not a sound solution, especially with a lot of deployments relying on stateless kubernetes objects, so it would pose a threat of inconsistency. If we can find a way to maintain state natively in OpenMetadata without having a need for new API endpoints then I am fine with it.

pmbrull:
We'll keep this PR on hold while we discuss possible options. Some notes from our conversation:

- Option A: handle lineage aggregation in the server, lineage reconciliation 
    - This would improve our Spark Lineage Agent as well
- Option B: naive approach -> aggregate state in the connector. That could have partial info, we cannot reprocess, we might miss events, what happens if it fails
    - E.g., Spark streaming may never have "complete" event, just running
    - that would work only for batch applications
    - when is the good moment to send the lineage information?? (e.g., in Flink)

pmbrull:
Hi thanks both for sharing the thoughts here. We have integrated these "half" events handling directly in the Lineage Agent, so there's no need to rely on any changes in the server to get this working.

Please @mgorsk1 feel free to take it for a spin. @ulixius9 please do add any pending details here or configs to get this set.

For now, I'll close this PR

Thank you

mgorsk1:
Can you point me towards the code you have in mind @pmbrull ? Thanks!

ulixius9:
we are yet to publish the code for this jar but we have built on top of the openlineage code and made sure the half lineages gets captured...

you can test out this jar and let us know your feedback

[openlineage-spark_2.12-1.22.0.jar.zip](https://github.com/user-attachments/files/19435090/openlineage-spark_2.12-1.22.0.jar.zip)

it works in the same way as mentioned in this doc: https://docs.open-metadata.org/latest/connectors/ingestion/lineage/spark-lineage#4.-configure-spark

but just instead of using `spark.openmetadata.transport` prefix you use `spark.openlineage.transport`

mgorsk1:
yeah we are already using this connector and had a moment where we introduced state management within but community was not happy about it :) looking forward to OL PR so we can test it with official release.

mgorsk1:
hey @ulixius9 any word on upstreaming the changes to OL connector you mentioned?

PR #16664, opened 2024-06-14
Fixes #16663: Fix the UserResource.patch's if condition (PATCH `v1/users/{id}`)

### LINKED ISSUE #16663: Security Vulnerability in UserRoles Patch API
**Affected module**
openmetadata-service, UI (RBAC)

**Describe the bug**
When a packet is intercepted and modified to activate `isAdmin=True` on Web Console(UI), any user without proper authorization can access the admin page through the Web Console(UI).

In this case, other APIs perform an additional authorization check at the backend level, providing protection from security threats. However, it has been identified that in the code that modifies Roles, there is a bug that allows an attacker to alter permissions.

**To Reproduce**
- [Edge point: UserResource.patch method](https://github.com/open-metadata/OpenMetadata/blob/1.4.0-release/openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java#L804-L847)
  - The authorization check is written as `path.equals("/roles")`.
```java
if (path.equals("/isAdmin") || path.equals("/isBot") || path.equals("/roles")) {
  authorizer.authorizeAdmin(securityContext);
  continue;
}
```

- However, the actual payload of the packet comes in as shown below, so it does not enter the corresponding statement.
```json
{"op": "remove", "path": "/roles/0/(some thing)"}
```

**Expected behavior**
By changing the statement `path.equals("/roles")` to `path.contains("/roles")`, the authorization change API will also be able to perform authorization verification at the backend.

**Version:**
 - OS: -
 - Python version: -
 - OpenMetadata version: 1.4.0-release (all versions)
 - OpenMetadata Ingestion package version: -

**Additional context**
We used the tools from the site below to test by modifying the packet.
- https://portswigger.net/burp

oringnam:
### Describe your changes:
Fixes #16663

Fix the if condition `path.equals("/roles")` to `path.contains("/roles")`
This is because the actual payload comes in as `{"op": "remove", "path": "/roles/0/(some thing)"}`, and the `equals` statement does not perform authorization verification for the roles modification API.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
@@ -832,7 +832,7 @@ public class UserResource extends EntityResource<User, UserRepository> {
       JsonObject patchOpObject = patchOp.asJsonObject();
       if (patchOpObject.containsKey("path") && patchOpObject.containsKey("value")) {
         String path = patchOpObject.getString("path");
-        if (path.equals("/isAdmin") || path.equals("/isBot") || path.equals("/roles")) {
+        if (path.equals("/isAdmin") || path.equals("/isBot") || path.contains("/roles")) {
           authorizer.authorizeAdmin(securityContext);
           continue;
         }


harshach:
@oringnam I have concern with checking contains as the match for roles. Can you describe how this is an issue with more details

harshach (approved):

oringnam:
@harshach 
There seems to be an issue with the Update User API (`PATCH /v1/users/{id}`). From what I've observed, the USERS API uses an Authorizer (I'll refer to the `DefaultAuthorizer` for this explanation) to double-check if the calling user truly has the necessary permissions with DB.

While this verification works seamlessly with the other `/v1/users` APIs, there appears to be a problem with the roles validation in the PATCH `/v1/users/{id}` endpoint. I suspect this may be due to an error in code auto-generation or during a copy-paste process.

### About the code
- [the origin code](https://github.com/open-metadata/OpenMetadata/blob/1.4.0-release/openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java#L804-L847)
```java
@PATCH
  @Path("/{id}")
  @Consumes(MediaType.APPLICATION_JSON_PATCH_JSON)
  @Operation(
      operationId = "patchUser",
      summary = "Update a user",
      description = "Update an existing user using JsonPatch.",
      externalDocs =
          @ExternalDocumentation(
              description = "JsonPatch RFC",
              url = "https://tools.ietf.org/html/rfc6902"))
  public Response patch(
      @Context UriInfo uriInfo,
      @Parameter(description = "Id of the user", schema = @Schema(type = "UUID")) @PathParam("id")
          UUID id,
      @RequestBody(
              description = "JsonPatch with array of operations",
              content =
                  @Content(
                      mediaType = MediaType.APPLICATION_JSON_PATCH_JSON,
                      examples = {
                        @ExampleObject("[{op:remove, path:/a},{op:add, path: /b, value: val}]")
                      }))
          JsonPatch patch) {
    for (JsonValue patchOp : patch.toJsonArray()) {
      JsonObject patchOpObject = patchOp.asJsonObject();
      if (patchOpObject.containsKey("path") && patchOpObject.containsKey("value")) {
        String path = patchOpObject.getString("path");
        if (path.equals("/isAdmin") || path.equals("/isBot") || path.equals("/roles")) {
          authorizer.authorizeAdmin(securityContext);
          continue;
        }
        // if path contains team, check if team is join able by any user
        if (patchOpObject.containsKey("op")
            && patchOpObject.getString("op").equals("add")
            && path.startsWith("/teams/")) {
          JsonObject value = null;
          try {
            value = patchOpObject.getJsonObject("value");
          } catch (Exception ex) {
            // ignore exception if value is not an object
          }
          if (value != null) {
            String teamId = value.getString("id");
            repository.validateTeamAddition(id, UUID.fromString(teamId));
            if (!repository.isTeamJoinable(teamId)) {
              authorizer.authorizeAdmin(securityContext); // Only admin can join closed teams
            }
          }
        }
      }
    }
    return patchInternal(uriInfo, securityContext, id, patch);
  }
```

The core issue lies in the security validation for updates involving `isAdmin`, `isBot`, and `roles` fields. Specifically, the condition:
```java
if (path.equals("/isAdmin") || path.equals("/isDoorBot") || path.equals("/roles")) {
    authorizer.authorizeAdmin(securityContext);
    continue;
}
```
is intended to ensure that only administrators can modify these sensitive attributes. However, the current implementation uses equals for isAdmin and isBot, which works correctly, but for roles, it only checks exact matches and does not account for nested paths such as /roles/0/(some thing) or /roles/1/(some thing) in the payload:
```json
{"op": "remove", "path": "/roles/0/(some thing)"}
{"op": "add", "path": "/roles/1/(some thing)"}
```
Because the path validation does not match these nested paths, the authorizer.authorizeAdmin(securityContext); check is skipped. This can allow unauthorized changes to user roles, as packets that should need admin privileges bypass the check.

### Potential Fixes
I've modified the path validation from `equals to contains` due to my limited understanding of the full range of potential PATCH USERS payloads. However, based on the payloads that have been observed so far, changing from `equals to startsWith` seems to be sufficient and unlikely to cause significant issues.

However, it would be best for someone with a deeper understanding of the OMD API specifications to review this change. I believe you are better equipped to judge whether this solution is adequate and ensures security and functionality.

PR #16695, opened 2024-06-18
[Under Development] - Slack App Integration

Siddhanttimeline:
Creating a distributed Slack app using Bolt-Servlet.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


conf/openmetadata.yaml
```diff
@@ -138,6 +138,18 @@ logging:
       timeZone: UTC
       maxFileSize: 50MB
 
+
+applications:
+  appsPrivateConfiguration:
+    - name: SlackApplication
+      preview: ${SLACKAPPLICATION_PREVIEW:-false}
+      parameters:
+        clientId: ${SLACK_APP_CLIENT_ID:-"7199163258083.7289624672561"}
```
mohityadav766, line 147: let's remove these and put into slackApp.tml , without the secrets

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/slack/SlackApp.java
```diff
@@ -0,0 +1,287 @@
...
+@Slf4j
+public class SlackApp extends AbstractNativeApplication {
+  private App slackAppInstance;
+  private SlackAppConfiguration appConfig;
+  private SystemRepository systemRepository;
+
+  private static final SecureRandom SECURE_RANDOM = new SecureRandom();
+  private static final Base64.Encoder BASE64_ENCODER = Base64.getUrlEncoder();
+
+  private static final String SLACK_OAUTH_BASE_URL = "https://slack.com/oauth/v2/authorize";
```
mohityadav766, line 44: can we use slack sdk instead of hardcoding urls

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/slack/SlackApp.java
```diff
@@ -0,0 +1,287 @@
...
+        LOG.error("Error executing Slack token exchange request: {}", e.getMessage());
+        return false;
+      }
+    } catch (Exception e) {
+      LOG.error("Error exchanging Slack token: {}", e.getMessage());
+      return false;
+    }
+  }
+
+  private RequestBody buildTokenExchangeRequestBody(String code) {
```
mohityadav766, line 152: we don't need to these manually bolt-idk can help us here

conf/openmetadata.yaml
mohityadav766: also can you confirm if the secret is needed, or this can be done on runtime on initiating the flow

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/slack/SlackApp.java
Siddhanttimeline: When using bolt-servlet, it needs to live with the server. The callback web servlet has to be loaded when the server starts with the initialized SlackApp. Otherwise, if we start the SlackApp, those servlets need to be dynamically loaded on the server.

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/slack/SlackApp.java
Siddhanttimeline: Noted.

conf/openmetadata.yaml
Siddhanttimeline: Checking.

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/slack/SlackApp.java
Siddhanttimeline: Updated.

conf/openmetadata.yaml
Siddhanttimeline: ClientSecret is required as they are used to identify and authenticate the application when interacting with Slack's API.

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/slack/SlackApp.java
Siddhanttimeline: Removed buildTokenExchangeRequestBody method and similar, now using slack-api-methods for the token exchange.

Siddhanttimeline:
Final draft completed.

PR #16756, opened 2024-06-21
MINOR: raise lineage error when table does not exist

sushi30:
### Describe your changes:

When lineage ingestion fails due to missing tables, it is not captured by the workflow. This change changes that to capture missing table errors and report them to the user. This is important because it might result in:
1. user error of choosing the wrong query log or service.
2. error with the query parser.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
+++ b/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
@@ -282,7 +282,7 @@ def _build_table_lineage(
     yield Either(right=lineage)
 
 
-# pylint: disable=too-many-arguments
+# pylint: disable=too-many-arguments,too-many-locals
 def _create_lineage_by_table_name(
     metadata: OpenMetadata,
     from_table: str,
--- a/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
+++ b/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
@@ -11,6 +11,7 @@
 """
 Helper functions to handle SQL lineage operations
 """
+import itertools
 import traceback
 from typing import Any, Iterable, List, Optional, Tuple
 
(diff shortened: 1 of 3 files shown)


ingestion/src/metadata/ingestion/lineage/sql_lineage.py
```diff
@@ -315,17 +315,25 @@ def _create_lineage_by_table_name(
...
-                    column_lineage_map=column_lineage_map,
-                    lineage_source=lineage_source,
-                )
+        for table_name, entity in (
+            (from_table, from_table_entities),
+            (to_table, to_table_entities),
+        ):
+            if entity is None:
+                raise RuntimeError(f"Table entity not found: [{table_name}]")
+
```
ulixius9, line 324: I'm not sure if I like idea of raising error here, consider this scenario

```
insert into abc as (select id from xyz union select id from pqr)
```

if i do not fine table `pqr` here, lineage between abc and xyz will also not get ingested, how about just logging that the table was not found?

IceS2 (dismissed):

ingestion/src/metadata/ingestion/lineage/sql_lineage.py
sushi30: good point. Ill add this as a test case and look for a solution

sushi30: @ulixius9 looked into this and added a test case [here](https://github.com/open-metadata/OpenMetadata/pull/16756/files#diff-4b742c40aa666c03af4c10d2111fdd67f00477f040a2d8f304b9e42e6e5dcbaaR5). If a query has multiple tables, it will register an error for failed ones and add the ones which exist.

ulixius9 (approved):

ulixius9 (approved):

PR #16762, opened 2024-06-21
[DuplicateLogin] User name issue for loggedInUserApi

### LINKED ISSUE #13470: nameHash conflict for users with different mail
if we have
- levy@gmail.com
- levy@yahoo.com

Both users can register to the platform, and their name will in both cases be `levy`, which conflicts with the `nameHash` of `user_entity` being a PK.

This is not a usual scenario since orgs restrict the user registration from their domain, but still we might want to review it

mohityadav766:
### Describe your changes:

Fixes #13470

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


mohityadav766:
@chirag-madlani can you check here, we are removing support for username based login

sushi30 (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -430,15 +432,17 @@ public User getCurrentLoggedInUser(
               schema = @Schema(type = "string", example = FIELDS))
           @QueryParam("fields")
           String fieldsParam) {
+    CatalogSecurityContext catalogSecurityContext =
+        (CatalogSecurityContext) containerRequestContext.getSecurityContext();
     Fields fields = getFields(fieldsParam);
-    String currentUserName = securityContext.getUserPrincipal().getName();
-    User user = repository.getByName(uriInfo, currentUserName, fields);
+    String currentEmail = ((CatalogPrincipal) catalogSecurityContext.getUserPrincipal()).getEmail();
```
harshach, line 438: do we get the email all the time?

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -463,9 +467,13 @@ public User getCurrentLoggedInUser(
         @ApiResponse(responseCode = "404", description = "User not found")
       })
   public List<EntityReference> getCurrentLoggedInUser(
-      @Context UriInfo uriInfo, @Context SecurityContext securityContext) {
-    String currentUserName = securityContext.getUserPrincipal().getName();
-    return repository.getGroupTeams(uriInfo, currentUserName);
+      @Context UriInfo uriInfo,
+      @Context SecurityContext securityContext,
+      @Context ContainerRequestContext containerRequestContext) {
+    CatalogSecurityContext catalogSecurityContext =
```
harshach, line 473: how is this different than securityContext?

openmetadata-service/src/main/java/org/openmetadata/service/security/CatalogOpenIdAuthorizationRequestFilter.java
```diff
@@ -62,14 +63,17 @@ protected boolean isHealthEndpoint(ContainerRequestContext containerRequestConte
     return uriInfo.getPath().equalsIgnoreCase(HEALTH_END_POINT);
   }
 
-  protected String extractAuthorizedUserName(MultivaluedMap<String, String> headers) {
-    LOG.debug("Request Headers:{}", headers);
+  protected String extractAuthorizedUserName(String openIdEmail) {
```
harshach, line 66: is this only applies to openId?

openmetadata-service/src/main/java/org/openmetadata/service/security/CatalogOpenIdAuthorizationRequestFilter.java
```diff
@@ -62,14 +63,17 @@ protected boolean isHealthEndpoint(ContainerRequestContext containerRequestConte
...
+    String[] openIdEmailParts = openIdEmail.split("@");
+    return openIdEmailParts[0];
```
harshach, line 68: should there be a exception thrown if the passedin String is not email?

harshach (commented):
left comments

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/UserRepository.java
```diff
@@ -126,8 +126,13 @@ public User getByEmail(UriInfo uriInfo, String email, Fields fields) {
     if (userString == null) {
       throw EntityNotFoundException.byMessage(CatalogExceptionMessage.entityNotFound(USER, email));
     }
-    return withHref(
-        uriInfo, setFieldsInternal(JsonUtils.readValue(userString, User.class), fields));
+    User user = JsonUtils.readValue(userString, User.class);
+    setFieldsInternal(user, fields);
+    setInheritedFields(user, fields);
+    // Clone the entity
+    User entityClone = JsonUtils.deepCopy(user, User.class);
```
harshach, line 133: @mohityadav766 whats the need for this?

mohityadav766: to get entity by Email in case of UserRepository

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
mohityadav766: yes we are building that in JwtFilter , and is fetched from the token itself

openmetadata-service/src/main/java/org/openmetadata/service/security/CatalogOpenIdAuthorizationRequestFilter.java
mohityadav766: this is only used in testing

openmetadata-service/src/main/java/org/openmetadata/service/security/CatalogOpenIdAuthorizationRequestFilter.java
mohityadav766: this is only for testing

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
mohityadav766: this is the whole request context with params and everything , SecurityContext cannot be casted to CatalogPrinicalSecurirtyContext , since it gets wrapped with some container class which only has package level access

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -523,7 +523,7 @@ public void validatePassword(String providedIdentity, User storedUser, String re
 
   @Override
   public User lookUserInProvider(String userName) {
-    User storedUser;
+    User storedUser = null;
     try {
       if (userName.contains("@")) {
```
harshach, line 528: do we need check this condition anymore? we should enforce that at JwtFilter and throw an exception if there is no email is present

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -571,7 +583,8 @@ public Response createUser(
     try {
       createdUserRes = create(uriInfo, securityContext, user);
     } catch (EntityNotFoundException ex) {
-      if (securityContext.getUserPrincipal().getName().equals(create.getName())) {
+      if (isSelfSignUpEnabled
```
harshach, line 586: we should throw a clear response back that the self signup is not enabled and UI show that clearly cc @chirag-madlani 
"Self Signup is not enabled. Please contact your admin" something on these lines

chirag-madlani: @harshach How does this sound
"Self-registration is currently disabled. Please contact your administrator for assistance with account creation."

mohityadav766: **"Self Signup is not available. Please contact your Administrator."** Added this message

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
mohityadav766: problem is in existing flow we used to have principal claims, where even if someone used [sub] claim first then we built the email ourself as [sub@principalDomain] , if I do as suggested above then it will break the existing user flow since he initially used [sub or maybe preferred_username] that itself is not an email (email is what we built) , on new claim mapping have added that check (see below). but for not breaking the existing flow I have not done that for earlier flow

image

ShaileshParmar11 (dismissed):

karanh37 (approved):

PR #16777, opened 2024-06-24
Fix Airflow Pipeline Toggle On

aaron-truehold-klor:
In [this thread](https://openmetadata.slack.com/archives/C02B6955S4S/p1718753793799799?thread_ts=1717612791.169119&cid=C02B6955S4S) we discovered that it was impossible to toggle an airflow pipeline back on after it was paused without manually editing the Openmetadata database. After some code reviewing, it was found that there was a bug in the code for the state change from "paused" to "unpaused" which would cause the Openmetadata service to become out-of-sync with the underlying Airflow process.

This should resolve the issue.

Thank you to @ulixius9 and @harshach for the suggestion to submit a PR!


harshach (approved):

ulixius9 (approved):
thanks for the pr @aaron-truehold-klor 🙏

Jason-Clark-FG:
Hi All, just a question out of curiosity. Is there any chance that this fix will get incorporated into the 1.4.x series? Thanks in advance! :-)

PR #16796, opened 2024-06-25
MINOR: fix getter for failed rows sample

sushi30:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
@@ -790,6 +790,7 @@ public class TestCaseRepository extends EntityRepository<TestCase> {
             "failedRowsSample",
             JsonUtils.pojoToJson(tableData));
     setFieldsInternal(testCase, Fields.EMPTY_FIELDS);
+    testCase = JsonUtils.readValue(JsonUtils.pojoToJson(testCase), TestCase.class);
     return testCase.withFailedRowsSample(tableData);
   }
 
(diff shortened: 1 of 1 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -790,6 +790,7 @@ public TestCase addFailedRowsSample(
             "failedRowsSample",
             JsonUtils.pojoToJson(tableData));
     setFieldsInternal(testCase, Fields.EMPTY_FIELDS);
+    testCase = JsonUtils.readValue(JsonUtils.pojoToJson(testCase), TestCase.class);
```
TeddyCr, line 793: isn't `testCase` already of type `TestCase`?

mohityadav766: @sushi30  you can directly do JsonUtils.deepCopy

sushi30: it is. the object needs to be deep copied to avoid modifying the object in the cache

sushi30: changed to use JsonUtils.deepCopy

TeddyCr (approved):

PR #16809, opened 2024-06-26
MINOR: add column diff for table diff test case

sushi30:
added column validation for table diff that will be carried out before running the row level diff. If a diff for the column exists, it will short circuit the test and report.

I worked on ... because ...

#
### Type of change:

- [ ] Improvement

#
### Checklist:


ingestion/src/metadata/data_quality/validations/runtime_param_setter/table_diff_params_setter.py
```diff
@@ -93,19 +114,23 @@ def get_service2_url(self, service1, table2, test_case):
     def get_extra_columns(
         self, key_columns: List[str], test_case
     ) -> Optional[List[str]]:
```
TeddyCr, line 116: Can you add some doc string to explain what we are doing here? 

Same with the other methods

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
```diff
@@ -136,15 +141,81 @@ def _run(self) -> TestCaseResult:
...
+                table1._schema[col1]  # pylint: disable=protected-access
+            )
+            if table2._schema.get(col2) is None:  # pylint: disable=protected-access
+                continue
+            col2_type = self.get_column_type(
+                table2._schema[col2]  # pylint: disable=protected-access
+            )
+
+            if col1_type != col2_type:
+                result.append(col1)
```
TeddyCr, line 172: Can you explain why we choose to return col1 as changed if types are not comparable?

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
```diff
@@ -136,15 +141,81 @@ def _run(self) -> TestCaseResult:
...
+        return result
+
+    @staticmethod
+    def get_column_type(column: SAColumn):
```
TeddyCr, line 176: Shouldn't we map the type to OM type?

TeddyCr, line 176: Shouldn't we return the OpenMetadata column type (vs the python type)

TeddyCr (commented):
I added a few comments. Can you also add the doc strings to your methods please?

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
sushi30: I am only returning the column name (which is the same for both tables). So it does not matter which one is returned.

sushi30:
> I added a few comments. Can you also add the doc strings to your methods please?

sure. isn't the CI suppose to throw a pylint error for methods without docstrings? I'll check why it hasnt.

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
sushi30: this is for resolving `comparability`. it gets the SQA type bc then we can resolve it to a `python_type` which hints to us of the types are comparable. It's still not bullet proof because the query engine might have nuances of its own but its the best I found for now.

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
sushi30: changed it to only iterate over columns from the main table. the removed/added is covered by other parts.

TeddyCr (approved):

PR #16815, opened 2024-06-27
Fix #15721: Added Override flag to force update of Description, Tags and Owner from Source System

### LINKED ISSUE #15721: Override flag to force update of Description, Tags and Owner from Source System
**Context**
OpenMetadata ingestion will use the metadata defined in its platform as the source of truth. On first ingestion the metadata from the source system will be ingested, but subsequent ingestion will not overrird `description`, `tags` and `owner` when pulling metadata from the source.

This means that any subsequent update on the source system will not be pulled to OpenMetadata. This has raised some feedback from users managing their metadata from the source (ref. https://openmetadata.slack.com/archives/C02DZK3QH8Q/p1710928357098439).

**A.C.**
- Add a flag in the metadata ingestion workflow to force update `descriptions`, `tags` and `owner` from source. The default behavior should have the flag turned off (i.e. keep the existing behavior as the default)

OnkarVO7:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/15721

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/models/patch_request.py
```diff
@@ -375,10 +393,20 @@ def build_patch(
...
                 restrict_update_fields
-            ).update(patch)
+            ).update(patch, override_metadata=override_metadata)
+
+            # if the only operation is to update the sourceHash, we'll skip the patch
+            # since this will only happen when we filter out the REMOVE and REPLACE ops
+            # and if the sourceHash is updated in this case further updates will not be processed
+            if (
+                len(updated_operations) == 1
+                and updated_operations[0]["path"] == "/sourceHash"
```
TeddyCr, line 407: should you do `updated_operations[0].get("path")` so worst case scenario it does not fail. I am not 100% of the specs of `updated_operations` feel free to let me know if it is ok as is

TeddyCr (approved):
LGTM 🚀

ingestion/src/metadata/ingestion/models/patch_request.py
OnkarVO7: I think it should be ok. Still I'll update it to get to be on safer side

TeddyCr (approved):

PR #16825, opened 2024-06-27
Fix #15163: Added SAP ERP Connector

### LINKED ISSUE #15163: SAP ERP Connector

OnkarVO7:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/15163

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/database/saperp/metadata.py
```diff
@@ -0,0 +1,297 @@
...
+from metadata.utils import fqn
+from metadata.utils.execution_time_tracker import calculate_execution_time_generator
+from metadata.utils.filters import filter_by_table
+from metadata.utils.helpers import clean_up_starting_ending_double_quotes_in_string
+from metadata.utils.logger import ingestion_logger
+
+logger = ingestion_logger()
+
+
+class SaperpSource(CommonDbSourceService):
```
sushi30, line 64: Is there a specific reason we are not using the `CommonNoSQLSource` instead of `CommonDbSourceService`. It seems more suitable for this kind source that does not support SQL.

OnkarVO7: Makes sense I think we can convert it to NoSQLSource

OnkarVO7: @sushi30 upon checking this, the connector would makes sense to use the SQLSource since SAP ERP tables can use SQL for querying.
We use API to get the metadata out but the system is based on SQL itself

sushi30 (dismissed):

ingestion/src/metadata/ingestion/source/database/saperp/metadata.py
sushi30: I dont see a reference to SQLalchemy anywhere in the submitted code. In this case we might be duplicating a lot of logic from the NoSQL source which generally uses APIs to extract metadata (things like filters). In any case I dont think its a blocker so I approved.

sushi30 (approved):

Issue #16829, opened 2024-06-28
Airbyte lineage between airbyte pipeline (api source) and BigQuery tables


aballiet:
**Is your feature request related to a problem? Please describe.**

Airbyte can be used to pull data from API sources which are not databases.

Having lineage between API pipelines and tables (BigQuery in my case) is super useful for OMD : 
- to know where data in tables come from
- be able to see when data has been refreshed through `executions` pulled in OMD

**Describe the solution you'd like**
In case of a BigQuery destination, OMD should look for database matching with the name of the `project_id` in Airbyte destination.

See below : 
<img width="1159" alt="image" src="https://github.com/open-metadata/OpenMetadata/assets/25565600/6a412646-0db7-43ed-93f7-051b912ac438">

**Describe alternatives you've considered**
None as I implemented a solution 👍 

**Additional context**
I am willing to provide a PR

PR #16830, opened 2024-06-28
feat: create airbyte to bigquery lineage

aballiet:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/16829

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/pipeline/airbyte/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/pipeline/airbyte/metadata.py
@@ -29,6 +29,7 @@ from metadata.generated.schema.entity.data.table import Table
 from metadata.generated.schema.entity.services.connections.pipeline.airbyteConnection import (
     AirbyteConnection,
 )
+from metadata.generated.schema.entity.data.database import Database
 from metadata.generated.schema.entity.services.databaseService import DatabaseService
 from metadata.generated.schema.metadataIngestion.workflow import (
     Source as WorkflowSource,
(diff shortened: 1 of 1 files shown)


ulixius9:
@aballiet  I've couple of questions regarding this pr:

1. I see that you are trying for find the project name and then finding the bigquery service in OpenMetadata related to that project, why do we need this approach? for airbyte we expect the name of them service in OpenMetadata would be same as the destination connection name in Airbyte, can you not maintain this consistent naming for this?
2. Also there is a fallback applied when source table involved in the table is not found, this might be applicable to your use case but other users might not expect such lineage any particular reason why the source table would not be found?

aballiet:
> @aballiet  I've couple of questions regarding this pr:
> 
> 1. I see that you are trying for find the project name and then finding the bigquery service in OpenMetadata related to that project, why do we need this approach? for airbyte we expect the name of them service in OpenMetadata would be same as the destination connection name in Airbyte, can you not maintain this consistent naming for this?
> 2. Also there is a fallback applied when source table involved in the table is not found, this might be applicable to your use case but other users might not expect such lineage any particular reason why the source table would not be found?  

IMO best is to have the most config handled by OMD directly. Expect the least from the user in terms of naming convention.

Yes destination name in our airbyte is not matching the database in OMD. 

Hard to onboard with the current setup when you have your airbyte already running with 100s sync.

So better implement a search for the correct database mapping the BigQuery config in Airbyte. It reduces need for convention.

For the fallback, let me come back to you.

We cannot use the code as is as currently omd does not support this lineage. Will need to wait 1.5 anyway

aballiet:
> @aballiet I've couple of questions regarding this pr:
> 
> 1. I see that you are trying for find the project name and then finding the bigquery service in OpenMetadata related to that project, why do we need this approach? for airbyte we expect the name of them service in OpenMetadata would be same as the destination connection name in Airbyte, can you not maintain this consistent naming for this?
> 2. Also there is a fallback applied when source table involved in the table is not found, this might be applicable to your use case but other users might not expect such lineage any particular reason why the source table would not be found?

1. I think the best approach to encourage adoption of the tool is to reduce the amount of `naming convention`. When setting up OpenMetadata, you may already have a lot of tools setup in your stack so better *use truth* from config instead of user entry to perform matching.

It's just a better design approach IMO to use Google BigQuery config to match with the correct database in OMD instead of a `name` in Airbyte that can change at anytime. Enforcing new naming convention is a blocker for us, we have 100s of Airbyte connection and will not change the names just because we have OMD.

2. It's not a fallback, it's the default behavior you have to support other destination using `name` to match (Snowflake, or any other available Airbyte destination). I just made the integration for BigQuery, ideally we would do it for every major destination supported in Airbyte.

Hope it makes more sense. In any case, we cannot merge this right now, as we need to wait OMD 1.5 to support lineage for pipelines.

ulixius9:
closing this as already fixed

PR #16861, opened 2024-07-01
FIX CL-548 - Encrypt JWT Token w/ Secrets Manager

pmbrull:
### Describe your changes:

Fixes https://github.com/open-metadata/openmetadata-collate/issues/548

This PR takes care of the following:
- We are now treating the JWT Tokens for our bots the same as any other sensitive info. By default, it'll be encrypted in the db and, if using SecretsManager, it'll be stored there instead of locally.
    - This means that, if using Secrets Manager, the JWT Token won't appear in the `dag_generated_config` files in Airflow anymore, and the ometa client can load it up dynamically, making it easier to secure external ingestions.
    - We need to get the secret from SM when checking the JWTFilter or in any other internal op where we requested the JWT token
    - When revoking a token, we stored it in the db as `""` (empty string). However, in the Secrets Manager we can't store empty strings. We're then sending directly `"null"`.
- There was a small error on the logic for JSON Schema annotations, which is highlighted below.

The end result is quite cool because even if you don't use the secrets manager on the server, you can still rely on the `get_secret_value` automation for the token to pick up the JWT dynamically.

Everything is explained in the docs

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


common/src/main/java/org/openmetadata/annotations/ExposedAnnotator.java
```diff
@@ -59,8 +61,12 @@ private void addMaskedFieldAnnotationIfApplies(JMethod jMethod, String propertyN
       Field outerClassField = JMethod.class.getDeclaredField("outer");
       outerClassField.setAccessible(true);
       JDefinedClass outerClass = (JDefinedClass) outerClassField.get(jMethod);
-      if (outerClass.fields().containsKey(propertyName)
-          && outerClass.fields().get(propertyName).annotations().stream()
+
+      TreeMap<String, JFieldVar> insensitiveFieldsMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+      insensitiveFieldsMap.putAll(outerClass.fields());
```
pmbrull, line 66: we had to apply a case-insensitive map here otherwise the `jwtToken` was not getting properly annotated with `Password`.

The issue was that the field is `jWTToken` on the generated class, and the setter/getters did not directly contain the `jWTToken` string in the name as they're called `getJWTToken` & `setJWTToken`. Checking without cases handles properly the starting lowercase `j` in this scenarios

harshach (approved):

PR #16911, opened 2024-07-03
Implemented TemplateProvider.

Siddhanttimeline:
### Describe your changes:

Implemented TemplateProvider.

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach:
@Siddhanttimeline  @mohityadav766 can we not use the system settings to implement email templates and make them available through the settings in UI itself. Instead of using config in server side.

openmetadata-service/src/main/java/org/openmetadata/service/util/DefaultTemplateProvider.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  @Override
+  public Template getTemplate(String templateName) throws IOException {
+    return templateConfiguration.getTemplate(templateName);
+  }
+
+  public static String getTemplatePath() {
+    SmtpSettings emailConfig =
+        SettingsCache.getSetting(SettingsType.EMAIL_CONFIGURATION, SmtpSettings.class);
```
Siddhanttimeline, line 39: @harshach We are reading the configurations from the SystemSettings only, which are being set from the UI. In the front end, we need to set this field templatePath and send its value along with the other configuration values.

harshach: I don't want email settings to be a server side config at all. Lets drive everything through the APIs and add a ./openmetadata-ops.sh that can pickup environment variables and update the email configs.

Siddhanttimeline: Noted.

Siddhanttimeline:
Adding the changes in different branch.

Issue #16652, opened 2024-06-13
GCS Storage Container support


MChamberlin:
**FEATURE REQUEST**: Support GCS (Google Cloud Storage) connector for Storage service.

GCS seems like an important integration to have. Currently only S3 is supported out-of-the-box. Can we please add a GCS connector?

Thanks!

harshach:
GCS Supported https://docs.open-metadata.org/v1.4.x/connectors/storage/gcs

MChamberlin:
@harshach 
Hm, I suspect the docs are wrong here. I do not see code for a GCS storage source connector [in any of the latest releases](https://github.com/open-metadata/OpenMetadata/tree/1.4.3-release/ingestion/src/metadata/ingestion/source/storage). The image links in the docs for adding a new source connector are broken. And indeed I see no GCS tile when trying to create a storage service while running the latest release:
![Screenshot 2024-06-25 at 2 55 02 PM](https://github.com/open-metadata/OpenMetadata/assets/4731479/61d03e00-f251-4185-b26a-cf29d52d3fda)

PR #16917, opened 2024-07-03
Fixes 16652: add GCS storage service

MChamberlin:
### Describe your changes:

Fixes [16652](https://github.com/open-metadata/OpenMetadata/issues/16652).

Add GCS (Google Cloud Storage) storage service connection. Documentation for this service (and for ADLS) [already existed](https://docs.open-metadata.org/v1.4.x/connectors/storage/gcs), but the actual implementation did not and the service [is marked as unsupported in the UI code](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-ui/src/main/resources/ui/src/utils/ServiceUtilClassBase.ts#L132).

I assume this documentation mismatch is an oversight. The ADLS documentation should probably be removed because that connector implementation does not exist either.

This PR adds an implementation for the GCS storage service connector. It tries to follow the general pattern of the S3 connector. It supports multiple project ids and bucket metrics via the [`google-cloud-monitoring` package](https://cloud.google.com/monitoring/docs/reference/libraries#client-libraries-install-python). 

Connector UI screenshots:

Screenshot 2024-07-03 at 11 07 18 AM

Screenshot 2024-07-03 at 11 08 52 AM

Screenshot of storage container with structured data (parsed from manifest file):
Screenshot 2024-07-03 at 11 23 59 AM

#
### Type of change:
- [x] New feature

#
### Checklist:


pmbrull:
Hi @MChamberlin thanks for the PR. Could you please check the python checkstyle? a `make py_format` might do the trick. Thanks

openmetadata-service/src/main/resources/json/data/testConnections/storage/gcs.json
```diff
@@ -11,25 +11,10 @@
       "mandatory": true
     },
     {
-      "name": "ListBlobs",
```
pmbrull, line 14: hi @MChamberlin why are we removing these? In order to sample the container files it's important we can list the blobs and get them I believe

MChamberlin: I removed these to bring the test logic in line with the existing S3 connector. I assume they were removed there because the additional tests require that state be tracked in the test client, which complicates things a bit. I can add them back though.

pmbrull: hi yes I think it makes sense to make a quick check of those variables. It would also make sense adding them to S3, but that'd be work for another PR :)

we could just call the GCS client to `list_blobs`, `get_bucket` etc. to ensure all the ingredients are working before running the connector.

Thanks!

pmbrull (dismissed):
Great PR, thanks @MChamberlin

openmetadata-service/src/main/resources/json/data/testConnections/storage/gcs.json
MChamberlin: Fixed in cd0e4e0

pmbrull (approved):
Thanks for the PR

PR #16943, opened 2024-07-05
MINOR: fix sample data issue with Pydantic v2 and refactor python integration tests

sushi30:
The profiler test for MSSQL was observing many errors due to unhandled data types:
- binary and varbinary -> changed [common type map](https://github.com/open-metadata/OpenMetadata/pull/16943/files#diff-bb3c60d6a4f252903d4e8b01c2070033846c1ebab127df459ac7604f7a941024R54-R55)
- bit -> use an explicit [bit type](https://github.com/open-metadata/OpenMetadata/pull/16943/files#diff-bb3c60d6a4f252903d4e8b01c2070033846c1ebab127df459ac7604f7a941024R54-R55)
- other types (`geometry(-1)` for example) -> created an [undetermined type](https://github.com/open-metadata/OpenMetadata/blob/f373f4f6ef5d0034c93ba622faf0d919f07710bd/ingestion/src/metadata/profiler/orm/types/undetermined_type.py) to catch all types unhandled by OM 
- XML -> added to the [NOT_COMPUTE set](https://github.com/open-metadata/OpenMetadata/blob/f373f4f6ef5d0034c93ba622faf0d919f07710bd/ingestion/src/metadata/profiler/orm/types/undetermined_type.py)

The errors for the binary/varbinary looked like this:

```python
[2024-07-09 12:56:10] DEBUG    {metadata.OMetaAPI:table_mixin:76} - Traceback (most recent call last):
  File "/Users/imri/projects/OpenMetadata-pycharm/ingestion/src/metadata/ingestion/ometa/mixins/table_mixin.py", line 73, in ingest_table_sample_data
    data=sample_data.model_dump_json(),
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/imri/projects/OpenMetadata-pycharm/ingestion/src/metadata/ingestion/models/custom_pydantic.py", line 65, in model_dump_json
    self.model_dump(
  File "/Users/imri/projects/OpenMetadata-pycharm/.venv/lib/python3.11/site-packages/pydantic/main.py", line 347, in model_dump
    return self.__pydantic_serializer__.to_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf7 in position 10: invalid utf-8
```

## Integration Tests Refactor

refactor tests and consolidate common functionality in `integrations.conftest`.
this enables writing tests more concisely.
demonstrated with postgres/mssql/trino

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/setup.py
```diff
@@ -342,6 +342,7 @@
     "pytest-order",
     # install dbt dependency
     "dbt-artifacts-parser",
+    "freezegun",
```
IceS2, line 345: Interesting, didn't know that library!

IceS2 (approved):

ingestion/setup.py
ayush-shah: +1 👀

ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
```diff
@@ -530,8 +530,15 @@ def fetch_sample_data(self, table, columns) -> TableData:
...
-        return sampler.fetch_sample_data(columns)
+        sample = sampler.fetch_sample_data(columns)
+        # TODO: this is innefficient, we should change the schema to be a string and
+        # manage the casting when creating the TableData object with the logic coupled to OpenMetdata
+        # DataType / Pydantic instead of letting the underlying ORM / DB client handle it.
+        for i, row in enumerate(sample.rows):
+            sample.rows[i] = list(row)
+            for j, value in enumerate(row):
+                sample.rows[i][j] = self._json_safe(value)
+        return sample
```
TeddyCr, line 541: @sushi30 what case are we handling in this scenario?

sushi30:
```
[2024-07-09 12:56:10] DEBUG    {metadata.OMetaAPI:table_mixin:76} - Traceback (most recent call last):
  File "/Users/imri/projects/OpenMetadata-pycharm/ingestion/src/metadata/ingestion/ometa/mixins/table_mixin.py", line 73, in ingest_table_sample_data
    data=sample_data.model_dump_json(),
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/imri/projects/OpenMetadata-pycharm/ingestion/src/metadata/ingestion/models/custom_pydantic.py", line 65, in model_dump_json
    self.model_dump(
  File "/Users/imri/projects/OpenMetadata-pycharm/.venv/lib/python3.11/site-packages/pydantic/main.py", line 347, in model_dump
    return self.__pydantic_serializer__.to_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf7 in position 10: invalid utf-8
```

also added to the PR description

TeddyCr: Do you know the type of the column and for what service it happens. Basically, [here](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/profiler/orm/types/custom_hex_byte_string.py) we convert `BYTES` column type result on SQA side to a string representation of the bytes. 

We might be missing this for a specific type/service which might cause the `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf7 in position 10: invalid utf-8` error.

sushi30: If I understand correctly, this requires defining a custom type map per database:
https://github.com/open-metadata/OpenMetadata/blob/3c36a3f1b27e52c2cb7785156d4feb189e7865b0/ingestion/src/metadata/profiler/orm/converter/snowflake/converter.py#L28

This will not scale as we add more connectors. We need something to catch this dynamically.

I can introduce a patch for mssql specifically but this fix will still be required regardless to avoid introducing un-handeled data types in the future. With current implementation, any introduction of a new data source or new data type can cause the sample ingestion to fail and there is no way of detecting it until its in service.

TeddyCr: It is not per service, but by type, but some services have their own specific type (hence the question around the service). In this case, we would create the custom type to handle the specific type that returns a raw bytes string. 

We should handle that decode error at the source using the SQA Custom type. This patch has O(n^2) efficiency so it can quickly become quite slow.

sushi30: Because each service can introduce its own arbitrary types, in the end it will be defined by service with some potential to reuse logic between services.

In this example, Snowflake defines a type of `BINARY` and mssql uses something like `BYTES` but another service might define `BYTESBUFFER` and so on.

The root issue here is that `TableData.rows` can contain any type of object so there is no way of guaranteeing that the [serializer running the the sink](https://github.com/open-metadata/OpenMetadata/blob/b960b60965593c584247f8bba3506295cc000377/ingestion/src/metadata/ingestion/ometa/mixins/tests_mixin.py#L344) will know how to handle it.

I can look at defining this at the serializer level in pydantic to replace the value with something like `<UNSERIALIZABLE_VALUE>`. This can be done by implementing a [CustomSerializer](https://docs.pydantic.dev/latest/concepts/serialization/#dictmodel-and-iteration) for the `TableData` type.

How does that sound?

TeddyCr: > Because each service can introduce its own arbitrary types
That is true at the marge. `BINARY` and `BYTES` are used across many different services. Creating a custom type to handle binary will then be used across all services. 

My point is that we should only serialize what we need to serialize and where we need to.

sushi30: Added handling using a custom pydantic serializer for `TableData`

ingestion/src/metadata/profiler/interface/nosql/profiler_interface.py
```diff
@@ -21,11 +21,10 @@
 
 from sqlalchemy import Column
 
-from metadata.generated.schema.entity.data.table import TableData
 from metadata.generated.schema.tests.customMetric import CustomMetric
 from metadata.profiler.adaptors.factory import factory
 from metadata.profiler.adaptors.nosql_adaptor import NoSQLAdaptor
-from metadata.profiler.api.models import ThreadPoolMetrics
+from metadata.profiler.api.models import SerializableTableData, ThreadPoolMetrics
```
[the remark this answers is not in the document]
sushi30: i am open to suggestions. but since `TableData` holds `any` in the schema, it needs to have some custom serialization logic so that it will be fail-proof. Another direction would be to hack it into the data model creation.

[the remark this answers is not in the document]
IceS2: It's a bit more verbose, but I think it'd be ok to just have a different class that does not inherit from TableData but is able to be casted into.

Something onlines of:

```python
class SerializableTableData(BaseModel):
    attr1: String
    attr2: Int
    attr3: Boolean

    def fromTableData(cls, TableData tableData) -> SerializableTableData:
        return cls(
            attr1=str(tableData.attr1),
            attr2=tableData.attr2,
            attr3=tableData.attr3
       )
```

I feel this should be safe enough so that if we modify TableData in a non-compatible way it should break. We could add some test to make sure of it as well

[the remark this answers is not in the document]
sushi30: we discussed a solution to change the TableData schema to `List[List[str]]`. It will be implemented in another PR.

TeddyCr (approved):

TeddyCr (approved):
LGTM 🚀

PR #16974, opened 2024-07-09
[Issue-16973] Fix Issue in Import TableCsv

mohityadav766:
### Describe your changes:

- This PR fixes the backend issue on validation in importing a CSV
- Pending Fix from the UI side for "" on empty cells

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/csv/EntityCsv.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/csv/EntityCsv.java
@@ -74,7 +74,7 @@ public abstract class EntityCsv<T extends EntityInterface> {
   private final String entityType;
   private final List<CsvHeader> csvHeaders;
   private final List<String> expectedHeaders;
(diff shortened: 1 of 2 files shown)


TeddyCr (commented):
LGTM 🚀 @mohityadav766 are we able to add a test to validate this logic works as expected and that we do not introduce regression in the future 😊

mohityadav766:
> LGTM 🚀 @mohityadav766 are we able to add a test to validate this logic works as expected and that we do not introduce regression in the future 😊

hi @TeddyCr , thanks for reviewing, from backend perspective it was all good there were empty cell issue from the UI , maybe we should cover those with cypress cc @chirag-madlani

mohityadav766:
merging this , @chirag-madlani let's add cypress via another PR

PR #16982, opened 2024-07-10
MINOR: Added Alation Sink connector

OnkarVO7:
### Describe your changes:

Added Alation Sink connector

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [x] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/api/step.py
```diff
@@ -211,7 +211,7 @@ def run(self) -> Iterable[Optional[Entity]]:
         in order to create a generator with `yield`.
         """
         try:
-            for result in self._iter():
+            for result in self._iter() or []:
```
pmbrull, line 214: not sure if we want this. Might be better to let the source blow up if we don't implement properly returning a generator

ingestion/src/metadata/ingestion/source/metadata/alationsink/client.py
```diff
@@ -0,0 +1,207 @@
...
+    def __init__(self, config: AlationSinkConnection):
+        self.config = config
+        self._auth_provider = AlationSinkAuthenticationProvider.create(config)
+        client_config: ClientConfig = ClientConfig(
+            base_url=clean_uri(config.hostPort),
+            auth_header="TOKEN",
+            api_version="integration",
+            auth_token=self._auth_provider.get_access_token,
+            auth_token_mode="",
+            verify=False,
```
pmbrull, line 117: should this be hardcoded as `False`? We might need to allow some tuning of the certs here

ingestion/src/metadata/ingestion/source/metadata/alationsink/client.py
```diff
@@ -0,0 +1,207 @@
...
+        self, api_url: str, data: Optional[Dict] = None, is_key_offset: bool = False
+    ) -> Optional[List[Any]]:
+        """
+        Method to paginate the entities
+        """
+        skip_key_name = "skip"
+        if is_key_offset:
+            skip_key_name = "offset"
+        entities = []
+        total_records = 100000000
```
pmbrull, line 132: this might be better a global var, easier to control? Plus is this the records we pick in one go?

ingestion/src/metadata/ingestion/source/metadata/alationsink/metadata.py
```diff
@@ -0,0 +1,397 @@
...
+                f"Unable to ingest columns for table [{model_str(om_table.name)}]: {exc}"
+            )
+
+    def ingest_tables(self, alation_datasource_id: int, om_schema: DatabaseSchema):
+        """
+        Method to ingest the tables
+        """
+        try:
+            # Iterate over all the tables in OpenMetadata
+            om_tables = list(
```
pmbrull, line 264: do we need to force the list here?

ingestion/src/metadata/examples/workflows/alationsink.yaml
```diff
@@ -0,0 +1,34 @@
...
+      paginationLimit: 10
+      # datasourceLinks: {
+      #   "23": "om_service_name.om_db_name",
+      #   "24": "om_service_name_two.om_db_name_two",
+      # }
+  sourceConfig:
+    config:
+        type: DatabaseMetadata
+sink:
+  type: metadata-rest
```
pmbrull, line 26: why is it metadata-rest?

ingestion/src/metadata/ingestion/api/step.py
OnkarVO7: Makes sense I'll remove that

ingestion/src/metadata/ingestion/source/metadata/alationsink/client.py
OnkarVO7: Got it I'll add ssl methods

ingestion/src/metadata/ingestion/source/metadata/alationsink/client.py
OnkarVO7: Since the alation APIs do not provide us how many total records are present, we are setting a upper limit on the total records.

The pagination logic will always and break the loop if the next API does not return any results

ingestion/src/metadata/examples/workflows/alationsink.yaml
OnkarVO7: Currently we don't have specific sink for the metadata sources

ingestion/src/metadata/ingestion/source/metadata/alationsink/metadata.py
OnkarVO7: Yes the list gets used again later [here](https://github.com/open-metadata/OpenMetadata/pull/16982/files#diff-b8127404335acba3b5bb1df9128cdb3928a500d205dbab6026695c352837a506R291)

ingestion/src/metadata/ingestion/source/metadata/alationsink/client.py
pmbrull: ok thanks, lets have this as a constant with this comment at the top

ingestion/src/metadata/examples/workflows/alationsink.yaml
pmbrull: we did not since we always relied to OM as a sink before. Since this is an Alation Sink, maybe we should create the right sink class?

pmbrull (approved):

PR #16992, opened 2024-07-10
MINOR: CI: added doctest

sushi30:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


TeddyCr:
What would be the benefit (vs using unit test)

sushi30:
@TeddyCr doctest execution in CI

sushi30:
The examples in the docs are tested so you get 2:1 documentation and testing. Its very useful for util functions with very defined scopes.

sushi30:
Widely used in python ecosystem (for example [pandas](https://github.com/pandas-dev/pandas/blob/d96646219618e007f64ee49e0a6e20f4aea761b5/ci/code_checks.sh#L60))

pmbrull:
> @sushi30 adding an example to the doc is a good use case. I think we should keep 1 testing strategy though so we can easily align as a team.
> 
> cc @pmbrull if you have any thoughts.

+1 on keeping it simple. Agree that we have room for improvement on the docstrings, but we should make it simpler for us to write the tests. Having too many options at this point will add too much noise both for us and the community

PR #17003, opened 2024-07-12
feat(#10933): API service UI

### LINKED ISSUE #10933: New Entity: API Services and support OpenAPI connector
I am having trouble to register metadata for data sources like  Interfaces, for example APIs.  Let's imagine an extreme (or not so) scenario where **company is ONLY using SaaS applications and they only have APIs** - **how its Data Catalog would look like? What metadata would it host**? Current OMD Entities and Connectors don't fully cater such scenario.

In general, I struggle to represent below scenarios:
- **For SaaS Applications** - as we are using more and more of SaaS application - we lose access to databases and ability to scan their data dictionaries. Those SaaS applications share data via APIs and they represent the data source. 
- **For Internal Applications**  -  
- - an application might have one to many databases, schemas and tables. And let's say it gives me 1000 attributes. And, naturally, I would like to inventorize that, because these are my data resources. It **represents what data the organization has**. 
- - However, that data is only accessible via interfaces (it might be an API (json schema, xml schema), an extract file, a view. And let's say - they cover 500 of attributes. And I would like to document that as well, because this shows my current **TRUE data accessibility**.
- However, in my data assets for that application I wouldn't not want to show that I have 1500 attributes. So - I do not want to duplicate data assets. 

So I would like to be able to document a) what data assets an organization has and b) which of them it can access and through which end-points. And for SaaS applications, I might only be able to register the end-points. This would require:

- **A new Entity Type - Interface**. Interface might represent an API, a WebHook, a file extract or maybe a View (is one has no option to scan the database).
- **A connector to harvest metadata from OpenAPI schemas** or being able to read metadata APIs that some SaaS apps  offer. For example - SalesForce has [metadata api.](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_intro.htm#:~:text=The%20main%20purpose%20of%20Metadata,work%20directly%20with%20business%20data.)

This would allow me to have metadata in Data Catalog  represented like: 
image

 - loading API information as Tables (via custom CSV connector), where I first extract relevant OpenAPI information and map the fields to Table Schema. 

**Additional context**

- A note on duplications - I will have API endpoints, or wehbooks, where same EventID will be repeated. Or when there is many interfaces for point-to-point integrations. 
- Data lineage becomes an important part as well as attributes might get renamed or transformed when prepared for interfaces.

Sachin-chaurasiya:
### Describe your changes:

Fixes #10933 

I worked on adding UI support for #10933

#
### Type of change:

- [x] New feature

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/utils/GlobalSettingsUtils.tsx
```diff
@@ -71,376 +33,6 @@ export interface SettingMenuItem {
   items?: SettingMenuItem[];
 }
 
-export const getGlobalSettingsMenuWithPermission = (
```
Sachin-chaurasiya, line 74: This is not being used and we are using GlobalSettingsClassBase.ts

chirag-madlani (approved):

PR #17008, opened 2024-07-12
MINOR - enable dynamic assertion dl

TeddyCr:
### Describe your changes:
- Refactor runtime parameter setter:
    - inject runtime parameter dependencies into the validator object
    - use builder classes to compose validator object
- Add runtime parameter setter support for pandas source
- Added  missing dependencies for tests

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Improvement


ingestion/src/metadata/data_quality/builders/pandas_validator_builder.py
```diff
@@ -0,0 +1,66 @@
...
+        self.validator_cls: Type[BaseTestValidator] = import_test_case_class(
+            entity_type,
+            "pandas",
+            self.test_case.testDefinition.fullyQualifiedName,  # type: ignore
+        )
+        self.reset()
+
+    def reset(self):
+        """Reset the builder"""
+        self._validator = self.validator_cls(
```
pmbrull, line 52: I might be missing something in the flow, but a couple of questions:
1. Why do we call the `self.reset()` in the `init`?
2. In which case the `reset` will be dependant on the builder? (at least now pandas and SQA have the same implementation as far as I can see)

TeddyCr: @pmbrull it is part of the recommend pattern to ensure you always work with a new object (i.e. an object that has not been composed yet).

They are different, pandas would import the `pandas` source folder while the SQA sources from `sqalchemy`. LMK if that makes sense.

pmbrull (approved):

pmbrull (approved):

PR #17015, opened 2024-07-15
Email template provider

Siddhanttimeline:
### Describe your changes:

Implemented TemplateProvider.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/resources/docstore/DocStoreResource.java
```diff
@@ -320,9 +327,40 @@ public Response createOrUpdate(
       @Context SecurityContext securityContext,
       @Valid CreateDocument cd) {
     Document doc = getDocument(cd, securityContext.getUserPrincipal().getName());
+
+    Response emailTemplateValidationResponse = null;
+    if (doc.getEntityType().equals(DefaultTemplateProvider.ENTITY_TYPE_EMAIL_TEMPLATE)) {
```
mohityadav766, line 332: move this condition to validateTemplate and if the entityType is not EMAIL_TEMPLATE return

mohityadav766 (commented):
@Siddhanttimeline this is not inline with our design, how will the UI know that there are placeholders , you should not be using regex , but rather come up with a schema for this 
That schema can have different fields like we talked about , this looks good from backend perspective, but will have issue on UI side

openmetadata-service/src/main/java/org/openmetadata/service/resources/docstore/DocStoreResource.java
Siddhanttimeline: Noted.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DocumentRepository.java
```diff
@@ -82,7 +110,16 @@ public DocumentUpdater(Document original, Document updated, Operation operation)
     @Transaction
     @Override
     public void entitySpecificUpdate() {
-      recordChange("data", original.getData(), updated.getData(), true);
+      if (updated.getEntityType().equals(DefaultTemplateProvider.ENTITY_TYPE_EMAIL_TEMPLATE)) {
```
mohityadav766, line 113: what are we doing here if the entityType is not EMAIL will we not allow updating data? we just had to add validation in case of EMAIL rest should have been same

openmetadata-spec/src/main/resources/json/schema/entity/docStore/document.json
```diff
@@ -10,7 +10,12 @@
     "data": {
       "javaType": "org.openmetadata.schema.entities.docStore.Data",
       "type": "object",
-      "additionalProperties": true
+      "additionalProperties": true,
```
mohityadav766, line 13: here if we are making data is of type emailTemplate, is that right?

mohityadav766: data is a generic data object that can store any type of data as per our need

Siddhanttimeline: >> here if we are making data is of type emailTemplate, is that right? 
no, I've created a new key emailTemplate inside data, and set its type as  emailTemplate.

mohityadav766: why data would have nameKey emailTemplate, it can store anything , we have created our type emailTemplate.json, but we cannot hardbind that on data saying data will have a name key emailTemplate, does all data exhibit this property .. no !

Siddhanttimeline: You are correct. I hadn't considered that. I am updating the code accordingly.

mohityadav766 (approved):

PR #17041, opened 2024-07-16
Fix #12677: Added Synapse Connector - docs and side docs

### LINKED ISSUE #12677: Add Synapse connector for OM
It would be great if we have a synapse connector for OM for our use case.

Add synapse as a connector.

**Additional context**

SumanMaharana:
### Describe your changes:

Fixes #12677 - Added Synapse Connector - docs and side docs

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [x] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/connections/test_connections.py
```diff
@@ -277,7 +277,7 @@ def test_connection_engine_step(connection: Engine) -> None:
     Generic step to validate the connection against a db
     """
     with connection.connect() as conn:
-        conn.execute(ConnTestFn())
+        conn.execute(ConnTestFn()).fetchall()
```
ulixius9, line 280: why fetchall, if we really need to fetch lets do fetchone

ulixius9 (dismissed):

ingestion/src/metadata/ingestion/connections/test_connections.py
ulixius9: also if this is only required for synapse and not other connectors then can we move it in the connection file of synapse?

SumanMaharana: yes will change to `fetchone`.
Had to change this as just using the `.execute` made the connection to be busy in some cases as it maybe sometimes returns a `generator` object leaving the connection object.
more info about it [here](https://stackoverflow.com/questions/46489749/whats-the-difference-between-using-c-fetchall-vs-just-assigning-c-execute)

ulixius9 (approved):

ShaileshParmar11 (approved):

PR #17055, opened 2024-07-17
Fixes #16501: Optimize Rows Sample Type Snowflake

### LINKED ISSUE #16501: When using "Rows" for "Profile Sample Type", Snowflake Profiler is inefficient
**Affected module**
Ingestion Framework

**Describe the bug**
When using "Rows" for "Profile Sample Type" in Profiler Ingestion, the ORM does not leverage Snowflake's SAMPLE feature, but instead creates a CTE using `WITH "TABLE_NAME_rnd" AS ( SELECT … MOD(0, xxxxx) AS random …. ORDER BY random LIMIT XXXX) …`.

This is very inefficient on very large tables, as it must perform the `MOD(0, xxxxx) as random` bit against every single row in the table before it can perform the pseudo sort (on a column named "random" which all consist of same value). In one instance, we have 80B+ records in a table, and it took over 8 hours to compute for every record on a medium sized Snowflake warehouse.

Switching it over to "Percentage"  for "Profile Sample Type" results in the CTE using `SAMPLE BERNOULLI`, which is an improvement, but still not ideal, as it must perform the calculation against large chunk of the data. Going back to the 80B+ records table example, when using 5%, it must compute `MOD(0, 100)` against 4B records. However, it was able to complete the request in 26 minutes on the same medium sized Snowflake warehouse.

Because we have table of varying size (some as small as 1000 records) in this schema, configuring to 1% sample is not an ideal solution. 

If we can set 1M records as sample threshold, and a path forward for this would be to leverage the  `TABLESAMPLE (XXXXX ROWS)` query in Snowflake in the `_rnd` CTE block. By leveraging the `TABLESAMPLE (XXXXX ROWS)` in the `_rnd` CTE, it would only need to perform the `MOD(0, XXXXX)` on the configured records (i.e.: 1M vs 4B). This query took 22 minutes to run on the same medium sized Snowflake warehouse. 

I do not fully grasp the requirement for `MOD(0, XXXXX)` bit, but even with the `MOD(0, XXXXX)` bit in place, I can't help but to wonder if this bit could be removed, but that might be another analysis at another time.

All test were done with warehouse from cold start with at least 15 minutes between suspend and wake up.

**To Reproduce**

1. Create a large table with a lot of data.
2. Configure "Profile Sample Type" to "Percentage", observe relatively long but acceptable run time;
3. Configure "Profile Sample Type" to "Rows", observe a much longer run time;
4. Test ORM generated query with minimal modification, observe relatively long but acceptable run time

**Expected behavior**

I wish to see "Rows" based "Profile Sample Type" to run without needing to load the entire table before discarding.

**Version:**
 - OS: Docker OMD 1.4.0

**Additional context**
Reference Slack Chat: https://openmetadata.slack.com/archives/C02B6955S4S/p1717185445543959

ayush-shah:
### Describe your changes:

Fixes #16501

I worked on optimizing row sample type in snowflake

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/profiler/processor/sampler/sqlalchemy/snowflake/sampler.py
+++ b/ingestion/src/metadata/profiler/processor/sampler/sqlalchemy/snowflake/sampler.py
@@ -47,7 +47,6 @@ class SnowflakeSampler(SQASampler):
             rnd = (
                 self._base_sample_query(
                     column,
-                    (ModuloFn(RandomNumFn(), 100)).label(RANDOM_LABEL),
                 )
                 .suffix_with(
                     f"SAMPLE BERNOULLI ({self.profile_sample or 100})",
@@ -55,7 +54,7 @@ class SnowflakeSampler(SQASampler):
                 .cte(f"{self.table.__tablename__}_rnd")
             )
             session_query = self.client.query(rnd)
-            return session_query.where(rnd.c.random <= self.profile_sample).cte(
+            return session_query.cte(
                 f"{self.table.__tablename__}_sample"
             )
 
--- a/ingestion/src/metadata/profiler/processor/sampler/sqlalchemy/snowflake/sampler.py
+++ b/ingestion/src/metadata/profiler/processor/sampler/sqlalchemy/snowflake/sampler.py
@@ -21,7 +21,6 @@ from sqlalchemy.sql.selectable import CTE
 from metadata.generated.schema.entity.data.table import ProfileSampleType
 from metadata.profiler.orm.functions.modulo import ModuloFn
 from metadata.profiler.orm.functions.random_num import RandomNumFn
-from metadata.profiler.orm.registry import Dialects
 from metadata.profiler.processor.handle_partition import partition_filter_handler
 from metadata.profiler.processor.sampler.sqlalchemy.sampler import (
     RANDOM_LABEL,
@@ -52,7 +51,6 @@ class SnowflakeSampler(SQASampler):
                 )
                 .suffix_with(
                     f"SAMPLE BERNOULLI ({self.profile_sample or 100})",
-                    dialect=Dialects.Snowflake,
                 )
                 .cte(f"{self.table.__tablename__}_rnd")
             )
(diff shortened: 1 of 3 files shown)


ingestion/src/metadata/profiler/processor/sampler/sqlalchemy/snowflake/sampler.py
TeddyCr, line 1: @ayush-shah can you move the one for the percentage out of the core sampler into here?

ayush-shah: Modified the same, Thanks 🙏

ingestion/src/metadata/profiler/processor/sampler/sqlalchemy/snowflake/sampler.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_sample_query(self, *, column=None) -> CTE:
+        """get query for sample data"""
+        # TABLESAMPLE SYSTEM is not supported for views
+        self.table = cast(Table, self.table)
+
+        if self.profile_sample_type == ProfileSampleType.PERCENTAGE:
+            rnd = (
+                self._base_sample_query(
+                    column,
+                    (ModuloFn(RandomNumFn(), 100)).label(RANDOM_LABEL),
```
TeddyCr, line 50: I don't think we need that anymore here as the sampling will be done from `SAMPLE BER....`

ingestion/src/metadata/profiler/processor/sampler/sqlalchemy/snowflake/sampler.py
```diff
@@ -0,0 +1,68 @@
...
+                )
+                .suffix_with(
+                    f"SAMPLE BERNOULLI ({self.profile_sample or 100})",
+                )
+                .cte(f"{self.table.__tablename__}_rnd")
+            )
+            session_query = self.client.query(rnd)
+            return session_query.where(rnd.c.random <= self.profile_sample).cte(
+                f"{self.table.__tablename__}_sample"
+            )
```
TeddyCr, line 60: Same thing on the random logic

TeddyCr (approved):

PR #17073, opened 2024-07-18
Fixes #17051: Dynamic import for Profiler Interface

### LINKED ISSUE #17051: pyhive required even when not used
**Affected module**
Ingestion

**Describe the bug**
When creating profiling job (for example on Trino service), `pyhive` dependency needs to be installed even if it's not used. This is because in `profiler_interface_factory.py` https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/profiler/interface/profiler_interface_factory.py#L61 `DatabricksProfilerInterface` is imported and, even if not used, it will throw `ModuleNotFoundError: No module named 'pyhive'` 

**To Reproduce**
pip3 install openmetadata-ingestion==1.4.0.0 and create profiling workflow.

**Expected behavior**
Profiling workflow does not fail on missing pyhive when profiling is done on Trino service.

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8] 1.4.0.0

**Additional context**
As a workaround `openmetadata-ingestion[hive]` dependency can be installed (even when not used) but I don't think it should be a target solution.

ayush-shah:
### Describe your changes:

Fixes #17051 

I worked on dynamically importing profiler interface on runtime

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


sushi30 (approved):
nit: I suggest defining the overrides at the connector level and not in the profiler submodule.

PR #17121, opened 2024-07-22
Minor: Bug fixes and improvements

aniketkatkar97:
I worked on:

- [x] Fixing the space showing when no icon is present for an entity type

**Before:**

Screenshot 2024-07-22 at 9 21 42 AM

**After:**

Screenshot 2024-07-22 at 9 13 22 AM

- [x] Fix the entity suggestion API failing in case of special characters searched (Tests are covered in automation playwright)

**Before:**

https://github.com/user-attachments/assets/3af86ea5-11c1-4492-9be2-148b4af35eab

**After:**

https://github.com/user-attachments/assets/6f6450ef-7239-4025-ad65-8dee9d53c12c

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/rest/miscAPI.ts
```diff
@@ -264,7 +265,9 @@ export const getAggregateFieldOptions = (
   value: string,
   q: string
 ) => {
-  const withWildCardValue = value ? `.*${value}.*` : '.*';
+  const withWildCardValue = value
```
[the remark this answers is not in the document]
aniketkatkar97: I implemented this as the API failed if we searched for a name with some special character like `&`. I changed it in this function itself to fix this issue wherever we are using this function to fetch options.
I found this out while writing test for automation. Then checked the advanced search and quick filters for the same. This will fix that issue in all places.

openmetadata-ui/src/main/resources/ui/src/utils/TableUtils.tsx
```diff
@@ -284,7 +289,8 @@ export const getEntityIcon = (
       break;
   }
 
-  return Icon ? <Icon className={className} style={style} /> : <></>;
+  // If icon is not found, return null
+  return Icon ? <Icon className={className} style={style} /> : null;
```
chirag-madlani, line 293: @aniketkatkar97 any reason to return null here?

aniketkatkar97: @chirag-madlani I changed this to null so that we can check if the icon is present for the passed entity name and perform actions according to that outside the function. For example, in the first issue described in the description above, I checked if the icon is present there to render the \<Col>, which removed the empty space where it's not needed

chirag-madlani (approved):

PR #17149, opened 2024-07-23
MINOR: Refactor output_handlers to a WorkflowOutputHandler class

IceS2:
Small refactor to reduce the amount of code and dependencies we have for the Workflows.
Instead of having `output_handler.py`, `application_output_handler.py` and `workflow_output_handler.py` files dealing with different parts of the output now we have it merged in one `WorkflowOutputHandler` class that will be instantiated when a Workflow is created.

This way the workflow is responsible for printing it's own status and summary with the help of this class instead of having both of them depending on each other.

TODO:
- [x] Create print_status(workflow) function to avoid breaking user's current scripts. Add a deprecation notice with the information on which release will be removed
- [x] Review the functions in workflow_output_handler that are not within the class
  - Extracted `WorkflowInitErrorHandler` from the `workflow_output_handler.py` file to keep the flow within a context.
- [x] Uncomment the workflow_output_handler.py file from basedpyright and fix any issues
- [x] Review/Add Tests

#
### Type of change:

- [x] Improvement

#
### Checklist:


ingestion/src/metadata/ingestion/ometa/client_utils.py
```diff
@@ -28,7 +28,7 @@
 
 def create_ometa_client(
     metadata_config: OpenMetadataConnection,
-) -> OpenMetadata:
+) -> OpenMetadata[T, C]:  # pyright: ignore[reportInvalidTypeVarUse]
```
TeddyCr, line 31: Do you know if the using `# type: ignore` works as well?

ingestion/src/metadata/workflow/application_output_handler.py
```diff
@@ -8,27 +8,14 @@
...
-    if workflow.runner.get_status().source_start_time:
-        log_ansi_encoded_string(
-            color=ANSI.BRIGHT_CYAN,
-            bold=True,
-            message="Workflow finished in time: "
-            f"{pretty_print_time_duration(time.time()-workflow.runner.get_status().source_start_time)}",
-        )
+@deprecated(message="Use 'workflow.print_status()' instead.", release="1.8")
+def print_status(workflow: BaseWorkflow):
+    workflow.print_status()
```
TeddyCr, line 21: ❤️

I think we could move the deprecation to 1.6 IMO. It gives people an entire release cycle to move their stuffs.

ingestion/src/metadata/workflow/base.py
```diff
@@ -106,6 +108,12 @@ def __init__(
 
         self.post_init()
 
+    @property
+    def output_handler(self) -> WorkflowOutputHandler:
+        if not self._output_handler:
+            self._output_handler = WorkflowOutputHandler()
+        return self._output_handler
```
TeddyCr, line 115: Could we instantiate it with the class in the `__init__` -- or even pass it as an argument to where we instantiate that class (so users could swap output handler if ever needed)?

ingestion/src/metadata/workflow/base.py
```diff
@@ -114,6 +122,10 @@ def ingestion_pipeline(self) -> Optional[IngestionPipeline]:
 
         return self._ingestion_pipeline
 
+    def with_output_handler(self, output_handler: WorkflowOutputHandler):
+        self._output_handler = output_handler
+        return self
```
TeddyCr, line 127: We could also use a setter method if we are going to keep it as a `property`

ingestion/src/metadata/workflow/workflow_init_error_handler.py
```diff
@@ -0,0 +1,183 @@
...
+    Workflow type enums based on the `metadata` CLI commands
+    """
+
+    INGEST = "ingest"
+    PROFILE = "profile"
+    TEST = "test"
+    LINEAGE = "lineage"
+    USAGE = "usage"
+    INSIGHT = "insight"
+    APP = "application"
```
TeddyCr, line 40: Should we unify this with `org.openmetadata.schema.entity.services.ingestionPipelines.PipelineType`?

ingestion/src/metadata/workflow/workflow_init_error_handler.py
TeddyCr, line 1: WDYT if making the method inside `WorkflowInitErrorHandler` as module functions? They seem to be all `staticmethod` I think we can bypass the overhead of object creation. 😊

TeddyCr (commented):
Thanks for the PR @IceS2, left a few comments/questions

ingestion/src/metadata/workflow/workflow_init_error_handler.py
IceS2: They were just functions, but I preferred them within a class just because they share the same context and they are all but one used by the "public" one.

I don't have a strong opinion thought. If you folks prefer having it as plain functions I'd be fine with it!

ingestion/src/metadata/workflow/base.py
IceS2: I think it makes sense. We can change it to a setter without returning self.

ingestion/src/metadata/workflow/base.py
IceS2: I would be fine passing it as an argument in the init and maybe having the logic there as well.

ingestion/src/metadata/workflow/workflow_init_error_handler.py
TeddyCr: Make sense either way. We can keep it in the class if we moved them from the module 😊

ingestion/src/metadata/ingestion/ometa/client_utils.py
IceS2: It is! https://docs.basedpyright.com/#/comments?id=line-level-diagnostic-suppression

TeddyCr (approved):

PR #17150, opened 2024-07-23
Fixes #17036: Ignore Enter keydown events for IME operation on FeedEditor

### LINKED ISSUE #17036: UI issue for Chinese/Japanese/Korean (CJK) text input
**Affected module**
UI

**Describe the bug**
Some (or all) text editors on the UI improperly handle keydown events on the Enter key for Chinese, Japanese, and Korean text input, causing unintended messages to be sent.
- In English text, the keydown event on the Enter key is almost obviously intended to send a message. This is the currently implemented behavior.
- In CJK text, the keydown event of the enter key does not always indicate the intent of the text sending. This is due to the behavior of software called [IME](https://en.wikipedia.org/wiki/Input_method).
  - It is necessary to distinguish between the Enter key input events caused by IEM operations and the events intended to send messages.

This behavior is uncomfortable for non-English users of OpenMetadata.

**To Reproduce**
It can be reproduced by entering CJK text into the text editor of any of the conversations.
Video: https://github.com/user-attachments/assets/2f13598e-a282-4184-9a0e-a4d5154a3845

**Expected behavior**
- For CJK text,
  - the message are *not* sent on the keydown event of Enter key due to IME operation.
  - the message are send on the keydown event of Enter key outside the IME operation.

**Version:**
 - OS: macOS
 - Python version: unknown
 - OpenMetadata version: 1.4.4
 - OpenMetadata Ingestion package version: unknown

**Additional context**

Debug:

The following table shows the event.key and event.keyCode values of the KeyDown event object during input the English or CJK (in this example, Japanese) text into the UI editor.
|index|English text `hello`|Japanese text `こんにちは`|
|-|-|-|
|0|key='h', keyCode=72|key='k', keyCode=229|
|1|key='e', keyCode=69|key='o', keyCode=229|
|2|key='l', keyCode=76|key='n', keyCode=229|
|3|key='l', keyCode=76|key='n', keyCode=229|
|4|key='o', keyCode=79|key='i', keyCode=229|
|5|key='Enter', keyCode=13 (*1)|key='t', keyCode=229|
|6|-|key='i', keyCode=229|
|7|-|key='h', keyCode=229|
|8|-|key='a', keyCode=229|
|9|-|key='Enter', keyCode=229 (*2)|

We observe that:
- (*1) For English text, the KeyDown event with { key='Enter', keyCode=13 } indents to send the message.
- For CJK text, all the KeyDown events has [keyCode=229 (this indicates IME operation)](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#keydown_events_with_ime).
- (*2) The event with { key='Enter', keyCode=229 } was caused by IME operation. At this time, the user does *not* intend to send a message.

Note that these experimental results were taken in Chrome (other browsers may behave differently).

Similar discussions:
- https://stackoverflow.com/q/74639605
- https://stackoverflow.com/q/7087858

Solution (?)
I have found this problem directly here.
https://github.com/open-metadata/OpenMetadata/blob/c2db1bbf0ff8e9e93741f528930d684ba077be0e/openmetadata-ui/src/main/resources/ui/src/components/ActivityFeed/FeedEditor/FeedEditor.tsx#L238-L247
I think the following changes would solve the problem, for example.
```diff
       // This logic will handle Enter key binding
-      if (e.key === 'Enter' && !e.shiftKey && !isMentionListOpen) {
+      if (e.key === 'Enter' && e.keyCode === 13 && !e.shiftKey && !isMentionListOpen) {
```
There are probably other components (event handlers) that should be modified like this.

fuzmish:
### Describe your changes:
Fixes #17036

- What changes did you make?
  - Ignore Enter keydown events caused by IME operations during CJK text input on KeyDown event handler.
    - Since [SyntheticKeyboardEvent does not have .isComposing property](https://github.com/facebook/react/issues/13104), I have used `nativeEvent` here.
- Why did you make them?
  - See #17036
  - I have added test cases to verify the issue is solved by this PR.
  - Also, I have tested the actual UI by running the server locally.

#
### Type of change:
- [x] Bug fix

#
### Checklist:


Sachin-chaurasiya (approved):
Great work @fuzmish  🙌🏻,
Thanks for the contributions ✨

PR #14087, opened 2023-11-23
feat: add support for doris datasource

chyueyi:
### Describe your changes:

feat <issue-12344>: add support for doris datasource

Apache doris is a very popular mpp database, we use doris as our datawarehouse,so I worked on doris datasource ingestion.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [x] My PR title is `Fixes <issue-number>: <short explanation>`
- [x] I have commented on my code, particularly in hard-to-understand areas. 
- [x] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

PR #17166, opened 2024-07-24
feat: Add support for exasol datasource

Nicoretti:
Hi @ all :wave:,

I have converted this PR from WIP to an actual PR. The basic SQLAlchemy-based metadata ingestion is now working and has been tested via CLI. Additionally, I have confirmed that the UI part should function correctly, though I have not tested the ingestion via UI yet, as I am not familiar with Airflow. I aimed to keep the initial PR for the ingestion plugin minimal. I plan to create additional PRs, such as one for lineage, in the future once we have the basic plugin merged.

If you require a basic set of unit/integration tests, it would be helpful to have some guidance on what kind of tests you typically expect to be included.

I mostly oriented myself on [this](https://github.com/open-metadata/OpenMetadata/pull/14087/files#diff-e3cff1f62f27d230c87adf51968f3e6eba58961b76e8d0cb9d4d1e823ff22165) PR.

### Describe your changes:

feat : Add support for exasol datasource

Exasol DB is a high-performance, in-memory, massively parallel processing (MPP) database specifically designed for advanced analytics and business intelligence applications.

### Type of change:
- [X] New feature
#
### Checklist:
- [ ] The issue properly describes why the new feature is needed, what's the goal, and how we are building it. Any discussion
- [x] I have updated the documentation.


ulixius9:
>Question: For the final PR, should the branch history be cleaned up so it can be rebased, or will it be squash merged anyway?

Hey @Nicoretti thanks for the pr, it will get squash merged so dont worry about the branch history

exasol-notes.md
```diff
@@ -0,0 +1,20 @@
...
+```shell
+nix develop
+```
+
+Regarding Python, you are free to choose whatever fits your needs best. During development, Rye was used for providing and managing the Python version
+(`.python-version`). For development itself, a virtual Python environment was used. In any case, make sure you have an appropriate Python environment
+and have activated it so the path, etc., are set correctly, e.g., via:
+
+```shell
+source .venv/bin/activate.fish
```
pmbrull, line 19: hi @Nicoretti thanks for the PR.

While you keep making progress on the connector itself, a couple of comments if I may:
1. we can add any specific developer notes [here](https://github.com/open-metadata/OpenMetadata/tree/main/openmetadata-docs/content/v1.5.x-SNAPSHOT/developers/contribute). Docs are md files in this same repo, so any specific consideration needed for exasol we can add it there
2. I'd avoid introducing - at least in this same PR - athe `flake.nix` and `flake.lock` or any new env management tool. This is something that would live in the whole project, and might require a different/extended discussion.

Thanks

Nicoretti: Hi @pmbrull,

Feel free to provide any feedback at any time, it is highly appreciated!

Sorry, it seems I haven't communicated my intentions well enough in the PR description,
so let me try to clarify your points from my perspective.

1. Thanks for pointing me to the exact location. I'll make sure the final and official documentation for the connector will end up there. For now, these are merely some notes for me or my colleagues, in case they need to pick up on the work. (Relevant pieces will be migrated to the place you mentioned, and irrelevant pieces will be purged for the "final" PR)

2. I wouldn't dare to just introduce a new complex tool out of the blue, especially `nix`, due to my kind of "love & hate" relationship with it :sweat_smile:. That said, it helps me keep my system clean and bootstrap a reproducible setup quickly. Due to the nature of `nix` flakes, they need to be in the git index to work (so I committed it for now). I'll make sure that once I remove the draft status of the PR, the `nix` files won't be part of it anymore.

:spiral_notepad:  Side Note: August is quite packed for me; therefore, it is likely that getting the PR out of draft status will happen earliest in September.

Hope that helps to clarify the details a bit better.

Best,
Nico

ingestion/setup.py
```diff
@@ -224,6 +224,7 @@
     "elasticsearch": {
         VERSIONS["elasticsearch8"],
     },  # also requires requests-aws4auth which is in base
+    "exasol": {"sqlalchemy_exasol>=5"},
```
pmbrull, line 227: not sure if we could pin this as `sqlalchemy_exasol~=5.0` to avoid it automatically going into 6.X and things breaking down.

ingestion/src/metadata/ingestion/source/database/exasol/connection.py
```diff
@@ -0,0 +1,87 @@
...
+    get_connection_args_common,
+)
+from metadata.ingestion.connections.test_connections import test_query
+from metadata.ingestion.ometa.ometa_api import OpenMetadata
+from metadata.utils.logger import ingestion_logger
+
+logger = ingestion_logger()
+
+
+def get_connection_url(connection: ExasolConnection) -> str:
```
pmbrull, line 24: we can add a test similar to https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/tests/unit/test_source_connection.py#L1098

ingestion/src/metadata/ingestion/source/database/exasol/connection.py
```diff
@@ -0,0 +1,87 @@
...
+    metadata: OpenMetadata,
+    engine: Engine,
+    service_connection: ExasolConnection,
+    automation_workflow: Optional[AutomationWorkflow] = None,
+) -> None:
+    """
+    Test connection. This can be executed either as part
+    of a metadata workflow or during an Automation Workflow
+    """
+    test_query(engine, "SELECT 1357;")
```
pmbrull, line 87: Curiosity: why 1357?

pmbrull (commented):
Hi @Nicoretti,

Appreciate the time here, great PR!

Left a quick comment. LMK if you need help.

Thanks!

ingestion/setup.py
Nicoretti: Good point! If it’s alright with you, I’d use `"sqlalchemy_exasol>=5,<6"` as constraint.

This approach is more flexible, particularly in scenarios where multiple major versions are compatible as dependency. I’m already aware that a major version change in `sqlalchemy_exasol` is on the horizon, which will be compatible for all users of the websocket-based sqla driver. However, this transition will necessitate a major version upgrade due to the breaking compatibility with legacy drivers.

ingestion/src/metadata/ingestion/source/database/exasol/connection.py
Nicoretti: **TL;DR: Leftover, I will adjust it to `SELECT 1;`.**

Details:
For basic connectivity tests, I usually use `SELECT 1;`. While this is a common approach, identifying my query on the DB side (e.g., audit log) can be tedious. So initially, I often go with `SELECT {RANDOM-NUMBER};`

@pmbrull Should I adjust it to `SELECT 1;` or should it be a more complex check as in the other ingestion plugins?

Side Note: I decided against a complex check because, from my point of view, what other plugins do goes far beyond just a simple connectivity check.

ingestion/src/metadata/ingestion/source/database/exasol/connection.py
Nicoretti: Thanks for the link! I'll add a similar test.

Do you have any suggestions or requirements regarding integration tests?

ingestion/src/metadata/ingestion/source/database/exasol/connection.py
pmbrull: hi thanks, agree. We stick to `SELECT 1` for other connectors. Even a `SELECT 42` might be standard (since it's a known joke) but these magic numbers might raise eyebrows. Thank you!

ingestion/src/metadata/ingestion/source/database/exasol/connection.py
pmbrull: Integration test should not be needed, since most of the connector is based on the "core" implementation which is covered already. thanks

pmbrull (approved):
Thanks, looks great!

pmbrull (approved):

PR #17216, opened 2024-07-29
Fixes #17221: Setuptools Version Fix, Resolve Ingestion Setup Failure

### LINKED ISSUE #17221: [BUG] Ingestion Packages and CI Pipelines breaking due to latest setuptools breaking change
**Affected module**
Ingestion

**Describe the bug**
The issue on GitHub (pypa/setuptools#4519) highlights that recent changes in setuptools (version 72.0.0) removed the test command, causing many packages to become un-installable during the default build process. This problem affects continuous integration (CI) pipelines and ingestion setups that do not pin a specific setuptools version. Suggested workarounds include downgrading setuptools which didn't work as there are many internal packages that use different versions of setuptools , using --no-build-isolation didn't work as we require extensive number of packages that we have not mentioned,

Workaround we are going with: 
- Adding the contraints file

For more details, visit the issue page: [GitHub Issue #4519](https://github.com/pypa/setuptools/issues/4519).

**To Reproduce**
Starting Ingestion with Setuptool>=72

**Expected behavior**
To work smoothly without breaking because of setuptools error

ayush-shah:
### Describe your changes:

Fixes #17221  Setuptools Issue

ref: https://github.com/pypa/setuptools/issues/4519

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/Makefile
```diff
@@ -11,23 +11,23 @@ endif
...
 
 .PHONY: install_all
 install_all:  ## Install the ingestion module with all dependencies
-	python -m pip install "$(INGESTION_DIR)[all]/"
+	PIP_CONSTRAINT=$(INGESTION_DIR)/constraints.txt python -m pip install "$(INGESTION_DIR)[all]/"
 
 .PHONY: install_apis
 install_apis:  ## Install the REST APIs module to the current environment
-	python -m pip install $(ROOT_DIR)/openmetadata-airflow-apis/ setuptools==69.0.2
+	PIP_CONSTRAINT=$(ROOT_DIR)/openmetadata-airflow-apis/constraints.txt python -m pip install $(ROOT_DIR)/openmetadata-airflow-apis/ setuptools==69.0.2
```
pmbrull, line 30: why are we adding these constraints here? Can't we handle it from the setup file?

pmbrull: or in pyproject.toml I think we have a requires https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/pyproject.toml#L2

I'd like to avoid having to look at versions to yet another place

pmbrull (changes requested):

ingestion/Makefile
ayush-shah: tried different solutions here, pyproject.toml setuptools version modification didn't work as expected, we already have that in place and it is still failing, the reach for requires doesn't impact 3rd party packages as far as i could test for the setuptools issue

ayush-shah: only solution that could impact 3rd party packages was the constraint one

ayush-shah: There is one more solution that i have yet to try i.e. --no-build-isolation

pmbrull (approved):

ingestion/Makefile
pmbrull: ok thanks, let's keep it as-is until setuptools fixes it on their end

PR #17254, opened 2024-07-31
Fetch templates for collate and oss.

Siddhanttimeline:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


mohityadav766:
It's ok for now , but we should come up with something for all entities

mohityadav766 (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -440,6 +443,31 @@ public final void initSeedDataFromResources() throws IOException {
...
 
   public final List<T> getEntitiesFromSeedData() throws IOException {
+    List<T> entitiesFromSeedData = new ArrayList<>();
+
+    if (entityType.equals(Entity.DOCUMENT)) {
+      SmtpSettings emailConfig =
+          SettingsCache.getSetting(SettingsType.EMAIL_CONFIGURATION, SmtpSettings.class);
+
+      switch (emailConfig.getTemplates()) {
+        case COLLATE -> {
```
pmbrull, line 453: instead of having this on a collate/OM toggle with the Enum, can't we specify a path or something as a more generic approach? This should give us some room to iterate on this and eventually allow users to modify templates from the UI

pmbrull (changes requested):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
pmbrull: I mean, we can keep the email templates in the json data dir under `openmetadata`, but the config could be a path we use to load the base resources rather than a hardcoded path with either 'collate' or 'openmetadata'

Siddhanttimeline: To allow users to edit the template, we have moved the email templates from .ftl to jsons and is saved in db(doc_store). https://github.com/open-metadata/OpenMetadata/pull/17015
We are reading the path from YAML to choose collate/OM emailTemplates jsons. Email YAML config is saved in the openmetadata_settings once and from there we are fetching the path.

pmbrull (approved):

PR #17282, opened 2024-08-02
Fix #16573: get table owners for databaricks & unitycatalog tables

### LINKED ISSUE #16573: Support ownership ingestion for Databricks
Currently tables ingested from Databricks do not get owners assigned, even though they exist in the source.

I would like both the [Databricks](https://docs.open-metadata.org/v1.4.x/connectors/database/databricks) and [Unity Catalog](https://docs.open-metadata.org/v1.4.x/connectors/database/unity-catalog) database connectors to support ingesting ownership from Databricks. This should support both individual users and groups as owners.

Alternatively we could manually set owners, but due to the volume of tables this is infeasible

harshsoni2024:
### Describe your changes:

Fix #16573 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
@@ -15,6 +15,7 @@ import traceback
 from copy import deepcopy
 from typing import Iterable, Optional, Tuple, Union
 
+from pydantic import EmailStr
 from pyhive.sqlalchemy_hive import _type_map
 from sqlalchemy import types, util
 from sqlalchemy.engine import reflection
@@ -51,7 +52,6 @@ from metadata.ingestion.source.database.databricks.queries import (
     DATABRICKS_GET_SCHEMA_TAGS,
     DATABRICKS_GET_TABLE_COMMENTS,
     DATABRICKS_GET_TABLE_TAGS,
-    DATABRICKS_TABLE_DESCRIBE,
     DATABRICKS_VIEW_DEFINITIONS,
 )
 from metadata.ingestion.source.database.external_table_lineage_mixin import (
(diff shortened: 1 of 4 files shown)


ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
```diff
@@ -661,3 +663,38 @@ def get_table_description(
...
+        query = DATABRICKS_TABLE_DESCRIBE.format(
+            catalog_name=self.context.get().database,
+            schema_name=self.context.get().database_schema,
+            table_name=table_name,
+        )
+        result = self.connection.engine.execute(query)
+        for row in result:
+            row_dict = dict(row)
+            if row_dict.get("col_name") == "Owner":
+                owner = row_dict.get("data_type")
```
SumanMaharana, line 691: Just to confirm are we getting owner detail under `data_type` key

ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
```diff
@@ -661,3 +663,38 @@ def get_table_description(
...
+
+    def _filter_owner_name(self, owner_name: str) -> str:
+        """remove unnecessary keyword from name"""
+        pattern = r"\(Unknown\)"
+        filtered_name = re.sub(pattern, "", owner_name).strip()
+        return filtered_name
+
+    def _check_if_email(self, email_id: str) -> bool:
+        """check if email string is valid"""
+        email_pattern = r"^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$"
```
pmbrull, line 675: do we need to do this here? Seems like a more generic check rather than keeping it only for databricks. How are we solving it for other systems such as postgres where usernames are not necessarily emails?

pmbrull (changes requested):

ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
harshsoni2024: In the databricks, we get owner name in both name/email type. So that I have applied here to filter out based on email first and then name.
While in other db/dashboard connection we have confirm that we get owner's email(from api/inspector method) and we're passing directly to get reference.
[domodb](https://github.com/open-metadata/OpenMetadata/blob/b84ce33b80cb43e885540a928f439a75771acbf6/ingestion/src/metadata/ingestion/source/database/domodatabase/metadata.py#L168), [iceberg](https://github.com/open-metadata/OpenMetadata/blob/373cafcda26b17108d0897c54773df97e271447e/ingestion/src/metadata/ingestion/source/database/iceberg/metadata.py#L231), [metabase](https://github.com/open-metadata/OpenMetadata/blob/c84baaf66a23da75f54d126b807fca2d2682b9e1/ingestion/src/metadata/ingestion/source/dashboard/metabase/metadata.py#L154).
I think we can put this method in `DatabaseServiceSource` or `DashboardServiceSource` & check with valid email whenever we're not sure. WDYT ?

pmbrull: - let's put this method in a common place. maybe https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/utils/helpers.py? I don't love it but we don't have a better structure atm.
- let's try to use pydantic [emailStr](https://docs.pydantic.dev/1.10/usage/types/#:~:text=requires%20email%2Dvalidator%20to%20be%20installed%3B%20the%20input%20string%20must%20be%20a%20valid%20email%20address%2C%20and%20the%20output%20is%20a%20simple%20string) instead of regex

ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
```diff
@@ -661,3 +663,41 @@ def get_table_description(
...
+                    break
+            if not owner:
+                return
+
+            owner = self._filter_owner_name(owner)
+            owner_ref = None
+            try:
+                owner_email = EmailStr._validate(owner)
+                owner_ref = self.metadata.get_reference_by_email(email=owner_email)
+            except Exception:
```
pmbrull, line 697: 1. we dont need to catch ANY exception here no? only the one triggered by wrong strings inside `EmailStr`
2. also, I don't think we need to run `._validate`. You might just be able to do `EmailStr(<string>)` and see if it blows up or not

harshsoni2024: 1. in case of different invalid emails it shows different error(Valueerror, SyntaxError) so I thought to apply general exception.
2. I checked with EmailStr(string) but it's not working in that way (maybe with v2 it's changed)
[change](https://github.com/pydantic/pydantic/issues/798#issuecomment-1992089526)

pmbrull (approved):

PR #17374, opened 2024-08-10
Minor: fixed params calculation for aborted test

ShaileshParmar11:
Minor: fixed params calculation for aborted test

Before:
image

After:
image

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/utils/DataQuality/TestSummaryGraphUtils.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/DataQuality/TestSummaryGraphUtils.ts
@@ -87,9 +87,14 @@ export const prepareChartData = ({
 
   dataPoints.reverse();
 
+  // get params from the result
+  const testCaseResultParams = testCaseResults.find(
+    (result) => result.testResultValue?.length
+  );
+
   return {
     information:
-      testCaseResults[0]?.testResultValue?.map((info, i) => ({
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/src/utils/DataQuality/TestSummaryGraphUtils.ts
```diff
@@ -87,9 +87,14 @@ export const prepareChartData = ({
 
   dataPoints.reverse();
 
+  // get params from the result
+  const testCaseResultParams = testCaseResults.find(
+    (result) => result.testResultValue?.length
+  );
+
```
Sachin-chaurasiya, line 94: @ShaileshParmar11 , this will return the first test case result that has a non-empty testResultValue. so if we have
```js
const testCaseResults = [
  { testResultValue: null },
  { testResultValue: [] },
  { testResultValue: [1, 2, 3] },
  { testResultValue: [4, 5] }
]; 
```

Then it will return the 3rd object and ignore the 4th one, is that intentional?

ShaileshParmar11: all the values in the test result will have the same parameter except when the test is aborted! so either it will be empty or have same length of value!

Sachin-chaurasiya (approved):

PR #17239, opened 2024-07-31
Fix #16434: SQL server date format issue

harshsoni2024:
### Describe your changes:

Fix #16434 

I worked on mssql date format issue when using different language

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

PR #17380, opened 2024-08-10
MINOR: Fix Datetime Conversion issue in usage/lineage

SumanMaharana:
### Describe your changes:

Fix Datetime Conversion issue in usage/lineage

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/database/mssql/usage.py
```diff
@@ -37,32 +25,3 @@ class MssqlUsageSource(MssqlQueryParserSource, UsageSource):
         AND lower(t.text) NOT LIKE '%%create%%function%%'
         AND lower(t.text) NOT LIKE '%%declare%%'
         """
```
harshsoni2024, line 27: this `get_sqlalchemy_engine_dateformat` would be useful incase of server in different language.
these were recent PRs we did regarding the same.
[lineage](https://github.com/open-metadata/OpenMetadata/pull/17239/files)
[usage](https://github.com/open-metadata/OpenMetadata/pull/17341/files)
removing timezone form datetime object is alright but we need to convert same string in appropriate format.

ayush-shah (commented):
Can we take care of lineage.py having similar changes? Thanks

ingestion/src/metadata/ingestion/source/database/mssql/lineage.py
```diff
@@ -26,10 +26,6 @@
 
 
 class MssqlLineageSource(MssqlQueryParserSource, LineageSource):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.start = self.start.replace(tzinfo=None)
-        self.end = self.end.replace(tzinfo=None)
```
SumanMaharana, line 32: `.replace(tzinfo=None)` is handled in `helpers.py` itself so we don't need it here

ingestion/src/metadata/utils/helpers.py
```diff
@@ -123,7 +123,7 @@ def get_start_and_end(duration: int = 0) -> Tuple[datetime, datetime]:
     Method to return start and end time based on duration
     """
 
-    today = datetime.now(timezone.utc)
+    today = datetime.now(timezone.utc).replace(tzinfo=None)
```
pmbrull, line 126: why pass the timezone and then remove it instead of `datetime.now()` without any specification?

SumanMaharana: just using `datetime.now()` will give the local time which can be different from the UTC time. to keep it standard we pass the timezone as `utc`

pmbrull (approved):

PR #17411, opened 2024-08-12
Minor: Kill active/idle connections after test connections

SumanMaharana:
### Describe your changes:

Minor: Kill active/idle connections after test connections

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/connections/test_connections.py
```diff
@@ -329,6 +329,8 @@ def test_connection_db_common(
         timeout_seconds=timeout_seconds,
     )
 
+    kill_active_connections(engine)
```
pmbrull, line 332: also this works for the db common, but should we call this in other services as well?

pmbrull (commented):
Let's review the idle connections in a service such as Postgres with multiple db ingestions and see if we are missing something open anywhere else

ingestion/src/metadata/ingestion/connections/test_connections.py
SumanMaharana: added for multi db as well

ingestion/src/metadata/ingestion/source/connections.py
```diff
@@ -47,3 +48,13 @@ def get_connection(connection: BaseModel) -> Any:
...
     """
     return get_connection_fn(connection)(connection)
+
+
+def kill_active_connections(engine: Engine):
+    """
+    Method to kill the active connections
+    as well as idle connections in the engine
+    """
+    active_conn = engine.pool.checkedout() + engine.pool.checkedin()
```
pmbrull, line 58: ok now maybe silly comment, but we're now passing this at the `set_inspector` stage.

Is it possible this will ever fail? Should we wrap this on a try/catch and log if something goes wrong? We don't want this cleanup stage to blow up the full ingestion

SumanMaharana: Will add a try catch block her

SumanMaharana: added

pmbrull (approved):

PR #17477, opened 2024-08-18
Updated pagination logic  to use  both name and id as cursors in listBefore and listAfter methods

sonika-shah:
### Describe your changes:

- updates the listBefore and listAfter methods to use both name and id (in a json object form) as cursor for pagination instead of only name.
- pagination changes have only been done where name was used as cursor, and not in methods where different column were used as cursor
```
{
  "name": "name value of Entity",
  "id": "id value of Entity"
}
```

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityDAO.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityDAO.java
@@ -13,7 +13,6 @@
 
 package org.openmetadata.service.jdbi3;
 
-import static org.openmetadata.common.utils.CommonUtil.nullOrEmpty;
 import static org.openmetadata.service.exception.CatalogExceptionMessage.entityNotFound;
 import static org.openmetadata.service.jdbi3.ListFilter.escape;
 import static org.openmetadata.service.jdbi3.ListFilter.escapeApostrophe;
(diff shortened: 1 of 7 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityDAO.java
```diff
@@ -160,80 +159,26 @@ int listCount(
...
               + "(<table>.name < :beforeName OR (<table>.name = :beforeName AND <table>.id < :beforeId)) "
-              + // Pagination by entity id or name (when entity have same name)
-              "ORDER BY <table>.name DESC, <table>.id  DESC "
-              + // Pagination by entity id or name (when entity have same name)
+              + // Pagination by entity name or id (when entity name same)
+              "ORDER BY <table>.name DESC,<table>.id DESC "
+              + // Pagination by entity name or id (when entity name same)
               "LIMIT :limit"
-              + ") last_rows_subquery ORDER BY name, id",
+              + ") last_rows_subquery ORDER BY name,id",
```
harshach, line 168: is this the right way to do here? what is it going to sort with?

harshach: is it better for us to use offset and limit instead of encrypted before & after?

sonika-shah: Using OFFSET like this—SELECT json FROM <table> <cond> ORDER BY name LIMIT :limit OFFSET :offset—can lead to a full table scan, which may cause performance issues as the dataset grows.

In contrast, using a cursor with > after or < before conditions does not always require a full table scan, offering better performance, especially with large datasets.

sonika-shah: And for the first question , order by name id, will take columns from the inner subquery we dont need to mention it in the outer query

harshach (approved):

PR #17490, opened 2024-08-19
Fixes #17489: Allow non numeric numbers to be sent via Json, Replace NaN value…

### LINKED ISSUE #17489: Profiler: Snowflake fails if it has NaN in the data
**Affected module**
Ingestion

**Describe the bug**
Snowflake fails if it has NaN in the data

**To Reproduce**
Ingest NaN, inf and -inf data in Snowflake db, run OM metadata ingestion, and then run profiler.

**Expected behavior**

It should pass without any errors.

Facing the below error:
metadata.ingestion.ometa.client.APIError: Non-standard token '-Infinity': enable `JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS` to allow

ayush-shah:
…s with None in SQAProfilerInterface

Replace NaN values with None in the SQAProfilerInterface class to maintain database parity. NaN values will be cast to null in OpenMetadata. This change ensures that data handling processes account for this conversion.

### Describe your changes:

Fixes #17489

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
+++ b/ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
@@ -451,16 +451,14 @@ class SQAProfilerInterface(ProfilerInterface, SQAInterfaceMixin):
                     column=metric_func.column,
                     sample=sample,
                 )
-                # Replace NaN values with None
-                logger.warning(
-                    "NaN Data Type Not Supported: NaN values will be cast to null in OpenMetadata"
-                    " to maintain database parity. Ensure your data handling processes account for this conversion."
-                )
-
-                row = {
-                    k: None if isinstance(v, float) and math.isnan(v) else v
-                    for k, v in dict(row).items()
-                }
+                if row:
+                    for k, v in row.items():
+                        # Replace NaN values with None
+                        if isinstance(v, float) and math.isnan(v):
+                            logger.warning(
+                                "NaN data detected and will be cast to null in OpenMetadata to maintain database parity"
+                            )
+                            row[k] = None
             except Exception as exc:
                 error = (
                     f"{metric_func.column if metric_func.column is not None else metric_func.table.__tablename__} "
--- a/ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
+++ b/ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
@@ -16,6 +16,7 @@ supporting sqlalchemy abstraction layer
 """
 
 import concurrent.futures
+import math
 import threading
 import traceback
 from collections import defaultdict
@@ -231,6 +232,13 @@ class SQAProfilerInterface(ProfilerInterface, SQAInterfaceMixin):
                     if not metric.is_window_metric()
                 ],
             )
+            # Replace NaN values with None
+            logger.warn(
+                "NaN Data Type Not Supported: NaN values will be cast to null in OpenMetadata"
+                " to maintain database parity. Ensure your data handling processes account for this conversion."
+            )
+
+            row = {k: None if math.isnan(v) else v for k, v in dict(row).items()}
             return dict(row)
         except (ProgrammingError, DBAPIError) as exc:
             return self._programming_error_static_metric(
--- a/ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
+++ b/ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
@@ -232,13 +232,6 @@ class SQAProfilerInterface(ProfilerInterface, SQAInterfaceMixin):
                     if not metric.is_window_metric()
                 ],
             )
-            # Replace NaN values with None
-            logger.warn(
-                "NaN Data Type Not Supported: NaN values will be cast to null in OpenMetadata"
-                " to maintain database parity. Ensure your data handling processes account for this conversion."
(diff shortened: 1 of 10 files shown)


ingestion/src/metadata/profiler/metrics/hybrid/histogram.py
```diff
@@ -119,7 +119,13 @@ def _get_bins(
...
             # freedman-diaconis rule
             bin_width = self._get_bin_width(float(res_iqr), res_row_count)  # type: ignore
-            num_bins = math.ceil((res_max - res_min) / bin_width)  # type: ignore
+            try:
+                num_bins = math.ceil((res_max - res_min) / bin_width)  # type: ignore
+            except OverflowError as warning:
+                logger.warn(
+                    f"An warning occurred while computing the number of bins: {warning}"
+                )
+                num_bins = max_bin_count
```
TeddyCr, line 128: Actually looking at this piece of code we won't be able to do anything if min/max is inf as we need it to compute the width of the bins. For now we can skip the histogram if we fall into this.

openmetadata-service/src/main/java/org/openmetadata/service/JsonPatchProvider.java
```diff
@@ -26,6 +27,9 @@ public ObjectMapper getContext(Class<?> type) {
     final ObjectMapper mapper = new ObjectMapper();
     mapper.registerModule(new JSR353Module());
     mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+    // This feature allows the parser to accept non-numeric numbers such as NaN,
+    // Infinity, and -Infinity in JSON input.
```
TeddyCr, line 31: can you add a link to this wiki https://github.com/FasterXML/jackson-core/wiki/JsonReadFeatures please?

ingestion/src/metadata/profiler/metrics/hybrid/histogram.py
```diff
@@ -125,7 +125,7 @@ def _get_bins(
                 logger.warn(
                     f"An warning occurred while computing the number of bins: {warning}"
                 )
-                num_bins = max_bin_count
+                return 0, bin_width
```
TeddyCr, line 128: what i meant here is that if min or max are inf then let's skip the whole thing entirely. Let's not add that try/except to handle this

ingestion/src/metadata/profiler/metrics/composed/null_ratio.py
```diff
@@ -52,6 +52,7 @@ def fn(self, res: Dict[str, Any]) -> Optional[float]:
 
         count = res.get(Count.name())
         null_count = res.get(NullCount.name())
-        if count + null_count == 0:
+        # Return None if count or null_count is None, or if their sum is zero
+        if count is None or null_count is None or count + null_count == 0:
```
sushi30, line 56: this should never be null. if it is then the method should fail and not hide it with `None`

ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
```diff
@@ -231,6 +232,13 @@ def _compute_static_metrics(
                     if not metric.is_window_metric()
                 ],
             )
+            # Replace NaN values with None
+            logger.warn(
+                "NaN Data Type Not Supported: NaN values will be cast to null in OpenMetadata"
+                " to maintain database parity. Ensure your data handling processes account for this conversion."
+            )
+
+            row = {k: None if math.isnan(v) else v for k, v in dict(row).items()}
```
sushi30, line 241: Can this be abstracted to modify any `nan` value submitted in other metrics as well? Maybe this can be applied at a higher level.

sushi30 (changes requested):

ingestion/src/metadata/profiler/metrics/hybrid/histogram.py
ayush-shah: Got it, Returning 0 returns None at the end which skips the histogram metric indirectly.

ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
ayush-shah: I will create a decorator that could allow to attach to metric functions ( static, composed, hybrid, etc )

ingestion/src/metadata/profiler/metrics/composed/null_ratio.py
ayush-shah: I will check where the issue is, it shouldn't return None, it should return 0..not sure if the interpretation within sqlalchemy is returning it 🤔

ingestion/src/metadata/profiler/metrics/hybrid/histogram.py
ayush-shah: I will modify the logic to skip the histogram entirely if min / max are `inf / -inf`

ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
ayush-shah: found a better place to add it, at the core level that takes care of the same

sushi30 (dismissed):

sushi30 (approved):

PR #17517, opened 2024-08-21
[WIP] Fixes <issue to open>: optimizing table profiler process for BQ

felipebmendes:
### Describe your changes:

Fixes <issue-number>

Context:
During the profile ingestion process, a separate query is executed for each table in the BigQuery project to retrieve metrics like rowCount, sizeInBytes, and columnCount. This approach leads to significant performance issues due to constraints like concurrent job limits and available slots in BigQuery.

Current Behavior:
The profile ingestion process retrieves table-level metrics by executing a query for each table individually. While the queries already leverage the __TABLES__ table, this method is inefficient for projects with many tables.

Proposed Solution:
We propose modifying the profile ingestion process to execute a single query on the __TABLES__ table that retrieves the required metrics for all tables in a project or schema. This would involve caching the results of this query and reading the data from the cache during the subsequent table iteration, rather than fetching metrics for each table individually.

#
### Type of change:

- [ ] Bug fix
- [X] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
+++ b/ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
@@ -16,7 +16,6 @@ supporting sqlalchemy abstraction layer
 """
 
 import concurrent.futures
-import math
 import threading
 import traceback
 from collections import defaultdict
@@ -75,6 +74,7 @@ class SQAProfilerInterface(ProfilerInterface, SQAInterfaceMixin):
         self,
         service_connection_config,
         ometa_client,
+        cache,
         entity,
         storage_config,
         profile_sample_config,
@@ -92,6 +92,7 @@ class SQAProfilerInterface(ProfilerInterface, SQAInterfaceMixin):
         super().__init__(
             service_connection_config,
             ometa_client,
+            cache,
             entity,
             storage_config,
             profile_sample_config,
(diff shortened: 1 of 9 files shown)


ingestion/src/metadata/profiler/orm/functions/table_metric_computer.py
```diff
@@ -309,29 +340,25 @@ def table_storage(self):
             where_clause,
         )
 
-        res = self.runner._session.execute(query).first()
+        res = self.runner._session.execute(query).mappings().all()
```
sushi30, line 343: BigQuery can have virtually infinite tables so there is an edge case where the process hosting the ingestion cannot contain this lise. Therefore this needs to be bound in some form. I can suggest some filtering strategies:

1. Grabbing a window "around" the specific table using "order by" and "between".
2. Using hash based filters.

In any case the cache should be by table id and not for the full `__TABLES__` object so that individual tables can be inserted and evicted based on usage.

sushi30 (changes requested):

ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
```diff
@@ -451,14 +453,6 @@ def compute_metrics_in_thread(
...
                     sample=sample,
                 )
-                if row:
-                    for k, v in row.items():
-                        # Replace NaN values with None
-                        if isinstance(v, float) and math.isnan(v):
-                            logger.warning(
-                                "NaN data detected and will be cast to null in OpenMetadata to maintain database parity"
-                            )
-                            row[k] = None
```
ayush-shah, line 461: Any reason we are removing this? 🤔

AntoineGlacet:
Is this stale?

pmbrull:
> Is this stale?

Looks like so. I'll be closing the PR. @AntoineGlacet feel free to pick it up

PR #17560, opened 2024-08-22
MINOR: Fix Percentage Difference Calculation and Tier Summary Calculation.

IceS2:
On the Percentage and Tier charts we were calculating the percentage difference wrongly. In order to calculate the true aggregated percentage difference we need to know the absolute values.

This PR aims to fix that by adding a work around with what we currently have, even though it should be improved.

---

It also fixes the Tier Sumary Calculation by moving the `NoTier` filter to the query itself instead of the filters.
This allows us to properly count all the documents for the division. Otherwise we'd always get 100%.

#
### Type of change:

- [x] Bug fix

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/components/DataInsight/DataInsightChartCard.tsx
```diff
@@ -372,7 +538,10 @@ export const DataInsightChartCard = ({
                           activeKeys.length ? activeKeys.includes(entity) : true
                         }
                         label={`${round(latestData[entity] ?? 0, 2)}${
-                          isPercentageGraph ? '%' : ''
+                          isPercentageGraph &&
+                          type !== SystemChartType.TotalDataAssetsByTier
+                            ? '%'
+                            : ''
```
IceS2, line 544: Although the Summary for Tier is a Percentage, the Numbers on the chart are absolute values.
Added this condition to fix it for now.

Should be improved!

openmetadata-ui/src/main/resources/ui/src/components/DataInsight/DataInsightSummary.tsx
```diff
@@ -202,12 +202,7 @@ const DataInsightSummary: FC<Props> = ({ chartFilter, onScrollToChart }) => {
                   total={0}
                   value={`${summary.latest}
                     ${
-                      [
-                        SystemChartType.TotalDataAssets,
-                        SystemChartType.TotalDataAssetsByTier,
-                      ].includes(summary.id)
-                        ? ''
-                        : '%'
+                      summary.id === SystemChartType.TotalDataAssets ? '' : '%'
```
IceS2, line 205: Since Summary for Tier is a percentage, I removed it from the condition.

harshach (approved):

PR #17614, opened 2024-08-28
Delete tags from search while deleting classification

sonika-shah:
### Describe your changes:

While deleting classification , we were able to still get the tags in search, this pr updates the childAlias for classification index

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/resources/elasticsearch/indexMapping.json
+++ b/openmetadata-service/src/main/resources/elasticsearch/indexMapping.json
@@ -144,7 +144,7 @@
     "indexMappingFile": "/elasticsearch/%s/classification_index_mapping.json",
     "alias": "classification",
     "parentAliases": ["all"],
-    "childAliases": []
+    "childAliases": ["tag"]
   },
   "user": {
     "indexName": "user_search_index",


openmetadata-service/src/main/java/org/openmetadata/service/search/SearchClient.java
```diff
@@ -68,7 +68,8 @@ public interface SearchClient {
   String SOFT_DELETE_RESTORE_SCRIPT = "ctx._source.put('deleted', '%s')";
   String REMOVE_TAGS_CHILDREN_SCRIPT =
       "for (int i = 0; i < ctx._source.tags.length; i++) { if (ctx._source.tags[i].tagFQN == params.fqn) { ctx._source.tags.remove(i) }}";
-
+  String REMOVE_TAGS_WITH_PARENT_SCRIPT =
```
harshach, line 71: why are we doing this with painless script. You can do this by searchByQuery and find the docs that has the tag and remove that tag only

sonika-shah: thanks @harshach for the suggestion, have updated the logic using searchByQuery and removed the painless script.

sonika-shah: we are going with painless script here  as in the tags field we have  no other information about the parent other than in the fullyQualifiedName, 
for example here, the only information for parent that we have is from `tagFQN` ,  that classification name is `departments` -
"tags" : [
            {
              "tagFQN" : "departments.marketing",
              "displayName" : "marketing",
              "name" : "marketing",
              "labelType" : "Manual",
              "description" : "marketing",
              "style" : { },
              "source" : "Classification",
              "state" : "Confirmed"
            }
          ],

harshach: @sonika-shah look into updateByQuery, painless is non-performant and adds on top of the results fetched. what you need to do is updatedByquery and fetch all the docs that has the tag thats deleted and remove it.

PR #17683, opened 2024-09-03
Fixes #15565 : Advanced Avro schema recursion depth issue

### LINKED ISSUE #15565: Recursion issue when parsing advanced Avro Schemas
**Affected module**
Ingestion

**Describe the bug**

When parsing more complex avro schema, where we have object 'Item' (from example below) which can have property 'itemList', which is additional list of 'Item' objects, this will cause infinite recursion issue and failure of whole ingestion.

Ideally such schema should be parsed properly however upon RecursionException it would be acceptable to skip it. It is not the case and whole Python process exists with SIGABORT signal - the suspected culprit here is C library used for avro parsing.

**To Reproduce**

```
from metadata.parsers.avro_parser import parse_avro_schema


data = """
{
  "type": "record",
  "name": "RecursionIssue",
  "namespace": "com.issue.recursion",
  "doc": "Schema with recursion issue",
  "fields": [
    {
      "name": "issue",
      "type": {
        "type": "record",
        "name": "Issue",
        "doc": "Global Schema Name",
        "fields": [
          {
            "name": "itemList",
            "default": null,
            "type": [
              "null",
              {
                "type": "array",
                "items": {
                  "type": "record",
                  "name": "Item",
                  "doc": "Item List  - Array of Sub Schema",
                  "fields": [
                    {
                      "name": "itemList",
                      "type": [
                        "null",
                        {
                          "type": "array",
                          "items": "Item"
                        }
                      ],
                      "default": null
                    }
                  ]
                }
              }
            ]
          }
        ]
      }
    }
  ]
}

"""
try:
    print(parse_avro_schema(data))
except RecursionError as e:
    print(e.args)
```

**Expected behavior**
Schema is parsed properly (or Python code doesn't fail with SIGABRT).

**Version:**
 - Python version: 3.9
 - OpenMetadata version: [e.g. 0.8] 1.3.1
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`] 1.3.1

**Additional context**

SumanMaharana:
### Describe your changes:

Fixes #15565 : Advanced Avro schema recursion depth issue

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/parsers/avro_parser.py
+++ b/ingestion/src/metadata/parsers/avro_parser.py
@@ -27,6 +27,7 @@ from metadata.utils.logger import ingestion_logger
 logger = ingestion_logger()
 
 RECORD_DATATYPE_NAME = "RECORD"
+PARSED_ALREADY = dict()
 
 
 def _parse_array_children(
@@ -241,6 +242,7 @@ def parse_avro_schema(
                 description=parsed_schema.doc,
             )
         ]
+        PARSED_ALREADY.clear()
         return models
     except Exception as exc:  # pylint: disable=broad-except
         logger.debug(traceback.format_exc())
@@ -255,6 +257,12 @@ def get_avro_fields(
     Recursively convert the parsed schema into required models
     """
     field_models = []
+
+    if parsed_schema.name in list(PARSED_ALREADY.keys()):
+        if PARSED_ALREADY[parsed_schema.name] == parsed_schema.type:
+            return
+    else:
+        PARSED_ALREADY.update({parsed_schema.name: parsed_schema.type})
 
     for field in parsed_schema.fields:
         try:
--- a/ingestion/tests/unit/test_avro_parser.py
+++ b/ingestion/tests/unit/test_avro_parser.py
@@ -530,6 +530,55 @@ RECORD_INSIDE_RECORD = """
 }
 """
 
+RECURSION_ISSUE_SAMPLE = """
+{
+  "type": "record",
+  "name": "RecursionIssue",
+  "namespace": "com.issue.recursion",
+  "doc": "Schema with recursion issue",
+  "fields": [
+    {
+      "name": "issue",
+      "type": {
+        "type": "record",
+        "name": "Issue",
+        "doc": "Global Schema Name",
+        "fields": [
+          {
+            "name": "itemList",
+            "default": null,
+            "type": [
+              "null",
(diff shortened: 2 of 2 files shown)


ingestion/src/metadata/parsers/avro_parser.py
```diff
@@ -27,6 +27,7 @@
 logger = ingestion_logger()
 
 RECORD_DATATYPE_NAME = "RECORD"
+PARSED_ALREADY = {}
```
sushi30, line 30: since we are using a global variable, add a test that runs this twice to validate there are no side effects.

sushi30 (changes requested):

ingestion/src/metadata/parsers/avro_parser.py
SumanMaharana: I'm doing a `PARSED_ALREADY.clear()` before we return the parsed schema to take care of it.
In the test this scenario has been already taken care of as we have mutiple sample schemas having the same name and type in it.

sushi30: why is it global?

sushi30 (changes requested):

ingestion/src/metadata/parsers/avro_parser.py
SumanMaharana: To parse the avro schema we are having an recursive approach for its fields.
To check if a schema is parsed already its declared out of function.

sushi30: If its a recursive function the correct way is to have the state in the function signature. Something like:

```python
def parse_avro_schema(
    schema: str, cls: Type[BaseModel] = FieldModel, already_parsed
) -> Optional[List[Union[FieldModel, Column]]]:
```

Then start with the initial state:

```python
parse_avro_schema(schema,cls,[])
```

SumanMaharana: yes right. i'll put it in a function signature

sushi30: you can set the the `already_parsed` as optional and initialize it to an empty list if its `None`. That way the first call will initialize the recursion stack.

PR #17697, opened 2024-09-04
minor: disable image upload support in description editor

Sachin-chaurasiya:
**Before**

https://github.com/user-attachments/assets/6049bc7c-063d-48c4-aa2e-dc3e0df23ab7

**After**

https://github.com/user-attachments/assets/b9b321d3-0ebf-40e8-9694-e77fce7b5a38


chirag-madlani (dismissed):

openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
```diff
@@ -1546,6 +1546,7 @@
     "hex-code-placeholder": "Select or enter HEX color code.",
     "hex-color-validation": "Input is not valid HEX code.",
     "hi-user-welcome-to": "Hi {{user}}, Welcome to",
+    "image-upload-error": "Image upload is not supported",
```
pmbrull, line 1549:
```suggestion
    "image-upload-error": "Image upload is not supported. Please use Markdown syntax for images available via URL.",
```

Or something like that? We should do this everywhere

pmbrull (changes requested):

pmbrull (approved):

chirag-madlani (approved):

PR #17706, opened 2024-09-04
GEN-1213 - logic to index test case result

TeddyCr:
### Describe your changes:
- implement indexing for test case results
- clean up IDE warnings
- Implement tests for test case results indexing

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Improvement


bootstrap/sql/migrations/native/1.6.0/mysql/schemaChanges.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE data_quality_data_time_series
+ADD COLUMN id VARCHAR(36) GENERATED ALWAYS AS (json ->> '$.id') STORED NOT NULL,
+ADD CONSTRAINT UNIQUE (id);
```
sonika-shah, line 3: It will be good if we can add index for the new column

TeddyCr: @sonika-shah the table already has an indices on fqn and ts (since most read would be performed filtered on these 2 values).

sonika-shah: @teddy, it will be okay currently . but  later on, if we add any queries based on the ‘id’ column, it would slow down the operation, even if its a simple insert/delete/update sql statement

sonika-shah (approved):

PR #17735, opened 2024-09-05
#16644: support resizable column to customProperty

### LINKED ISSUE #16644: Support flex column for custom property view
On the custom property view of the entity page we should allow user to resize the column width similar to how we do it for the connection page -- see screen recording

https://github.com/open-metadata/OpenMetadata/assets/13626425/052ce272-42f4-4c4d-8810-0788dd01b56b

Ashish8689:
### Describe your changes:

Fixes #16644

- I worked on support resizable column to customProperty
- Added new library to support Table Column re-sizable
- Mock the antd re-sizebale  column library globally

https://github.com/user-attachments/assets/e18ef290-1fe0-4a0f-b87a-623ba671d7a8

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/components/common/Table/table.less
Sachin-chaurasiya, line 1: we already have table.less in `styles/component` , can we use that?

openmetadata-ui/src/main/resources/ui/webpack.config.dev.js
Sachin-chaurasiya, line 1: @Ashish8689 , why do we need this change, and can we update the prod webpack config as well?

Ashish8689: The library was using styles with `.css` which threw loader issue, so to fix that have included in the CSS parser here!

chirag-madlani (approved):

PR #17754, opened 2024-09-06
GEN-309: OpenAPI Connector

harshsoni2024:
### Describe your changes:

Fixes <issue-number>

I worked on OpenAPI Ingestion Connector

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/api/rest/connection.py
```diff
@@ -0,0 +1,68 @@
...
+    service_connection: RESTConnection,
+    automation_workflow: Optional[AutomationWorkflow] = None,
+) -> None:
+    """
+    Test connection. This can be executed either as part
+    of a metadata workflow or during an Automation Workflow
+    """
+
+    def custom_url_exec():
+        if client.status_code == 200:
```
pmbrull, line 55: is this enough? Aren't we expecting any specific shape? I guess that if I put `www.google.com` i might get a 200 but it's not a real service that I can parse no? 🤔

ingestion/src/metadata/ingestion/source/api/rest/metadata.py
```diff
@@ -0,0 +1,238 @@
...
+                continue
+            yield collection
+
+    def yield_api_collection(
+        self, collection: dict
+    ) -> Iterable[Either[CreateAPICollectionRequest]]:
+        """Method to return api collection Entities"""
+        try:
+            collection_request = CreateAPICollectionRequest(
+                name=EntityName(collection.get("name")),
```
pmbrull, line 97: let's prepare pydantic models, this is error-prone:
1. `collection.get("name")` might be `None`
2. `EntityName` does not allow `None`
3. The error better tell us the data is not behaving properly, rather than enigmatic pydantic errors.

Let's properly parse the payload and validate early (shift left as many errors as possible), then let the rest of the flow be "dumb" on how to treat the results, instead of having to be aware all the time about the shape of our data

pmbrull (changes requested):

Sachin-chaurasiya (dismissed):
The current UI changes look good, and I’ll handle the remaining UI changes in a separate patch.

pmbrull (approved):

Sachin-chaurasiya (approved):

PR #17779, opened 2024-09-10
MINOR: Optimise Snowflake Test Connection

ulixius9:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/bigquery/connection.py
+++ b/ingestion/src/metadata/ingestion/source/database/bigquery/connection.py
@@ -17,6 +17,7 @@ from datetime import datetime
 from functools import partial
 from typing import Optional
 
+from google.api_core.exceptions import NotFound
 from google.cloud.datacatalog_v1 import PolicyTagManagerClient
 from sqlalchemy.engine import Engine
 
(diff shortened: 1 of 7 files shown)


TeddyCr (dismissed):

ingestion/src/metadata/pii/scanners/ner_scanner.py
```diff
@@ -44,7 +44,7 @@ class NLPEngineModel(BaseModel):
     """Required to pass the nlp_engine as {"lang_code": "en", "model_name": "en_core_web_lg"}"""
 
     lang_code: str
-    model_name: str
+    nlp_model_name: str
```
ulixius9, line 47: to fix this warning

image

ingestion/src/metadata/ingestion/source/database/bigquery/connection.py
```diff
@@ -170,3 +171,28 @@ def test_connection_inner(engine):
...
+                tables = client.list_tables(dataset.reference, page_size=1)
+                for table in tables:
+                    if table.table_type in item_types:
+                        break
+            except NotFound:
+                # It's possible that the dataset was deleted between when we
+                # fetched the list of datasets and when we try to list the
+                # tables from it. See:
+                # https://github.com/googleapis/python-bigquery-sqlalchemy/issues/105
+                pass
```
sushi30, line 197: I suggest adding a debug message here so that if we hit this edge case we get some signal.

ulixius9: I think it should be fine, because this is a replica of function within bigquery dialect

in the test connection we can have the logger but the actual function would just pass it anyways 🤔

OnkarVO7 (approved):

PR #17809, opened 2024-09-11
MINOR: profiler e2e fix

harshsoni2024:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


SumanMaharana (approved):
LGTM.

ingestion/src/metadata/profiler/processor/core.py
```diff
@@ -603,7 +603,9 @@ def get_profile(self) -> CreateTableProfileRequest:
 
             if self._system_results:
                 system_profile = [
-                    SystemProfile(**system_result)
+                    system_result
+                    if isinstance(system_result, SystemProfile)
+                    else SystemProfile(**system_result)
```
ayush-shah, line 608: This seems like something we should refactor to have a standard return type i.e. SystemProfile

SumanMaharana: do u mean `metric_type != MetricTypes.System.value` ?

sushi30: With pydantic better to use:
```suggestion
                    system_result
                    if isinstance(system_result, SystemProfile)
                    else SystemProfile.model_validate(system_result)
```

harshsoni2024: so if metric_type = MetricTypes.System.value and still profile=None or [] it will still not convert profile into {} on line `395`

ingestion/src/metadata/profiler/processor/core.py
```diff
@@ -603,7 +603,9 @@ def get_profile(self) -> CreateTableProfileRequest:
 
             if self._system_results:
                 system_profile = [
-                    SystemProfile(**system_result)
+                    system_result
+                    if isinstance(system_result, SystemProfile)
+                    else SystemProfile.model_validate(system_result)
```
sushi30, line 608: checked pydantic API. this works:

```python
from metadata.ingestion.models.custom_pydantic import BaseModel
class Model(BaseModel):
    a: int
    
m1 = Model.model_validate(Model(a=1))
m2 = Model.model_validate({"a": 1})
assert m1 == m2
```

so you can just use:

```suggestion
                    SystemProfile.model_validate(system_result)
```

sushi30 (approved):

PR #17812, opened 2024-09-11
minor: update custom property entities data model, data product and database schema icon

Sachin-chaurasiya:
image


chirag-madlani (commented):
Can we also remove older icons?

harshach (approved):

chirag-madlani (approved):

PR #17855, opened 2024-09-16
Add Sigma Dashboard Connector

SumanMaharana:
### Describe your changes:

Added Sigma Dashboard Connector

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/dashboard/sigma/client.py
```diff
@@ -0,0 +1,177 @@
...
+    client: REST
+
+    def __init__(self, config: SigmaConnection):
+        self.config = config
+        token_api_key = str(
+            b64encode(
+                f"{self.config.clientId}:{self.config.clientSecret.get_secret_value()}".encode(
+                    "utf-8"
+                )
+            ).decode("utf-8")
```
pmbrull, line 56: we have a UTF_8 constant

ingestion/src/metadata/ingestion/source/dashboard/sigma/client.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+
+        headers = {
```
pmbrull, line 59: should this be a constant in this module?

ingestion/src/metadata/ingestion/source/dashboard/sigma/client.py
```diff
@@ -0,0 +1,177 @@
...
+            "accept": "application/json",
+            "Content-type": "application/x-www-form-urlencoded",
+        }
+
+        token_config = ClientConfig(
+            base_url=clean_uri(config.hostPort),
+            api_version=config.apiVersion,
+            auth_header="Authorization",
```
pmbrull, line 67: this could be a constant too

ingestion/src/metadata/ingestion/source/dashboard/sigma/client.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+
+        self.client = REST(client_config)
+
+    def get_auth_token(self) -> Tuple[str, int]:
+        """
+        generate auth token
+        """
+        payload = {"grant_type": "client_credentials"}
```
pmbrull, line 88: same, constant

pmbrull: and why do we return a tuple with 0 as a constant int? Maybe a pydantic model to have clear named values could help understand what's going on

ingestion/src/metadata/ingestion/source/dashboard/sigma/client.py
```diff
@@ -0,0 +1,177 @@
...
+        result = AuthToken(**self.token_client.post("/auth/token", data=payload))
+        return result.access_token, 0
+
+    def get_dashboards(self) -> Optional[List[Workbook]]:
+        """
+        method to fetch dashboards from api
+        """
+        result = WorkBookResponseDetails(**self.client.get("/workbooks"))
```
pmbrull, line 96: if client.get returns `None` we'd get an unpacking error here. Maybe we should handle this a bit better + use `model_validate` instead of directly unpacking

ingestion/src/metadata/ingestion/source/dashboard/sigma/client.py
```diff
@@ -0,0 +1,177 @@
...
+        if result:
+            return result.entries
+
+    def get_dashboard_detail(self, workbook_id: str) -> Optional[WorkbookDetails]:
+        """
+        method to fetch dashboard details from api
+        """
+        try:
+            result = WorkbookDetails(**self.client.get(f"/workbooks/{workbook_id}"))
```
pmbrull, line 105: same as above, let's apply everywhere

ingestion/src/metadata/ingestion/source/dashboard/sigma/models.py
```diff
@@ -0,0 +1,95 @@
...
+
+class AuthToken(BaseModel):
+    access_token: str
+    refresh_token: str
+    token_type: str
+
+
+class Workbook(BaseModel):
+    workbookId: str
+    name: Optional[str] = ""
```
pmbrull, line 27: should we default to `None` instead of empty strings?

ingestion/src/metadata/ingestion/source/dashboard/sigma/metadata.py
```diff
@@ -0,0 +1,228 @@
...
+                            table_name=node.name,
+                            database_name="",
+                        )
+                        from_entity = self.metadata.get_by_name(
+                            entity=Table,
+                            fqn=from_fqn,
+                        )
+                        if from_entity and to_entity:
+                            yield self._get_add_lineage_request(
+                                to_entity=to_entity, from_entity=from_entity
```
pmbrull, line 200: any way we can get column level lineage? Also, does sigma have the concept of DashboardDataModels? Or dashboards are directly connected to the source data

pmbrull (changes requested):

ingestion/src/metadata/ingestion/source/dashboard/sigma/metadata.py
SumanMaharana: sigma does have a concept of DashboardDataModels but it is introducted recently and its still in beta.
currently dashboards are directly connected to source data in sigma.
DashboardDataModels as in beta in sigma does not have any apis for them currently to fetch metadata from.

ingestion/src/metadata/ingestion/source/dashboard/sigma/client.py
SumanMaharana: i will apply that everywhere

ingestion/src/metadata/ingestion/source/dashboard/sigma/client.py
SumanMaharana: In `ClientConfig` for` auth_token` we require a Callable which returns a tuple with containing a auth token as str and expires_in as int. when we don't have an expires in we pass 0 as the expires_in time here.

ingestion/src/metadata/ingestion/source/dashboard/sigma/models.py
SumanMaharana: yes will default to None here

ingestion/src/metadata/ingestion/source/dashboard/sigma/client.py
pmbrull: then we could have a pydantic model explaining that one field is token, the second `expires_in` or something to make it clear

ingestion/src/metadata/ingestion/source/dashboard/sigma/metadata.py
pmbrull: ok got it, let's put it in a comment in the method as well as in the documentation. Any way we can get col-level lineage though?

SumanMaharana: ill check on the column level lineage in the sigma APIs

SumanMaharana: As of now we can build column level lineage between dashboard data models and the table

OnkarVO7 (approved):

pmbrull (approved):

OnkarVO7 (approved):

ShaileshParmar11 (approved):

PR #17866, opened 2024-09-16
[Issue-13782] Add Search RBAC config option in settings

mohityadav766:
### Describe your changes:

Part of Issue-13782

https://github.com/user-attachments/assets/568fe21b-2578-4742-b664-d5324c69bb1a

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


mohityadav766:
@chirag-madlani can someone take this up from UI we need to a simple setting with a toggle under Settings -> Access Control -> RBAC Search cc @harshach

chirag-madlani:
> @chirag-madlani can someone take this up from UI we need to a simple setting with a toggle under Settings -> Access Control -> RBAC Search cc @harshach

Having just a toggle in a page will not look good. will check with @Rupchavan here tomorrow

openmetadata-spec/src/main/resources/json/schema/configuration/rbacSearchConfiguration.json
harshach, line 1: @mohityadav766 Lets call this as SearchSettings. I want to add more configs to control the relevancy.
One of the options EnableAccessControl toggle

mohityadav766: makes sense @harshach , updated

openmetadata-service/src/main/java/org/openmetadata/service/search/security/RBACConditionEvaluator.java
```diff
@@ -43,8 +46,9 @@ public OMQueryBuilder evaluateConditions(SubjectContext subjectContext) {
         it.hasNext(); ) {
       SubjectContext.PolicyContext context = it.next();
       for (CompiledRule rule : context.getRules()) {
-        if (rule.getCondition() != null
-            && rule.getOperations().stream().anyMatch(SEARCH_RELEVANT_OPS::contains)) {
+        if ((rule.getCondition() != null
+                && rule.getOperations().stream().anyMatch(SEARCH_RELEVANT_OPS::contains))
```
mohityadav766, line 50: @harshach here i cannot find another way, since the domain only access role  uses All resources and on all metadata operation, any better ideas ?

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -2322,7 +2308,13 @@ public Object getLowLevelClient() {
 
   private void buildSearchRBACQuery(
       SubjectContext subjectContext, SearchSourceBuilder searchSourceBuilder) {
-    if (subjectContext != null && !subjectContext.isAdmin()) {
+    if (Boolean.TRUE.equals(
```
harshach, line 2311: @mohityadav766 lets create a common method for check this, given we have it in both

openmetadata-service/src/main/java/org/openmetadata/service/search/security/RBACConditionEvaluator.java
harshach: @mohityadav766 not sure I follow, what case are we missing?

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
mohityadav766: okay @harshach

openmetadata-service/src/main/java/org/openmetadata/service/search/security/RBACConditionEvaluator.java
mohityadav766: @harshach what i mean was, that domain restrictions apply to all type of metadataoperation, while the search to es mappings only is applied when VIEW_BASIC or VIEW_ALL operation is performed, so i had to manually look for DomainAccessRule and apply condition bypassing or that is below : 
```if ((rule.getCondition() != null
                && rule.getOperations().stream().anyMatch(SEARCH_RELEVANT_OPS::contains))
            || rule.getName().equalsIgnoreCase(DOMAIN_ONLY_ACCESS_RULE))```
            
 lmk if this makes sense

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
mohityadav766: resolved

openmetadata-service/src/main/java/org/openmetadata/service/search/security/RBACConditionEvaluator.java
harshach: @mohityadav766 the domain only access rule should come in from our policies while we are looping on them.
The goal of this check is when to enforce a rule on search related queries. The reason to do this is search related access is only related for view or view all. If someone has edit only rule that by default allows them to view atleast.

harshach:
@chirag-madlani lets addd beta flag around the search rbac setting in UI

chirag-madlani:
Duplicate #18338

PR #17871, opened 2024-09-17
Improve Search Relevancy for tier, usage and use stemming to generate ngrams

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/MlModelIndex.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/MlModelIndex.java
@@ -32,7 +32,6 @@ public class MlModelIndex implements SearchIndex {
     ParseTags parseTags = new ParseTags(Entity.getEntityTags(Entity.MLMODEL, mlModel));
     Map<String, Object> commonAttributes = getCommonAttributesMap(mlModel, Entity.MLMODEL);
     doc.putAll(commonAttributes);
-    doc.put("displayName", mlModel.getDisplayName());
     doc.put("tags", parseTags.getTags());
     doc.put("tier", parseTags.getTierTag());
     doc.put("serviceType", mlModel.getServiceType());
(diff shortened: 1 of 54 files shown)


ulixius9 (approved):

openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/MlModelIndex.java
```diff
@@ -32,9 +32,7 @@ public Map<String, Object> buildSearchIndexDocInternal(Map<String, Object> doc)
     ParseTags parseTags = new ParseTags(Entity.getEntityTags(Entity.MLMODEL, mlModel));
     Map<String, Object> commonAttributes = getCommonAttributesMap(mlModel, Entity.MLMODEL);
     doc.putAll(commonAttributes);
-    doc.put(
```
sonika-shah, line 35: @harshach,  wanted to understand why we are changing the logic for displayName. In cases where displayName is missing, we usually use name as a fallback.

The reason for this is that we will now be using displayName.keyword in most places. If an entity exists but doesn’t have a displayName, it will not appear in the aggregation results. (when field passed in aggregate api = displayName.keyword)

We can move this common logic to SearchIndexApp.getCommonAttributesMap to avoid repeating it in all index files.

image

harshach: our search query counts as the search string appears twice and scores it twice

sonika-shah: @harshach I think we are avoiding the double scoring here - screenshot added below 

, so it should be okay, if we have  
doc.put("displayName",
        mlModel.getDisplayName() != null ? mlModel.getDisplayName() : mlModel.getName());  present
        
        
image

harshach: the above is not going work. We need to make a tie breaker between displayName and name.
for now added displayName back, we will default to displayName

sonika-shah (approved):

PR #17923, opened 2024-09-19
MINOR: fix mssql integration test

sushi30:
use sql server 2022 due to something like https://github.com/microsoft/mssql-docker/issues/881 or https://github.com/microsoft/mssql-docker/issues/441

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/setup.py
+++ b/ingestion/setup.py
@@ -370,7 +370,7 @@ test = {
     VERSIONS["grpc-tools"],
     VERSIONS["neo4j"],
     "testcontainers==3.7.1;python_version<'3.9'",
-    "testcontainers==4.8.0;python_version>='3.9'",
+    "testcontainers==4.8.1;python_version>='3.9'",
     "minio==7.2.5",
     *plugins["mlflow"],
     *plugins["datalake-s3"],
--- a/ingestion/tests/integration/sql_server/conftest.py
+++ b/ingestion/tests/integration/sql_server/conftest.py
@@ -3,6 +3,9 @@ import shutil
 import tempfile
 
 import pytest
+from sqlalchemy import create_engine, text
+from testcontainers.mssql import SqlServerContainer
+
 from _openmetadata_testutils.helpers.docker import copy_dir_to_container, try_bind
 from metadata.generated.schema.api.services.createDatabaseService import (
     CreateDatabaseServiceRequest,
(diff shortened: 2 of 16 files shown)


pmbrull (dismissed):

ingestion/setup.py
```diff
@@ -370,7 +370,7 @@
     VERSIONS["grpc-tools"],
     VERSIONS["neo4j"],
     "testcontainers==3.7.1;python_version<'3.9'",
-    "testcontainers==4.8.0;python_version>='3.9'",
+    "testcontainers==4.8.1;python_version>='3.9'",
```
TeddyCr, line 373: Should we just fix it on minor version so we keep adding patches release? 

```suggestion
    "testcontainers~=4.8.0;python_version>='3.9'",
```

ingestion/tests/integration/sql_server/conftest.py
```diff
@@ -22,24 +23,55 @@
 from ..conftest import ingestion_config as base_ingestion_config
 
 
-@pytest.fixture(scope="module")
```
TeddyCr, line 25: More a question on my end. Wouldn't `scope=module` make it so the fixture would be destroyed at the end of the last test in `sql_server/test_lineage.py` for example?

sushi30: Indeed. With containers a best practice is to have **session** (since its most expensive to setup) level fixtures for the containers and **module** level fixtures for data entities like databases or tables.

sushi30: But for this case the data stays static throughout the tests so its fine like this.

TeddyCr (approved):
LGTM

ingestion/setup.py
TeddyCr: We can do that later

sushi30: done

sushi30: didnt make the commit... I will file a separate PR for this.

sushi30: https://github.com/open-metadata/OpenMetadata/pull/17938

PR #17969, opened 2024-09-24
GEN-1215: DI Populate Test Case Result & Resolution Status

ulixius9:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/workflows/dataQuality/DataQualityWorkflow.java
```diff
@@ -0,0 +1,195 @@
...
+import org.openmetadata.service.util.ResultList;
+import org.openmetadata.service.workflows.interfaces.Processor;
+import org.openmetadata.service.workflows.interfaces.Sink;
+import org.openmetadata.service.workflows.interfaces.Source;
+import org.openmetadata.service.workflows.searchIndex.PaginatedEntityTimeSeriesSource;
+
+@Slf4j
+public class DataQualityWorkflow {
+  public static final String DATA_STREAM_KEY = "DataStreamKey";
+  private final int retentionDays = 30;
```
TeddyCr, line 41: How do we use this value? Just want to make sure we will not delete index data that are older than 30 days. 😊

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchClient.java
```diff
@@ -108,6 +108,8 @@ public interface SearchClient {
 
   void createAliases(IndexMapping indexMapping);
 
+  void addIndexAlias(IndexMapping indexMapping, String aliasName);
```
TeddyCr, line 111: As part of the `searchRepository.createIndex()` we will add the alias defined in the `indexMappin.json` -- see below. Would it be worth adding maybe a `dataInsightAlias` key to that file and the corresponding object so that as part of `searchRepository.createIndex()` we add the corresponding DI alias. I am thinking it would allow to scale adding new index to the DI more easily and limit some code duplication -- WDYT?

```
  public void createIndex(IndexMapping indexMapping) {
    try {
      if (!indexExists(indexMapping)) {
        String indexMappingContent = getIndexMapping(indexMapping);
        searchClient.createIndex(indexMapping, indexMappingContent);
        searchClient.createAliases(indexMapping);
      } 
```

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -283,6 +287,24 @@ public void createIndex(IndexMapping indexMapping, String indexMappingContent) {
     }
   }
 
+  @Override
+  public void addIndexAlias(IndexMapping indexMapping, String aliasName) {
```
TeddyCr, line 291: Same thing than mentioned in the SearchClient.java

openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OpenSearchClient.java
```diff
@@ -275,6 +279,24 @@ public void createIndex(IndexMapping indexMapping, String indexMappingContent) {
     }
   }
 
+  @Override
+  public void addIndexAlias(IndexMapping indexMapping, String aliasName) {
```
TeddyCr, line 283: Same thing here than mentioned in the SearchClient

TeddyCr (commented):
@ulixius9 left a few comments. It might be worth checking the maven psql failing test as well since we changed some description null logic I believe.

image

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/workflows/dataQuality/DataQualityWorkflow.java
ulixius9: this restricts user for putting backfill date older then 30 days, do we here want to allow user to enable backfill upto 
any date?

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchClient.java
ulixius9: I'm not sure how this will work, whenever we add a new entry to `indexMappin.json`  it will try to first create a index and then try to create alias, in our case the index should already exist and we want to create alias for multiple indexes as part of DI app so not sure how it will work 🤔.

Also I feel the current logic is scalable as we only have to add the entity [type here](https://github.com/open-metadata/OpenMetadata/blob/bab93a84d269808466e326d45970ae47f984a38f/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java#L81) (which we would have to do anyways if we plan to add a new index to DI) as soon as we add that the alias will get created automatically.

and I see the code duplication I think I can take care of that 🤔

TeddyCr: Sounds good.

TeddyCr (commented):
@ulixius9 LGMT. Can you confirm if the Maven test failures are not linked to this change?

```
    } else {
       entityMap.put("hasDescription", CommonUtil.nullOrEmpty(entity.getDescription()) ? 0 : 1);
```

Once this is confirmed and the check style is fixed I can approve it 😊

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/workflows/dataQuality/DataQualityWorkflow.java
TeddyCr: Yes, we should

TeddyCr:
@ulixius9 LGTM, Let me know once you have resolved the conflicts and I'll approve it.

TeddyCr (approved):

PR #17970, opened 2024-09-24
update webpack dependencies

karanh37:
### Describe your changes:

Fixes Build timeout

Webpack 5 is using more memory in our CI due to memory leaks found [here](https://github.com/webpack/webpack/issues/13550) 
The CI builds are failing due to this. We don't want to increase the resources and as a result we are adding this fix.
The problem is with jest worker leaking memory.
For fix, we have made resolution of terser webpack plugin to 5.1.1 which uses older version of jest worker v26.

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/webpack.config.prod.js
+++ b/openmetadata-ui/src/main/resources/ui/webpack.config.prod.js
@@ -31,10 +31,6 @@ module.exports = {
   // Input configuration
   entry: ['@babel/polyfill', path.join(__dirname, 'src/index.tsx')],
 
-  cache: {
-    type: 'filesystem', // Enable caching on filesystem
-  },
-
   // Output configuration
   output: {
     path: outputPath,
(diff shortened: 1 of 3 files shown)


openmetadata-ui/src/main/resources/ui/webpack.config.prod.js
```diff
@@ -31,10 +31,6 @@ module.exports = {
   // Input configuration
   entry: ['@babel/polyfill', path.join(__dirname, 'src/index.tsx')],
 
-  cache: {
```
Sachin-chaurasiya, line 34: @karanh37 , I believe we have this config in the dev config as well, do we need that?

Sachin-chaurasiya (approved):

openmetadata-ui/src/main/resources/ui/webpack.config.prod.js
karanh37: No this is not needed. Will remove that in another PR. 
Thanks for highlighting

PR #18072, opened 2024-10-01
MINOR - Add App Limits

pmbrull:
### Describe your changes:

Main changes here:
- Updated `apps_extension_time_series` to work in a more generic approach. We were using it only for `status` information. Now we're adding the `extension` and `appName` fields so it can be used for storing limits information.
- There are migrations in there to ensure current info can follow the new structure
- We have updated the DAO logic for extension to have this generic behavior
- We have an `AppLimits` interface that each app can use to define their own limits
- The `apps_extension_time_series` table will store generic payloads of shape `AppExtension`. Limits and Status information are just extensions of this payload. It's up to each app to figure out what and how to store there.
    - However, for limits, each app is only allowed to store limits for its own ID for safety
- We're using this implementation for MetaPilot in Collate in another PR

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/AppRepository.java
```diff
@@ -199,6 +307,57 @@ public ResultList<AppRunRecord> listAppRuns(UUID appId, int limitParam, int offs
...
 
+  public <T> ResultList<T> listAllAppExtension(
+      App app, boolean byName, Class<T> clazz, AppExtension.ExtensionType extensionType) {
+    return listAppExtension(app, byName, Integer.MAX_VALUE, 0, clazz, extensionType);
+  }
+
+  public <T> T getLatestExtension(
+      App app, boolean byName, Class<T> clazz, AppExtension.ExtensionType extensionType) {
+    List<String> result;
+    if (byName) {
```
mohityadav766, line 318: @pmbrull we should spearate these into different funct getlatestExtensionByName and getlatestExtensionById

openmetadata-service/src/main/java/org/openmetadata/service/limits/AppLimits.java
```diff
@@ -0,0 +1,85 @@
...
+      this.dao.insert(
+          JsonUtils.pojoToJson(limitsExtension), AppExtension.ExtensionType.LIMITS.toString());
+    } catch (Exception e) {
+      LOG.error("Error inserting app limits for {}: {}", this.getApp().getName(), e.getMessage());
+    }
+  }
+
+  // Parse the app limits defined in the Private Configuration
+  // Let each App parse and store however is needed
+  public abstract void parseAppLimits() throws Exception;
```
mohityadav766, line 78: Exception should be avoided, if needed we can create a new one
AppLimitsParseException()

karanh37 (approved):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/AppRepository.java
pmbrull: wanted to avoid copy-pasting the same method, but fair enough, on it.

Sachin-chaurasiya (approved):

mohityadav766 (approved):

PR #18083, opened 2024-10-03
GEN-1552: Postgres stored procedures support

harshsoni2024:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/postgres/queries.py
+++ b/ingestion/src/metadata/ingestion/source/database/postgres/queries.py
@@ -296,3 +296,14 @@ POSTGRES_GET_JSON_FIELDS = """
     FROM
         aggregate_hierarchy;
 """
+
+POSTGRES_GET_STORED_PROCEDURES = """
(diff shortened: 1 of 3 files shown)


ingestion/src/metadata/ingestion/source/database/postgres/queries.py
```diff
@@ -296,3 +296,22 @@
...
+    WHERE prokind = 'p';
+"""
+
+POSTGRES_GET_STORED_PROCEDURES_QUERIES = """
+    SELECT 
+        total_exec_time,
+        query
+    FROM pg_stat_statements
+    WHERE query LIKE 'CALL%';
+"""
```
ulixius9, line 317: we do not want the query which calls sp, we will not be able to extract queries that were executed during sp execution so lets just add a comment for that and skip this query call.

PR #18098, opened 2024-10-03
MINOR: Fix Ingestion Dockerfile Compatibility by changing IF syntax

IceS2:
Depending on how we are building the Docker Images, a couple commands would fail as the syntax was not compatible with SH.
This PR aims to fix that by changing the syntax to a more portable one.

It was tested locally by building and running the ingestion dockerfile using zsh, bash and sh.

#
### Type of change:

- [x] Bug fix

#
### Checklist:


tutte (approved):
LGTM :rocket:

SumanMaharana (approved):
LGTM 👌

PR #18115, opened 2024-10-04
MINOR: add column case sensitivity parameter

sushi30:
- added handling for case-insensitive columns
- added handling for different numeric types (int/float/Decimal)
- added handling of boolean test case parameters

## Boolean test case parameters

Appear to be sent as json strings:

image

Therefore I added handling as string types.

## Validation

- [x] integration tests
- [x] migrations tested on snapshot from snadbox 

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/data_quality/validations/base_test_handler.py
```diff
@@ -87,7 +88,7 @@ def get_test_case_param_value(
             (param.value for param in test_case_param_vals if param.name == name), None
         )
 
-        if not value:
+        if value is None:
```
TeddyCr, line 91: is it expected to move from considering `[]`,`""`, `None` to only considering `None`?

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
```diff
@@ -75,6 +76,18 @@ def masked(s: str, mask: bool = True) -> str:
...
+
+    Args:
+        t: type to check
+
+    Returns:
+        True if the type is numeric otherwise False
+    """
+    return t in [int, float, Decimal]
+
+
```
TeddyCr, line 90: Are we able to use https://github.com/open-metadata/OpenMetadata/blob/aa7b95a515a6a72b7bad5f49aa19963b85c25be7/ingestion/src/metadata/profiler/orm/registry.py#L159 ?

ingestion/src/metadata/utils/collections.py
```diff
@@ -0,0 +1,27 @@
...
+class CaseInsensitiveString(str):
+    """
+    A case-insensitive string. Useful for case-insensitive comparisons like SQL.
+    """
+
+    def __eq__(self, other):
+        return self.lower() == other.lower()
+
+    def __hash__(self):
+        return hash(self.lower())
```
TeddyCr, line 15: You should use `.casefold()` for caseless matching. https://docs.python.org/3/library/stdtypes.html#str.casefold.

e.g.
```python
>>> "fooßBar".lower()
'fooßbar'
>>> "fooßBar".casefold()
'foossbar'
```

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
sushi30: no. that is specific for sqlalchemy types

ingestion/src/metadata/data_quality/validations/base_test_handler.py
sushi30: retained original behavior and changed how the parameter is consumed

sushi30:
@TeddyCr addressed your comments:

- use casefold()
- reverted change to test case parameters
- added special handling for boolean test case parameters

ingestion/src/metadata/data_quality/validations/utils.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import Callable, Optional, TypeVar, Union
+
+from metadata.generated.schema.tests.testCase import TestCaseParameterValue
+
+T = TypeVar("T", bound=Callable)
+R = TypeVar("R")
+
+
+def get_test_case_param_value(
+    test_case_param_vals: list[TestCaseParameterValue],
```
TeddyCr, line 13: I think you may need to make that `List[...]` vs `list[...]`

TeddyCr (dismissed):

ingestion/src/metadata/data_quality/validations/base_test_handler.py
```diff
@@ -65,37 +66,18 @@ def run_validation(self) -> TestCaseResult:
...
     def get_test_case_param_value(
-        self,
-        test_case_param_vals: list[TestCaseParameterValue],
+        test_case_param_vals: List[TestCaseParameterValue],
         name: str,
         type_: T,
         default: Optional[R] = None,
         pre_processor: Optional[Callable] = None,
     ) -> Optional[Union[R, T]]:
-        """Give a column and a type return the value with the appropriate type casting for the
```
TeddyCr, line 76: Should we keep this doc string?

sushi30: I moved it to the underlying utils library bc this doesnt have much meat. didnt want to keep redundant material. wdyt?

sushi30: here: https://github.com/open-metadata/OpenMetadata/pull/18115/files#diff-81b87512f0e5a37568c837e4c3631ae91431312ff2b185e864aa18a260d1c8faR20

TeddyCr (approved):

pmbrull (approved):

PR #18172, opened 2024-10-08
MINOR: open links in new tab in DataAssetsHeader

fuzmish:
### Describe your changes:

(No issue opened explicitly)

- What changes did you make?
  - Change the links in the DataAssetsHeader component to open in a new tab.
- Why did you make them?
  - Some entities have properties with link information to external web resources (e.g., [Dashboard.sourceUrl](https://github.com/open-metadata/OpenMetadata/blob/4233abf41587b436402d1abec2bcc2d72c913f64/openmetadata-spec/src/main/resources/json/schema/entity/data/dashboard.json#L70)), which may be rendered as `<a>` links in the DataAssetsHeader component.
  - Most or all of these are resources in a separate location from the OpenMetadata UI, and it is preferred that the links be opened in a new tab. (Currently, they open in the same tab.)
  - I have added the unit test for the ExtraInfoLink component.
  - I have tested the behavior in the local development UI server.

References:
- [rel=noopener - HTML: HyperText Markup Language | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener)
- [Why rel='noopener noreferrer'?](https://stackoverflow.com/a/50709724)

#
### Type of change:
- [x] Improvement

#
### Checklist:

- [x] ~~My PR title is `Fixes <issue-number>: <short explanation>`~~
- [x] ~~For JSON Schema changes: I updated the migration scripts or explained why it is not needed.~~
- [x] ~~For connector/ingestion changes: I updated the documentation.~~


openmetadata-ui/src/main/resources/ui/src/utils/DataAssetsHeader.utils.tsx
```diff
@@ -178,13 +180,15 @@ export const getDataAssetsHeaderInfo = (
...
             <ExtraInfoLink
+              newTab
               href={mlModelDetail.server}
               label={t('label.server')}
               value={mlModelDetail.server}
             />
           )}
           {mlModelDetail.dashboard && (
             <ExtraInfoLink
+              newTab
```
chirag-madlani, line 191: This won't be new tab, it's internal route

openmetadata-ui/src/main/resources/ui/src/utils/DataAssetsHeader.utils.tsx
```diff
@@ -422,6 +427,7 @@ export const getDataAssetsHeaderInfo = (
         <>
           {apiCollection.endpointURL && (
             <ExtraInfoLink
+              newTab
```
chirag-madlani, line 430: @Sachin-chaurasiya can you confirm if `apiCollection.endpointURL` will be external or internal?

openmetadata-ui/src/main/resources/ui/src/utils/DataAssetsHeader.utils.tsx
```diff
@@ -451,6 +457,7 @@ export const getDataAssetsHeaderInfo = (
           )}
           {apiEndpoint.endpointURL && (
             <ExtraInfoLink
+              newTab
```
chirag-madlani, line 460: same as above

openmetadata-ui/src/main/resources/ui/src/utils/DataAssetsHeader.utils.tsx
Sachin-chaurasiya, line 191: @fuzmish, I believe this is an internal path, so we don’t need the newTab as true here.

Sachin-chaurasiya (commented):
Thanks for the PR @fuzmish, left one comment and rest looks good.

openmetadata-ui/src/main/resources/ui/src/utils/DataAssetsHeader.utils.tsx
Sachin-chaurasiya: Yes,@chirag-madlani that will be the external link.

openmetadata-ui/src/main/resources/ui/src/utils/DataAssetsHeader.utils.tsx
Sachin-chaurasiya: Same here, it will be external link.

openmetadata-ui/src/main/resources/ui/src/utils/DataAssetsHeader.utils.tsx
fuzmish: Thank you for your comment. a4bd542 I've fixed it.

chirag-madlani (approved):

PR #18234, opened 2024-10-11
Fixed dbt Manifest and Run results parsing

SumanMaharana:
### Describe your changes:

Fixed dbt Manifest and Run results parsing

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/dbt/dbt_service.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt/dbt_service.py
@@ -13,7 +13,7 @@ DBT service Topology.
 """
 
 from abc import ABC, abstractmethod
-from typing import Iterable
+from typing import Iterable, List
 
 from dbt_artifacts_parser.parser import parse_catalog, parse_manifest, parse_run_results
 from pydantic import Field
@@ -37,6 +37,10 @@ from metadata.ingestion.models.topology import (
     TopologyNode,
 )
 from metadata.ingestion.source.database.database_service import DataModelLink
+from metadata.ingestion.source.database.dbt.constants import (
+    REQUIRED_NODE_KEYS,
+    REQUIRED_RESULTS_KEYS,
+)
 from metadata.ingestion.source.database.dbt.dbt_config import get_dbt_details
 from metadata.ingestion.source.database.dbt.models import (
     DbtFiles,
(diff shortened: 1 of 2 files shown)


ingestion/src/metadata/ingestion/source/database/dbt/dbt_service.py
```diff
@@ -169,51 +173,28 @@ def remove_manifest_non_required_keys(self, manifest_dict: dict):
...
-            "compiled_code",
-            "compiled_sql",
-            "raw_code",
-            "raw_sql",
-            "language",
-        }
+        for field in ["nodes", "sources"]:
+            for node, value in manifest_dict.get(  # pylint: disable=unused-variable
+                field
+            ).items():
```
TeddyCr, line 179: @SumanMaharana here we should probably check `manifest_dict.get(...` returns a dict and is not Null -- we can also update the method signature `def remove_manifest_non_required_keys(self, manifest_dict: dict):` if we are expecting indeed a `Dict[str, dict]`.


Otherwise we'll get 
```
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'items'
```
or 
```
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'items'
```

ingestion/src/metadata/ingestion/source/database/dbt/dbt_service.py
```diff
@@ -225,6 +206,9 @@ def get_dbt_objects(self) -> Iterable[DbtObjects]:
         self.remove_manifest_non_required_keys(
             manifest_dict=self.context.get().dbt_file.dbt_manifest
         )
+        self.remove_run_result_non_required_keys(
+            run_results=self.context.get().dbt_file.dbt_run_results
```
TeddyCr, line 210: Is it possible to end up with a None value when calling `self.context.get()`? If so we should handle it so we don't end up with some `AttribiuteErrro`

ingestion/src/metadata/ingestion/source/database/dbt/dbt_service.py
SumanMaharana: @TeddyCr `manifest_dict.get(...` will always return a dict as its defined that way in dbt.

ingestion/src/metadata/ingestion/source/database/dbt/dbt_service.py
SumanMaharana: Added the check there

PR #18270, opened 2024-10-15
Minor : Fix exception in search due to exception in  database.displayName and databaseSchema.aggregation

sonika-shah:
### Describe your changes:

Fixes  exception on  search query for all
image

Changes done :
- segregate search builder for dataAsset and all, created new method for dataAssetSearchBuilder
- fixed the missing serviceType in test_suite_index

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


TeddyCr:
@sonika-shah if I am not mistaking you can always do terms/aggregation against this field with `displayName.keyword` since we have this mapping

```
          "displayName": {
             "type": "text",
             "fields": {
               "keyword": {
                 "type": "keyword",
             }
           },
```
if we change the mapping for display name to keyword, then search query such as `foo` for a `My Foo Table` display name will not return a match.

harshach (approved):

PR #18272, opened 2024-10-15
FIx #17933: Superset fix chart query for mysql con.

### LINKED ISSUE #17933: Superset MySQL API Connection Fails on getCharts Due to Incorrect Backend Query in OpenMetadata
I upgraded to the latest OpenMetadata version (1.5.2), which OM experts communicated to me, and tried to establish a connection with Apache Superset using the MySQL API. However, it still shows an error when fetching charts.

Superset version: 4.0.2
MySQL version: 8.0.30
Open metadata Version: 1.5.2
OpenMetadata Ingestion package 1.5.2
Screenshot 2024-09-20 115230
I will mention the below query which works fine OM expert provided it:

select 
	s.id,
	s.slice_name,
	s.description,
    	s.datasource_id,
    	s.viz_type,
	t.table_name,
	t.schema,
	db.database_name,
    db.sqlalchemy_uri
from 
	slices s left join `tables` t 
on s.datasource_id  = t.id and s.datasource_type = 'table' 
	left join `dbs` db 
on  db.id = t.database_id

So, the OM team needs to fix the query.

harshsoni2024:
### Describe your changes:

Fixes #17933 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/superset/connection.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/superset/connection.py
@@ -84,7 +84,12 @@ def test_connection(
     else:
         test_fn["CheckAccess"] = partial(test_connection_engine_step, client)
         test_fn["GetDashboards"] = partial(test_query, client, FETCH_DASHBOARDS_TEST)
-        test_fn["GetCharts"] = partial(test_query, client, FETCH_ALL_CHARTS_TEST)
+        if isinstance(service_connection.connection, MysqlConnection):
+            test_fn["GetCharts"] = partial(
+                test_query, client, FETCH_ALL_CHARTS_TEST.replace('"', "`")
+            )
+        else:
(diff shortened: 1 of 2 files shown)


OnkarVO7 (approved):

IceS2 (commented):
Looks good, but maybe we could have the logic in one place, doing something like:

```
def getSupersetQuery(String query, Connection connection):
  if isinstance(connection, MysqlConnection):
    return query.replace('"', "`")
  return query
```

harshsoni2024:
> Looks good, but maybe we could have the logic in one place, doing something like:
> 
> ```
> def getSupersetQuery(String query, Connection connection):
>   if isinstance(connection, MysqlConnection):
>     return query.replace('"', "`")
>   return query
> ```

I'll do this in next changes sure. Need to cherry-pick this branch today 😅

PR #18280, opened 2024-10-15
GEN-1166 - Improve Ingestion Workflow Error Summary

pmbrull:
### Describe your changes:

Adding success % to the status list

```
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Workflow Mysql Summary:
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Processed records: 91
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Updated records: 0
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Warnings: 0
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Errors: 0
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Success %: 100.0
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Workflow OpenMetadata Summary:
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Processed records: 64
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Updated records: 0
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Warnings: 0
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Errors: 27
[2024-10-15 15:28:50] INFO     {metadata.Utils:logger:181} - Success %: 70.33
```
and unifying how we compute success %.

Also, final % is based as the mean of the success of all steps. Before, we only gave "wiggle room" for failures on the source, but not on any other step. Now, the SUCCESS_THRESHOLD is applied to all steps, and final success % is based on all the steps.

Finally, we are allowing to configure the success % in the workflowConfig instead of having it as a constant in the code.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/workflow/workflow_output_handler.py
+++ b/ingestion/src/metadata/workflow/workflow_output_handler.py
@@ -14,6 +14,7 @@ Module handles the output messages from different workflows
 """
 
 import time
+from statistics import mean
 from typing import Any, Dict, List, Optional, Type, Union
 
 from pydantic import BaseModel
@@ -116,14 +117,10 @@ class WorkflowOutputHandler:
 
     def _print_summary(self, steps: List[Step]):
         failures: List[Failure] = []
-        total_records: int = 0
-        total_errors: int = 0
 
         for step in steps:
             step_summary = Summary.from_step(step)
 
-            total_records += step_summary.records or 0
-            total_errors += step_summary.errors or 0
             failures.append(
                 Failure(name=step.name, failures=step.get_status().failures)
             )
@@ -141,15 +138,18 @@ class WorkflowOutputHandler:
                 log_ansi_encoded_string(message=f"Filtered: {step_summary.filtered}")
 
             log_ansi_encoded_string(message=f"Errors: {step_summary.errors}")
+            log_ansi_encoded_string(
+                message=f"Success %: {step.get_status().calculate_success()}"
+            )
 
         self._print_failures_if_apply(failures)
 
-        total_success = max(total_records, 1)
+        # If nothing is processed, we'll have a success of 100%
+        success_pct = mean([step.get_status().calculate_success() for step in steps])
         log_ansi_encoded_string(
             color=ANSI.BRIGHT_CYAN,
             bold=True,
-            message="Success %: "
-            + f"{round(total_success * 100 / (total_success + total_errors), 2)}",
+            message="Workflow Success %: " + f"{round(success_pct, 2)}",
         )
 
     def _print_debug_summary(self, steps: List[Step]):
--- a/ingestion/src/metadata/workflow/workflow_output_handler.py
+++ b/ingestion/src/metadata/workflow/workflow_output_handler.py
@@ -115,8 +115,11 @@ class WorkflowOutputHandler:
 
         self._print_summary(steps)
 
-    def _print_summary(self, steps: List[Step]):
+    def _print_summary(self, steps: List[Step]) -> None:
         failures: List[Failure] = []
+        if not steps:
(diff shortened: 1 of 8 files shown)


openmetadata-docs/content/partials/v1.6/deployment/upgrade/upgrade-prerequisites.md
```diff
@@ -113,75 +113,14 @@ We believe this update will bring greater consistency and clarity to our version
...
-
-To reduce the processing time for out-of-the-box workflows, we are disabling the include DDL by default, whereas before, it was enabled, which potentially led to long-running workflows.
-
-### Secrets Manager
-Starting with the release 1.5.0, the JWT Token for the bots will be sent to the Secrets Manager if you configured one. It won't appear anymore in your dag_generated_configs in Airflow.
-
-### Python SDK
-The `metadata insight` command has been removed. Since Data Insights application was moved to be an internal system application instead of relying on external pipelines the SDK command to run the pipeline was removed.
+Now, we're changing this behavior to consider the success rate of all the steps involved in the workflow. The UI will
+then show more `Partial Success` statuses rather than `Failed`, properly reflecting the real state of the workflow.
```
sushi30, line 126: I dont understand how this impacts me as a user. Will I have more failing ingestions? Less? We can take an example ingestion and show how it would reflect differently now compared to before this change.

ingestion/src/metadata/workflow/workflow_output_handler.py
```diff
@@ -141,15 +138,18 @@ def _print_summary(self, steps: List[Step]):
...
             log_ansi_encoded_string(message=f"Errors: {step_summary.errors}")
+            log_ansi_encoded_string(
+                message=f"Success %: {step.get_status().calculate_success()}"
+            )
 
         self._print_failures_if_apply(failures)
 
-        total_success = max(total_records, 1)
+        # If nothing is processed, we'll have a success of 100%
+        success_pct = mean([step.get_status().calculate_success() for step in steps])
```
sushi30, line 148: this will fail with this error if steps is empty. 
```
statistics.StatisticsError: mean requires at least one data point
```

Not sure if we currently have such a use-case but worth handling this just in case and ask the users to check config or send them to slack.

pmbrull: should not happen since we pass the steps by hand but fair enough, will handle that

sushi30 (approved):

openmetadata-docs/content/partials/v1.6/deployment/upgrade/upgrade-prerequisites.md
sushi30: approved, as a side quest I suggest we clarify definition of "partial success" or do away with it. From my experience ETL processes have a status of `success` / `failure` so the user knows if they need to inverine or not. I have suspicions that "partial success" can confuse platform users of whether they need to do something about their pipelines.

Examples of this design can be seen in:

- [Bash - nonzero for failure](https://tldp.org/LDP/abs/html/exit-status.html)
- [Airflow state machine](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/tasks.html#task-instances)
- [AWS Lambda](https://docs.aws.amazon.com/lambda/latest/dg/functions-states.html)
- [Dagster](https://docs.dagster.io/_apidocs/execution#dagster.DagsterEvent.is_failure) - a bit less orthodox but a boolean value.

PR #18290, opened 2024-10-16
Minor : Fix exception in search due to exception in database.displayN…

sonika-shah:
### Describe your changes:

Minor : Fix exception in search due to exception in database.displayName and databaseSchema.aggregation

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
@@ -1305,6 +1305,23 @@ public class ElasticSearchClient implements SearchClient {
         AggregationBuilders.terms("databaseSchema.name.keyword")
             .field("databaseSchema.name.keyword")
             .size(MAX_AGGREGATE_SIZE));
+    return addAggregation(searchSourceBuilder);
+  }
+
+  private static SearchSourceBuilder buildDataAssetsSearchBuilder(
+      String query, int from, int size) {
+    QueryStringQueryBuilder queryStringBuilder =
+        buildSearchQueryBuilder(query, SearchIndex.getAllFields());
+    FunctionScoreQueryBuilder queryBuilder = boostScore(queryStringBuilder);
+    SearchSourceBuilder searchSourceBuilder = searchBuilder(queryBuilder, null, from, size);
+    searchSourceBuilder.aggregation(
+        AggregationBuilders.terms("database.name.keyword")
+            .field("database.name.keyword")
+            .size(MAX_AGGREGATE_SIZE));
+    searchSourceBuilder.aggregation(
+        AggregationBuilders.terms("databaseSchema.name.keyword")
+            .field("databaseSchema.name.keyword")
+            .size(MAX_AGGREGATE_SIZE));
     searchSourceBuilder.aggregation(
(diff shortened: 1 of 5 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -1297,6 +1297,23 @@ private static SearchSourceBuilder buildSearchAcrossIndexesBuilder(
...
     SearchSourceBuilder searchSourceBuilder = searchBuilder(queryBuilder, null, from, size);
+    searchSourceBuilder.aggregation(
+        AggregationBuilders.terms("database.name.keyword")
+            .field("database.name.keyword")
+            .size(MAX_AGGREGATE_SIZE));
+    searchSourceBuilder.aggregation(
+        AggregationBuilders.terms("databaseSchema.name.keyword")
+            .field("databaseSchema.name.keyword")
+            .size(MAX_AGGREGATE_SIZE));
+    return addAggregation(searchSourceBuilder);
```
TeddyCr, line 1308: Should we do the aggregation on the fullyQualifiedName? Using name I believe if we have `service.database.schema` and `otherService.database.schema` then they'll both get aggregated under the same if we use `name` I believe.

openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OpenSearchClient.java
```diff
@@ -1282,6 +1282,24 @@ private static SearchSourceBuilder buildSearchAcrossIndexesBuilder(
...
     SearchSourceBuilder searchSourceBuilder = searchBuilder(queryBuilder, null, from, size);
+    searchSourceBuilder.aggregation(
+        AggregationBuilders.terms("database.name.keyword")
+            .field("database.name.keyword")
+            .size(MAX_AGGREGATE_SIZE));
+    searchSourceBuilder.aggregation(
+        AggregationBuilders.terms("databaseSchema.name.keyword")
+            .field("databaseSchema.name.keyword")
+            .size(MAX_AGGREGATE_SIZE));
+    return addAggregation(searchSourceBuilder);
```
TeddyCr, line 1293: Same comment here https://github.com/open-metadata/OpenMetadata/pull/18290/files#r1803006416

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
sonika-shah: hi @TeddyCr thanks for pointing your concern is valid, but for our usage we are not  dependent
only on a single aggregation while getting result, we check the condition with other aggregations as well
image

sonika-shah: have tried to replicate the same scenario with same database and schema, seems to be working fine.
Also will be removing the support for aggregations on name.keyword, as in most scenario we are moving to show results based on displayName

image

TeddyCr (approved):

harshach (approved):

PR #18292, opened 2024-10-16
FIX #17042 - Allow non admin/bot to read profiler global config

### LINKED ISSUE #17042: Enable reading global profiler settings not only by admins and bots

We have a feature on our platform where we enable users to execute profiling jobs themselves. We provide curated workflow template, users update table and schema name and execute ad-hoc jobs for their tables. This was working fine until we upgraded to OM 1.4.0.0, where https://github.com/open-metadata/OpenMetadata/pull/15889 was introduced. Now, regular users (not admins or bots) cannot execute profiling jobs as they get 403 error on fetching global profiler config.

Requesting global profiler config (`get /api/v1/system/settings/profilerConfiguration`) is not restricted only to admins and bots. Proposed approaches:
- move this resource out of system settings
- create resource `ProfilerConfig` (or more generic `SystemSettings`) with operations `create` `edit` `read` like with other resources

What's particularly interesting about aforementioned implementation is that `authorizeAdminOrBot` method is used only once throughout whole OM service - in said endpoint.

Since we are grouping permissions using Teams/Groups (we have Teams A, B, C and we assign users to their respective teams, then we assign DatabaseSchema owner to a team. For example schema `transactions` from `Trino` service is owned by team A and all members of team A can edit metadata in transactions schema tables) we considered extending functionality of OpenMetadata with scoped bot users (so we could create bot `X` that would be a member of team `A` - this bot would inherit permissions of the team but would be treated as bot) https://github.com/open-metadata/OpenMetadata/issues/15891.

**Additional context**
We follow shift-left paradigm, so instead of running profiling jobs within OM Airflow, we instead allow users to do this with their personal credentials and on their own desired cadence. This is very important security-wise, as our OM instance service connections cannot use accounts (NPAs) that have access to actual data so moving this responsibility to end users is our only way to get profiling data into OM.

TeddyCr:
### Describe your changes:
Fixes #17042 
- Introduce `ViewProfilerGlobalConfig` permission to allow non admin/bot to read global profiler  config on ingestion when using their personal token -- `ViewAll` will allow reading this config. It is possible to add `ViewAll` and explicitly deny `ViewProfilerGlobalConfig` if one chooses to
- fix customer properties package name from `customproperties` to `customProperties`

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Bug fix


harshach (approved):

openmetadata-spec/src/main/resources/json/schema/type/customProperties/enumConfig.json
```diff
@@ -3,7 +3,7 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "EnumConfig",
   "type": "object",
-  "javaType": "org.openmetadata.schema.type.customproperties.EnumConfig",
+  "javaType": "org.openmetadata.schema.type.customProperties.EnumConfig",
```
sushi30, line 6: why are we changing this class path?

TeddyCr: because the java package generated is customProperties (and not `customproperties`)

sushi30 (approved):
nit: add reason for changing  of `customProperties` -> `customproperties` so that it we have it in our commit log.

TeddyCr:
> nit: add reason for changing of `customProperties` -> `customproperties` so that it we have it in our commit log.

@sushi30 it is already the description of this commit 243afd0feb6912bdccd524c44c91987f0c124303

PR #18322, opened 2024-10-18
[GEN-356] Use ServiceSpec for loading sources based on connectors

sushi30:
### Introducing the OpenMetadata service specification.

Main things to look out for:

- Removal of the `import_x_class` in favor of a [central `BaseSpec.get_for_source`](https://github.com/open-metadata/OpenMetadata/pull/18322/files#diff-61be2a367dfcecf6276b7d000dd450c972325909b8cdfac1402d29fd4929d4f0R50) which will provide the centralized repository for all the source's interfaces.
- [Breaking change documentation](https://github.com/open-metadata/OpenMetadata/pull/18322/files#diff-92c58296a428163f247ba74d2397fb9a4e1818e6e6659c9d34295dca004369abR20)
- Adding the [`system_metrics_computer_class` field in the `SQAProfilerInterface`](https://github.com/open-metadata/OpenMetadata/pull/18322/files#diff-349d411eee3a3f5548ec34b1de6a9513f65d4dcda8e790dffa1201e2f38aec83R10)
- New profiler implementations for redshift, biquery, snowflake to accomodate for the new system metric source
- use source classes that can be overridden in system profiles
- use a manifest class instead of factory to specify which class to resolve for connectors
- example usage can be seen in redshift and snowflake

### TODO

- [x] Implement the BigQuery system metric source class  
- [ ] run e2e tests ([link to GA](https://github.com/open-metadata/OpenMetadata/actions/workflows/py-cli-e2e-tests.yml?query=branch%3Asystem-metrics-refactor))

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/database/redshift/manifest.py
```diff
@@ -0,0 +1,6 @@
+from metadata.ingestion.source.database.redshift.profiler.profiler import (
+    RedshiftProfiler,
+)
+from metadata.utils.manifest import BaseManifest, get_class_path
+
+RedshiftManifest = BaseManifest(profler_class=get_class_path(RedshiftProfiler))
```
TeddyCr, line 6: I am thinking since we control the framework and limit "meta" management of class import and what not when creating additional classes, can we just dynamically import the class based on the service type?

So if I need to add a new class to support some functionality just for Athena, then I just need to implement that class and I don't need to thinking about managing this `manifest.py` file for the specific source.

sushi30: Using a dynamic import approach for all our classes raises to issues:

1. How does one implement a profiler for a [CustomDatabase](https://docs.open-metadata.org/latest/connectors/custom-connectors) in this case?
The idea behind this implementation is to have a default map at the [service type level](https://github.com/open-metadata/OpenMetadata/pull/18322/files#diff-a30dbc8e9bba8105e9857501e0e29627b792608d416e01c65a7aa99af22d2e72R79) and allow the contributors to have more flexibility when developing other modules for their connectors.

2. Our dynamic module resolution is not type safe. One such example is that `BigQuery` is sometimes resolved as `bigquery`, `Bigquery`, or `BigQuery`.  A dynamic import path approach requires the user to have this knowledge and implementing each separate class with these conventions in mind. One can easily forget what is the right casing fora class name and use the wrong conventions when writing the class (for example `BigQuerySource` instead of `BigquerySource`). Without type-safety the only way we can such errors is at runtime. With the current approach there is a single point of failure (the manifest). The connectors themselves are contained within a system that can be validated independently.

ingestion/src/metadata/profiler/metrics/system/system.py
```diff
@@ -58,207 +38,81 @@ def recursive_dic():
...
+    """Empty system metrics source that can be used as a default. Just returns an empty list of system metrics
+    for any resource."""
+
+    def __init__(self, *args, **kwargs):
+        kwargs.pop("session", None)
+        if len(args) > 0:
+            logger.warning("Received unexpected arguments: %s", args)
+        if len(kwargs) > 0:
+            logger.warning("Received unexpected keyword arguments: %s", kwargs)
+        super().__init__()
```
TeddyCr, line 72: Any reason to not define a strong interface here directly __init__(self, session)? 

Also, I think `super().__init__()` will be redundant here.

ingestion/src/metadata/profiler/metrics/system/system.py
```diff
@@ -58,207 +38,81 @@ def recursive_dic():
...
+        """Get update queries"""
+        return []
+
+    def get_kwargs(self, *args, **kwargs):
+        return {}
+
+
+class SystemMetricsComputer(EmptySystemMetricsSource):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
TeddyCr, line 92: That will not be needed I believe

ingestion/src/metadata/utils/lru_cache.py
```diff
@@ -13,47 +13,108 @@
...
+                key = key_func(*args, **kwargs)
+                if key in self:
+                    return self.get(key)
+                value = func(*args, **kwargs)
+                self.put(key, value)
+                return value
+
+            return wrapped
+
+        return wrapper
```
TeddyCr, line 120: Have we considered simply using `@functools.lru_cache` https://docs.python.org/3/library/functools.html#functools.lru_cache? I believe this is what it does.

ingestion/src/metadata/ingestion/source/database/bigquery/profiler/profiler.py
```diff
@@ -0,0 +1,26 @@
...
+)
+from metadata.profiler.interface.sqlalchemy.bigquery.profiler_interface import (
+    BigQueryProfilerInterface,
+)
+from metadata.profiler.metrics.system.system import System
+from metadata.profiler.processor.runner import QueryRunner
+
+
+class BigQueryProfiler(BigQueryProfilerInterface):
+    system_metrics_computer_class = BigQuerySystemMetricsComputer
```
TeddyCr, line 15: Should the metric computation definition be coupled with the profiler class? I believe right now this is somewhat decouple where the metrics computation logic lives in the metrics module and the metric module will either use the base definition and use the one related to the source -- but there is no metric injection directly into the profiler.

ingestion/src/metadata/profiler/metrics/system/system.py
```diff
@@ -58,207 +38,81 @@ def recursive_dic():
...
+
+    def get_system_metrics(self, *args, **kwargs) -> List[SystemProfile]:
+        """Return system metrics for a given table. Actual passed object can be a variety of types based
+        on the underlying infrastructure. For example, in the case of SQLalchemy, it can be a Table object
+        and in the case of Mongo, it can be a collection object."""
+        kwargs = super().get_kwargs(*args, **kwargs)
+        return (
+            super().get_inserts(**kwargs)
+            + super().get_deletes(**kwargs)
+            + super().get_updates(**kwargs)
```
TeddyCr, line 102: Why do we have this very loose interface and we are not enforcing specific interface that children should implement.

ingestion/src/metadata/utils/importer.py
```diff
@@ -302,3 +288,18 @@ def import_side_effects(self, *modules):
...
     SideEffectsLoader().import_side_effects(*modules)
+
+
+def import_system_metrics_computer(db_service: DatabaseService):
+    """
+    Import the system metrics profile class
+    """
+    try:
+        return import_from_module(
+            "metadata.ingestion.source.database.{}.profiler.system.SystemMetricsComputer".format(  # pylint: disable=consider-using-f-string
```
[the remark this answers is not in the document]
pmbrull: e.g., I understand we can keep tight imports on the `BaseSpec`, and the workflows themselves make sense since we have it based on YAMLs we don't want to change atm, but these other internal pieces such as the system metrics, why dynamic here?

ingestion/src/metadata/utils/service_spec/service_spec.py
```diff
@@ -0,0 +1,91 @@
...
+
+    profiler_class: Optional[str]
+    metadata_source_class: str
+
+    @model_validator(mode="before")
+    @classmethod
+    def transform_fields(cls, values):
+        """This allows us to pass in the class directly instead of the string representation of the class. The
+        validator will convert the class to a string representation of the class."""
+        for field in list(cls.model_fields.keys()):
```
pmbrull, line 45: maybe some more thoughts on this:
1. if the only use case was about defining manifests as yaml/json -> do we even want to consider this at this point? maybe we can keep it simpler for first round and if the use case comes, then jump into it
2. if we end up needing this, some tests would be great

ingestion/src/metadata/utils/service_spec/service_spec.py
```diff
@@ -0,0 +1,91 @@
...
+                    service_type.name.lower(),
+                    get_module_dir(source_type),
+                    "service_spec",
+                )
+            )
+        )
+
+
+def get_class_path(module):
+    return module.__module__ + "." + module.__name__
```
pmbrull, line 78: maybe a good method to have on `importer` instead

ingestion/src/metadata/profiler/metrics/system/system.py
pmbrull: if this is only to keep a generic enough default, i'd remove the init 🤷

pmbrull: in any case noone would inherit this no? if we jump into non-SQA sources then they'd need their own class and whatnot

ingestion/src/metadata/profiler/metrics/system/system.py
sushi30: This is a dependency injection pattern that uses `super`. This method which is type-safe and deterministic. More information in [this video](https://www.youtube.com/watch?v=EiOglTERPEo).

```
                                                        
                                                        
              ┌───────────────────────┐                 
              │                       │                 
              │  DependencyInterface  │                 
              │                       │                 
              └──────▲───────▲────────┘                 
                     │       │                          
                     │       │                          
    ┌──────────────┬─┘       └──┬────────────────────┐  
    │              │            │                    │  
    │   Procedure  │            │ ConcreteDependency │  
    │              │            │                    │  
    └───────────▲──┘            └──▲─────────────────┘  
                │                  │                    
                │                  │                    
               ┌┴─────────────────┬┘                    
               │                  │         ▼           
               │  Implementation  │                     
               │                  │                     
               └──────────────────┘                     
```

ingestion/src/metadata/utils/importer.py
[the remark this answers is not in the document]
sushi30: this is leftover from development. its not used anywhere. deleted.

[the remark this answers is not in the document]
sushi30: Generally. We should start removing all these dynamic import and stay only with with `BaseSpec.get_for_source` or other variations of it that rely on `ServiceSpec`.

ingestion/src/metadata/utils/service_spec/service_spec.py
sushi30: This is what's being used de-facto in all our current OpenMetadata. Its tested in the integration and e2e tests. The only reason to use strings is to give developers the option to specify classes that are not part of the OpenMetadata module tree. We already do this for [CustomDatabaseConnection](https://github.com/open-metadata/OpenMetadata/blob/f0f74dbe5903bdace840320d68be5aa9c589a260/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/customDatabaseConnection.json#L24) but we allow this only for the `Metadata Source` class. This will allow us to allow importing classes for other workflows.

sushi30: In any case implemented some unit tests

sushi30:
@pmbrull [re __init](https://github.com/open-metadata/OpenMetadata/pull/18322#discussion_r1810432308)

I changed how the constructors are called to remove this. LMK if this addresses your comment.

ingestion/src/metadata/profiler/metrics/system/system.py
TeddyCr:
```
from metadata.mixins.sqalchemy.sqa_mixin import SQAInterfaceMixin
class MyClass(SQAInterfaceMixin):
    pass
```
But this is a MixIn right, so it is never intended to be used like this -- a MixIn will never be used in single inheritance? 

To get back to the initial question, why can't we have 

```
def get_system_metrics(self):
    inserts = self.get_inserts(...)
    deletes = self.get_deletes(...)
    updates = self.get_updates(...)
    return SystemMetrics()

@abastrcatclass
def get_inserts(self):
    raise NoteImplementedError

...
```
With strong interface children need to implement

ingestion/src/metadata/profiler/metrics/system/system.py
sushi30: It's required for using the `super()` dependency injection method. The one downside of this method. I'll add a comment for it.

sushi30: P.S: Something similar is used in [python's ABC class](https://github.com/python/cpython/blob/de0d5c6e2e12f24ade1ccc457afaf5fb2c650c64/Lib/abc.py#L106)

ingestion/src/metadata/profiler/metrics/system/system.py
sushi30: My goal was to enable extending functionality of classes and dependency injection while keeping the system type safe, deterministic and modularized. More specifically:

1. We have a common interface but we dont want to couple how we get the data (example: sqlalchemy) with the framework's logic. 
2. We have some functionality is required at the bottom of the diamond (see diagram). Notice how they all require the sql_alchemy session? With linear inheritance (like you suggested) all the implementing children will need to get this sql alchemy session even though some will not need it (or need another object). We are currently doing this using the `runner` object and the `table` object which are not type safe (the profiler needs to guess what object its getting).
3. Note how each system metrics source implements the cache provider with a different generic type and uses it differently. This framework allows us to create an independent state in each object but have common functionality. I dont see how this can be achieved with a `CacheMixin` but I am open to suggestions.
4. The solution we currently have in place is the `SQAInterfaceMixin` which is (1) not type-safe, (2) uses `self` which disables functionality changes downstream, any change requires changing the parent class and therefore **all the children** and it has no method of managing state as it relies on the state of the inheriting child.
5. This is the idea of dependency injection. Each implementing class takes what it needs to accomplish the task and is independent of how others are accomplishing it.

Regarding your example. I fail to notice how it achieves the following:
1. Allowing default cases for classes that do not need to implement the system profile interface (all profilers need to inherit it).
2. Allowing common functionality (example: using the `Metrics` objects) while allowing different methods for retrieving the data.
3. Enabling state for implementing children while unifying functionality. Using `utils` will not achieve this because as the `utils` module grows one needs to guess with operations are relevant for the object one is developing. This method limits the available functionality to the scope of the what's available for the object.

TeddyCr (approved):

ingestion/src/metadata/profiler/metrics/system/system.py
TeddyCr: That is a bit different. It overrides `__new__` not `__init__`. The idea, and what they do in ABC, is override `__new__` if you need to do something after class instantiation. When you override `__new__` you do need to call `super().__new__` so you instantiate the class -- otherwise you'll get `None` instance. 

```
 
class Foo:
    def __init__(self) -> None:
        self.a = 1
        self.b # 2

    def __new__(cls):
        cls = super().__new__(cls)
        cls.b = 2
        return cls

f = Foo()
f.a # 1
f.b # 2
```

Here we call `super().__init__(...)` but if we don't have that method here it will call the one of the parent.

ingestion/src/metadata/ingestion/source/database/redshift/manifest.py
pmbrull: hi let's stick with strict imports here. It adds a bit of more "boilerplate" when we're developing a new connector but removes magical layers from the code

PR #18333, opened 2024-10-21
Fix #18332: Do not consolidate changes for Parent Entities

### LINKED ISSUE #18332: Do not consolidate changes for Parent Entities
**Feature**
In OpenMetadata, we capture every change to a data asset and version it. This provides a great view for users, owners and data goverance teams to understand the evolution of a data asset , including who changed and when. This change event also triggers a thread for conversations.
We introduced a consolidation of entity versioning if the change happens with-in 10mins by the same user against the same data asset. This provides a way to reduce unnecessary noise and reduced the no.of versions an asset goes through.
However this puts an issue when a parent entity change is consolidated. Whenever there is a change to a data asset, we generate a change event which used to propagate changes to children entities. For example a change to classification to disable should disable all of its children tags in search index. 
However if the enable/disable happens with-in a session time by the same user only the first change is applied not the second change because of consolidation.

In this ticket, we will not apply change consolidation to parent entities.

Parent entities less likely to change unlike their children, example database metadata is less likely change compared to Table. 
So this exception for Parent entities should be ok in not having huge affect in entity versions.

harshach:
### Describe your changes:

Fixes <issue-number>

Backend : 
- Fix the tags not visible after re-enabling them from disabled state within a span of 10min from same user

UI : 
- Fix the dropdown not visible after search or when no tags are available

Issue :  

https://github.com/user-attachments/assets/8bc5906d-c20f-4843-87d1-bdee01d5895b

Resolved : 

https://github.com/user-attachments/assets/03c7d8af-dfd8-4a21-98d8-559d29aa3986

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/Entity.java
```diff
@@ -251,6 +252,20 @@ public final class Entity {
...
+            MLMODEL_SERVICE,
+            PIPELINE_SERVICE,
+            API_SERVICE,
+            STORAGE_SERVICE,
+            METADATA_SERVICE,
+            SEARCH_SERVICE,
+            DATABASE,
+            DATABASE_SCHEMA,
+            CLASSIFICATION,
+            GLOSSARY));
```
Ashish8689, line 268: @harshach API_COLLECTION and MESSAGING_SERVICE are missing in this list.

PR #18456, opened 2024-10-29
Fixes GEN-1260: Add Validators while creating table to escape special characters

ayush-shah:
### Describe your changes:

Fixes GEN-1260
This allows support for entities that have `::` or `>`
It is an incremental change that is shipped for Table, Columns ( for metadata, profiling and data quality )
Additionally it now supports names with quotes and spaces

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/models/custom_pydantic.py
```diff
@@ -30,13 +32,187 @@
...
+            Column,
+            Table,
+            TableConstraint,
+        )
+        from metadata.profiler.processor.core import ColumnProfile, ProfilerResponse
+        from metadata.profiler.source.metadata import ProfilerSourceAndEntity
+        from metadata.utils.entity_link import CustomColumnName
+
+        @singledispatch
+        def validate_name_and_transform(_):
```
pmbrull, line 75: what's the benefit of having the validation logic inside the `BaseModel`? We could define the methods outside and call them from here right? Or is it about circular imports?

ingestion/src/metadata/ingestion/models/custom_pydantic.py
```diff
@@ -30,13 +32,187 @@
...
+            """Single dispatch function to process name and display name based on object type."""
+
+        @validate_name_and_transform.register
+        def _(obj: CreateTableRequest):
```
pmbrull, line 79: one thing we could try is to not use a typed dispatch, but directly check the fields with `gettatr` -> We could have a list of fields for us to validate, or a few use cases and make something like:
```
if gettatr(obj, "columns"): # then check the names of the columns
if gettatr(obj, "fields"): # same for topics,
```
etc

pmbrull: might help us avoid this weird import order we have about BaseModel -> Table -> BaseModel

ingestion/src/metadata/ingestion/models/custom_pydantic.py
```diff
@@ -30,13 +32,187 @@
...
+                                constraint_column
+                            )
+
+        if not values:
+            return
+
+        try:
+
+            validate_name_and_transform(values)
+            # if class_name.startswith("Create") or class_name in CREATE_ADJACENT_MODELS:
```
pmbrull, line 193: if not needed let's clean the comment

ingestion/src/metadata/ingestion/models/custom_pydantic.py
```diff
@@ -30,13 +32,187 @@
 SECRET = "secret:"
 JSON_ENCODERS = "json_encoders"
 
+RESTRICTED_KEYWORDS = ["::", ">"]
+RESERVED_COLON_KEYWORD = "__reserved__colon__"
```
sushi30, line 36: What happens if a table or column has this identifier in the name? If we are not planning to support such a scheme its best to fail for entities that have these strings in the original names.

ingestion/src/metadata/ingestion/models/custom_pydantic.py
```diff
@@ -30,13 +32,187 @@
...
+        @singledispatch
+        def validate_name_and_transform(_):
+            """Single dispatch function to process name and display name based on object type."""
+
+        @validate_name_and_transform.register
+        def _(obj: CreateTableRequest):
+            """Process CreateTableRequest type."""
+            process_table_and_columns(obj, is_create=True)
+
+        @validate_name_and_transform.register
```
sushi30, line 83: Instead of implementing this on a per-entity basis might make more sense to hook into the serialization process and modify any `basic.EntityName`types.

When we add new entities in the future, one will need to explicitly add it here. Makes more sense to have this functionality baked in so that high-level implementation dont require any awareness to this logic.

pmbrull: that's a good one actually. @ayush-shah similar to the comment above. Regardless of the object we can traverse its schema, and if it's instance of EntityName we just convert the ::

ingestion/src/metadata/data_quality/api/models.py
sushi30, line 1: this is being skipped by the format ci but I believe this imports are not required

ingestion/src/metadata/data_quality/processor/test_case_runner.py
sushi30, line 1: whitespace change

ingestion/src/metadata/ingestion/models/custom_pydantic.py
```diff
@@ -37,6 +45,29 @@ class BaseModel(PydanticBaseModel):
...
+
+            if cls.__name__ in CREATE_ADJACENT_MODELS or cls.__name__.startswith(
+                "Create"
+            ):
+                values = validate_name_and_transform(values, replace_separators)
+            elif cls.__name__ in FETCH_MODELS:
+                values = validate_name_and_transform(values, revert_separators)
+
+        except Exception as exc:
+            logger.warning("Exception while parsing special characters: %s", exc)
```
sushi30, line 68: prefer to re-raise and crash. otherwise the system is in an unknown state.

ayush-shah: re raised and crashed, if it faces any error

sushi30 (approved):

sushi30 (approved):

PR #18493, opened 2024-11-02
[WIP] Fix Successfully Sent Change Events for Alerts For Debuggability API

Siddhanttimeline:
Retrieving a list of `changeEvent` entries that represent successfully sent change events from the `change_event` table, filtering these events based on an offset range, and ensuring they do not have corresponding failed entries in the `consumers_dlq` table. The offset range helps identify the starting point from which an alert began polling the changeEvent entries, ensuring that only relevant records are fetched for the alert process.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/changeEvent/AbstractEventConsumer.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/changeEvent/AbstractEventConsumer.java
@@ -55,6 +55,8 @@ public abstract class AbstractEventConsumer
   public static final String FAILED_EVENT_EXTENSION = "eventSubscription.failedEvent";
 
   private long offset = -1;
+  private long startingOffset = -1;
+
   private AlertMetrics alertMetrics;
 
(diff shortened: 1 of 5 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/events/subscription/EventSubscriptionResource.java
```diff
@@ -760,10 +770,16 @@ public Response getEventSubscriptionDiagnosticInfoById(
       @Context UriInfo uriInfo,
       @Context SecurityContext securityContext,
       @Parameter(description = "Maximum number of unprocessed events returned")
-          @DefaultValue("100")
+          @DefaultValue("15")
```
harshach, line 773: any reason for 15 specific?

harshach:
@Siddhanttimeline how are we going to make sure there are only 100 events ever in this table successful_sent_change_events. Otherwise it will continue to grow large sizes

mohityadav766 (approved):

PR #18556, opened 2024-11-08
Improve parallelization in SearchIndexApp

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

**Search Index Application**

https://github.com/user-attachments/assets/5e2fcca7-9649-4c23-9322-ee107281b0d3

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/SearchIndexApp.java
```diff
@@ -239,174 +268,252 @@ public void updateRecordToDb(JobExecutionContext jobExecutionContext) {
...
-                            .withMessage(e.getMessage()));
-                  }
-                  sendUpdates(jobExecutionContext);
-                  LOG.error("Unexpected error during reindexing for entity {}", entityType, e);
-                }
-              });
-
-      futures.add(future);
-    }
+    int numProducers = jobData.getEntities().size();
```
mohityadav766, line 285: @harshach number of producers can be 39 , if the user select all entities might not be a good idea to select this equal to entities count

harshach: what would be the issue with that?

pmbrull (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/SearchIndexApp.java
mohityadav766: for moderate operations it should be ok , but in cases where there are long running process which can be in our cases fixed thread pool doesn't scale well , and newCachedThreadPool is more suited

PR #18610, opened 2024-11-12
Create new Auto Classification Workflow

pmbrull:
### Describe your changes:

- We are separating the `sampler` out of the Profiler. Now, the sampler is an input parameter for both Profiler and Test Suite. It will help us have the new sampling workflow
- We are removing the factories for the sampler and test suite interface, and relying on the Service Spec instead
- New Auto Classification Workflow and implementation for the IngestionPipeline
- Removed `processPIISensitive`, `confidence` and `generateSampleData` from the profiler

I need to port the implementation for the ingestion to Argo in Collate https://github.com/open-metadata/openmetadata-collate/pull/923

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/data_quality/validations/runtime_param_setter/param_setter.py
```diff
@@ -15,22 +15,21 @@
...
+    You can see an example implementation in `TableDiffParamsSetter`.
     """
 
     def __init__(
         self,
         ometa_client: OpenMetadata,
         service_connection_config,
         table_entity: Table,
-        sampler: SQASampler,
+        sampler,
```
pmbrull, line 32: had to remove it due to a cyclic import

pmbrull: we can figure it out later

ingestion/src/metadata/data_quality/interface/sqlalchemy/sqa_test_suite_interface.py
```diff
@@ -49,23 +49,25 @@ def __init__(
...
-        self.table_entity = table_entity
-        self.service_connection_config = service_connection_config
+        super().__init__(
+            service_connection_config,
+            ometa_client,
+            sampler,
+            table_entity,
+        )
         self.create_session()
-        self._table = self._convert_table_to_orm_object(sqa_metadata)
```
TeddyCr, line 60: Have we kept the logic to pass the Metadata object when creating orm object. We need it to prevent naming conflict when creating the orm object if I well remember

ingestion/src/metadata/sampler/processor.py
```diff
@@ -0,0 +1,179 @@
...
+
+NON_SQA_DATABASE_CONNECTIONS = ("Datalake",)
+
+
+@lru_cache
+def _get_sqa_metadata(conn_type: str) -> Optional[MetaData]:
+    """Set sqlalchemy metadata"""
+    if conn_type not in NON_SQA_DATABASE_CONNECTIONS:
+        return MetaData()
+    return None
```
TeddyCr, line 55: Would this cache the same metadata object? If so we would need to need to remove the caching. We need to make sure we get a new Metadata() instance each time otherwise we get some conflicts in orm object names if I well remember.

Basically the metadata instance will hold the orm object. if you have the same table name but from different schema / database then you run into issues

ingestion/src/metadata/sampler/sampler_interface.py
```diff
@@ -0,0 +1,246 @@
...
+        sample_data_count: Optional[int] = SAMPLE_DATA_DEFAULT_COUNT,
+        **__,
+    ):
+        self.ometa_client = ometa_client
+        self._sample = None
+        self._columns: Optional[List[SQALikeColumn]] = None
+        self.sample_config = sample_config
+
+        if not self.sample_config.profile_sample:
+            self.sample_config.profile_sample = 100
```
TeddyCr, line 77: I think we should enforce sample type = percent if we'll do that.

TeddyCr (commented):
@pmbrull thanks for the PR. Left a few comments/questions. 

If I understand correctyl we won't ingest sample data from the profiler at all anymore?

ingestion/src/metadata/data_quality/interface/sqlalchemy/sqa_test_suite_interface.py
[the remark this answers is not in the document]
pmbrull: the ORM table is now an input argument https://github.com/open-metadata/OpenMetadata/pull/18610/files#diff-e8a8a48b444e6246572ad762618b52eed69b4b7a75ffba896cf8314f56c15990R132

Before we sent the `sqa_metadata` object and let the interfaces (profiler/test suite) create the ORM. Now, since the sampler needs the ORM table too, we create it at the `source` and pass it as an arg to the 3 interfaces

ingestion/src/metadata/ingestion/sink/metadata_rest.py
```diff
@@ -549,30 +550,16 @@ def write_life_cycle_data(self, record: OMetaLifeCycleData) -> Either[Entity]:
...
-        record.profile.columnProfile = column_profile
-
-        table = self.metadata.ingest_profile_data(
-            table=record.table,
-            profile_request=record.profile,
-        )
-        logger.debug(
-            f"Successfully ingested profile metrics for {record.table.fullyQualifiedName.root}"
-        )
-
```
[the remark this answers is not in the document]
pmbrull: this logic was never applied for the sample data 🤔 not sure if it would apply. Will add it just in case

ingestion/src/metadata/mixins/sqalchemy/sqa_mixin.py
```diff
@@ -52,19 +50,6 @@ def _get_engine(self):
...
-    ) -> DeclarativeMeta:
-        """Given a table entity return a SQA ORM object
-
-        Args:
-            sqa_metadata_obj: sqa metadata registry
-        Returns:
-            DeclarativeMeta
-        """
-        return ometa_to_sqa_orm(self.table_entity, self.ometa_client, sqa_metadata_obj)
-
```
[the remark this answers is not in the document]
pmbrull: correct, we just moved the ORM as an input arg https://github.com/open-metadata/OpenMetadata/pull/18610/files#diff-5e6d3e3344f81391a572db8c2cc3f48eb6ab298a0555ae18cb923d1cb7530b76R113

ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py
```diff
@@ -393,19 +375,6 @@ def _compute_system_metrics(
...
-        table,
-    ):
-        """Create thread safe runner"""
-        if not hasattr(thread_local, "sampler"):
-            thread_local.sampler = self._get_sampler(
-                table=table,
-                session=session,
-            )
-        return thread_local.sampler
-
```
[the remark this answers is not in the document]
pmbrull: we would need to test that tbh. I could use a hand validating if the error still exists. Since the sampler is an input arg now, does it change some internal state during the computation?

I guess before the issue was that we created a sampler object in each thread that needed to instantiate the same sample data. Right now it would not be the case right

ingestion/src/metadata/sampler/processor.py
pmbrull: I was trying to replicate this behavior https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/profiler/source/database/base/profiler_source.py#L78 of the `ProfilerSource` "freezing" a single MetaData object when it was instantiated. If I got that right, the ProfilerSource is created once at the DB Service level, and then each table would reuse the same MetaData object?

ingestion/src/metadata/sampler/sampler_interface.py
pmbrull: will check

ingestion/src/metadata/sampler/processor.py
pmbrull: otherwise, I am just as happy to remove the caching if there's no downside

ingestion/src/metadata/sampler/sampler_interface.py
pmbrull: the default for the sampling type is already percentage. Do you have some sane default in the case it's ROW?

ingestion/src/metadata/mixins/sqalchemy/sqa_mixin.py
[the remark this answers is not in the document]
TeddyCr: We can think about it for later, but keeping it with sqlalchemy logic may make sense as this is a pure SQA function.

ingestion/src/metadata/ingestion/sink/metadata_rest.py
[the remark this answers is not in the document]
TeddyCr: Sounds good. I don't think we have any BQ integration test directly in the codebase but you can check the `nested_struct` table in the `sdoh_cms_dual_eligible_enrollment` schema

TeddyCr (dismissed):

TeddyCr (approved):

PR #18660, opened 2024-11-15
ISSUE-16094: fix s3 storage parquet structureFormat ingestion

KylixSerg:
### Describe your changes:

Fixes 16094
Short blurb explaining:
This aims at fixing the s3 storage ingestion for parquet structured containers, when the ingestion flow tries to extract columns from a sample file under the dataPath it may happen that a non structureFormat file is picked, example a _SUCCESS file or that the prefix is empty but the s3 lifecycle hasnt kicked in yet, this will result in the column extraction understandably failing as the file cannot be loaded into a dataframe.

For that, do not fail when the column extraction fails but log a warning and continue with the remainder of the paths in the manifest, this is already an improvement on the status quo, but ideally when the column extraction fails, the ingestion flow should have a rety mechanism where it will rety a couple other files under the prefix if available before giving up, and moving on to another prefix.

#
### Type of change:

- [ X] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

- [X ] My PR title is `Fixes <issue-number>: <short explanation>`
- [X ] I have commented on my code, particularly in hard-to-understand areas. 


sushi30:
We need to choose one of:
 1. ignore failed files are
 2. allow for including/excluding patterns so that the user can define which files to skip.
 
The risk of performing the "catch return None" (1) is that when the user mis-configures the connector (example: chooses a directory with CSV files when they were looking for PARQUET files) they will receive a "successful" workflow with nothing ingested instead of an error.

We've seen reverse issues where users are complaining about successful workflows that have empty results and digging through the logs is not necessarily the best UX for these cases.   

Whichever solution we should stick to it throughout the platform as much as we can.

KylixSerg:
@sushi30 Thanks for checking this out!

imo I would avoid option 2, the issue I see with that approach from the way we're using OpenMetadata in my organization and that is probably similar to what a lot of folks might have setup, is that the ingestion pipeline (generating the manifests, configuring the connector...) is maintained by a different set of engineers than those publishing the data, with approach number 2, we will open up the door for users of the platform to introduce breaking changes to the connector's setup, ex: 

filter out any *_STATUS files(which in itself might be problematic, as it might require changes to a lot of pipelines in case there is no convention on naming files), then say the user uploads some other file like _READY, if the ingestion then picks up that file then it will break the ingestion job.

Furthermore, there might be setups(as it is in my use case) where even different teams are all publishing to the same bucket, therefore a single job for the same s3 bucket, making it so that if a user introduces a non conventional file name to their publishing location, that also might break it for other users publishing to the same bucket.

Personally I like option 1 even considering it's downside 

> We've seen reverse issues where users are complaining about successful workflows that have empty results and digging through the logs is not necessarily the best UX for these cases.

Because the responsibility of making the ingestion work would fall exclusively on the engineers maintaining the connection, generating the manifest.

sushi30:
Thank you @KylixSerg for taking the time to look it into this solution and provide the context. OpenMetadata aims to increase effective communication between teams. The issue you highlights the kind of discussions we prefer to be handled within the organization instead of letting the platform absorb the discrepancies. Object storage systems have fairly stable conventions in place (like the `_SUCCESS` file created by spark, using linux-like paths, or hive-like parititon directories) and teams will adopt some of them and might create some of their own. With OpenMetadata, teams are encouraged discuss and agree on the technical specifications that define their data pipelines and, as a result, how the metadata is consumed by the platforms (or others like it).

If multiple teams are publishing the same bucket, how are any other upstream processes reading the objects? I imagine there needs to be some coordination to avoid serving the wrong files to the wrong consumers.

As a counter-example, what happens if a process start dumping corrupted parquet files or CSV files instead of parquet? We expect our platform to be a participant in safe data practices and raise an error for this case rather than be complacent and pass the responsibility to upstream actions that ultimately will result in data we cannot trust.

## Suggested Implementation

Given these principal. I suggest the following implementation:
1. Add a global and/or local `containerFilterPatterns` to the [containerMetadataConfig schema](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/metadataIngestion/storage/containerMetadataConfig.json). This will be similar to our [filterPatterns](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/metadataIngestion/storage/containerMetadataConfig.json) used to filter with regular expressions.
2. Add the appropriate logic to exclude these files in the [S3Source class](https://github.com/open-metadata/OpenMetadata/blob/840a102887a44ecce08f6e86c74c36a8c1418168/ingestion/src/metadata/ingestion/source/storage/s3/metadata.py#L334). Once this logic is demonstrated for s3 we can create a separate delivery to generalize the concept for other object storage systems.

KylixSerg:
@sushi30 

> If multiple teams are publishing the same bucket, how are any other upstream processes reading the objects? I imagine there needs to be some coordination to avoid serving the wrong files to the wrong consumers.

The upstream process will break but only for that particular segment of the bucket, what OpenMetadata calls container, but it definitely does not break upstream for all users of a bucket, for example if there is a process that would create glue tables out of s3 prefixes, and that process requires a set of partitions on the prefix, then said process will surely not break because a single tenant of the bucket didn't know how to effectively use the platform, similarly I think a single tenant of OpenMetadata should  not have the ability to break the OpenMetadata integration.

> As a counter-example, what happens if a process start dumping corrupted parquet files or CSV files instead of parquet? We expect our platform to be a participant in safe data practices and raise an error for this case rather than be complacent and pass the responsibility to upstream actions that ultimately will result in data we cannot trust.

With your counter-example in mind, I do agree that the user should be aware of failures in the pipeline, like you laid out, bad parquet or a completely different file type than what was expected, but my ultimate point is that a multi-tenant setup should be setup so that one tenant's setup should not break it for other people using it.

I would suggest we go with your suggested implementation but with the extra point of setting it up so that we dont have single points of failure in the pipeline ie

3. Do not fail the whole ingestion job when a single container fails, make sure to properly highlight which containers failed

wdyt?

KylixSerg:
Hey @TeddyCr 

any updates on this?

sushi30:
@KylixSerg. That makes sense. You can see how a failed status is communicated in postgres [as an example](https://github.com/open-metadata/OpenMetadata/blob/026f39c5c79e773d419c92dcca67a8b91433b2de/ingestion/src/metadata/ingestion/source/database/postgres/metadata.py#L324). The workflow was some internal logic that fails a pipeline if the number of failed statuses are too large. With this change there is no need to implement the filter at this stage. You can add it to your code as it is only instead of ignoring the error, registering as a failure.

You can also join our [#contributor channel in slack](https://slack.open-metadata.org/) if you need any help in implementing this.

KylixSerg:
@sushi30 @TeddyCr Alright folks, I made the necessary changes to unify the approach of reporting failures - if you could please have a look 🙏

pmbrull:
awesome thanks @KylixSerg, bumped the branch to get the latest CI runs, let's see if everything passes properly. Thanks for the PR

PR #18640, opened 2024-11-14
MINOR: Add the possibility to have no Schedule and trigger the workflow manually

IceS2:
Adding the possibility to trigger the workflow manually.

#
### Type of change:

- [x] Improvement

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

PR #18664, opened 2024-11-15
MINOR: Improve PeriodicBatchEntityTrigger Workflows performance

IceS2:
In order to Increase the performance of the Flowable Workflows (Mainly for the **PeriodicBatchEntityTrigger**), we are:

- Updating the default Async Executor Flowable configurations to handle more workload
  The default configurations are pretty bad for an asset intensive production environment and thus we are updating them.

- Refactoring how we call the **MainWorkflow** from the **TriggerWorkflow**
  We are now relying on `CallActivity` to call the **MainWorkflow** instead of doing it programatically through a `JavaDelegate` on a `ServiceTask`.
  
  This is pretty important for the **PeriodicBatchEntityTrigger**, where we are now relying on `MultiInstanceLoopCharacteristics` for the concurrent task execution instead of dealing with it manually.

  It is a BPMN2.0 natural way of solving this issues that I didn't realized it would work for a whole workflow before.

  As a plus, it also takes care of waiting for the **MainWorkflow** to finish before finishing the **TriggerWorkflow**.

- Small fix to avoid patching an Entity if the updated Certification would be the same as the one they currently have.

#
### Type of change:

- [x] Improvement

#
### Checklist:


pmbrull (approved):

IceS2:
The test that is failing for the CI is being fixed here: https://github.com/open-metadata/OpenMetadata/pull/18640/files

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
```diff
@@ -40,6 +40,10 @@ private WorkflowHandler(OpenMetadataApplicationConfig config) {
     ProcessEngineConfiguration processEngineConfiguration =
         new StandaloneProcessEngineConfiguration()
             .setAsyncExecutorActivate(true)
+            .setAsyncExecutorCorePoolSize(50)
```
harshach, line 43: should these be configurable?

IceS2: I think they should, but we can iterate on that after having a decent default for production load.
I was thinking about having a `Workflows Setting` kind of thing within the OpenMetadata settings.
What do you think?

PR #18716, opened 2024-11-21
MINOR: ref(data-quality): modularized test case validator import

sushi30:
Modularizing test case importer so it can be overridden in Collate.

- removed test_suite_factory
- implemented TestCaseImporter
- removed SQAValidatorBuilder and PandasValidatorBuilder in favor of a SourceType enum
- removed the orm table creation from test suite source

waiting on:
- [x] https://github.com/open-metadata/OpenMetadata/pull/18641

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/data_quality/builders/i_validator_builder.py
TeddyCr, line 1: If we are moving away from independent builders implementing an interface for the client to a centralized one, let's rename the interface to `validator_builder.py` and `class ValidatorBuilder` -- since it is not an interface anymore

ingestion/src/metadata/profiler/orm/converter/base.py
```diff
@@ -156,7 +169,7 @@ def ometa_to_sqa_orm(
 
     if not isinstance(orm, DeclarativeMeta):
         raise ValueError("OMeta to ORM did not create a DeclarativeMeta")
-
+    global_orm_registry[table.fullyQualifiedName.root] = orm
```
TeddyCr, line 172: I think we should be cautious with keeping all form objects in memory (from what I understand this does)

ingestion/src/metadata/profiler/orm/converter/base.py
```diff
@@ -198,3 +211,22 @@ def get_orm_database(table: Table, metadata: OpenMetadata) -> str:
...
+        self, table: Table, service_conn_config: BaseModel, ometa_client: OpenMetadata
+    ) -> Optional[DeclarativeMeta]:
+        """Build the ORM table if needed for the sampler and profiler interfaces"""
+        if table.fullyQualifiedName.root in self.registry:
+            return self.registry[table.fullyQualifiedName.root]
+        if service_conn_config.type.value not in NON_SQA_DATABASE_CONNECTIONS:
+            orm_obj = ometa_to_sqa_orm(table, ometa_client)
+            self.registry[table.fullyQualifiedName.root] = orm_obj
+            return orm_obj
+        return None
```
TeddyCr, line 232: This used to be in the interface SQAMixIn. This is SQA specific, if we are going to move it again, let's move it back to the MixIn IMO where all SQA ops live.

sushi30: This requires a global state. do you want to add it to the SQA mixin? IMO the SQAMixin is becoming a [bloater](https://refactoring.guru/refactoring/smells/bloaters) by capturing all functionality related to to SQA and it should be broken down to more bit-sizes pieces.

If you have a concrete example of how to implement this in the `SQAMixin` along with the state please share so I can see what you mean.

ingestion/src/metadata/profiler/orm/converter/base.py
sushi30: I agree but could not a find a better solution for now. see [this comment](https://github.com/open-metadata/OpenMetadata/pull/18716/files/94619e24f57a4c87740c1dfb8b8dabf679405af1#diff-22074ecf4c9865187ab56afe336d18334aabdaea68c4abc37314f4cf9b33b7bbR217-R219) and LMK if you have another solution which is more scalable.

I am also not convinced that the current implementation is any better since it initialisez the orm at an early stage in the cycle so it might also be keeping them in memory. Was this explored via a load test?

ingestion/src/metadata/data_quality/builders/i_validator_builder.py
sushi30: done

ingestion/src/metadata/profiler/orm/converter/base.py
sushi30: moved to `SQAMixin`

ingestion/src/metadata/profiler/orm/converter/base.py
sushi30: using `sampler.table` within TestSuiteInterface

ingestion/src/metadata/data_quality/runner/base_test_suite_source.py
```diff
@@ -122,7 +113,6 @@ def create_data_quality_interface(self) -> TestSuiteInterface:
             source_config_type=self.service_conn_config.type.value,
```
TeddyCr, line 113: Can you move these to their own methods? `self.get_sampler` and ``self.get_validator_class`

TeddyCr: On the collate side we can then just modify these two

sushi30: The importer injection [is achieved here](https://github.com/open-metadata/openmetadata-collate/pull/923/files#diff-cb6c2264a7b5993fb58a79a09e87c3a662c3decde41d13a32dc4581b66fdfa62R34).

sushi30: and [here](https://github.com/open-metadata/openmetadata-collate/pull/923/files#diff-0fa83b3b608553c1bd66e2707132b47d4aed1ef4ec81f9ad61c0789db36b6416R18-R21)

PR #18729, opened 2024-11-21
Improvement : Teams api to update and remove users

sonika-shah:
### Describe your changes:

Improve performance of updating users in the team

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/TeamResource.java
```diff
@@ -666,6 +668,69 @@ public CsvImportResult importCsv(
     return importCsvInternal(securityContext, name, csv, dryRun);
   }
 
+  @PUT
```
harshach, line 671: @sonika-shah lets add a GET and remove users from fields for TeamResource, UI should make the call to get the user list with pagination

sonika-shah: image
yes @harshach , will have a separate api to GET the users list for a particular team with pagination, will also need the above PUT to update users in the team  (to avoid using the patch call to update users in a team)

PR #18732, opened 2024-11-21
Fix: mstr removed dependency issues

SumanMaharana:
### Describe your changes:

Fix: mstr removed dependency issues
mstrio-py had lot of dependencies which do not align with other OSS entity 
removed that dependency by using the requests lib

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/setup.py
+++ b/ingestion/setup.py
@@ -304,7 +304,7 @@ plugins: Dict[str, Set[str]] = {
         "psycopg2-binary",
         VERSIONS["geoalchemy2"],
     },
-    "mstr": {"mstr-rest-requests==0.14.1"},
+    "mstr": {},  # uses requests
     "sagemaker": {VERSIONS["boto3"]},
     "salesforce": {"simple_salesforce~=1.11"},
     "sample-data": {VERSIONS["avro"], VERSIONS["grpc-tools"]},
--- a/ingestion/src/metadata/ingestion/source/dashboard/mstr/client.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/mstr/client.py
@@ -117,6 +117,21 @@ class MSTRClient:
                 f"Failed to fetch the auth header and cookies due to [{exc}], please validate credentials"
             )
 
+    def close_api_session(self) -> None:
+        """
+        Closes the active api session
+        """
+        try:
+            close_api_session = self.client.post(
+                path="/auth/logout",
+            )
+            if close_api_session.ok:
+                logger.info("API Session Closed Successfully")
+
+        except Exception as exc:
+            logger.debug(traceback.format_exc())
+            logger.warning(f"Failed to close the api sesison due to [{exc}]")
+
     def is_project_name(self) -> bool:
         return bool(self.config.projectName)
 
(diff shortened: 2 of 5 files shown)


ingestion/setup.py
```diff
@@ -304,7 +304,7 @@
         "psycopg2-binary",
         VERSIONS["geoalchemy2"],
     },
-    "mstr": {"mstr-rest-requests==0.14.1"},
+    "mstr": {},  # uses requests
```
pmbrull, line 307: we can remove this then and clean the docs

ingestion/src/metadata/ingestion/source/dashboard/mstr/client.py
```diff
@@ -45,32 +46,92 @@ class MSTRClient:
...
+    def _get_auth_header_and_cookies(self) -> Optional[AuthHeaderCookie]:
+        """
+        Send a request to authenticate the user and get headers and
+        """
+        try:
+            data = {
+                "username": self.config.username,
+                "password": self.config.password.get_secret_value(),
+                "loginMode": 8,
+                "applicationType": 35,
```
pmbrull, line 80: we should document in the code itself the magic numbers (or better add a meaningful constant).

also if there any doc references from mstr we can add them here too

ingestion/src/metadata/ingestion/source/dashboard/mstr/client.py
```diff
@@ -45,32 +46,92 @@ class MSTRClient:
...
+            logger.debug(traceback.format_exc())
+            logger.error(
+                f"Failed to set user session due to [{exc}], please validate credentials"
+            )
+        except Exception as exc:
+            logger.debug(traceback.format_exc())
+            logger.error(
+                f"Failed to fetch the auth header and cookies due to [{exc}], please validate credentials"
+            )
+        return False
```
pmbrull, line 119: should this blow up if we cant set the session properly? will the ingestion work without this?

ingestion/src/metadata/ingestion/source/dashboard/mstr/client.py
```diff
@@ -125,30 +179,23 @@ def get_search_results_list(
...
-                },
+            data = {
+                "project_id": project_id,
+                "type": object_type,
+                "getAncestors": False,
+                "offset": 0,
+                "limit": -1,
+                "certifiedStatus": "ALL",
+                "isCrossCluster": False,
+                "result.hidden": False,
```
pmbrull, line 190: some info about this would help too

ingestion/src/metadata/ingestion/source/dashboard/mstr/client.py
pmbrull: then it should break IMO, not just return False. We need better exception management to know why things break at the right place

ingestion/src/metadata/ingestion/source/dashboard/mstr/client.py
pmbrull: Let's put constants, e.g., `LOGIN_MODE_POTATO = 8`

ingestion/src/metadata/ingestion/source/dashboard/mstr/client.py
SumanMaharana: yes right ill do the changes

ingestion/src/metadata/ingestion/source/dashboard/mstr/client.py
SumanMaharana: Added the constant

pmbrull (approved):

pmbrull (approved):

PR #18812, opened 2024-11-27
Revert "Add alert bar component"

chirag-madlani:
Reverts open-metadata/OpenMetadata#17230


aniketkatkar97 (approved):

ShaileshParmar11 (approved):

openmetadata-ui/src/main/resources/ui/src/styles/variables.less
```diff
@@ -110,10 +110,6 @@
 @team-avatar-bg: #0950c51a;
 @om-navbar-height: ~'var(--ant-navbar-height)';
 @sidebar-width: 60px;
-@error-bg-color: rgb(from @error-color r g b / 0.1);
-@success-bg-color: rgb(from @success-color r g b / 0.1);
```
Sachin-chaurasiya, line 114: These two variables are used in other places so can we keep them?

chirag-madlani: @Sachin-chaurasiya it's a revert we don't need changes which introduce as this PR. It will bring back previous changes which we need

Sachin-chaurasiya: @chirag-madlani 
POV: those variables are being used in `banner.less` file and now when I check it has two entries so good to go.

Sachin-chaurasiya (approved):

PR #18825, opened 2024-11-28
GEN-2157: Snowflake increased cost issue

harshsoni2024:
### Describe your changes:

Fixes <issue-number>

Here are some of the findings:
1) Simply maintaining an open connection with engine.connect() does not keep the warehouse active.
2) Only active queries or operations (like metadata queries or table scans) trigger warehouse usage.
- Snowflake warehouse has auto suspend/resume feature which is enable by default [doc](https://docs.snowflake.com/en/user-guide/warehouses-overview#auto-suspension-and-auto-resumption)

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -61,3 +66,33 @@ def get_stored_procedure_queries_dict(self) -> Dict[str, List[QueryByProcedure]]
...
+        """
+        for engine in self.get_engine():
+            rows = None
+            with engine.connect() as conn:
+                rows = conn.execute(
+                    self.get_sql_statement(
+                        start_time=self.start,
+                        end_time=self.end,
+                    )
+                )
```
ulixius9, line 83: do we not need to do fetch all?

harshsoni2024: yes it would be better since we are exiting the connection but not sure if count of rows is too large then it would cause issue or not !

ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -61,3 +66,33 @@ def get_stored_procedure_queries_dict(self) -> Dict[str, List[QueryByProcedure]]
...
+        """
+        for engine in self.get_engine():
+            rows = None
+            with engine.connect() as conn:
+                rows = conn.execute(
+                    self.get_sql_statement(
+                        start_time=self.start,
+                        end_time=self.end,
+                    )
+                ).fetchall()
```
pmbrull, line 83: this can kill the memory. Try server side cursors. example

```
with engine.connect() as conn:
    result = conn.execution_options(stream_results=True, max_row_buffer=100).execute(
        text("select * from public.employees")
    )
    counter = 0
    for row in result:
        counter += 1
print(counter)
```

pmbrull (changes requested):

pmbrull (approved):

PR #18839, opened 2024-11-28
MINOR - Fix sqa table reference

TeddyCr:
### Describe your changes:
- Use ORM table as the raw_dataset reference
- Explicitly reference dataset in column retrieval
- Remove aliased object creation

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Bug fix


ingestion/src/metadata/sampler/sqlalchemy/sampler.py
```diff
@@ -230,7 +229,7 @@ def _rdn_sample_from_user_query(self) -> Query:
 
     def _partitioned_table(self) -> Query:
         """Return the Query object for partitioned tables"""
-        return aliased(self.raw_dataset, self.get_partitioned_query().subquery())
+        return self.get_partitioned_query()
```
sushi30, line 232: Why is this method required? I suppose we can just use `self.get_partitioned_query()`

sushi30 (changes requested):

pmbrull (approved):

PR #18860, opened 2024-11-29
Minor: fixed connection modal style

ShaileshParmar11:
improved the spacing in connection component

image

image

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Modals/IngestionRunDetailsModal/IngestionRunDetailsModal.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Modals/IngestionRunDetailsModal/IngestionRunDetailsModal.tsx
@@ -11,7 +11,7 @@
  *  limitations under the License.
  */
 
-import { Button, Modal, Typography } from 'antd';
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/src/components/Modals/IngestionRunDetailsModal/IngestionRunDetailsModal.tsx
```diff
@@ -88,24 +88,28 @@ function IngestionRunDetailsModal({
...
+                  testConnectionStepResult={{
+                    name: failure.name,
+                    passed: false,
+                    mandatory: false,
+                    message: failure.error,
+                    errorLog: failure.stackTrace,
+                  }}
+                />
+              </Col>
+            )) ?? []}
```
Sachin-chaurasiya, line 111: @ShaileshParmar11,as we are returning the JSX I believe we don’t need this empty array as a fallback.

Moreover, I believe we should return null if there are no items to render.

Sachin-chaurasiya (approved):

PR #18864, opened 2024-12-01
Minor : Fix #18606 prevent overwriting style for PUT request

### LINKED ISSUE #18606: styling (color, icon) is lost if a glossaryTerm is updated via the bulk upload?
**Affected module**
UI

**Describe the bug**
If make changes to the glossaryTerm via upload to lose any applied styling (color, icon).

**Expected behavior**
**Or maintain style via import
Or don't change if uploaded via import**

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 1.5.4]

**Additional context**

sonika-shah:
### Describe your changes:

Fixes #18606

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -2924,6 +2925,14 @@ private static List<EntityReference> getEntityReferences(List<EntityReference> r
 
     private void updateStyle() {
       if (supportsStyle) {
+        Style originalStyle = original.getStyle();
+        Style updatedStyle = updated.getStyle();
+
+        if (originalStyle == updatedStyle) return;
+        if (operation == Operation.PUT && updatedStyle == null) {
```
harshach, line 2932: Lets not add these kind of conditions. PUT as definition requires the full payload.If someone is using PUT and. not passing all the required details then we will replace it.
What use-case are we trying to solve here

sonika-shah: @harshach  while importing glossary csv the existing style of the glossary terms are getting removed, 

while  importing we either create or update entity, while updating entity internally a PUT request is being sent , and it overrides the field for which it does not find value in `updated` entity.

harshach: @sonika-shah that shouldn't be the reason to add these kind of chcecks to allow specific thing to work.
The PR should be to load the attributes of glossaryTerm first and override with the import ones  but not allow PUT to override certain behavior

harshach:
@sonika-shah lets revert to the previous version. That looks better

PR #18877, opened 2024-12-02
MINOR - FQN encoding in ometa_api, TestSuite pipeline creation & serialization of test case results

pmbrull:
### Describe your changes:

Fixes <issue-number>

Solving three issues:
1. We were not encoding the FQN in some places in the ometa_api Mixins. This made the ingestion pipeline status update fail in the client
2. Pandas validator might return numpy types (e.g., for the `df[col].count()` test) for the passed/failed rows, which cannot be automatically serialized. The types should be `int`, so we're ensuring the cast there. I'd rather cast it generally than missing some test. 
3. If we tried to run a YAML passing an Ingestion Pipeline that was not yet created in the UI, the creation logic failed, since we were not properly referencing the TestSuite service type

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/tests/integration/datalake/conftest.py
```diff
@@ -84,7 +84,7 @@
         "sourceConfig": {
             "config": {
                 "type": "TestSuite",
-                "entityFullyQualifiedName": f'datalake_for_integration_tests.default.{BUCKET_NAME}."users.csv"',
+                "entityFullyQualifiedName": f'datalake_for_integration_tests.default.{BUCKET_NAME}."users/users.csv"',
```
pmbrull, line 87: adding the file in a directory to test the encoding properly

ingestion/src/metadata/data_quality/validations/base_test_handler.py
```diff
@@ -118,8 +118,8 @@ def get_test_case_result_object(  # pylint: disable=too-many-arguments
             failed_rows = (
                 failed_rows if failed_rows is not None else (row_count - passed_rows)
             )
-            test_case_result.passedRows = passed_rows
-            test_case_result.failedRows = failed_rows
+            test_case_result.passedRows = int(passed_rows)
+            test_case_result.failedRows = int(failed_rows)
             test_case_result.passedRowsPercentage = (passed_rows / row_count) * 100
             test_case_result.failedRowsPercentage = (failed_rows / row_count) * 100  # type: ignore
```
TeddyCr, line 124: should we cast these as float just in case?

TeddyCr (dismissed):

TeddyCr (approved):

PR #18904, opened 2024-12-03
[GEN-2187] fix(data-diff): added MD5 handling for bigquery

sushi30:
- added MD5 handling for bigquery
- use URL instead of Engine because it requires less steps and less prone to failure

Adding e2e test in separate PR

Fixes:

```python
  File "/usr/local/lib/python3.10/site-packages/data_diff/databases/bigquery.py", line 270, in _query_atom
    result = self._client.query(sql_code).result()
  File "/usr/local/lib/python3.10/site-packages/google/cloud/bigquery/job/query.py", line 1681, in result
    while not is_job_done():
  File "/usr/local/lib/python3.10/site-packages/google/api_core/retry/retry_unary.py", line 293, in retry_wrapped_func
    return retry_target(
  File "/usr/local/lib/python3.10/site-packages/google/api_core/retry/retry_unary.py", line 153, in retry_target
    _retry_error_helper(
  File "/usr/local/lib/python3.10/site-packages/google/api_core/retry/retry_base.py", line 212, in _retry_error_helper
    raise final_exc from source_exc
  File "/usr/local/lib/python3.10/site-packages/google/api_core/retry/retry_unary.py", line 144, in retry_target
    result = target()
  File "/usr/local/lib/python3.10/site-packages/google/cloud/bigquery/job/query.py", line 1630, in is_job_done
    raise job_failed_exception
google.api_core.exceptions.BadRequest: 400 No matching signature for operator < for argument types: BYTES, STRING. STRING and BYTES are different types that are not directly comparable. To write a BYTES literal, use a b-prefixed literal such as b'bytes value'. Supported signature: ANY < ANY at [1:101]; reason: invalidQuery, location: query, message: No matching signature for operator < for argument types: BYTES, STRING. STRING and BYTES are different types that are not directly comparable. To write a BYTES literal, use a b-prefixed literal such as b'bytes value'. Supported signature: ANY < ANY at [1:101]
```

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
```diff
@@ -83,9 +84,9 @@ def build_sample_where_clause(
     reduced_concat = reduce(
         lambda c1, c2: c1.concat(c2), sql_alchemy_columns + [literal(salt)]
     )
-    sqa_dialect = create_engine(
+    sqa_dialect = make_url(
```
TeddyCr, line 87: I forgot if you have access to the runner where this is called, if so you can pass the dialect directly from the session `session.bind.dialect.name` I believe

ingestion/src/metadata/profiler/orm/functions/md5.py
```diff
@@ -33,7 +31,12 @@ def _(element, compiler, **kw):
...
 
 
-@compiles(MD5, Dialects.MSSQL)
+@compiles(MD5, PythonDialects.MSSQL.value)
 def _(element, compiler, **kw):
     # TODO requires separate where clauses for each table
     return f"CONVERT(VARCHAR(8), HashBytes('MD5', {compiler.process(element.clauses, **kw)}), 2)"
+
+
+@compiles(MD5, PythonDialects.BigQuery.value)
```
TeddyCr, line 40: Should we check for all the major ones at least to see if the default way the md5 function is compiled works as intented?

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
sushi30: If you are referring to `self.runner_session` then I would rather avoid using that as it is a protected parameter and this validator does not make use of `sqlalchemy` except for compiling the where clause so I would avoid exposing it to interfaces it doesn't require.

ingestion/src/metadata/profiler/orm/functions/md5.py
TeddyCr: I think we only support it for 10 connectors, we can test it for all 10
- Trino
- Athena
- MySQL
- Hana
- Oracle

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
sushi30: It's a different dialect for `table1` and `table2`. I prefer to have a centralized point for getting the dialect instead of using different methods for each table.

TeddyCr (approved):

PR #18990, opened 2024-12-10
MINOR: Handle Flowable History Cleanup

IceS2:
This PR aims to do two different things:

- First of all, it aims to provide a good default for keeping the flowable historic tables nice and tidy to avoid them growing way too much
    This is done by setting the following config, with the default being cleaning history of processes after 7 days they ended.

    ```java
    processEngineConfiguration
        .setEnableHistoryCleaning(true)
        .setCleanInstancesEndedAfter(
            Duration.ofDays(
                workflowSettings.getHistoryCleanUpConfiguration().getCleanAfterNumberOfDays()));
    ```
- Second, it parametrizes the main configurations that should be defined depending on the different use cases
    This was done by setting a new SystemSetting: WorkflowSetting.
    The main thing I'm not sold on is how I'm updating the `ProcessEngine` after a setting change is done. I'd love any inputs since it seems it is the first time we are doing this.

### Type of change:

- [x] Improvement

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
@@ -2,6 +2,7 @@ package org.openmetadata.service.governance.workflows;
 
 import static org.openmetadata.service.governance.workflows.elements.TriggerFactory.getTriggerWorkflowId;
 
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


sushi30 (approved):
I suggest having a pub/sub mechanism in the `SystemRepository` allowing objects like the `WorkflowHandler` to listen to configuration changes. So `postUpdate` should call all the subscribers instead of explicit objects and `WorkflowHandler` can choose what to do with t he update based on the `SettingsType`. This will keep the dependency tree linear (`SystemRepository` -> `WorkflowHandler`). This might also be useful in the context of the Application framework.

But that can be done on another iteration.

mohityadav766 (approved):

PR #19057, opened 2024-12-14
Correct Imports for Class Converter Factory

mohityadav766:
### Describe your changes:

Fix Import for Mysql and Redshift Connection

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/mysqlConnection.json
+++ b/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/mysqlConnection.json
@@ -44,6 +44,7 @@
     "authType": {
       "title": "Auth Configuration Type",
       "description": "Choose Auth Config Type.",
+      "mask":  true,
       "oneOf": [
         {
           "$ref": "./common/basicAuth.json"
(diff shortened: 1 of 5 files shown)


openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/mysqlConnection.json
```diff
@@ -44,6 +44,7 @@
     "authType": {
       "title": "Auth Configuration Type",
       "description": "Choose Auth Config Type.",
+      "mask":  true,
```
pmbrull, line 47: do we need the `mask` at this level?

mohityadav766: it works without mask as well, i was checking at other connection sons we are adding it similarly
if doesn't make much sense we can remove it

mohityadav766: @pmbrull lmk wdyt

harshach (approved):

Issue #18806, opened 2024-11-27
Add Table Type filter to the main page


baohan1999:
**Is your feature request related to a problem? Please describe.**
For our organization, we have both table and view as table types, and each of the types will have a specific purpose for different teams. Analytics teams mostly care about the views, while other teams will need both tables and views, so it is quite inconvenient for us (vast number of users) to filter out the view/table by "Advanced Search" as is at the moment. It is also quite complicated for our users if they are not familiar with OM Advanced Search yet too while the frequency of this filter using is very high as I described above.
**AS-IS**
![image](https://github.com/user-attachments/assets/f3663251-b160-4a49-aa4d-83751d6242b1)

**Describe the solution you'd like**
We suggest that the filter for "Table Type" should be added to the main page, next to the filter for Data Assets/ Domain/ Owner/ Tag etc. 
![image](https://github.com/user-attachments/assets/d5100177-6b8b-410e-9183-0c279d1aceb0)

baohan1999:
cc @harshach @ayush-shah

harshach:
@karanh37 we wanted to drive this through searchSettings. Let me know what support you need from backend

karanh37:
@harshach - We already are getting aggregations in response. We will need allowedAggregations to provide a list from which can user can select from.

nqvuong1998:
Hi @harshach @karanh37 , will this feature release in 1.7.x?

smga3000:
Thank you for the feature request @baohan1999 it helps grow the project for everyone. In the meantime, the best way to keep track of all the latest updates is by bookmarking [OpenMetadata](https://github.com/open-metadata/OpenMetadata) with a star! And, if you haven’t already, I'd like to invite you to join our upcoming virtual community meetup next Wednesday. We’ll be sharing the OMD Q1 roadmap along with a community spotlight featuring real-world use cases from Wix. Hope to see you there!

chirag-madlani:
@aniketkatkar97 validate and close it

aniketkatkar97:
Since we show all the assets on the explore page upfront, we only keep the filters that are common to all the assets. Asset-specific filters (e.g. tableType for table assets) are still shown in the case of the explore view after searching for any text.

PR #19103, opened 2024-12-16
Search Settings with Ranking config

### LINKED ISSUE #13914: Enable search rank customization by making weights configurable and allowing custom properties to be used
In our organisation, the ranking of search results does not correspond with user expectations. There are at least two personas that use search: people that know an exact table name so they can lookup table information; but there are also personas that search for the best table for something while they don't know the exact table name. So, for example: someone searching for "booking" will always have some source table ranked at the top (even with 0 usage) while a post-processed table called fact_booking with high usage would be better suited for this person. 
I know that every org might have different requirements for the search ranking configuration; but currently the weights of entity ranking are hard coded.

I would like a way to optionally override/configure ES weights/factors per entity attribute, so that we can optimize the search for our use case. 

N/A

**Additional context**
Hardcoded weights code: https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TableIndex.java#L116
Slack conversation: https://openmetadata.slack.com/archives/C02B6955S4S/p1699503942769939

harshach:
### Describe your changes:

Fixes #13914

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/exception/SystemSettingsException.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/exception/SystemSettingsException.java
@@ -1,0 +1,12 @@
+package org.openmetadata.service.exception;
+
+import javax.ws.rs.core.Response;
+import org.openmetadata.sdk.exception.WebServiceException;
+
+public class SystemSettingsException extends WebServiceException {
+  private static final String ERROR_TYPE = "SYSTEM_SETTINGS_EXCEPTION";
+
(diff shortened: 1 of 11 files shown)


karanh37:
@harshach - We should also store default quick filters. Ref ticket - https://github.com/open-metadata/OpenMetadata/issues/18806

PR #19105, opened 2024-12-17
Update mask_literals_with_sqlparse function to handle multiple SQL st…

mkohei:
…atements

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/lineage/masker.py
+++ b/ingestion/src/metadata/ingestion/lineage/masker.py
@@ -38,11 +38,11 @@ def mask_literals_with_sqlparse(query: str):
     logger = get_logger()
 
     try:
-        parsed = sqlparse.parse(query)  # Parse the query
+        # Parse all statements in the query
+        parsed_statements = sqlparse.parse(query)
 
-        if not parsed:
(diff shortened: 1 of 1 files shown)


TeddyCr (approved):

mohittilala:
https://github.com/open-metadata/OpenMetadata/issues/19122#issuecomment-3776552975
> Hi [@mkohei](https://github.com/mkohei), parser is designed to only parse single sql at a time and lineage also considers first sql from multi-statement queries, and mostly, we process single sql at a time in lineage workflow from query history or csv file so I'm not sure if we need to add support for multi-statement queries here.
> 
> Can you share more about when this is required to be handled? The changes need to be done on parser side as well to support this if it's indeed a valid use-case.

@mkohei I reviewed the PR and noticed your implementation will not yield expected masked query results since parser doesn't support multi-statement parsing. Please confirm if we can close this PR?

mohittilala:
Closing this PR as per above comment.

PR #19127, opened 2024-12-18
Feature: Docker Host Retry

keshavmohta09:
### Describe your changes:

I implemented a retry mechanism for the service connection, replacing localhost with host.docker.internal if a connection error occurs.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/automations/runner.py
```diff
@@ -60,19 +63,40 @@ def _(
...
+    try:
+        connection = get_connection(request.connection.config)
+
+        host_port_str = str(request.connection.config.hostPort or "")
+        if "localhost" in host_port_str:
+            result = test_connection_fn(metadata, connection, request.connection.config)
+            raise_test_connection_exception(result)
+
+        test_connection_fn(
+            metadata, connection, request.connection.config, automation_workflow
```
ulixius9, line 83: does this mean we run the test connection twice in case of localhost and when no exception is thrown?

keshavmohta09: Yes, we need to do this to show the result of test_connection on the frontend

ulixius9 (approved):

PR #19207, opened 2025-01-02
Fixes #18816 -  Airbyte api endpoint

### LINKED ISSUE #18816: Airbyte v1.0 API endpoint changed
**Affected module**
Ingestion Framework

**Describe the bug**
I'm running Open Metadata on a Virtual Machine using Ubuntu 24.04 and using the provided docker-compose file. For now, I'm using all the provided services from the compose file except for the database, which is an external Postgres.
I wanted to create a connection between Open Metadata and our self hosted Airbyte service, but the test connection fails with the following error message:
`403 Client Error: Forbidden for url: https://<companydomain>//api/v1/workspaces/list`

Based on this documentation, in the newer versions of Airbyte, the endpoint reference has changed to [(Ref):](https://docs.airbyte.com/using-airbyte/configuring-api-access#step-3-operate-airbyte-via-api)
`/api/public/v1/workspaces`
So the `public` part is missing.

Can you please also double check the double slashes before the api keyword in the link. I haven't provided an additional one in the host column, but still have 2 in the endpoint link:
`//api/v1/workspaces/list`

I think the change has been introduced since v1.0 version of Airbyte, the older version probably still use the old endpoint.
I was able to hit the new endpoint with curl command:
`curl -v https://<companydomain>/api/public/v1/workspaces \
     -H "Content-Type: application/json" \
     -H "Authorization: Bearer <token>"`

**To Reproduce**

Print screen about the error message:
image

**Expected behavior**
Based on the correct host and credentials, the test connection should work by hitting the newer endpoint for newer Airbyte versions.

**Version:**
 - OpenMetadata version: 1.5.7
 - OpenMetadata Ingestion package version: built-in ingestion framework

SumanMaharana:
### Describe your changes:

Fixes #18816 

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/pipeline/airbyte/client.py
+++ b/ingestion/src/metadata/ingestion/source/pipeline/airbyte/client.py
@@ -20,6 +20,7 @@ from metadata.generated.schema.entity.services.connections.pipeline.airbyteConne
 from metadata.ingestion.ometa.client import REST, APIError, ClientConfig
 from metadata.utils.constants import AUTHORIZATION_HEADER, NO_ACCESS_TOKEN
 from metadata.utils.credentials import generate_http_basic_token
+from metadata.utils.helpers import clean_uri
 
 
 class AirbyteClient:
@@ -30,8 +31,8 @@ class AirbyteClient:
     def __init__(self, config: AirbyteConnection):
         self.config = config
         client_config: ClientConfig = ClientConfig(
-            base_url=str(self.config.hostPort),
-            api_version="api/v1",
+            base_url=clean_uri(self.config.hostPort),
+            api_version=self.config.apiVersion,
             auth_header=AUTHORIZATION_HEADER,
             auth_token=lambda: (NO_ACCESS_TOKEN, 0),
         )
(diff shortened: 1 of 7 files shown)


ingestion/src/metadata/ingestion/source/pipeline/airbyte/client.py
```diff
@@ -30,8 +31,8 @@ class AirbyteClient:
     def __init__(self, config: AirbyteConnection):
         self.config = config
         client_config: ClientConfig = ClientConfig(
-            base_url=str(self.config.hostPort),
-            api_version="api/v1",
+            base_url=clean_uri(self.config.hostPort),
+            api_version=self.config.apiVersion,
```
ulixius9, line 35: what are the other options user would input here instead of "api/v1"?

SumanMaharana: "api/public/v1"

ulixius9 (approved):

chirag-madlani (approved):

PR #19340, opened 2025-01-13
MINOR - createMany endpoint for test case resource

TeddyCr:
### Describe your changes:
- Added create many endpoint to test case resource to create `n` test cases in 1 request

**Benchmark Sequential creation vs createMany creation**
**Setup:** Create 100 test cases spread across `n` entities (e.g. if `n=2` then create 50 test cases in each entities). 

Picture 1

e.g. creating 100 test cases for 1 entities is ~30% faster on average with `createMany` than sequential API, and 99p is ~20% faster with createMany.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
```diff
@@ -652,6 +653,62 @@ public Response create(
     return Response.created(test.getHref()).entity(test).build();
   }
 
+  @POST
+  @Path("/createMany")
```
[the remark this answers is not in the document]
harshach: or just bulk
looking at some examples 
image

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java
```diff
@@ -298,6 +298,31 @@ public void createEntity(EntityInterface entity) {
     }
   }
 
+  public void createEntities(List<EntityInterface> entities) {
```
harshach, line 301: Lets tag this method as @Transaction

harshach: IMO , where we are are inserting into DB and calling this method should be wrapped in a method and tag with @transaction that way if ES fails it can rollback the Database

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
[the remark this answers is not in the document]
TeddyCr: That works too, I was looking at Zendesk API, they use createMany{Type}. bulk create works too. 👍

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java
harshach: Ignore above, looks like its handled

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -1953,6 +1955,42 @@ public void createEntity(String indexName, String docId, String doc) {
...
+                    "Failed to create entities in ElasticSearch: {}",
+                    bulkItemResponses.buildFailureMessage());
+              } else {
+                LOG.debug("Successfully created {} entities in ElasticSearch", docsAndIds.size());
+              }
+            }
+
+            @Override
+            public void onFailure(Exception e) {
+              LOG.error("Failed to create entities in ElasticSearch", e);
```
harshach, line 1987: we may want to throw the exception back here to let the transaction know to revert the DB if there is unrecoverable exception

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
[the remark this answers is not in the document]
harshach: ahh ok, either one works, we can keep this

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
TeddyCr: @harshach I try to implement and did a bit of investigation as the ops don't get rolled back (more details [here](https://github.com/open-metadata/OpenMetadata/issues/19369)). Let me know if you have any thoughts about it.

PR #19366, opened 2025-01-15
Fix #18834: Introduce new setting OpenMetadataBaseUrlConfiguration to set the server endpoint

### LINKED ISSUE #18834: Slack Entity URLs Not Working If OpenMetadata URL Is Not Set in SMTP UI
**Affected module**
- Alerts and Notifications

 **Describe the bug**

- The URLs for entities in Slack messages are dependent on the OpenMetadata URL configured in the SMTP UI. If the OpenMetadata URL is not set, the entity URLs in the Slack messages fail to work.

**To Reproduce**
- Ensure that the OpenMetadata URL is not set in the SMTP UI.
- Send a message to Slack containing an entity link.
- Click on the entity link in Slack.
- Observe that the link does not work and the entity cannot be accessed.

Siddhanttimeline:
### Describe your changes:

Fixes #18834

1. Removed OM Base URL configuration from smtpSettings.
2. Introduced a new OpenMetadataBaseUrlConfiguration to set the server endpoint.

https://github.com/user-attachments/assets/6f1ea1d3-b8d4-4d71-8f9d-a7a66e1ad72f

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


mohityadav766:
@Siddhanttimeline also you need to update openmetadata_test.yaml for the same

openmetadata-service/src/main/java/org/openmetadata/service/util/email/EmailUtil.java
```diff
@@ -440,6 +440,15 @@ public static Boolean isValidEmail(String email) {
     return email.matches("^[\\w-\\.]+@([\\w-]+\\.)+[\\w-]{2,4}$");
   }
 
+  public static String getOMBaseURL() {
```
mohityadav766, line 443: we need to add this global config in the yaml as well 
as when we configure cluster this isn't a step that the user needs to be doing

mohityadav766 (changes requested):
please check comments

openmetadata-service/src/main/java/org/openmetadata/service/util/email/EmailUtil.java
Siddhanttimeline: Noted, addressing.

conf/openmetadata.yaml
```diff
@@ -383,6 +382,9 @@ email:
   transportationStrategy: ${SMTP_SERVER_STRATEGY:-"SMTP_TLS"}
   templates: ${TEMPLATES:-"openmetadata"}
 
+openMetadataBaseUrlConfiguration:
```
mohityadav766, line 385: @harshach would it make sense to create something like below
global: 
    serverBaseUrl: {}
    other configs: {} 
or we should keep it separate? 
cc @Siddhanttimeline

harshach: @mohityadav766 @Siddhanttimeline this shouldn't be part of server configs, we should let them configure in system settings and provide openmetadata-ops.sh to configure to help with devops automation

mohityadav766 (approved):

PR #19376, opened 2025-01-15
fix the api should not called after time out in Test connection

Ashish8689:
### Describe your changes:

Fixes <issue-number>

- fix the API calls that were being made even after the timeout of the Test Connection.
- Issue arrived due to the trying to use` variable reference` being passed to the function and assigning the interval id there. Which doesn't work like it.

Issue : 

https://github.com/user-attachments/assets/51282a26-b267-4971-bcb5-999a4a0fc7c8

Resolved : 

https://github.com/user-attachments/assets/c2d9ed8d-6106-4354-a459-13f92799c71d

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/components/common/TestConnection/TestConnection.test.tsx
```diff
@@ -331,6 +331,9 @@ describe('Test Connection Component', () => {
     expect(
       screen.getByText('message.test-connection-taking-too-long')
     ).toBeInTheDocument();
+
+    // 59 since it will make this amount of call, and after timeout it should not make more api calls
+    expect(getWorkflowById).toHaveBeenCalledTimes(59);
```
ShaileshParmar11, line 336: i think we should also add negative assertion here
expect(getWorkflowById).not.toHaveBeenCalledTimes(60);

Ashish8689: updated!

ShaileshParmar11 (approved):

PR #19423, opened 2025-01-17
FIX #19386 & #19388: Fixing Data Insights index mapping

### LINKED ISSUE #19386: Fix Mappings for Data Insights
Since for Data Insights we are using one Data Stream for each Entity Type, we should not rely on dynamic mappings for most of the attributes.
We should have static mappings for each entity.

As part of this task we also need to expose a command to devops, to be able to recreate the indexes and backfill the data.

### LINKED ISSUE #19388: Reduce the amount of columns we are indexing for Data Insights
We need to take a look at the columns we are indexing on Data Insights in order to reduce it by removing the ones that don't really make sense.

IceS2:
This PR aims to solve three things:
1. First and foremost, it aims to make the Data Insights indexes more reliable, by using more static mapping, in order to rely less on the dynamic mapping.
2. Meanwhile it also exposes a command for OpenMetadataOps to recreate and backfill Data Insights indexes.
3. Finally, it aims to reduce the amount of data that we are indexing, by reducing the amount of attributes in the data insights documents.

Fix #19386 
Fix #19388 

#
### Type of change:

- [x] Improvement

#
### Checklist:


openmetadata-service/src/main/resources/dataInsights/config.json
TeddyCr, line 1: Basically if we want to add new fields/metrics (or support for a new entity) we'll need to add those to this file, is that correct?

IceS2: In order to support a new entity actually we have a list on the Java Application itself (It'd be a good idea to remove it and basically support based on the list)

But yes, in order to add any new fields, we'd need to add it there. That way we'd control the amount of data indexed and also we're fetching the mapping that we define in our not-data-insights indexes to reduce the amount of dynamic mapping

openmetadata-service/src/main/java/org/openmetadata/service/util/OpenMetadataOperations.java
```diff
@@ -513,6 +517,104 @@ private int executeSearchReindexApp(
     return result;
   }
 
+  @Command(
```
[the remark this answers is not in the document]
IceS2: It is already there:

      @Option(
              names = {"--start-date"},
              description = "Start Date to backfill from.")
          String startDate,
      @Option(
              names = {"--end-date"},
              description = "End Date to backfill to.")
          String endDate) {

mohityadav766 (approved):

PR #19476, opened 2025-01-22
fix the richTextEditor output in case on no data save

Ashish8689:
### Describe your changes:

Fixes <issue-number>

- Issue : On any Description update, when data is being cleared and save it used to have empty `<p></p>` tag in the editor response.
- I worked on fixing the richTextEditor output to be undefined in case of all data is cleared.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


karanh37 (approved):

openmetadata-ui/src/main/resources/ui/src/components/common/RichTextEditor/RichTextEditor.tsx
```diff
@@ -47,7 +47,7 @@ const RichTextEditor = forwardRef<EditorContentRef, RichTextEditorProp>(
         const htmlContent = editorRef.current?.editor?.getHTML() ?? '';
         const backendFormat = formatContent(htmlContent, 'server');
 
-        return backendFormat;
+        return backendFormat === '<p></p>' ? '' : backendFormat;
```
chirag-madlani, line 50: I feel we can control this inside formatContent itself instead having condition on component end

PR #19487, opened 2025-01-23
Fix #19393 :postgres parse json schemas

### LINKED ISSUE #19393: Json parsing in huge tables in Postgres is a source of important performance problems
**Affected module**
It affects the Ingestion Framework.

**Describe the bug**

After upgrading to OpenMetadata >= 1.6.0, the metadata ingestion process for some of our Postgres databases has become really resource-hungry for the databases, to an extent of taking down one of the databases during the ingestion process execution.

After debugging the problem, we have found that the root cause is the following query (https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/database/postgres/queries.py#L215):

```
    WITH RECURSIVE json_hierarchy AS (
        SELECT
            key AS path,
            json_typeof(value) AS type,
            value,
            json_build_object() AS properties,
            key AS title
        FROM
            {table_name} tbd,
            LATERAL json_each({column_name}::json)
    ),
    build_hierarchy AS (
        SELECT
            path,
            type,
            title,
            CASE
                WHEN type = 'object' THEN
                    json_build_object(
                        'title', title,
                        'type', 'object',
                        'properties', (
                            SELECT json_object_agg(
                                key,
                                json_build_object(
                                    'title', key,
                                    'type', json_typeof(value),
                                    'properties', (
                                        CASE
                                            WHEN json_typeof(value) = 'object' THEN
                                                (
                                                    SELECT json_object_agg(
                                                        key,
                                                        json_build_object(
                                                            'title', key,
                                                            'type', json_typeof(value),
                                                            'properties',
                                                            json_build_object()
                                                        )
                                                    )
                                                    FROM json_each(value::json) AS sub_key_value
                                                )
                                            ELSE json_build_object()
                                        END
                                    )
                                )
                            )
                            FROM json_each(value::json) AS key_value
                        )
                    )
                WHEN type = 'array' THEN
                    json_build_object(
                        'title', title,
                        'type', 'array',
                        'properties', json_build_object()
                    )
                ELSE
                    json_build_object(
                        'title', title,
                        'type', type
                    )
            END AS hierarchy
        FROM
            json_hierarchy
    ),
    aggregate_hierarchy AS (
        select
            json_build_object(
            'title','{column_name}',
            'type','object',
            'properties',
            json_object_agg(
                path,
                hierarchy
            )) AS result
        FROM
            build_hierarchy
    )
    SELECT
        result
    FROM
        aggregate_hierarchy;
```

Which was introduced by https://github.com/open-metadata/OpenMetadata/pull/17645 (Fixes: Parse postgres json column fields) in OpenMetadata  1.6.0. 

The query aim is to extract the fields contained in the json columns. The problem is that the query is really expensive when you have tables with a large number of rows (we have some tables with json columns that contains tens of millions of rows), as it processes each row json content. 

For tables with millions of rows, this translates to a query using GBs of memory and lots of CPU. So after upgrading to OpenMetadata 1.6.0 we have moved from a process that was quite lightweight, basically getting the needed metadata from Postgres internal tables, to a process that is processing large volumes of data, provoking huge performance problems to the databases.

We'd expect this kind of functionality to be introduced as an option (opt-in) or at least to have a mechanism to disable it (which we have not found it after reviewing the PR and the OpenMetadata documentation and codebase). 

**To Reproduce**

To reproduce the problem you need a Postgres database, and create a table with a json/jsonb column, filling it with a dataset of some millions of rows.

When executing the metadata ingestion process using  `openmetadata-ingestion[postgres]>=1.6.0.0`, monitoring the database (CPU, memory, etc.), you can see that the ingestion process, otherwise quite lightweight, is really resource-hungry, taking a significant amount of resources and time.

**Expected behavior**

The functionality is working fine for small tables, but the approach is really problematic for huge tables. I'd expect this functionality to be disabled by default ("opt-in") or at least to have a way to disable it.

**Version:**
 - OS: Linux (Alpine)
 - Python version: 3.10
 - OpenMetadata version: >= 1.6.0
 - OpenMetadata Ingestion package version: `openmetadata-ingestion[postgres]>=1.6.0.0`

SumanMaharana:
### Describe your changes:

Fixes #19393 : json column parsing in postgres causes performance issues.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


TeddyCr (approved):
@SumanMaharana will we reimplement it with a more efficient approach?

SumanMaharana:
> @SumanMaharana will we reimplement it with a more efficient approach? Will it break anything / limit some information available for JSON field in postgres?

@TeddyCr 
will we reimplement it with a more efficient approach? yes
Will it break anything / limit some information available for JSON field in postgres? Not break anything just now the json columns will be shown as column type `json` directly.
1. if we parse those json schemas through taking a sample we are not sure if the sample contains all the json schemas also the select query will still take up time as we will have to query each json column and then parse that column. suppose we have 3000 tables containing a single json column the query running time is estimated to be about `1 hr 40 mins`.
2. This wasn't asked by any users and was only limited to postgres so removed those changes for now as its creating performance issues.

TeddyCr (approved):

PR #19497, opened 2025-01-23
fix the glossary column width sizes for the resizable columns

Ashish8689:
### Describe your changes:

- I worked on fixing the glossary column width sizes to the previous one, before resizing was added to it.
- Used the table width and then calculated the column width based on the percentage we previously used.

Issue : 

https://github.com/user-attachments/assets/ff2fec4c-f793-4f22-ac71-ef16a824a0fe

Resolved : 

https://github.com/user-attachments/assets/65161fd2-756c-4c02-83a9-205cc19f2639

https://github.com/user-attachments/assets/071113eb-9c91-4450-b721-45c951e1f3a8

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/components/Glossary/GlossaryTermTab/GlossaryTermTab.component.tsx
```diff
@@ -796,6 +821,12 @@ const GlossaryTermTab = ({
     return expandedRowKeys.length === expandableKeys.length;
   }, [expandedRowKeys, expandableKeys]);
 
+  useEffect(() => {
+    if (tableRef.current) {
+      setTableWidth(tableRef.current.offsetWidth);
+    }
+  }, []);
```
chirag-madlani, line 828: Do we need this?

openmetadata-ui/src/main/resources/ui/src/components/Glossary/GlossaryTermTab/GlossaryTermTab.component.tsx
```diff
@@ -147,6 +158,20 @@ const GlossaryTermTab = ({
...
+      description: calculatePercentageFromValue(
+        tableWidth,
+        permissions.Create ? 21 : 33
+      ),
+      reviewers: calculatePercentageFromValue(tableWidth, 33),
+      synonyms: calculatePercentageFromValue(tableWidth, 33),
+      owners: calculatePercentageFromValue(tableWidth, 17),
+      status: calculatePercentageFromValue(tableWidth, 12),
+    };
+  }, [permissions.Create, tableWidth]);
```
chirag-madlani, line 173: Move this to a util to get columns width

openmetadata-ui/src/main/resources/ui/src/components/Glossary/GlossaryTermTab/GlossaryTermTab.component.tsx
Ashish8689: Yes, we do since we are setting the column width based on table width which we are getting from Table Reference.  And changes in the ref value, don't re-render the document. So to make the change work have used `useEffect`, since it will trigger after a render and till then we will have tableRef value.

chirag-madlani: You can use `tableRef.current` as deps

Ashish8689: Have tried that, but have not worked.

chirag-madlani (approved):

PR #19510, opened 2025-01-24
GEN -19341 :  Improvement  -  Avoid usage of CONCAT in WHERE clause

### LINKED ISSUE #19341: Use immutable string concatenation in CollectionDAO to preserve indices
The use of CONCAT below breaks the index on tag_usage and possibly elsewhere within CollectionDAO.java. This is because CONCAT is not an immutable function and thus forces that query in getTagsByPrefix to do a sequential scan of the block rather than scan the index on targetFQNHash. Concatenation via the || operator is immutable and preserves the index. I suspect there are other cases of this performance risk based on the broad usage of CONCAT in the DAO file. 

https://github.com/open-metadata/OpenMetadata/blob/21dd3e4342225e559a48e74f4366e25bf9a1f3f7/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java#L3198

For example:

```
WHERE
  combined_data.targetFQNHash LIKE CONCAT('e5f0d95794bf67e9a3b2ee1fbaa57334.58b2896ff829ab76c7ca559cfabe6e85','.82ee2627aa49f798f3eb5df64a60');
```

The CONCAT filter produces the following plan:

```
Append  (cost=0.55..699.91 rows=2 width=400)
   ->  Subquery Scan on "*SELECT* 1"  (cost=0.55..14.74 rows=1 width=228)
         ->  Nested Loop  (cost=0.55..14.73 rows=1 width=2391)
               ->  Index Scan using tag_usage_source_tagfqnhash_targetfqnhash_key on tag_usage tu  (cost=0.41..6.18 rows=1 width=262)
                     Index Cond: (source = 1)
                     Filter: ((targetfqnhash)::text ~~ concat('e5f0d95794bf67e9a3b2ee1fbaa57334.58b2896ff829ab76c7ca559cfabe6e85', '.82ee2627aa49f798f3eb5df64a60'))
               ->  Index Scan using index_glossary_term_entity_deleted on glossary_term_entity gterm  (cost=0.14..8.16 rows=1 width=548)
                     Index Cond: ((fqnhash)::text = (tu.tagfqnhash)::text)
   ->  Subquery Scan on "*SELECT* 2"  (cost=0.00..685.16 rows=1 width=571)
         ->  Nested Loop  (cost=0.00..685.15 rows=1 width=2734)
               Join Filter: ((ta.fqnhash)::text = (tu_1.tagfqnhash)::text)
               ->  Seq Scan on tag_usage tu_1  (cost=0.00..677.66 rows=1 width=262)
                     Filter: ((source = 0) AND ((targetfqnhash)::text ~~ concat('e5f0d95794bf67e9a3b2ee1fbaa57334.58b2896ff829ab76c7ca559cfabe6e85', '.82ee2627aa49f798f3eb5df64a60')))
               ->  Seq Scan on tag ta  (cost=0.00..6.66 rows=66 width=441)
```

whereas the || operator is included in the index condition:

```
WHERE
  combined_data.targetFQNHash LIKE 'e5f0d95794bf67e9a3b2ee1fbaa57334.58b2896ff829ab76c7ca559cfabe6e85' || '.82ee2627aa49f798f3eb5df64a60';
```

```
Append  (cost=0.55..30.12 rows=2 width=400)
   ->  Subquery Scan on "*SELECT* 1"  (cost=0.55..14.74 rows=1 width=228)
         ->  Nested Loop  (cost=0.55..14.73 rows=1 width=2391)
               ->  Index Scan using tag_usage_source_tagfqnhash_targetfqnhash_key on tag_usage tu  (cost=0.41..6.18 rows=1 width=262)
                     Index Cond: ((source = 1) AND ((targetfqnhash)::text = 'e5f0d95794bf67e9a3b2ee1fbaa57334.58b2896ff829ab76c7ca559cfabe6e85.82ee2627aa49f798f3eb5df64a60'::text))
                     Filter: ((targetfqnhash)::text ~~ 'e5f0d95794bf67e9a3b2ee1fbaa57334.58b2896ff829ab76c7ca559cfabe6e85.82ee2627aa49f798f3eb5df64a60'::text)
               ->  Index Scan using index_glossary_term_entity_deleted on glossary_term_entity gterm  (cost=0.14..8.16 rows=1 width=548)
                     Index Cond: ((fqnhash)::text = (tu.tagfqnhash)::text)
   ->  Subquery Scan on "*SELECT* 2"  (cost=8.45..15.37 rows=1 width=571)
         ->  Hash Join  (cost=8.45..15.36 rows=1 width=2734)
               Hash Cond: ((ta.fqnhash)::text = (tu_1.tagfqnhash)::text)
               ->  Seq Scan on tag ta  (cost=0.00..6.66 rows=66 width=441)
               ->  Hash  (cost=8.43..8.43 rows=1 width=262)
                     ->  Index Scan using tag_usage_targetfqnhash_tagfqn on tag_usage tu_1  (cost=0.41..8.43 rows=1 width=262)
                           Index Cond: ((targetfqnhash)::text = 'e5f0d95794bf67e9a3b2ee1fbaa57334.58b2896ff829ab76c7ca559cfabe6e85.82ee2627aa49f798f3eb5df64a60'::text)
                           Filter: (((targetfqnhash)::text ~~ 'e5f0d95794bf67e9a3b2ee1fbaa57334.58b2896ff829ab76c7ca559cfabe6e85.82ee2627aa49f798f3eb5df64a60'::text) AND (source = 0))
```

sonika-shah:
### Describe your changes:

Fixes #19341 

- Introduced new annotation to perfom the CONCAT operation,  to avoid using concatenation in the WHERE clause, as it can prevent the use of indexes and lead to full table scans. Full details are mentioned in the issue here: [#19341](https://github.com/open-metadata/OpenMetadata/issues/19341).

For postgresql, it seems to show better performance when use without CONCAT clause but for Mysql it showed similar performance if the correct indexes are in place

Other approaches tried - 
	1.	`Using other means for string concat operation.`
	•	For postgresql we directly have || operator to do concatenation , but in mysql there is no direct way of doing string concatenation , we can only use + operator only when PIPES_AS_CONCAT mode is enabled
	2.	`Precompute the final concatenated strings :`
	•	While this improves performance, it makes code verbose and harder to maintain as it is done at repository level.

So to maintain consistency across codebase introduced  the below new annotation to handle concatenation logic at the DAO level instead of repository level:
		`@BindConcat`: For general concatenation of multiple input strings and string variables., with 2 optional field original and hash 

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -746,20 +748,27 @@ void insertMany(
...
             + "FROM entity_extension "
-            + "WHERE extension LIKE CONCAT(:extensionPrefix, '%') "
+            + "WHERE extension LIKE :extension "
             + "ORDER BY id, extension")
     @RegisterRowMapper(ExtensionWithIdAndSchemaRowMapper.class)
     List<ExtensionWithIdAndSchemaObject> getExtensionsByPrefixBatch(
-        @Bind("extensionPrefix") String extensionPrefix);
+        @BindConcat(
+                value = "extension",
+                parts = {":extensionPrefix", ".%"})
```
mohityadav766, line 770: are we adding a . extra ?

sonika-shah: yes there's an extra dot here , thanks for pointing out

openmetadata-service/src/main/java/org/openmetadata/service/util/jdbi/BindConcat.java
```diff
@@ -0,0 +1,74 @@
...
+                      + (placeholderIndex + 1)
+                      + " but got "
+                      + placeholders.length);
+            }
+            String dynamicValue = placeholders[placeholderIndex++];
+            if (dynamicValue == null) {
+              containsNull = true;
+              break;
+            }
+            concatenatedResult.append(dynamicValue);
```
mohityadav766, line 59: we can supply some conditional var in @Bind contract and based on that if hashing is require hash it or use plain value

mohityadav766: add some boolean in contract , hash = true/ false

PR #19554, opened 2025-01-28
[WIP] Allow client restart if required

mohityadav766:
### Describe your changes:

- Allow client restart if required

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach:
@mohityadav766 lets not add this. We shouldn't be introducing this type of change. it won't help us catch actual bug and fix it.

mohityadav766:
> @mohityadav766 lets not add this. We shouldn't be introducing this type of change. it won't help us catch actual bug and fix it.

Coming up with the actual change , need to introduce ClientHttpClirntAshnc with some Safe consumer class(as per the es docs) , upgrading es clients will fix this , but that would be bigger change

harshach:
lets not worry about this change. Lets only introduce the metric to monitor it

harshach:
@mohityadav766 closing this out, we need to find a root cause not having to do this with an API call

PR #19574, opened 2025-01-28
Alert Fixes

### LINKED ISSUE #19685: Alert Issues
**Affected module**

**Describe the bug**
- Error or exception in polling and publish does the offset to move ahead 
- we are making unnecessary calls to the database for getting initial and starting offset
- diagnostic information needs to be added on the UI
- allow the fixes or syncing from alerts page in case the publisher falls short

**To Reproduce**

**Expected behavior**

**Version:**
 - Python version:

**Additional context**

mohityadav766:
### Describe your changes:
Fixes  #19685

- Handle poll in try/catch , and error 

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/changeEvent/AbstractEventConsumer.java
```diff
@@ -256,18 +256,34 @@ public List<ChangeEvent> pollEvents(long offset, long batchSize) {
...
+      batchSize = batch.size();
+      eventsWithReceivers.putAll(createEventsWithReceivers(batch));
       // Publish Events
       if (!eventsWithReceivers.isEmpty()) {
         alertMetrics.withTotalEvents(alertMetrics.getTotalEvents() + eventsWithReceivers.size());
         publishEvents(eventsWithReceivers);
       }
     } catch (Exception e) {
-      LOG.error("Error in executing the Job : {} ", e.getMessage());
+      long currentLatest = Entity.getCollectionDAO().changeEventDAO().getLatestOffset();
```
harshach, line 272: do we need to make a call to db here?

harshach: isnt' offset is the current offset and all we need to increment the batch?

mohityadav766: Ex: - Change event has 50 entries table
current offset is let's say 20 , batch size 10
reading from 20 to  with batch size 10 if the fails , then i update the offset to 20 + 10 because 50 (total or latest offset) - current > batch size so we next batch of events
but let's say current is 45 and batch is 10 
reading from 45 there is a failure in this case 50-45 < batch size so the current offset needs to be updated to offset + remanining items(5)

mohityadav766: and at any failure we need the latest offset of change_event

harshach: we do have offset of the message no? 
if we record what is the offset its failing then we should be able to move currentOffset = failedOffset+1.
Trrying to see if we can avoid polling db

mohityadav766: it's in batch , we are reading in batches which are passed , so it is difficult to add per message at that layer

harshach: can we throw a custom exception to catch, exception should contain the offset it failed on?

mohityadav766: done , removed the db call as well

PR #19591, opened 2025-01-29
MINOR: feat(event-subscriptions): event consumer extension

sushi30:
## Introducing Dependency Injection for Quartz Jobs

Quartz jobs might require different dependencies to function. These dependencies are handled via the [CustomJobFactory](https://github.com/open-metadata/OpenMetadata/blob/fe300b78b8c0f893533f85af3762d91f799dd0ad/openmetadata-service/src/main/java/org/openmetadata/service/apps/scheduler/CustomJobFactory.java#L14) for OpenMetadata applications.

While this is convient as it allows to share state from the running app to the jobs which are stateless, it couples all applications implementing the [`AbstractNativeApplication`](https://github.com/open-metadata/OpenMetadata/blob/fe300b78b8c0f893533f85af3762d91f799dd0ad/openmetadata-service/src/main/java/org/openmetadata/service/apps/AbstractNativeApplication.java#L55) to the dependencies (in this case, `CollectionDao` and `SearchRepository`).

The solution in this PR uses a simple [DIContainer](https://github.com/open-metadata/OpenMetadata/blob/e354c4056d3264ec15a461614f7ff2b0cba191b5/openmetadata-service/src/main/java/org/openmetadata/service/util/DIContainer.java) class that can register arbitrary dependencies. The dependencies can be restricted based on some policy at a later time. Adding a new dependency does not require changing any of the implementing classes, and each application can use only the dependencies it requires.

## Changes

- implemented DI for event consumers
- added custom config for event consumers
- initialize the event scheduler along with application
- added workflow custom type

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplication.java
sushi30, line 1: EventSubscriptionScheduler will by used by apps so it is initialized here.

openmetadata-service/src/main/java/org/openmetadata/service/apps/scheduler/AppScheduler.java
```diff
@@ -92,7 +92,7 @@ private AppScheduler(
     // Add OMJob Listener
     this.scheduler
         .getListenerManager()
-        .addJobListener(new OmAppJobListener(dao), jobGroupEquals(APPS_JOB_GROUP));
+        .addJobListener(new OmAppJobListener(), jobGroupEquals(APPS_JOB_GROUP));
```
sushi30, line 95: removed dao dependency. use repository instead

openmetadata-service/src/main/java/org/openmetadata/service/apps/scheduler/OmAppJobListener.java
sushi30, line 1: removed `AbstractOmAppJobListener` as it was not adding any value. we can use the concrete class directly instead.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/AppRepository.java
sushi30, line 1: allow apps to have children EventSubscription that are deployed or removed based on the app configuration

openmetadata-service/src/main/java/org/openmetadata/service/resources/automations/WorkflowMapper.java
```diff
@@ -2,30 +2,19 @@
...
-  private final OpenMetadataApplicationConfig openMetadataApplicationConfig;
-
-  public WorkflowMapper(OpenMetadataApplicationConfig config) {
-    this.openMetadataApplicationConfig = config;
-  }
 
   @Override
   public Workflow createToEntity(CreateWorkflow create, String user) {
-    OpenMetadataConnection openMetadataServerConnection =
-        new OpenMetadataConnectionBuilder(openMetadataApplicationConfig).build();
```
sushi30, line 20: we don't need `openMetadataApplicationConfig` stored in the workflow. It gets added at runtime by the `pipleineServiceClient`.

openmetadata-service/src/main/java/org/openmetadata/service/resources/events/subscription/EventSubscriptionResource.java
```diff
@@ -102,7 +102,7 @@
         "The `Events` are changes to metadata and are sent when entities are created, modified, or updated. External systems can subscribe to events using event subscription API over Webhooks, Slack, or Microsoft Teams.")
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
-@Collection(name = "events/subscriptions")
+@Collection(name = "events/subscriptions", order = 7) // needs to initialize before applications
```
sushi30, line 105: since applications can deploy event subscriptions, we need it to be initialized earlier in the cycle.

openmetadata-service/src/main/java/org/openmetadata/service/secrets/converter/WorkflowClassConverter.java
```diff
@@ -30,7 +31,8 @@ public WorkflowClassConverter() {
   public Object convert(Object object) {
     Workflow workflow = (Workflow) JsonUtils.convertValue(object, this.clazz);
 
-    tryToConvertOrFail(workflow.getRequest(), List.of(TestServiceConnectionRequest.class))
+    tryToConvertOrFail(
+            workflow.getRequest(), List.of(TestServiceConnectionRequest.class, Map.class))
```
sushi30, line 35: allow workflow configurations to be arbitrary maps so they can be extended on collate without exposing the interface on OpenMetadata

openmetadata-spec/src/main/resources/json/schema/events/eventSubscription.json
sushi30, line 1: added `Custom` type and arbitrary `config` so we extend in Collate.

pmbrull (dismissed):

pmbrull (approved):

chirag-madlani (approved):

PR #19616, opened 2025-01-31
issue-19112: API service schema fields of object type not listed

### LINKED ISSUE #19112: Support for Ingesting Fields with $ref in Schema Properties in API Ingestion
-- Currently, while processing schema objects, we only support fields that have a type defined directly within the properties dictionary. For example:

```
{
  "type": "object",
  "required": ["name"],
  "properties": {
    "name": {
      "type": "string"
    },
    "address": {
      "$ref": "#/components/schemas/Address"
    },
    "age": {
      "type": "integer",
      "format": "int32",
      "minimum": 0
    }
  }
}
```
In the above example, only name and age are ingested, while address is skipped because it uses a $ref to reference another schema and lacks a direct type.

From the referenced schema, fields like address could potentially map to other schemas but are currently not identified as valid fields for ingestion.

-- Introduce support for fields with $ref in future versions. This would involve resolving references to their respective schemas and incorporating the resolved schema into the ingestion process.

Use Case:

- Ensure all valid fields, including those referencing other schemas via $ref, are ingested.
- Improve schema processing coverage and allow for better representation of nested or complex data structures.

**Additional context**

harshsoni2024:
### Describe your changes:

Fixes #19112 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/api/rest/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/api/rest/metadata.py
@@ -262,16 +262,15 @@ class RestSource(ApiServiceSource):
             fetched_fields = []
             for key, val in schema_fields.get("properties", {}).items():
                 dtype = val.get("type")
-                if not dtype:
-                    continue
-                fetched_fields.append(
-                    FieldModel(
-                        name=key,
-                        dataType=DataTypeTopic[dtype.upper()]
+                if dtype:
+                    parsed_dtype = (
+                        DataTypeTopic[dtype.upper()]
                         if dtype.upper() in DataTypeTopic.__members__
-                        else DataTypeTopic.UNKNOWN,
+                        else DataTypeTopic.UNKNOWN
                     )
-                )
+                else:
+                    parsed_dtype = DataTypeTopic.UNKNOWN
+                fetched_fields.append(FieldModel(name=key, dataType=parsed_dtype))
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/ingestion/source/api/rest/metadata.py
```diff
@@ -195,23 +195,22 @@ def _generate_collection_url(self, collection_name: str) -> Optional[AnyUrl]:
...
-            return AnyUrl(base_url)
+        return AnyUrl(base_url)
 
     def _get_api_request_method(self, method_type: str) -> Optional[str]:
         """fetch endpoint request method"""
         try:
             return ApiRequestMethod[method_type.upper()]
         except KeyError as err:
-            logger.info(f"Keyerror while fetching request method: {err}")
+            logger.error(f"Keyerror while fetching request method: {err}")
```
SumanMaharana, line 213: why we changed from info to error for logging ? why not using warning ?

harshsoni2024: We just changed to be visible when it runs on loglevel info.

SumanMaharana (approved):

PR #19618, opened 2025-01-31
feat: use table_id instead of table_name for fetching Superset dataset columns

mgorsk1:
### Describe your changes:

Fixes <issue-number>

I worked on changing query for fetching dataset columns from relying on dataset **name** to dataset **id**. Implementation relying on dataset names is not reliable due to below:
- there might be multiple datasets with same name but different id. dataset id is the only proper attributes on which joining between tables and table_columns tables should happen.
- it relied on the assumption that dataset name is lowercase, which might not be (and usually is not) true

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ulixius9 (dismissed):

mgorsk1:
fixed tests and linting @ulixius9

ulixius9 (approved):

PR #19704, opened 2025-02-07
issue-1047: Avoid default schema, database metadata

harshsoni2024:
### Describe your changes:
List of databases:
- Postgres
- Snowflake
- MSSQL
- MYSQL

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/snowflake/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/snowflake/metadata.py
@@ -60,6 +60,7 @@ from metadata.ingestion.ometa.ometa_api import OpenMetadata
 from metadata.ingestion.source.database.column_type_parser import create_sqlalchemy_type
 from metadata.ingestion.source.database.common_db_source import (
     CommonDbSourceService,
+    DefaultExcludeFilters,
     TableNameAndType,
 )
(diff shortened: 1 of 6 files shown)


ingestion/src/metadata/ingestion/source/database/snowflake/metadata.py
```diff
@@ -160,6 +162,10 @@ def __init__(
         pipeline_name,
         incremental_configuration: IncrementalConfig,
     ):
+        self.default_entities = {
+            "database": {"excludes": DEFAULT_EXCLUDE_DATABASES},
+            "schema": {"excludes": DEFAULT_EXCLUDE_SCHEMAS},
+        }
```
ulixius9, line 168: when we apply a default filter it will always be a exclude filter so i would prefer making it this way 


```suggestion
        self.default_exclude_filters = {
            "database": DEFAULT_EXCLUDE_DATABASES,
            "schema": DEFAULT_EXCLUDE_SCHEMAS,
        }
```

also instead of maintaining a dictionary just create a pydantic object for this purpose 

which we can initialise this common_db init and override in child classes if required

PR #19711, opened 2025-02-07
MINOR: fix: close client after query

sushi30:
use context clients in SQL sampler to close the connection once the query is complete

## Example in e2e tests

image

Running e2e here:
https://github.com/open-metadata/OpenMetadata/actions/workflows/py-cli-e2e-tests.yml?query=branch%3Asampler-close-connection

### Before

Screenshot 2025-02-10 at 14 30 28

### After

Screenshot 2025-02-10 at 14 30 25

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


IceS2 (dismissed):

pmbrull (approved):

TeddyCr:
@sushi30 you should not have to create a context manager yourself and use it in the methods that return `CTE` type (as at this stage nothing is sent to the db. You should be able to do something like the below on line 178 in `sqlaLchemy.sampler.py`.  You can check what we do here in the sqa profiler_interface https://github.com/open-metadata/OpenMetadata/blob/5bb8c5cfe3df212af443b30f0f247e5368217387/ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py#L399-L400

```
        Session = self.session_factory  # pylint: disable=invalid-name
        with Session() as session:
            ...
```

More details on sessions management here https://docs.sqlalchemy.org/en/20/orm/session_basics.html#basics-of-using-a-session

TeddyCr (approved):

PR #19751, opened 2025-02-11
MINOR - Hybrid Ingestion Agent - Link ingestion agents to ingestion pipelines

gyohub:
### Describe your changes:

Adds the `IngestionAgent` field to the `IngestionPipeline` schema. The field is a simple entity reference.
There's no db migrations since the field can be null.  
This is required for the Collate's Hybrid ingestion agent feature.

Fixes GEN-2117

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
@@ -39,6 +39,7 @@ import org.openmetadata.schema.type.ChangeEvent;
 import org.openmetadata.schema.type.EntityReference;
 import org.openmetadata.schema.type.FieldChange;
 import org.openmetadata.schema.type.Include;
+import org.openmetadata.schema.type.Relationship;
 import org.openmetadata.sdk.PipelineServiceClientInterface;
 import org.openmetadata.service.Entity;
 import org.openmetadata.service.OpenMetadataApplicationConfig;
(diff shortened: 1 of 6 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
```diff
@@ -151,6 +151,7 @@ public void storeEntity(IngestionPipeline ingestionPipeline, boolean update) {
   @Override
   public void storeRelationships(IngestionPipeline ingestionPipeline) {
     addServiceRelationship(ingestionPipeline, ingestionPipeline.getService());
+    addServiceRelationship(ingestionPipeline, ingestionPipeline.getIngestionAgent());
```
pmbrull, line 154: is this the right relationship? it adds the `CONTAINS` relationship which might not be the best representation here. Maybe we can use HAS?

pmbrull (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
gyohub: Good call. I just changed

openmetadata-spec/src/main/resources/json/schema/entity/services/ingestionPipelines/ingestionPipeline.json
```diff
@@ -221,6 +221,10 @@
     "applicationType": {
       "description": "Type of the application when pipelineType is 'application'.",
       "type": "string"
+    },
+    "ingestionAgent" : {
```
harshach, line 225: This requires migration right cc @pmbrull

gyohub: @harshach db migration is not required since this field is nullable since any ingestion pipeline that is not hybrid won't have an agent reference.

pmbrull (approved):

PR #19752, opened 2025-02-11
Fix #19680: mask query method to return none if failed

### LINKED ISSUE #19680: mask query method should not return original query in case of exception
**Affected module**
Ingestion

**Describe the bug**
I am not sure whether it's a bug or wrongly defined logic, but I wanted to raise concern over [**mask_query** method](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/lineage/masker.py#L112). In it's current form, **mask_query** is a method that:
- will return **masked** query if masking succeeds
- will return **original** query if masking does not succeed 

Which I think is wrong (the second part). If a user relies on this method and masking was not possible, then it would return original query (possibly with sensitive data) without users knowledge. In automated scenarios this might cause leakage of sensitive information when masking fails but returned query is pushed for further processing (for example when enriching lineage edge). The method ideally either returns `None` or raises an exception if masking fails.

**To Reproduce**

**Expected behavior**
**mask_query** does return `None` when masking fails.

**Version:**
 - Python version:

**Additional context**

ulixius9:
### Describe your changes:

Fix #19680: mask query method to return none if failed

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/lineage/masker.py
+++ b/ingestion/src/metadata/ingestion/lineage/masker.py
@@ -127,4 +127,4 @@ def mask_query(
     except Exception as exc:
         logger.debug(f"Failed to mask query with sqlfluff: {exc}")
         logger.debug(traceback.format_exc())
-    return query
+    return None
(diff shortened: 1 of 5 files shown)


SumanMaharana (approved):
LGTM

PR #19754, opened 2025-02-11
Fixes 19755: Publish app config with status

sushi30:
### Describe your changes:

Publish appConfig as part of the status for ingestion pipelines.

Delivers https://github.com/open-metadata/OpenMetadata/issues/19755

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/models/custom_pydantic.py
```diff
@@ -110,6 +111,11 @@ def model_dump_json(  # pylint: disable=too-many-arguments
             ensure_ascii=True,
         )
 
+    def model_dump_masked(self, *args, **kwargs):
```
pmbrull, line 114: will this mask all the secrets when serialized? what about when we try to create a service from a YAML?

sushi30: It will only marked fields marked as `secret` like passwords, tokens etc.. Not the whole service. We can add more complex for things like service connections if we want to handle them differently.

sushi30: For example [this password field](https://github.com/open-metadata/OpenMetadata/blob/f80eaf3a2653135bd7028d214b447d81f66a0b15/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/common/basicAuth.json#L13).

pmbrull: for now i would at least keep the option to see if storeCreds: true/false in the workflow config and control the secrets. If we were to create a service from a YAML with this config, the server would just receive ****

sushi30: In this case we are dumping the configuration and sending it to the DB but we don't necessarily want to handle secrets. There should not be any secrets in `appConfig` but having this kind of functionality makes it future-proof so that if something sneaks in there it will be masked.

I don't think this functionality should be used for managing the workflow configuration.

Are you suggesting to dump everything in plaintext for this use-case?

pmbrull: what I'm thinking is that our `model_dump_json()` should accept an extra parameter to mask or not secrets on-demand. For the appConfig you mention, we can add the logic to `mask=True` while keeping a default mask=False for any other request so far, or at least for our POST in the ometa_api

sushi30: [addressed here](
https://github.com/open-metadata/OpenMetadata/pull/19754/files#diff-3a95c7a1e98ed0dbc1733ebe04fbc466d557a22bb34c0dbd2e85a86c83ebc9a4R116-R127)

pmbrull (approved):

pmbrull (approved):

PR #19846, opened 2025-02-17
Asset uploader

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/config/S3Configuration.java
```diff
@@ -0,0 +1,49 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import javax.validation.constraints.AssertTrue;
+import javax.validation.constraints.NotBlank;
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.commons.lang3.StringUtils;
+
+@Getter
+@Setter
+public class S3Configuration {
```
pmbrull, line 12: Could we reuse this somehow in here? https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/security/credentials/awsCredentials.json#L7

We might just need to add the bucket and keep centralized creds management.

Same would be true for Azure/GCP.

pmbrull: we might be able to configure the supported storages in JSON Schema and reuse those defs in Collate for the factories etc. as well

conf/openmetadata.yaml
```diff
@@ -156,6 +156,10 @@ database:
   evictionInterval: ${DB_CONNECTION_EVICTION_INTERVAL:-5 minutes}
   minIdleTime: ${DB_CONNECTION_MIN_IDLE_TIME:-1 minute}
 
+objectStorage:
+  enabled: false
+  provider: NOOP
```
akash-jain-10, line 161:
```suggestion
  provider: NOOP # Supported values - "s3", "azure", or "noop" (for local testing)
```

harshach: its meant to be no configurable in case of OSS

PR #19912, opened 2025-02-21
Fixes #15696: Added Powerbi filter query

### LINKED ISSUE #15696: filter Power BI workspaces before metadata ingestion
**Feature**
Allow push down of filtering in Power BI ingestion.

**Describe the task**

Currently, in the Power BI ingestion pipeline, any workspace (aka project in OM) filters are applied AFTER the metadata was ingested. This causes the issue, that if there are many workspaces in a Power BI tenant, then the API used (https://learn.microsoft.com/en-us/rest/api/power-bi/admin/groups-get-groups-as-admin) to ingest the metadata will hit the API limits very quickly, resulting in timeout and failures.

image
image

The feature request is to allow pushing down the filtering to the Power BI level, but using a different API with an odata filter or something similar - https://learn.microsoft.com/en-us/rest/api/power-bi/groups/get-groups#get-a-list-of-workspaces-using-a-filter-example

image

akashverma0786:
### Describe your changes:

Fixes #15696 

Short blurb explaining:
- What changes did you make?
    Added method to add filter query in powerbi api call to filter out projects on call itself
   A method that changes ProjectFilterPattern(if provided) into odata condition
- Why did you make them?
    To allow pushing down the filtering to the Power BI level, using a different API with an odata filter.
    Test for method

Project Filter pattern:- 
Screenshot 2025-02-24 at 18 05 26

Before odata filter:- 
Screenshot 2025-02-24 at 17 34 39

After filter:- 
Screenshot 2025-02-24 at 17 37 37

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/powerbi/client.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/powerbi/client.py
@@ -294,7 +294,7 @@ class PowerBiApiClient:
                 f"Error converting regex '{regex}' to OData condition: {exc}"
             )
             return ""
-        
+
     def create_filter_query(self, filter_pattern) -> Optional[str]:
         """
         Create a complete filter query for workspaces from filter_pattern
--- a/ingestion/src/metadata/ingestion/source/dashboard/powerbi/client.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/powerbi/client.py
@@ -266,7 +266,7 @@ class PowerBiApiClient:
             # Exact match
             if regex.startswith("^") and regex.endswith("$"):
                 literal = regex[1:-1]
-                return f"name eq '{literal}'"
+                return f"trim(name) eq '{literal}'"
 
             # Starts with
             if regex.startswith("^"):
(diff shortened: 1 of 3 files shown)


ingestion/src/metadata/ingestion/source/dashboard/powerbi/client.py
```diff
@@ -164,7 +165,7 @@ def fetch_dashboards(self) -> Optional[List[PowerBIDashboard]]:
             response_data = self.client.get("/myorg/admin/dashboards")
             response = DashboardsResponse(**response_data)
             return response.value
-        group = self.fetch_all_workspaces()[0]
+        group = self.fetch_all_workspaces(None)[0]
```
harshsoni2024, line 168: better to make `filter_pattern` optional in function def

ingestion/src/metadata/ingestion/source/dashboard/powerbi/metadata.py
```diff
@@ -171,7 +171,8 @@ def get_admin_workspace_data(self) -> Optional[List[Group]]:
         fetch all the workspace ids
         """
         groups = []
-        workspaces = self.client.api_client.fetch_all_workspaces()
+        filter_pattern = self.source_config.projectFilterPattern
```
harshsoni2024, line 174: same logic should be applied in non-admin [method](https://github.com/open-metadata/OpenMetadata/blob/8deccf6698e25875827b6751b177f31cbe69a681/ingestion/src/metadata/ingestion/source/dashboard/powerbi/metadata.py#L135)

PR #19946, opened 2025-02-24
GEN-2891 - Add ingestion agent relationship to service entities

gyohub:
### Describe your changes:

Part of the work on the [Collate Hybrid Ingestion Agent](https://www.notion.so/collate-inc/Add-ingestion-agent-entity-reference-to-Service-entities-1a1b45723024803a8d6cdef48d7b160a?pvs=4)

Fixes GEN-2891

The ingestion agent can be assigned at service level too.
a DB migration is not required as there's no need for a new table column.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/APIServiceRepository.java
```diff
@@ -17,4 +18,17 @@ public APIServiceRepository() {
         ServiceType.API);
     supportsSearch = true;
   }
+
+  @Override
+  public void storeRelationships(ApiService service) {
```
pmbrull, line 23: not sure eh? but could this live under `ServiceEntityRepository` instead of in all the service repos?

pmbrull: oki understood, thanks

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ServiceEntityRepository.java
```diff
@@ -81,7 +82,18 @@ public void storeEntity(T service, boolean update) {
...
+  }
+
+  private void addIngestionAgentRelationship(T service) {
+    if (service.getIngestionAgent() != null) {
+      addRelationship(
+          service.getId(),
+          service.getIngestionAgent().getId(),
+          entityType,
+          service.getIngestionAgent().getType(),
+          Relationship.HAS);
```
pmbrull, line 95: dont kill me but maybe instead of `HAS` we can have a `USES` relationship? Other than that LGTM

gyohub: haha no worries at all, I will change it. @pmbrull should we do the same for the IngestionPipeline relationship?

chirag-madlani (approved):

PR #20040, opened 2025-03-01
feat(ui): support /logout path to perform logout from api redirect

chirag-madlani:
### Describe your changes:

Fixes: It fix issue with refresh failure, From BE it will send redirect to serverURL which didn't clear userInfo and ends up to infinite loader on UI side 

To avoid that, created `/logout` path redirecting there from BE so that UI will clear the LoggedIn info and throw the user to the sign in page for a smoother experience 

https://github.com/user-attachments/assets/202e4105-7096-4437-bd82-2b293cb06da7

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/security/AuthenticationCodeFlowHandler.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/security/AuthenticationCodeFlowHandler.java
@@ -349,7 +349,7 @@ public class AuthenticationCodeFlowHandler {
       if (session != null) {
         LOG.debug("Invalidating the session for logout");
         session.invalidate();
-        httpServletResponse.sendRedirect(serverUrl);
+        httpServletResponse.sendRedirect(serverUrl + "/logout");
       } else {
         LOG.error("No session store available for this web context");
       }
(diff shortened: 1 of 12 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/security/AuthenticationCodeFlowHandler.java
```diff
@@ -349,7 +349,7 @@ public void handleLogout(
       if (session != null) {
         LOG.debug("Invalidating the session for logout");
         session.invalidate();
-        httpServletResponse.sendRedirect(serverUrl);
+        httpServletResponse.sendRedirect(serverUrl + "/logout");
```
mohityadav766, line 352: Better to take it as part of configuration, otherwise if this changes that would need code change?

chirag-madlani: It's handled on Frontend so not sure how frequent it may change
Here:
https://github.com/open-metadata/OpenMetadata/blob/55b5001bb5a0729dd3fa7daa26e22ee07120d0d5/openmetadata-ui/src/main/resources/ui/src/constants/constants.ts#L134

mohityadav766 (approved):

aniketkatkar97 (approved):

PR #20091, opened 2025-03-05
MINOR: fix wrong suggestion count being fetched on table change

Ashish8689:
### Describe your changes:

- I worked fixing wrong suggestion count being fetched on Table entity if switch to another table from one table. Due to Provider being wrapped on the TableDetailPage!

Issue : 

https://github.com/user-attachments/assets/217e723e-2843-46aa-9ae3-023a01a0c811

Resolved : 

https://github.com/user-attachments/assets/04013c15-c947-4e34-b9ed-1ec42e9e786c

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Suggestions/SuggestionsProvider/SuggestionsProvider.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Suggestions/SuggestionsProvider/SuggestionsProvider.tsx
@@ -24,6 +24,7 @@ import React, {
   useState,
 } from 'react';
 import { useTranslation } from 'react-i18next';
+import { PAGE_SIZE } from '../../../constants/constants';
 import { usePermissionProvider } from '../../../context/PermissionProvider/PermissionProvider';
 import {
   Suggestion,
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/src/components/Suggestions/SuggestionsProvider/SuggestionsProvider.tsx
```diff
@@ -62,48 +63,51 @@ const SuggestionsProvider = ({ children }: { children?: ReactNode }) => {
   const publish = usePub();
 
   const [loading, setLoading] = useState(false);
-  const [suggestionLimit, setSuggestionLimit] = useState<number>(10);
+  const [suggestionLimit, setSuggestionLimit] = useState<number>(PAGE_SIZE);
```
chirag-madlani, line 66: Why we are using state and not using usePaging hook?

Ashish8689: Here we don't require any Next, Previous, or limit changing mechanism. State is only set to get all the value from total and fetch them single time. On the user action.

PR #20135, opened 2025-03-07
MINOR: fix suggestion more button throwing error

Ashish8689:
### Describe your changes:

- I worked on fixing suggestion more button throwing error, due to button onClick reference were passed to the function where the function expect some arguments.
- This will fix the playwright's failure as well on main.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Suggestions/SuggestionsSlider/SuggestionsSlider.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Suggestions/SuggestionsSlider/SuggestionsSlider.tsx
@@ -45,7 +45,7 @@ const SuggestionsSlider = () => {
           data-testid="more-suggestion-button"
           loading={loading}
           type="primary"
-          onClick={fetchSuggestions}>
+          onClick={() => fetchSuggestions()}>
           {t('label.plus-count-more', {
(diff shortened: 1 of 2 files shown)


ShaileshParmar11 (approved):

openmetadata-ui/src/main/resources/ui/src/components/Suggestions/SuggestionsSlider/SuggestionsSlider.tsx
```diff
@@ -45,7 +45,7 @@ const SuggestionsSlider = () => {
           data-testid="more-suggestion-button"
           loading={loading}
           type="primary"
-          onClick={fetchSuggestions}>
+          onClick={() => fetchSuggestions()}>
```
chirag-madlani, line 48: What we fixes here? Like there were params that causing issue?

Ashish8689: Yes, so we have one argument in the  `fetchSuggestions` which is optional. And since here it is directly aligned with the onClick so the onClick reference is going as the param which was causing the issue.

PR #20141, opened 2025-03-07
issue-19892: parse powerbi table source

### LINKED ISSUE #19892: PowerBI connector - PowerBI and Snowflake lineage
- The current logic appears to match Power BI models to Snowflake tables based purely on the table name, without considering:

  - The database and schema context 
  - Whether the Power BI model actually connects to Snowflake at all
In our current workspace, we have seen cases where a Power BI model with no Snowflake connection is incorrectly matched to a Snowflake table of the same name. This is misleading and can result in incorrect lineage tracking.

The issue stems from the fact that the lineage extraction logic only looks at one API response, without cross-referencing the full connection details. To resolve this, it would require querying an additional API endpoint to retrieve metadata from the .bim file of the cube, which contains the full source mapping.

- We’ve observed an issue with how OpenMetadata establishes lineage between Power BI models and Snowflake tables, and we believe an enhancement is necessary to ensure accurate dependency tracking.

**Additional context**

harshsoni2024:
### Describe your changes:

Fixes #19892

PowerBI parse table source for correct schema, table name for snowflake & redshift

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/dashboard/powerbi/metadata.py
```diff
@@ -616,6 +617,81 @@ def _get_data_model_column_fqn(
...
+                schema and table
+            ):  # atlease 2 entities schema and table should be fetched
+                return {"database": database, "schema": schema, "table": table}
+            return None
+        except Exception as exc:
+            logger.debug(f"Error to parse redshift table source: {exc}")
+            logger.debug(traceback.format_exc())
+        return None
+
+    def _parse_snowflake_source(self, source_expression: str) -> Optional[dict]:
```
pmbrull, line 643: we need unit tests on these things. It's easy to add, and will serve as an example of how we expect the data in PBI to look like.

Also this snowflake source implementation does not seem to reference snowflake anywhere. Would this match any other source?

pmbrull (changes requested):

pmbrull (approved):

Issue #18269, opened 2024-10-15
Users without access are able perform bulk Add/Remove Assets using API


Prajwal214:
**Affected module**
Does it impact the UI, backend or Ingestion Framework? 
-- Backend

**Describe the bug**
-- Users who do not have the necessary access permissions are able to perform bulk add or remove operations on assets through the API.

**To Reproduce**

Screenshots or steps to reproduce

```
curl -X PUT "http://localhost:8585/api/v1/teams/Test_Team1/assets/add" \
 -H "accept: application/json"\
 -H "authorization: Bearer Auth Token"\
 -H "content-type: application/json" \
 -d '{"assets":[{"deleted":false,"description":"","displayName":"Prajwal","fullyQualifiedName":"","href":"http://localhost:8585/api/v1/users/612b464b-99c5-4abe-b2a6-ed3987bba66b","id":"612b464b-99c5-4abe-b2a6-ed3987bba66b","inherited":false,"name":"prajwal1","type":"user"}]}' \
```

**Expected behavior**
-- only the users with Access should be able to perform the action.

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8] OM v1.5.6
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

mohityadav766:
closing this as this was fixed , more test are added here https://github.com/open-metadata/OpenMetadata/pull/25170

PR #20150, opened 2025-03-08
fix-18269 Users without access are able perform bulk Add/Remove Asset…

sonika-shah:
…s using API

### Describe your changes:

Fixes [#18269](https://github.com/open-metadata/OpenMetadata/issues/18269)

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/TeamResource.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/TeamResource.java
@@ -14,6 +14,7 @@
 package org.openmetadata.service.resources.teams;
 
 import static org.openmetadata.common.utils.CommonUtil.listOf;
+import static org.openmetadata.service.security.DefaultAuthorizer.getSubjectContext;
 
 import io.dropwizard.jersey.PATCH;
 import io.swagger.v3.oas.annotations.ExternalDocumentation;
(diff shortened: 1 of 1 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/TeamResource.java
```diff
@@ -749,4 +756,42 @@ public Response importCsvAsync(
       String csv) {
     return importCsvInternalAsync(securityContext, name, csv, dryRun);
   }
+
+  public final void verifyUserPermission(SecurityContext securityContext) {
```
mohityadav766, line 760: hi @sonika-shah can we add it as a different operation itself ADD_ASSETS?
this seems like to much to handle single permission of Adding assets!

mohityadav766: don't want to build if else, we have flexible roles and policies which can be used to associate operation to different resource

mohityadav766 (changes requested):
needs some work

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/TeamResource.java
sonika-shah: hi @mohityadav766 , will check with current implementations, when we add multiple operations to the operation_context, it treats them as an AND condition—all permissions must be present to authorize.  

that’s why I implemented it so that if any of the listed operations is present, the user is authorized.

sonika-shah:
this  is already fixed with  pr : Missing Permission on Adding Users to team (#20768)

PR #20169, opened 2025-03-10
fix: return None for BQ partition handler

TeddyCr:
### Describe your changes:
- Fixes issue with `Other` partition in BQ partition handler. Return `None` vs raising an error if partition type is not supported

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Bug fix


sushi30 (approved):

ingestion/src/metadata/sampler/partition.py
```diff
@@ -164,8 +167,12 @@ def _handle_bigquery_partition(
                 partitionIntegerRangeEnd=10000,
             )
         # TODO: Allow External Hive Partitioning for profiler
-        raise TypeError(
-            f"Unsupported partition type {partition.intervalType}. Skipping table"
+        logger.debug(
```
sushi30, line 170: might be better to set this as `info` so that it is visible be default.

PR #20299, opened 2025-03-17
ISSUE #19185 -- Allow user to choose non random sample

### LINKED ISSUE #19185: Allow User to Choose Whether data should be sampled
**Affected module**
Profiler Ingestion

**Describe the bug**
The table data sampling implementation in Profiler Ingestion job produces a SQL statement like the one below:

```sql
/* {"app": "OpenMetadata", "version": "1.5.10.0"} */
SELECT
    ......
    MOD(ABS(RANDOM()) * 100, 41026) AS random
  FROM my_table
  ORDER BY random
  LIMIT 100
```

This is an expensive query because the sort operation in SQL (`ORDER BY random`) is performed before the row limit (`LIMIT 100`).
Unlike other databases, AWS Redshift doesn't have a native and effective table sampling method, so I believe many OM users would prefer to simply `LIMIT N` and sacrifice true random sampling for better performance. 

**To Reproduce**
- OM v1.5.10
- A Redshift (or a Postgres?) db with a large table. (100+ columns and  billion rows.)
- Setup a Profiler Ingestion job, disable every option **except** `Generate Sample Data`.
- Set `Profile Sample Type=ROWS` and `Profile Sample=100`
- Trigger job one time.

**Expected behavior**
Table data sampling should adopt a lightweight approach for any database.
Even if true and deterministic random sampling is crucial feature, an optional opt-out would be great for cost effectiveness.

**Version:**
 - OS: n/a (official Docker image)
 - Python version: n/a (official Docker image)
 - OpenMetadata version: v1.5.10
 - OpenMetadata Ingestion package version: openmetadata/ingestion:1.5.10

**Additional context**

TeddyCr:
### Describe your changes:
Fixes #19185 

- Allow users to choose non random sample on `ROWS` sample type

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Improvement


sushi30:
I suggest adding this as another [sampling method](https://github.com/open-metadata/OpenMetadata/blob/d709018c93384652627f143e2c4f5db570ee8dea/openmetadata-spec/src/main/resources/json/schema/entity/data/table.json#L22) instead of another field. Otherwise we have ambiguous cases like:

```yaml
randomizedSample: true
samplingMethod: system
```

sushi30 (approved):

TeddyCr:
> I suggest adding this as another [sampling method](https://github.com/open-metadata/OpenMetadata/blob/d709018c93384652627f143e2c4f5db570ee8dea/openmetadata-spec/src/main/resources/json/schema/entity/data/table.json#L22) instead of another field. Otherwise we have ambiguous cases like:
> 
> ```yaml
> randomizedSample: true
> samplingMethod: system
> ```

Not sure that would make sense. I'll move it to the advanced settings and add the appropriate doc in the ingestion config section

sushi30 (approved):

PR #20304, opened 2025-03-17
feat: add created by for test case entities

TeddyCr:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-spec/src/main/resources/json/schema/tests/testCase.json
```diff
@@ -89,6 +89,10 @@
       "$ref": "../type/entityReferenceList.json",
       "default": null
     },
+    "createdBy": {
```
harshach, line 92: does it require any migration for existing tests?

harshach: the use-case I am thinking is , if there are existing customers with lot of teset cass and then later adopted the AI recommenation , we should be able to showcase how many are created by users vs AI. I think if we don’t have it populated for users this might not look right. wdyt?

TeddyCr: Makes sense. Let me add this migration then. I'll use the updatedby field as the ref.

ulixius9 (approved):

mohityadav766 (approved):

aniketkatkar97 (approved):

PR #20318, opened 2025-03-18
saml documentation updated

### LINKED ISSUE #19162: SAML documentation corrections
**Is some content missing, wrong or not clear?**

The documentation for the following fails to mention that the AUTHENTICATION_PROVIDER has to be set to saml (lowercase) before SAML will work.

https://docs.open-metadata.org/latest/deployment/security/saml/azure

On step 4, it states "Set up for SAML is done, you should be routed to your IDP on trying to Sign-in."  This doesn't happen until AUTHENTICATION_PROVIDER is set to saml.

https://docs.open-metadata.org/latest/deployment/security/saml/docker

The title of this article is "Keycloak SSO for Docker" but this isn't necessarily about Keycloak, which is in a different section of the documentation.

`SAML_IDP_CERTIFICATE=/path/to/the/certificate`

Using a path for this doesn't appear to work had to pass the entire certificate here as a string instead of the path to the file.  When using a path it tried to parse the path as a base64 encoded cert instead of the contents.

Update the documentation to mention need for AUTHENTICATION_PROVIDER and that the SAML_IDP_CERTIFICATE either needs to actually support a path in the source code or be updated to pass the certificate in as a string instead of a path.

safayavatsal:
### Describe your changes:

Fixes #19162

SAML documentation corrections for azure

I worked on SAML documentation corrections

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [x] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-docs/content/v1.5.x/deployment/security/saml/aws.md
+++ b/openmetadata-docs/content/v1.5.x/deployment/security/saml/aws.md
@@ -83,7 +83,7 @@ Security requirements for your **production** environment:
       keyStorePassword: ${SAML_KEYSTORE_PASSWORD:-""}
 ```
 
-- Populate the above config from xml metadata
+- Populate the above config from [xml metadata](/deployment/security/saml/xml_file)
 
 {% image src="/images/v1.5/deployment/security/saml/aws/saml-aws-8.png" alt="populate-metadata" /%}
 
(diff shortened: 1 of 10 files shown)


pmbrull (approved):
awesome ty!

PR #20351, opened 2025-03-19
Bug fix for user profile page for Mentions Tab rendering and Diff View issue

shrushti2000:
Fixes issue related to Mentions tab not getting rendered properly for the selected Thread
and flickering issue happening for Diff component.

Issue
Screenshot 2025-03-19 at 6 24 14 PM

Fix
Screenshot 2025-03-19 at 6 24 33 PM

Issue
https://github.com/user-attachments/assets/a7667e3b-4502-46a3-9ab4-78a8e00758b7
Fix

https://github.com/user-attachments/assets/7a58b3b6-6fca-4605-a58a-40c9c2d131dc

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/pages/TasksPage/shared/DiffViewNew.tsx
```diff
@@ -66,50 +66,73 @@ export const DiffViewNew = ({
...
-    );
-  });
+    return `diff-${index}-normal`;
+  };
+
+  const elements = useMemo(
+    () =>
+      diffArr.map((diff, index) => {
+        const key = getDiffKey(diff, index);
+
```
Ashish8689, line 84: @shrushti2000 apart from the index don't we have any other thing that we use inplace of index ?

shrushti2000: Previously when we  were using uniqueId() for the key it  was causing  some re rendering issue. The Diff View was getting rendered on every interaction

Ashish8689: Understood but don't we have any unique value from `diff` that we can use and avoid index here?

Ashish8689: we can use value right here, since it will be different for every one ?

shrushti2000: updated the key with diff-${diff.value}-${diff.removed}-${diff.added}-added` combination

Ashish8689 (approved):

Issue #19629, opened 2025-01-31
API to fetch last login of the User


Prajwal214:
**Is your feature request related to a problem? Please describe.**
-- We attempted to retrieve the Last Login details of users via the API, but this information is currently unavailable. It is essential for us to track user activity, and we would like the ability to extract this data through the API.

**Describe the solution you'd like**
-- The API should provide the Last Login timestamp for each user.
- The data should be accessible via a dedicated endpoint or as part of the user metadata.

**Describe alternatives you've considered**

**Additional context**

PR #20390, opened 2025-03-24
Add last login time in User when logging in 

mohityadav766:
### Describe your changes:

Add work for  API to fetch last login of the User
#19629

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


mohityadav766:
@chirag-madlani this takes care of SAML, Basic Auth, ldap and Confidential client 
for public client type SSO this needs to be taken care of on UI , you can patch login time on /callback to user with PATCH api
"lastLoginTime"

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
```diff
@@ -692,6 +695,7 @@ public Response createOrUpdateUser(
           securityContext,
           getResourceContextByName(user.getFullyQualifiedName()),
           new OperationContext(entityType, MetadataOperation.CREATE));
+      user.withLastLoginTime(System.currentTimeMillis());
```
sushi30, line 698: Is this recording the user's last login time? Not sure it makes sense since in a scenario where someone else (like an admin) creates the user but the user themself do not log in and create a session.

openmetadata-service/src/main/java/org/openmetadata/service/security/auth/BasicAuthenticator.java
```diff
@@ -469,6 +469,7 @@ public JwtResponse loginUser(LoginRequest loginRequest) throws IOException, Temp
     checkIfLoginBlocked(email);
     User storedUser = lookUserInProvider(email, loginRequest.getPassword());
     validatePassword(email, loginRequest.getPassword(), storedUser);
+    Entity.updateUserLastLoginTime(storedUser, System.currentTimeMillis());
```
harshach, line 472: any reason to create this method in Entity ? instead of adding it to UserRepostiory, we can always refer Entity.getUserRepository()

mohityadav766: updated @harshach

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/UserResource.java
mohityadav766: @sushi30 updated the PR, it should be there in @Post and if selfSignUp is enabled , thanks for the review

sushi30 (approved):

sushi30 (approved):

PR #20399, opened 2025-03-24
Fix : Delete user playwright test 

shrushti2000:
Fixed an issue where the flaky test "Admin soft & hard delete and restore user from profile page"  was failing to locate the alert-icon due to a page redirection causing the alert to close before it was found.

 Fix:-
Updated the  logic to wait for all toast elements (alert-bar, alert-icon, and alert-icon-close) after the delete action. So that the alert elements are detected properly

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/utils/user.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/utils/user.ts
@@ -202,7 +202,7 @@ export const hardDeleteUserProfilePage = async (
   await page.click('[data-testid="confirm-button"]');
 
   await deleteResponse;
-
+  await page.waitForLoadState('networkidle');
   await toastNotification(page, /deleted successfully!/);
 };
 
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/playwright/utils/user.ts
```diff
@@ -196,13 +196,18 @@ export const hardDeleteUserProfilePage = async (
   await page.check('[data-testid="hard-delete"]');
   await page.fill('[data-testid="confirmation-text-input"]', 'DELETE');
 
+  const toastPromises = [
+    page.waitForSelector('[data-testid="alert-bar"]'),
+    page.waitForSelector('[data-testid="alert-icon"]'),
+    page.waitForSelector('[data-testid="alert-icon-close"]'),
+  ];
```
ShaileshParmar11, line 203: not sure why we want to wait for all of this?

shrushti2000: Its not able to find the alert icons as it is getting closed and we are also redirecting the user to landing page.

ShaileshParmar11 (approved):

PR #20432, opened 2025-03-26
Fixes: Ingest System Tags

keshavmohta09:
### Describe your changes:

Fixes <issue-number>

I added support to use OMD system tags during ingestion instead of creating new ones. For databases that provide tags in lowercase while OpenMetadata uses title/upper case, we now map them to system tags directly.

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/ometa/mixins/es_mixin.py
```diff
@@ -201,6 +201,37 @@ def es_search_from_fqn(
             fields=fields,
         )
 
+    def es_search_from_name(
```
ulixius9, line 204: why can't we use the search from fqn?

for classifications name will be same as fqn

keshavmohta09: For bigquery we get tag name and classification name both in lowercase so we have to check tag names also So I used for both

ulixius9: yes but in case of classification you can pass fqn: my_classification and for tag you can use fqn: *.my_tag_name no?

keshavmohta09: we can use an existing method that checks fqn.
changes done

ulixius9 (approved):

PR #20441, opened 2025-03-26
Updated the count in tabs for containers

sweta1308:
Updated to have a count of Schema and Children tabs for container.

image

image

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/pages/ContainerPage/ContainerPage.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/pages/ContainerPage/ContainerPage.tsx
@@ -52,6 +52,7 @@ import { FeedCounts } from '../../interface/feed.interface';
 import {
   addContainerFollower,
   getContainerByName,
+  getContainerChildrenByName,
   patchContainerDetails,
   removeContainerFollower,
   restoreContainer,
@@ -94,6 +95,7 @@ const ContainerPage = () => {
   const [feedCount, setFeedCount] = useState<FeedCounts>(
     FEED_COUNT_INITIAL_DATA
   );
+  const [childrenCount, setChildrenCount] = useState<number>(0);
 
   const fetchContainerDetail = async (containerFQN: string) => {
     setIsLoading(true);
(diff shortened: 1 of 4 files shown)


openmetadata-ui/src/main/resources/ui/src/pages/ContainerPage/ContainerPage.tsx
```diff
@@ -166,6 +168,24 @@ const ContainerPage = () => {
...
 
+  // Fetch children count to show it in Tab label
+  const fetchContainerChildren = useCallback(
+    async (pagingOffset?: number) => {
+      try {
+        const { paging } = await getContainerChildrenByName(
+          decodedContainerName,
+          {
+            offset: pagingOffset ?? 0,
+          }
```
chirag-madlani, line 179: Not sure what's pagingOffset and we need it. Also can we set limit as 0 to avoid any data fetching here?

openmetadata-ui/src/main/resources/ui/src/pages/ContainerPage/ContainerPage.tsx
```diff
@@ -490,6 +511,7 @@ const ContainerPage = () => {
   // Effects
   useEffect(() => {
     fetchResourcePermission(decodedContainerName);
+    fetchContainerChildren();
```
Ashish8689, line 514: In case of only children tab, do we want this to be fetch since we will be fetching the Children with data we can use that data right ?

sweta1308: Yes fetch is necessary because the ChildrenTab component is independent of this ContainerPage parent component and also used in multiple places and not in ContainerPage. So, either we have to move the logic to Generic Provider or context. So, we need to fetch data here, like we are doing for Stored Procedure tab in Database Schema.

sweta1308: Also, as per @chirag-madlani 's comments, I have updated the fetch call to not fetch data and only the count.

openmetadata-ui/src/main/resources/ui/src/pages/ContainerPage/ContainerPage.tsx
sweta1308: Updated

chirag-madlani (approved):

PR #20486, opened 2025-03-28
Fix : Feeds not getting updated on user profile change.

shrushti2000:
Fixed the issue around feeds not getting updated on changing the user profile in User Profile page. The feeds were getting updated only on refresh.

Issue :-

When the user was changed, the loader state was not reseted to true , due to which it was trying to render the child components with stale values, resulting into making feeds api call with previous/old user id.

Fix:-
 Have reseted the loader state on change of the user 

https://github.com/user-attachments/assets/426f0c60-893a-4c7d-b92b-a7155ab95f13

Apart from this , the PR also fixes issue around feeds api call being made twice (one normal feeds call and one with 'after' payload )on initial load on user page , because of the intersection observer being visible even before the feeds are loaded.
Fix:- 
The intersection observer is only attached  now after the initial feeds are loaded.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/pages/UserPage/UserPage.component.tsx
```diff
@@ -179,7 +180,7 @@ const UserPage = () => {
     fetchUserData();
   }, [username]);
 
-  if (isLoading) {
+  if (userData?.name !== username || isLoading) {
```
chirag-madlani, line 183: Not sure on this condition, Isn't data fetching should be handling loading.
I think we can remove this condition

shrushti2000: There was 1 additional feeds(normal) call happening when then user profile was changed along with the feeds call happening for that particular user.
This additional condition was added to prevent that additional feeds api call.

chirag-madlani (approved):

PR #20498, opened 2025-03-28
MINOR: Adding a ManualRetry Workflow Node in order to be able to retry a given workflow from a given step

IceS2:
In order to be able to Retry the **Day One Experience Workflow** from a given step in case critical Agents fail for external reasons (Wrong connection setup, missing permissions, etc) we developed a `Manual Retry` workflow step that allows us to retry a given workflow instance from a given step.

Also added a couple info logs for more clarity on what Agents are created/deployed/ran/failing

#
### Type of change:

- [x] New feature

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
@@ -1,5 +1,9 @@
 package org.openmetadata.service.governance.workflows;
 
+import static org.openmetadata.service.governance.workflows.Workflow.GLOBAL_NAMESPACE;
+import static org.openmetadata.service.governance.workflows.Workflow.MANUAL_RETRY_MESSAGE;
+import static org.openmetadata.service.governance.workflows.Workflow.RETRY_LOCK_VARIABLE;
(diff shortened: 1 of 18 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
```diff
@@ -438,4 +444,83 @@ public void terminateWorkflow(String workflowName) {
...
+        LOG.info(
+            String.format("WorkflowInstance %s retrying from %s", workflowInstance, stepToRetry));
+        break;
+      } else {
+        retries++;
+        try {
+          Thread.sleep(retryIntervalMillis);
+          retryIntervalMillis = retryIntervalMillis * 2;
+        } catch (InterruptedException ex) {
+          throw new RuntimeException(ex);
```
IceS2, line 521: @mohityadav766 is there anything we could do here to avoid blocking the request

pmbrull (approved):

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
mohityadav766: Do we not have retryable in flowable?

PR #20512, opened 2025-03-29
MINOR: apply new tab design overall application where required

Ashish8689:
### Describe your changes:

- I worked on using new tab design overall application where required

image
image

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ShaileshParmar11 (commented):
Instead of adding class to each and every tab, can we find some way where we can make change to global style and  new style apply to across the application, it will help us to do future development fast, 
With current approach we have remember to add this new class to tabs every time we use that.

Ashish8689:
> Instead of adding class to each and every tab, can we find some way where we can make change to global style and new style apply to across the application, it will help us to do future development fast, With current approach we have remember to add this new class to tabs every time we use that.

We won't be doing this since there is multiple varient of tab, and we don't need to apply this everywhere. So will be sticking with this.

openmetadata-ui/src/main/resources/ui/src/styles/components/tabs.less
```diff
@@ -78,8 +78,6 @@
 
   .ant-tabs-tabpane {
     margin-top: @padding-sm;
-    border-radius: @border-radius-sm;
-    background-color: @white;
```
chirag-madlani, line 82: We need bg color no?

chirag-madlani (approved):

PR #20573, opened 2025-04-02
Fix: dbt cloud latest run execution

SumanMaharana:
### Describe your changes:

Fix: dbt cloud latest run execution

Added a new field `numberOfRuns` to control how many runs to fetch. (Users may have as many as 6k runs)
Added order by created time to get the latest runs first

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-spec/src/main/resources/json/schema/entity/services/connections/pipeline/dbtCloudConnection.json
```diff
@@ -55,6 +55,12 @@
           "type": "string"
         }
       },
+      "numberOfRuns": {
+        "title": "Number of Runs",
+        "description": "Number of runs to fetch from DBT cloud",
+        "type": "integer",
+        "default": null
```
harshsoni2024, line 62: should we provide any default value ? otherwise it may cause issue [here](https://github.com/open-metadata/OpenMetadata/pull/20573/files#diff-2cd5e16ae1b5a6f7d219b19e347007a31053e4a8859ea6a4ba8e3e58f01551cfR170)

pmbrull (approved):

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/pipeline/dbtCloudConnection.json
SumanMaharana: we do have a default value there as `null`.
and it's handled with checks so when its null there is no limit and it will fetch all the runs available.
I have ran it with various combination of values and it works well
e.g. with default i.e. null, 0, 1, 50, 100, 200, 103

harshsoni2024 (dismissed):

chirag-madlani (dismissed):

harshsoni2024 (approved):

chirag-madlani (approved):

PR #20601, opened 2025-04-03
MINOR: move the export png logic to utils to work independently

Ashish8689:
### Describe your changes:

- I worked on moving the export logic to work independently, so it won't clash with other export type functionality.
- Added support for the manage button in DataInsightHeader to further extends on collate.
- Also fix the onClick event  on ManageButtonLabel when it is disbaled and should not be clickable and perform action.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/utils/Export/ExportUtils.ts
```diff
@@ -23,3 +26,49 @@ export const downloadImageFromBase64 = (
...
+          }
+        : {}),
+    },
+  })
+    .then((base64Image: string) => {
+      downloadImageFromBase64(base64Image, name, ExportTypes.PNG);
+    })
+    .catch((error) => {
+      throw error;
+    });
```
chirag-madlani, line 73: It's not making any sense to catch error and throw it again

PR #20613, opened 2025-04-03
Minor: Day one trigger button on service details page

aniketkatkar97:
I worked on the following:

- [x] Added the Trigger Day one application workflow for a service

https://github.com/user-attachments/assets/b5975ec6-43cc-4309-b9f8-0421f2d89dd6

- [x] Fixed the styling issue on the main tab

**Before:**

Screenshot 2025-04-03 at 6 58 29 PM

**After:**

Screenshot 2025-04-03 at 6 58 46 PM

- [x] Fixed the spacing issue on the Add Service Page

**Before:**

Screenshot 2025-04-04 at 2 08 54 PM

**After:**

Screenshot 2025-04-04 at 2 08 10 PM

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


chirag-madlani (approved):

pmbrull:
@aniketkatkar97 do we want the button to be enabled all the time? under which circumstances would we want to trigger it again?

Might be cleaner to keep it only within the Day 1 banner whenever we find an issue with the execution. Once Day 1 is done, not sure there's value on re-executing.

Happy to hear your thoughts

aniketkatkar97:
@pmbrull I'm working on adding a condition there, I forgot to add it earlier. So we will always show the button, but disable it when the workflow is already running.

The reason for not keeping it in the banner is that there is a possibility that the banner will not be displayed for some services. (ex. already existing services that the user ingested). Also, say even after the successful run, the user wants to trigger the workflow for changed configurations, they should be able to trigger that.

So after a discussion we decided to have it in the header itself, and disable it if the workflow is already in progress.
cc. @harshach

IceS2:
> @pmbrull I'm working on adding a condition there, I forgot to add it earlier. So we will always show the button, but disable it when the workflow is already running.
> 
> The reason for not keeping it in the banner is that there is a possibility that the banner will not be displayed for some services. (ex. already existing services that the user configured before 1.7). Also, say even after the successful run, the user wants to trigger the workflow for changed configurations, they should be able to trigger that.
> 
> So after a discussion we decided to have it in the header itself, and disable it if the workflow is already in progress. cc. @harshach

My main concern with this approach is that... do we want the users to feel like they should be restarting the workflow from there instead of having the pipelines configured, scheduled and that's it?

I think it might be too much a `first citizen` on the page when it should be a "retry from start" button, while we figure our the best way to retry from certain points.

chirag-madlani (approved):

PR #20725, opened 2025-04-09
[Fix-19741] Azure SSO callback getRefreshToken null 

mohityadav766:
### Describe your changes:

Updated code to handle null refresh token

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


TeddyCr (approved):
```
      if (refreshToken == null || refreshToken.getValue() == null) {
         throw new TechnicalException("No refresh token available to request new access token.");
       }
```

If we fall into this, the user will get logged out at that point correct?

mohityadav766:
> ```
>       if (refreshToken == null || refreshToken.getValue() == null) {
>          throw new TechnicalException("No refresh token available to request new access token.");
>        }
> ```
> 
> If we fall into this, the user will get logged out at that point correct? 

Yes @TeddyCr , yes it should logout the user.
The problem of not having a refresh token on login is mostly a config mistake with either the scope missing or the refresh disables on the idp side. We log that on login if the refresh token is null

PR #20736, opened 2025-04-09
Fixes Issue 20189: different check for successful connection for api

chrisrayrayne:
### Describe your changes:

Fixes 20189

I worked on the connection for REST API collections because http responses can be wrong, so a not so strict behaviour opens up a lot of doors

Please check, if this change lives up to your requirements and levels

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/api/rest/connection.py
```diff
@@ -68,21 +66,24 @@ def test_connection(
     """
 
     def custom_url_exec():
-        if client.status_code == 200 and any(
-            content_type in client.headers.get("content-type")
-            for content_type in ACCEPTED_CONTENT_TYPES
-        ):
+        if client.status_code == 200:
```
harshsoni2024, line 69: we will still need to verify content-type if whole json is according to openapi spec content.

chrisrayrayne: I moved the schema check to the custom_schema_exec, because to me, this is more appropriate. Connection check, in my opinion, just checks if connection is successful, but not the content or schema.

As a webserver, you can set false Content-Type, for example havin a json in a text/xml, is wrong, but this code would still be able to parse the json in the body.

As said in the issue, I do a proposal. If this is not what you want, we can delete this PR and close the issue.

ingestion/src/metadata/ingestion/source/api/rest/connection.py
```diff
@@ -68,21 +66,24 @@ def test_connection(
...
+            "Failed to connect to the JSON schema url. Please check the url and credentials. Status Code was: " + str(client.status_code)
         )
 
     def custom_schema_exec():
-        if client.json().get("openapi") is not None:
-            return []
-        raise InvalidOpenAPISchemaError(
-            "Provided schema is not valid OpenAPI JSON schema"
-        )
+        try:
```
harshsoni2024, line 76: here also I think if response status is 200 OK & content type matched then this single condition would be enough.

You can test these things for some invalid json schemas 
https://api.github.com/users/octocat
https://jsonplaceholder.typicode.com/posts/1

chrisrayrayne: Sure, will do that. Thanks for your response.
I did test with a Google JSON OpenAPI Spec, a Google YAML OpenAPI Spec and an invalid URL. These tests will be a good benefit.

ingestion/src/metadata/ingestion/source/api/rest/connection.py
harshsoni2024: Okay, I see you are suggesting to move content-type check in custom_schema_exec.
But my concern is if url with false content-type is provided then it would still fail on next check(custom_schema_exec)? if I'm understanding correctly.

chrisrayrayne: Maybe I can give just a little background, why I suggest to move the schema check out of the connection-step. When I was troubleshooting my BitBucket-connection, it always failed in connection-step. Also, my fault, I did not see, that only JSON OpenAPI specs are supported. So I had to look at the URL, the Token and, because I misunderstood the step, also the Content-Type. For me it would have been much easier to have less checks and the status code in the log.

And for the content-type check, I see the point to pin down very easy, if it should be a json or not. But, as I saw, you already had to extend the compatible content-types, and there might come even more. With the URL https://api.github.com/users/octocat I get a content-type 'application/json; charset=utf-8' which actually states a good JSON, but it would be fail on the now implemented check. Of course, this is not an OpenAPI spec, but still a valid JSON.
Also, not checking the content-type, rather just the content, there might also be the opportunity to have file-server-connection, when they just tell no content-type (my experience) or with just application/octet-stream.

ingestion/src/metadata/ingestion/source/api/rest/connection.py
harshsoni2024: Yes sure, if that's the case we are open to changes 🙂.

ingestion/src/metadata/ingestion/source/api/rest/connection.py
harshsoni2024: Yes I see the point of removing content-type validation from this check because if JSON is still valid then it should pass. Also we're validating "openapi" key in next check so it will handle json spec validation.
Can you give any example if possible of the last point you mentioned about file-server connection where it's valid OpenAPI spec but not as json content-type ? So we can test with that and approve this.

harshsoni2024 (approved):

PR #20785, opened 2025-04-12
reverted table fixed header changes

Ashish8689:
### Describe your changes:

- I worked on reverted table fixed header changes

https://github.com/user-attachments/assets/b9e7d9f3-b812-480f-ae0e-5f16ce19f902

image

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/common/Table/Table.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/common/Table/Table.tsx
@@ -311,11 +311,6 @@ const Table = <T extends Record<string, unknown>>(
           ref={ref}
           tableLayout="fixed"
           {...resizingTableProps}
-          scroll={{
-            y: 740,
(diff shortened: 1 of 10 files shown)


openmetadata-ui/src/main/resources/ui/src/components/common/Table/Table.tsx
```diff
@@ -311,11 +311,6 @@ const Table = <T extends Record<string, unknown>>(
           ref={ref}
           tableLayout="fixed"
           {...resizingTableProps}
-          scroll={{
```
chirag-madlani, line 314: I think we should allow to set scroll from parent no?

Ashish8689: so, from the rest we are apply the scroll from the parent component.
image

chirag-madlani (approved):

PR #20788, opened 2025-04-12
fix(ingestion): Kinesis missing nexttoken 

belarusian:
### Describe your changes:

  Fixes issue with Kinesis ingestion failing when there are no more shards

  

Kinesis connector fails with a Pydantic validation error if there are no more shards to list. The AWS Kinesis API doesn't include
a `NextToken` field in its response when there are no more shards, but our `KinesisPartitions` model expects this field to be present, causing validation errors like:

  ValidationError: 1 validation error for KinesisPartitions
  NextToken Field required [type=missing, input_value={'Shards': [{'ShardId': '...'}, 'RetryAttempts': 0}}, input_type=dict]

  While the `KinesisPartitions` model defines `NextToken` as `Optional[str]`, Pydantic still requires the field to be present in the input dictionary. My change ensures the `NextToken` field always exists in the API response dictionary before model validation, preventing the validation error.

  I tested this change by running the Kinesis connector against a stream with few shards and validating that the ingestion completes successfully.

  

  #
  ### Type of change:
  
  - [x] Bug fix
  - [ ] Improvement
  - [ ] New feature
  - [ ] Documentation

  #
  ### Checklist:
  
  - [x] My PR title is `fix(ingestion): Handle missing NextToken field in Kinesis response`

  

  
### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
@@ -612,6 +612,7 @@ class DashboardServiceSource(TopologyRunnerMixin, Source, ABC):
         patch_request = PatchRequest(
             original_entity=original_entity,
             new_entity=original_entity.model_copy(update=create_request.__dict__),
+            override_metadata=self.source_config.overrideMetadata,
         )
         if isinstance(original_entity, Dashboard):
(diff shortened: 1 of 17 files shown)


pmbrull:
cc @SumanMaharana

harshsoni2024 (approved):

PR #20903, opened 2025-04-22
Allow for domain-level permissions while creating/updating dataProduct

sonika-shah:
### Describe your changes:

Changes  - 
-	Allow checking for root-level/container parent during resource creation for self-containing entities.
	-	Entities like GlossaryTerm, Tag, Domain, and Container already have their parent field set to themselves →
(e.g., GlossaryTerm → GlossaryTerm / Glossary (in case of root term), Tag → Tag / Classification, Domain → Domain, Container → Container)
as defined in [EntityRepository.java](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java#L2441) (overridden per entity).
	-	So, in the case of nested entities, they are currently unable to check root parent permissions.
With this fix, we will also check for the root-level parent when verifying permissions.

Fixes  the following usecase - 

Usecase - 
- added new rule in Datconsumer to allow only domain owners to add data Product to particular domain

image

Earlier it used to fail for domain owner  - 

image

With current pr changes - 

a. if its a domain owner 
image

b . if its not a domain owner 

image

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


sushi30 (dismissed):

chirag-madlani (approved):

karanh37 (dismissed):

karanh37 (approved):

chirag-madlani (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/ResourceContext.java
```diff
@@ -73,7 +81,36 @@ public List<EntityReference> getOwners() {
...
+    }
+
+    return owners;
+  }
+
+  private EntityInterface resolveParentEntity(T entity) {
+    Fields fields = new Fields(new HashSet<>(Collections.singleton(FIELD_OWNERS)));
+    try {
+      EntityReference parentReference =
+          switch (entityRepository.getEntityType()) {
```
harshach, line 100: shouldn't this be for schema and database as well?

openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/ResourceContext.java
```diff
@@ -73,7 +81,36 @@ public List<EntityReference> getOwners() {
     } else if (Entity.USER.equals(entityRepository.getEntityType())) {
       return List.of(entity.getEntityReference()); // Owner for a user is same as the user
     }
-    return entity.getOwners();
+
+    // Check for parent owners
+    List<EntityReference> owners = entity.getOwners();
+    EntityInterface parentEntity = resolveParentEntity(entity);
+    if (parentEntity != null && parentEntity.getOwners() != null) {
```
harshach, line 88: did we check the performance if we are looking up for parent everytime?

sonika-shah: we have kept it fail fast — it checks only for case match and fetches the field of parent only if it matches switch case

openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/ResourceContext.java
sonika-shah: For now as discussed in standup we have kept changes limited to dataProduct <-> domain,  glossary <-> glossaryTerm, tag <-> classification for 1.6.11

karanh37 (approved):

PR #20918, opened 2025-04-22
Fix: User playwright test

shrushti2000:
This PR fixes playwright test  'Check permissions for Data Steward ' of  User.spec
The test was failing because it was not able to find the table checkbox in Assets Filter. The cause of this issue was the aggregate call with size limit 10 happening before the query call sometimes.(which was not having  table in response).
As a fix for this test, have added wait for the query api call.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/utils/user.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/utils/user.ts
@@ -575,6 +575,7 @@ export const checkStewardServicesPermissions = async (page: Page) => {
   // Click on the sidebar item for Explore again
   await sidebarClick(page, SidebarItem.EXPLORE);
 
+  await page.waitForResponse('/api/v1/search/query?q=*');
   // Perform search actions
   await page.click('[data-testid="search-dropdown-Data Assets"]');
   await page.locator('[data-testid="table-checkbox"]').scrollIntoViewIfNeeded();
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/playwright/utils/user.ts
```diff
@@ -575,6 +575,7 @@ export const checkStewardServicesPermissions = async (page: Page) => {
   // Click on the sidebar item for Explore again
   await sidebarClick(page, SidebarItem.EXPLORE);
 
+  await page.waitForResponse('/api/v1/search/query?q=*');
```
Ashish8689, line 578:
```suggestion
 const queryResponse =  page.waitForResponse('/api/v1/search/query?q=*');
 await sidebarClick(page, SidebarItem.EXPLORE);
 await queryResponse;
 
```

Ashish8689: @shrushti2000 let's await like this, as in the previous case, there can be a chance as, before reaching API await line the API might be resolved, and test will broke on the await line.

ShaileshParmar11: Agree with @Ashish8689, Please update the changes @shrushti2000

Ashish8689 (approved):

PR #20922, opened 2025-04-22
Fixes 20876: BigQuery String Type

### LINKED ISSUE #20876: BigQuery’s STRING type is mapped to VARCHAR
This mapping can't show the users the native data type on a BigQuery table which has STRING column.

**Describe the bug**
BigQuery’s STRING type is mapped to VARCHAR

**To Reproduce**

Ingest metadata from a table which contains STRING column, then checking it in OpenMetadata, the STRING column shows VARCHAR type

**Expected behavior**
I need to retain the native data type

**Version:**
 - OpenMetadata version: 1.6.5

keshavmohta09:
### Describe your changes:

Fixes #20876 

I worked on a fix to use the same BigQuery string type as the datatype display

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
```diff
@@ -154,6 +154,18 @@ def _array_sys_data_type_repr(col_type):
...
+    Get the system data type for the column type
+    """
+    if isinstance(col_type, String):
+        return "string"
+    if str(col_type) == "ARRAY":
+        return _array_sys_data_type_repr(col_type)
+
+    return str(col_type)
+
+
```
ulixius9, line 168: actually a better way to do this would be to use `field.field_type` we should replicate the logic in function `_get_sqla_column_type` for building the display data type string

keshavmohta09: `_get_sqla_column_type` returns the `String()` class, which gets converted to `varchar` when cast to a string. 
In `SqlColumnHandlerMixin`, the data type display is based on `system_data_type`. 
If I use field.field_type generically for all data types, it could unintentionally affect the data type display for other types as well.

ulixius9: yes but that should be the expected behaviour, purpose of `system_data_type` is to show exactly what we have on the source side, for example if you have a column with type `FLOAT64` it should show `FLOAT64` but the current code will convert it into `FLOAT` which should not happen

ulixius9 (approved):

PR #21049, opened 2025-04-30
Fix(ui): Task assignee new styling

dhruvjsx:
I worked on new styling of Task assignee. Also removed fixed width from owner component and made it adapt from its parent 

Before:
Screenshot 2025-04-30 at 11 27 51 AM
Screenshot 2025-04-30 at 11 28 02 AM
Screenshot 2025-05-13 at 1 27 54 PM

After:

https://github.com/user-attachments/assets/7d69f03d-fb26-44c9-a2ea-8c5b3d4cddda

Screenshot 2025-05-13 at 1 29 39 PM

#
### Type of change:

- [x] Improvement

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/components/ActivityFeed/TaskFeedCard/TaskFeedCardNew.component.tsx
```diff
@@ -359,12 +358,12 @@ const TaskFeedCard = ({
                       ? 'task-card-assignee'
                       : ''
                   }`}>
-                  <AssigneesIcon height={20} width={20} />
                   <OwnerLabel
                     avatarSize={24}
                     isCompactView={false}
                     owners={feed?.task?.assignees}
                     showLabel={false}
+                    showMultipleType={{}}
```
chirag-madlani, line 366: Why we need to pass empty object as prop?

openmetadata-ui/src/main/resources/ui/src/components/common/OwnerLabel/OwnerLabel.interface.ts
```diff
@@ -29,4 +29,11 @@ export interface OwnerLabelProps {
   tooltipText?: string;
   isCompactView?: boolean;
   avatarSize?: number;
+  showMultipleType?: {
```
chirag-madlani, line 32: showMultipleType seems a boolean from name and acting as object
Can you come up with something generic? 
 Like for ownerLabel it will be like `showEdit`? and caller may identify it based on different condition

openmetadata-ui/src/main/resources/ui/src/components/ActivityFeed/TaskFeedCard/TaskFeedCardNew.component.tsx
dhruvjsx: Object was introduced because we wanted to show edit component which has few dependency. but for this particular card we wanted the same view but not edit. So to have a non-falsy value I gave it an empty object.

openmetadata-ui/src/main/resources/ui/src/components/common/OwnerAvtar/OwnerAvatar.tsx
```diff
@@ -13,27 +13,77 @@
...
+  if (multiEntityConfig) {
+    return (
+      <div className="flex w-max-full items-center gap-2">
+        {owner.type === OwnerType.TEAM ? (
+          <div className="d-flex gap-2 multi-team-container w-max-full items-center">
+            <Icon
+              className="owner-team-icon"
+              component={AssigneesIcon}
+              data-testid={!isCompactView && getEntityName(owner)}
+              style={{ fontSize: '16px', color: 'white' }}
```
ShaileshParmar11, line 56: instead of style can you please add class?

openmetadata-ui/src/main/resources/ui/src/components/common/OwnerLabel/OwnerLabel.component.tsx
```diff
@@ -85,16 +94,107 @@ export const OwnerLabel = ({
...
+  const renderMultipleType = useMemo(() => {
+    return (
+      <div className="flex-wrap w-full d-flex relative items-center">
+        <div className="flex w-full gap-2 flex-wrap relative">
+          {showMultipleTypeTeam.map((owner, index) => (
+            <div
+              className="w-max-full"
+              key={owner.id}
+              style={{
+                zIndex: showMultipleTypeTeam.length - index,
```
ShaileshParmar11, line 115: why zindex is based on length? it can effect on modals,

openmetadata-ui/src/main/resources/ui/src/components/common/OwnerLabel/OwnerLabel.component.tsx
```diff
@@ -85,16 +94,107 @@ export const OwnerLabel = ({
...
+                marginRight: '-4px',
+                position: 'relative',
```
ShaileshParmar11, line 117: instead of style lets use class

ShaileshParmar11 (commented):
added few comment

openmetadata-ui/src/main/resources/ui/src/components/common/OwnerLabel/OwnerLabel.component.tsx
dhruvjsx: The length won't be more then 4 since we are using slice above

Ashish8689: @dhruvjsx we don't want to use z-index, since it can be overlayed on some element if there will be something on it. Instead follow the reverse method and CSS flex-row-reverse to achieve the same UX and get rid of the z-index here.

openmetadata-ui/src/main/resources/ui/src/components/common/OwnerLabel/OwnerLabel.component.tsx
```diff
@@ -87,9 +90,92 @@ export const OwnerLabel = ({
...
+            </div>
+          ))}
+          <div className="flex relative">
+            {showMultipleTypeVisibleUser.map((owner, index) => (
+              <div
+                className="owner-item-container relative"
+                key={owner.id}
+                style={{
+                  zIndex: showMultipleTypeVisibleUser.length - index,
+                }}>
```
Ashish8689, line 125: Do we need to use zIndex here ?

PR #21074, opened 2025-05-04
MINOR: data sample ingestion bigquery

TeddyCr:
### Describe your changes:
**What**
Fix 1.7.0 regression when ingesting sample data.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Bug fix


ingestion/src/metadata/sampler/sqlalchemy/bigquery/sampler.py
```diff
@@ -130,62 +130,3 @@ def get_sample_query(self, *, column=None) -> Query:
             )
 
         return super().get_sample_query(column=column)
-
-    def fetch_sample_data(self, columns: Optional[List[Column]] = None) -> TableData:
```
TeddyCr, line 134: the multi project issue was fixed [here](https://github.com/open-metadata/OpenMetadata/pull/20779). This is not needed and was ofuscating the base implementation

ingestion/src/metadata/sampler/sqlalchemy/sampler.py
```diff
@@ -199,25 +199,12 @@ def fetch_sample_data(self, columns: Optional[List[Column]] = None) -> TableData
                 if col.name != RANDOM_LABEL and col.name in names
             ]
         with self.get_client() as client:
-            try:
```
TeddyCr, line 202: The `try` / `catch` offuscate some error -- unsure why we introduced this.

e.g. passing a `MONTH` partition on a BQ timestamp will throw an error (as only DAY and HOUR can be used), which will be caught by the `Exception` which in turn will run the query against the raw dataset and the user will see a `partition field required ...` error message (instead of the correct one)

harshach (approved):

pmbrull (approved):

PR #21162, opened 2025-05-13
chore: fix flaky Entity playwright tests

pranita09:
### Describe your changes:

Fix flaky tests:
- Follow & Un-follow entity
- Delete Service -> Hard Delete

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/playwright/utils/common.ts
```diff
@@ -122,8 +123,12 @@ export const toastNotification = async (
 ) => {
   await expect(page.getByTestId('alert-bar')).toHaveText(message, { timeout });
 
+  await page.waitForSelector('[data-testid="alert-icon"]');
+
   await expect(page.getByTestId('alert-icon')).toBeVisible();
 
+  await page.waitForSelector('[data-testid="alert-icon-close"]');
+
   await expect(page.getByTestId('alert-icon-close')).toBeVisible();
```
ShaileshParmar11, line 132: Why are two separate wait conditions needed here? If the alert-icon exists, the close button should also be available within the same alert. Additionally, we first check for the message and then introduce another wait condition. If the alert-bar exists, shouldn't the icon and close button be present within the same alert bar?

ShaileshParmar11 (changes requested):

openmetadata-ui/src/main/resources/ui/playwright/utils/common.ts
pranita09: Yes, corrected the code. Instead of two separate wait conditions for alert-icon and alert-icon-close, used single wait for alert-bar. So both icon and close button will be visible.

ShaileshParmar11 (approved):

PR #21200, opened 2025-05-14
Feature/pii column classifier

pmenal:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/setup.py
+++ b/ingestion/setup.py
@@ -437,6 +437,7 @@ test = {
     "python-liquid",
     VERSIONS["google-cloud-bigtable"],
     *plugins["bigquery"],
+    "faker>=37.0.0",
 }
 
 if sys.version_info >= (3, 9):
--- a/ingestion/src/metadata/pii/feature_extraction.py
+++ b/ingestion/src/metadata/pii/feature_extraction.py
@@ -1,0 +1,237 @@
+#  Copyright 2025 Collate
+#  Licensed under the Collate Community License, Version 1.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/LICENSE
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+import inspect
+import logging
+import re
+from collections import defaultdict
+from typing import (
+    Dict,
+    FrozenSet,
+    Hashable,
+    Iterable,
+    Mapping,
+    Optional,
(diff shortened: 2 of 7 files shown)


ingestion/src/metadata/pii/algorithms/classifiers.py
```diff
@@ -0,0 +1,171 @@
...
+        sample_data: Sequence[Any],
+        column_name: Optional[str] = None,
+        column_data_type: Optional[DataType] = None,
+    ) -> Mapping[T, float]:
+        """
+        Predict the scores for the given column and sample data of the column.
+        The scores are a mapping of class labels to their respective scores:
+        higher scores indicate a higher likelihood of the class for the given inputs.
+        """
+        pass
```
pmbrull, line 67: no need for `pass` if we add the docstring

ingestion/src/metadata/pii/algorithms/classifiers.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def predict_scores(
+        self,
+        sample_data: Sequence[Any],
+        column_name: Optional[str] = None,
+        column_data_type: Optional[DataType] = None,
+    ) -> Mapping[PIITag, float]:
+
+        if column_data_type is not None and is_non_pii_datatype(column_data_type):
+            return {}
```
pmbrull, line 102: nit - instead of empty structures, should we return `None` and handle that wherever this is used?

ingestion/src/metadata/pii/algorithms/presidio_utils.py
```diff
@@ -0,0 +1,117 @@
...
+        "model_name": model_name,
+    }
+
+    nlp_engine = SpacyNlpEngine(models=[model])
+    analyzer_engine = AnalyzerEngine(
+        nlp_engine=nlp_engine, supported_languages=[SUPPORTED_LANG]
+    )
+    for recognizer in _get_all_pattern_recognizers():
+        # Add the recognizer to the engine
+        recognizer.supported_language = SUPPORTED_LANG
```
pmbrull, line 54: nice, let's add a comment saying "we're patching this at runtime to allow registering all recognizers" or smt

ingestion/src/metadata/pii/algorithms/classifiers.py
pmenal: Fixed!

ingestion/src/metadata/pii/algorithms/classifiers.py
pmenal: But then you should check if the resulting mapping is empty anyway. I prefer to keep the returning type as simple as possible.

ingestion/src/metadata/pii/algorithms/presidio_utils.py
pmenal: Done!

PR #21222, opened 2025-05-15
Update ingestion runner fetch condition

sweta1308:
Updated the Add Service Page to not fetch the Ingestion Runner list when pipeline status i not Hybrid.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/constants/constants.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/constants/constants.ts
@@ -84,6 +84,7 @@ export const NO_DATA_PLACEHOLDER = '--';
 export const PIPE_SYMBOL = '|';
 export const NO_DATA = '-';
 export const STAR_OMD_USER = 'STAR_OMD_USER';
+export const AIRFLOW_HYBRID = 'Hybrid';
 
 export const TOUR_SEARCH_TERM = 'dim_a';
 export const ERROR500 = t('message.something-went-wrong');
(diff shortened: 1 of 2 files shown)


karanh37 (dismissed):

chirag-madlani (approved):

openmetadata-ui/src/main/resources/ui/src/components/Settings/Services/ServiceConfig/ConnectionConfigForm.tsx
```diff
@@ -50,10 +56,11 @@ const ConnectionConfigForm = ({
   disableTestConnection = false,
 }: Readonly<ConnectionConfigFormProps>) => {
   const { inlineAlertDetails } = useApplicationStore();
+  const [ingestionRunner, setIngestionRunner] = useState<string | undefined>();
 
   const formRef = useRef<Form<ConfigData>>(null);
 
-  const { isAirflowAvailable } = useAirflowStatus();
+  const { isAirflowAvailable, platform } = useAirflowStatus();
```
chirag-madlani, line 63: I think we don't need this state variable, You can use useMemo to determine it's value instead useEffect

openmetadata-ui/src/main/resources/ui/src/constants/constants.ts
```diff
@@ -84,6 +84,10 @@ export const NO_DATA_PLACEHOLDER = '--';
 export const PIPE_SYMBOL = '|';
 export const NO_DATA = '-';
 export const STAR_OMD_USER = 'STAR_OMD_USER';
+export const AIRFLOW_HYBRID = 'Hybrid';
+export const COLLATE_SAAS = 'CollateSaaS';
+export const COLLATE_SAAS_RUNNER = 'Collate SaaS Runner';
```
chirag-madlani, line 89: Not sure with this const, is it coming from BE?

openmetadata-ui/src/main/resources/ui/src/components/Settings/Services/ServiceConfig/ConnectionConfigForm.tsx
sweta1308: We cannot replace the state and useEffect in this specific case with useMemo because it doesn't track asynchronous changes, like formRef.current?.state?.formData. This won't rerender when the value of form changes. So, we need useEffect for the same.

chirag-madlani (approved):

PR #21249, opened 2025-05-16
MINOR: Add data insights migrations to remove stale objects

IceS2:
Adding migrations and required methods on the search clients to remove the Stale Data Insights Objects:
- On 1.6.6  we implemented support fo `CLUSTER_ALIAS`, but migrations were missing. This PR aims to delete the Data Streams, ILM Policy, Component Templates and Index Templates without `CLUSTER_ALIAS` if this is present.
- On 1.7.1 we removed the necessity of an ILM. This PR also aims to dettach and delete any ILM.

#
### Type of change:

- [x] Improvement

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
@@ -2579,52 +2579,56 @@ public class ElasticSearchClient implements SearchClient {
   }
 
   @Override
-  public void removeILMFromIndexTemplate(String templateName) throws IOException {
+  public void removeILMFromComponentTemplate(String componentTemplateName) throws IOException {
     try {
-      // 1. Get the existing template
-      Request getRequest = new Request("GET", "/_index_template/" + templateName);
+      // 1. Get the existing component template
+      Request getRequest = new Request("GET", "/_component_template/" + componentTemplateName);
       es.org.elasticsearch.client.Response getResponse =
           client.getLowLevelClient().performRequest(getRequest);
       String responseBody = org.apache.http.util.EntityUtils.toString(getResponse.getEntity());
       com.fasterxml.jackson.databind.JsonNode templateNode =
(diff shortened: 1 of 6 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -2403,4 +2407,232 @@ private ElasticSearchSourceBuilderFactory getSearchBuilderFactory() {
...
+
+      if (!templateNode.has("component_templates")
+          || templateNode.get("component_templates").isEmpty()) {
+        LOG.warn("Component template {} does not exist", componentTemplateName);
+        return;
+      }
+
+      // 2. Update the template in place
+      com.fasterxml.jackson.databind.JsonNode template =
+          templateNode.get("component_templates").get(0).get("component_template");
```
pmbrull, line 2619: nit - maybe "component_templates" could be a constant

pmbrull: same with other pieces we're calling a few times

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v172/MigrationUtil.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+  @SneakyThrows
+  private static void deleteDataInsightsDataStreams(SearchClient searchClient) {
+    try {
+      // Get and delete data streams that don't have cluster alias
+      List<String> dataStreams =
+          searchClient.getDataStreams(String.format("%s-*", DATA_INSIGHTS_PREFIX));
+      for (String name : dataStreams) {
```
pmbrull, line 64: lets review the behavior of any of these ops to ensure they don't fail in the second run (e.g., imagine force migrations, or if a cluster is "new" and does not have old ILM etc.). We should manage these cases gracefully and log accordingly

pmbrull (approved):

PR #21253, opened 2025-05-17
ISSUE #20962 - Issues with test case alerts

### LINKED ISSUE #20962: Alert Triggered on Both Failure and Success of the Test Case
**Affected module**
-- backend

**Describe the bug**
-- We have an alert configured to trigger **only on failure**. Earlier today, the alert was triggered correctly due to a failure. However, since that initial failure, we are now receiving notifications **even when the alert passes**, which is not expected behavior.

**To Reproduce**

-- 1. Set an alert to trigger only on failure.
2. Trigger a failure (alert correctly fires).
3. Let the condition pass in subsequent runs.
4. Observe that notifications continue even when the condition passes.

**Expected behavior**
--- The alert should only trigger notifications when the condition (failure) is met. Success events should not send notifications if the alert is set to only trigger on failure.

**Version:**
 - Python version:

**Additional context**

### LINKED ISSUE #21121: Bug: Test case name having issues with alerts not working - REGEX
**Affected module**
Backend

**Describe the bug**

**To Reproduce**
The name should have space in fqn for the test case / test suite + adding it to alerts, the alerts do not get triggered

FQN:
Core Prod.hades.creditApplication.customField.Test NULL PD Scores (pdv2) for Primary Borrower

TeddyCr:
### Describe your changes:

Fixes #20962 && fixes #21121

**What**
- Do not consolidate test case and test suite change events 👉 in local tests, I see change events being consolidated, where the change to the status goes from success to failed but is reverted to success (or inversingly) and hence the alerts is either triggered for a successful test or not triggered for a failing 
**[Maybe a better way to handle it than defaulting to False?]**
- record change to testCase testCaseResult field
- remove non-indexable fields on testSuites field

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
@@ -759,7 +759,7 @@ public class TestCaseRepository extends EntityRepository<TestCase> {
           original.getUseDynamicAssertion(),
           updated.getUseDynamicAssertion());
       recordChange("testCaseStatus", original.getTestCaseStatus(), updated.getTestCaseStatus());
-      recordChange("testCaseStatus", original.getTestCaseResult(), updated.getTestCaseResult());
+      recordChange("testCaseResult", original.getTestCaseResult(), updated.getTestCaseResult());
     }
   }
 
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
@@ -710,6 +710,11 @@ public class TestCaseRepository extends EntityRepository<TestCase> {
       super(original, updated, operation);
     }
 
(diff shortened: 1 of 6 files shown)


aji-aju (approved):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -754,6 +759,7 @@ public void entitySpecificUpdate(boolean consolidatingChanges) {
           original.getUseDynamicAssertion(),
           updated.getUseDynamicAssertion());
       recordChange("testCaseStatus", original.getTestCaseStatus(), updated.getTestCaseStatus());
+      recordChange("testCaseStatus", original.getTestCaseResult(), updated.getTestCaseResult());
```
sonika-shah, line 762: @TeddyCr , i think it should be` recordChange("testCaseResult"`...

sonika-shah (approved):

PR #21435, opened 2025-05-28
FIX - Automation Workflows should not be updated by the SM & cleanup …

pmbrull:
…migration

### Describe your changes:

- UI should always be cleaning up the test connection after calling it. Those workflows are ephemeral.
- We're adding a migration here just in case while we investigate why workflows are dangling
- There's also no need to run the secrets manager on those workflows, since they need to be cleaned up anyway

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


IceS2 (approved):

openmetadata-service/src/main/java/org/openmetadata/service/secrets/SecretsManagerUpdateService.java
```diff
@@ -103,14 +102,6 @@ private void updateIngestionPipelines() {
     retrieveIngestionPipelines().forEach(this::updateIngestionPipeline);
   }
 
-  private void updateWorkflows() {
```
mohityadav766, line 106: @pmbrull is this only used for automation_workflow  for test connection? just thinking from the persepective if there are schema changes relevant to worflow , and we have encrypted something it will to be run from this flow to have correct update new schema which is encrypted?

pmbrull: automation_workflow should always be empty. it's only used to pass info between server (UI) <> client. No data should be kept there so we can remove the SM updates

PR #21490, opened 2025-05-30
Fix(UI): Update the file upload option input for service config

aniketkatkar97:
Fixes https://github.com/open-metadata/openmetadata-collate/issues/1619

I worked on adding an alternative to file upload, to directly input the file content

https://github.com/user-attachments/assets/1b4d026f-c887-41b3-94a4-51b63ead3626

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/components/common/Form/JSONSchema/JsonSchemaWidgets/PasswordWidget.tsx
```diff
@@ -25,28 +37,71 @@ const PasswordWidget: FC<WidgetProps> = (props) => {
...
-      placeholder={rest.placeholder}
-      readOnly={rest.readonly}
-      required={rest.required}
-      value={passwordWidgetValue}
-      onBlur={() => onBlur(rest.id, rest.value)}
-      onChange={(e) => onChange(e.target.value)}
-      onFocus={() => onFocus(rest.id, rest.value)}
-    />
+    <>
+      {isInputTypeFile ? (
```
chirag-madlani, line 67: Thinking out loud, but don't you think uiType file will always be rendering this widget now?
Have you check that actually we needed this behvaiour for all the places where we have uiType File? 
I may call it fileOrInput to be more clear

aniketkatkar97: We are currently using this only for the certificate inputs now. But I agree with you here. We should make that clear in the UI type if we want an alternative. Thanks.

chirag-madlani (approved):

PR #21503, opened 2025-06-01
 Fix #21457: SearchIndexing: Limit of total fields [1000] has been exceeded

### LINKED ISSUE #21457: SearchIndexing: Limit of total fields [1000] has been exceeded
**Affected module**
Backend, UI and Ingestion

**Describe the bug**
- OpenSearchException[OpenSearch exception [type=illegal_argument_exception, reason=Limit of total fields [3000] has been exceeded]]

**To Reproduce**
Extensive lineage with too many nodes and edges, column level lineages

reduce `index.mapping.total_fields.limit`

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/resources/elasticsearch/en/table_index_mapping.json
```diff
@@ -615,11 +616,67 @@
           }
         }
       },
-      "lineage": {
-        "type" : "object"
+      "upstreamLineage": {
```
mohityadav766, line 618: @harshach this needs to be added for other entities as well,  also dynamic : false might cause issues , since pipeline whole object is being put or we need to cleanup  below method in LineageRepository

  public static void addPipelineDetails(EsLineageData lineageData, EntityReference pipelineRef) {
    if (nullOrEmpty(pipelineRef)) {
      lineageData.setPipeline(null);
    } else {
      Pair<String, Map<String, Object>> pipelineOrStoredProcedure =
          getPipelineOrStoredProcedure(pipelineRef, List.of("changeDescription"));
      lineageData.setPipelineEntityType(pipelineOrStoredProcedure.getLeft());
      lineageData.setPipeline(pipelineOrStoredProcedure.getRight());
    }
  }

harshach: why would it cause the issue?

mohityadav766: setting "dynamic": false on a field (here, on the upstreamLineage object) prevents any fields that are not explicitly defined in the properties from being added to the mapping or indexed.

harshach: @mohityadav766 my question for you to review this to understand what we are missing, I want this to be explicitly mapped otherwise its leaving everything to be indexed which is what we don't want right

harshach:
@karanh37 @mohityadav766 
1. verify if the upstreamLineage is applied to all the indexes we need
2. check if the upstreamLineage definition in these indexes will work for all the searches we support
3. @ayush-shah need test data so that this can be reproduced

PR #21512, opened 2025-06-02
Scim Interface and resource in OM

aji-aju:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/resources/scim/ScimResource.java
```diff
@@ -0,0 +1,249 @@
+package org.openmetadata.service.resources.scim;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import javax.ws.rs.*;
```
harshach: avoid wild-card imports

openmetadata-service/src/main/java/org/openmetadata/service/resources/scim/ScimResource.java
```diff
@@ -0,0 +1,249 @@
...
+  @GET
+  @Path("/")
+  @Operation(
+      operationId = "getServiceProviderConfig",
+      summary = "Get SCIM Service Provider Config",
+      description = "Returns the SCIM service provider configuration.",
+      responses =
+          @ApiResponse(responseCode = "200", description = "SCIM Service Provider Configuration"))
+  public Response getServiceProviderConfig() {
+    return provisioningService.getServiceProviderConfig();
```
harshach, line 61: are we going to return to this configuration to all users? without authorization check? is this part of SCIM spec?

openmetadata-service/src/main/java/org/openmetadata/service/resources/scim/ScimResource.java
```diff
@@ -0,0 +1,249 @@
...
+
+  @GET
+  @Path("/Users")
+  @Operation(
+      operationId = "listScimUsers",
+      summary = "List SCIM users",
+      description = "Lists SCIM users based on optional filters.",
+      responses = @ApiResponse(responseCode = "200", description = "List of SCIM Users"))
+  public Response listUsers(@Context UriInfo uriInfo) {
+    return provisioningService.listUsers(uriInfo);
```
harshach, line 88: again seems like there is no authZ check here? is that how SCIM operates?

openmetadata-service/src/main/java/org/openmetadata/service/resources/scim/ScimResource.java
```diff
@@ -0,0 +1,249 @@
...
+      operationId = "createScimGroup",
+      summary = "Create SCIM group",
+      description = "Creates a new SCIM group.",
+      responses = {
+        @ApiResponse(responseCode = "201", description = "Group created"),
+        @ApiResponse(responseCode = "400", description = "Invalid group input")
+      })
+  public Response createGroup(
+      ScimGroup group, @Context UriInfo uriInfo, @Context SecurityContext securityContext) {
+    return provisioningService.createGroup(group, uriInfo, securityContext);
```
harshach, line 227: how are we making sure these group hierarchy is being matintained? what happes if the user belongs to multiple groups in a hierarchy?

aji-aju: Hi @harshach,
We will not maintain group hierarchy as part of SCIM provisioning, as the SCIM 2.0 specification does not natively support nested group relationships.
In our implementation:
If a nested group already exists in OpenMetadata with type "Group", the user will be added directly to that group, regardless of its hierarchy.
If no such group exists (by name), a new group will be created directly under the root organization.
Additionally, while Azure AD allows nested groups to contain both users and groups, OpenMetadata does not support assigning users directly to parent groups. These structural differences can lead to inconsistencies, so we’ve kept our behavior flat.
I've already communicated this behaviour to HDI and am currently waiting for their response.

openmetadata-service/src/main/java/org/openmetadata/service/resources/scim/ScimResource.java
aji-aju: Removed wildcard imports

openmetadata-service/src/main/java/org/openmetadata/service/resources/scim/ScimResource.java
aji-aju: Added authorization. Checks if the logged in user is admin/ bot. Only then scim APIs can be accessed.

openmetadata-service/src/main/java/org/openmetadata/service/resources/scim/ScimResource.java
aji-aju: Added authorization. Checks if the logged in user is admin/ bot. Only then scim APIs can be accessed.

mohityadav766 (approved):

mohityadav766:
Scim Resource Test @aji-aju will handle in a different PR

PR #21581, opened 2025-06-04
MINOR: Initial implementation for our Connection Class

IceS2:
# Implements **BaseConnection Class**

BaseConection should be our base class on which we build the connection interface that will be used around the ingestion framework.

This is just the start that aims to organize how we manage connections around the code. In this state I've only done the following:

- Created an interface to return a `Client`  (SQLAlchemy Engine, AWS Client, etc)
- Created an interface for the `test_connection`
- Migrated the MySQL Connector code to this structure
- Updated the ServiceSpec specification to allow us to define a Connection Class for each connector.
- Updated the code related with test_connection and get_connection to take into account this change. As of now it will try to fetch the Connection class via the ServiceSpec and if it is not able it will fallback to the previous method.

#
### Type of change:

- [x] Improvement

#
### Checklist:


ingestion/src/metadata/ingestion/source/connections.py
```diff
@@ -28,10 +30,71 @@
...
+    connection: BaseModel,
+) -> Optional[Type[BaseConnection]]:
+    """
+    Helper method to get the connection class from the connection spec.
+    Returns the connection class if successful, None otherwise.
+    """
+    from metadata.utils.service_spec.service_spec import (  # pylint: disable=import-outside-toplevel
+        BaseSpec,
+        import_connection_class,
+    )
```
TeddyCr, line 45: Is this here as it creates some import issue if it is a top level import?

ingestion/src/metadata/ingestion/source/connections.py
```diff
@@ -41,11 +104,18 @@ def get_test_connection_fn(connection: BaseModel) -> Callable:
...
+    test_fn = _get_test_fn_from_service_spec(connection)
+    if test_fn:
+        return test_fn
+    # Fallback to default
     return import_connection_fn(
         connection=connection, function_name=TEST_CONNECTION_FN_NAME
     )
 
 
+# ------------------------------------------------------------
```
TeddyCr: nit: i think we can remove this

ingestion/src/metadata/ingestion/source/connections.py
IceS2: Exactly

TeddyCr (dismissed):

ingestion/tests/unit/test_importer.py
```diff
@@ -92,21 +88,3 @@ def test_import_bulk_sink_type(self) -> None:
...
-            connection=connection, function_name="get_connection"
-        )
-        self.assertIsNotNone(get_connection_fn)
-
-        self.assertRaises(
-            AttributeError,
-            import_connection_fn,
-            connection=connection,
-            function_name="random",
-        )
```
IceS2, line 112: Decided to remove this specific UnitTest since this method became an internal implementation from other methods such as the `get_connection_fn`.

Currently, this method would work for most scenarios but as we migrate this will work less and less and I believe it is not worth it to test this bit while we already test the `get_connection_fn` in many different places.

PR #21633, opened 2025-06-07
fix tier and tag playwright for Entity.spec

karanh37:
### Describe your changes:

Worked on Tier and Tag search test failure on Explore page.
Reason of failure was when filter is applied there are more than 10 records and resulting entity is not on page 1.
Added fix so that filter is applied for unique Tier/Tag so that it will appear on 1st page.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


chirag-madlani (approved):

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Entity.spec.ts
```diff
@@ -232,7 +233,14 @@ entities.forEach((EntityClass) => {
     test('Tag Add, Update and Remove', async ({ page }) => {
       test.slow(true);
 
-      await entity.tag(page, 'PersonalData.Personal', 'PII.None', entity);
+      await entity.tag(
+        page,
+        'PersonalData.Personal',
+        EntityDataClass.tag1.data.displayName,
+        EntityDataClass.tag1.data.name,
+        EntityDataClass.tag1.data.classification,
```
karanh37: no need to pass both, you should pass only fqn, and in the subsequent no concatenation is required

pranita09: Tried passing fqn, but fullyQualifiedName does not exist for TagData.
Screenshot 2025-06-09 at 3 01 32 PM

karanh37: It would be under tag1.responseData.fullyQualifiedName

pranita09: Used responseData instead of data for fullyQualifiedName

openmetadata-ui/src/main/resources/ui/playwright/support/entity/EntityClass.ts
```diff
@@ -203,18 +203,24 @@ export class EntityClass {
...
   }
 
-  async tier(page: Page, tier1: string, tier2: string, entity?: EntityClass) {
+  async tier(
+    page: Page,
+    tier1: string,
+    tier2: string,
+    tier2Fqn?: string,
+    entity?: EntityClass
+  ) {
```
ShaileshParmar11, line 212: Since you have added tier2Fqn in between, please make sure we are checking all the places about this change.

openmetadata-ui/src/main/resources/ui/playwright/support/entity/EntityClass.ts
```diff
@@ -226,13 +232,29 @@ export class EntityClass {
     await updateDescription(page, description);
   }
 
-  async tag(page: Page, tag1: string, tag2: string, entity?: EntityClass) {
+  async tag(
+    page: Page,
+    tag1: string,
+    tag2: string,
+    tag2Fqn?: string,
+    entity?: EntityClass
```
ShaileshParmar11, line 240: same goes here

openmetadata-ui/src/main/resources/ui/playwright/support/entity/EntityClass.ts
pranita09: Yes verified for other test cases as well i.e EntityDataConsumer.spec, EntityDataSteward.spec, ServiceEntity.spec, etc

openmetadata-ui/src/main/resources/ui/playwright/support/entity/EntityClass.ts
pranita09: Yes, same here as well, verified for other test cases i.e EntityDataConsumer.spec, EntityDataSteward.spec, ServiceEntity.spec, etc as they are optional params passing only inside Entity.spec.

PR #21638, opened 2025-06-08
issue-21370: db2 custom driver installation

### LINKED ISSUE #21370: Install the IBM_DB2 compatible driver if version is provided by the user
There's a licensing issue with the DB2 connector.

Although now it is possible to upload the license file, the installed DB2 driver bundled with the ibm_db package should match the license version.
A mismatch here may lead to unexpected behaviour or compatibility issues.

You can check a bit more about how to install a given driver in the docs: https://github.com/ibmdb/python-ibmdb/?tab=readme-ov-file#installation

I propose that we install the driver programatically based on a configuration passed by the user. You can check more about how to do that here: https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program

harshsoni2024:
### Describe your changes:

Fixes #21370 
DB2 custom version driver installation along with `ibm_db` package.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/db2/connection.py
+++ b/ingestion/src/metadata/ingestion/source/database/db2/connection.py
@@ -33,13 +33,28 @@ from metadata.ingestion.connections.builders import (
 )
 from metadata.ingestion.connections.test_connections import test_connection_db_common
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
+from metadata.ingestion.source.database.db2.utils import (
+    check_clidriver_version,
+    install_clidriver,
+)
 from metadata.utils.constants import THREE_MIN, UTF_8
+from metadata.utils.logger import ingestion_logger
+
+logger = ingestion_logger()
 
 
 def get_connection(connection: Db2Connection) -> Engine:
     """
     Create connection
     """
+    # Install ibm_db with specific version
+    clidriver_version = connection.clidriverVersion
+
+    if clidriver_version:
+        clidriver_version = check_clidriver_version(clidriver_version)
+        if clidriver_version:
+            install_clidriver(clidriver_version.value)
+
     # prepare license
     # pylint: disable=import-outside-toplevel
     if connection.license and connection.licenseFileName:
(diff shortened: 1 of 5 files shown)


ingestion/src/metadata/ingestion/source/database/db2/utils.py
```diff
@@ -61,3 +85,104 @@ def get_unique_constraints(
...
+        try:
+            with urlopen(url) as _:
+                return True
+        except URLError:
+            return False
+
+    if system == "darwin":  # macOS
+        machine = platform.machine().lower()
+        if machine == "arm64":  # Apple Silicon
+            default_clidriver_url = f"/{BASE_CLIDRIVER_URL}/macarm64_odbc_cli.tar.gz"
```
SumanMaharana: Are we sure we need the  `/`  before `BASE_CLIDRIVER_URL`
`BASE_CLIDRIVER_URL` starts with `https://`.
This results in URLs like:
`/https://public.dhe.ibm.com/ibmdl/export/pub/software/data/db2/drivers/odbc_cli/...` which is not valid.

Ashish8689 (dismissed):

ingestion/src/metadata/ingestion/source/database/db2/connection.py
```diff
@@ -33,13 +33,28 @@
...
 def get_connection(connection: Db2Connection) -> Engine:
     """
     Create connection
     """
+    # Install ibm_db with specific version
+    clidriver_version = connection.clidriverVersion
+
+    if clidriver_version:
+        clidriver_version = check_clidriver_version(clidriver_version)
+        if clidriver_version:
```
SumanMaharana, line 55: when an invalid version is specified we are returning `None`. 
so do we proceed without installing any driver ?

ingestion/src/metadata/ingestion/source/database/db2/utils.py
harshsoni2024: seems typo, will fix it

ingestion/src/metadata/ingestion/source/database/db2/connection.py
harshsoni2024: no, so ibm_db package comes with default/latest driver installed, in case user want to change driver from supported versions then we install it.
Will be adding doc on another pr.

SumanMaharana (approved):

PR #21719, opened 2025-06-11
MINOR: Implement dependency injection on ingestion

IceS2:
# Implement Dependency Injection on the Ingestion Framework

Created a new dependency injection system (dependency_injector.py) that:
- Provides a type-safe way to inject dependencies using Python's type hints
- Uses a singleton container to manage dependencies
- Supports dependency registration, overriding, and removal
- Handles thread-safe operations with RLock

Updated __init__.py to:
- Initialize the dependency container
- Register the default source loader as a dependency

Modified service_spec.py to:
- Use the new dependency injection system for the get_for_source method
- Inject the SourceLoader dependency, making it more flexible and testable
- Allow for easier mocking and testing of source loading functionality

This change aims to help us override behaviour when needed by injecting an overriding dependency.
A better explanation on how the Dependency Injection system here works is given on the package README.md

## Alternatives explored

- Using a third party package to take care of the dependency:
  - https://github.com/ivankorobkov/python-inject/tree/master
  - https://python-dependency-injector.ets-labs.org/containers/declarative.html
  
 Although both pacakges seem quite interesting, for what we need I decided to go with a short and simple in-house implementation to avoid adding another dependency to the already huge list we've got.

#
### Type of change:

- [x] Improvement

#
### Checklist:


ingestion/src/metadata/utils/dependency_injector/dependency_injector.py
```diff
@@ -0,0 +1,343 @@
...
+            return db.query(f"SELECT * FROM users WHERE id = {user_id}")
+
+        # Dependencies can also be passed explicitly
+        get_user(user_id=1, db=Database("postgresql://localhost:5432"))
+        ```
+    """
+
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        container = DependencyContainer()
```
TeddyCr, line 277: should that be hardcoded? What if I wanted a different container?

TeddyCr (commented):
Do we need the lock? I can't think of a use case where the @ingject would be invoked in different threads

ingestion/src/metadata/utils/dependency_injector/README.md
```diff
@@ -0,0 +1,154 @@
...
+Use the `@inject` decorator and `Inject` type to automatically inject dependencies:
+
+```python
+@inject
+def get_user(user_id: int, db: Inject[Database]) -> dict:
+    return db.query(f"SELECT * FROM users WHERE id = {user_id}")
+
+# The db parameter will be automatically injected
+user = get_user(user_id=1)
+```
```
TeddyCr, line 53: Do we have a use case for that?

TeddyCr (dismissed):

ingestion/src/metadata/utils/dependency_injector/README.md
IceS2: For what? Not sure I followed the question. This is just a README with a dummy example on how to use the module

ingestion/src/metadata/utils/dependency_injector/dependency_injector.py
IceS2: DependencyContainer is a Singleton that gets created as soon as `metadata` is imported.

On itself it is a container for all the dependencies that we might need to register, that could be overriden.

Do you have any use case where you could need a different one?

PR #21738, opened 2025-06-12
MINOR: Add injection to profiler

IceS2:
Waiting on #21719 
---

# Extract Profiler dependencies and use Dependency Injection

Following on the Dependency Injection PRs, here I am extracting logic that needs to be injected to be easily overriden for the Profiler flow. This is mainly:

- Metrics: The Enum we use to fetch the different metrics from the Registry
- ProfilerResolver: How we actually resolve which Sampler and ProfilerInterface are going to use

#
### Type of change:

- [x] Improvement

#
### Checklist:


ingestion/src/metadata/data_quality/validations/table/pandas/tableRowInsertedCountToBeBetween.py
TeddyCr, line 1: Any reasons to add that?

ingestion/src/metadata/ingestion/source/connections.py
```diff
@@ -113,6 +113,9 @@ def get_test_connection_fn(connection: BaseModel) -> Callable:
     )
 
 
+# ------------------------------------------------------------
```
TeddyCr, line 116: can we clean this up?

ingestion/src/metadata/profiler/processor/default.py
```diff
@@ -12,55 +12,54 @@
...
+from metadata.generated.schema.settings.settings import Settings
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
 from metadata.profiler.interface.profiler_interface import ProfilerInterface
 from metadata.profiler.metrics.core import Metric, add_props
-from metadata.profiler.metrics.registry import Metrics
 from metadata.profiler.processor.core import Profiler
+from metadata.profiler.registry import MetricRegistry
 
 
 def get_default_metrics(
```
TeddyCr, line 29: why do we use the metric registry directly here and not use the @inject?

ingestion/src/metadata/profiler/processor/metric_filter.py
```diff
@@ -33,22 +33,35 @@
...
     def __init__(
         self,
         metrics: Tuple[Type[TMetric]],
         global_profiler_config: Optional[ProfilerConfiguration] = None,
         table_profiler_config: Optional[TableProfilerConfig] = None,
         column_profiler_config: Optional[List[ColumnProfilerConfig]] = None,
+        metrics_registry: Inject[Type[MetricRegistry]] = None,
     ):
+        if metrics_registry is None:
+            raise DependencyNotFoundError(
```
TeddyCr, line 59: Do you need to this check here since you are doing in the wrapper in `dependency_injector.py`

```
            # Check if it's an Inject type
            if is_inject_type(param_type):
                dependency_type = extract_inject_arg(param_type)
                dependency = container.get(dependency_type)
                if dependency is None:
                    raise DependencyNotFoundError(
                        f"Dependency of type {dependency_type} not found in container. "
                        f"Make sure to register it using container.register({dependency_type.__name__}, ...)"
                    )
                kwargs[param_name] = dependency
```

ingestion/src/metadata/utils/dependency_injector/dependency_injector.py
```diff
@@ -206,10 +216,9 @@ def get(self, dependency_type: Type[Any]) -> Optional[Any]:
             ```
         """
         with self._lock:
-            type_name = dependency_type.__name__
-            factory = self._overrides.get(type_name) or self._dependencies.get(
-                type_name
-            )
+            factory = self._overrides.get(
```
TeddyCr, line 219: do we check overrides first as Collate dependencies will be registered in there?

ingestion/src/metadata/data_quality/validations/table/pandas/tableRowInsertedCountToBeBetween.py
IceS2: This is being used in the Pandas profiler to filter partitions and I wanted to reuse it on the spark one since I'm not implemented dataquality for now. It seemed decent enough to add here. If you prefer I could add a random utility method outside.

ingestion/src/metadata/ingestion/source/connections.py
IceS2: I had done that. My bad, I'll do it again xD

ingestion/src/metadata/profiler/processor/default.py
IceS2: Just because it is being passed from another place where it is being injected. We could go eitherway

ingestion/src/metadata/profiler/processor/metric_filter.py
IceS2: It is mostly for linter purposes, but I could remove it. At Runtime it shouldn't be an issue

ingestion/src/metadata/utils/dependency_injector/dependency_injector.py
IceS2: Well, we check overrides first as overrides is made to override registered dependencies. It will work for Collate but would also work for testing purposes

ingestion/src/metadata/profiler/processor/metric_filter.py
TeddyCr: I see, I think we can remove the raise if it is handled somewhere else. We can let the linter cry if it reduces the amount of lines

TeddyCr (approved):

ingestion/src/metadata/profiler/processor/metric_filter.py
IceS2: There's still one way this could happen. Although we're using dependency injection, you can manually override it by passing it a value. You could technically call the function by passing `None` and it would be valid and wouldn't be caught by the decorator as if you are explicitely passing a dependency it shouldn't try to inject it

PR #21760, opened 2025-06-13
Fix #1681 - Deleted Filter in Automations Not Working as Expected

sonika-shah:
### Describe your changes:

 Deleted Filter in Automations Not Working as Expected as the deleted parameter is conflicting with the queryFilter condition for deleted that is already applied in the API

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
@@ -509,7 +509,6 @@ public class GlossaryTermRepository extends EntityRepository<GlossaryTerm> {
               .withFetchSource(true)
               .withTrackTotalHits(false)
               .withSortFieldParam("_score")
-              .withDeleted(false)
               .withSortOrder("desc")
               .withIncludeSourceFields(new ArrayList<>());
       Response response = searchRepository.search(searchRequest, null);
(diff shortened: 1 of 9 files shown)


openmetadata-ui/src/main/resources/ui/playwright/utils/glossary.ts
```diff
@@ -1003,7 +1003,7 @@ export const createDescriptionTaskForGlossary = async (
     await assigneeField.click();
 
     const userSearchResponse = page.waitForResponse(
-      `/api/v1/search/query?q=*${value.assignee}**&index=user_search_index%2Cteam_search_index`
+      `/api/v1/search/query?q=*${value.assignee}**&index=user_search_index%2Cteam_search_index*`
```
harshach, line 1006: why ae we adding * at the end? @sonika-shah

sonika-shah: cc - @aniketkatkar97 ^

aniketkatkar97: @harshach it's the matcher we use in playwright tests to detect the API call where we are waiting. Since we have introduced deleted param here, we needed to update the API string. To make it future proof I have used `*` to match any parameter that we add.

chirag-madlani (approved):

PR #21796, opened 2025-06-17
Chore: Add github action to generate and commit types based on schema changes

aniketkatkar97:
I worked on the following:

**Issue:**
Earlier, we had type generation in the pre-commit step when JSON schema changes are made. This led to 2 problems.

1. Increased time for the commit.
2. The type generation for the schemas where the changed schema was referenced to were not updating.

**Fix:**
- Added a GitHub workflow to generate the types from the JSON schemas and commit the same in the PR.
- This will only run if changes are detected in the JSON schema folder or the generated types folder in the UI. (i.e. "openmetadata-spec/src/main/resources/json/schema/**" & "openmetadata-ui/src/main/resources/ui/src/generated/**")

**Tested Scenarios:**

- Should only trigger if changes are detected in above mentioned folder.
- If no type changes are detected after workflow completion, no commit should be done.
- The initial workflow triggered should be cancelled if a new commit is pushed to the PR.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


.github/workflows/typescript-type-generation.yml
```diff
@@ -3,18 +3,76 @@ name: TypeScript Type Generation
...
+          git checkout ${{ github.event.inputs.branch }}
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "18"
+
+      - name: Install dependencies
+        run: |
+          yarn install --frozen-lockfile
```
chirag-madlani, line 42: Don't we need to mention working directory here?

.github/workflows/typescript-type-generation.yml
```diff
@@ -3,18 +3,76 @@ name: TypeScript Type Generation
...
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+
+      - name: Commit and push changes
+        if: steps.git-check.outputs.changes == 'true'
+        run: |
+          git add openmetadata-ui/src/main/resources/ui/src/generated/
+          git commit -m "Update generated TypeScript types"
+          git push --force-with-lease
```
chirag-madlani: This will commit to main?

.github/workflows/typescript-type-generation.yml
aniketkatkar97: We want it to run in the root, so haven't mentioned the directory.

.github/workflows/typescript-type-generation.yml
aniketkatkar97: No, it'll commit to the branch we have checked out. i.e. will in the PRs

.github/workflows/typescript-type-generation.yml
```diff
@@ -3,18 +3,76 @@ name: TypeScript Type Generation
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.branch || github.event.pull_request.head.ref || github.ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Switch to branch (if manual trigger)
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          git checkout ${{ github.event.inputs.branch }}
```
akash-jain-10: This should not be required! `actions/checkout` step will take care of cloning the repo with right branch!

.github/workflows/typescript-type-generation.yml
akash-jain-10: Add `working-directory`!

akash-jain-10 (approved):

PR #21827, opened 2025-06-18
Fix explain turned by default, use dfs_query_then_fetch in cases of sharding of search cluster

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/SearchIndex.java
```diff
@@ -94,9 +94,6 @@ default Set<String> getExcludedFields() {
 
   default Map<String, Object> getCommonAttributesMap(EntityInterface entity, String entityType) {
     Map<String, Object> map = new HashMap<>();
-    map.put(
```
mohityadav766: any reason we are removing this from here?

PR #21832, opened 2025-06-18
Fix:#21647 Inherited Tags from Glossary Term should be shown in put column response

### LINKED ISSUE #21647: Tags inherited from Glossary Terms are not removed when the Term is unlinked from an Asset
**Affected module**
Does it impact the UI, backend or Ingestion Framework? -- Backend

**Describe the bug**
-- When a glossary term is added to an asset, the associated tags of that term are automatically inherited by the asset. However, upon removing the glossary term, the inherited tags remain linked to the asset, which creates inconsistency. These tags should be automatically removed if they were inherited and not directly added.

**To Reproduce**

- Add a Glossary Term (e.g., PhoneNumber) to an asset (e.g., CustomerPhoneEdpEvent).
- Ensure the term has associated tags (e.g., Sensitive, Personal) — these tags get automatically added to the asset.
- Remove the Glossary Term from the asset.
- Observe: The tags (Sensitive, Personal) still remain on the asset.

**Expected behavior**
- Tags inherited from a glossary term should be automatically removed from an asset when the term is unlinked — unless those tags were manually added independently.

**Version:**
 - Python version:

**Additional context**

yan-3005:
### Describe your changes:

Fixes #21647

Inconsistency in the put response of column when displaying inherited tags

https://github.com/user-attachments/assets/4858c1a5-fc58-4230-b800-6cafd240a536

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ColumnRepository.java
```diff
@@ -143,6 +144,8 @@ private Column updateTableColumn(
 
     tableRepository.patch(uriInfo, parentEntityRef.getId(), user, jsonPatch);
 
+    // Include Derived Tags in response
+    column.setTags(addDerivedTags(column.getTags()));
```
sonika-shah: Instead of adding this condition at very end, please add it with the relevant block of code

in the same file  line 127

 ```
if (updateColumn.getTags() != null) {
      column.setTags(updateColumn.getTags()); // Empty array = remove all tags
    }
```

sonika-shah (approved):

PR #21840, opened 2025-06-18
MINOR: Extend profile workflow config to allow engine configuration

IceS2:
# Summary

## Engine Configuration Structure
- Native Engine: Simple configuration with type "Native"
- Spark Engine: Includes remote URL and additional config properties
- Default behavior maintains Native engine when no processing engine is specified
- Existing pipelines continue to work without modification

## Repository Priority System
- Created a priority-based repository selection system
- Allows for multiple repository implementations with different priorities

# Changes

## Schema Extensions (JSON Schema Files)
- Created new engine configuration schemas
    - nativeEngineConfig.json - Configuration for the native metadata ingestion engine
    - sparkEngineConfig.json - Configuration for Spark Engine runner with remote URL and config properties
- Updated existing pipeline schemas
    - Added processingEngine field to databaseServiceProfilerPipeline.json
    - Added processingEngine field to ingestionPipeline.json
    - Added processingEngine field to createIngestionPipeline.json

## Python Code Changes (Ingestion Module)
- Profiler Resolver Updates
    - Modified profiler_resolver.py to accept ProcessingEngine type instead of string
    - Updated method signatures to use the new engine configuration type
- Profiler Source Interface
    - Enhanced profiler_source_interface.py to return ProcessingEngine type
    - Added default Native engine configuration when none is provided
- Sampler Interface
    - Extended sampler_interface.py to accept processing_engine parameter
    - Updated constructor and factory methods to handle the new engine configuration

## Java Backend Changes
- Repository System Enhancement
    - Enhanced the Repository.java annotation with name() and priority() attributes
    - Updated Entity.java to handle named repositories with priority-based selection
    - Added explicit repository name to IngestionPipelineRepository.java
- Ingestion Pipeline Management
    - Updated IngestionPipelineRepository.java to include processingEngine in update/patch fields
    - Modified IngestionPipelineMapper.java to map the new processing engine field

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/Entity.java
```diff
@@ -665,7 +665,28 @@ private static List<Class<?>> getRepositories() {
             .acceptPackages(PACKAGES.toArray(new String[0]))
             .scan()) {
       ClassInfoList classList = scanResult.getClassesWithAnnotation(Repository.class);
-      return classList.loadClasses();
+
```
TeddyCr, line 668: what does this do?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/Repository.java
```diff
@@ -9,4 +9,9 @@
 @Inherited
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE, ElementType.CONSTRUCTOR})
-public @interface Repository {}
+public @interface Repository {
```
TeddyCr, line 12: Same thing here, why do we need this change?

TeddyCr (commented):
Added a few questions

openmetadata-service/src/main/java/org/openmetadata/service/Entity.java
IceS2: It's a bit explained in the description. Basically what we did with @mohityadav766  is creating a system where we could override the repositories if needed be.

This would avoid adding logic to the OSS code that does nothing just because we need it in a wrapper.

This way we could do something like:

```
@repository(name = "ingestionPipelineRepository", priority = 0)
class MyIngestionPipelineRepository extends IngestionPipelineRepository {
   public void newFunctionality() {
   }
}
```

and that would override the default one with the extra functionality.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/Repository.java
IceS2: Same as the above answer

TeddyCr (approved):

ShaileshParmar11 (approved):

PR #21844, opened 2025-06-18
fix(test): AUT failures for Policies.spec and Roles.spec

pranita09:
### Describe your changes:

- This PR fixes AUT failures for Policies.spec and Roles.spec caused due to pagination.
- The element we need to locate is not present on the page, so we need to check next pages for it.
- Created a util function `getLocatorWithPagination` to find the element and used it inside tests.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/utils/roles.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/utils/roles.ts
@@ -45,10 +45,14 @@ export const getElementWithPagination = async (
     }
 
     const nextBtn = page.locator('[data-testid="next"]');
+    hasNext = !(await nextBtn.getAttribute('disabled'));
+
+    if (!hasNext) {
+      throw new Error('Element not found and no more pages to paginate.');
+    }
+
     await nextBtn.click();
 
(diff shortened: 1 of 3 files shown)


openmetadata-ui/src/main/resources/ui/playwright/utils/roles.ts
```diff
@@ -28,3 +28,27 @@ export const removePolicyFromRole = async (
...
+    }
+
+    const nextBtn = page.locator('[data-testid="next"]');
+    await nextBtn.click();
+
+    await page.waitForSelector('[data-testid="loader"]', { state: 'detached' });
+
+    hasNext = await nextBtn.isVisible();
+  }
+};
```
ShaileshParmar11: What if the element is on the 3rd or 4th page?

pranita09: Adjusted the function so that it will find the element even if it is on 3rd or 4th page.
Also confirmed the same scenario for 3rd page on UI.

PR #21867, opened 2025-06-19
Feature #20137 Move Glossary Terms to any level including other glossaries

### LINKED ISSUE #20137: It is inconvenient to transfer the term to the root level of the glossary
I ran into a problem related to the inconvenience of moving a child term to the root of the glossary.
If you select "Change Parent Term" from the term page, you can select only another parent term in the drop-down window.
On the glossary page, you can drag a term to the root level, but this will only work if you drag it to the headings of the table. This may not be easy if the glossary is quite large.

The best solution would be to add the ability to put the term at the root level on the menu screen "Change Parent Term".
Also on this screen, it would be great to add to the drop-down menu with terms the ability to enter the name of the desired term, so as not to scroll through the list.

An alternative option would be to add the ability to drag a term to the root level on the glossary page without having to drag it to the headings.

Image

yan-3005:
### Describe your changes:

Fixes #20137

1. Move Glossary Terms across Glossaries.
2. Move under any glossary term or root of the glossary
3. Async Implementation

Cases covered:

1. Tags, Assets assigned to a glossary term will be moved with after the term is moved.
2. Domain is inherited from Glossary to Glossary Term, but they still can have separate domains. So assume Glossary A -> Term A, Glossary B -> Term B. Both glossaries have domains assigned and terms are inheriting it. If Term B is moved under Glossary A, it will inherit the domain of Glossary A.
3. If Term has a domain assigned to it, it will retain that domain.

https://github.com/user-attachments/assets/5c616dca-184c-44ec-9be4-f1afb1be4022

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OSLineageGraphBuilder.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OSLineageGraphBuilder.java
@@ -254,7 +254,7 @@ public class OSLineageGraphBuilder {
             NodeInformation nodeInformation = result.getNodes().get(fqnFromHash);
             nodeInformation.setPaging(
                 new LayerPaging().withEntityDownstreamCount((int) bucket.getDocCount()));
-            result.getNodes().put(bucket.getKeyAsString(), nodeInformation);
+            result.getNodes().put(fqnFromHash, nodeInformation);
           }
         }
 


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -780,39 +779,52 @@ private void closeApprovalTask(GlossaryTerm entity, String comment) {
...
+    // Collect all FQN hashes for the glossary term and its children
     Set<String> targetFQNHashesFromDb =
         new HashSet<>(
             daoCollection.tagUsageDAO().getTargetFQNHashForTag(updated.getFullyQualifiedName()));
-
-    List<GlossaryTerm> childTerms =
-        getNestedTerms(updated); // get old value of children term from database
     for (GlossaryTerm child : childTerms) {
-      targetFQNHashesFromDb.addAll( // for each child term find the targetFQNHashes of assets
+      targetFQNHashesFromDb.addAll(
```
sonika-shah: here instead of going via loop we can make one DB call to fetch all targetFqnHashes using the common FQN prefix. All child terms share the same parent prefix, so this we can do in single db call which will be faster and cleaner.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -780,39 +779,52 @@ private void closeApprovalTask(GlossaryTerm entity, String comment) {
...
+
+    // Build a mapping of oldFqn -> newFqn for all terms (including children)
+    Map<String, GlossaryTerm> oldFqnToUpdatedTerm = new HashMap<>();
+    for (GlossaryTerm term : allTerms) {
+      String newFqn = term.getFullyQualifiedName();
+      // Reconstruct the old FQN by replacing the new parent FQN prefix with the old parent FQN
+      // prefix
+      String oldFqn =
+          newFqn.replaceFirst(
+              java.util.regex.Pattern.quote(newParentFqn),
```
sonika-shah: This replacement should work with all possible cases of naming glossary terms, we allow glossaryTerms to have (.) character in name .That should also work here

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -780,39 +779,52 @@ private void closeApprovalTask(GlossaryTerm entity, String comment) {
...
+    for (Map.Entry<String, GlossaryTerm> entry : oldFqnToUpdatedTerm.entrySet()) {
+      String oldFqn = entry.getKey();
+      GlossaryTerm term = entry.getValue();
+      String newFqn = term.getFullyQualifiedName();
+      String newDisplayName = term.getDisplayName();
+      String newName = term.getName();
+      String newDescription = term.getDescription();
+      Map<String, EntityReference> usage =
+          getGlossaryUsageFromES(oldFqn, targetFQNHashesFromDb.size(), false);
+      for (EntityReference ref : usage.values()) {
```
sonika-shah: Again here we can avoid looping through each term , and use common fqnprefix for updating

yan-3005: Fixed it with a bulk update doc, instead of iterations

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
yan-3005: Works

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
yan-3005: This is not needed now

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -1081,6 +1054,24 @@ public void entitySpecificUpdate(boolean consolidatingChanges) {
...
     }
 
+    /**
+     * Move a glossary term to a new parent or glossary. Only parent or glossary can be changed.
+     */
+    @Transaction
+    public void moveAndStore() {
+      boolean consolidateChanges = consolidateChanges(original, updated, operation);
+      incrementalChange();
+      if (consolidateChanges) {
```
sonika-shah: why are we consolidating and reverting the changes here ?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -1325,4 +1317,38 @@ private void invalidateTerm(UUID termId) {
       }
     }
   }
+
+  /**
+   * Move a glossary term to a new parent or glossary. Only parent or glossary can be changed.
+   */
+  public GlossaryTerm moveGlossaryTerm(UUID id, MoveGlossaryTermRequest moveRequest, String user) {
```
sonika-shah, line 1335: how will the logic work when we make simple patch request to update glossaryTerm name?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
yan-3005: Okay, on further analysis, this is not needed since our operation would be "async" and PUT, and each move operation is a significant change and needs its own audit trail and versioning, removing this, thanks for pointing it out

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
yan-3005: Patch requests are not modified, they would still work the same way, this flow is only for moveGlossaryTerm which is particular for the new API

harshach:
@yan-3005 make sure there are tests for this including playwright tests cc @sonika-shah

yan-3005:
@harshach Hi, yes Unit Test cases would be in the same PR, right now, I implemented it for sync, I will make it async and include tests before merging to main. I will also inform for playwright tests.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -1325,4 +1326,37 @@ private void invalidateTerm(UUID termId) {
...
+
+    EntityReference parent = moveRequest.getParent();
+    if (parent.getType().equalsIgnoreCase("glossary")) {
+      // Move to root of the glossary
+      Glossary glossary = Entity.getEntity(GLOSSARY, parent.getId(), "*", Include.ALL);
+      updated.setParent(null);
+      updated.setGlossary(glossary.getEntityReference());
+    } else if (parent.getType().equalsIgnoreCase("glossaryTerm")) {
+      // Can be of same glossary or a different glossary
+      GlossaryTerm parentTerm = Entity.getEntity(GLOSSARY_TERM, parent.getId(), "*", Include.ALL);
```
sonika-shah: why are we getting all the fields of glossaryTerm , when we pass  *  to Entity.getEntity() it brings all the fields associated with the term , which is not required
 
 we should fetch only the required fields

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -1325,4 +1326,37 @@ private void invalidateTerm(UUID termId) {
...
+      updated.setParent(parentTerm.getEntityReference());
+      Glossary glossary =
+          Entity.getEntity(GLOSSARY, parentTerm.getGlossary().getId(), "*", Include.ALL);
```
sonika-shah: same here we dont need to fetch all * fields

openmetadata-service/src/main/java/org/openmetadata/service/util/WebsocketNotificationHandler.java
```diff
@@ -270,4 +270,38 @@ public static void sendDeleteOperationFailedNotification(
...
+    UUID userId = getUserIdFromSecurityContext(securityContext);
+    LOG.error(
+        "[AsyncMove] Move operation failed - jobId: {}, userId:{} ,entity: {}, error: {}",
+        jobId,
+        userId,
+        entity.getName(),
+        error);
+    if (userId != null) {
+      WebSocketManager.getInstance()
+          .sendToOne(userId, WebSocketManager.MOVE_ENTITY_CHANNEL, jsonMessage);
```
sonika-shah: can we have a better name, specific to the glossaryTerm , MOVE_ENTITY_CHANNEL seems generic

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
yan-3005: Done

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
yan-3005: Sure, done

openmetadata-service/src/main/java/org/openmetadata/service/util/WebsocketNotificationHandler.java
yan-3005: Changed to move_glossary_Term_channel

openmetadata-ui/src/main/resources/ui/src/components/Modals/ChangeParentHierarchy/ChangeParentHierarchy.component.tsx
```diff
@@ -11,79 +11,161 @@
...
+        socket.off(SOCKET_EVENTS.MOVE_GLOSSARY_TERM_CHANNEL);
+      }
+    };
+  }, [socket, handleMoveJobUpdate]);
 
   return (
     <Modal
       open
       cancelText={t('label.cancel')}
+      closable={false}
```
chirag-madlani, line 166: Provide onClose so that it can close on Esc

sonika-shah (dismissed):

yan-3005:
@harshach @mohityadav766 Please review

openmetadata-service/src/main/java/org/openmetadata/csv/EntityCsv.java
```diff
@@ -916,9 +917,16 @@ protected void createEntity(
     if (Boolean.FALSE.equals(importResult.getDryRun())) {
       try {
         // In case of updating entity , prepareInternal as update=True
-        repository.prepareInternal(
-            entity,
-            repository.findByNameOrNull(entity.getFullyQualifiedName(), Include.ALL) != null);
+        boolean update =
+            repository.findByNameOrNull(entity.getFullyQualifiedName(), Include.ALL) != null
```
mohityadav766: move this code to entity utility, create a function that return a boolean, you pass the id and type and it can tell whether that entity exists other than not

openmetadata-service/src/main/java/org/openmetadata/csv/EntityCsv.java
```diff
@@ -936,6 +944,19 @@ protected void createEntity(
           repository.findByNameOrNull(entity.getFullyQualifiedName(), Include.NON_DELETED) == null
               ? Response.Status.CREATED
               : Response.Status.OK;
+      // Glossary term's parent can change causing fqn to change, so we need to fetch by name to
+      // ensure
+      if (responseStatus.equals(Response.Status.CREATED) && entity instanceof GlossaryTerm) {
```
mohityadav766: reuse same function i meantioned above

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -1113,6 +1114,22 @@ public final PutResponse<T> createOrUpdate(UriInfo uriInfo, T updated, String up
   public final PutResponse<T> createOrUpdateForImport(
       UriInfo uriInfo, T updated, String updatedBy) {
     T original = findByNameOrNull(updated.getFullyQualifiedName(), ALL);
+    // Glossary term's parent can change causing fqn to change, so we need to fetch by name to
+    // ensure
+    if (original == null && updated instanceof GlossaryTerm) {
```
mohityadav766: this code applies to applies to all imports we should not add isntanceOf here, rather override this for specific reposutory and do use or change in there

openmetadata-service/src/main/java/org/openmetadata/service/util/MoveGlossaryTermResponse.java
```diff
@@ -0,0 +1,23 @@
+package org.openmetadata.service.util;
+
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@NoArgsConstructor
```
mohityadav766, line 7: can you confirm there is some background job schema that cannot be used here

openmetadata-service/src/main/java/org/openmetadata/csv/EntityCsv.java
yan-3005: @mohityadav766 This was an already existing code, It exists in EntityRepository and used from there.Refactored the code to include instance of glossary term which i will remove and place it in the corresponding layer

openmetadata-service/src/main/java/org/openmetadata/service/util/MoveGlossaryTermResponse.java
yan-3005: @mohityadav766 No, there are other messages similar to this class for Delete and Import, but the entityName field is not there in Import, and Delete has extra booleans which is not maintainable, This is clean and maintainable, we can also extend it in future for other entities if any.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
yan-3005: Addressed

openmetadata-service/src/main/java/org/openmetadata/csv/EntityCsv.java
yan-3005: Addressed

openmetadata-service/src/main/java/org/openmetadata/csv/EntityCsv.java
yan-3005: Addressed

yan-3005:
Addressed review comments

mohityadav766 (approved):

PR #21874, opened 2025-06-20
feat: Use Docker volumes for local DB persistence

mgorsk1:
### Describe your changes:

Fixes <issue-number>

I worked on refactoring local docker compose files to use docker volumes for db services because it's more convenient to rely on docker volumes than mounting paths. It improves the overall developer experience and stability.

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/docker/development/docker-compose-gcp.yml
+++ b/docker/development/docker-compose-gcp.yml
@@ -19,6 +19,7 @@ volumes:
   ingestion-volume-dags:
   ingestion-volume-tmp:
   es-data:
+  mysql-data:
 services:
   mysql:
     build:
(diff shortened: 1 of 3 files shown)


harshach:
This will break existing users who are using the volumes as set previously. Any reason for this change? @mgorsk1

mgorsk1:
Indeed, however I assumed these docker compose files are used primarily for development so it shouldn't be much of a problem. 

Reasons for a change that I consider important:

- Improved permission management: Mounting paths directly can lead to issues where folders are created with root permissions. This may result in permission-related problems, especially when cleanup or modification is required by users without elevated privileges (which is also my case 😅).
- Consistency and ease of management: All other mounts are currently handled via Docker volumes. Aligning with this approach offers a more consistent and user-friendly experience, simplifying maintenance and reducing potential configuration friction. **Currently there is one mounting point that wouldn't be visible in docket volume list and cannot be managed through this API.**

When developing OM I always locally refactor these to use volumes and wanted to make this change persistent in the code base.

TeddyCr:
Thanks @mgorsk1 for the contribution. We really appreciate it, however at this time we are going to close this PR based on @harshach comment.

PR #21909, opened 2025-06-23
Fix #21908: Add Persona configuration per User

### LINKED ISSUE #21908: Add Persona configuration per User
**Feature**
Allow users to extend persona derived landing page to make their own

**Describe the task**
Currently, admin can set a landing page persona. We are improving the landing page to make it more usable for business users, non technical users alike. To improve this we are adding new widgets and layout. 
In this landing page layout
1. User should be able to set top bar background panel
2. In the future user should be able override the widgets that admin set for the persona
3. As part of this task we want to override the header background per persona that user has

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-spec/src/main/resources/json/schema/entity/teams/user.json
```diff
@@ -155,6 +155,14 @@
     "lastLoginTime": {
       "description": "Last time the user logged in.",
       "$ref": "../../type/basic.json#/definitions/timestamp"
+    },
```
mohityadav766: IMO we should use some other table for user preferences .
We have "personaPreference" as one type , bad example but maybe appearance and color choice can be as per user choice.

mohityadav766: we can even have search preferences at user level

mohityadav766: just want this to be in a way where it's flexible enought to accomodate other type of per user preferences

harshach: @mohityadav766 you want to store preferences in another table?

mohityadav766: yes

mohityadav766 (dismissed):

PR #21973, opened 2025-06-26
Fix #21569 : Pagination state lost after multiple back navigations in table list view

### LINKED ISSUE #21569: Pagination state lost after multiple back navigations in table list view
**Affected module**
 UI

**Describe the bug**

I noticed a bug related to navigation history. As shown in the screenshot, if I’m on page 1 and then navigate to page 2 or 3 to browse tables (e.g., I select table 1, then go back to select another), when I hit "back" again, it returns me to page 1 instead of the page I was previously browsing.

**To Reproduce**

Go to the Table list in OpenMetadata

Navigate to Page 2 or 3

Click to view any table

Use the browser back button → you return to the correct page

Click on another table

Use back again → you're unexpectedly returned to Page 1

https://github.com/user-attachments/assets/5e81a1ca-daaa-43ca-be7b-8ad8c956239d

**Expected behavior**

The system should return to the exact page I was previously browsing, even after navigating multiple times. The pagination state should be retained consistently when using browser back navigation

**Version:**
 - Python version:
 - OpenMetadata version: 1.7.1
 - OpenMetadata Ingestion package version: 1.7.1.1

**Additional context**

shrushti2000:
#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/pages/DatabaseSchemaPage/SchemaTablesTab.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/pages/DatabaseSchemaPage/SchemaTablesTab.tsx
@@ -135,7 +135,10 @@ function SchemaTablesTab({
 
   const handleShowDeletedTables = (value: boolean) => {
     setFilters({ showDeletedTables: value });
-    handlePageChange(INITIAL_PAGING_VALUE);
+    handlePageChange(INITIAL_PAGING_VALUE, {
+      cursorType: null,
+      cursorValue: undefined,
+    });
   };
 
   const getSchemaTables = useCallback(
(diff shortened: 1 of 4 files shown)


openmetadata-ui/src/main/resources/ui/src/pages/DatabaseSchemaPage/SchemaTablesTab.tsx
```diff
@@ -174,7 +177,6 @@ function SchemaTablesTab({
           pageSize
         );
       }
-      handlePageChange(currentPage);
```
chirag-madlani, line 177: What's the reason for removing page changes? Handle by useEffect?

shrushti2000: `handlePageChange(currentPage);` its removed from the tablePaginationHandler and not useeffect.
 `handlePageChange(currentPage)`; line at the end of the tablePaginationHandler function was redundant.
We are already updating the cursorData in tablePaginationHandler once, it was unnecessarily updating it twice.

PR #21976, opened 2025-06-26
fix(ui): Update bg color for the tab components

sweta1308:
**Before:** 

https://github.com/user-attachments/assets/ca297249-e939-499d-81cc-acd23443621c

**After:** 

https://github.com/user-attachments/assets/ad3226c8-3ce8-4b02-96ef-d3e78adc56f9

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Customization/GenericTab/generic-tab.less
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Customization/GenericTab/generic-tab.less
@@ -13,7 +13,7 @@
 @import (reference) url('antd/dist/antd.variable.less');
 @import (reference) url('../../../styles/variables.less');
 
-:not(.edit-dashboard-page-content) .grid-container {
+.grid-container {
   position: relative;
   height: auto !important;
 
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/src/styles/components/tabs.less
```diff
@@ -108,3 +108,11 @@
...
   }
 }
+
+.entity-details-page-tabs {
+  .ant-tabs.tabs-new {
+    .ant-tabs-tabpane .data-quality-content-panel {
+      background-color: @white;
+    }
+  }
+}
```
chirag-madlani: I think we can have this for data-qulity less file and not at global level

sweta1308: removed this from here

openmetadata-ui/src/main/resources/ui/src/components/Database/SampleDataTable/SampleDataTable.component.tsx
```diff
@@ -214,7 +214,7 @@ const SampleDataTable = ({
 
   return (
     <div
-      className={classNames('p-md border-default border-radius-sm', {
+      className={classNames('p-md border-default border-radius-sm bg-white', {
```
chirag-madlani, line 217: Why we are passing this to both like on tsx and on less file too

sweta1308: The less file one is for the error placeholder and the tsx one is for the table component. And both are rendered separately, like error placeholder one is inside an if statement outside of main page return statement and main table is in the return.

PR #21983, opened 2025-06-26
MINOR: Update Trino Connection to fix data diff

IceS2:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/data_quality/validations/runtime_param_setter/base_diff_params_setter.py
```diff
@@ -1,15 +1,50 @@
...
+            "metadata.{}.source.{}.{}.{}.ServiceSpec".format(  # pylint: disable=C0209
+                "ingestion",
+                self.service_type.name.lower(),
+                get_module_dir(self.source_type),
+                "service_spec",
+            )
+        )
+
+    def get_data_diff_class(self) -> Type["BaseTableParameter"]:
+        return import_from_module(self.service_spec.data_diff)
```
TeddyCr, line 44: Do you know if that would be handle automatically in Collate? I believe we have this class there as well https://github.com/open-metadata/openmetadata-collate/blob/main/ingestion-extension/src/metadata/data_quality/validations/table/sqlalchemy/collate/tableDiff.py

ingestion/src/metadata/ingestion/connections/connection.py
```diff
@@ -41,12 +41,23 @@ class BaseConnection(ABC, Generic[S, C]):
...
 
     service_connection: S
+    _client: Optional[C]
 
     def __init__(self, service_connection: S) -> None:
         self.service_connection = service_connection
+        self._client = None
+
+    @property
+    def client(self) -> C:
```
TeddyCr, line 51: nit: should that be Optional[C] (since `get_client` is not implemented by default?

ingestion/src/metadata/data_quality/validations/runtime_param_setter/base_diff_params_setter.py
IceS2: I'm not sure. I think we're importing the one from OSS always. I didn't change this logic. I just moved it around:

```def get_for_source(
    service_type: ServiceType, source_type: str, from_: str = "ingestion"
):
    return import_from_module(
        "metadata.{}.source.{}.{}.{}.ServiceSpec".format(  # pylint: disable=C0209
            from_,
            service_type.name.lower(),
            get_module_dir(source_type),
            "service_spec",
        )
    )


class TableDiffParamsSetter(RuntimeParameterSetter):
	@@ -62,23 +49,16 @@ def __init__(self, *args, **kwargs):
        }

    def get_parameters(self, test_case) -> TableDiffRuntimeParameters:
        # Using the specs class method causes circular import as TestSuiteInterface
        # imports RuntimeParameterSetter
        cls_path = get_for_source(
            ServiceType.Database,
            source_type=self.service_connection_config.type.value.lower(),
        ).data_diff
        cls = import_from_module(cls_path)()```

ingestion/src/metadata/ingestion/connections/connection.py
IceS2: I think a client is a mus though, no? `_get_client` is abstract but It returns C

TeddyCr (dismissed):

PR #22068, opened 2025-07-01
Removing external app private config from ingestion pipeline

aji-aju:
### Describe your changes:

Fixes:

- mask collate ai app private_config in app runner
- use secret manager to store/retrieve secret values of private_config
- relevant Collate PR: https://github.com/open-metadata/openmetadata-collate/pull/2275

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/tests/unit/application/test_application_private_config.py
+++ b/ingestion/tests/unit/application/test_application_private_config.py
@@ -14,7 +14,7 @@ Unit tests for Application private config retrieval from secrets manager
 """
 import json
 from unittest import TestCase
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock
 
 from metadata.generated.schema.entity.services.connections.metadata.openMetadataConnection import (
     OpenMetadataConnection,
@@ -23,8 +23,8 @@ from metadata.generated.schema.metadataIngestion.application import (
     OpenMetadataApplicationConfig,
 )
 from metadata.generated.schema.metadataIngestion.workflow import WorkflowConfig
+from metadata.ingestion.models.custom_pydantic import _CustomSecretStr
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
-from metadata.utils.secrets.secrets_manager_factory import SecretsManagerFactory
 from metadata.workflow.application import AppRunner
 
 MOCK_PIPELINE_FQN = "OpenMetadata.CollateAIApplication"
@@ -40,10 +40,10 @@ class MockAppRunner(AppRunner):
     ):
         self.app_config = config.appConfig.root if config.appConfig else None
         self.metadata = metadata
-        # Use ingestionPipelineFQN instead of applicationFqn
+        # Use applicationFqn from config
         self.private_config = self._retrieve_app_private_config(
             config.appPrivateConfig.root if config.appPrivateConfig else None,
-            config.ingestionPipelineFQN,
+            config.applicationFqn.root if config.applicationFqn else None,
         )
 
     def run(self) -> None:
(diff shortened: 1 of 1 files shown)


mohityadav766 (dismissed):
lgtm

mohityadav766:
@aji-aju can you check this is not impacting other apps

ingestion/src/metadata/workflow/application.py
```diff
@@ -22,13 +26,20 @@
...
 from metadata.utils.logger import ingestion_logger
+from metadata.utils.secrets.db_secrets_manager import DBSecretsManager
+from metadata.utils.secrets.secrets_manager_factory import SecretsManagerFactory
 from metadata.workflow.base import BaseWorkflow
 
 logger = ingestion_logger()
 
+FERNET_KEY = "FERNET_KEY"
+# default ferney key provided in openmetadata.yaml
+DEFAULT_FERNET_KEY = "jJ/9sz0g0OHxsfxOoSfdFdmk3ysNmPRnH3TUAbz3IHA="
```
pmbrull: this should not be here. client should not care about the key

aji-aju:
> @aji-aju can you check this is not impacting other apps

Hi @mohityadav766 , Verified by running Search Indexing App, Data Retention, Collate AI Tier Agent. No impact noticed.

pmbrull (changes requested):
lets apply the discussed changes using the bot reading logic

openmetadata-service/src/main/java/org/openmetadata/service/apps/AbstractNativeApplication.java
```diff
@@ -205,6 +210,20 @@ private void createAndBindIngestionPipeline(
...
 
+    // Get encrypted private config
+    Object privateConfig = this.getApp().getPrivateConfiguration();
+    if (privateConfig != null) {
+      String privateConfigJson = JsonUtils.pojoToJson(privateConfig);
+      SecretsManager secretsManager = SecretsManagerFactory.getSecretsManager();
+      if (secretsManager instanceof ExternalSecretsManager) {
+        String baseSecretId =
+            "external-app-" + this.getApp().getName().toLowerCase() + "-private-config";
+        privateConfig = "secret:" + baseSecretId;
```
pmbrull: should we have such important keys in here? `secret:` is a constant already in the SecretsManager class. I feel a bit uneasy with the `external-app-` prefix handling directly here. Not sure if we could centralize this logic on the SecretsMaanger itself vs. having it scattered

ingestion/src/metadata/workflow/application.py
```diff
@@ -46,13 +50,105 @@ def __init__(
...
+            ingestion_pipeline_fqn: Fully qualified name of the ingestion pipeline (e.g., "OpenMetadata.AppName")
+
+        Returns:
+            Dictionary containing the private config or None if not found
+        """
+        try:
+            if isinstance(private_config, dict):
+                logger.debug("Private config is already a dictionary")
+                return private_config
+            # Extract app name from ingestion pipeline FQN
```
pmbrull: should we have an AppName in the appConfig instead?

ingestion/src/metadata/workflow/application.py
```diff
@@ -46,13 +50,105 @@ def __init__(
...
+                    parsed_private_config = self._mask_private_config_token(
+                        fetched_private_config
+                    )
+                    logger.debug(
+                        "Successfully retrieved app private config from API using bot token"
+                    )
+                    return parsed_private_config
+                elif isinstance(
+                    fetched_private_config, str
+                ) and fetched_private_config.startswith("secret:"):
```
pmbrull: why are we doing this here if our custom pydantic class already reads the secrets properly?

openmetadata-service/src/main/java/org/openmetadata/service/apps/ApplicationHandler.java
```diff
@@ -338,4 +379,71 @@ public void removeOldJobs(App app) throws SchedulerException {
...
+          e.getMessage());
+      LOG.warn(
+          "External app {} will load private config from resource files instead of Secrets Manager",
+          app.getName());
+    }
+  }
+
+  private void updateIngestionPipelinePrivateConfig(App app) {
+    try {
+      String fqn = FullyQualifiedName.add("OpenMetadata", app.getName());
```
pmbrull, line 426: why are we magically adding FQN here?

ingestion/src/metadata/workflow/application.py
harshsoni2024: Initially added that as a safety check since we are checking inside secret manager after that.
There can be only two possible cases where private_config (from api) comes as dictionary or string with "secret:" prefix.

openmetadata-service/src/main/java/org/openmetadata/service/apps/ApplicationHandler.java
aji-aju: This is the currently followed pattern for external app FQN.

Reference: https://github.com/open-metadata/OpenMetadata/blob/1bb9cdc4ab9436a6d0c0b737f33a930f601d2e72/openmetadata-service/src/main/java/org/openmetadata/service/apps/AbstractNativeApplication.java#L58C31-L58C43


https://github.com/open-metadata/OpenMetadata/blob/1bb9cdc4ab9436a6d0c0b737f33a930f601d2e72/openmetadata-service/src/main/java/org/openmetadata/service/apps/AbstractNativeApplication.java#L162

PR #22155, opened 2025-07-04
MINOR: Improve Array Sampler for UC & DBX

ulixius9:
### Describe your changes:

The unity catalog and databricks connector fails with OOM when a table has an array field which may contain huge data, with this pr we are only getting max 10 items from array instead of fetching all the items form it

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/sampler/sqlalchemy/sampler.py
+++ b/ingestion/src/metadata/sampler/sqlalchemy/sampler.py
@@ -262,7 +262,7 @@ class SQASampler(SamplerInterface, SQAInterfaceMixin):
                     if self._handle_array_column(col):
                         slice_expression = self._get_slice_expression(col)
                         select_columns.append(slice_expression)
-                        logger.info(
+                        logger.debug(
                             f"Limiting array column {col.name} to {max_elements} elements to prevent OOM"
                         )
                     else:
--- a/ingestion/src/metadata/sampler/sqlalchemy/sampler.py
+++ b/ingestion/src/metadata/sampler/sqlalchemy/sampler.py
@@ -248,45 +248,34 @@ class SQASampler(SamplerInterface, SQAInterfaceMixin):
             ]
 
         with self.get_client_context() as client:
-            # Check if we have array columns that need special handling
-            array_columns = [
-                col for col in sqa_columns if self._handle_array_column(col)
-            ]
 
-            if array_columns:
-                # Handle array columns with special query modification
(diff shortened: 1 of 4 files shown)


ingestion/src/metadata/sampler/sqlalchemy/sampler.py
```diff
@@ -215,16 +248,60 @@ def fetch_sample_data(self, columns: Optional[List[Column]] = None) -> TableData
             ]
 
         with self.get_client_context() as client:
```
TeddyCr, line 250: Can you move it out of the base sampler and put it in DBX and UC sampler. You can extract the sql into its own and override it in DBX and UC sampler. This way we don't mix specific code with the generic implementation

pmbrull (approved):

ingestion/src/metadata/sampler/sqlalchemy/sampler.py
ulixius9: I did this in base sampler because I think we will face the same problems for other connectors as well for same situation in such case we just have to write the logic for implementing the slicing

should I still move it to UC & DBX and we can later decide how we want to handle for other connectors?

TeddyCr: The problem here is that _base_sample_query is used not just for fetching sample data but it is the base query used when we'll run the profiler / DQ is sampled. There are few things that will have an impatc on the profiler / DQ overall:
1. We are iterating over the all the columns of the table twice -- it is not uncommon to have tables with a large amount of columns
2. If a service has an array but does not implement `_get_slice_expression` we iterate over all the column for nothing 
3. This will go through all the rows, that will become a bottleneck
```python
        if array_columns:
             for row in sqa_sample:
                 processed_row = []
                 for i, col in enumerate(sqa_columns):
                     value = row[i]
                     if self._handle_array_column(col):
                         processed_value = self._process_array_value(value)
                         processed_row.append(processed_value)
                     else:
                         processed_row.append(value)
                 processed_rows.append(processed_row)
         else:
             processed_rows = [list(row) for row in sqa_sample]
```

TeddyCr (changes requested):
@ulixius9 there a few points we need to address, otheriwse we'll be introducing new issues.

PR #22201, opened 2025-07-07
[FIX] 21910: Update amundsen ingest - remove extra Either

### LINKED ISSUE #21910: Multiple errors when trying to ingest metadata from Amundsen during a migration
**Affected module**
Ingestion Framework

**Describe the bug**

Issues with the ingestion from Amundsen (in order to migrate from one catalog to the other):

1. There is a typo in the snowflake config in:
	https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/utils/metadata_service_helper.py. 
	It should be `service_name` and not `service_nmae`
2. Seems like the error I'm seeing during the Amundsen ingestion is coming from the [amundsen metadata](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/metadata/amundsen/metadata.py) module.
	- For _yield_create_database and _yield_create_database_schema to work, the elements that come from the Amundsen query need to have a matching case. 
	- In my Amundsen data, everything is lowercase, and the generation of these yields came empty. (And the logging when this happens is not clear) (AttributeError: 'tuple' object has no attribute 'left')
3. For those _yield methods mentioned before, there is a "naming matching" that needs to happen:
	- table["database"] from the Amundsen data has to be the "Database Service Name" in OpenMetadata.
	- table["cluster"] from the Amundsen data has to be the Database Name in OpenMetadata, but the names of the variables in _yield_create_database makes it "not that easy" to follow.
4. Even after manually adjusting everything mentioned before, It seems like something is happening when calling self._iter() in /home/airflow/.local/lib/python3.10/site-packages/metadata/ingestion/api/step.py because the "result" object is still this ('left',None) tuple.

**To Reproduce**

Trying to ingest metadata from Amundsen using Openmetadata v1.7.2.

Image

**Expected behavior**

- Being able to ingest metadata from Amundsen.
- Or, if something fails, have a better understanding of where the error is coming from, since the logging "AttributeError: 'tuple' object has no attribute 'left'" is not really showing where the error could be coming from.

**Version:**
 - Python version: 3.10
 - OpenMetadata version: docker.getcollate.io/openmetadata/server:1.7.2
 - OpenMetadata Ingestion package version: docker.getcollate.io/openmetadata/ingestion:1.7.2

mikekutzma:
Addresses: #21910

Removed 'Either' that currently wraps the generators being yieded from in the amundsen `metadata.py` from the `AmundsenSource` class. I think these are not necessary and causing the issue since the objects being yielded inside those methods (`self.create_table_entity`, etc) are all yielding `Either` objects.

Tested on my local instance and this resolved the issue with the service not being able to ingest.

### Describe your changes:

Fixes #21910

#
### Type of change:

- [x] Bug fix

#
### Checklist:

### DIFF
--- a/bin/openmetadata-server-start.sh
+++ b/bin/openmetadata-server-start.sh
@@ -97,15 +97,23 @@ if [ "x$OPENMETADATA_DEBUG" != "x" ]; then
 fi
 
 # GC options for Java 21
-GC_LOG_FILE_NAME='openmetadata-gc.log'
-if [ -z "$OPENMETADATA_GC_LOG_OPTS" ]; then
-  OPENMETADATA_GC_LOG_OPTS="-Xlog:gc*,gc+heap=info,gc+ergo=debug:file=$LOG_DIR/$GC_LOG_FILE_NAME:time,level,tags:filecount=10,filesize=102400"
-fi
+OPENMETADATA_GC_LOG_OPTS="-Xlog:gc*,gc+heap=info,gc+ergo=debug:file=${LOG_DIR:-/var/log/openmetadata}/openmetadata-gc.log:time,level,tags:filecount=10,filesize=102m"
 
+NOW_EPOCH=$(date +%s)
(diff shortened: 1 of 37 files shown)


harshach:
thanks for the contribution @mikekutzma

ulixius9 (approved):

PR #22212, opened 2025-07-08
Fix(UI): Current User store crash 

dhruvjsx:
Application was getting crashed because old user didn't had `language` key on their local storage which was introduced in latest changes. Added  its fix for any missing key we'll spread the default values.

https://github.com/user-attachments/assets/3e85a4c7-57b2-44ea-81c9-62e59e06e25a

### Describe your changes:

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/NavBar/NavBar.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/NavBar/NavBar.tsx
@@ -511,7 +511,8 @@ const NavBar = () => {
               <Button
                 className="flex-center gap-2 p-x-xs font-medium"
                 type="text">
-                {upperCase(language.split('-')[0])} <DropDownIcon width={12} />
+                {language ? upperCase(language.split('-')[0]) : ''}{' '}
+                <DropDownIcon width={12} />
               </Button>
             </Dropdown>
             <Dropdown
--- a/openmetadata-ui/src/main/resources/ui/src/hooks/currentUserStore/useCurrentUserStore.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/hooks/currentUserStore/useCurrentUserStore.ts
@@ -98,7 +98,9 @@ export const useCurrentUserPreferences = () => {
   }
 
   return {
-    preferences: preferences[currentUser.name] || defaultPreferences,
+    preferences: preferences[currentUser.name]
+      ? { ...defaultPreferences, ...preferences[currentUser.name] }
+      : defaultPreferences,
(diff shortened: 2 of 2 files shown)


openmetadata-ui/src/main/resources/ui/src/components/NavBar/NavBar.tsx
```diff
@@ -511,7 +511,8 @@ const NavBar = () => {
               <Button
                 className="flex-center gap-2 p-x-xs font-medium"
                 type="text">
-                {upperCase(language.split('-')[0])} <DropDownIcon width={12} />
+                {language ? upperCase(language.split('-')[0]) : ''}{' '}
```
aniketkatkar97, line 514: why do we need last `{' '}`?

openmetadata-ui/src/main/resources/ui/src/hooks/currentUserStore/useCurrentUserStore.ts
```diff
@@ -98,7 +98,9 @@ export const useCurrentUserPreferences = () => {
   }
 
   return {
-    preferences: preferences[currentUser.name] || defaultPreferences,
+    preferences: preferences[currentUser.name]
+      ? { ...defaultPreferences, ...preferences[currentUser.name] }
+      : defaultPreferences,
```
aniketkatkar97, line 103: what's the reason for this change?

openmetadata-ui/src/main/resources/ui/src/components/NavBar/NavBar.tsx
dhruvjsx: It was automatically added while committing will remove it

openmetadata-ui/src/main/resources/ui/src/hooks/currentUserStore/useCurrentUserStore.ts
dhruvjsx: We introduced language key in currentUserStore but all users who already had previously stored value in their local storage did not had language key. Which was causing the crash. 

Now we make sure that if any of the key is missing we are sending default value for that particular key

chirag-madlani: @dhruvjsx for your reference you may use `get` function from lodash for handling fallback cases

chirag-madlani (approved):

PR #22278, opened 2025-07-10
Add domain filter in search request

vvavepacket:
### Describe your changes:

Fixes 22276

I worked on issue 22276 because I want users with Domain Access Only Role should only see entities that are belonging to their assigned domains. User should not see entities belonging to other domains.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
@@ -366,6 +366,13 @@ public class ElasticSearchClient implements SearchClient {
   public Response doSearch(
       SearchRequest request, SubjectContext subjectContext, SearchSettings searchSettings)
       throws IOException {
+    // Override with domain query
+    if (request.getDomains().stream().findFirst().isPresent()) {
+      String domain = request.getDomains().get(0).getFullyQualifiedName();
+      String newQuery = request.getQuery() + " AND (domain.fullyQualifiedName:\""+domain+"\")";
+      request.setQuery(newQuery);
+    }
+
     String indexName = Entity.getSearchRepository().getIndexNameWithoutAlias(request.getIndex());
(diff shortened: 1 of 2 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -366,6 +366,13 @@ public Response previewSearch(
   public Response doSearch(
       SearchRequest request, SubjectContext subjectContext, SearchSettings searchSettings)
       throws IOException {
+    // Override with domain query
+    if (request.getDomains().stream().findFirst().isPresent()) {
```
harshach, line 370: you can do request.getDomains().getFirst().isPresent()?

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -366,6 +366,13 @@ public Response previewSearch(
...
+      String domain = request.getDomains().get(0).getFullyQualifiedName();
```
harshach, line 371: same here request.getDomains().get(0)

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -366,6 +366,13 @@ public Response previewSearch(
...
+      String newQuery = request.getQuery() + " AND (domain.fullyQualifiedName:\""+domain+"\")";
```
harshach, line 372: we need to stop thinking that we will be using queryString, currently queryString being deprecated in our searches we already use multimatch, please check that and adjust accodingly don't want to pass this in this format

mohityadav766:
this should be fixed, enableSearchRbac will do this 
Anyone part of a domain will only see domains assets

PR #22291, opened 2025-07-11
Fix #22990: Add ETag Support for Optimistic Concurrency Control in PATCH Operations

### LINKED ISSUE #22290: Add ETag Support for Optimistic Concurrency Control in PATCH Operations
**Feature**
 Currently, OpenMetadata lacks optimistic concurrency control for entity updates, which can lead to silent data loss when multiple users or processes update the same entity
  concurrently. This issue proposes adding ETag (Entity Tag) support to prevent concurrent update conflicts.

**Describe the task**

  When two users attempt to update the same entity simultaneously:
  - Both read version 0.4 of an entity
  - User A adds a description to a column
  - User B adds tags to the entity
  - Both submit their changes
  - The last write wins, causing the first user's changes to be silently lost

  This is particularly problematic in environments with:
  - Multiple users collaborating on metadata
  - Automated ingestion pipelines running concurrently
  - Bulk update operations

**Proposed Solution**

  Implement HTTP ETag headers for optimistic concurrency control:

  1. GET responses include an ETag header with a hash of the entity's current state
  2. PATCH requests can include an If-Match header with the ETag value
  3. If the entity has been modified since the ETag was generated, return 412 Precondition Failed
  4. Clients can retry with the latest version of the entity

**Implementation Details**

  1. ETag Generation
   - Generate ETag from entity version and updatedAt timestamp
  - Format: "{hash(entityId + version + updatedAt)}"
  - Add compression suffix handling (e.g., --gzip)

  2. Request Flow

  GET /api/v1/tables/{id}
  Response Headers:
  ETag: "0915370b84b4481f"

  PATCH /api/v1/tables/{id}
  Request Headers:
  If-Match: "0915370b84b4481f"

  Response:
    200 OK - If ETag matches
    412 Precondition Failed - If entity was modified

3. Core Components

  New Classes:
  - EntityETag - Utility for ETag generation and validation
  - ETagRequestFilter - JAX-RS filter to extract If-Match headers
  - ETagResponseFilter - JAX-RS filter to add ETag headers
  - PreconditionFailedException - Exception for 412 responses

  Modified Classes:
  - EntityResource - Add ETag support to PATCH endpoints
  - EntityRepository - Add optimistic locking in patch methods
  - EntityDAO - Add version-aware update methods

**Example Client Usage**

 - Get entity with ETag
```
  response = requests.get(f"{base_url}/api/v1/tables/{table_id}")
  etag = response.headers.get('ETag')
  table = response.json()
```

 - Modify entity
 ```
  table['description'] = 'Updated description'
```

 - Update with ETag validation
```
  headers = {'If-Match': etag}
  response = requests.patch(
      f"{base_url}/api/v1/tables/{table_id}",
      headers=headers,
      json=patch_data
  )

  if response.status_code == 412:
      # Entity was modified, retry with latest version
      print("Conflict detected, retrying...")
```
  ---

### PR DESCRIPTION
**Pull request by harshach**
### Describe your changes:

Fixes #22290 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### CODE REVIEW

**Review by pmbrull — DISMISSED** [review]

File: package.json [hunk 1]
```diff
@@ -1,20 +0,0 @@
-{
-  "name": "open-metadata",
```
mohityadav766: is this on purpose?

harshach: why do we have package.json at the root of the project? cc @chirag-madlani @karanh37 any ideas or some left over from merges

chirag-madlani: It was added to the quickType dependencies, which are used for type generation. 
It's added to the root so that it has access to the spec folder without any permission issues

PR #22319, opened 2025-07-12
WIP: Glossary Workflows - State Maintenance

yan-3005:
### Describe your changes:

Introduced State by State Maintenance for Glossary Approval Workflow
**Previous Limitations**:
Since Glossary Terms can be huge in number, storing each state as a record for every glossary term in the workflow instance state table would result in humungous number of records leading to extreme resource consumption.

**Thought Process**:
Instead of storing each stage as a separate record, the approach is to store each stage in the same workflowInstace table by appending or modifying the stages array in the same workflowInstance record. So, in short, one Workflow Instance would have all the stages/states in itself instead of the workflow instance states table which is being used now incase of autopilot application.

**Implementation**:

1. Introduced two extra configs for optimisticLocking retries and maximum states in a workflow instance object
retries -> If two parallel edges try to modify the record, we shouldnt allow the race, instead make a retry of the second edge with the updated version of the workflow instance.
maximum states: limiting to 100, inorder not to overgrow the maximum states to store in a workflow instance making it a very big json, so limiting the states incase of bigger workflows.

2. if **storeState** config is true, it follows the existing logic of storing each record in the table, If storeState is false, then the states are appended to the same workflowInstance document allowing for states to be stored in the same document for glossary approval allowing for Backward Compatibility

3. Optimised the states API to send back the response depending on the config of storeState, if true, follow the old logic, if false, reroute to the workflowInstance repository (the optimised one).

4. Added a migration in v190 (for now) to redeploy the workflow with new configs and new listeners

5. Unit Test cases for the Listener and integration test case for two of the workflow process

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowInstanceStageListener.java
```diff
@@ -52,23 +52,149 @@ private void addNewStage(
...
     String stage =
         Optional.ofNullable(execution.getCurrentActivityId()).orElse(workflowDefinitionName);
-    UUID workflowInstanceStateId =
-        workflowInstanceStateRepository.addNewStageToInstance(
-            stage,
-            workflowInstanceExecutionId,
-            workflowInstanceId,
+
+    // Fetch workflow definition to check config
+    org.openmetadata.service.jdbi3.WorkflowDefinitionRepository workflowDefinitionRepository =
```
harshach: lets import the classes, do not use full reference to classes

yan-3005:
Not needed anymore, following this https://github.com/open-metadata/OpenMetadata/pull/22402

PR #22350, opened 2025-07-14
Username extraction from jwtprincipalclaimsmapping, if email format

aji-aju:
… given

### Describe your changes:

Currently for username extraction, if email is given in jwt principal claims mapping, it is taken as it is. We need to extract just the username part if email is given:
eg: 1234@collate.io -> We need to extract "1234" for username.

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


mohityadav766:
@aji-aju can you add details in the PR for the fix?

harshach:
@mohityadav766 don't we have one Utils class to return the username from email. It looks like this code is being duplicated multiple places

aji-aju:
> @aji-aju can you add details in the PR for the fix?

Hi @mohityadav766  Added description. When we are using username fro jwtclaims mapping, username needs to be extracted out if it is in email format

aji-aju:
@harshach @mohityadav766 , Removed reduntant method

mohityadav766 (approved):

PR #22426, opened 2025-07-17
Fix - Data discrepancy between frontend and SDK extraction for domain api assets field

### LINKED ISSUE #21091: Data discrepancy between frontend and SDK extraction
Hi,

sonika-shah:
### Describe your changes:

Fixes #21091 
- Added new field  allAssets for domain entity to allow listing  current domain assets + its subdomain assets . This is required because in UI for any of the domain asset page, we show assets from the current domain and all its subdomain

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


sonika-shah:
Closing the pr, actual changes present in https://github.com/open-metadata/OpenMetadata/pull/22885

PR #22437, opened 2025-07-17
Fix: #19907 Glossary Term Approval Workflow - Trigger Filters - JsonLogic along with Exclude Fields filter

### LINKED ISSUE #19907: Auto approve glossaryTerm changes if they are done by a reviewer.
A reviewer is often the Data Steward or Data Governance team member.  

Currently, if the change is done by a reviewer, the term goes into  InReview and Approval task is generated.  
Reviewers/Data Stewards often does mass updates (for example add a new tag/change tag) via the upload. This means that the same Steward then gets 100s of approval tasks. 

If change is performed by a reviewer, it should be auto approved. 

Image

Running a separate process via API to approve. Not very effective, almost would be a clean-up work done daily. 

**Additional context**

yan-3005:
### Describe your changes:

No Approval Workflow triggered in case of terms modified by reviewers.

Implementation: 
In the Trigger Filter, add JsonLogic instead of plain exclude/include. This will help in a lot of operations and extended functionality to ignore workflow triggers.

Changed the Glossary Approval Workflow to include filter as a field to indicate the jsonFilters along with exclude fields. Hybrid Logic , By default the filter is empty with {and:[]}
Example exclude and filter
```
"exclude": ["reviewers"],
"filter": "{\"or\":[{\"isReviewer\":{\"var\":\"updatedBy\"}}]}"
```
Added Nodes for auto approval
image

Fixes #19907

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/EventBasedEntityTrigger.java
```diff
@@ -165,18 +168,34 @@ private CallActivity getWorkflowTrigger(String triggerWorkflowId, String mainWor
...
+            new FieldExtensionBuilder()
+                .fieldName("filterExpr")
+                .fieldValue(JsonUtils.pojoToJson(triggerConfig.getFilter()))
+                .build();
+        serviceTask.getFieldExtensions().add(filterExpr);
+      } else if (triggerConfig.getExclude() != null) {
+        // Use legacy exclude path
+        FieldExtension excludedFilterExpr =
+            new FieldExtensionBuilder()
+                .fieldName("excludedFilterExpr")
```
IceS2: Should we have this? Shouldn't we deal with this in migration and transform this into the filter? 🤔

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/impl/FilterEntityImpl.java
```diff
@@ -16,22 +17,40 @@
 import org.openmetadata.schema.utils.JsonUtils;
 import org.openmetadata.service.Entity;
 import org.openmetadata.service.governance.workflows.WorkflowVariableHandler;
+import org.openmetadata.service.governance.workflows.utils.JsonLogicUtils;
 import org.openmetadata.service.resources.feeds.MessageParser;
 
 public class FilterEntityImpl implements JavaDelegate {
   private Expression excludedFilterExpr;
+  private Expression filterExpr;
```
IceS2: Like I mentioned above, I'll just repalce the exclude filter for the filterExpr to be honest

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowEventConsumer.java
```diff
@@ -68,6 +69,13 @@ public void sendMessage(ChangeEvent event) throws EventPublisherException {
             getNamespacedVariableName(GLOBAL_NAMESPACE, RELATED_ENTITY_VARIABLE),
             entityLink.getLinkString());
 
+        // Add updatedBy information from the ChangeEvent
+        if (event.getUserName() != null) {
+          variables.put(
+              getNamespacedVariableName(GLOBAL_NAMESPACE, UPDATED_BY_VARIABLE),
+              event.getUserName());
+        }
+
```
IceS2: Do we need to get this from here? Can't we get it from the EntityLink? Build the entity, fetch the UpdatedBy?

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/utils/JsonLogicUtils.java
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    // Add change-related variables if needed
+    if (requiredVars.contains("changedFields") || requiredVars.contains("isCreateEvent")) {
+      addChangeInfo(rules, entity);
+    }
+
+    LOG.debug("Built rules for variables: {} -> {}", requiredVars, rules.keySet());
+    return rules;
+  }
```
IceS2: Why do we need to extract the variables and build a jsonlogic context? 🤔 

It seems a bit convoluted vs applying it to the Map dierctly, but I've never created a custom operator.

@pmbrull thoughts?

openmetadata-service/src/main/resources/json/data/governance/workflows/GlossaryApprovalWorkflow.json
```diff
@@ -11,7 +11,7 @@
     "config": {
       "entityType": "glossaryTerm",
       "events": ["Created", "Updated"],
-      "exclude": ["reviewers"]
+      "filter": "{\"or\":[{\"var\":\"isReviewer\"},{\"excludeFields\":[\"reviewers\"]}]}"
```
IceS2: Again, haven't created a new JsonLogic operator but I'd expected it to look like the example [here](https://jsonlogic.com/add_operation.html):

```
var plus = function(a,b){ return a+b; };
jsonLogic.add_operation("plus", plus);

jsonLogic.apply({"plus":[23, 19]}); //Returns 42
```

Something on the lines of:

{"isReviewer": {"var": "updatedBy"}}

openmetadata-spec/src/main/resources/json/schema/governance/workflows/elements/triggers/eventBasedEntityTrigger.json
```diff
@@ -36,12 +36,17 @@
           "uniqueItems": true
         },
         "exclude": {
-          "title": "Exclude Fields",
-          "description": "Select fields that should not trigger the workflow if only them are modified.",
+          "title": "Exclude Fields (Deprecated)",
+          "description": "Deprecated: Use 'filter' instead. Select fields that should not trigger the workflow if only them are modified.",
```
IceS2: Again not sure we need to deprecate it or just transform it to the new filter instead. @pmbrull thoughts?

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/utils/JsonLogicUtils.java
pmbrull: maybe this helps https://github.com/open-metadata/OpenMetadata/pull/22108/files#diff-6726d015640ba86e88ebd041dc4096374abe13dee9ee128ca682aa2740815a7bR27

this method here seems a bit hard to follow

openmetadata-spec/src/main/resources/json/schema/governance/workflows/elements/triggers/eventBasedEntityTrigger.json
pmbrull: just migrate it, as any other place in the code. We can deprecate (and keep for some time) APIs, but data should be migrated transparently for users

yan-3005: Yep, using migration!

openmetadata-service/src/main/resources/json/data/governance/workflows/GlossaryApprovalWorkflow.json
yan-3005: Summary from my end would help answer this question!

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/utils/JsonLogicUtils.java
yan-3005: Thanks, utilising that singleton

yan-3005:
@pmbrull  thanks a ton for the PR, using the same singleton but with only my necessary functions.
I am planning to have the custom Operations this way, as the lambda function is taking only a **FUNCTION** and not a **BIFUNCTION** , so this operator evaluator is becoming stateless or context unaware, so necessary to have variables to be sent in the rule the same way the referenece PR is working on the owners, let me know if this works or you guys have something in mind, @IceS2  review as well please.
isReviewer: 
```
{
  "isReviewer": [
    { "var": "reviewers" },
    { "var": "updatedBy" }
  ]
}
```
excludeFields:
```
{
  "excludeFields": [
    { "var": "changeDescription" },
    ["reviewers", "description"]
  ]
}
```

yan-3005:
Working on it, sill in wip

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/EventBasedEntityTrigger.java
```diff
@@ -165,18 +168,34 @@ private CallActivity getWorkflowTrigger(String triggerWorkflowId, String mainWor
...
+      if (triggerConfig.getFilter() != null && !triggerConfig.getFilter().trim().isEmpty()) {
+        // Use JSON Logic path
+        FieldExtension filterExpr =
+            new FieldExtensionBuilder()
+                .fieldName("filterExpr")
+                .fieldValue(JsonUtils.pojoToJson(triggerConfig.getFilter()))
+                .build();
+        serviceTask.getFieldExtensions().add(filterExpr);
+      }
+    }
```
IceS2, line 222: Have you tested this if one of them is null? Im not 100% sure about the behaviour when you have an Expression defined but not passed within the task.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/impl/FilterEntityImpl.java
```diff
@@ -17,40 +17,69 @@
...
-    ChangeDescription changeDescription = oChangeDescription.get();
-
-    List<FieldChange> changedFields = changeDescription.getFieldsAdded();
-    changedFields.addAll(changeDescription.getFieldsDeleted());
-    changedFields.addAll(changeDescription.getFieldsUpdated());
-    return changedFields.isEmpty()
-        || changedFields.stream()
-            .anyMatch(changedField -> !excludedFilter.contains(changedField.getName()));
+
+    return !jsonFilter;
```
IceS2: Nitpick:
Its not that straightforward to understand with the negations everywhere. I think we should aim to write easier to read code.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -1420,4 +1423,81 @@ public boolean isUpdateForImport(GlossaryTerm entity) {
...
+    }
+  }
+
+  private boolean shouldAutoApproveGlossaryTerm(GlossaryTerm glossaryTerm) {
+    // If there are no reviewers, auto-approve regardless of workflow configuration
+    List<EntityReference> reviewers = glossaryTerm.getReviewers();
+    if (reviewers == null || reviewers.isEmpty()) {
+      LOG.debug(
+          "No reviewers found - auto-approving term: {}", glossaryTerm.getFullyQualifiedName());
+      return true;
```
IceS2: I would simplify this to avoid Auto Approving.

I believe every term should start in DRAFT and follow whatever configuration the user defines within the workflow.

One argument to back this approach is a feedback we got about the case where someone created a glossary term and FORGOT to add reviewers. In this case it would go to approved but it should not, it should go to whatever status they set in the workflow for this specific case.

IceS2: This would also simplify the module here as it does not need to be aware of workflow filters whatsoever

openmetadata-service/src/main/resources/json/data/governance/workflows/GlossaryApprovalWorkflow.json
```diff
@@ -11,7 +11,8 @@
     "config": {
       "entityType": "glossaryTerm",
       "events": ["Created", "Updated"],
-      "exclude": ["reviewers"]
+      "exclude": ["reviewers"],
+      "filter": "{\"or\":[{\"isReviewer\":{\"var\":\"updatedBy\"}}]}"
```
IceS2: Are we going to support this OR in the UI?

yan-3005: yes we can support it now

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/impl/FilterEntityImpl.java
yan-3005: In this case, this is necessary instead of changing the jsonLogic. Noted for future changes

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/EventBasedEntityTrigger.java
yan-3005: Now I am going to have an empty filter! something along the lines {and: []} since we are having new nodes, so this shouldnt cause an issue

openmetadata-service/src/main/resources/json/data/governance/workflows/GlossaryApprovalWorkflow.json
```diff
@@ -11,7 +11,8 @@
     "config": {
       "entityType": "glossaryTerm",
       "events": ["Created", "Updated"],
-      "exclude": ["reviewers"]
+      "exclude": ["reviewers"],
+      "filter": "{\"and\":[]}"
```
IceS2: Does this make sense to be the default or could we have it null? I'm not 100% sure

IceS2 (approved):
I think it looks good. Added a minor comment.

One important thing I see here is that we should communicate properly since this is changing something in the middle of what our users are using!

openmetadata-service/src/main/resources/json/data/governance/workflows/GlossaryApprovalWorkflow.json
yan-3005: Let it be this way! If the filter is null, if we try to patch it , it gives key not found, this should be fine, as UI also can render this

karanh37: keep the default value as empty string

yan-3005: okay changed he default value to empty string

pmbrull (approved):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/WorkflowInstanceStateRepository.java
```diff
@@ -176,6 +177,51 @@ public void updateStageWithTask(UUID taskId, UUID workflowInstanceStateId) {
...
+          workflowDefinitionRepository.get(
+              null, instance.getWorkflowDefinitionId(), EntityUtil.Fields.EMPTY_FIELDS);
+
+      // Query all states for this workflow instance
+      long endTs = System.currentTimeMillis();
+      long startTs = endTs - (7L * 24 * 60 * 60 * 1000);
+
+      ResultList<WorkflowInstanceState> instanceStates =
+          listWorkflowInstanceStateForInstance(
+              workflowDefinition.getName(), workflowInstanceId, null, startTs, endTs, 1000, false);
```
pmbrull, line 207: these few lines about getting the state are used a couple of places. Would be good to have a common method so we don't have discrepancies by listing with different times in different places

pmbrull (dismissed):

aniketkatkar97 (approved):

pmbrull (approved):

aniketkatkar97 (approved):

PR #22471, opened 2025-07-20
Fix cluster alias tests

harshach:
Summary of Test Coverage

  We have comprehensive test coverage for the clusterAlias feature:

  ✅ Core Tests

  1. IndexMappingTest - Unit tests for all clusterAlias methods
  2. SearchIndexAppTest - Tests for reindexing with cluster alias
  3. DataInsightsAppTest - Tests for data insights indexes and data streams
  4. SearchResourceTest - End-to-end tests for search operations

  ✅ Key Scenarios Covered

  1. Index name generation with/without cluster alias
  2. Multiple index resolution
  3. Data stream naming for DataInsights
  4. Search operations (single index, multiple indexes, global search)
  5. Backward compatibility (empty/null cluster alias)
  6. Special characters in cluster alias
  7. Different entity types (tables, dashboards, topics, etc.)

  ✅ Improvements

  1. Removed parameterless getIndexName() method from IndexMapping
  2. Enforced that all code must use getIndexName(clusterAlias)
  3. Fixed the one place in SearchRepository that was calling getIndexName() without cluster alias

  ✅ What's Protected

  - Direct index access is now impossible without cluster alias
  - UI remains completely unaware of cluster alias (transparent backend handling)
  - All search operations properly resolve index names
  - Data insights and time series data properly namespaced

  The test suite ensures that the clusterAlias feature works correctly across all components, enabling proper index segregation in shared Elasticsearch/OpenSearch clusters.

### DIFF
--- a/openmetadata-service/src/test/java/org/openmetadata/service/apps/bundles/searchIndex/BulkSinkTest.java
+++ b/openmetadata-service/src/test/java/org/openmetadata/service/apps/bundles/searchIndex/BulkSinkTest.java
@@ -1,0 +1,238 @@
+package org.openmetadata.service.apps.bundles.searchIndex;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


harshach:
@mohityadav766 INdexMapping is not being used in those right?

mohityadav766:
> @mohityadav766 INdexMapping is not being used in those right?

that's where the indexing broke as well 
we are calling below code in those classes
    IndexMapping indexMapping = searchRepository.getIndexMapping(entityType);
    if (indexMapping == null) {
      LOG.debug("No index mapping found for entityType '{}'. Skipping indexing.", entityType);
      return;
    }
    String indexName = indexMapping.getIndexName(searchRepository.getClusterAlias());

mohityadav766:
closing the es/os is upgraded , also lot of pieces are moved to new classes

PR #22474, opened 2025-07-21
fixes 16053: implemented timeout for airflow lineage backend

### LINKED ISSUE #16053: [Airflow Lineage Backend] Add timeout on request client
**Feature**
- Currently, [Airflow Lineage Backend](https://docs.open-metadata.org/v1.2.x/connectors/pipeline/airflow/lineage-backend) does  not have a timeout configuration available. This results to task succeeding but stalling indefinitely due to connection problem between Airflow worker and OM.

**Describe the task**
- Please add a timeout & retry configuration.
  - When requesting from Airflow worker to OM, after configured timeout duration, request should be closed, returning proper Exception.
  - When request fails, lineage ingestion should be retried for configured retrial counts.

Bierbarbar:
### Describe your changes:

Fixes #16053

I worked on fixing because #16053, because we stumbled over it in our own deployed instance

#
### Type of change:

- [X] Bug fix
- [X] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


CLAUDE.md
```diff
@@ -2,111 +2,139 @@
 
 This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
 
-## About OpenMetadata
+## About OpenMetadata Ingestion
```
pmbrull: hi @Bierbarbar should we keep the current logic and add further details on Ingestion instead of removing the existing docs?

Bierbarbar: Hey @pmbrull I'm currently testing arround with claude code a bit. It wasn't intended to submit this file as well. I think if we want to provide such informations for claude we need to rework the file anyways.

ingestion/src/metadata/ingestion/ometa/ometa_api.py
```diff
@@ -161,14 +162,25 @@ def __init__(
...
+                base_url=self.config.hostPort,
+                api_version=self.config.apiVersion,
+                auth_header="Authorization",
+                extra_headers=extra_headers,
+                auth_token=self._auth_provider.get_access_token,
+                verify=get_verify_ssl(self.config.sslConfig),
+            )
+        else:
+            # Update the provided client_config with required OpenMetadata connection settings
+            client_config.base_url = self.config.hostPort
```
pmbrull: sorry reading this again, might be simpler to follow if we directly use the incoming ClientConfig as-is. I'm imagining the flow that if i'm passing a config, i'd like to stick to that, vs. having certain parts overriden by the `init` call.

In any case, we should be capable of getting these configs externally since they come from the `OpenMetadataConnection`. wdyt?

Bierbarbar: The current implementation does not allow to overwrite any other fields beside retry, timeout and timeout_wait. So currently there is no chance that a user overwrites them in that place. The `OpenMetadataConnection` that is generated from `entity/services/connections/metadata/openMetadataConnection.json` does not provide the timeout and retry options as they might be specific for pythons request lib? Not sure here, but I felt it not correct to extend the java class.

pmbrull: ye sorry what I aimed here for was to replicate the `ClientConfig` building logic that happens inside the ometa_api.__init__, but outside.

Or maybe instead of accepting a `client_config: Optional[ClientConfig] = None,` in the init, have a generic `extra_params` as a dict that can be embedded directly within the `ClientConfig`

Bierbarbar: Sure i can refactor this that we do not pass a client config, but instead `client_config_extra_args` dict. I thought it would be better to use the client config object as it is more understandable/pythonic and more failsafe here. 

Would you still prefer the dict - then I'm going to refactor it.

pmbrull: train of thought is it seems strange that we allow to pass a config, but then some of the params of that configuration are going to be overriden by the class. I'd expect if using that constructor that I can either let the class handle the defaults, but if I'm adding something extra, it'd be kept.

I understand your mention around dicts, so as long as we keep the behavior that "inputs prevail", we can handle that via ClientConfig as well.

Thank you!!

Bierbarbar: I just thought about overwriting the config values only if they are set in the client config. But this does not make sense, as you also pass the config object. I might ending up adding the dictionary...

Bierbarbar: @pmbrull i tried to follow the dictionary approach :)

pmbrull (approved):
Thank you!

ingestion/src/metadata/ingestion/ometa/ometa_api.py
pmbrull: thanks for the chat, approved

pmbrull:
hi @Bierbarbar! I see tests are failing on the "setup" step. That usually means there's some issues bringing Airflow up on the development docker compose. Did you have a chance to validate it?

Let us know if you need a hand there

Thanks

Bierbarbar:
> hi @Bierbarbar! I see tests are failing on the "setup" step. That usually means there's some issues bringing Airflow up on the development docker compose. Did you have a chance to validate it?
> 
> Let us know if you need a hand there
> 
> Thanks

@pmbrull i will have a look - i could something todo with the refactor in the dictionary - i will keep you updated

Bierbarbar:
> > hi @Bierbarbar! I see tests are failing on the "setup" step. That usually means there's some issues bringing Airflow up on the development docker compose. Did you have a chance to validate it?
> > Let us know if you need a hand there
> > Thanks
> 
> @pmbrull i will have a look - i could something todo with the refactor in the dictionary - i will keep you updated

@pmbrull i fixed it

pmbrull (approved):

pmbrull:
cherry picked manually to 1.8.7

PR #22491, opened 2025-07-21
FIX #22099: enable 'Column values to be in set' test case for boolean columns

### LINKED ISSUE #22099: Bug: "Column values to be in set" Test Case Unavailable for Boolean Columns
The "Column values to be in set" test case type, which is ideal for validating that a column's values strictly adhere to a predefined set of allowed values, is currently not available for columns identified as a `boolean` data type. This prevents users from easily enforcing strict `true` or `false` validation on boolean columns directly at the column level.

**Impact:**

* **Incomplete Validation:** Users cannot apply a simple, declarative test case to ensure boolean columns contain only `true` or `false`, leading to potential data quality issues if other values (e.g., `1`, `0`, `T`, `F`, `null`, strings) are present.
* **Workaround Inefficiency:** Users are forced to implement workarounds, such as writing custom SQL queries at the table level, which is less intuitive, more cumbersome, and less granular than a column-level test case.
* **Reduced Usability:** The absence of this expected functionality for boolean columns diminishes the completeness and ease of use of the data quality test suite.

**Steps to Reproduce:**

1.  Navigate to a data asset (e.g., a table) that contains a column identified as a `boolean` type.
2.  Attempt to add a new data quality test case specifically for this boolean column.
3.  Observe the list of available test case types.

**Expected Behavior:**

The "Column values to be in set" test case type should be available for selection when configuring a test case for a `boolean` column. This would allow users to specify `true` and `false` as the only allowed values, ensuring strict boolean validation.

**Actual Behavior:**

The "Column values to be in set" test case type is not presented as an option for boolean columns.

**Workaround:**

Currently, the only way to achieve this validation is by using a "Custom SQL Query" test case at the **table level**, targeting the specific boolean column.

Example of a custom SQL query workaround:
```sql
SELECT COUNT(*)
FROM your_table_name
WHERE your_boolean_column NOT IN (TRUE, FALSE);
```

IceS2:
This PR fixes issue #22099 by enabling the "Column values to be in set" test case for boolean columns.

image

## Changes
- Added "BOOLEAN" to supportedDataTypes array in columnValuesToBeInSet.json

## Impact
- Users can now apply strict boolean validation at column level
- Eliminates need for workaround custom SQL queries
- Improves data quality testing completeness for boolean columns

## Testing
- Python validation logic already supports boolean values
- No additional backend changes required

Closes #22099


IceS2:
@claude, can you add some concise tests to this new feature? Rely on the already written tests for it.
They should be within the python module

TeddyCr (approved):

PR #22516, opened 2025-07-22
Fix #22511: k8s secret support for Secrets Manager

### LINKED ISSUE #22511: k8s secret support for Secrets Manager
```
SECRET_NAME="my-secret"
NAMESPACE="default"
TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
CACERT="/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"

curl --cacert $CACERT \
     -H "Authorization: Bearer $TOKEN" \
     -H "Accept: application/json" \
     https://kubernetes.default.svc/api/v1/namespaces/$NAMESPACE/secrets/$SECRET_NAME
```

support these kind of calls with a full Secrets Manager implementation:
- server can create/get secrets
- ingestion can get secrets

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/conf/openmetadata.yaml
+++ b/conf/openmetadata.yaml
@@ -374,7 +374,7 @@ fernetConfiguration:
   fernetKey: ${FERNET_KEY:-jJ/9sz0g0OHxsfxOoSfdFdmk3ysNmPRnH3TUAbz3IHA=}
 
 secretsManagerConfiguration:
-  secretsManager: ${SECRET_MANAGER:-db} # Possible values are "db", "managed-aws","aws", "managed-aws-ssm", "aws-ssm", "managed-azure-kv", "azure-kv", "in-memory", "gcp"
+  secretsManager: ${SECRET_MANAGER:-db} # Possible values are "db", "managed-aws","aws", "managed-aws-ssm", "aws-ssm", "managed-azure-kv", "azure-kv", "in-memory", "gcp", "kubernetes"
   prefix: ${SECRET_MANAGER_PREFIX:-""} # Define the secret key ID as /<prefix>/<clusterName>/<key>
   tags: ${SECRET_MANAGER_TAGS:-[]} # Add tags to the created resource. Format is `[key1:value1,key2:value2,...]`
 # it will use the default auth provider for the secrets' manager service if parameters are not set
--- a/conf/openmetadata.yaml
+++ b/conf/openmetadata.yaml
@@ -390,6 +390,10 @@ secretsManagerConfiguration:
     vaultName: ${OM_SM_VAULT_NAME:-""}
     ## For GCP
     projectId: ${OM_SM_PROJECT_ID:-""}
+    ## For Kubernetes
+    namespace: ${OM_SM_NAMESPACE:-"default"}
+    kubeconfigPath: ${OM_SM_KUBECONFIG_PATH:-""}
+    inCluster: ${OM_SM_IN_CLUSTER:-"false"}
 
 health:
   delayedShutdownHandlerEnabled: true
--- a/ingestion/src/metadata/utils/secrets/kubernetes_secrets_manager.py
+++ b/ingestion/src/metadata/utils/secrets/kubernetes_secrets_manager.py
@@ -1,0 +1,188 @@
+#  Copyright 2025 Collate
+#  Licensed under the Collate Community License, Version 1.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/LICENSE
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""
+Kubernetes Secrets Manager implementation
+"""
+import base64
+import os
+import re
+import traceback
+from abc import ABC
+from typing import Optional
+
+from kubernetes import client, config
+from kubernetes.client.exceptions import ApiException
+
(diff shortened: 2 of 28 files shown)


ingestion/src/metadata/utils/secrets/kubernetes_secrets_manager.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def load_credentials(self) -> Optional[dict]:
+        """Load the provider credentials based on the loader type"""
+        try:
+            loader_fn = secrets_manager_client_loader.registry.get(self.loader.value)
+            return loader_fn()
+        except Exception as err:
+            raise SecretsManagerConfigException(f"Error loading credentials - [{err}]")
+
+    def _sanitize_secret_name(self, secret_name: str) -> str:
```
pmbrull: not sure we need sanitization on the ingestion side. Ingestion will receive the secret_id from the BE, never create them on its own, so we should take care that the secret id is correct on the BE side when storing the secret not here

ingestion/src/metadata/utils/secrets/kubernetes_secrets_manager.py
```diff
@@ -0,0 +1,188 @@
...
+@secrets_manager_client_loader.add(SecretsManagerClientLoader.env.value)
+def _() -> Optional[dict]:
+    namespace = os.getenv("KUBERNETES_NAMESPACE", "default")
+    in_cluster = os.getenv("KUBERNETES_IN_CLUSTER", "false").lower() == "true"
+    kubeconfig_path = os.getenv("KUBERNETES_KUBECONFIG_PATH")
+
+    return {
+        "namespace": namespace,
+        "in_cluster": in_cluster,
+        "kubeconfig_path": kubeconfig_path,
```
pmbrull: we could create a JSON Schema here instead of having the dictionary. Similar to how we have the AWS params and Azure

openmetadata-service/src/main/java/org/openmetadata/service/secrets/KubernetesSecretsManager.java
```diff
@@ -0,0 +1,279 @@
...
+    sanitized = sanitized.toLowerCase();
+    sanitized = sanitized.replaceAll("-+", "-");
+    sanitized = sanitized.replaceAll("^-+|-+$", "");
+
+    if (sanitized.length() > 253) {
+      String hash = Integer.toHexString(secretName.hashCode());
+      sanitized = sanitized.substring(0, 240) + "-" + hash;
+    }
+    if (!sanitized.matches("^[a-z0-9].*")) {
+      sanitized = "om-" + sanitized;
```
pmbrull: I believe we already have a prefix `prefix: ${SECRET_MANAGER_PREFIX:-""} # Define the secret key ID as /<prefix>/<clusterName>/<key>
` we could use here

conf/openmetadata.yaml
```diff
@@ -390,6 +390,10 @@ secretsManagerConfiguration:
     vaultName: ${OM_SM_VAULT_NAME:-""}
     ## For GCP
     projectId: ${OM_SM_PROJECT_ID:-""}
+    ## For Kubernetes
+    namespace: ${OM_SM_NAMESPACE:-"default"}
+    kubeconfigPath: ${OM_SM_KUBECONFIG_PATH:-""}
+    inCluster: ${OM_SM_IN_CLUSTER:-"false"}
```
pmbrull, line 396: do we need a token here?

pmbrull: well I guess the configPath should solve everything, but we'd need to validate this from @tutte if instead of a file we need to pass some cluster config by param?

pmbrull: ok for our main case the `inCluster` would be always `true` and then we don't need any further info since we're in k8s directly

ingestion/src/metadata/utils/secrets/kubernetes_secrets_manager.py
```diff
@@ -0,0 +1,188 @@
...
+            kubeconfig_path = self.credentials.get("kubeconfig_path")
+            if kubeconfig_path:
+                config.load_kube_config(config_file=kubeconfig_path)
+                logger.info(f"Using kubeconfig from path: {kubeconfig_path}")
+            else:
+                config.load_kube_config()
+                logger.info("Using default kubeconfig")
+
+        self.client = client.CoreV1Api()
+        self.namespace = self.credentials.get("namespace", "default")
```
tutte: Namespace could default to the current namespace where the application is running.
```
def _get_current_namespace(self) -> str:
  """
  :return: The namespace where the application service account is running or default if it can't be retrieved
  """
  try:
      with open('/var/run/secrets/kubernetes.io/serviceaccount/namespace') as f:
            return f.read().strip()
  except:
      pass
  logger.info("Can't read the current namespace from in-cluster kubernetes. Is the service account configured?")
  return "default"
```
And then
```
    self.namespace = self.credentials.get("namespace", _get_current_namespace())
```

openmetadata-spec/src/main/resources/json/schema/security/credentials/kubernetesCredentials.json
```diff
@@ -0,0 +1,27 @@
...
+    "javaType": "org.openmetadata.schema.security.credentials.KubernetesCredentials",
+    "properties": {
+      "namespace": {
+        "title": "Namespace",
+        "description": "The namespace of the Kubernetes cluster",
+        "type": "string"
+      },
+      "inCluster": {
+        "title": "In Cluster",
+        "description": "Whether the Kubernetes cluster is in cluster",
```
pmbrull: i guess this is whether the app (server, ingestino) are running as part of the k8s cluster holding the secrets or not. we could clarify this in the description

tutte (approved):

PR #22620, opened 2025-07-28
Fix: #20662 Display Name sort for DataAssets

### LINKED ISSUE #20662: Add Sort Order (e.g., Alphabetical) Support for Data Assets like Metrics
Hi team,

The problem is that there is not any sort order available for data assets in openmetadata right now. As a result, the user is confused at the first glance and is not able to find the desired Metric for example, So he should always use the search box at the top.

Image.

I ask you to consider adding alphabetical sort as the default sorting order for data assets and other sorting orders. For instance, based on the last update time , etc...

Thank you very much for your time and consideration .
Warm Regards,

yan-3005:
### Describe your changes:

Fixes #20662

Sort Order for DisplayName

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/SearchIndex.java
```diff
@@ -94,7 +94,9 @@ default Map<String, Object> getCommonAttributesMap(EntityInterface entity, Strin
     Map<String, Object> map = new HashMap<>();
     map.put(
         "displayName",
-        entity.getDisplayName() != null ? entity.getDisplayName() : entity.getName());
+        entity.getDisplayName() != null && !entity.getDisplayName().isBlank()
```
sonika-shah, line 100: @yan-3005  we can use nullOrEmpty instead of 2 condition check

yan-3005: the common utils doesn't check for isBlank, so lets have isBlank, I will update the commonUtils to handle isBlank as well

PR #22646, opened 2025-07-29
MINOR: Update table profile config to add spark configs. Update spark config to add temp path

IceS2:
Minor update on table profiler configuration and spark config to add a temp path configuration.

#
### Type of change:

- [x] Improvement

#
### Checklist:


ingestion/src/metadata/profiler/source/database/base/profiler_source.py
```diff
@@ -61,16 +61,23 @@ class ProfilerSource(ProfilerSourceInterface):
...
+    @inject
     def __init__(
         self,
         config: OpenMetadataWorkflowConfig,
         database: Database,
         ometa_client: OpenMetadata,
         global_profiler_configuration: ProfilerConfiguration,
+        profiler_config_class: Inject[Type[ProfilerProcessorConfig]] = None,
     ):
+        if profiler_config_class is None:
```
TeddyCr, line 73: What would be the behavior if there is actually no config? Would an empty ProfilerConfig object be passed?

openmetadata-spec/src/main/resources/json/schema/metadataIngestion/engine/sparkEngineConfig.json
```diff
@@ -16,8 +16,19 @@
...
       },
       "config": {
-        "description": "Additional Spark configuration properties as key-value pairs.",
-        "$ref": "../../type/basic.json#/definitions/map"
+        "type": "object",
+        "properties": {
+          "tempPath": {
+            "description": "Temporary path to store the data.",
+            "type": "string",
+            "default": "/tmp/openmetadata"
```
TeddyCr, line 24: This is on the spark clsuter right?

IceS2: Not necessarily, I'll even update the description with one thing...

It is a path that needs to be acessible for all the nodes.

with EMR I used s3, with local I used /tmp on a shared volume.

ingestion/src/metadata/profiler/source/database/base/profiler_source.py
IceS2: That is the class we're using to validate the config. Not the config itself. It should never be null as we are injecting it. If it is null there is a big big issue 🤣

PR #22764, opened 2025-08-05
Fix-22668 Glossary term missing from UI in columns search

### LINKED ISSUE #22668: Glossary term missing from UI in columns searching
**Affected module**
UI

**Describe the bug**

After adding glossary terms to a column, after a term is linked, it's not visible when search this column. We have observed that the issue occurs specifically when searching for a particular column. If we scroll down manually instead of using the search function, the glossary term linkage is visible.

**To Reproduce**

* Go column glossary terms
* Add a term a update
* Refresh the page and search the column, will see the glossary term is not visible.

**Expected behavior**

**Version:**
 - OpenMetadata version: v1.8.3

**Additional context**

sonika-shah:
### Describe your changes:

Fixes  #22668 

image

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -2028,6 +2028,10 @@ private ResultList<Column> searchTableColumnsInternal(
       }
     }
 
+    if (fields.contains("tags") || fields.contains("*")) {
```
harshach, line 2031: why are only doing this for "tags" shouldn't this apply to other fields as well. 
Also check if it applies to DashboardDataModel as well

sonika-shah: For dashboardDataModel it was already present , it was missing only for tables
we’re using  the /column/search api for fetching tags and customMetrics of columns — no other fields

harshach (approved):

PR #22885, opened 2025-08-11
Fix #21091 - Add allAssets field to Domain to include assets from subdomains

### LINKED ISSUE #21091: Data discrepancy between frontend and SDK extraction

sonika-shah:
### Describe your changes:

Fixes #21091

Summary-
 Added a new allAssets field to the Domain entity that aggregates assets from the domain and all its sub-domains recursively, providing a hierarchical view
  of assets under a domain.
  
  

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
@@ -1322,6 +1322,15 @@ public interface CollectionDAO {
       return findToBatchAllTypesWithCondition(fromIds, relation, condition);
     }
 
+    default List<EntityRelationshipObject> findToBatchExcludingType(
+        List<String> fromIds, int relation, String excludeEntityType) {
+      String condition = "";
+      if (excludeEntityType != null && !excludeEntityType.isEmpty()) {
+        condition += " AND toEntity <> '" + excludeEntityType.replace("'", "''") + "'";
+      }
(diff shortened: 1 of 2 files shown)


sonika-shah:
Closing this PR as the functionality already exists.                                                                                                                          
                                                                                                                                                                                
  We have dedicated domain assets APIs that already return assets from the domain and all nested subdomains:                                                                    
                                                                                                                                                                                
  - GET /domains/{id}/assets                                                                                                                                                    
  - GET /domains/name/{fqn}/assets                                                                                                                                              
  - GET /domains/assets/counts                                                                                                                                                  
                                                                                                                                                                                
  These endpoints use a hierarchy query (term match + prefix match on domains.fullyQualifiedName) to automatically include assets from subdomains. See                          
  https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/search/QueryFilterBuilder.java#L152-L167 and              
  https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/search/InheritedFieldEntitySearch.java#L198-L208 which    
  sets supportsHierarchy=true.                                                                                                                                                  
                                                                                                                                                                                
  Adding an allAssets field to the Domain entity would be redundant and less efficient compared to the existing paginated API approach.

PR #22924, opened 2025-08-13
MINOR: Adding missing fallback value for gx integration

IceS2:
### Describe your changes:

This fix aims to provide a fallback table_name value for when it is missing in the checkpoint results. If the checkpoint results contain more than one table and no table_names, all the results will be pushed to the fallback table_name.

If this is a common pattern for gx, we should think about allowing the users to provide a map instead.

#
### Type of change:

- [x] Bug fix

#
### Checklist:


harshach:
@IceS2 add "To Release" label once its ready

PR #22931, opened 2025-08-13
Fix Issue for removed user updating an entity

mohityadav766:
### Describe your changes:

- Fixed the function to be a lenient in check for the user.
- in case the users are delted after updating the entity it should not fail the entities

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
@@ -47,7 +47,7 @@ import static org.openmetadata.service.Entity.FIELD_TAGS;
 import static org.openmetadata.service.Entity.FIELD_VOTES;
 import static org.openmetadata.service.Entity.TEAM;
 import static org.openmetadata.service.Entity.USER;
-import static org.openmetadata.service.Entity.getEntityByName;
+import static org.openmetadata.service.Entity.findEntityByNameOrNull;
 import static org.openmetadata.service.Entity.getEntityFields;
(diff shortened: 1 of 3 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -3558,10 +3558,16 @@ public EntityUpdater(
       this.original = original;
       this.updated = updated;
       this.operation = operation;
-      this.updatingUser =
-          updated.getUpdatedBy().equalsIgnoreCase(ADMIN_USER_NAME)
+      User originalUser =
```
harshach: for every call are we going ot make a user_entity call? 
that will affect the performance severely

harshach (changes requested):
we can’ keep pulling user information for every request they make

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
mohityadav766: tbh haven't checked, it has been like this since our initial implementations, will see it's usage and remove the findEntityByName call in case it's not required

mohityadav766: i just added a condition to create a temp user , in case we can't find the user in the system , since that causes issues, in case the user gets deleted !

yan-3005 (approved):

PR #22932, opened 2025-08-13
FIXES #20888: Validate EntityLink column existence during test case creation

### LINKED ISSUE #20888: DQ - Validate EntityLink during Test creation
re https://github.com/open-metadata/openmetadata-collate/pull/1487

looks like the BE allows us to create tests even when the EntityLink does not give a column that exists in the table.

We've found that when playing with the column casing (e.g., passing lowercase in link when in OM the column is ingested in uppercase).

Then, the test can be created but the actual execution will fail with an empty error on the ingestion side.

IceS2:
**Summary**

This PR fixes issue #20888 where test cases could be created with invalid column references due to case sensitivity issues.

**Changes Made**
1. Added column existence validation to `TestCaseRepository.validateColumnTestCase()`

**Testing**

The fix ensures that:
- Column names are validated case-sensitive
- Test cases with non-existent columns are rejected during creation


TeddyCr:
@IceS2 could we add some tests before merging it please?

PR #22933, opened 2025-08-13
Fix Consistent masking of PII in table column profiling

### LINKED ISSUE #1883: Extend policy to attach it to a location
Lifecycle policies are attached to Locations such as S3 or GCS buckets.

Extend policy schema such that it can be attached to a Locaiton

manerow:
Fixes https://github.com/open-metadata/openmetadata-collate/issues/1883

This PR fixes inconsistent visibility of **PII-Sensitive** information in the **Data Observability → Column Profile** UI:

* **Unauthorized users** (without PII permission) could view the **column detail** profile, even though the **overview** properly masked it.
* **Authorized users** (admins or owners) could see sensitive column profiles in the **detail** view, but not consistently in the **overview**.

### What Changed
* **Centralized PII authorization logic** by moving the `authorizePII` check into `PIIMasker`, ensuring it uses the latest authorization patterns from the `Authorizer` class.
* **Refactored** `PIIMasker.getTableProfile(...)` to accept the **fully qualified table name (FQN)**, so it can fetch the correct `owners` for the authorization check:

  ```java
  public static List<Column> getTableProfile(
      String fqn, List<Column> columns, Authorizer authorizer, SecurityContext securityContext) {
    Table table = Entity.getEntityByName(Entity.TABLE, fqn, "owners", Include.ALL);
    boolean authorizePII = authorizer.authorizePII(securityContext, table.getOwners());
    return maskColumnsIfNotAuthorized(columns, authorizePII);
  }
  ```

* This ensures the `Authorizer.authorizePII(...)` always gets the actual `owners` list, which was previously null or incomplete from `findByName(fqn, ALL)`.

* **Maintains compatibility** with **system/job executions** (where `SecurityContext` is absent) by **overloading**:

  * A `TableRepository.getLatestTableProfile(...)` variant that accepts `(authorizer, securityContext)` for standard flows, and
  * A `PIIMasker.getTableProfile` overload that accepts a `boolean authorizePII` to bypass masking when needed.
   

  There are jobs that invoke `getLatestTableProfile()` without a `SecurityContext`, overloading it avoids breaking existing workflows and preserves backward compatibility

### Why It Matters

PII masking must respect established roles and owner policies consistently across all views:

* Users **without** PII permissions must **never** see sensitive profiles (neither in detail nor overview).
* Users **with** permission should **always** see them.
  *As specified in this comment:* [[see issue discussion](https://github.com/open-metadata/OpenMetadata/issues/10830#issuecomment-1488798742)](https://github.com/open-metadata/OpenMetadata/issues/10830#issuecomment-1488798742).

### Trade-Offs

* Adds **an extra query** per call to fetch the `owners`, ensuring accurate authorization. This is necessary to unify behavior across overview and detail views.

### Validation

* Tested as **unauthorized user**: PII-Sensitive columns are masked everywhere.
* Tested as **admin/table owner**: full visibility of PII profiles in both overview and detail.
* Verified **non-PII columns** behave as before.

---

### Categories

* [x] Bug fix
* [ ] Improvement
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation

### Checklist:
* [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
* [ ] My PR title is `Fixes <issue-number>: <short explanation>`
* [x] I have commented on my code, particularly in hard-to-understand areas.
* [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.
* [x] I have added a test that covers the exact scenario we are fixing.

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
@@ -54,6 +54,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -758,8 +759,26 @@ public class TableRepository extends EntityRepository<Table> {
   }
 
   public Table getLatestTableProfile(
+      String fqn,
+      boolean includeColumnProfile,
+      Authorizer authorizer,
+      SecurityContext securityContext) {
+    return getLatestTableProfileInternal(
+        fqn,
+        includeColumnProfile,
+        t -> PIIMasker.getTableProfile(fqn, t.getColumns(), authorizer, securityContext));
+  }
+
+  public Table getLatestTableProfile(
       String fqn, boolean authorizePII, boolean includeColumnProfile) {
+    return getLatestTableProfileInternal(
+        fqn, includeColumnProfile, t -> PIIMasker.getTableProfile(t.getColumns(), authorizePII));
+  }
+
+  private Table getLatestTableProfileInternal(
+      String fqn, boolean includeColumnProfile, Function<Table, List<Column>> maskFn) {
     Table table = findByName(fqn, ALL);
+
     TableProfile tableProfile =
         JsonUtils.readValue(
             daoCollection
@@ -767,16 +786,16 @@ public class TableRepository extends EntityRepository<Table> {
                 .getLatestExtension(table.getFullyQualifiedName(), TABLE_PROFILE_EXTENSION),
             TableProfile.class);
     table.setProfile(tableProfile);
+
     if (includeColumnProfile) {
       setColumnProfile(table.getColumns());
     }
 
-    // Set the column tags. Will be used to hide the data
-    if (!authorizePII) {
-      populateEntityFieldTags(entityType, table.getColumns(), table.getFullyQualifiedName(), true);
-      table.setColumns(PIIMasker.getTableProfile(table.getColumns()));
-    }
+    // Always populate field tags; the masking strategy decides what to do with them
+    populateEntityFieldTags(entityType, table.getColumns(), table.getFullyQualifiedName(), true);
 
+    // Apply caller-provided masking strategy
+    table.setColumns(maskFn.apply(table));
     return table;
   }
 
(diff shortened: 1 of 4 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -750,26 +758,44 @@ private void setColumnProfile(List<Column> columnList) {
...
+        fqn,
+        includeColumnProfile,
+        t -> PIIMasker.getTableProfile(fqn, t.getColumns(), authorizer, securityContext));
+  }
+
   public Table getLatestTableProfile(
       String fqn, boolean authorizePII, boolean includeColumnProfile) {
+    return getLatestTableProfileInternal(
+        fqn, includeColumnProfile, t -> PIIMasker.getTableProfile(t.getColumns(), authorizePII));
+  }
```
TeddyCr, line 776: Do we still need this method?

openmetadata-service/src/main/java/org/openmetadata/service/security/mask/PIIMasker.java
```diff
@@ -130,27 +131,46 @@ public static SearchIndex getSampleData(SearchIndex searchIndex) {
...
+      String fqn, List<Column> columns, Authorizer authorizer, SecurityContext securityContext) {
+    Table table = Entity.getEntityByName(Entity.TABLE, fqn, "owners", Include.ALL);
+    List<EntityReference> owners = table.getOwners();
+    boolean authorizePII = authorizer.authorizePII(securityContext, owners);
+    return maskColumnsIfNotAuthorized(columns, authorizePII);
+  }
+
+  public static List<Column> getTableProfile(List<Column> columns, boolean authorizePII) {
+    return maskColumnsIfNotAuthorized(columns, authorizePII);
+  }
```
TeddyCr, line 144: Same question here, can we just rely on the method with the new signature?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
manerow: Thanks for the question! We still need this overload because it’s used by `DatabaseServiceTablesProcessor`, which runs in a background job. That job doesn’t initialize a `SecurityContext` or provide an `Authorizer`, and the `Processor::process` contract doesn’t accept a `SecurityContext` parameter.

Given those constraints, the processor can’t call the new `getLatestTableProfile(...)` signature that requires a `SecurityContext`. Changing the `process` contract would impact other processors, and injecting a `SecurityContext`/`Authorizer` into `DatabaseServiceTablesProcessor` isn’t currently wired up in the job.

So, for now we’re keeping this method to support that code path. We can follow up with a separate refactor to pass `SecurityContext`/`Authorizer` into processors (or adjust the contract) and then remove this overload.

openmetadata-service/src/main/java/org/openmetadata/service/security/mask/PIIMasker.java
manerow: For reference, see: https://github.com/open-metadata/OpenMetadata/pull/22933#discussion_r2281823927

PR #23023, opened 2025-08-20
Feat# Implementation of Custom Workflows

yan-3005:
### Describe your changes:

Allow creation of customizable workflows with 
1. Automated Tasks (checkEntityAttributes, setEntityAttribute, rollbackEntity, dataCompleteness)
2. UserApprovalTask (Request Approval Task)

Customizable triggers:
1. EventBasedEntity
2. PeriodicBatch

Update the Existing Workflows to adapt to the new style of workflow nodes.

Tests around the same

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/bootstrap/sql/migrations/native/1.11.0/mysql/postDataMigrationSQLScript.sql
+++ b/bootstrap/sql/migrations/native/1.11.0/mysql/postDataMigrationSQLScript.sql
@@ -1,0 +1,61 @@
+-- Correct the table diff test definition
+-- This is to include the new parameter table2.keyColumns
+UPDATE test_definition
+SET json = JSON_SET(
+        json,
+        '$.parameterDefinition',
+        JSON_ARRAY(
+                JSON_OBJECT(
+                        'name', 'keyColumns',
+                        'displayName', 'Table 1\'s key columns',
+            'description', 'The columns to use as the key for the comparison. If not provided, it will be resolved from the primary key or unique columns. The tuples created from the key columns must be unique.',
+            'dataType', 'ARRAY',
+            'required', false
+        ),
+        JSON_OBJECT(
+            'name', 'table2',
+            'displayName', 'Table 2',
+            'description', 'Fully qualified name of the table to compare against.',
+            'dataType', 'STRING',
+            'required', true
+        ),
+        JSON_OBJECT(
+            'name', 'table2.keyColumns',
+            'displayName', 'Table 2\'s key columns',
+                        'description', 'The columns in table 2 to use as comparison. If not provided, it will default to `Key Columns`, risking errors if the key columns\' names have changed.',
+            'dataType', 'ARRAY',
+            'required', false
+        ),
+        JSON_OBJECT(
+            'name', 'threshold',
+            'displayName', 'Threshold',
+            'description', 'Threshold to use to determine if the test passes or fails (defaults to 0).',
+            'dataType', 'NUMBER',
+            'required', false
+        ),
+        JSON_OBJECT(
+            'name', 'useColumns',
+            'displayName', 'Use Columns',
+            'description', 'Limits the scope of the test to this list of columns. If not provided, all columns will be used except the key columns.',
+            'dataType', 'ARRAY',
+            'required', false
+        ),
+        JSON_OBJECT(
+            'name', 'where',
+            'displayName', 'SQL Where Clause',
+            'description', 'Use this where clause to filter the rows to compare.',
+            'dataType', 'STRING',
+            'required', false
+        ),
+        JSON_OBJECT(
+            'name', 'caseSensitiveColumns',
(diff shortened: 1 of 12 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowEventConsumer.java
```diff
@@ -32,7 +32,55 @@ public class WorkflowEventConsumer implements Destination<ChangeEvent> {
   // ENTITY_RESTORED.
   private static List<EventType> validEventTypes =
       List.of(EventType.ENTITY_CREATED, EventType.ENTITY_UPDATED);
```
IceS2, line 34: Small thought here. Eventually we could have workflows that rely in ENTITY_DELETED, but those are tricky as we can't just get the entity from the database since it was deleted.

Just a small thing for you to think about (=

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
```diff
@@ -298,19 +390,410 @@ public void resolveTask(UUID taskId) {
...
+    if (currentUser != null) {
+      return currentUser;
+    }
+
+    // Try namespaced versions
+    for (String key : variables.keySet()) {
+      if (key.endsWith("_updatedBy")) {
+        currentUser = (String) variables.get(key);
+        if (currentUser != null) {
+          return currentUser;
```
IceS2, line 786: Couldn't this be an issue if you have multiple tasks with updatedBy and you're fetching the first one that might not be the one needed for a given node? 

This is the main reason we created the input_namespace_map, so that the users are able to tell us from which node a given variable should be fetched

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
```diff
@@ -298,19 +390,410 @@ public void resolveTask(UUID taskId) {
...
+    if (rejectionThreshold == null) {
+      rejectionThreshold = 1;
+    }
+
+    // Get current approval tracking variables
+    @SuppressWarnings("unchecked")
+    List<Map<String, Object>> approvals =
+        (List<Map<String, Object>>) taskService.getVariable(task.getId(), "approvals");
+    if (approvals == null) {
+      approvals = new ArrayList<>();
```
IceS2: Maybe we could save a map in the variable with approved: list<user>, rejected: list<user>

This way you don't need to try to get the current user to know if it had already approved or rejected, but you just check the list

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
yan-3005: Noice, yes. WorkflowHandler didnt know the node it was associated with, I will try to store the namespace variable directly in the task and see how that goes
`if (inputNamespaceMap != null && !inputNamespaceMap.isEmpty()) {
       execution.setVariable("inputNamespaceMap", inputNamespaceMap);
   }`

yan-3005: Update: Forgot to mention, already in the resolve task from GlossaryTermRepo or any other repo where resolve task is possible, we send the updatedBy variable, this is a fallback but still seeing on how to optimise this to keep it polished. 
Edit1: I think I will remove this completely, there is no use in this fallback actually, i thought of retrieving it but this handler doesn't even know what all nodes are there, not a great idea to have fall back

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
yan-3005: Implemented this solution

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
yan-3005: Implemented the suggestion of nodeNameSpace

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowEventConsumer.java
yan-3005: For now, this is tricky, i will have it in mind for this and come up with a solution after this

IceS2 (approved):

karanh37 (approved):

sonika-shah (approved):

PR #23120, opened 2025-08-27
FIXES #21808: Add domain filtering support to incident manager

### LINKED ISSUE #21808: Allow filtering of incidents by domain
As a User,
I want to be able to filter the Incidents in Incident Manager by Domain.

This should require a change from the Backend, by if needed, adding the domain property to the incidents. In theory it should be the only change needed as the UI should be entirely filtered by the domain selected on top of the page.

We should, however, double check that this filter works as intended.

IceS2:
## Summary

  This PR implements domain filtering for TestCaseResolutionStatus entities in the incident manager, allowing users to filter incidents by the domain of their associated TestCase. The solution promotes inherited domains from TestCase to the top-level of the search index and implements a search-based filtering approach.

  ## Problem

  TestCaseResolutionStatus entities do not have direct domain associations, but users need to filter incidents by domain.

  The incident manager needed a way to filter TestCaseResolutionStatus records by the domain of their parent TestCase without complex SQL joins across multiple entity relationships.

  ## Solution

  We implemented **Option A: Domain Promotion in Search Index** - a search-based filtering approach that promotes inherited domains to the top-level of the TestCaseResolutionStatus search index.

  ### Key Changes

  1. **Search Index Enhancement** (`TestCaseResolutionStatusIndex.java`)
     - Modified `setParentRelationships()` to promote domains from TestCase to top-level `domains` field
     - Enables standard domain filtering through `SearchListFilter.getDomainCondition()`

  2. **Search Endpoint** (`TestCaseResolutionStatusResource.java`)
     - Added `/search/list` endpoint with domain filtering support
     - Implemented `latest` parameter with aggregation by `testCase.fullyQualifiedName.keyword`

  3. **Search Filter Enhancement** (`SearchListFilter.java`)
     - Added `getTestCaseResolutionStatusCondition()` method
     - Supports filtering by `testCaseResolutionStatusType`, `assignee`, `testCaseFqn`, and `originEntityFQN`

  4. **Repository Updates**
     - **TestCaseResolutionStatusRepository**: Added field exclusion for search-specific fields (`@timestamp`, `domains`, `testCase`, etc.)

  5. **Frontend Integration**
     - Updated `incidentManagerAPI.ts` to support domain filtering parameters
     - Modified `IncidentManager.component.tsx` to use search-based API

  ## Why This Approach?

  ### Rejected Alternatives

  **Option B: Complex SQL Joins**
  - Required 3+ table joins across entity relationships
  - Performance concerns with large datasets
  - Complex query maintenance
  - Doesn't align with existing TestCase domain filtering patterns

  **Option C: Repository-Level Domain Resolution**
  - Would require significant changes to core repository patterns
  - Risk of breaking existing functionality
  - Inconsistent with how other entities handle inherited properties

  ### Benefits of Chosen Solution

  1. **Consistency**: Follows the same pattern used by TestCase domain filtering (search-based)
  2. **Performance**: Leverages pre-indexed search data instead of complex SQL joins
  3. **Maintainability**: Reuses existing `SearchListFilter` infrastructure
  4. **Scalability**: Search-based approach scales better with large datasets
  5. **Non-Breaking**: Doesn't modify existing database queries or entity schemas

#
### Type of change:

- [x] Improvement

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/data_quality/validations/runtime_param_setter/base_diff_params_setter.py
+++ b/ingestion/src/metadata/data_quality/validations/runtime_param_setter/base_diff_params_setter.py
@@ -3,16 +3,24 @@
 from typing import List, Optional, Set, Type, Union
 from urllib.parse import urlparse
 
+from sqlalchemy.engine import make_url
+
 from metadata.data_quality.validations.models import Column, TableParameter
 from metadata.generated.schema.entity.data.table import Table
-from metadata.generated.schema.entity.services.databaseService import DatabaseService
+from metadata.generated.schema.entity.services.databaseService import (
+    DatabaseService,
+    DatabaseServiceType,
+)
 from metadata.generated.schema.entity.services.serviceType import ServiceType
 from metadata.ingestion.connections.connection import BaseConnection
 from metadata.ingestion.source.connections import get_connection
-from metadata.profiler.orm.registry import Dialects
+from metadata.profiler.orm.registry import Dialects, PythonDialects
 from metadata.utils import fqn
 from metadata.utils.collections import CaseInsensitiveList
 from metadata.utils.importer import get_module_dir, import_from_module
+from metadata.utils.logger import test_suite_logger
+
+logger = test_suite_logger()
 
 
 # TODO: Refactor to avoid the circular import that makes us unable to use the BaseSpec class and the helper methods.
@@ -68,7 +76,9 @@ class BaseTableParameter:
         """
         return TableParameter(
             database_service_type=service.serviceType,
-            path=self.get_data_diff_table_path(entity.fullyQualifiedName.root),
+            path=self.get_data_diff_table_path(
+                entity.fullyQualifiedName.root, service.serviceType
+            ),
             serviceUrl=self.get_data_diff_url(
                 service,
                 entity.fullyQualifiedName.root,
@@ -85,7 +95,9 @@ class BaseTableParameter:
         )
 
     @staticmethod
-    def get_data_diff_table_path(table_fqn: str) -> str:
+    def get_data_diff_table_path(
+        table_fqn: str, service_type: DatabaseServiceType
+    ) -> str:
         """Get the data diff table path.
 
         Args:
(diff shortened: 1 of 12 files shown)


TeddyCr (approved):

TeddyCr (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityTimeSeriesRepository.java
```diff
@@ -434,6 +431,42 @@ public ResultList<T> listLatestFromSearch(
...
+    SearchAggregationNode root = new SearchAggregationNode("root", "root", null);
+
+    // Create terms aggregation by groupBy field
+    SearchAggregationNode termsAgg = SearchAggregation.terms("byTerms", groupBy);
+
+    // Add latest_overall (replaces the old "latest" aggregation for parsing compatibility)
+    termsAgg.addChild(SearchAggregation.topHits("latest", 1, "timestamp", "desc"));
+
+    // Add max_timestamp for bucket selector
+    termsAgg.addChild(SearchAggregation.max("max_timestamp", "timestamp"));
```
TeddyCr, line 445: Is it possible I don't have a timestamp for this entity in the index?

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchAggregation.java
```diff
@@ -18,6 +19,77 @@ public SearchAggregation(
     this.rawAggregationMetadata = rawAggregationMetadata;
   }
 
+  /**
```
TeddyCr, line 22: Can we use the `buildAggregation` method of the clients?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityTimeSeriesRepository.java
IceS2: As it is a timeseries it should not be possible, right?

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchAggregation.java
IceS2: Not sure I understood what you mean. Can you explain further?

The idea here is that we are able to create the aggregations without using strings, which should make it easier to debug and then delegate to the clients to translate that into either ES or OS.

TeddyCr: Make sense actually, since we don't need to handle "free form" aggregation.

PR #23145, opened 2025-08-28
fix(ui): fixes 22938 missing input sanitization of suggestions query

### LINKED ISSUE #22938: Searching with UI after custom properties does not work as expected
**Affected module**
The search on the explore page.

**Describe the bug**
Serach queries with custom properties on the explore page fail due to an invalid escape character in the query. But suggestion in the search bar work.

**To Reproduce**
1. Create a custom property
2. Add a value for the custom property to a container (did not test it for other entities)
3. Search on the search bar with the query syntax => Suggestion is correct
4. Press enter to search

Image

The following requests are made from the frontend:
- Search Query when tying in the bar: http://localhost:8585/api/v1/search/query?q=extension.test:abc&index=dataAsset&from=0&size=15&query_filter=%7B%7D
- Search Query when pressed enter: http://localhost:8585/api/v1/search/query?q=extension.test\:abc&index=dataAsset&from=0&size=15&deleted=false&sort_field=_score&sort_order=desc&exclude_source_fields=columns

The `:`-char is replaced with `\:`. This could be the issue.

**Expected behavior**
When press enter in the search bar, the result which where suggested should show up.

**Version:**
 - OS: n/a
 - Python version: n/a
 - OpenMetadata version: 1.8.10
 - Docker image: openmetadata/server:1.8.10
 - Browser: Microsoft Edge v139

**Additional context**
If this is due to some required input sanitation, then the bug is with the sanitation as it is not applied for the suggestions.

jojo599k:
### Describe your changes:

Fixes #22938 

I worked on input sanitization because I noticed it was broken when testing search queries for custom properties.

The input sanitization was not applied for search queries, when they where issued by the global search bar to recommend suggestions.

I used the code from `src/main/resources/ui/src/utils/StringsUtils.ts` which is used in `src/main/resources/ui/src/utils/ExploreUtils.tsx` called from `src/main/resources/ui/src/pages/ExplorePage/ExplorePageV1.component.tsx`.

I did not add tests as I did not found any for either `ExplorePageV1` nor `ExploreUtils`. It would be good if these were still added. But for that, I need support on where exactly this should happen.

#
### Type of change:
- [X] Bug fix

#
### Checklist:

#### Bug fix

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/AppBar/Suggestions.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/AppBar/Suggestions.tsx
@@ -53,6 +53,7 @@ import {
   getGroupLabel,
   getSuggestionElement,
 } from '../../utils/SearchUtils';
+import { escapeESReservedCharacters } from '../../utils/StringsUtils';
 import { showErrorToast } from '../../utils/ToastUtils';
 import Loader from '../common/Loader/Loader';
 
@@ -288,7 +289,7 @@ const Suggestions = ({
       setIsLoading(true);
 
       const res = await searchQuery({
-        query: searchText,
+        query: escapeESReservedCharacters(searchText),
         searchIndex: searchCriteria ?? SearchIndex.DATA_ASSET,
         queryFilter: quickFilter,
         pageSize: PAGE_SIZE_BASE,


harshach:
Thanks @jojo599k for the contribution

PR #23149, opened 2025-08-29
Fix #23010: BigQuery Project ID  Selection While Querying For AutoClassification & Profiler Workflows

### LINKED ISSUE #23010: BQ auto-classification pipelines failing with ADC
**Affected module**
Ingestion Framework

**Describe the bug**
My team has deployed OpenMetadata in the GKE cluster. We have added a multi-project BQ database service, where we are using Application Default Credentials for the authentication, with impersonated Service Account that has all the necessary permissions. We have noticed that auto-classification pipelines work different than when we use GCP Credentials and Service Account key for authentication, because with ADC all queries are run in our default project, instead of the projects configured in the service settings. This results in auto-classification and profiler pipelines failure, because we get a lot of "Dataset Not Found" errors, as these datasets don't exist in the default project and in the queries their projects are not specified in the FROM clause. 

**To Reproduce**
Run auto-classification in the multi-project BQ database service (or single-project but without billing project ID set up), authenticated with ADC. 

**Expected behavior**
Auto-classification and profiler queries should work in the BQ ingestion service, where ADC is used for authentication.

**Version:**
 - OpenMetadata version: 1.9.1
 - OpenMetadata Ingestion package version: 1.9.1

keshavmohta09:
### Describe your changes:

Fixes #23010 

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


TeddyCr:
Have you check what the session is returning for a BQ with a billing project ID? We build the session from the connection here `ingestion/src/metadata/sampler/sqlalchemy/sampler.py` and we'll use the connection for the source returned by the logic in `connection.py`. I am assuming the session should point to the correct billing ID from what I was seeing.

```python
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._table = self.build_table_orm(
            self.entity, self.service_connection_config, self.ometa_client
        )
        self.session_factory = create_and_bind_thread_safe_session(self.connection)
  ```

Same thing in the `ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py`, what project does the session points to?

```python
class SQAProfilerInterface(ProfilerInterface, SQAInterfaceMixin):
    """
    Interface to interact with registry supporting
    sqlalchemy.
    """

    # pylint: disable=too-many-arguments
    def __init__(
        self,
        service_connection_config: Union[DatabaseConnection, DatalakeConnection],
        ometa_client: OpenMetadata,
        entity: Table,
        source_config: DatabaseServiceProfilerPipeline,
        sampler: SamplerInterface,
        thread_count: int = 5,
        timeout_seconds: int = 43200,
        **kwargs,
    ):
        """Instantiate SQA Interface object"""
        self.session_factory = None
        self.session = None

        super().__init__(
            ometa_client=ometa_client,
            entity=entity,
            source_config=source_config,
            sampler=sampler,
            thread_count=thread_count,
        )

        self._table = self.sampler.raw_dataset
        self.create_session()
        self.system_metrics_class = SystemMetricsRegistry.get(
            self.session.get_bind().dialect
        )

    def create_session(self):
        self.session_factory = self._session_factory()
        self.session = self.session_factory()

    @property
    def table(self):
        return self._table

    def _session_factory(self) -> scoped_session:
        """Create thread safe session that will be automatically
        garbage collected once the application thread ends
        """
        return create_and_bind_thread_safe_session(self.connection)
```

keshavmohta09:
Hi @TeddyCr 

In `bigquery/connection.py`, this logic is already handled - whenever a billing project ID is provided, we build the SQLAlchemy engine with that project id, and the **session is created using that engine object**. That's why, in this PR, we have changed the query execution from `schema` to `database.schema` to ensure the full path is used.

keshavmohta09:
Closing this PR since the issue is covered in [this PR](https://github.com/open-metadata/OpenMetadata/pull/23233)

PR #23218, opened 2025-09-03
Allow Alerts to notify Lineage downstream entities stakeholders

manerow:
I worked on extending OpenMetadata's alert notification system to include downstream stakeholders because currently, when an alert is triggered on a data asset, only the direct stakeholders of that asset are notified.
  This misses critical downstream consumers who depend on the affected data.

  **Key changes:**
  - Added `notifyDownstream` and `downstreamDepth` fields to `SubscriptionDestination` schema to enable opt-in downstream notifications
  - Created `LineageGraphExplorer` utility class for cycle-safe lineage graph traversal using DFS with configurable depth limits
  - Modified `SubscriptionUtil.getTargetsForAlert()` to expand notifications to downstream entity stakeholders when enabled
  - Updated all publisher classes (Email, Slack, Teams, GChat, Generic) to pass the full `SubscriptionDestination` object

  **Testing approach:**
  - Created comprehensive integration tests for `LineageGraphExplorer` with real database and lineage data
  - Tested linear traversal, cycle detection, depth limiting, and complex mixed-entity graphs
  - Verified backward compatibility - all existing notification behavior remains unchanged
  - Ran existing publisher tests to ensure no regressions
  
### Additional Notes:
  - No database migration required - the `event_subscription_entity` table stores EventSubscription objects as JSON, automatically accommodating new schema fields
  - Feature is opt-in with `notifyDownstream: false` by default, ensuring zero impact on existing subscriptions
  - Depth limiting and cycle detection prevent performance issues in complex lineage graphs
  - Excludes EXTERNAL category from downstream expansion to prevent notifications to external systems about indirect relationships

### Type of change:
  - [ ] Bug fix
  - [ ] Improvement
  - [x] New feature
  - [ ] Documentation

  ### Checklist:
  
  - [ ] The issue properly describes why the new feature is needed, what's the goal, and how we are building it. Any discussion

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/util/LineageGraphExplorer.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/util/LineageGraphExplorer.java
@@ -13,15 +13,15 @@ import org.openmetadata.service.Entity;
 import org.openmetadata.service.jdbi3.CollectionDAO;
 
 @Slf4j
-public final class DownstreamTraversalService {
+public final class LineageGraphExplorer {
   private final CollectionDAO dao;
   private static final int RELATIONSHIP_UPSTREAM = Relationship.UPSTREAM.ordinal();
 
-  public DownstreamTraversalService(CollectionDAO dao) {
+  public LineageGraphExplorer(CollectionDAO dao) {
     this.dao = dao;
   }
 
-  public Set<EntityReference> findDownstream(UUID id, String type, Integer maxDepth) {
+  public Set<EntityReference> findUniqueEntitiesDownstream(UUID id, String type, Integer maxDepth) {
     if (maxDepth != null && maxDepth == 0) {
       return Set.of();
     }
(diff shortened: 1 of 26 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/util/LineageGraphExplorer.java
```diff
@@ -0,0 +1,70 @@
...
+public final class LineageGraphExplorer {
+  private final CollectionDAO dao;
+  private static final int RELATIONSHIP_UPSTREAM = Relationship.UPSTREAM.ordinal();
+
+  public LineageGraphExplorer(CollectionDAO dao) {
+    this.dao = dao;
+  }
+
+  public Set<EntityReference> findUniqueEntitiesDownstream(UUID id, String type, Integer maxDepth) {
+    if (maxDepth != null && maxDepth == 0) {
```
TeddyCr: should this be `(maxDepth != null || maxDepth == 0)`

TeddyCr: Should we also do `maxDepth <= 0`

manerow: Thanks!

* `(maxDepth != null || maxDepth == 0)` could cause an NPE.
* Agreed on `<= 0`. For non-null values we should stop when `maxDepth <= 0`.

Since `null` means unlimited traversal and `0` (or negative) means no traversal, I’ll keep it as:

```java
if (maxDepth != null && maxDepth <= 0) {
  return Set.of();
}
```

PR #23229, opened 2025-09-03
change contract update to patch call from put

Ashish8689:
### Describe your changes:

- Updated the DataContract now with PATCH call instead of PUT
- Supported DisplayName update from the name input while edit
- Supported Collapsed state if their is no schema entries

https://github.com/user-attachments/assets/49fcca40-1bbd-4908-af87-ff6dc07e06c0

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
@@ -970,7 +970,7 @@ test.describe('Data Contracts', () => {
 
       await test.step('Save contract and validate for schema', async () => {
         const saveContractResponse = page.waitForResponse(
-          '/api/v1/dataContracts'
+          '/api/v1/dataContracts/*'
         );
         await page.getByTestId('save-contract-btn').click();
 
@@ -1014,7 +1014,7 @@ test.describe('Data Contracts', () => {
         ).not.toBeChecked();
 
         const saveContractResponse = page.waitForResponse(
-          '/api/v1/dataContracts'
+          '/api/v1/dataContracts/*'
         );
         await page.getByTestId('save-contract-btn').click();
 
@@ -1056,7 +1056,7 @@ test.describe('Data Contracts', () => {
           }
 
           const saveContractResponse = page.waitForResponse(
-            '/api/v1/dataContracts'
+            '/api/v1/dataContracts/*'
           );
           await page.getByTestId('save-contract-btn').click();
 


openmetadata-ui/src/main/resources/ui/playwright/utils/entity.ts
```diff
@@ -140,20 +144,31 @@ export const addOwner = async ({
...
-    const patchRequest = page.waitForResponse(`/api/v1/${endpoint}/*`);
-    await page.getByTestId('selectable-list-update-btn').click();
-    await patchRequest;
+      const patchRequest = page.waitForResponse(`/api/v1/${endpoint}/*`);
+      await page.getByTestId('selectable-list-update-btn').click();
+      await patchRequest;
+    } else {
+      await page.getByRole('listitem', { name: owner, exact: true }).click();
+      await page.getByTestId('selectable-list-update-btn').click();
+    }
```
chirag-madlani: Things are getting complicated here, can we try to keep it simple?

chirag-madlani: would recommend to create separate function rather then keep adding flags

openmetadata-ui/src/main/resources/ui/src/components/DataContract/AddDataContract/AddDataContract.tsx
```diff
@@ -81,20 +81,26 @@ const AddDataContract: React.FC<{
...
-            entityStatus: EntityStatus.Approved,
-          }));
+      if (contract) {
+        const jsonPatch = compare(contract, {
+          ...contract,
+          ...formValues,
+          displayName: formValues.name,
+        });
+
+        await updateContract(contract?.id, jsonPatch);
```
chirag-madlani: We also need to skip the API if JSON patch is empty

openmetadata-ui/src/main/resources/ui/src/components/DataContract/AddDataContract/AddDataContract.tsx
```diff
@@ -81,20 +81,30 @@ const AddDataContract: React.FC<{
...
+      if (contract) {
+        const jsonPatch = compare(contract, {
+          ...contract,
+          ...formValues,
+          displayName: formValues.name,
+        });
+
+        if (isEmpty(jsonPatch)) {
+          return;
+        }
```
chirag-madlani: If we do return user will stuck on edit mode only no?

chirag-madlani (approved):

PR #23241, opened 2025-09-04
feat(search): custom properties are now searchable in any query

jojo599k:
### Describe your changes:

Fixes n/a

I worked on the search because i wanted the data assets to be found by custom properties in the global search.

Currently the search (`/api/v1/search/query`) does not include the custom properties fields.
This change aims to make them searchable as they getting index in ElasticSearch.

The configuration was changed to include all custom properties reflected by `"extension*"` as a new field, which matches all custom properties. This was done as the custom properties are registered by the user and not known at compile-time. The boost value of `2.0` was chosen as it should have (imo) a similar priority as the description, as though it could be even higher.
As an alternative it could be considered to edit the `SearchSettings` config(!) at runtime whenever a a custom property is added by the user. This would be more complicated but would allow to get rid off the second change.

The second change is to include `.lenient(true)` in the queries generated to ElasticSearch (or OpenSearch) to circumvent the problem that custom properties could be non-text-searchable types like `datetime-cp`. This ignores fields in a query which are not of the correct type. The question is whether such a change is desired.

I have not added any test already as the current test suite for the search module seem to be broken for me on main. It also seem in not so great shape. For the same reason I cannot confirm that my changes broke any tests, I only tested in manually and it seems to work correctly.
Is there something in work on the tests or can I help with that?

#
### Type of change:
- [X] Improvement

#
### Checklist:

#### Improvement

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchSourceBuilderFactory.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchSourceBuilderFactory.java
@@ -80,7 +80,8 @@ public class ElasticSearchSourceBuilderFactory
             .fuzziness(Fuzziness.AUTO)
             .fuzzyMaxExpansions(10)
             .fuzzyPrefixLength(3)
-            .tieBreaker(DEFAULT_TIE_BREAKER);
+            .tieBreaker(DEFAULT_TIE_BREAKER)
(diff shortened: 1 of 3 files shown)


IceS2:
Hey @jojo599k, thanks a lot for the collaboration here. I'm closing the PR in favour of the following issue: https://github.com/open-metadata/OpenMetadata/issues/23315

We should tackle this by allowing the users to customize if they want to enable this or not.

PR #23296, opened 2025-09-08
fix(ui): Fixes 22939: handle table-cp as multiple custom properties in advanced search

### LINKED ISSUE #22939: Advanced Searched Filter generates faulty Elasticsearch query for custom property of type table-cp
**Affected module**
The advanced Search filter of the UI

**Describe the bug**
Its impossible to find values of custom properties which are of type `table-cp`.

**To Reproduce**
1. Create a custom property of type `table-cp`
2. Add a value for the custom property to a container (did not test it for other entities)
3. Use the advanced search filter to search for the value

**Expected behavior**
The container should be findable by the value of the custom property

**Version:**
 - OS: Windows
 - Python version: n/a
 - OpenMetadata version: 1.8.10
 - Docker image: openmetadata/server:1.8.10
 - Browser: Microsoft Edge v139

**Additional context**
The generated query filter for Elasticsearch if the following:
```json
{
  ...
        "must": [
            {
                "wildcard": {
                    "extension.processAttributes": {
                        "value": "*MAT-789*"
                    }
                }
            },
            {
                "term": {
                    "entityType": "container"
                }
            }
        ]
 ...
}
```
which would work if the `table-cp` property was a string but it is a table like so
```json
{
...
    "extension": {
        "processAttributes": {
            "rows": [
                {
                    "key": "sample-id",
                    "value": "MAT-789"
                },
                {
                    "key": "magnification-range",
                    "value": "1000x-50000x"
                },
                {
                    "key": "acceleration-voltage",
                    "value": "15kV"
                }
            ],
            "columns": [
                "key",
                "value"
            ]
        }
    }
}
```
An Elasticsearch query which would work would be the following for searching a value in the table.
```json
{
  ...
      "must": [
        {
          "wildcard": {
            "extension.processAttributes.rows.<COLUMN_NAME>.keyword": "MAT-789"
          }
        },
  ...
}
```
For this to work, it should either be possible to select the column to search through in the UI or the `table-cp` types must be stored as nested in Elasticsearch.

jojo599k:
### Describe your changes:

Fixes #22939

I worked on the advanced search with table custom property because it was not possible to find anything written in a `table-cp`.

There can now be multiple search field for each actual custom property. With my change only the `table-cp` uses this (for now).
Because of this change it is now possible search for each column individually in the advanced search modal and thus make the `table-cp`.

Screenshot_20250908_165938-1

This change was done as it seemed to be simple and contained (only the ui is affected). Alternatively it could be considered to create `table-cp` as `nested` in ElasticSearch to allow it to be searchable even when not specifying a column. Nevertheless I think this is a meaningful change and the alternative could be done "on top" if needed.

Is it needed to write Playwright test for this change? 

#
### Type of change:
- [X] Bug fix

#
### Checklist:


harshach:
@jojo599k thanks for the contribution. @chirag-madlani @karanh37 please review

openmetadata-ui/src/main/resources/ui/src/utils/AdvancedSearchClassBase.ts
```diff
@@ -1030,76 +1030,123 @@ class AdvancedSearchClassBase {
...
+        ];
+
+      case 'table-cp':
+        if (config && typeof config === 'object' && 'columns' in config) {
+          if (
+            config.columns &&
+            Array.isArray(config.columns) &&
+            config.columns.every((item) => typeof item === 'string')
+          ) {
+            return columns.map((col) => ({
```
jojo599k, line 1308:
```suggestion
            return config.columns.map((col) => ({
```

chirag-madlani:
@Ashish8689 @sonika-shah Can you check this one

aniketkatkar97:
Hey @jojo599k, thanks for your contribution. We are taking care of this along with improvements for other custom property types. So closing this one out.

PR #23302, opened 2025-09-09
Domain policy must be evaluated during PolicyEvaluator

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,32 @@ ingestion/.nox/
 # Temporary files
 *.tmp
 *.temp 
+
+# Claude Flow generated files
+.claude/settings.local.json
+.mcp.json
+claude-flow.config.json
+.swarm/
+.hive-mind/
+memory/claude-flow-data.json
+memory/sessions/*
+!memory/sessions/README.md
+memory/agents/*
+!memory/agents/README.md
+coordination/memory_bank/*
+coordination/subtasks/*
+coordination/orchestration/*
+*.db
+*.db-journal
+*.db-wal
+*.sqlite
+*.sqlite-journal
+*.sqlite-wal
+claude-flow
+claude-flow.bat
+claude-flow.ps1
+hive-mind-prompt-*.txt
+.claude/
+.claude-flow/
+docs
+memory
--- a/openmetadata-service/src/main/java/org/openmetadata/service/security/DefaultAuthorizer.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/security/DefaultAuthorizer.java
@@ -16,7 +16,6 @@ package org.openmetadata.service.security;
 import static org.openmetadata.common.utils.CommonUtil.nullOrEmpty;
 import static org.openmetadata.schema.type.Permission.Access.ALLOW;
 import static org.openmetadata.service.exception.CatalogExceptionMessage.notAdmin;
-import static org.openmetadata.service.jdbi3.RoleRepository.DOMAIN_ONLY_ACCESS_ROLE;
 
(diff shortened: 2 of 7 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/CreateResourceContext.java
```diff
@@ -78,6 +78,12 @@ public EntityInterface getEntity() {
 
   @Override
   public List<EntityReference> getDomains() {
+    // First check if the entity being created already has domains assigned
```
mohityadav766: even if the entity has assigned domains ? the parent domains are still inherited right?

openmetadata-service/src/main/java/org/openmetadata/service/Entity.java
```diff
@@ -334,6 +334,19 @@ public final class Entity {
             SPREADSHEET));
   }
 
+  /**
+   * Entities that skip domain checks - includes system-provided entities and entities where domain field is not fully supported.
+   */
+  public static final Set<String> SKIP_DOMAIN_CHECK_ENTITY_LIST =
```
mohityadav766: we can check with **supportsDomains** from EnityRepository ?

openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java
```diff
@@ -100,6 +99,8 @@ public final class EntityUtil {
   public static final Comparator<CustomProperty> compareCustomProperty =
       Comparator.comparing(CustomProperty::getName);
 
+  private static final String DOMAIN_ONLY_ACCESS_ROLE = "DomainOnlyAccessRole";
```
mohityadav766: the whole idea of the PR was to remove this hardcoding?!

openmetadata-service/src/main/java/org/openmetadata/service/Entity.java
sonika-shah: the reason is  supportsDomains checks for `domains` field in the the json file , coming from here -`this.supportsDomains = allowedFields.contains(FIELD_DOMAINS);  `
so for all these entities even when we have domains field supported in their json, but its fully not supported from UI or in backend (not being used in either, patch,put , create)

frame bettre

openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java
sonika-shah: @mohityadav766 this is being used for post filtering by domain  in list api ,for cases like listing glossary/tags/domain - where the result can belong to multiple cross-domains

we need to handle it like this in EntityUtil and cannot be handled from policy evaluator
```
if (resourceContext.getEntity() == null || resourceContext.getEntity().getId() == null) {
      LOG.info(
          "hasDomain() - List operation detected (no specific resource), returning true for post-filtering");
      return true;
    }
```

openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/CreateResourceContext.java
sonika-shah: yes correct , for CreateResourceContext  we need to consider both assigned domains + inherited domains. 
Fixed it!

PR #23335, opened 2025-09-10
MINOR - Data Contracts only execute tests without results

pmbrull:
### Describe your changes:

Update the logic on when we create tests suites, and which ones need to be executed. We'll prioritize getting last results from tests if any

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

---

## Summary by Gitar

- **Data contract validation optimization:**
  - Split execution into two paths: tests WITH results compile existing data, tests WITHOUT results execute via DQ pipeline
  - New filtering methods `filterTestsWithResults()` and `filterTestsWithoutResults()` in `DataContractRepository.java`
- **Bidirectional entity references:**
  - Added `dataContract` field to `testCase.json` schema enabling test cases to reference their parent contracts
  - Lifecycle hooks (`postCreate`, `postUpdate`, `postDelete`) maintain references automatically
- **Database operations enhancement:**
  - Added 3 DAO methods in `CollectionDAO.java` for managing test case references with database-specific implementations (PostgreSQL `jsonb_set` vs MySQL `JSON_SET`)
- **UI component updates:**
  - `ContractQualityCard` now uses `latestContractResults.qualityValidation` prop instead of fetching test suite summary
  - Fetches test cases by entity link rather than test suite ID

---


pmbrull:
this requires more thought on how sync data between tests / DQ. will bring in 1.10.X

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataContractRepository.java
```diff
@@ -208,6 +212,10 @@ protected void postUpdate(DataContract original, DataContract updated) {
     }
 
     postCreateOrUpdate(updated);
+    // Update test cases with dataContract reference for new/updated quality expectations
+    updateTestCasesWithDataContract(updated);
+    // Remove dataContract reference from test cases no longer in quality expectations
+    removeDataContractFromOldTestCases(original, updated);
```
TeddyCr, line 222: why do we update and remove?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataContractRepository.java
```diff
@@ -224,20 +234,27 @@ protected void postDelete(DataContract dataContract, boolean hardDelete) {
...
-        TestSuiteRepository testSuiteRepository =
-            (TestSuiteRepository) Entity.getEntityRepository(Entity.TEST_SUITE);
-        testSuiteRepository.createOrUpdate(null, testSuite, ADMIN_USER_NAME);
-        if (!pipeline.getDeployed()) {
-          prepareAndDeployIngestionPipeline(pipeline, testSuite);
+      // Create the test suite with only the tests pending to be executed
+      List<TestCase> tests = getTestsWithResults(dataContract);
+      List<TestCase> testsWithoutResults = filterTestsWithoutResults(tests);
+      if (!nullOrEmpty(testsWithoutResults)) {
+        TestSuite testSuite = getOrCreateTestSuite(dataContract);
```
TeddyCr, line 245: I was looking at that method, I was wondering if:
1. it should be in the testSuite repository directly in the test suite repo?
2. I was also seeing test suites can have data contract? I think the link should be to the test case itself no?

IMO, you unite of execution should not be relying on creating a testSuite. You should probably create a separate pipeline executing just these test cases with missing results

```json
{
  "airflowConfig": {},
  "displayName": "orders_TestSuite_5YDtkHrr",
  "name": "e7ab1c28-1192-410a-a6e2-e0eaad9e0c97",
  "loggerLevel": "INFO",
  "pipelineType": "TestSuite",
  "raiseOnError": true,
  "service": {
    "id": "eec1120a-bed9-4a34-9c10-1708be968ade",
    "type": "testSuite"
  },
  "sourceConfig": {
    "config": {
      "type": "TestSuite",
      "entityFullyQualifiedName": "Meetup.dev.dbt_jaffle.orders",
      "testCases": [
        "relationships_orders_customer_id__customer_id__ref_customers_",
        "orders_table_row_count_to_equal_jn_xw"
      ]
    }
  }
}
```

If we want to make it seamless for a user we can even delete that pipeline after execution. 🤷

It may also reduce the complexity there is in managing testSuite from the DC POV.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataContractRepository.java
pmbrull: first step is to add the dataContract link to the new test cases in the DQ expectations
second is to remove it from those that have been removed from the DQ expectations

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataContractRepository.java
pmbrull: if we execute some tests via the contract (first time they are created, not yet executed) shouldn't we rely on the suite? Thinking that if we go to the ingestion pipeline directly and run it, we are "skipping steps" that the suite does for us automatically and any changes on how we execute test cases would need to be handled here as well

TeddyCr (approved):

pmbrull:
IDK why table tests are failing, I must have broken something. Will land this in .1

pmbrull:
porting changes to a fresh branch

PR #23425, opened 2025-09-16
Introduce unified ES/OS client API + separate index management

Bhanuagr:
### What’s changed

- Introduced new Java API client support for both Elasticsearch and OpenSearch
- Initialized a “new client” in addition to existing legacy clients. 
- Added a new IndexManagementClient interface to centralize index operations: creation, updates, alias management, deletion etc. 
- Extended SearchClient so it now extends the new IndexManagementClient interface. 
- Updated implementations in ElasticSearchClient / OpenSearchClient + corresponding IndexManager classes. 
- Refactored index‑management logic into its own dedicated manager (moving logic out of client classes where appropriate) to improve separation of concerns.

### Why

- To pave the way for optimisations or replacing legacy client usage without breaking existing functionality.
- To allow better modularity and cleaner interface boundaries between search/query vs index management.
- To support both ES and OS more uniformly, making future enhancements or swaps easier.

<issue-number>

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-spec/src/main/resources/elasticsearch/en/test_case_result_index_mapping.json
```diff
@@ -495,6 +498,5 @@
         }
       }
     }
-  },
-  "priority": 500
```
Bhanuagr, line 500: memo: `priority` is at the top level, but it must be nested under settings.index
- Elasticsearch doesn’t allow arbitrary fields at the top-level of the index creation JSON.
- The priority setting is a part of index settings, not the request metadata.
- The elasticsearch-java client is strict with unknown fields — unlike older REST clients that silently ignored them.

PR #23483, opened 2025-09-19
Remove LMAX Disruptor Dependencies

### LINKED ISSUE #22326: Remove LMAX Disruptor Dependencies
**Affected module**
In our project we have LmAx Dirputor dependency but we can move them to Entity lifecycleDispatcher.java , and remove the LmAx completely.

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Version:**
 - Python version:

**Additional context**

mohityadav766:
Fixes #22326

Removed LMAX Disruptor dependencies and migrated functionality to existing EntityLifecycleEventDispatcher system.

The EntityLifecycleEventDispatcher provides superior performance using virtual threads and better configurability compared to the LMAX Disruptor approach.

---

## Summary by Gitar

- **Dependency removal:**
  - Removed `com.lmax:disruptor` dependency from `pom.xml` and `openmetadata-service/pom.xml`
- **Event system migration:**
  - Deleted `EventPubSub.java` (LMAX Disruptor-based event bus) and created `EventMonitorHandler.java` implementing `EntityLifecycleEventHandler`
  - Migrated `EventMonitorPublisher` to new handler pattern with async execution and priority 1000
- **Interface simplification:**
  - Removed Disruptor dependencies from `EventPublisher` interface, added lifecycle methods `onStart()` and `onShutdown()`
- **Application lifecycle:**
  - Updated startup/shutdown to use `EntityLifecycleEventDispatcher.getInstance()` instead of `EventPubSub`

---


openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplication.java
```diff
@@ -1108,7 +1103,8 @@ public void start() {
     public void stop() throws InterruptedException, SchedulerException {
       LOG.info("Cache with Id Stats {}", EntityRepository.CACHE_WITH_ID.stats());
       LOG.info("Cache with name Stats {}", EntityRepository.CACHE_WITH_NAME.stats());
-      EventPubSub.shutdown();
+      EntityLifecycleEventDispatcher.getInstance().shutdown();
+      AppScheduler.shutDown();
```
manerow, line 1107: **Bug: Duplicate `AppScheduler.shutDown()` call**

`AppScheduler.shutDown()` is called here and again 3 lines below (line 1110). This line was incorrectly added when replacing `EventPubSub.shutdown()`. Remove this one — the existing call at line 1110 is sufficient.

openmetadata-service/src/main/java/org/openmetadata/service/monitoring/EventMonitorHandler.java
```diff
@@ -0,0 +1,106 @@
...
+      eventMonitor.pushMetric(changeEvent);
+    }
+  }
+
+  @Override
+  public void onEntitySoftDeletedOrRestored(
+      EntityInterface entity, boolean isDeleted, SubjectContext subjectContext) {
+    if (Entity.INGESTION_PIPELINE.equals(entity.getEntityReference().getType())) {
+      EventType eventType = isDeleted ? EventType.ENTITY_SOFT_DELETED : EventType.ENTITY_RESTORED;
+      ChangeEvent changeEvent = createChangeEvent(entity, eventType);
```
manerow, line 65: **`shutdown()` is never called**

`EntityLifecycleEventDispatcher.shutdown()` only shuts down its executor thread pool — it does not invoke any lifecycle methods on registered handlers. This means `eventMonitor.close()` will never be called and the resource leaks on application shutdown.

Either call `eventMonitorHandler.shutdown()` explicitly in `ManagedShutdown.stop()`, or update `EntityLifecycleEventDispatcher.shutdown()` to call a shutdown hook on all registered handlers.

openmetadata-service/src/main/java/org/openmetadata/service/monitoring/EventMonitorHandler.java
```diff
@@ -0,0 +1,106 @@
...
+
+  @Override
+  public Set<String> getSupportedEntityTypes() {
+    return Set.of(Entity.INGESTION_PIPELINE);
+  }
+
+  private ChangeEvent createChangeEvent(EntityInterface entity, EventType eventType) {
+    return new ChangeEvent()
+        .withEventType(eventType)
+        .withEntityType(entity.getEntityReference().getType())
```
manerow, line 93: **Narrowing of scope compared to the old publisher**

The old `EventMonitorPublisher` received all change events from the Disruptor ring buffer. This new handler is filtered to only `INGESTION_PIPELINE` entities. If this is intentional (the old publisher also only acted on ingestion pipelines internally), it's worth a brief note. If other entity types should also be monitored, this filter would silently drop them.

openmetadata-service/src/main/java/org/openmetadata/service/monitoring/EventMonitorHandler.java
```diff
@@ -0,0 +1,106 @@
...
+        .withTimestamp(System.currentTimeMillis())
+        .withCurrentVersion(entity.getVersion())
+        .withEntity(entity);
+  }
+
+  public void shutdown() {
+    eventMonitor.close();
+    LOG.info("Event Monitor Handler shutdown completed");
+  }
+}
```
manerow, line 106: Nit: missing newline at end of file.

manerow (changes requested):
**Build is broken — `AuditLogEventPublisher` not migrated**

This PR deletes `EventPubSub` and removes the `batchSize` constructor from `AbstractEventPublisher`, but `AuditLogEventPublisher` was not updated. It still calls `super(batchSize)` and is registered via `EventPubSub.addEventHandler()` in `OpenMetadataApplication.java:971`. This causes two compilation errors that have been failing CI since this PR was opened.

`AuditLogEventPublisher` needs to be either migrated to `EntityLifecycleEventDispatcher` (like `EventMonitorHandler`) or the registration mechanism needs to be reworked. Without this, audit logging is silently broken — there's no event delivery path for it.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #23511, opened 2025-09-22
Undo entity status migration, show entityStatus by default

yan-3005:
### Describe your changes:

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -1186,10 +1186,8 @@ public final T setFieldsInternal(T entity, Fields fields) {
     entity.setReviewers(
         fields.contains(FIELD_REVIEWERS) ? getReviewers(entity) : entity.getReviewers());
     entity.setVotes(fields.contains(FIELD_VOTES) ? getVotes(entity) : entity.getVotes());
-    if (fields.contains(FIELD_ENTITY_STATUS)) {
-      if (entity.getEntityStatus() == null) {
-        entity.setEntityStatus(EntityStatus.APPROVED);
-      }
+    if (entity.getEntityStatus() == null) {
```
mohityadav766: does that means all entities are approved?
Maybe a Unavailable or other option would have been better

yan-3005: @mohityadav766 yeah by default we are having it as Approved. Can you explain why you think another enum is good here? By default having it as Approved would make sense because those entities don't have workflows, if they had one, then they would have the node to send the status to IN REVIEW and go to APPROVED based on user suggestions. Also OSS users when requesting via API would see Approved

mohityadav766: Reason for other enum - if something isn't processed how is that Approved?
Also if entity has entityStatus or not we could control similar to isSupportOwners /isSupportDomain in entityRepository based on entityStatus field presence in json schema , the entities which doesn't support "entityStatus" should not have that field in their json

yan-3005: @mohityadav766 Addressed the review comments, Please review once and approve

mohityadav766 (approved):

mohityadav766 (approved):

karanh37 (approved):

PR #23515, opened 2025-09-23
zero‑downtime reindex

### LINKED ISSUE #23514: zero‑downtime reindex
Reindexing today is disruptive: when an admin selects “recreate indexes,” our CLI/UI drops each Elasticsearch/OpenSearch index before rebuilding. For the entire duration of the reindex job
  the search UI is effectively empty—there’s no data to surface. On production datasets
  this can last almost an hr depending on amount of data.

  The new staged-index workflow eliminates the blackout. Instead of deleting the live index up front, we spin up a parallel index (e.g., table_search_index_rebuild_<timestamp>), backfill
  it, and only swap aliases when the data copy succeeds. The swap is instantaneous: global aliases like all, dataAsset, and the entity-specific alias move in one operation, the old index
  is deleted, and users continue to see consistent search results.

  ### Key benefits

  - Zero downtime: Search remains fully functional while reindexing runs.
  - Safer rollouts: If staging fails, the live index is untouched—no partial rebuilds or unexpected outages.
  - Consistent experience: Global filters (domain, service, asset type) stay populated, so the Explore page doesn’t collapse mid-job.
  - Future-proofing: The alias swap makes it easier to add validation and rollback hooks without impacting availability.

  ### What changed technically

  - DefaultRecreateHandler now builds a staged index per entity, tracks its canonical alias plus parent aliases, and swaps them atomically when the stage completes.
  - A new ReindexContext carries canonical index, the previous alias targets, and staged index metadata across the lifecycle.
  - Alias operations became idempotent—404s are treated as no-ops, and we selectively remove aliases only from indexes that belong to the entity being promoted.
  - Integration tests were added to SearchIndexAppTest to enforce the alias swap contract.

harshach:
### Describe your changes:

Fixes #23514 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/SearchIndexApp.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/SearchIndexApp.java
@@ -10,6 +10,7 @@ import static org.openmetadata.service.apps.scheduler.OmAppJobListener.APP_CONFI
 import static org.openmetadata.service.apps.scheduler.OmAppJobListener.APP_RUN_STATS;
 import static org.openmetadata.service.apps.scheduler.OmAppJobListener.WEBSOCKET_STATUS_CHANNEL;
 import static org.openmetadata.service.socket.WebSocketManager.SEARCH_INDEX_JOB_BROADCAST_CHANNEL;
+import static org.openmetadata.service.workflows.searchIndex.ReindexingUtil.TARGET_INDEX_KEY;
 import static org.openmetadata.service.workflows.searchIndex.ReindexingUtil.isDataInsightIndex;
 
 import com.fasterxml.jackson.core.type.TypeReference;
@@ -97,7 +98,6 @@ public class SearchIndexApp extends AbstractNativeApplication {
   private static final String DISABLED = "Disabled";
   private static final String RECREATE_INDEX = "recreateIndex";
   private static final String ENTITY_TYPE_KEY = "entityType";
-  private static final String TARGET_INDEX_KEY = "targetIndex";
 
   // String constants to avoid duplication
   private static final String APP_SCHEDULE_RUN = "AppScheduleRun";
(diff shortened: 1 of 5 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/SearchIndexApp.java
```diff
@@ -95,6 +97,7 @@ public ReindexingException(String message, Throwable cause) {
   private static final String DISABLED = "Disabled";
   private static final String RECREATE_INDEX = "recreateIndex";
   private static final String ENTITY_TYPE_KEY = "entityType";
+  private static final String TARGET_INDEX_KEY = "targetIndex";
```
Bhanuagr: should we define this variable in constant as used in multiple places ?

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchClient.java
```diff
@@ -393,6 +393,18 @@ public interface SearchClient<T> {
 
   void createAliases(IndexMapping indexMapping);
 
+  void createIndex(String indexName, String indexMappingContent);
```
Bhanuagr, line 396: we already have below method present, can we reuse same one ?
`void createIndex(IndexMapping indexMapping, String indexMappingContent);`

same for all below methods also

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchClient.java
```diff
@@ -328,6 +328,129 @@ public void createAliases(IndexMapping indexMapping) {
     }
   }
 
+  @Override
+  public void createIndex(String indexName, String indexMappingContent) {
```
Bhanuagr, line 332: we already have below method present, can we reuse same one ?
void createIndex(IndexMapping indexMapping, String indexMappingContent);

same for all below methods also

harshach: @Bhanuagr  We intentionally kept the dedicated createIndex(String indexName, String indexMappingContent) overload because the existing createIndex(IndexMapping, ...) routine is tightly coupled to
  canonical indexes:

  - It always creates the index using the canonical name baked into the IndexMapping; we need to spin up a temporary staging index with a timestamped name.
  - After creating the index it automatically attaches the canonical alias set. During staging we must avoid binding any aliases until the index is fully backfilled—otherwise live traffic
  would start flowing to a partially populated index.

  Reusing the canonical overload would either force us to mutate the IndexMapping (risking side effects) or delete the aliases it just created. The string-based method lets us create the
  staging index with the right name and without alias side-effects, so we can explicitly control when aliases move during promotion.

PR #23523, opened 2025-09-23
Fix classification scoring

edg956:
### Describe your changes:

Fixes a likely reason for some false positives in classification.

The normalization function was not taking into account that very low scores (0.1) should be discarded.

#
### Type of change:
- [X] Bug fix

#
### Checklist:

- [X] I have added a test that covers the exact scenario we are fixing. For complex issues, comment the issue number in the test for future reference.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/generated/api/automations/createWorkflow.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/generated/api/automations/createWorkflow.ts
@@ -4332,6 +4332,10 @@ export interface TagLabel {
      */
     name?: string;
     /**
+     * An explanation of why this tag was proposed, specially for autoclassification tags
+     */
+    reason?: string;
+    /**
      * Label is from Tags or Glossary.
      */
     source: TagSource;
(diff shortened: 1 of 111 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -4796,27 +4803,30 @@ default void applyTagsBatch(List<TagLabel> tagLabels, String targetFQN) {
...
         connectionType = POSTGRES)
     void applyTagsBatchInternal(
         @Bind("source") List<Integer> sources,
         @Bind("tagFQN") List<String> tagFQNs,
         @Bind("tagFQNHash") List<String> tagFQNHashes,
         @Bind("targetFQNHash") List<String> targetFQNHashes,
         @Bind("labelType") List<Integer> labelTypes,
-        @Bind("state") List<Integer> states);
+        @Bind("state") List<Integer> states,
+        @Bind("reason") List<String> reasons);
```
pmbrull, line 4832: are we expecting the reason to always be there? what happens if we add a tag via the UI?

edg956: Well spotted. Let me try and see what happens

edg956: @pmbrull I've added tests covering this situation

pmbrull (approved):

pmbrull (dismissed):

chirag-madlani (approved):

PR #23529, opened 2025-09-23
Fixes #23397: Thin Slice for Dimensionality on Data Quality

### LINKED ISSUE #23397: [Data Quality Dimensionality] Implement Thin Slice E2E
- [x] Update JSONSchema to take into account TestCaseDimensions
- [x] Implement Ingestion - We will resort to a Group By query for each dimension, returning at most 10 different dimensions, with the rest grouped
    - [x] Update one Validator to handle Dimensions
    - [x] Implement logic to group least meaningful dimensions into Others
- [x] Implement Backend - We are going to have a new TimeSeriesEntity `TestCaseDimensionResult` that will be managed by the TestCaseResult in order to keep it in the same transaction while we enable a more granular control for fetching the dimension results
    - [x] Implement flow to store TestCaseDimensions
    - [x] Implement flow to retrieve TestCaseDimensions

IceS2:
### Describe your changes:

  Fixes #23397

  I worked on implementing **Dimensional Data Quality** for OpenMetadata, which enables breaking down test results by dimension columns (e.g.,
  region, category, status) to provide granular insights into data quality issues. This feature helps identify specific data segments where quality
  problems occur most frequently.

  ### Major Implementation Areas:

  #### 1. **Schema and Model Updates**
  - Added `dimensionColumns` field to TestCase schema for specifying grouping columns
  - Added `dimensionResults` field to TestCaseResult for storing dimensional breakdowns
  - Created `DimensionResult` and `TestCaseDimensionResult` models
  - Added database migration scripts for MySQL and PostgreSQL (1.11.0)

  #### 2. **Impact Score Algorithm (New Module: `impact_score.py`)**
  - Implemented sophisticated impact scoring to identify most problematic dimensions
  - Formula: `impact = failure_rate² × volume_factor × sample_weight / 1.5`
  - Uses tiered linear scaling for database compatibility (avoids log functions)
  - Handles "Others" aggregation for low-impact dimensions

  #### 3. **Base Test Handler Refactoring**
  - Implemented Template Method Pattern in `BaseTestValidator`
  - Split validation into `_run_validation()` (main) and `_run_dimensional_validation()` (dimensional)
  - Added helper methods:
    - `is_dimensional_test()` - checks if dimensions are configured
    - `get_dimension_result_object()` - creates dimension results with percentages
    - `validate_dimension_columns()` - validates dimension columns exist
    - `_convert_to_test_case_dimension_results()` - converts results for storage

  #### 4. **Mixin Enhancements**
  - **SQLAlchemy Mixin**: Added `_execute_with_others_aggregation()` for CTE-based dimensional queries
  - **Pandas Mixin**: Added dimensional validation support methods
  - Both mixins now support impact score calculations

  #### 5. **Full Dimensional Implementation for Key Validators**
  - **columnValuesToBeInSet** (SQLAlchemy & Pandas) - Complete with impact scoring
  - **columnValuesToBeUnique** (SQLAlchemy & Pandas) - Complete with impact scoring
  - Added dimensional validation stubs to 25+ validators for future expansion

  #### 6. **Error Handling & Resilience**
  - Dimensional validation errors don't abort the main test (only log warnings)
  - Handles missing dimension columns gracefully
  - Supports validators that haven't implemented dimensional logic (NotImplementedError)
  - Backward compatible - tests work identically without dimension columns

  #### 7. **Comprehensive Test Suite**
  - Added extensive tests in `test_base_handler.py`:
    - Template method pattern tests
    - Error handling scenarios
    - Edge cases (0 rows, percentage calculations)
    - Parameterized tests for different scenarios
  - Fixed and enhanced `test_validations_databases.py`:
    - Realistic dimensional validation scenarios
    - Proper SQLAlchemy column mocking
    - Backward compatibility verification

  ### Technical Highlights:
  - **Non-breaking**: Dimensional validation is completely optional
  - **Performance optimized**: Uses CTEs and efficient SQL aggregations
  - **Database agnostic**: Works with SQLite, MySQL, PostgreSQL
  - **Extensible**: Easy to add dimensional support to more validators

  ### Files Modified:
  - 55+ files modified including:
    - 20+ validator implementations
    - Database migration scripts
    - Test suites
    - Schema definitions
    - Base handler and mixins

  #
  ### Type of change:
  - [x] New feature

  #
  ### Checklist:

  ### New feature
  - [x] The issue properly describes why the new feature is needed, what's the goal, and how we are building it. Any discussion or decision-making process
   is reflected in the issue.

### DIFF
--- a/ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/kafka_parser.py
+++ b/ingestion/src/metadata/ingestion/source/pipeline/databrickspipeline/kafka_parser.py
@@ -33,6 +33,12 @@ VARIABLE_ASSIGNMENT_PATTERN = re.compile(
     re.MULTILINE,
 )
 
+# Pattern to extract boolean variable assignments like: snapshot_required = True
+BOOL_ASSIGNMENT_PATTERN = re.compile(
+    r"^\s*([A-Za-z_][A-Za-z0-9_]*)\s*=\s*(True|False)\s*$",
+    re.MULTILINE,
+)
+
 # Pattern to extract DLT table decorators: @dlt.table(name="table_name", ...) or @dlt.table(name=func())
 DLT_TABLE_PATTERN = re.compile(
     r"@dlt\.table\s*\(",
@@ -98,16 +104,25 @@ def _extract_variables(source_code: str) -> dict:
     Examples:
         TOPIC = "events"
         KAFKA_BROKER = "localhost:9092"
+        snapshot_required = True
 
-    Returns dict like: {"TOPIC": "events", "KAFKA_BROKER": "localhost:9092"}
+    Returns dict like: {"TOPIC": "events", "KAFKA_BROKER": "localhost:9092", "snapshot_required": "True"}
     """
     variables = {}
     try:
+        # Extract string variables
         for match in VARIABLE_ASSIGNMENT_PATTERN.finditer(source_code):
             var_name = match.group(1)
             var_value = match.group(2)
             variables[var_name] = var_value
             logger.debug(f"Found variable: {var_name} = {var_value}")
+
+        # Extract boolean variables
+        for match in BOOL_ASSIGNMENT_PATTERN.finditer(source_code):
+            var_name = match.group(1)
+            var_value = match.group(2)
+            variables[var_name] = var_value
+            logger.debug(f"Found boolean variable: {var_name} = {var_value}")
     except Exception as exc:
         logger.debug(f"Error extracting variables: {exc}")
     return variables
@@ -313,24 +328,43 @@ def _infer_table_name_from_function(
     Strategies:
     1. Look for entity_name variable and use it to build table name
     2. Extract keywords from function name (e.g., "event_log" from "generate_event_log_table_name")
+    3. Handle Materializer pattern: entity_name + suffix from function name
     """
     try:
         # Extract variables to find entity_name or similar
         variables = _extract_variables(source_code)
 
-        # Strategy 1: Use entity_name variable if present
+        # Strategy 1: Materializer pattern - entity_name + suffix from function
+        # Handles: @dlt.table(name=materializer.generate_event_log_table_name())
+        # where entity_name = "customerEvent" should produce "customerevent_event_log"
         entity_name = (
             variables.get("entity_name")
             or variables.get("entity")
             or variables.get("table_name")
         )
+
+        if entity_name and "generate_event_log_table_name" in function_call.lower():
+            table_name = f"{entity_name.lower()}_event_log"
+            logger.debug(
+                f"Inferred event_log table from Materializer pattern: {table_name}"
+            )
+            return table_name
+
+        if entity_name and "generate_snapshot_table_name" in function_call.lower():
+            table_name = f"{entity_name.lower()}_snapshot"
+            logger.debug(
+                f"Inferred snapshot table from Materializer pattern: {table_name}"
+            )
+            return table_name
+
+        # Strategy 2: Use entity_name variable if present (fallback)
         if entity_name:
             logger.debug(
                 f"Inferred table name from entity_name variable: {entity_name}"
             )
             return entity_name
 
-        # Strategy 2: Extract from function name (e.g., "event_log" from "generate_event_log_table_name")
+        # Strategy 3: Extract from function name (e.g., "event_log" from "generate_event_log_table_name")
         # Common patterns: generate_X_table_name, create_X_table, build_X_dataframe
         match = re.search(
             r"(?:generate|create|build)_([a-z_]+?)(?:_table|_dataframe)",
@@ -421,7 +455,21 @@ def extract_dlt_table_dependencies(source_code: str) -> List[DLTTableDependency]
                     continue
 
                 # Check if it reads from Kafka
+                # Direct pattern: spark.readStream.format("kafka")
                 reads_from_kafka = bool(KAFKA_STREAM_PATTERN.search(function_block))
+
+                # Materializer pattern: materializer.build_event_log_dataframe()
+                # This method internally reads from Kafka, so if we find this pattern
+                # and the table name matches event_log pattern, mark as Kafka reader
+                if (
+                    not reads_from_kafka
+                    and "materializer.build_event_log_dataframe" in function_block
+                ):
+                    if "event_log" in table_name:
+                        reads_from_kafka = True
+                        logger.debug(
+                            f"Table {table_name} reads from Kafka via Materializer"
+                        )
 
                 # Check if it reads from S3
                 s3_locations = []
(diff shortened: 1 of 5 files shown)


ingestion/src/metadata/data_quality/validations/column/base/columnValuesToBeUnique.py
```diff
@@ -75,8 +79,66 @@ def run_validation(self) -> TestCaseResult:
...
+                        column, dimension_col, metrics_to_compute
+                    )
+
+                    dimension_results.extend(single_dimension_results)
+
+                except Exception as exc:
+                    logger.warning(
+                        f"Error executing dimensional query for column {dimension_column}: {exc}"
+                    )
+                    continue
```
TeddyCr, line 213: I think we can extract this part, it looks like it is common in multiple files.

ingestion/src/metadata/data_quality/validations/column/base/columnValuesToBeUnique.py
```diff
@@ -99,3 +161,37 @@ def _get_unique_count(self, metric: Metrics, column: Union[SQALikeColumn, Column
             Tuple[int, int]:
         """
         raise NotImplementedError
+
+    @abstractmethod
+    def _execute_dimensional_query(
```
TeddyCr: a bit of a nit but we should remove the `query` of the method name as the base class are inherited not just by query type interface.

ingestion/src/metadata/data_quality/validations/column/pandas/columnValuesToBeInSet.py
```diff
@@ -65,3 +91,116 @@ def compute_row_count(self, column: SQALikeColumn):
             NotImplementedError:
         """
         return self._compute_row_count(self.runner, column)
+
+    def _execute_dimensional_query(
```
TeddyCr: It seem we are reimplementing the logic to compute the data quality test. That is going to create overhead down the road as we'll need to make sure that we correctly keep the same logic in sync in 2 places. We should think about how we can have the logic to compute the result in 1 place.

(same thing with the other implementations)

ingestion/src/metadata/data_quality/validations/column/sqlalchemy/columnValuesToBeInSet.py
```diff
@@ -61,3 +78,118 @@ def compute_row_count(self, column: Column):
...
+            for metric_name, metric in metrics_to_compute.items():
+                metric_instance = metric.value(column)
+                if metric_name == "count_in_set":
+                    metric_instance.values = allowed_values
+                metric_expressions[metric_name] = metric_instance.fn()
+
+            from metadata.data_quality.validations.base_test_handler import (
+                DIMENSION_FAILED_COUNT_KEY,
+                DIMENSION_TOTAL_COUNT_KEY,
+            )
```
TeddyCr: Do you know why we have these import in the middle of the code?

ingestion/src/metadata/data_quality/validations/column/sqlalchemy/columnValuesToBeInSet.py
```diff
@@ -61,3 +78,118 @@ def compute_row_count(self, column: Column):
...
+        """
+        from sqlalchemy import func
+
+        from metadata.data_quality.validations.column.base.columnValuesToBeInSet import (
+            ALLOWED_VALUE_COUNT,
+        )
+        from metadata.data_quality.validations.impact_score import (
+            DEFAULT_TOP_DIMENSIONS,
+        )
+        from metadata.generated.schema.tests.basic import TestResultValue
```
TeddyCr: It seems we are reimporting func. Do you know why we have these imports in the middle of the code?

ingestion/src/metadata/data_quality/validations/mixins/sqa_validator_mixin.py
```diff
@@ -126,3 +144,146 @@ def _compute_row_count(self, runner: QueryRunner, column: Column, **kwargs):
...
+                    (
+                        dimension_stats.c.dimension_value.in_(
+                            select([top_dimensions.c.dimension_value])
+                        ),
+                        dimension_stats.c.dimension_value,
+                    )
+                ],
+                else_=DIMENSION_OTHERS_LABEL,
+            ).label(DIMENSION_GROUP_LABEL)
+        ]
```
TeddyCr, line 232: Do you think we'll have performance gain to split this in 2 operations:
1. get top `n` %/# of rows with the highest impact score
2. get the rest

This way we don't need to have a CASE / WHEN / THEN.

ingestion/src/metadata/data_quality/validations/mixins/sqa_validator_mixin.py
```diff
@@ -126,3 +144,146 @@ def _compute_row_count(self, runner: QueryRunner, column: Column, **kwargs):
...
+
+        categorized = (
+            select(categorized_columns)
+            .select_from(dimension_stats)
+            .cte(CTE_CATEGORIZED)
+        )
+
+        # Final query: Aggregate by dimension_group
+        final_columns = [
+            getattr(categorized.c, DIMENSION_GROUP_LABEL).label(DIMENSION_VALUE_KEY)
```
TeddyCr, line 247: If I understand correctly this will always return `n + 1` where `n` is the number of top impact categories we want to return and 1 is the `other` category?

ingestion/src/metadata/data_quality/validations/table/base/tableColumnCountToEqual.py
```diff
@@ -62,6 +67,19 @@ def run_validation(self) -> TestCaseResult:
...
+        This method should implement the dimensional logic specific to each test type.
+        It will be called automatically by the template method when dimensionColumns
+        are configured in the test case.
+
+        Returns:
+            List[DimensionResult]: List of dimension-specific test results
+        """
+        # Default implementation returns empty list
+        return []
+
```
TeddyCr: What do you think of moving this to `BaseTestValidator`?

ingestion/src/metadata/data_quality/validations/base_test_handler.py
```diff
@@ -55,12 +70,111 @@ def __init__(
...
+            if not self.are_dimension_columns_valid():
+                # Don't abort the main test, just skip dimensional validation
+                # The main test result is still valid
+                return test_result
+
+            try:
+                dimension_results = self._run_dimensional_validation()
+                if dimension_results:
+                    logger.debug(
+                        f"Dimensional validation completed with {len(dimension_results)} results"
```
TeddyCr, line 156: It seems that we would be executing the test twice:
1. the full test
2. the dimensional test

I believe we should be able to capture the full test results from the dimensional test results?

Otherwise we are adding a bit of overhead.

ingestion/src/metadata/data_quality/validations/mixins/sqa_validator_mixin.py
IceS2: Hey @TeddyCr , I did not understand what you meant.

ingestion/src/metadata/data_quality/validations/mixins/sqa_validator_mixin.py
IceS2: Exactly

ingestion/src/metadata/data_quality/validations/base_test_handler.py
IceS2: We are doing exactly that. I'm not sure all teh tests are aggregatable. Do we always rely on failed number of rows vs total number of rows?

If this is the case, I imagine we could execute only the dimensional tests and aggregate from the first dimensional test if any.

I think it makes the code a bit more convoluted as we will be executing the test n + 1 times (where n is the number of dimensions configured)

ingestion/src/metadata/data_quality/validations/column/sqlalchemy/columnValuesToBeInSet.py
IceS2: Bloop 😅

ingestion/src/metadata/data_quality/validations/column/sqlalchemy/columnValuesToBeInSet.py
IceS2: Bloop 😅

ingestion/src/metadata/data_quality/validations/column/pandas/columnValuesToBeInSet.py
IceS2: I understand what you mean, but I would rather start with duplicated implementation and after see what we can extract that start the implementation by trying to guess what would be best, as I'm not that familiar with Data Quality.

That was the main approach

ingestion/src/metadata/data_quality/validations/column/base/columnValuesToBeUnique.py
IceS2: Fair enough!

ingestion/src/metadata/data_quality/validations/column/base/columnValuesToBeUnique.py
IceS2: Similar to what I've answered below, I was not worrying about that at the moment as it would be more visible when implementing more validators and I'd understand better where to extract it to

ingestion/src/metadata/data_quality/validations/table/base/tableColumnCountToEqual.py
IceS2: You mean having the method as a default implementation instead of an abstract method? I think it'd be a good idea to be honest 🤣

ingestion/src/metadata/data_quality/validations/base_test_handler.py
TeddyCr: Do we always rely on failed number of rows vs total number of rows?

> We do not, maybe we can worry about that optimization later if we hear complaints.

ingestion/src/metadata/data_quality/validations/mixins/sqa_validator_mixin.py
TeddyCr: SOMETHING LIKE

1. First op
```sql
SELECT foo FROM table ORDER BY impactScore LIMIT `n`
```

2. Second op
```sql
SELECT foo AS Other FROM table WHERE foo not IN (<results from op 1>)
```

TeddyCr (dismissed):

PR #23588, opened 2025-09-26
Add tooltip for icon

ugouhar:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ugouhar:
The translations are generated using Claude and I have verified each translation with google translate. All convey the same meaning with very minor changes in phrases. Some examples:

1. Icon aspect ratio should be 1:1 and recommended size should be 64 x 64 px
2. The icon aspect ratio should be 1:1 and the recommended size should be 64 x 64 px
3. Icon aspect ratio is 1:1, recommended size is 64 x 64 px.

openmetadata-ui/src/main/resources/ui/src/components/common/IconPicker/MUIIconPicker.tsx
```diff
@@ -129,7 +131,31 @@ const MUIIconPicker: FC<MUIIconPickerProps> = ({
 
   return (
     <FormControl component="fieldset" disabled={disabled}>
-      {label && <FormLabel>{label}</FormLabel>}
+      {label && (
```
karanh37, line 133: we are hardcoding this. 
The common component should expect a tooltip from the usage, if i want to reuse this icon picker else i should be able to customise it

openmetadata-ui/src/main/resources/ui/src/components/common/IconPicker/MUIIconPicker.tsx
```diff
@@ -129,7 +130,12 @@ const MUIIconPicker: FC<MUIIconPickerProps> = ({
...
+        <Box sx={{ display: 'inline-flex' }}>
+          <FormLabel>{label}</FormLabel>
+          {toolTip}
+        </Box>
+      )}
```
ugouhar, line 138: Here `FormLabel` will be wrapped with `Box` also when `toolTip` is not present but that won't make any visual difference

PR #23589, opened 2025-09-26
Identify when Airflow DAG import errors in CI

edg956:
### Describe your changes:

Fixes the issue mentioned in this [Slack thread](https://collate-inc.slack.com/archives/C044A8L48P2/p1757934328709299).

I improved the loop that waits for airflow to import the sample data DAG. This is so that if a change in `ingestion` inadvertently introduces a bug that breaks `Setup Openmetadata Test Environment` step we can quickly identify it in the error log.

#
### Type of change:
- [X] Improvement

#
### Checklist:
- [X] I have tested this locally

### DIFF
--- a/docker/run_local_docker.sh
+++ b/docker/run_local_docker.sh
@@ -10,6 +10,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+RED='\033[0;31m'
+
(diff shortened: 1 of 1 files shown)


tutte (approved):
LGTM, thanks @edg956

PR #23590, opened 2025-09-26
MINOR - Fix S3 logging from ingestion pipelines

pmbrull:
### Describe your changes:

- fix invalid endpoint when initializing logs
- allow using `enableStreamableLogs` via YAML ingestion
- use the proper configuration as part of the workflow instead of relying on the Ingestion Pipeline API
- update airflow apis to handle enableStreamableLogging
- fix logging lifecycle to include the workflow summary. This means moving the print_summary inside the workflow call before `stop`. Added the note in Breaking Changes, even if it's just esthetic.

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach (approved):

conf/openmetadata.yaml
```diff
@@ -361,16 +361,20 @@ pipelineServiceClientConfiguration:
     certificatePath: ${PIPELINE_SERVICE_CLIENT_SSL_CERT_PATH:-""} # Local path for the Pipeline Service Client
   logStorageConfiguration:
     type: ${PIPELINE_SERVICE_CLIENT_LOG_TYPE:-"default"} # Possible values are "default", "s3"
+    enabled: ${PIPELINE_SERVICE_CLIENT_LOG_ENABLED:-false} # Enable it for pipelines deployed in the server
     # if type is s3, provide the following configuration
     bucketName: ${PIPELINE_SERVICE_CLIENT_LOG_BUCKET_NAME:-""}
     # optional path within the bucket to store the logs
     prefix: ${PIPELINE_SERVICE_CLIENT_LOG_PREFIX:-""}
-    region: ${PIPELINE_SERVICE_CLIENT_LOG_REGION:-""}
+    enableServerSideEncryption: ${PIPELINE_SERVICE_CLIENT_LOG_SSE_ENABLED:-true}
```
harshach: shouldn’t this be false by default

openmetadata-ui/src/main/resources/ui/src/generated/configuration/logStorageConfiguration.ts
```diff
@@ -108,6 +116,14 @@ export interface AWSCredentials {
     profileName?: string;
 }
 
+/**
+ * Server-side encryption algorithm (if applicable)
+ */
+export enum SSEAlgorithm {
+    Aes256 = "AES256",
```
harshach, line 123: can we increase the SSE algo here instead of 256

PR #23607, opened 2025-09-29
SSO minor improvements around field level errors

shrushti2000:
1.  Updated request Scope field to array field input which was initially an normal text input field
Screenshot 2025-09-29 at 5 25 18 PM

2. Added validation for public key url

https://github.com/user-attachments/assets/d04b20f2-e748-4d42-b525-bcecc7792971

3. Removes Required field restriction on Provider Name in case of configured sso's. Since there might be sso's which are already configured , not having provider names. (Ex current sandbox-beta)

Provider Name(required) restriction will only be applied for Fresh configurations.
Screenshot 2025-09-29 at 4 57 30 PM

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/components/SettingsSso/SSOConfigurationForm/SsoConfigurationFormArrayFieldTemplate.tsx
```diff
@@ -151,9 +197,19 @@ const SsoConfigurationFormArrayFieldTemplate = (props: FieldProps) => {
...
-          onBlur={() => props.onBlur(id, value)}
-          onChange={(value) => {
-            props.onChange(value);
+          onBlur={() => {
+            const convertedValue = isScopeField ? value.join(' ') : value;
+            props.onBlur(id, convertedValue);
+          }}
+          onChange={(newValue) => {
+            // Handle scope field conversion from array (UI) to string (backend)
+            const convertedValue = isScopeField
```
Ashish8689: we should move this out for the render and same with onBlur!

shrushti2000: addressed

Ashish8689 (approved):

openmetadata-ui/src/main/resources/ui/src/utils/SSOUtils.ts
```diff
@@ -23,6 +23,26 @@ import SamlIcon from '../assets/svg/ic-saml.svg';
...
+
+  try {
+    const url = new URL(urlString.trim());
+
+    // Check if the protocol is http or https and has a valid hostname
+    return ['http:', 'https:'].includes(url.protocol) && !!url.hostname;
+  } catch {
+    return false;
+  }
+};
```
chirag-madlani, line 36: Can't we use a regexp here?

shrushti2000: addressed, updated to use regex

chirag-madlani (approved):

PR #23609, opened 2025-09-29
feat: Create reusable EntityAvatar component for consistent icon rend…

satishwaghela:


openmetadata-ui/src/main/resources/ui/src/components/common/EntityAvatar/EntityAvatar.tsx
```diff
@@ -0,0 +1,123 @@
...
+        }}>
+        <IconComponent size={size * 0.6} style={{ strokeWidth: 1.5 }} />
+      </Avatar>
+    );
+  }
+
+  // Default icons when no iconURL is provided
+  let DefaultIcon;
+  if (isSubDomain) {
+    DefaultIcon = SubDomainIcon;
```
karanh37, line 105: Can we handle this with some map of entity and its icon, rather than hardcoding here

PR #23637, opened 2025-09-30
Fixes: Drive Service

keshavmohta09:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


sonika-shah (approved):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SpreadsheetRepository.java
```diff
@@ -339,7 +341,11 @@ protected void createEntity(CSVPrinter printer, List<CSVRecord> csvRecords) thro
                       Pair.of(9, TagLabel.TagSource.CLASSIFICATION),
                       Pair.of(10, TagLabel.TagSource.GLOSSARY))))
           .withDomains(getDomains(printer, csvRecord, 11))
-          .withDataProducts(getDataProducts(printer, csvRecord, 12));
+          .withDataProducts(getDataProducts(printer, csvRecord, 12))
+          .withCreatedTime(
```
yan-3005, line 345: We are adding createdTime and modifiedTime in the csvImport and export, are you sure we need this?
Also we need to add the fields in the spreadsheetCsvDocumentation which is missing

keshavmohta09: I added `createdTime` and `modifiedTime` by following the same approach we already use for `mimeType`, since it’s the same use case. My focus was only on filling the missing information needed for display in the UI, so I didn't update `spreadsheetCsvDocumentation`. **If required, we can include those updates in a future PR.**

PR #23711, opened 2025-10-04
Fix #23710: GlossaryTerm - Circular references cause API to hang: Self-referential or circular parent relationships in glossary terms cause infinite loops when querying with directChildrenOf parameter

### LINKED ISSUE #23710: GlossaryTerm  - Circular references cause API to hang: Self-referential or circular parent relationships in glossary terms cause infinite loops when querying with directChildrenOf parameter
**Affected module**
GlossaryTerm

**Describe the bug**
## Bug 1: Circular Reference Infinite Loop

  Current Behavior

  - API endpoint /api/v1/glossaryTerms?directChildrenOf=<term> hangs indefinitely and throws UNHANDLED_SERVER_EXCEPTION
  - No error logs generated, just repeated timeout requests
  - Root cause: Database contained self-referential relationship (Account → Account) in entity_relationship table

  Example of corrupted data:
  -- Self-referential circular relationship
  SELECT * FROM entity_relationship
  WHERE fromid = toid = '24d7b98a-460b-4a3b-a0f5-1d1568bb1139'
  AND relation = 0;

  Root Cause

  The validateHierarchy() method only validated FQN strings, not actual database relationship graphs:
  // OLD - only checked FQN prefix matching
  if (term.getFullyQualifiedName().startsWith(parentTerm.getFullyQualifiedName())) {
    throw new IllegalArgumentException("Circular reference");
  }

  This failed to detect:
  - Self-references (term pointing to itself)
  - Indirect circular chains (A → B → C → A)
  - Corrupted data created outside normal validation flows

## Bug 2: Incorrect childrenCount 

When fetching a glossary term with the childrenCount field:
  - Only direct children are counted
  - Nested descendants (grandchildren, great-grandchildren, etc.) are excluded from the count

  Example:
  term1
  ├── term1.1
  │   ├── term1.1.1
  │   └── term1.1.2
  └── term1.2

  Currently term1.childrenCount = 2 (only counting term1.1 and term1.2)

  **Expected Behavior**

  The childrenCount field should show the total count of ALL nested descendants at all levels.

  Example:
  Using the same hierarchy above, term1.childrenCount should be 4 (term1.1, term1.1.1, term1.1.2, term1.2)

harshach:
…

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
@@ -199,10 +199,8 @@ public class GlossaryTermRepository extends EntityRepository<GlossaryTerm> {
   }
 
   private Integer getChildrenCount(GlossaryTerm term) {
-    return daoCollection
-        .relationshipDAO()
-        .findTo(term.getId(), GLOSSARY_TERM, Relationship.CONTAINS.ordinal(), GLOSSARY_TERM)
-        .size();
+    // Count all nested terms using FQN prefix matching for efficiency
+    return daoCollection.glossaryTermDAO().countNestedTerms(term.getFullyQualifiedName());
   }
 
   private List<EntityReference> getRelatedTerms(GlossaryTerm entity) {
@@ -222,14 +220,10 @@ public class GlossaryTermRepository extends EntityRepository<GlossaryTerm> {
       parentReviewers = parentTerm.getReviewers();
       entity.setParent(parentTerm.getEntityReference());
     }
-
     // Validate glossary
     Glossary glossary = Entity.getEntity(entity.getGlossary(), "reviewers", Include.NON_DELETED);
     entity.setGlossary(glossary.getEntityReference());
-    parentReviewers = parentReviewers != null ? parentReviewers : glossary.getReviewers();
-
     validateHierarchy(entity);
-
     // Validate related terms
     EntityUtil.populateEntityReferences(entity.getRelatedTerms());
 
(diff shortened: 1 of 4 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -222,14 +220,10 @@ public void prepare(GlossaryTerm entity, boolean update) {
       parentReviewers = parentTerm.getReviewers();
       entity.setParent(parentTerm.getEntityReference());
     }
-
     // Validate glossary
     Glossary glossary = Entity.getEntity(entity.getGlossary(), "reviewers", Include.NON_DELETED);
     entity.setGlossary(glossary.getEntityReference());
-    parentReviewers = parentReviewers != null ? parentReviewers : glossary.getReviewers();
```
sonika-shah, line 229: I guess this should not be removed , in case term does not have its reviewers , it should inherit from glossary

yan-3005: Not needed here, this is needed for setting the entityStatus, hence a left over, I cleaned up the remaining areas as well

pmbrull (approved):

yan-3005:
Cherry Picked to 1.9.13
 https://github.com/open-metadata/OpenMetadata/commit/6228549832d11c0f747bf9b05f51efa5cb62b6db

PR #23758, opened 2025-10-07
Playwright: Fix flaky user profile persona interactions

shah-harshit:
### Describe your changes:

Fix flaky user persona interactions along with the curated assets tests

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Users.spec.ts
```diff
@@ -947,7 +950,7 @@ test.describe('User Profile Dropdown Persona Interactions', () => {
   });
 });
 
-test.describe('User Profile Persona Interactions', () => {
+test.describe.serial('User Profile Persona Interactions', () => {
```
ShaileshParmar11, line 953: Why do we need this in serial mode?

openmetadata-ui/src/main/resources/ui/playwright/utils/widgetFilters.ts
```diff
@@ -145,21 +179,33 @@ export const verifyDataProductsFilters = async (
   await expect(sortDropdown).toBeVisible();
 
   const aToZFilter = page.waitForResponse(
-    '/api/v1/search/query?q=*&index=data_product*&sort_field=name.keyword&sort_order=asc'
```
ShaileshParmar11, line 148: Aren't both the same?

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Users.spec.ts
shah-harshit: There are two tests:

1. The first test uses existing personas in the users list — it performs operations on them and verifies their removal at the end.
2. The second test re-adds those personas, since they were removed in the first test.

If the execution order changes, the personas may not be available as expected, causing the tests to fail.

openmetadata-ui/src/main/resources/ui/playwright/utils/widgetFilters.ts
shah-harshit: I noticed this becomes flaky sometimes. Also, we have used includes for API responses in all the widget filters so to make it consistent I made these changes.

Issue #23778, opened 2025-10-08
bug: LoggerLevel does not work for ingestion pipelines


aballiet:
**Affected module**
Ingestion Framework

**Describe the bug**
Setting 

```yaml
workflowConfig:
        loggerLevel: ERROR
```

Does not affect logger, we still have warning logs:

<img width="1250" height="935" alt="Image" src="https://github.com/user-attachments/assets/2fce2489-3808-4421-b299-fea1d5113d5f" />

Which generates in this case a lot of logs...

**To Reproduce**

Run airbyte ingestion pipeline for exemple as it generates warning due to usage of a deprecated function https://github.com/open-metadata/OpenMetadata/blob/a6ac42371d8677ffd993f7b3f3f9f908400c4b50/ingestion/src/metadata/utils/time_utils.py#L146-L155

called here: https://github.com/open-metadata/OpenMetadata/blob/a6ac42371d8677ffd993f7b3f3f9f908400c4b50/ingestion/src/metadata/ingestion/source/pipeline/airbyte/metadata.py#L155

**Expected behavior**

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8]
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

PR #23784, opened 2025-10-08
fix: logger level should work for deprecation warnings

aballiet:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/23778

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


pmbrull:
hi @AntoineGlacet 👋 wouldn't we want deprecation warnings to be present to make sure we're not missing them and act on the fixes? We don't want end users to play with the SDK/connectors and end up with missing processes once we remove deprecated methods. 

Might be better for us to clean the actual airbyte function. wdyt?

aballiet:
> wouldn't we want deprecation warnings to be present to make sure we're not missing them and act on the fixes? We don't want end users to play with the SDK/connectors and end up with missing processes once we remove deprecated methods. 

@pmbrull I get your point but on the other side the logger is then not working as expected and may generate a lot of unnecessary logs. When passing a log level to errors, we agree to hide the warning so it's a risk the user is willing to take. Default level is INFO


> Might be better for us to clean the actual airbyte function. wdyt?

Ofc it would be nice too! I can do the PR

pmbrull (dismissed):

pmbrull:
> > wouldn't we want deprecation warnings to be present to make sure we're not missing them and act on the fixes? We don't want end users to play with the SDK/connectors and end up with missing processes once we remove deprecated methods.
> 
> @pmbrull I get your point but on the other side the logger is then not working as expected and may generate a lot of unnecessary logs. When passing a log level to errors, we agree to hide the warning so it's a risk the user is willing to take. Default level is INFO
> 
> > Might be better for us to clean the actual airbyte function. wdyt?
> 
> Ofc it would be nice too! I can do the PR

oki lets move with this then. We just went ahead with the initial one as a possible solution, but what you mention makes sense. If someone sticks at `ERROR` is "at your own peril". Understand it can clutter logs, and in this case for our internal connectors the work would be on us anyway and not the end user.

If you have the time to clean airbyte it'd be awesome, since you're a heavy user so if it works for your case I'm sure it'll tick all the boxes :)

Thanks!

PR #23785, opened 2025-10-08
fix(ui): updated default font size to 14px and grey color based on style guide

ShaileshParmar11:
-  Updated the default font size to 14px in the application
- Updated grey color based on the style guide

1. app.less - Sets the actual browser font size
```
html {
  font-size: 14px;
}
```
This sets the root font size for the entire HTML document
Affects all elements that use rem units (1rem = 14px instead of default 16px)
Applies to regular HTML elements, Ant Design components, and custom CSS

2. createMuiTheme.ts - Tells MUI how to calculate rem values
```
typography: {
  fontSize: 14,
  htmlFontSize: 14,
  // ...
}
```
htmlFontSize: 14 tells MUI's theme system what the root HTML font size is
MUI uses this value to calculate rem units correctly in its components
fontSize: 14 sets the base font size for typography variants
Without this, MUI would still assume 1rem = 16px (the default) and its components would appear larger than intended

#### Why both are required? 
If you only change app.less:
- Browser renders with 14px base
- But MUI still calculates assuming 16px
- Result: MUI components have incorrect sizing

If you only change createMuiTheme.ts:
- MUI calculates correctly for 14px
- But browser still uses 16px as the root
- Result: MUI calculations are wrong because the actual root font size doesn't match

Together: Both the actual font size and MUI's understanding of it are aligned, ensuring consistent 14px base across all components.


openmetadata-ui/src/main/resources/ui/src/styles/app.less
```diff
@@ -15,6 +15,10 @@
 
 /* Generic */
 
+html {
+  font-size: 14px;
+}
+
```
chirag-madlani: We don't need this

openmetadata-ui-core-components/src/main/resources/ui/src/colors/defaultColors.ts
```diff
@@ -64,8 +64,8 @@ export const defaultColors: ThemeColors = {
     725: '#363636', // custom shade for search icon
     750: '#404040', // custom shade
     800: '#252B37',
-    900: '#101828',
-    950: '#0C111D',
+    900: '#181D27',
+    950: '#0A0D12',
```
chirag-madlani, line 68: Source for this change?

openmetadata-ui/src/main/resources/ui/src/styles/app.less
ShaileshParmar11: Please check PR description i have added information

openmetadata-ui-core-components/src/main/resources/ui/src/colors/defaultColors.ts
ShaileshParmar11: https://www.figma.com/design/q2HveWEgd55xkItAppFSaO/Design-System?node-id=1023-36350&p=f&m=dev

openmetadata-ui/src/main/resources/ui/src/styles/app.less
chirag-madlani: AntD alread do the same, that's not needed can you check again?

ShaileshParmar11: Good point, let me check

ShaileshParmar11: push the changes

PR #23788, opened 2025-10-08
Fixes 23104: Showing validation error if user enter duplicate service name

### LINKED ISSUE #23104: Misleading error while creating a service connection with a name (say "abc") that already exists
**Describe the bug**

- Let's suppose there is an existing service (of any type) with service name as "test service". 

- Now if I create a new service connection (of any type) with same name i.e. "test service", I see an error in test-connection saying - "Cannot convert [LinkedHashMap]...", which is misleading. 

- As a user it's not possible for me to figure out that the issues is duplicate service name (in OMD or Collate)

**Expected behavior**
Ideally there should have been a validation error saying - "duplicate service name" or similar before even letting the user add config for the new service they are trying to setup. 

**Version:**
 - OpenMetadata version: 1.9.2

**Additional context**

Image

Rohit0301:
### Describe your changes:

Fixes #23104 

**Changes Made**
Added a validation check during service creation to prevent duplicate service names.
Displayed a clear and descriptive error message for duplicates.

https://github.com/user-attachments/assets/ed210e2c-0d17-4491-83b6-0d18d05fae06

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/utils/ServiceUtils.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/ServiceUtils.tsx
@@ -132,11 +132,11 @@ export const setServiceSchemaCount = (
   Promise.allSettled(promises)
     .then((results) => {
       let count = 0;
-      results.forEach((result) => {
+      for (const result of results) {
         if (result.status === PROMISE_STATE.FULFILLED) {
           count += result.value?.paging?.total || 0;
         }
-      });
+      }
       callback(count);
     })
     .catch((err: AxiosError) => showErrorToast(err));
@@ -153,11 +153,11 @@ export const setServiceTableCount = (
   Promise.allSettled(promises)
     .then((results) => {
       let count = 0;
-      results.forEach((result) => {
+      for (const result of results) {
         if (result.status === PROMISE_STATE.FULFILLED) {
           count += result.value?.paging?.total || 0;
         }
-      });
+      }
       callback(count);
     })
     .catch((err: AxiosError) => showErrorToast(err));
@@ -611,7 +611,6 @@ export const searchServiceByExactName = async (
       hits: { hits },
     } = await searchService({
       search: name,
-      wildcard: false,
       searchIndex: getSearchIndexFromService(serviceName),
     });
 
--- a/openmetadata-ui/src/main/resources/ui/src/rest/serviceAPI.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/rest/serviceAPI.ts
@@ -192,8 +192,6 @@ export const searchService = async ({
   currentPage?: number;
   queryFilter?: Record<string, unknown>;
   deleted?: boolean;
-  trackTotalHits?: boolean;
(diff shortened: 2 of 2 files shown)


openmetadata-ui/src/main/resources/ui/src/utils/ServiceUtils.tsx
```diff
@@ -582,3 +584,63 @@ export const getAddServiceEntityBreadcrumb = (
...
+export const searchServiceWithName = async (
+  name: string,
+  serviceName: string
+) => {
+  try {
+    const {
+      hits: { hits },
+    } = await searchService({
+      search: name,
+      limit: 10000,
```
aniketkatkar97: We should not be putting such large limits without any specific reason. Also can we use constants for such inputs?

openmetadata-ui/src/main/resources/ui/src/utils/ServiceUtils.tsx
```diff
@@ -582,3 +584,63 @@ export const getAddServiceEntityBreadcrumb = (
...
+      searchIndex: getSearchIndexFromService(serviceName),
+    });
+
+    const services = hits.map(
+      ({ _source }) => _source as DatabaseServiceSearchSource
+    );
```
aniketkatkar97, line 619: We are also iterating over all the hits, so if there are 10000 results as per the above limit, it this scalable?

openmetadata-ui/src/main/resources/ui/src/utils/ServiceUtils.tsx
aniketkatkar97: If we are searching for a specific name, should the limit be this big?

Rohit0301: Removed the limit

openmetadata-ui/src/main/resources/ui/src/utils/ServiceUtils.tsx
Rohit0301: Updated the search to perform an exact match. Since it’s an exact match, there will be at most one result returned

openmetadata-ui/src/main/resources/ui/src/utils/ServiceUtils.tsx
```diff
@@ -582,3 +584,66 @@ export const getAddServiceEntityBreadcrumb = (
...
+  const searchedServices = await searchServiceByExactName(name, serviceName);
+
+  const isServiceNamePresent = searchedServices.some(
+    (service) => service.name === name
+  );
+
+  if (isServiceNamePresent) {
+    return t('message.entity-already-exists', {
+      entity: t('label.name'),
+    });
```
aniketkatkar97, line 644: Instead of `Name already exists`, it should be `Service already exists`

aniketkatkar97 (approved):

PR #23841, opened 2025-10-10
Fixes #23840: Duplicate tags in ChangeDescription

### LINKED ISSUE #23840: Duplicate tags in ChangeEvent notifications for Entity tag updates
# **Affected module**

**Backend**

## **Describe the bug**

ChangeEvent notifications for table column tag updates report duplicate tags in the ChangeDescription. When updating tags on a table column, the `fieldsAdded` array
contains duplicate entries of the same tag, even though the entity itself stores only one instance.

## **To Reproduce**

1. Update tags on a table column (e.g., change from `PII.Sensitive` to `PII.None`)
2. Observe the ChangeDescription in the generated ChangeEvent
3. The `fieldsAdded` array shows duplicate tag entries:

```plaintext
fieldsAdded=[
  FieldChange[name=columns.xxx.tags, newValue=[
    {"tagFQN":"PII.None","name":"None",...},
    {"tagFQN":"PII.None","name":"None",...}  // duplicate
  ]]
]
```

## **Expected behavior**

Each tag should appear only once in the ChangeDescription `fieldsAdded` array, matching the actual entity state.

## **Root cause**

`EntityRepository.updateTags()` (line 4020-4067) pre-calculates `addedTags` and `deletedTags`, then passes these pre-populated lists to `recordListChange()`.
However, `recordListChange()` is designed to calculate diffs itself by adding to the provided lists, causing duplicates.
Other implementations like `GlossaryTermRepository` correctly pass empty lists to `recordListChange()`.

## **Version**
* **OpenMetadata version:** main branch

manerow:
### Describe your changes

Fixes #23840

Fixed duplicate tags appearing in ChangeEvent notifications when updating table column tags by passing empty lists to `recordListChange()` in
`EntityRepository.updateTags()`.

The bug occurred because `updateTags()` pre-calculated `addedTags` and `deletedTags` for database operations, then passed these pre-populated lists to
`recordListChange()`. However, `recordListChange()` is designed to calculate diffs itself by adding to the provided lists, causing duplicates.  
This fix aligns `EntityRepository` with the correct pattern used by `GlossaryTermRepository` and all other callers of `recordListChange()`.

Tested by debugging the ChangeEvent generation flow for column tag updates and verifying that tags now appear only once in the ChangeDescription.

---

### Type of change

- [x] Bug fix  
- [ ] Improvement  
- [ ] New feature  
- [ ] Documentation  

---

### Checklist

### DIFF
--- a/bootstrap/openmetadata-ops.sh
+++ b/bootstrap/openmetadata-ops.sh
@@ -54,4 +54,5 @@ else
 fi
 
 ${JAVA} -Dbootstrap.dir=$BOOTSTRAP_DIR  -cp ${CLASSPATH} ${OPENMETADATA_SETUP_MAIN_CLASS} -c $CONFIG_FILE_PATH "$@"
+exit $?
 
--- a/ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
@@ -418,14 +418,6 @@ class BigquerySource(LifeCycleQueryMixin, CommonDbSourceService, MultiDBSource):
         logic on how to handle table types, e.g., material views,...
         """
 
-        if self.incremental.enabled:
-            view_names = [
-                view_name
-                for view_name in view_names
-                if view_name
-                in self.incremental_table_processor.get_not_deleted(schema_name)
-            ]
-
         table_name_and_types = []
         for table_type, query in {
             TableType.View: BIGQUERY_GET_VIEW_NAMES,
--- a/ingestion/tests/unit/topology/database/test_bigquery.py
+++ b/ingestion/tests/unit/topology/database/test_bigquery.py
@@ -542,6 +542,80 @@ class BigqueryUnitTest(TestCase):
                 for either in self.bq_source.yield_table((table[0], table[1]))
             ]
 
+    def test_query_view_names_and_types_with_incremental(self):
+        """
+        Test query_view_names_and_types with incremental extraction enabled
+        """
+        self.bq_source.context.get().__dict__["database"] = MOCK_DB_NAME
+        self.bq_source.context.get().__dict__[
+            "database_schema"
+        ] = MOCK_DATABASE_SCHEMA.name.root
+
+        mock_view_results = [("view1",), ("view2",), ("view3",)]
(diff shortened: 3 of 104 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -4064,7 +4064,8 @@ protected void updateTags(
       }
 
       // Record changes for audit trail
-      recordListChange(fieldName, origTags, updatedTags, addedTags, deletedTags, tagLabelMatch);
+      recordListChange(
+          fieldName, origTags, updatedTags, new ArrayList<>(), new ArrayList<>(), tagLabelMatch);
```
Bhanuagr, line 4068: in `recordListChange` method impl, we can consider using set so there would be no duplicate so in any case no tags would be missed or duplicate, WDYT ?

manerow: Thanks for the review! I don't think Set deduplication is needed here. Here's why:

The root cause is architectural, not missing validation:
`updateTags()` computes diffs (lines 4020-4054) → applies DB changes → then passes those pre-populated lists to `recordListChange()`, which computes the diff again and adds to the same lists.

All other callers do it correctly:
- `updateOwners()`, `updateDomains()`, `updateToRelationships()`, etc. pass empty lists to `recordListChange()`
- Only `updateTags()` breaks this pattern

Why I would not not use Set?
- Masks the double-computation bug rather than fixing it. We give the chance to future callers to keep calling the method incorrectly.
- Adds unnecessary overhead (list→set→list conversion) on every entity update
- Makes the method API unclear: "Pass empty lists... but also we'll dedupe if you don't"

The fix:
```
  // Before: passes pre-populated lists (bug)
  recordListChange(fieldName, origTags, updatedTags, addedTags, deletedTags, tagLabelMatch);
```

```
  // After: passes empty lists (follows existing pattern)
  recordListChange(fieldName, origTags, updatedTags, new ArrayList<>(), new ArrayList<>(), tagLabelMatch);
```

This aligns `updateTags()` with how every other caller uses `recordListChange()`.

PR #23850, opened 2025-10-11
[DQaC] Simplified API

edg956:
## Summary

Implements a simplified API for running data quality tests as code against OpenMetadata tables.

## Changes by Commit

1. **Extend `metadata.sdk.configure` function**
   - Enhanced SDK configuration to support more flexible client setup

2. **Create convenience classes for existing `TestDefinition`s**
   - Added helper classes wrapping standard test definitions for easier programmatic use

3. **Added a configuration builder for `TestSuiteWorkflow` configurations**
   - Dynamically builds the configuration, just like it would for CLI runs
   - Decouples input (e.g: cli workflow, data quality checks code) from building the end config
 
4. **Implemented a results capturing processor**
   - Acts as a proxy class that captures the output of a `Processor`'s `_run()` for later usage

5. **Implement `TestRunner` interface to run test cases as code**
   - Added `TestRunner` class for defining and executing DQ tests
   - Internally uses the configuration builder
   - Instantiates `TestSuiteWorkflow` and decorates its processor (aka `TestCaseRunner`) to capture its output
   - Created example usage script demonstrating the SDK
   - Added comprehensive integration tests validating actual test execution against PostgreSQL
   - Includes support for table-level tests (row count, column count) and column-level tests (uniqueness, null checks, value ranges)

## Tests
- [X] Unit tests for `TestRunner`
- [X] Integration test running a few test cases

### DIFF
--- a/ingestion/src/metadata/sdk/examples/builder_end_to_end.py
+++ b/ingestion/src/metadata/sdk/examples/builder_end_to_end.py
@@ -10,6 +10,7 @@ Run:
 """
 from __future__ import annotations
 
+import logging
 from dataclasses import dataclass, field
 from typing import List, Optional
 
@@ -44,6 +45,8 @@ from metadata.sdk.entities.databaseschemas import DatabaseSchemas
 from metadata.sdk.entities.glossaries import Glossaries
 from metadata.sdk.entities.tables import Tables
 
+logger = logging.getLogger(__name__)
+
 # ------------------------
 # Example-local builders
 # ------------------------
@@ -51,6 +54,8 @@ from metadata.sdk.entities.tables import Tables
 
 @dataclass
 class DatabaseServiceBuilderPy:
+    """Builder for creating database service requests."""
+
     name_val: Optional[str] = None
     description_val: Optional[str] = None
     type_val: Optional[DatabaseServiceType] = None
@@ -71,6 +76,7 @@ class DatabaseServiceBuilderPy:
     def mysql_connection(
         self, host_port: str, username: str, database: Optional[str] = None
     ) -> "DatabaseServiceBuilderPy":
+        """Configure a MySQL connection for the database service."""
         conn = DatabaseConnection(
             config=MysqlConnection(
                 type=MySQLType.Mysql,
@@ -127,6 +133,8 @@ class DatabaseServiceBuilderPy:
 
 @dataclass
 class DatabaseBuilderPy:
+    """Builder for creating database requests."""
+
     name_val: Optional[str] = None
     description_val: Optional[str] = None
     service_fqn_val: Optional[str] = None
@@ -173,6 +181,8 @@ class DatabaseBuilderPy:
 
 @dataclass
 class SchemaBuilderPy:
+    """Builder for creating database schema requests."""
+
     name_val: Optional[str] = None
     description_val: Optional[str] = None
     database_fqn_val: Optional[str] = None
@@ -218,6 +228,8 @@ class SchemaBuilderPy:
 
 @dataclass
 class TableBuilderPy:
+    """Builder for creating table requests."""
+
     name_val: Optional[str] = None
     description_val: Optional[str] = None
     schema_fqn_val: Optional[str] = None
@@ -238,6 +250,7 @@ class TableBuilderPy:
     def add_column(
         self, name: str, dtype: ColumnDataType, *, length: Optional[int] = None
     ) -> "TableBuilderPy":
+        """Add a column to the table."""
         col = Column(
             name=ColumnName(name),
             displayName=None,
@@ -299,6 +312,7 @@ class TableBuilderPy:
 
 
 def main() -> None:
+    """Run the builder-style end-to-end example."""
     config = OpenMetadataConfig(
         server_url="http://localhost:8585",
         jwt_token="YOUR_JWT_OR_API_KEY",
@@ -380,7 +394,7 @@ def main() -> None:
     # apply
     _ = Glossaries.import_csv(glossary_name).with_data(csv_text).execute()
 
-    print("Completed builder-based example successfully.")
+    logger.info("Completed builder-based example successfully.")
 
 
 if __name__ == "__main__":
--- a/ingestion/src/metadata/sdk/examples/dq_as_code_example.py
+++ b/ingestion/src/metadata/sdk/examples/dq_as_code_example.py
@@ -1,0 +1,67 @@
+"""
+Example: Data Quality as Code with OpenMetadata SDK
+
+This example demonstrates how to run data quality tests programmatically
+using the simplified DQ as Code API.
+"""
+
+# pyright: reportUnusedCallResult=false
+# pylint: disable=W5001
+from metadata.sdk import configure
+from metadata.sdk.data_quality import (
+    ColumnValuesToBeNotNull,
+    ColumnValuesToBeUnique,
+    TableColumnCountToBeBetween,
+    TableRowCountToBeBetween,
+    TestRunner,
+)
+
+# Configure SDK connection
+configure(
+    host="http://localhost:8585/api",
+    jwt_token="your-jwt-token-here",
+)
+
+# Initialize test runner for a specific table
+runner = TestRunner.for_table("MySQL.default.openmetadata_db.bot_entity")
+
+# Add multiple tests
+runner.add_test(
+    TableColumnCountToBeBetween(min_count=10).with_description(
+        "Ensure bot_entity table has at least 10 columns"
+    )
+)
+
+runner.add_test(
+    TableRowCountToBeBetween(min_count=1, max_count=1000).with_description(
+        "Check bot_entity has reasonable number of rows"
+    )
+)
+
+runner.add_test(
+    ColumnValuesToBeUnique(column="id")
+    .with_name("bot_entity_id_unique")
+    .with_compute_row_count(True)
+)
+
+runner.add_test(ColumnValuesToBeNotNull(column="name"))
+
+# Execute all tests
+print("Running data quality tests...")
+results = runner.run()
+
+# Process results
+for result in results:
+    test_case = result.testCase
+    test_result = result.testCaseResult
+
+    print(f"\nTest: {test_case.name.root}")
+    print(f"Status: {test_result.testCaseStatus}")
+    print(f"Result: {test_result.result}")
+
+    if test_result.passedRows is not None:
+        print(f"Passed Rows: {test_result.passedRows}")
+        print(f"Failed Rows: {test_result.failedRows}")
+        print(f"Pass Rate: {test_result.passedRowsPercentage:.2f}%")
+
+print(f"\n✅ All tests completed! {len(results)} tests executed.")


harshach:
@edg956 can we add the same for Java sdk as well. Doesn't need to be in this PR but parity of the APIs would be great

edg956:
> @edg956 can we add the same for Java sdk as well. Doesn't need to be in this PR but parity of the APIs would be great

Makes sense. I'll work on it

ingestion/src/metadata/sdk/data_quality/tests/base_tests.py
```diff
@@ -0,0 +1,190 @@
...
+            TestCaseDefinition instance
+        """
+        return TestCaseDefinition(
+            testDefinitionName=self.test_definition_name,
+            name=self.name or self.test_definition_name,
+            description=self.description,
+            displayName=self.display_name,
+            parameterValues=self.parameters,
+            computePassedFailedRowCount=self.compute_passed_failed_row_count,
+        )
```
TeddyCr, line 123: ~~Depending on what we want to do here we should use either `CreateTestCase` or `TestCase` from the generated schema package.~~

Are you using TestCaseDefinition as you are passing them to the workflowConfig when executing the tests?

BTW, using the workflow config will automatically pull the existing test cases defined against that entity in OM - if I remember correctly. This is something we should explicitly mention.

ingestion/src/metadata/sdk/data_quality/tests/base_tests.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        return TestCaseDefinition(
+            testDefinitionName=self.test_definition_name,
+            name=self.name or self.test_definition_name,
+            columnName=self.column_name,
+            description=self.description,
+            displayName=self.display_name,
+            parameterValues=self.parameters,
+            computePassedFailedRowCount=self.compute_passed_failed_row_count,
+        )
```
TeddyCr, line 188: Same comment than in BaseTest class

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+    def for_service(cls, table_fqn: str) -> Self:
+        """Alias for for_table(). Initialize runner for a specific table FQN.
+
+        Args:
+            table_fqn: Fully qualified name of the table (e.g., "MySQL.default.db.table")
+
+        Returns:
+            TestRunner instance
+        """
+        return cls.for_table(table_fqn)
```
TeddyCr: What's the difference between for_service and for_table? And why can we pass the client in for_table, but not for_service?

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+        with open(file_path, "r", encoding="utf-8") as stream:
+            return cls.from_yaml_string(stream.read(), use_connection_from_yaml, client)
+
+    @classmethod
+    def from_yaml_string(
+        cls,
+        string: str,
+        use_connection_from_yaml: bool = False,
+        client: Optional[OMeta[Any, Any]] = None,
+    ) -> Self:
```
TeddyCr: I think these 2 should be combined. The only deviation is one first need to read a file and not the other, but we should be able to hide this logic and just have a `from_yaml`. Wdyt?

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+        """Fetch table entity and service connection from OpenMetadata.
+
+        This method retrieves the table entity and associated database service connection
+        from OpenMetadata. It validates that the table exists and has a properly configured
+        service connection.
+
+        Raises:
+            ValueError: If table not found, service not found, or connection not configured
+        """
+        self.table_entity = self.tables.get_by_name(
```
TeddyCr: [nit] can we rename `self.tables` and the like to `self.table_client` (so to have clarity that this is a client and not holding actual table objects.

TeddyCr: then `self.table_entity` can be `self.table` -- since holds an actual entity Object

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+                "service",
+                "database",
+            ],
+        )
+
+        if self.table_entity is None:
+            raise ValueError(
+                f"Table '{self.table_fqn}' not found in OpenMetadata. "
+                + "Please verify the FQN is correct (format: service.database.schema.table)"  # noqa: E501
+            )
```
TeddyCr: we could expand the message as the client may not be correct. Not sure about that statement, but at least from the backend POV, it will throw an error:
e.g. 404 if the fqn is not correct, 401 if unauthorized, etc. if this is propagated from `self.tables.get_by_name`, then let's just propagate this back

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+        assert (
+            self.table_entity.service is not None
+        ), f"Service not found for table '{self.table_fqn}'"
```
TeddyCr: Let's check the schema definition of Table, but I believe service might be required

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+
+        service_id = self.table_entity.service.id
+        service = self.services.get_by_id(DatabaseService, service_id)
```
TeddyCr: [nit] let's rename it to `self.service_client`

TeddyCr: same comment than the table as well on checking if the HTTP error are propagated correctly from the Python client

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+            raise ValueError(f"Service '{service_id}' not found in OpenMetadata")
+
+        if service.connection is None:
+            raise ValueError(
+                f"Service '{service_id}' does not have a connection configured"
+            )
+
+        self.service_connection = service.connection
+
+    def add_test(self, test_definition: BaseTest) -> Self:
```
TeddyCr: If we think about that operation as something similar to python `append` let's not return the object.

e.g.
```
l = [1,2,3]
l.append(4) // this does not return the `l` object
```

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+        Args:
+            test_definition: Test definition instance (e.g., TableColumnCountToBeBetween)
+
+        Returns:
+            Self for method chaining
+        """
+        self.test_definitions.append(test_definition.to_test_case_definition())
+        return self
+
+    def add_tests(self, *test_definitions: BaseTest) -> Self:
```
TeddyCr: Same comment than above

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+            >>> runner.add_tests(
+            ...     TableRowCountToBeBetween(min_count=100),
+            ...     ColumnValuesToBeNotNull(column="user_id")
+            ... )
+        """
+        for test_definition in test_definitions:
+            _ = self.add_test(test_definition)
+        return self
+
+    def _build_test_case(self, test_def: TestCaseDefinition) -> TestCase:
```
TeddyCr: I don't think this will be needed. We should directly work with `CreateTestCase` or `TestCase` depending on what we want to do.

If we'll need to handle creating the testCase in OM, we'll need the `CreateTestCase`  object.

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+            TestCase object ready for execution
+        """
+        column_name = test_def.columnName
+        entity_link_str = (
+            entity_link.get_entity_link(  # pyright: ignore[reportUnknownMemberType]
+                Table,
+                fqn=self.table_fqn,
+                column_name=column_name,
+            )
+        )
```
TeddyCr: Have you checked what happens if column_name=null? Do we get the entity link as expected i.e. `<E#::table::my.table.fqn>` or does it returns something like `<E#::table::my.table.fqn::columns::None>`

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+
+        test_definition = self.tests.get_by_name(
```
TeddyCr: same thing on the client naming and propagating the errors from the client

ingestion/src/metadata/sdk/data_quality/runner.py
edg956: Actually I wanted to remove this method but I forgot about it

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,439 @@
...
+            description=test_def.description,
+            fullyQualifiedName=f"{self.table_fqn}.{test_def.name}",
+            testDefinition=test_definition_ref,
+            entityLink=EntityLink(entity_link_str),
+            testSuite=test_suite_ref,
+            parameterValues=test_def.parameterValues
+            if test_def.parameterValues
+            else None,
+            computePassedFailedRowCount=test_def.computePassedFailedRowCount,
+        )
```
TeddyCr: ~~I think we need to handle a few scenarios:~~
~~1. The test case exists already in OM - then we need to simply retrieve it from there. at least we should have some sort of check to make sure we don't create duplicate~~
~~2. the test case does not exist but there is already a test suite in OM, in this case we should probably get that test suite and not create a new one~~
~~3. There is no test suite in OM for our "TestRunner" execution, then we should create it and use it when creating our test cases~~

Actually if you are using the workflow config, from what I understand, we should allow users to specify the `forceUpdate` parameters. This will update the test case if it exist or create it if it does not

```
class TestSuiteProcessorConfig(ConfigModel):
    """class for the processor config"""

    testCases: Optional[List[TestCaseDefinition]] = None
    forceUpdate: Optional[bool] = False
```

ingestion/src/metadata/sdk/data_quality/runner.py
edg956: I copied this pattern from soda, but no strong opinion about keeping both. You suggest supporting only reading from a file, right? That sounds good for me

ingestion/src/metadata/sdk/data_quality/runner.py
edg956: I will look into it. If no information is given, I'll rewrite this to make it clear that the error might be due to any of those situations

ingestion/src/metadata/sdk/data_quality/runner.py
edg956: It's optional

image

It's probably safe to assume it's going to come populated, but pyright complains

ingestion/src/metadata/sdk/data_quality/runner.py
edg956: It returns it as expected 
image

ingestion/src/metadata/sdk/config.py
```diff
@@ -43,6 +44,43 @@ def builder(cls) -> "OpenMetadataConfigBuilder":
...
+        )
+        if not server_url:
+            raise ValueError(
+                "Server URL must be provided via 'OPENMETADATA_HOST' or "
+                + "'OPENMETADATA_SERVER_URL' environment variable"
+            )
+
+        jwt_token = os.environ.get("OPENMETADATA_JWT_TOKEN") or os.environ.get(
+            "OPENMETADATA_API_KEY"
+        )
```
TeddyCr, line 69: why use 2 different env var and not just 1?

ingestion/src/metadata/sdk/data_quality/runner.py
TeddyCr: I meant let's support both, but combine them into 1 method.

ingestion/src/metadata/sdk/data_quality/runner.py
TeddyCr: Let's just add `# type: ignore` so we don't add un necessary lines.

ingestion/src/metadata/sdk/config.py
edg956: Since we're extending an already existing interface that used both, I thought it was sensible to add the possibility to use both

edg956:
@TeddyCr 

In reply to:

> Depending on what we want to do here we should use either CreateTestCase or TestCase from the generated schema package.

and

> I don't think this will be needed. We should directly work with `CreateTestCase` or `TestCase` depending on what we want to do.
> 
> If we'll need to handle creating the testCase in OM, we'll need the `CreateTestCase` object.

I have not really added an integration with OM to create tests cases. At most we're checking is that the test definition exists in OM.

The more I think about it the more I believe we can reuse the TestSuitePipeline workflow so that it automatically creates tests in OM and what not. In a sense, it'd be like running `metadata test -c ...` on a test suite built in code.

Maybe it's easier to reuse all that logic and have these two interfaces behave similarly. That requires refactoring, but then it also simplifies the wrapper.

What do you think?

edg956:
@TeddyCr 

> If I understand correctly, you are using the workflow config to execute the test cases, so you'll be adding the testCases the user adds via `add_test` to the workflow config and then execute it?

Yes, though the workflow config is rather "dummy". Based on my previous response, perhaps the better way forward is to have this facade as a wrapper around calling and running the test ingestion workflow. Then, syncing results in OM would be "free" because it's already implemented.

On second thought, I'm not sure if we can capture errors correctly

ingestion/src/metadata/sdk/data_quality/tests/base_tests.py
edg956: > Are you using TestCaseDefinition as you are passing them to the workflowConfig when executing the tests?

I'm not passing them to workflowConfig, I'm building these directly and calling them in line 435 `result = runner.run_and_handle(test_case)`.

So the code never takes the tests cases from `processor.config`

> BTW, using the workflow config will automatically pull the existing test cases defined against that entity in OM - if I remember correctly. This is something we should explicitly mention.

This is something I'm not doing. Pulling those tests cases happens in the `TestSuiteWorkflow` itself via the `TestCaseRunner`, which I am not using

TeddyCr: Ok makes sense. I think we can go with `TestCase` object directly then, if I am not mistaking.

ingestion/src/metadata/sdk/config.py
TeddyCr: Make sense.

ingestion/src/metadata/sdk/data_quality/tests/base_tests.py
edg956: This method was just a convenience for dealing with code-based and yaml-based test cases uniformly.

I think it's better to deal with test cases internally. That way, if we decide to actually create them in OM, we could use `CreateTestCase` internally without having to update this method

edg956:
@TeddyCr looking at the comments you've made regarding the usage of `TestCase` and `CreateTestCase` I think you were expecting test cases to be automatically created when running code, is that right?

So running `TestRunner` should behave somewhat similar to `TestCaseRunner` when it fetches test cases like in this picture

image

I started this feature by the simplest API where `TestCase`s are not created in OM, that's why I'm building them in `_build_test_case`, but you say I should not need that method and deal directly with TestCases? If so, then by first creating them in OM?

I think building them in the runner before actually performing the validation is a good way to separate inputs (`TestCaseDefinition` from yaml-based config or `BaseTest`-based configs) from the implementation details (isolated test cases or test cases that are first created from OM)

Does it make sense for you to keep this separation? Then the interface remains relatively simple and we can decide how to deal with `TestCase`s internally

TeddyCr:
> @TeddyCr looking at the comments you've made regarding the usage of `TestCase` and `CreateTestCase` I think you were expecting test cases to be automatically created when running code, is that right?
> 
> So running `TestRunner` should behave somewhat similar to `TestCaseRunner` when it fetches test cases like in this picture
> 
> image
> I started this feature by the simplest API where `TestCase`s are not created in OM, that's why I'm building them in `_build_test_case`, but you say I should not need that method and deal directly with TestCases? If so, then by first creating them in OM?
> 
> I think building them in the runner before actually performing the validation is a good way to separate inputs (`TestCaseDefinition` from yaml-based config or `BaseTest`-based configs) from the implementation details (isolated test cases or test cases that are first created from OM)
> 
> Does it make sense for you to keep this separation? Then the interface remains relatively simple and we can decide how to deal with `TestCase`s internally

Sounds good. 👍

ingestion/src/metadata/sdk/data_quality/runner.py
```diff
@@ -0,0 +1,264 @@
...
+
+        config = self.config_builder.build()
+
+        workflow = TestSuiteWorkflow.create(  # pyright: ignore[reportUnknownMemberType]
+            config.model_dump()  # pyright: ignore[reportUnknownMemberType]
+        )
+
+        original_processor = workflow.steps[0]
+        result_capturer = ResultCapturingProcessor(original_processor)
+        workflow.steps = (result_capturer,) + workflow.steps[1:]
```
edg956, line 260: This suggestion from Claude is probably the part I'm least happy with, but I also did not want to make any drastic change to `TestSuiteWorkflow` or parent classes.

I also did not want to subclass it for this very specific case, though it might just be cleaner.

I added types on `workflow.steps` so the dependency here is explicit and type safe. If we ever change the workflow's steps, then `pyright` should point us at places where the change breaks.

Open for suggestions here

TeddyCr:
> @TeddyCr
> 
> In reply to:
> 
> > Depending on what we want to do here we should use either CreateTestCase or TestCase from the generated schema package.
> and
> 

> > I don't think this will be needed. We should directly work with CreateTestCase or TestCase depending on what we want 


> to do.
> 
> If we'll need to handle creating the testCase in OM, we'll need the CreateTestCase object.
> I have not really added an integration with OM to create tests cases. At most we're checking is that the test definition exists in OM.
> 
> The more I think about it the more I believe we can reuse the TestSuitePipeline workflow so that it automatically creates tests in OM and what not. In a sense, it'd be like running metadata test -c ... on a test suite built in code.
> 
> Maybe it's easier to reuse all that logic and have these two interfaces behave similarly. That requires refactoring, but then it also simplifies the wrapper.
> 
> What do you think?

if the user wants to push the results back to OM automatically these tests will need to exists. I think reusing it would make sense so that we keep it aligned. We would still simplify the API for a developer but we maintain some consistency.

edg956:
> > @TeddyCr
> > In reply to:
> > > Depending on what we want to do here we should use either CreateTestCase or TestCase from the generated schema package.
> > > and
> 
> > > I don't think this will be needed. We should directly work with CreateTestCase or TestCase depending on what we want
> 
> > to do.
> > If we'll need to handle creating the testCase in OM, we'll need the CreateTestCase object.
> > I have not really added an integration with OM to create tests cases. At most we're checking is that the test definition exists in OM.
> > The more I think about it the more I believe we can reuse the TestSuitePipeline workflow so that it automatically creates tests in OM and what not. In a sense, it'd be like running metadata test -c ... on a test suite built in code.
> > Maybe it's easier to reuse all that logic and have these two interfaces behave similarly. That requires refactoring, but then it also simplifies the wrapper.
> > What do you think?
> 
> if the user wants to push the results back to OM automatically these tests will need to exists. I think reusing it would make sense so that we keep it aligned. We would still simplify the API for a developer but we maintain some consistency.

Thank you. I think after reworking this PR I achieve just that

ingestion/src/metadata/sdk/data_quality/runner.py
TeddyCr: @edg956 it is ok to me to have it like that. Maybe add a comment as to why we made that tradeoff so we remember

PR #23856, opened 2025-10-13
 Add assets API and deprecate inline assets field for Domain and Dataproduct

### LINKED ISSUE #21091: Data discrepancy between frontend and SDK extraction

sonika-shah:
### Describe your changes:

Fixes #21091

Summary of changes-
 New API Endpoints (GET) for fetching assets from search (paginated result - max cap on 1000)
   - /api/v1/domains/{id}/assets
   - /api/v1/dataProducts/{id}/assets
    - /api/v1/glossaryTerms/{id}/assets
    - /api/v1/tags/{id}/assets
    - /api/v1/teams/{id}/assets
    - /api/v1/users/{id}/assets
    
    
   - /api/v1/domains/{name}/assets
  - /api/v1/dataProducts/{name}/assets
    - /api/v1/glossaryTerms/{id name/assets
    - /api/v1/tags/{name}/assets
    - /api/v1/teams/{name}/assets
    - /api/v1/users/{name}/assets
    
  2.  Marked assets field as deprecated in domain and dataProduct/creatDataProduct schemas
   - Domain:
     Previously:  we used assets field to fetch assets using  fetchers during GET operations
     Now: removed field fetchers and fetchAssets() method - assets are only retrieved via dedicated /assets endpoint
    
    
 -   DataProduct:
     Previously: assets could be added/removed via PUT/PATCH on the assets field in the dataProduct entity
    Now: Removed all asset update logic from PUT/PATCH operations, assets must now be added/removed  through the bulk assets API only(/dataProducts/{id}/assets) - to keep the dataProduct behavioru consistent with domain (both use dedicated endpoints instead of inline field updates)

  

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/SearchIndexApp.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/SearchIndexApp.java
@@ -17,7 +17,6 @@ import static org.openmetadata.service.workflows.searchIndex.ReindexingUtil.isDa
 import com.fasterxml.jackson.core.type.TypeReference;
 import jakarta.ws.rs.core.Response;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -69,8 +68,6 @@ import org.openmetadata.service.search.RecreateIndexHandler.ReindexContext;
 import org.openmetadata.service.search.SearchClusterMetrics;
 import org.openmetadata.service.search.SearchRepository;
 import org.openmetadata.service.socket.WebSocketManager;
-import org.openmetadata.service.util.EntityUtil;
-import org.openmetadata.service.util.EntityUtil.Fields;
 import org.openmetadata.service.util.FullyQualifiedName;
 import org.openmetadata.service.util.RestUtil;
 import org.openmetadata.service.workflows.interfaces.Source;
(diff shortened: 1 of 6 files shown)


ingestion/src/metadata/ingestion/ometa/mixins/tag_glossary_mixin.py
```diff
@@ -0,0 +1,54 @@
...
+
+        Args:
+            fqn: Fully qualified name of the tag
+            limit: Maximum number of assets to return (default 10, max 1000)
+            offset: Offset from which to start returning results (default 0)
+
+        Returns:
+            API response as a dictionary containing paginated assets
+        """
+        path = f"/tags/name/{fqn}/assets"
```
harshsoni2024: can we implement exception handling here & on other methods also in case of api failures?
```
try:
    path = f"/tags/name/{fqn}/assets"
    params = {"limit": limit, "offset": offset}
    return self.client.get(path, params)
except Exception as exc:
    logger.debug(f"API failure for getting tag assets::{exc}")
return {}
 ```

sonika-shah: thanks @harshsoni2024 i will check

harshsoni2024 (approved):

Bhanuagr (dismissed):

aniketkatkar97 (dismissed):

PR #23864, opened 2025-10-13
remove unwanted semantic rule and supported more playwright test

Ashish8689:
### Describe your changes:

- Removed Service and Name rule from the semantic, since they weren't be used by someone as validation will be around that respective entity.
- Fix DataProducts validation not working.
- Added Playwright test for `Version Rule` and `DataProducts`
- Remove the `is_null` and `is_not_null` condition from the Tags, Tier and Glossary. Since they all are pointing to only one source i.e `Tag` and the user will always be mislead around the operation they will put in for this rule.

#
### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
@@ -1158,6 +1158,8 @@ test.describe('Data Contracts', () => {
   test('Semantic with Not_Contains Operator should work for Tier, Tag and Glossary', async ({
     page,
   }) => {
+    test.slow(true);
+
     const table = new TableClass();
     const testClassification = new ClassificationClass();
     const testTag = new TagClass({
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/src/components/DataContract/ContractSemanticFormTab/ContractSemanticFormTab.tsx
```diff
@@ -203,9 +203,7 @@ export const ContractSemanticFormTab: React.FC<{
           <Form.List name="semantics">
             {(fields, { add }) => {
               // Store the add function so it can be used outside
-              if (!addFunctionRef.current) {
-                addFunctionRef.current = add;
-              }
+              addFunctionRef.current ??= add;
```
chirag-madlani, line 206: You can add null check here to fix code smell instead removing conditino completly

PR #23892, opened 2025-10-14
Improve Tableau logging

SumanMaharana:
### Describe your changes:

Improve Tableau logging

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/tableau/client.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/tableau/client.py
@@ -332,7 +332,7 @@ class TableauClient:
                 )
             )
             if datasources_graphql_result and datasources_graphql_result.get("data"):
-                if datasources_graphql_result["data"]["workbooks"]:
+                if datasources_graphql_result["data"].get("workbooks"):
                     tableau_datasource_connection = TableauDatasourcesConnection(
                         **datasources_graphql_result["data"]["workbooks"][0]
                     )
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/ingestion/source/dashboard/tableau/client.py
```diff
@@ -331,14 +331,18 @@ def _query_datasources(
...
+                if datasources_graphql_result["data"]["workbooks"]:
                     tableau_datasource_connection = TableauDatasourcesConnection(
                         **datasources_graphql_result["data"]["workbooks"][0]
                     )
                     return tableau_datasource_connection.embeddedDatasourcesConnection
+                else:
+                    logger.warning(
+                        f"No Datasources found in GraphQL datasources query result for the workbook {dashboard_id}. "
+                        "If this is a recently created or updated workbook, it may take some time "
+                        "to become available for querying via the GraphQL API."
```
keshavmohta09: Should we log the data present in `datasources_graphql_result["data"]` so that we can inspect it when workbooks are not present? This might help us identify if there's another key we can use as a fallback later.

Also, should we handle a possible keyerror by using `datasources_graphql_result.get["data"].get("workbooks")` instead?

SumanMaharana: we are using graphql query to query that data so we are sure that these cases doesnt happen
the workbooks will always be there if we have data

SumanMaharana: the data inside `datasources_graphql_result["data"]` will be empty 

```
"data": {
    "workbooks": []
  }
```
but we can sure log the `datasources_graphql_result` here

keshavmohta09 (approved):

PR #23903, opened 2025-10-15
Proper error message when bearer token is empty

aji-aju:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/security/JwtFilter.java
```diff
@@ -259,6 +259,9 @@ protected static String extractToken(MultivaluedMap<String, String> headers) {
     }
     // Extract the bearer token
     if (source.startsWith(TOKEN_PREFIX)) {
+      if (source.length() <= TOKEN_PREFIX.length() + 1) {
```
Bhanuagr: This code block is duplicate, can we create one function to extract token and handle failure inside that function to avoid duplicate code
same code is using below also (ref: [github](https://github.com/open-metadata/OpenMetadata/pull/23903/files#diff-56ec8fe549c6bd81ca57a3c166cc75cf05d928b06a7a6d9d93630fd5fc55271bR277-R279))

aji-aju: Done

Issue #23925, opened 2025-10-16
Postgres docker compose does not work turn-key


LewisCowlesMotive:
**Is some content missing, wrong or not clear?**
Basically the docs do not seem to mention that postgres will require exclusive access to port 5432. Within the container, the port should not matter, but per-host, this can cause issues bringing up the container.

Additionally, the volume mount for postgres doesn't follow the same conventions as the other stateful services. I am uncertain why, but making it follow those conventions fixes on my setup and is internally consistent with the rest of the docker-compose file for postgres.

**Describe the solution you'd like**
I'm making a PR so you'll see it. It's just fixing the docker-compose file to bind to 5439 on the users host. A Note should be added but there was no clear "contribute to docs" on the docs, so I have settled for the file recommended.
https://github.com/open-metadata/OpenMetadata/pull/23926

**Additional context**

Without these edits this will not run on my machine, which would mean I would be unable to evaluate OpenMetadata

PR #23926, opened 2025-10-16
Fixes 23925: docs - Update docker-compose-postgres.yml

LewisCowlesMotive:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/23925

I worked on changing the default port, and volume mechanism for postgres because the container would not start on my machine.

By using a non-standard postgres port, this is more likely to run on others machines.
By following the covention used elsewhere within the compose file, the volume permissions are no longer an issue.

### Type of change:

- [X] Documentation

### Checklist:

- [X] I have commented on my code, particularly in hard-to-understand areas. 

### Documentation:

I have considered use-cases outside of my own and remain open to feedback and contributing wider changes around this proposed fix

### DIFF
--- a/docker/docker-compose-quickstart/docker-compose-postgres.yml
+++ b/docker/docker-compose-quickstart/docker-compose-postgres.yml
@@ -15,6 +15,7 @@ volumes:
   ingestion-volume-dags:
   ingestion-volume-tmp:
   es-data:
+  db-data-postgres:
 services:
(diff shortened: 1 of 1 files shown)


TeddyCr:
Thanks for contribution to the project. We are going to keep the default port as is and let users change them if it conflicts with services running on their machine

PR #23931, opened 2025-10-16
Fixes #23930: LDAP connection pool bind

### LINKED ISSUE #23930: LDAP Connection Pool doesn't revert to auth user
**Affected module**
Backend

**Describe the bug**
When connecting to OpenMetadata server via LDAP, an authorizing DN and password with elevated LDAP privileges is specified in the config, and the connection pool is initialized with this. However, when a user logs in, the whole connection pool rebinds as the user, who may not have elevated privileges. When this happens, all subsequent logins fail with the (incorrect) error "Invalid username/password".

**To Reproduce**

Assume an LDAP server with 3 users:
- cn=directory manager (elevated privileges)
- uid=alice,dc=example,dc=org (elevated privileges)
- uid=bob,dc=example,dc=org

Example OpenMetadata server env:
```env
AUTHENTICATION_PROVIDER=ldap
AUTHENTICATION_LDAP_HOST=ldap.example.org
AUTHENTICATION_LDAP_PORT=636
AUTHENTICATION_LOOKUP_ADMIN_DN='cn=directory manager'
AUTHENTICATION_LOOKUP_ADMIN_PWD=SuperSecretPassword!
AUTHENTICATION_USER_MAIL_ATTR=mail
AUTHENTICATION_USER_LOOKUP_BASEDN=dc=example,dc=org
AUTHENTICATION_LDAP_SSL_ENABLED=true
AUTHENTICATION_LDAP_TRUSTSTORE_TYPE=TrustAll
AUTHENTICATION_LDAP_ALLOWED_HOSTNAMES=[ldap.example.org]
```

Scenario 1: Alice logs in first
1. Log in as Alice using her LDAP credentials. **This should be successful.**
2. Log in as Bob using his LDAP credentials. **This should be successful.**
3. Log in as either Alice or Bob using LDAP credentials. **This should fail.**

Scenario 2: Bob logs in first
1. Log in as Bob using his LDAP credentials. **This should be successful.**
2. Log in as either Alice or Bob using LDAP credentials. **This should fail.**

**Expected behavior**
In both of the above scenarios, Alice and Bob should both be able to log in over any span of time in any order (provided their LDAP credentials are still valid).

**Version:**
 - OS: OSX 15.5 and Ubuntu 22.04 LTS
 - Python version: 3.10
 - OpenMetadata version: 1.8, 1.9

**Additional context**
I have a PR coming shortly.

bwinston-sdp:
### Describe your changes:

Fixes #23930: LDAP connection pool bind issue

I changed the method call from `bind(username, pw)` to `bindAndRevertAuthentication(username, pw)`, to keep the connection with the authorized DN.

#
### Type of change:

- [X] Bug fix

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LdapAuthenticator.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/security/auth/LdapAuthenticator.java
@@ -200,7 +200,7 @@ public class LdapAuthenticator implements AuthenticatorHandler {
     // performed in LDAP , the storedUser's name set as DN of the User in Ldap
     BindResult bindingResult = null;
     try {
-      bindingResult = ldapLookupConnectionPool.bind(userDn, reqPassword);
+      bindingResult = ldapLookupConnectionPool.bindAndRevertAuthentication(userDn, reqPassword);
       if (Objects.equals(bindingResult.getResultCode().getName(), ResultCode.SUCCESS.getName())) {
         return;
       }


bwinston-sdp:
It looks like I may be late to the party, and the error may have been addressed last week by #23690, though that PR doesn't use the LDAP connection pool for authenticating the user at all, instead creating a new connection whenever validating a password. I don't think it's necessarily a bad strategy to spin up a new connection for each user auth, but it seems strange to me to explicitly set up a connection pool (and offer the configuration of the pool size to the end user) and then only use it for _some_ of the LDAP integration -- my assumption would be that the pool handled everything. IMO it should be all or nothing, but that's just my 2p. Would love to hear reasons why it shouldn't be that way!

bwinston-sdp:
@aji-aju wondering if you have any thoughts here, or if I should just close this.

mohityadav766:
@aji-aju does this PR makes sense?

TeddyCr:
closing as fixed in https://github.com/open-metadata/OpenMetadata/pull/23690

bwinston-sdp:
I'm fine with this being closed, but I don't know that I agree that #23690 "fixed" the issue -- I'll point out [my previous comment](https://github.com/open-metadata/OpenMetadata/pull/23931#issuecomment-3416759801) again.

PR #24001, opened 2025-10-24
Clean up recognizers migration data

edg956:
### Describe your changes:

This PR cleans up a pattern-breaking change I had introduced previously by placing some json data in a wrong directory.

Now, not only is the data coming from the expected place (`openmetadata-service.src.main.resources.json.data`), but it's the same source we use as seed for creating tags with recognizers.

#
### Type of change:
- [X] Improvement

#
### Checklist:

- [X] I have tested in both MySql and Postgres that the migrations do indeed happen

### DIFF
--- a/openmetadata-service/src/main/resources/json/data/tags/piiTagsWithRecognizers.json
+++ b/openmetadata-service/src/main/resources/json/data/tags/piiTagsWithRecognizers.json
@@ -409,7 +409,7 @@
           {
             "name": "AuAcnRecognizer",
             "displayName": "Au Acn Recognizer",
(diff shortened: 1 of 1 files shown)


edg956:
> https://github.com/open-metadata/OpenMetadata/blob/main/bootstrap/sql/migrations/native/1.11.0/postgres/postDataMigrationSQLScript.sql

Because the recognizers definition for `PII.Sensitive` is so long that it breaks the record keeping. That's why I copied the behaviour of the migration but truncated the query here:

```java
      migrationDAO.upsertServerMigrationSQL(
          migrationFile.version, truncatedQuery, hash(truncatedQuery));
```

Issue #24006, opened 2025-10-24
Implement a simple API for validating DataFrames


edg956:
**Feature**
This includes a facade for running tests in python from code and returning the results. For simplicity of this task, implement only short circuit mode

PR #24009, opened 2025-10-24
Simplified API for validating DataFrames

edg956:
### Describe your changes:

Implements #24006 

This PR includes a facade to validate pandas DataFrames with a similar API as #23850, using a short-circuit execution mode

#### Commit by commit
- Refactored old tests to accommodate new ones
- Added models and a method to publish results to OM
- Added an engine for running tests
- Implemented the facade to validate dataframes
- Added examples

#
### Type of change:
- [X] New feature

#
### Checklist:

- [X] The issue properly describes why the new feature is needed, what's the goal, and how we are building it. Any discussion

### DIFF
--- a/ingestion/src/metadata/sdk/data_quality/dataframes/validation_results.py
+++ b/ingestion/src/metadata/sdk/data_quality/dataframes/validation_results.py
@@ -26,7 +26,7 @@ from metadata.sdk.data_quality.dataframes.models import MockTestCase
 from metadata.utils.entity_link import (
     get_entity_link,  # pyright: ignore[reportUnknownVariableType]
 )
-from metadata.utils.entity_link import maybe_get_column_from
+from metadata.utils.entity_link import get_column_name_or_none
 
 logger = logging.getLogger(__name__)
 
@@ -103,7 +103,7 @@ class ValidationResult(BaseModel):
                     entity_link=get_entity_link(
                         Table,
                         table_fqn,
-                        column_name=maybe_get_column_from(test_case.entityLink.root),
+                        column_name=get_column_name_or_none(test_case.entityLink.root),
                     ),
                     test_definition_fqn=test_case.testDefinition.fullyQualifiedName,
                     test_case_parameter_values=test_case.parameterValues,
--- a/ingestion/src/metadata/utils/entity_link.py
+++ b/ingestion/src/metadata/utils/entity_link.py
@@ -106,7 +106,7 @@ def get_table_or_column_fqn(entity_link: str) -> str:
     )
 
 
-def maybe_get_column_from(entity_link: str) -> Optional[str]:
+def get_column_name_or_none(entity_link: str) -> Optional[str]:
     """It attempts to get a column from an entity link
 
     Args:


ingestion/src/metadata/data_quality/validations/column/pandas/__init__.py
TeddyCr, line 1: We designed the data_quality module as a namespace package so users can implement their own tests logic in a separate namespace package and be able to run it through the UI. Would this create some issues? More details here https://medium.com/openmetadata/leveraging-the-power-of-openmetadata-data-quality-framework-385ba2d8eaf

ingestion/src/metadata/sdk/data_quality/dataframes/__init__.py
TeddyCr, line 1: Let's not use wildcard imports

ingestion/src/metadata/data_quality/validations/table/pandas/__init__.py
TeddyCr, line 1: same comment than above

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validation_engine.py
```diff
@@ -0,0 +1,146 @@
...
+            )
+        )
+
+        return TestValidationResult(
+            test_name=test_definition.name,
+            test_type=test_definition.testDefinitionName,
+            status=TestCaseStatus.Aborted,
+            passed_rows=0,
+            failed_rows=0,
+            total_rows=0,
```
TeddyCr: We should pass the actual failure / pass count as potentially a user could pass chunks of data to the validator

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validation_engine.py
```diff
@@ -0,0 +1,146 @@
...
+            test_case_result=TestCaseResult(  # pyright: ignore[reportCallIssue]
+                timestamp=int(time.time() * 1000),
+                testCaseStatus=TestCaseStatus.Aborted,
+                result=f"Test execution failed: {str(exception)}",
+            ),
+        )
+
+    def _build_validation_result(
+        self, test_results: List[TestValidationResult], execution_time_ms: float
+    ) -> ValidationResult:
```
TeddyCr, line 102: Why are we not using the TestCasResult object and implementing a new ValidationResultObject?

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validator.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def add_tests(self, *tests: BaseTest) -> None:
+        """Add multiple test definitions at once.
+
+        Args:
+            *tests: Variable number of test definitions
+        """
+        self._test_definitions.extend(tests)
+
+    def validate(
```
TeddyCr, line 107: I believe we discuss allowing users to pass `on_success` and `on_failure` callback functions

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validator.py
```diff
@@ -0,0 +1,85 @@
...
+        Args:
+            df: DataFrame to validate
+            mode: Validation mode (`FailureMode.ShortCircuit` stops on first failure)
+
+        Returns:
+            ValidationResult with outcomes for all tests
+        """
+        from metadata.sdk.data_quality.dataframes.dataframe_validation_engine import (
+            DataFrameValidationEngine,
+        )
```
TeddyCr: Any reason to not have this as a top level import

ingestion/src/metadata/sdk/data_quality/dataframes/validation_results.py
```diff
@@ -0,0 +1,91 @@
...
+
+"""DataFrame validation result models."""
+
+from dataclasses import dataclass
+from typing import List, Optional
+
+from metadata.generated.schema.tests.basic import TestCaseResult, TestCaseStatus
+
+
+@dataclass
```
TeddyCr: Let;s not use data class. We should use pedantic. Why not use the existing TestCaseResult object?

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validation_engine.py
edg956: I was thinking of the report error failure mode and add error indices, but I think I will remove this because of what we discussed about validators not reporting row-level errors.

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validator.py
```diff
@@ -0,0 +1,85 @@
...
+from typing import List
+
+from pandas import DataFrame
+
+from metadata.sdk.data_quality.dataframes.validation_results import ValidationResult
+from metadata.sdk.data_quality.tests.base_tests import BaseTest
+
+
+class FailureMode(Enum):
+    ShortCircuit = "short-circuit"
```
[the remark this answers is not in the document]
edg956: We will support other modes in the future

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validator.py
```diff
@@ -0,0 +1,85 @@
...
+
+        Args:
+            *tests: Variable number of test definitions
+        """
+        self._test_definitions.extend(tests)
+
+    def validate(
+        self,
+        df: DataFrame,
+        mode: FailureMode = FailureMode.ShortCircuit,
```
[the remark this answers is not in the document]
edg956: I'll fix the test

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validator.py
edg956: I should've cleaned this up. I was getting some circular import errors

ingestion/src/metadata/sdk/data_quality/dataframes/validation_results.py
edg956: Will remove this class altogether

ingestion/src/metadata/data_quality/validations/column/pandas/__init__.py
edg956: Good point. This might break that

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validator.py
```diff
@@ -0,0 +1,159 @@
...
+        Args:
+            data: An iterable of pandas DataFrames
+            on_success: Callback to execute after successful validation
+            on_failure: Callback to execute after failed validation
+            mode: Validation mode (`FailureMode.ShortCircuit` stops on first failure)
+
+        Returns:
+            None
+        """
+        self._check_full_table_tests_included()
```
edg956, line 185: Using this here because, like the warning above explains, some tests need the whole DataFrame to return a valid response. This however means we need to have the whole dataframe in memory, for which the `on_success`/`on_failure` interface may be less clear.

I decided to let application code be explicit when validation is to be run in chunks (e.g: in cases when data source is too big for memory), for which the callbacks make a nice UX.

See tests for the expected behavior of this warning

ingestion/src/metadata/ingestion/ometa/mixins/tests_mixin.py
```diff
@@ -379,3 +407,13 @@ def ingest_inspection_query(
...
         return TestCase(**resp)
+
+    def delete_test_case(self, test_case_fqn: str) -> None:
+        """Delete a test case
+        Args:
+            test_case_fqn: Fully qualified name of the test case to delete
+        """
+        url = f"{self.get_suffix(TestCase)}/name/{quote(test_case_fqn)}"
+        url += f"?recursive=true"
+        url += f"&hardDelete=true"
```
TeddyCr: Let's make `hardDelete` an argument to the function, we can default to `false` I would say to be on the safe side

ingestion/src/metadata/sdk/data_quality/dataframes/models.py
```diff
@@ -0,0 +1,44 @@
+from uuid import uuid4
+
+from metadata.generated.schema.tests.testCase import TestCase
+from metadata.generated.schema.type.entityReference import EntityReference
+from metadata.sdk.data_quality import BaseTest, ColumnTest
+
+
+class MockTestCase(TestCase):
```
TeddyCr, line 8: Why do we have a MockTestCase and not use TestCase directly?

ingestion/src/metadata/sdk/data_quality/dataframes/open_metadata.py
TeddyCr, line 1: should we keep this close to the ometa client in the sdk? and maybe rename it to something more explicit or make it part of the client

ingestion/src/metadata/sdk/data_quality/dataframes/validation_results.py
```diff
@@ -0,0 +1,212 @@
...
+            for result in self.test_results
+            if result.testCaseStatus == TestCaseStatus.Success
+        ]
+
+    @property
+    def test_results(self) -> List[TestCaseResult]:
+        """Get all test results."""
+        return [result for _, result in self.test_cases_and_results]
+
+    def publish_to_openmetadata(
```
TeddyCr: let's not name it openmetadata as Collate users may use it as well

ingestion/src/metadata/sdk/data_quality/dataframes/validators.py
TeddyCr, line 1: it is ok to keep it like that for now, but let's think about how we can make so that if I add a new tests then it gets picked up here automatically (i.e. I don't need to manually think about adding it here)

ingestion/src/metadata/sdk/data_quality/dataframes/models.py
edg956: Easier recognition of test cases mocked because defined in code instead of being retrieved from OM

ingestion/src/metadata/sdk/data_quality/dataframes/open_metadata.py
edg956: I considered that, but I wasn't sure if it was worth extending the API like that considering it's really tied to ValidationResults.

On second thought, it might just be a good idea to have it sit there

ingestion/src/metadata/sdk/data_quality/dataframes/validation_results.py
edg956: Do we have a standard way to refer to it? Or is it something like "publish" just enough?

ingestion/src/metadata/sdk/data_quality/dataframes/validators.py
edg956: Agreed. I didn't want to address this here, but I think using something like a Metaclass for automatic registering of Pandas subclasses could do. Just an idea

ingestion/src/metadata/sdk/data_quality/dataframes/open_metadata.py
edg956: That'd require to import `ValidationResult` in the client. I'll move this to the `ValidationResult` class so it's explicit about it

ingestion/src/metadata/ingestion/ometa/mixins/tests_mixin.py
```diff
@@ -379,3 +408,25 @@ def ingest_inspection_query(
             data=inspection_query,
         )
         return TestCase(**resp)
+
+    def delete_test_case(
+        self,
+        test_case_fqn: str,
+        recursive: bool = False,
```
TeddyCr: recursive should be `true` by default though if you check my previous comment -- only hard should be `false`

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validation_engine.py
TeddyCr: will you be using the TestCaseResult class or going with the ValidationResult one?

ingestion/src/metadata/sdk/data_quality/dataframes/models.py
TeddyCr: 🤔 does it matter though ?

ingestion/src/metadata/sdk/data_quality/dataframes/dataframe_validation_engine.py
edg956: The validation result will contain `TestCaseResult`s as they come from the test case runs. Now the code looks a little different and `TestValidationResult` does not exist

ingestion/src/metadata/utils/entity_link.py
```diff
@@ -106,6 +106,21 @@ def get_table_or_column_fqn(entity_link: str) -> str:
     )
 
 
+def maybe_get_column_from(entity_link: str) -> Optional[str]:
```
TeddyCr: nit: get_column_from_or_none might be more explicit

ingestion/src/metadata/ingestion/ometa/mixins/tests_mixin.py
edg956: My mistake!

ingestion/src/metadata/sdk/data_quality/dataframes/models.py
edg956: Well it helps us recognize which test cases we need to create in OpenMetadata to push results for later. We use it [here](https://github.com/open-metadata/OpenMetadata/blob/3985d27aeec2b3250fce516cd24bb1956bdbc45d/ingestion/src/metadata/sdk/data_quality/dataframes/validation_results.py#L100).

Alternatives are: use a sentinel value in some field to recognize those instances, which I'm not 100% sure I like, or always call `get_or_create`, in which case we'd still need some logic to make sure we override things like `entityLink` only for test cases we created from code.

Because of the delayed sync with OM (users can decide to push results after validation has run), those test cases know nothing of the table they refer to in OM, so we mock values like `<E::table::dataframe_validator::column::foo>`.

Thoughts?

PR #24038, opened 2025-10-28
Fix: Search Slowness when painless scripts aggregates for terms and classifications

harshach:
…

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-spec/src/main/resources/elasticsearch/en/security_service_index_mapping.json
+++ b/openmetadata-spec/src/main/resources/elasticsearch/en/security_service_index_mapping.json
@@ -249,6 +249,14 @@
               }
             }
           },
+          "classificationTags": {
+            "type": "keyword",
+            "normalizer": "lowercase_normalizer"
+          },
+          "glossaryTags": {
+            "type": "keyword",
+            "normalizer": "lowercase_normalizer"
+          },
           "tags": {
             "properties": {
               "tagFQN": {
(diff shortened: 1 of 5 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/drives/SpreadsheetResource.java.backup
Bhanuagr, line 1: why are we pushing backup file also ?

openmetadata-spec/src/main/resources/elasticsearch/en/security_service_index_mapping.json
```diff
@@ -573,4 +573,4 @@
           }
         }
     }
-}
\ No newline at end of file
+}
```
Bhanuagr, line 584: Here classification and glossary tags config not added, is it intentional or missed ?

mohityadav766 (dismissed):

Bhanuagr (commented):
It looks like definitely this change would improve performance but as per thread, there was only performance issue observed in OS for ES working fine.  Changes added would improve perf for both cases but not sure if that is the reason for slowness observed only for OS

openmetadata-service/src/main/java/org/openmetadata/service/resources/drives/SpreadsheetResource.java.backup
mohityadav766: it is removed, it was there in previous 
This PR removed it

openmetadata-spec/src/main/resources/elasticsearch/en/security_service_index_mapping.json
mohityadav766: addressed

PR #24048, opened 2025-10-28
Minor: Handle empty buckets in GCS connection tests

ayush-shah:
This pull request improves the robustness of the blob listing and retrieval logic in the `list_blobs` and `get_blob` methods for GCS connections. The main change is to handle empty buckets gracefully and avoid errors when attempting to retrieve blobs from buckets with no contents.

Improvements to blob listing and retrieval:

* Updated `list_blobs` to catch `StopIteration` when a bucket is empty, logging a debug message to confirm that list permissions are valid even if no blobs are found.
* Added a check in `get_blob` to ensure that blob retrieval is only attempted if a blob was found during the listing phase, preventing errors for empty buckets.


ulixius9 (approved):

ayush-shah:
Pytest failure not related to PR, merging it in

PR #24093, opened 2025-10-30
Fixes 22920: Enhance connection class import handling

ayush-shah:
This pull request addresses a critical bug in the `get_connection_class` function related to case sensitivity in module imports for certain database services, most notably SAS, SQLite, and SSAS. The fix implements a robust fallback mechanism that first attempts to import the connection class using the standard camelCase convention and, if that fails, retries with an all-lowercase module name. This ensures compatibility across different filesystems (case-sensitive and case-insensitive) and resolves issues encountered on Linux systems. The update is thoroughly validated with comprehensive unit tests covering standard cases, edge cases, and performance.

**Bug Fix: Case-Sensitive Module Import Handling**
- Enhanced `get_connection_class` in `parser.py` to automatically fall back to all-lowercase module names if the standard camelCase import fails, fixing issues for services like SAS, SQLite, and SSAS on case-sensitive filesystems.

**Testing: Comprehensive Validation of Import Logic**
- Added a new test suite `test_parser_connection_fallback.py` with detailed tests and documentation to verify both standard and fallback import paths, including edge cases and performance validation.
- Introduced `test_parser_connection_class.py` with unit tests that check the correct resolution of connection classes for a wide range of service types, including those affected by the bug.
- Updated `test_workflow_parse.py` to explicitly test all-uppercase source types (SAS, SSAS), ensuring the fallback mechanism works as expected in integrated workflows.


ayush-shah:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/open-metadata/OpenMetadata/pull/24093#pullrequestreview-3398637909)

PR #24126, opened 2025-11-03
Add bulk apis to improve ingestion performance

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/distributed/argo_orchestrator.py
```diff
@@ -0,0 +1,313 @@
...
+            tasks.append(
+                {
+                    "name": f"discover-{task_name}",
+                    "template": "discover",
+                    "arguments": {
+                        "parameters": [{"name": "entity-type", "value": entity_type}]
+                    },
+                }
+            )
+
```
TeddyCr, line 123: What do you mean here 1 discovery task per entity type? Would a workflow have only 1 entity type?

ingestion/src/metadata/distributed/argo_orchestrator.py
```diff
@@ -0,0 +1,313 @@
...
+        if hasattr(config, "model_dump"):
+            # Pydantic v2 model
+            config_dict = config.model_dump(mode="json", by_alias=True)
+        elif hasattr(config, "dict"):
+            # Pydantic v1 model or dict with nested models
+            config_dict = json.loads(
+                json.dumps(
+                    config, default=lambda o: o.dict() if hasattr(o, "dict") else str(o)
+                )
+            )
```
TeddyCr, line 265: We don't use Pydantic V1 so we can remove this.

ingestion/src/metadata/distributed/local_executor.py
```diff
@@ -0,0 +1,286 @@
...
+        logger.info("=" * 80)
+        logger.info("LOCAL DISTRIBUTED EXECUTION SUMMARY")
+        logger.info("=" * 80)
+        logger.info(f"Total entities: {total_entities}")
+        logger.info(f"Successful: {successful}")
+        logger.info(f"Failed: {failed}")
+        logger.info(f"Success rate: {(successful/total_entities)*100:.1f}%")
+        logger.info(f"Duration: {duration:.1f} seconds")
+        logger.info(f"Throughput: {total_entities/duration:.1f} entities/second")
+        logger.info("=" * 80)
```
TeddyCr, line 162: We should probably see if we can reuse the existing logging logic  with the `Status()` object we use for example in https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/api/step.py across the new implementation

ingestion/src/metadata/distributed_cmd/discover.py
TeddyCr, line 1: I believe you will need to the command entry point here https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/pyproject.toml#L36-L37

ingestion/src/metadata/examples/workflows/redshift.yaml
TeddyCr, line 1: Any reasons to delete this? I believe it is used in the logs to provide a working example of a config file when a user config is not working.

ingestion/src/metadata/ingestion/api/distributed.py
```diff
@@ -0,0 +1,233 @@
...
+    @abstractmethod
+    def get_parallelizable_entity_types(self) -> List[str]:
+        """
+        Return entity types that can be processed in parallel.
+
+        Returns:
+            List of entity type names
+
+        Examples:
+            - Database source: ["table", "stored_procedure", "database_schema"]
```
TeddyCr, line 146: Can parralelize this way ? For example I can't process tables before I processed the schema the tables belong to, same thing with charts, etc.

ingestion/src/metadata/ingestion/ometa/mixins/table_mixin.py
```diff
@@ -335,3 +336,40 @@ def create_or_update_custom_metric(
...
+            use_async (bool): Use backend async processing (default: True)
+
+        Returns:
+            BulkOperationResult: Result containing success/failure details
+        """
+        import json
+
+        from metadata.generated.schema.type.bulkOperationResult import (
+            BulkOperationResult,
+        )
```
TeddyCr, line 356: Can we move the generic imports back to top level?

ingestion/src/metadata/ingestion/sink/metadata_rest.py
```diff
@@ -123,6 +131,34 @@ def __init__(self, config: MetadataRestSinkConfig, metadata: OpenMetadata):
...
+            )
+
+            # Timing statistics
+            self.bulk_api_total_time = 0.0
+            self.bulk_api_call_count = 0
+            self.bulk_api_total_tables = 0
+
+            # Start worker threads
+            for i in range(self.config.async_pipeline_workers):
+                worker = threading.Thread(
```
TeddyCr, line 156: We should use ThreadPoolExecutor to manage threading here (instead of Thread) with a context manager so we don't need to worry about clean up. 

Also, out of curiosity, instead of having the thread share a common object and locking it when a thread reads from it (if I understand correctly), could we just dispatch the entity to create directly to the threads so we don't have a central blocking object?

ingestion/src/metadata/ingestion/sink/metadata_rest.py
```diff
@@ -748,7 +863,115 @@ def write_pipeline_usage(self, pipeline_usage: PipelineUsage) -> Either[Pipeline
...
+            logger.info("BULK API PERFORMANCE METRICS")
+            logger.info("=" * 70)
+            logger.info(f"Total bulk API calls: {self.bulk_api_call_count}")
+            logger.info(f"Total tables processed: {self.bulk_api_total_tables}")
+            logger.info(
+                f"Total time in bulk API: {self.bulk_api_total_time:.2f}s ({self.bulk_api_total_time/60:.2f}m)"
+            )
+            logger.info(f"Average time per bulk call: {avg_time_per_call:.2f}s")
+            logger.info(f"Average throughput: {avg_tables_per_sec:.1f} tables/sec")
+            logger.info("=" * 70)
```
TeddyCr, line 977: same comment here regarding the logging.

ingestion/tests/unit/sdk/test_test_case_entity.py.bak
TeddyCr, line 1: why are we deleting some integration and unit tests?

TeddyCr:
Regarding the distribution logic, it might be good to have it in a different IMO. It will make the bulk upload easier to cherry / bring in a new release and test that we have not impacted anything.

Also regarding the distributed logic, maybe this is the idea and I missed it, but should the python only responsible for:
1. finding out how many pods to spin up based on how many entities we have
2. modify the ingestion config at runtime (i.e. filtering based on the entities we are going to process )
3. spin up `n` argo workflow pods at runtime with the patch ingestion config?

PR #24173, opened 2025-11-05
Feat(UI): Notification template UI

aniketkatkar97:
Worked on the UI changes needed for Template UI addition in https://github.com/open-metadata/openmetadata-collate/pull/2328

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/locale/languages/de-de.json
+++ b/openmetadata-ui/src/main/resources/ui/src/locale/languages/de-de.json
@@ -2764,6 +2764,8 @@
     "re-indexing-error": "Fehler beim Re-Indexieren!",
     "re-indexing-started": "Re-Indexierung gestartet",
     "re-indexing-stopped": "Re-Indexierung gestoppt",
+    "reset-entity-error": "Fehler beim Zurücksetzen von {{entity}}!",
+    "reset-entity-success": "{{entity}} erfolgreich zurückgesetzt.",
     "reset-password-success": "Passwort erfolgreich zurückgesetzt!",
     "restore-entity-success": "{{entity}} erfolgreich wiederhergestellt.",
     "stop-re-indexing-error": "Beim Versuch, die Re-Indexierung zu stoppen, ist ein Fehler aufgetreten!",
--- a/openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
+++ b/openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
@@ -2764,6 +2764,8 @@
     "re-indexing-error": "Error while re indexing!",
     "re-indexing-started": "Re-index started",
     "re-indexing-stopped": "Re-index stopped",
+    "reset-entity-error": "Error while resetting {{entity}}!",
+    "reset-entity-success": "{{entity}} reset successfully.",
     "reset-password-success": "Password reset successfully!",
     "restore-entity-success": "{{entity}} restored successfully.",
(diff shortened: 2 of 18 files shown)


openmetadata-ui/src/main/resources/ui/src/components/BlockEditor/BubbleMenu/BubbleMenu.tsx
```diff
@@ -141,7 +141,11 @@ const BubbleMenu: FC<BubbleMenuProps> = ({ editor, toggleLink }) => {
                 'is-format-active': isActive(),
               })}
               type="text"
-              onClick={command}>
+              onMouseDown={(e) => {
```
karanh37, line 144: any reason for using mousedown ?

openmetadata-ui/src/main/resources/ui/src/components/BlockEditor/block-editor.less
```diff
@@ -322,6 +322,33 @@
...
+// Handlebars highlighting
+.om-handlebars {
+  background-color: #f0f7ff;
+  color: #0066cc;
+  padding: 2px 4px;
+  border-radius: 3px;
+  font-family: 'Monaco', 'Courier New', monospace;
+  font-size: 0.9em;
+  font-weight: 500;
+  border: 1px solid #b3d9ff;
```
karanh37: theme colors ?

openmetadata-ui/src/main/resources/ui/src/pages/AddNotificationPage/AddNotificationPage.tsx
```diff
@@ -203,6 +217,39 @@ const AddNotificationPage = () => {
...
+    []
+  );
+
+  const extraFormButtons = useMemo(
+    () => alertsClassBase.getAddAlertFormExtraButtons(),
+    []
+  );
+
+  const fetchTemplates = useCallback(async () => {
+    setLoadingCount((count) => count + 1);
```
karanh37: why using loading count as number and not boolean ?

karanh37 (changes requested):

openmetadata-ui/src/main/resources/ui/src/components/BlockEditor/BubbleMenu/BubbleMenu.tsx
aniketkatkar97: Yes, the mouseDown event fires before the click event and before focus changes, so we can intercept it and prevent the default focus behaviour. I'll add the same explanation in the comment.

openmetadata-ui/src/main/resources/ui/src/pages/AddNotificationPage/AddNotificationPage.tsx
aniketkatkar97: We have multiple API calls being made on the page, using a single boolean state will result in flaky behaviour of loader if any one of the call is finished first but others are still going. Also adding multiple states is also not scalable hence using the count strategy. We are doing this in other places as well.

karanh37: can we not use promise all ?

aniketkatkar97: Promise all will be useful if we have all calls in one place. It's not the same here. Let me see how we can improve this.

PR #24300, opened 2025-11-12
Fix search bar typing race conditions

### LINKED ISSUE #23830: Bug: User List Page Search Input Loses Characters During Typing
**Affected module**
Does it impact the UI, backend or Ingestion Framework? UI

**Describe the bug**

The search input on the User List page has a poor user experience where typed characters disappear during fast typing or when the component is in a loading state. This makes it difficult for users to search for users/admins effectively.

**To Reproduce**
1) Navigate to Settings > Users (or Admins)
2) Start typing in the search input field
3) Type quickly or type while the component is loading
4) Observed: Letters get removed from the input field

https://github.com/user-attachments/assets/ef313a61-f70d-477f-b647-6daf5e574220

**Expected behavior**
- The search input should maintain all typed characters smoothly during typing
- Fast typing should be handled gracefully with debouncing
- The input should not reset during loading states or component re-renders
- The search should feel responsive and natural

**Version:**
 - Python version:

**Additional context**

harsh-vador:
### Describe your changes:

Fixes #23830 

Fixed character scrambling during fast typing in SearchBar component and stale search results in UserListPageV1.

Root Cause:
- Lodash debounce creates closures that capture the search text parameter value
- Multiple debounced calls queued during fast typing would execute with stale captured values
- API responses from earlier requests could overwrite results from later requests

Solution
- SearchBar Component:
  - Use ref (`searchTextRef`) to store current search value synchronously
  - Debounced callback reads from ref instead of closure parameter
  - All queued debounced calls read the latest value from the same ref

**Before**: `debounceOnSearch(searchText)` → closure captures stale value
**After**: `debounceOnSearch()` → reads `searchTextRef.current` (always latest)

- UserListPageV1
  - Implement request ID tracking using `latestSearchIdRef`
  - Each search request gets unique incremental ID
  - Only process results if request ID matches latest ID
  - Prevents stale responses from displaying incorrect data

## Changes
- `SearchBar.component.tsx`: Simplified to use single ref for current search text
- `UserListPageV1.tsx`: Added request ID tracking to prevent race conditions

https://github.com/user-attachments/assets/98df380c-b66d-4857-9da5-8d1585498315

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/src/pages/UserListPage/UserListPageV1.tsx
```diff
@@ -450,8 +460,7 @@ const UserListPageV1 = () => {
         type: t('label.user'),
       })}...`,
       searchValue: searchValue,
-      typingInterval: 500,
-      urlSearchKey: 'user',
+      typingInterval: 350,
```
ShaileshParmar11, line 461: how we come up with this number? any specific calculation?

harsh-vador: 350ms provides a better balance - fast enough to feel responsive, slow enough to avoid excessive API calls

chirag-madlani (approved):

PR #24337, opened 2025-11-13
Improvements to accuracy of auto-classification

### LINKED ISSUE #24296: Identify and correct the amount false positives in the Auto Classification process
**Affected module**
Ingestion

**Describe the bug**
We've received reports from users that the classifier is incorrectly labelling some columns with the wrong entities or PII level

**To Reproduce**

We can't really reproduce yet because we don't have access to data, we're supposed to catch them now

**Expected behavior**

Columns should be labelled according to their data without tags that are obviously wrong

**Version:**
 - OS: running on our docker images
 - Python version: 3.9, 3.10
 - OpenMetadata version: reported in >1.9, could affect earlier versions
 - OpenMetadata Ingestion package version: >1.9.x, could affect earlier versions

**Additional context**
For collate employees, ask me for the internal communications for more context

edg956:
### Describe your changes:

Fixes #24296 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/__init__.py
```diff
@@ -11,6 +11,8 @@
 """
 OpenMetadata package initialization.
 """
+__path__ = __import__("pkgutil").extend_path(__path__, __name__)
```
edg956: I shall remove it before merging, as it is not my intention to formally introduce this

pmbrull (approved):

PR #24340, opened 2025-11-13
Fixes 404 error when loading dashboards or other entities that reference deleted glossary terms or tags

Bhanuagr:
## Root Cause

  The error occurs during tag fetching in TagLabelUtil:

  1. addDerivedTags() method - When fetching derived/related tags for glossary terms, if a referenced term doesn't exist, EntityNotFoundException is thrown
  2. populateTagLabel() method - When batch-fetching glossary term entities, if any term is missing, the entire operation fails

These exceptions propagate up through the call stack and cause the entire getByName API to fail.

## Solution

Added comprehensive error handling in TagLabelUtil.java to gracefully skip missing glossary terms:

  - addDerivedTags(): Wrapped getDerivedTags() call in try-catch to log warning and continue processing other tags
  - populateTagLabel(): Separated tag and term fetching into individual try-catch blocks to isolate failures
  - applyTagCommonFields(): Gracefully handles missing tags/glossary terms during field enrichment
  - Missing glossary terms are now logged and gracefully skipped
  - Dashboard loads successfully with all other details intact
  

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/Entity.java
```diff
@@ -787,7 +787,7 @@ public static <T extends FieldInterface> void populateEntityFieldTags(
         if (columnTag == null) {
           c.setTags(new ArrayList<>());
         } else {
-          c.setTags(addDerivedTags(columnTag));
+          c.setTags(addDerivedTagsGracefully(columnTag));
```
mohityadav766, line 790: let's use the same name

Bhanuagr: what do you mean by same name ? function name need to be different if same parameters list to pass, right ?

mohityadav766: addDerivedTags we are updating all the occurance 
we can still call that addDerivedTags  instaed of addDerivedTagsGracefully, it's a minor nit nothing on functionality "Gracefully" can be removed i meant

PR #24352, opened 2025-11-14
Fixes #24349: Fix email configuration templates default value from 'collate' to 'openmetadata'

### LINKED ISSUE #24349: Email Template Resolution Failure Caused by Wrong Default (templates="collate") in v1.9.x
A configuration bug affected OSS users who deployed between **v1.9.x and v1.10.x**. The issue was originally raised in an OM Slack Support thread.

### **The Issue**

Some users ended up with `$.templates = "collate"` in their `openmetadata_settings` table (`configType = "emailConfiguration"`).
This caused email template resolution failures — for example, testing the SMTP connection resulted in:

> **"Email template not found with name: testMail"**

The system was incorrectly looking for templates in the non-existent `emailTemplates/collate/` directory, while the OSS build only includes `emailTemplates/openmetadata/`.
*(This was before NotificationTemplates existed.)*

### **Root Cause**

Two commits bracket the issue:

* **[a50cddf665](https://github.com/open-metadata/OpenMetadata/commit/a50cddf665544b0934ce79a539aecbb00bd541c8)** (May 12, 2025)
  Introduced `conf/operations.yaml` with the **wrong default**:
  `templates: ${TEMPLATES:-"collate"}` ❌

* **[079343acfd](https://github.com/open-metadata/OpenMetadata/commit/079343acfd993b0d66d632dde7fbb74bd6eeee68)** (July 8, 2025)
  Corrected it to:
  `templates: ${TEMPLATES:-"openmetadata"}` ✅

### **Who's Affected**

Users who deployed **v1.9.x between May and July 2025**.
Because this was a misconfigured default—not a migration—existing DB records were never automatically updated, so affected deployments still contain the incorrect `"collate"` value even after upgrading.

manerow:
Fixes #24349

This PR fixes the incorrect default `emailConfiguration.templates` value for OSS users deployed between **v1.8.0 and v1.8.2**. Previously, deployments in the May–July 2025 window would set `templates` to `"collate"` instead of `"openmetadata"`, causing email template resolution failures.

I worked on creating a migration to ensure the default value is always `"openmetadata"` when creating new `emailConfiguration` records. This ensures fresh affected configurations use the correct templates directory.

The change was tested by deploying a fresh OpenMetadata instance with a cleared `openmetadata_settings` table and verifying that `emailConfiguration.templates` is correctly set to `"openmetadata"` and email template lookups succeed.

### Type of change:

* [x] Bug fix
* [ ] Improvement
* [ ] New feature
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ] Documentation

### Checklist:

* [x] I have read the [**[CONTRIBUTING](https://docs.open-metadata.org/developers/contribute)**](https://docs.open-metadata.org/developers/contribute) document.
* [x ] My PR title is `Fixes <issue-number>: <short explanation>`
* [x] I have commented on my code, particularly in hard-to-understand areas.
* [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

### DIFF
--- a/bootstrap/sql/migrations/native/1.10.8/mysql/postDataMigrationSQLScript.sql
+++ b/bootstrap/sql/migrations/native/1.10.8/mysql/postDataMigrationSQLScript.sql
@@ -1,0 +1,6 @@
+-- Fix emailConfiguration templates field for users affected by v1.8.0 bug
+-- See commit: https://github.com/open-metadata/OpenMetadata/commit/a50cddf665544b0934ce79a539aecbb00bd541c8
+-- This migration ensures all emailConfiguration records have the correct 'openmetadata' value
+UPDATE openmetadata_settings
+SET json = JSON_SET(json, '$.templates', 'openmetadata')
+WHERE configType = 'emailConfiguration';
--- a/bootstrap/sql/migrations/native/1.10.8/postgres/postDataMigrationSQLScript.sql
+++ b/bootstrap/sql/migrations/native/1.10.8/postgres/postDataMigrationSQLScript.sql
@@ -1,0 +1,6 @@
+-- Fix emailConfiguration templates field for users affected by v1.8.0 bug
+-- See commit: https://github.com/open-metadata/OpenMetadata/commit/a50cddf665544b0934ce79a539aecbb00bd541c8
+-- This migration ensures all emailConfiguration records have the correct 'openmetadata' value
+UPDATE openmetadata_settings
+SET json = jsonb_set(json, '{templates}', '"openmetadata"')
+WHERE configType = 'emailConfiguration';


bootstrap/sql/migrations/native/1.10.7/mysql/schemaChanges.sql
```diff
@@ -6,4 +6,11 @@ TRUNCATE TABLE ACT_HI_VARINST;
 TRUNCATE TABLE ACT_HI_DETAIL;
 TRUNCATE TABLE ACT_HI_COMMENT;
 TRUNCATE TABLE ACT_HI_ATTACHMENT;
-TRUNCATE TABLE ACT_HI_IDENTITYLINK;
```
aji-aju, line 9: Can you verify whether this migration will cause any issue for collate users?

manerow: Thanks for pointing out. I've created [this PR](https://github.com/open-metadata/openmetadata-collate/pull/2349/files) in collate to fix potential issue.

PR #24415, opened 2025-11-18
Fix auto classification tests

edg956:
### Describe your changes:

- This changes a model that's more apt for recognizing entities in [multilanguage settings](https://spacy.io/models/xx), which is useful for picking up sensitive data like people's names in non-western contexts.
- It also improves `NhsRecognizer` to discard matches when the context points to a timestamp column
- Added test cases

#
### Type of change
- [X] Bug fix

#
### Checklist:
- [X] I have added a test that covers the exact scenario we are fixing. For complex issues, comment the issue number in the test for future reference.

### DIFF
--- a/ingestion/src/metadata/pii/algorithms/presidio_utils.py
+++ b/ingestion/src/metadata/pii/algorithms/presidio_utils.py
@@ -13,9 +13,10 @@ Utilities for working with the Presidio Library.
 """
 import inspect
 import logging
-from typing import Any, Callable, Iterable, List, Optional, Type, Union, cast
+from typing import Any, Callable, Iterable, List, Optional, Set, Type, Union, cast
 
 import spacy
+from dateutil import parser
 from presidio_analyzer import (
     AnalyzerEngine,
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/pii/algorithms/presidio_utils.py
```diff
@@ -28,20 +29,27 @@
...
+)
 from metadata.utils.dispatch import class_register
 from metadata.utils.logger import pii_logger
 
 logger = pii_logger()
 
 
 def load_nlp_engine(
-    model_name: str = SPACY_EN_MODEL, supported_language: str = SUPPORTED_LANG
+    model_name: str = SPACY_MULTILANGUAGE_MODEL,
```
edg956, line 51: This model has about the same accuracy as the previous one, although limited to NER recognition
- [Multi language](https://spacy.io/models/xx#xx_ent_wiki_sm)
- [English](https://spacy.io/models/en#en_core_web_md)

PR #24423, opened 2025-11-18
fix(ui): tier count for quick filters

chirag-madlani:
### Describe your changes:

Fixes: Tier count not appearing on Explore filters

Screenshot 2025-11-18 at 7 48 47 PM

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ExploreQuickFilters.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ExploreQuickFilters.spec.ts
@@ -20,7 +20,11 @@ import {
   createNewPage,
   redirectToHomePage,
 } from '../../utils/common';
-import { assignTag } from '../../utils/entity';
+import {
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ExploreQuickFilters.spec.ts
```diff
@@ -43,6 +49,7 @@ test.beforeAll('Setup pre-requests', async ({ browser }) => {
     table.endpoint,
     'KnowledgePanel.Tags'
   );
+  await assignTier(page, 'Tier5', table.endpoint);
   await afterAction();
```
aniketkatkar97, line 53: Since it's in prerequisites, better to do it via API call rather than from UI to save time, we might even not need to make it slow then.

PR #24462, opened 2025-11-20
Lineage node column pagination

ugouhar:
### Describe your changes:

- Clicking on a node shows the lineage for that node only, highlights the traced nodes and edges
- When the column layer is applied, column-level lineage is visible.
- Columns in nodes now have pagination with each page showing max of 5 columns
- Only edges where both source and target are present on the current pages in nodes are visible.
- When the page changes, visible edges are updated accordingly.
- Hovering on a column shows the lineage for that column only, highlights the traced columns and edges in blue, and hides all other column-to-column and node-to-node edges.
- When a column is hovered and the traced columns are on a different page, those traced columns appear at the bottom of the current page with a divider separating them from the current page's columns.
- Clicking on a column selects it and shows the lineage for that column only, highlights the traced columns and edges in purple, and hides all other column-to-column and node-to-node edges.
- When a column is selected, hovering over another column has no hover effect.
- When a column is selected, clicking on another column selects that column.
- Clicking on a column to column edge, highlight that edge, source and target columns and hides all other edges
- When a column is searched inside a node, only the edges for matching columns in that node are visible.
- In edit mode, when connecting a column to another column, the source column gets selected and its traced columns and edges are highlighted in purple. All other edges are hidden. When dropping the connection, the target column is highlighted in blue for easy identification.

https://www.loom.com/share/f56679eaae5c4550aaec1318022cf4fb

### New integration tests
New integration test for column level lineage are added:
- Verify column visibility across pagination pages
- Verify edges when no column is hovered or selected
- Verify columns and edges when a column is hovered
- Verify columns and edges when a column is clicked

### Arrow color
The arrow at the end of each edge is an SVG marker. React Flow creates only one marker and reuses it everywhere. By setting the `color` property on `markerEnd` for column-level lineage edges, a separate marker is created specifically for those edges. This results in two markers, but the only way to distinguish them is through the `nth-child` selector—which is why the CSS change uses that approach.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Entity/EntityLineage/custom-node.less
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Entity/EntityLineage/custom-node.less
@@ -618,7 +618,7 @@
   z-index: 1; // Adding as column text and data type should be behind handles
 
   &.react-flow__handle-left {
-    left: 0;
+    left: -6px;
   }
 
   &.react-flow__handle-right {


openmetadata-ui/src/main/resources/ui/src/components/Entity/EntityLineage/CustomNodeV1.test.utils.tsx
chirag-madlani, line 1: Please keep file name as *.test.ts or *.test.tsx

openmetadata-ui/src/main/resources/ui/src/components/Entity/EntityLineage/NodeChildren/NodeChildren.component.tsx
```diff
@@ -32,8 +36,181 @@ import { getTestCaseExecutionSummary } from '../../../../rest/testAPI';
...
+}
+
+const CustomPaginatedList = ({
+  items,
+  filteredColumns,
+  nodeId,
+  page,
+  setPage,
+}: CustomPaginatedListProps) => {
+  const ITEMS_PER_PAGE = 5;
```
chirag-madlani: Move to const

openmetadata-ui/src/main/resources/ui/src/components/Entity/EntityLineage/custom-node.less
```diff
@@ -339,29 +350,44 @@
...
+        &:hover {
+          background-color: #d1e9ff;
+        }
+      }
+      &.any-column-selected {
         .custom-node-column-container.custom-node-header-column-tracing {
-          background-color: #ffffff;
-          border-color: #1570ef;
+          background-color: #e0eaff;
+          border-left-color: #444ce7;
```
chirag-madlani: Can we use variables instead?

chirag-madlani:
@ugouhar checked recording, are we missing Arrow at end of the Edge?

ugouhar:
> @ugouhar checked recording, are we missing Arrow at end of the Edge?

@chirag-madlani I checked, it's there, z-index needs to be updated

openmetadata-ui/src/main/resources/ui/src/components/Entity/EntityLineage/CustomNodeV1.test.utils.tsx
ugouhar: It's only a utility file and contains helper function for actual test file `CustomNodeV1.test.tsx`. The recommended name suggested by Claude is `CustomNodeV1.test.utils.ts`. To avoid any confusion I will move those helper function inside the main test file.

PR #24479, opened 2025-11-21
Fix NPE in IngestionPipelineRepository update methods for null fields

ayush-shah:
This pull request improves the robustness of equality checks in the `IngestionPipelineRepository` by replacing direct `.equals()` calls with `Objects.equals()`. This change helps prevent potential `NullPointerException`s when comparing possibly null values in several update methods.

Key improvements to equality checks:

* Replaced `.equals()` with `Objects.equals()` in update methods (`updateAirflowConfig`, `updateLogLevel`, `updateEnableStreamableLogs`, `updateDeployed`, `updateRaiseOnError`, and `updateEnabled`) to safely handle null values and avoid `NullPointerException`s.
* Added the necessary `import java.util.Objects;` statement to support the new equality checks.

### DIFF
--- a/.github/workflows/docker-openmetadata-ingestion.yml
+++ b/.github/workflows/docker-openmetadata-ingestion.yml
@@ -28,6 +28,16 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: false
+          swap-storage: true
+          docker-images: false
       - name: Check out the Repo
         uses: actions/checkout@v4
 
--- a/.github/workflows/ui-core-components-tests.yml
+++ b/.github/workflows/ui-core-components-tests.yml
@@ -1,0 +1,49 @@
+#  Copyright 2025 Collate
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  http://www.apache.org/licenses/LICENSE-2.0
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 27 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
```diff
@@ -551,33 +551,36 @@ private void updateAirflowConfig(
     }
 
     private void updateLogLevel(LogLevels origLevel, LogLevels updatedLevel) {
-      if (updatedLevel != null && !origLevel.equals(updatedLevel)) {
+      if (updatedLevel != null && (origLevel == null || !origLevel.equals(updatedLevel))) {
```
sonika-shah: @ayush-shah  better if we can use !Objects.equals(origEnabled, updatedEnabled) instead of  !Objects.equals(origEnabled, updatedEnabled) , it would take care of the NPE

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
```diff
@@ -551,33 +551,36 @@ private void updateAirflowConfig(
...
         recordChange("loggerLevel", origLevel, updatedLevel);
       }
     }
 
     private void updateEnableStreamableLogs(
         Boolean origEnableStreamableLogs, Boolean updatedEnableStreamableLogs) {
       if (updatedEnableStreamableLogs != null
-          && !origEnableStreamableLogs.equals(updatedEnableStreamableLogs)) {
+          && (origEnableStreamableLogs == null
+              || !origEnableStreamableLogs.equals(updatedEnableStreamableLogs))) {
```
sonika-shah: similarly here we can do - if (updatedEnableStreamableLogs != null
          && !Objects.equals(origEnableStreamableLogs, updatedEnableStreamableLogs)) {

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
ayush-shah: Thanks, will apply that.

sonika-shah (approved):

Issue #24544, opened 2025-11-25
Consider multiple `Classification`s in `TagProcessor` and improve the explanation behind a classification


edg956:
**Is your feature request related to a problem? Please describe.**

Part of allowing dynamic recognizers in auto-classification and the on-going improvements we've been doing on the auto-classification process, we need to ensure two things:

1. Ensure `TagProcessor` takes into account more than just PII classifications
2. Improve the explanation we give to users when a tag is chosen for a column

PR #24545, opened 2025-11-25
Allow multiple classifications in `TagProcessor`

edg956:
## Describe your changes:

Part 1 of #24544: supporting multiple classifications in `TagProcessor`

I worked on refactoring the work I'd done before because the implementation didn't properly support taking into account multiple classifications, it only obtained tags and picked the highest scored tag, regardless of whether a potential `General.CreditCard` tag was winning over a more important `PII.Sensitive` tag.

This is a lengthy PR in part because I added `factory-boy` to create tests fixtures. I recommend going commit by commit for a focused review.

### Changes by commit
- Added `factory-boy` and created fixture factories
- Minor adjustment to a docker util to ease discovery of misconfigured `testcontainers`
- **Added a `ConflictResolver` class to solve conflicts between tags of the same classification**
- **Refactored `TagClassifier` to abstract the scoring logic**
- **Abstracted how we retrieve available `Classification`s and `Tag`s**
- **Used everything built up to now to refactor `TagProcessor`**
- **Added an integration test**
- Added a migration to enable the `PII` classification
- Moved a test dependency to `_openmetadata_testutils`
- **Fixed broken mutually exclusive classifications on reruns**

#
### Type of change:
- [X] Improvement

#
### Checklist:
- [X] I have commented on my code, particularly in hard-to-understand areas.

### DIFF
--- a/ingestion/src/metadata/pii/tag_processor.py
+++ b/ingestion/src/metadata/pii/tag_processor.py
@@ -89,6 +89,58 @@ class TagProcessor(AutoClassificationProcessor):
 
         return tag_label
 
+    def filter_tags_to_analyze(
+        self, column: Column, candidate_tags: List[Tag]
+    ) -> List[Tag]:
+        """
+        Filter candidate tags based on already-applied tags and mutually exclusive
+        classification constraints.
+
+        Returns only tags that should be analyzed for this column.
+        """
+        existing_tag_fqns = {
+            tag.tagFQN.root for tag in (column.tags or []) if tag.tagFQN
+        }
+
+        # Build classification lookup map
+        classification_map = {
+            classification.fullyQualifiedName.root: classification
+            for classification in self.enabled_classifications
+        }
+
+        # Identify mutually exclusive classifications that already have tags applied
+        mutually_exclusive_classifications_with_tags = set()
+        for existing_tag_fqn in existing_tag_fqns:
+            if "." in existing_tag_fqn:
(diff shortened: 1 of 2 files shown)


ingestion/src/_openmetadata_testutils/factories/metadata/generated/schema/api/classification/create_classification.py
```diff
@@ -0,0 +1,34 @@
...
+)
+from metadata.generated.schema.api.classification.createClassification import (
+    AutoClassificationConfig,
+    ConflictResolution,
+    CreateClassificationRequest,
+)
+from metadata.generated.schema.type.basic import EntityName
+
+
+class AutoClassificationConfigFactory(factory.Factory):
```
pmbrull, line 15: do we want/need these classes under `generated/`? should we find another place for them?

edg956: Well this is defined in `_openmetadata_utils`, not `metadata/generated`. The idea of keeping such structure is that it's easier to look for factoryboy factories for your generated models.

Of course, not so relevant in the context of IDEs suggesting imports, but if you have to go through the directory structure it might make it more intuitive.

e.g: since our generated class `AutoClassificationConfig` is defined in `metadata/generated/schema/api/classification/create_classification.py`, we can expect to find it in the "same place" in `_openmetadata_utils/factories`.

Does it make sense?

pmbrull (approved):

PR #24659, opened 2025-12-02
Fix using wrong indices for lineage

Bhanuagr:
### Root Cause

The lineage search was failing because Service entities are indexed in separate service-specific indices (api_service_search_index_*), not in the dataAsset alias that the lineage code was searching.

### Fix:

Changed all lineage search operations to use GLOBAL_SEARCH_ALIAS ("all") instead of DATA_ASSET_SEARCH_ALIAS ("dataAsset"). This allows the lineage API to search across all entity types, including:
  - Data assets (tables, topics, dashboards, etc.)
  - Service entities (apiService, databaseService, etc.)
  - All other indexed entities

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/conf/openmetadata.yaml
+++ b/conf/openmetadata.yaml
@@ -294,7 +294,7 @@ authenticationConfiguration:
   # Force secure flag on session cookies even when not using HTTPS directly.
   # Enable this when running behind a proxy/load balancer that handles SSL termination.
   # Default: false (secure flag only set when HTTPS is detected)
-  forceSecureSessionCookie: ${FORCE_SECURE_SESSION_COOKIE:-true}
+  forceSecureSessionCookie: ${FORCE_SECURE_SESSION_COOKIE:-false}
   oidcConfiguration:
     id: ${OIDC_CLIENT_ID:-""}
     type: ${OIDC_TYPE:-""} # google, azure etc.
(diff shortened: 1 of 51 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OSLineageGraphBuilder.java
```diff
@@ -862,7 +862,7 @@ private void getEntitiesAtSpecificDepthWithPagination(
       SearchRequest searchRequest =
           getSearchRequest(
               request.getDirection(),
-              DATA_ASSET_SEARCH_ALIAS,
+              GLOBAL_SEARCH_ALIAS,
```
harshach, line 865: any reason we are changing this to GLOBAL, the goal is to have dataAsset only search no? is it because of services are not part of dataAsset alias?

Bhanuagr: yes services are not part of dataasset causing return empty result, verified changes by running ES queries and also verified from UI

harshach: @Bhanuagr thinking out loud given global alias contains everything , should we create another alias umbreall where services and dataAssets combined together. Having global alias might add unnecessary things into this?

harshach:
@chirag-madlani @mohityadav766 this requires through testing not merging into 1.11.0

harshach (approved):

PR #24729, opened 2025-12-08
Add SQLGlot parser support

mohittilala:
### Describe your changes:

Implement SQLGlot analyzer with SQLFluff/SQLParse fallback, query hash tracking, and optimized masking.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
+++ b/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
@@ -43,11 +43,7 @@ from metadata.generated.schema.type.entityLineage import (
 from metadata.generated.schema.type.entityLineage import Source as LineageSource
 from metadata.generated.schema.type.entityReference import EntityReference
 from metadata.ingestion.api.models import Either
-from metadata.ingestion.lineage.models import (
-    Dialect,
-    QueryParsingError,
-    QueryParsingFailures,
-)
+from metadata.ingestion.lineage.models import Dialect
 from metadata.ingestion.lineage.parser import LINEAGE_PARSING_TIMEOUT, LineageParser
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
 from metadata.utils import fqn
@@ -860,7 +856,7 @@ def get_lineage_by_query(
     Now supports cross-database lineage by accepting a list of service names.
     """
     column_lineage = {}
-    query_parsing_failures = QueryParsingFailures()
+
     if service_name and isinstance(service_name, str):
         service_names = [service_name]
         logger.warning(
@@ -949,17 +945,6 @@ def get_lineage_by_query(
                             graph=graph,
                             schema_fallback=schema_fallback,
                         )
-        if not lineage_parser.query_parsing_success:
-            logger.debug(
-                f"[{query_hash}] Query parsing failed while getting lineage by query:"
-                f" {lineage_parser.query_parsing_failure_reason}"
-            )
-            query_parsing_failures.add(
-                QueryParsingError(
-                    query=masked_query or query,
-                    error=lineage_parser.query_parsing_failure_reason,
-                )
-            )
     except Exception as exc:
         yield Either(
             left=StackTraceError(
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/ingestion/lineage/parser.py
```diff
@@ -69,6 +78,7 @@ def __init__(
         timeout_seconds: int = LINEAGE_PARSING_TIMEOUT,
     ):
         self.query = query
+        self.query_hash = self.get_query_hash(query)
```
ulixius9, line 81: is the purpose of query hash only for logging?

ingestion/src/metadata/ingestion/lineage/masker.py
```diff
@@ -120,30 +134,86 @@ def mask_query(
...
             return masked_query_cache.get((query, dialect))
         if parser_required and not parser:
+            logger.warning(
+                f"{hash_prefix}Query masking skipped as no parser available."
+            )
             return None
-        if not parser:
+
+        # Since SqlGlot generalizes query structures/syntax, we will use
+        # SqlParse for masking if SqlGlot is used for parsing
```
ulixius9, line 173: not sure if i understand this fully?? why can't we build a making with sqlglot - using sqlparse here would not achieve the performance improvements we are expecting

ingestion/src/metadata/ingestion/lineage/parser.py
ulixius9: also how does it help in debugging? since we are logging it everywhere

mohittilala: since we are using multithreading to parse queries, we usually don't get idea on which query is taking more time and in which step.

we can't append query in every log message so with query hash we are adding it to every logs and just search which query is performing poorly and OOM issue tracking will be easier.

ingestion/src/metadata/ingestion/lineage/masker.py
mohittilala: sqlglot normalizes sqls after parsing and doesn't retain original sql content.

for reference (masking with sqlglot):
```
Original Query: SELECT * FROM user WHERE id=1234 AND name='Alice' AND birthdate=DATE '2023-01-01';

E       AssertionError: 
E       	[SqlGlot] Expected Masked Query: SELECT * FROM user WHERE id=? AND name=? AND birthdate=DATE ?;
E       	[SqlGlot] Actual Masked Query: SELECT * FROM user WHERE id = ? AND name = ? AND birthdate = CAST(? AS DATE)
```

here, type casting `birthdate=DATE ?` is normalized to `birthdate = CAST(? AS DATE)` and there is no option for us to disable this behaviour as it's intended use-case of sqlglot.

ideally, this is not an issue but when sql is complex and sqlglot does any mistake in converting sql then we wouldn't even get the original sql so can't rely 100% on the parsing/conversion capabilities of sqlglot.

mohittilala: Code reference:

```
from sqlglot import parse_one, exp

sql = "SELECT * FROM user WHERE id=1234 AND name='Alice' AND birthdate=DATE '2023-01-01';"

expression:exp.Expression = parse_one(sql)
transformed = expression.transform(
    lambda n: exp.Placeholder() if isinstance(n, exp.Literal) else n
)
masked_sql = transformed.sql()

print(f"Original Query: {sql}")
print(f"Masked Query:   {masked_sql}")
```

mohittilala: sqlparse parsing time is almost same as sqlglot so here it would be 2x time than expectation but no workaround for now.

mohittilala:
playwright failures here are not related to pr changes so we are good to merge.

---

2 failed playwright tests

[chromium] › playwright/e2e/Pages/Glossary.spec.ts:2088:7 › Glossary tests › Create glossary, change language to Dutch, and delete glossary
    -- Fix PR: https://github.com/open-metadata/OpenMetadata/pull/24927

[chromium] › playwright/e2e/Pages/Lineage.spec.ts:96:7 › Lineage creation from Container entity
    -- Fix PR: https://github.com/open-metadata/OpenMetadata/pull/24896

pmbrull (approved):

PR #24808, opened 2025-12-12
test(ui): fix glossary and domain playwright

karanh37:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/utils/glossary.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/utils/glossary.ts
@@ -723,8 +723,8 @@ export const validateGlossaryTerm = async (
     await expect(page.locator(termSelector)).toContainText(term.name);
     await expect(page.locator(statusSelector)).toBeVisible();
 
-    // If status is Draft or In Review, then check for either or
-    if (status === 'Draft' || status === 'In Review') {
+    // If status is Draft, then check for either Draft or In Review
+    if (status === 'Draft') {
       await expect(page.locator(statusSelector)).toContainText(
         /Draft|In Review/
       );


openmetadata-ui/src/main/resources/ui/playwright/utils/glossary.ts
```diff
@@ -722,14 +722,22 @@ export const validateGlossaryTerm = async (
     await expect(page.locator(termSelector)).toBeVisible();
     await expect(page.locator(termSelector)).toContainText(term.name);
     await expect(page.locator(statusSelector)).toBeVisible();
-    await expect(page.locator(statusSelector)).toContainText(status);
+
+    // If status is Draft or In Review, then check for either or
+    if (status === 'Draft' || status === 'In Review') {
+      await expect(page.locator(statusSelector)).toContainText(
+        /Draft|In Review/
```
shah-harshit, line 729: How are if and else different? They look same to me. The text here is same as the status.

karanh37: When glossary is created at that a workflow is triggered which is async in nature.
The starting state of Glossary is 'Draft', but it can happen workflow quickly syncs the default state and make it as 'In Review', so when i'm testing for creation scenario it should be either or.

But there are cases where i explicitly Approve/Reject glossary in those cases i want to check for status.
So else  is for explicit check

shah-harshit: Since this quick change will happen only in case of Draft, we can remove "in review" from the if condition right? In case of "In review", it should be "In review" only and not "draft".

karanh37: Yeah makes sense, updated the condition

PR #24817, opened 2025-12-12
Tagging explanation

### LINKED ISSUE #24880: Improve metadata to applied tags
# Improvement Request
## Affected Module
Auto Classification

## Explanation

We have very little visibility on the output of the Auto Classification agent. Results from presidio are not explained and debugging false positives requires an exercise of simulating scenarios with an environment as similar to the production environment as possible.

## Goal

1. Improve the explanation given by `TagProcessor` and `PIIProcessor` to the generated `TagLabel`s 
2. Add a timestamp to understand when was the tag applied
3. Add a field to track the name of the user that produced the tag

edg956:
### Describe your changes:

Fixes #24880 

#### Changes By Commit
1. **Added a new field to track insertion timestamps in `tag_usage`, retrieved in `TagLabel.appliedAt`**
2. **Create proper explanations from presidio results**
3. **Refactored `TagProcessor` to capture explanations from content and column analysis**
4. **Added a mechanism to capture the results produced by presidio**
5. Applied PR comments by moving values to a constants file
6. **Added `TagLabel.appliedBy` field to know who created the tag**
7. Updated typescript types
8. Fixed python tests
9. Fixed java tests

In **bold** the relevant commits

#
### Type of change:
- [X] Improvement

#
### Checklist:

---

## Summary by Gitar

- **Tag audit trail:**
  - Added `appliedAt` timestamp and `appliedBy` identifier to `tag_usage` table with MySQL/PostgreSQL migrations
- **PII explanation generation:**
  - `explain_recognition_results()` in `presidio_utils.py` formats Presidio detection results into textual explanations with recognizer names, counts, and pattern scores
- **Analysis refactoring:**
  - `TagAnalyzer` returns `TagAnalysis` objects containing tag, score, and explanation instead of raw floats
- **Performance optimization:**
  - Module-level cache in `get_cached_analyzer_engine()` prevents reloading spaCy models (~40MB each)

---


ingestion/src/metadata/pii/algorithms/presidio_utils.py
```diff
@@ -290,3 +303,58 @@ def analyze(
...
+        sorted(results, key=_get_getter),
+        key=_get_getter,
+    )
+
+    textual_explanation = ""
+    for recognizer_identifier, group in grouped_results:
+        group_list = list(group)
+
+        recognizer_name: str = cast(
+            Dict[str, str], group_list[0].recognition_metadata
```
harshach, line 329: is this the safe way to do it , directly subscripting on group_list, doesn't look like we are checking if it has any values

ingestion/src/metadata/pii/algorithms/presidio_utils.py
```diff
@@ -290,3 +303,58 @@ def analyze(
...
+        ).get("recognizer_name", recognizer_identifier)
```
harshach: should be a constant string?

ingestion/src/metadata/pii/algorithms/presidio_utils.py
```diff
@@ -290,3 +303,58 @@ def analyze(
...
+                is None  # pyright: ignore[reportUnnecessaryComparison]
+            ):
+                continue
+
+            patterns_matched.add(
+                (result.analysis_explanation.pattern, result.analysis_explanation.score)
+            )
+
+        if patterns_matched:
+            textual_explanation += "Patterns matched:\n"
```
harshach: same would like to see these defined clearly in a single file with constants

openmetadata-spec/src/main/resources/json/schema/type/tagLabel.json
```diff
@@ -57,6 +57,11 @@
     "reason": {
       "type": "string",
       "description": "An explanation of why this tag was proposed, specially for autoclassification tags"
+    },
```
harshach, line 60: shall we add more details who it is applied this , could be a bot, AI, or a human

edg956: Makes sense 👍

ingestion/src/metadata/pii/algorithms/presidio_utils.py
edg956: `groupby` never returns empty group iterables. Even if trying to group an empty sequence, in which case iterating the result of `groupby` would result in 0 loops

openmetadata-spec/src/main/resources/json/schema/type/tagLabel.json
edg956: I've followed the convention to add the user's name

openmetadata-service/src/main/java/org/openmetadata/service/util/JsonPatchUtils.java
```diff
@@ -189,4 +187,56 @@ public static MetadataOperation getMetadataOperation(String path) {
...
+      } else {
+        // Keep other operations unchanged
+        modifiedOpsBuilder.add(op);
+      }
+    }
+
+    // Create new JsonPatch from modified operations
+    JsonArray modifiedOps = modifiedOpsBuilder.build();
+    return Json.createPatch(modifiedOps);
+  }
```
edg956: Perhaps a little hacky but a way to patch new tags without having to drill down `user` into the stack call where the SQL statement is produced

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -26,16 +26,7 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
```
mohityadav766: we should avaoid wildcard

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -218,12 +218,8 @@
 import org.openmetadata.service.search.SearchSortFilter;
 import org.openmetadata.service.security.AuthorizationException;
 import org.openmetadata.service.security.policyevaluator.SubjectContext;
-import org.openmetadata.service.util.EntityETag;
-import org.openmetadata.service.util.EntityUtil;
+import org.openmetadata.service.util.*;
```
mohityadav766: avoid wildcards

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -4561,11 +4552,11 @@ List<String> getTagsStartingWithPrefix(
   interface TagUsageDAO {
     @ConnectionAwareSqlUpdate(
         value =
-            "INSERT IGNORE INTO tag_usage (source, tagFQN, tagFQNHash, targetFQNHash, labelType, state, reason) VALUES (:source, :tagFQN, :tagFQNHash, :targetFQNHash, :labelType, :state, :reason)",
+            "INSERT IGNORE INTO tag_usage (source, tagFQN, tagFQNHash, targetFQNHash, labelType, state, reason, appliedBy) VALUES (:source, :tagFQN, :tagFQNHash, :targetFQNHash, :labelType, :state, :reason, :appliedBy)",
```
mohityadav766, line 4570: applied By is not mandatorily available while checking Table tags for example we want to be able to list all the tags that asset has

openmetadata-service/src/main/java/org/openmetadata/service/util/JsonPatchUtils.java
edg956: I'll change this

edg956: Use `EntityUpdater.updatingUser` instead

bootstrap/sql/migrations/native/1.11.4/postgres/schemaChanges.sql
```diff
@@ -5,3 +5,7 @@ ADD COLUMN appliedAt TIMESTAMP;
 
 ALTER TABLE tag_usage
 ALTER COLUMN appliedAt SET DEFAULT NOW();
+
+-- Include `tag_usage.appliedBy` field
+ALTER TABLE tag_usage
+ADD COLUMN appliedBy VARCHAR(64) DEFAULT 'Unknown';
```
edg956: Use admin here

openmetadata-service/src/main/java/org/openmetadata/service/util/JsonPatchUtils.java
```diff
@@ -17,10 +17,7 @@
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.fge.jsonpatch.JsonPatchException;
-import jakarta.json.JsonObject;
-import jakarta.json.JsonPatch;
-import jakarta.json.JsonString;
-import jakarta.json.JsonValue;
+import jakarta.json.*;
```
pmbrull: remove the *

pmbrull (approved):

pmbrull (approved):

PR #24836, opened 2025-12-15
Fix#2489: Entity Type not being sent inside the `EntityReference` object

aniketkatkar97:
Fixes https://github.com/open-metadata/openmetadata-collate/issues/2489

Earlier, the `type` field in the entity reference object formed by a common util function was being formulated by the `searchIndex` coming from the searchAPI by searchIndex->entityType mapping. This was causing the issue since the search index can be different from the earlier assumptions. ( earlier expected 'domain_search_index` -> randomised 'domain_search_index_bs9d92'. 

I worked on removing the dependency on the search index and directly utilising the `entityType` field from the ES source.

### DIFF
--- a/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
+++ b/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
@@ -66,12 +66,24 @@ CUTOFF_NODES = 20
 
 
 # pylint: disable=too-many-function-args,protected-access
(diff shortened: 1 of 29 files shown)


aniketkatkar97:
@copilot can you fix the Sonar issues in the AlertsUtil.tsx reported [here](https://sonarcloud.io/project/issues?id=open-metadata-ui&pullRequest=24836&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)?

PR #24845, opened 2025-12-16
fix(test): flaky right panel flow

harsh-vador:
### Describe your changes:

Fix: Array field updates causing 500 errors in entity updates

### Problem

When updating array fields (tags, glossary terms, etc.) in entity summary panels, the application was generating JSON Patch operations with specific array indices that caused 500 errors from the backend. Error example:
`{
  "code": 500,
  "errorType": "UNHANDLED_SERVER_EXCEPTION",
  "message": "An exception with message [Error processing JSON Patch] was thrown while processing request."
}`

Patch operation:
`[
  {
    "op": "add",
    "path": "/tags/3",
    "value": {...}
  }
]`

### Root Cause
- The fast-json-patch library's compare() function generates add operations with specific array indices (e.g., /tags/3) when adding items to arrays. The backend's JSON Patch implementation rejects these operations when the target index doesn't exist yet.

### Solution
- Modified EntityUpdateUtils.ts to detect array index-based operations and convert them to a single replace operation for the entire array:
  - Detects when jsonPatch contains add or remove operations with paths ending in array indices (/\d+$)
  - For array fields with such operations, replaces the entire patch with a single replace operation
  - Non-array fields and patches without index operations remain unchanged

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/AbstractNativeApplication.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/AbstractNativeApplication.java
@@ -34,6 +34,7 @@ import org.openmetadata.service.Entity;
 import org.openmetadata.service.apps.scheduler.AppScheduler;
 import org.openmetadata.service.apps.scheduler.OmAppJobListener;
 import org.openmetadata.service.exception.EntityNotFoundException;
+import org.openmetadata.service.fernet.Fernet;
 import org.openmetadata.service.jdbi3.AppRepository;
 import org.openmetadata.service.jdbi3.CollectionDAO;
 import org.openmetadata.service.jdbi3.IngestionPipelineRepository;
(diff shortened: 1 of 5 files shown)


openmetadata-ui/src/main/resources/ui/src/utils/EntityUpdateUtils.ts
```diff
@@ -56,12 +56,26 @@ export const updateEntityField = async <T>({
...
+  if (Array.isArray(newValue) && hasArrayIndexOperations) {
+    jsonPatch = [
+      {
+        op: 'replace',
+        path: `/${fieldName}`,
+        value: newValue,
+      },
+    ];
+  }
+
```
aniketkatkar97, line 78: @harsh-vador we don't need to do any custom handling while creating patch objects. Whatever is generated by the `compare` function should be good

PR #24853, opened 2025-12-16
Enable column layer when entering edit mode

ugouhar:
### Describe your changes:
- The current column-level lineage had one small confusing user experience. When users entered edit mode without the column layer applied, they could manually expand individual columns and connect them. However, the newly added edges weren't visible until the column layer was applied.
- To improve the experience, I'm automatically enabling the column layer when users enter edit mode. This makes sense—users entering edit mode likely intend to add or delete column-level edges.
- When exiting edit mode, the column layer state is preserved. Users may switch between layers and toggle the column layer multiple times, so it makes sense to keep it as is when exiting edit mode.

https://github.com/user-attachments/assets/b2d13b70-1a77-4dbc-91c4-bcf5514cd492

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Lineage.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Lineage.spec.ts
@@ -34,6 +34,7 @@ import {
   addColumnLineage,
   addPipelineBetweenNodes,
   applyPipelineFromModal,
+  clickLineageNode,
   connectEdgeBetweenNodes,
   connectEdgeBetweenNodesViaAPI,
   deleteEdge,
@@ -143,9 +144,8 @@ for (const EntityClass of entities) {
             entity,
             'entityResponseData.fullyQualifiedName'
           );
-          await page
-            .locator(`[data-testid="lineage-node-${toNodeFqn}"]`)
-            .click();
+
+          await clickLineageNode(page, toNodeFqn);
 
           await expect(
             page
@@ -170,9 +170,8 @@ for (const EntityClass of entities) {
           currentEntity,
           'entityResponseData.fullyQualifiedName'
         );
-        await page
-          .locator(`[data-testid="lineage-node-${fromNodeFqn}"]`)
-          .click({ position: { x: 5, y: 5 } });
+
+        await clickLineageNode(page, fromNodeFqn);
 
         for (const entity of entities) {
           await applyPipelineFromModal(page, currentEntity, entity, pipeline);
@@ -527,7 +526,7 @@ test('Verify table search with special characters as handled', async ({
 
     await expect(page.locator('[data-testid="lineage-details"]')).toBeVisible();
 
-    await page.locator(`[data-testid="lineage-node-${dbFqn}"]`).click();
+    await clickLineageNode(page, dbFqn);
     await page.waitForLoadState('networkidle');
 
     await expect(
@@ -691,9 +690,8 @@ test('Verify cycle lineage should be handled properly', async ({ page }) => {
 });
 
 test('Verify column layer is applied on entering edit mode', async ({
-  browser,
+  page,
 }) => {
-  const { page } = await createNewPage(browser);
   const { apiContext, afterAction } = await getApiContext(page);
   const table = new TableClass();
 
@@ -734,9 +732,8 @@ test('Verify column layer is applied on entering edit mode', async ({
 });
 
 test('Verify there is no traced nodes and columns on exiting edit mode', async ({
-  browser,
+  page,
 }) => {
-  const { page } = await createNewPage(browser);
   const { apiContext, afterAction } = await getApiContext(page);
   const table = new TableClass();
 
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Lineage.spec.ts
```diff
@@ -168,7 +168,7 @@ for (const EntityClass of entities) {
         );
         await page
           .locator(`[data-testid="lineage-node-${fromNodeFqn}"]`)
-          .click();
+          .click({ position: { x: 5, y: 5 } });
```
ugouhar: In this PR, columns are automatically expanded in edit mode. At line 170, the test clicks on the node by clicking at the center. However, since columns are expanded, Playwright was clicking on the column instead of the node, causing the test to fail. To fix this, I'm clicking at the top-left position to ensure the node is clicked

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Lineage.spec.ts
```diff
@@ -732,7 +732,7 @@ test.describe.serial('Test pagination in column level lineage', () => {
 
     await addPipelineBetweenNodes(page, table1, table2);
 
-    await activateColumnLayer(page);
+    await rearrangeNodes(page);
```
ugouhar, line 838: `addPipelineBetweenNodes` step includes entering edit mode and when edit mode is entered through edit lineage button, column layer is automatically applied. So we just need to rearrange the nodes.

openmetadata-ui/src/main/resources/ui/playwright/utils/lineage.ts
```diff
@@ -90,6 +90,16 @@ export const verifyColumnLayerInactive = async (page: Page) => {
...
+
+  const isColumnLayerSelected = await page
+    .locator('[data-testid="lineage-layer-column-btn"]')
+    .evaluate((el) => el.classList.contains('Mui-selected'));
+
+  if (isColumnLayerSelected) {
+    await clickOutside(page);
+    return;
+  }
+
```
ugouhar, line 102: Previously, this function opened the layers option and clicked "Column layer." Although named `activateColumnLayer`, it was actually toggling the column layer. I updated it to only activate the column layer by returning early if it's already applied—otherwise, it applies the column layer.

openmetadata-ui/src/main/resources/ui/src/context/LineageProvider/LineageProvider.tsx
```diff
@@ -1122,12 +1122,23 @@ const LineageProvider = ({ children }: LineageProviderProps) => {
...
+      LineageLayer.ColumnLevelLineage
+    );
+
+    if (!isEditMode && !hasColumnLayer) {
+      setActiveLayer((pre) => uniq([LineageLayer.ColumnLevelLineage, ...pre]));
+    } else if (isEditMode && hasColumnLayer) {
+      setTracedNodes([]);
+      setTracedColumns([]);
+    }
+
```
ugouhar, line 1135: The first `if` checks whether the user is entering edit mode but column layer is not applied—if true, it applies the column layer.
In the `else` block, when the user exits edit mode, traced nodes and columns are reset to empty. This handles cases where the user clicks on nodes or columns while in edit mode, leaving traced nodes and columns populated that need to be cleared upon exit.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Lineage.spec.ts
chirag-madlani: @ugouhar wondering if we can modify the selector to aim the header portion of the node instead whole container? WDYT?

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Lineage.spec.ts
```diff
@@ -693,6 +694,112 @@ test('Verify cycle lineage should be handled properly', async ({ browser }) => {
   }
 });
 
+test('Verify column layer is applied on entering edit mode', async ({
+  browser,
```
chirag-madlani: In playwright each test will have it's own page can we confuse page directly instead creating within test?

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Lineage.spec.ts
```diff
@@ -693,6 +694,112 @@ test('Verify cycle lineage should be handled properly', async ({ browser }) => {
...
+      }
+    );
+  } finally {
+    await table.delete(apiContext);
+    await afterAction();
+  }
+});
+
+test('Verify there is no traced nodes and columns on exiting edit mode', async ({
+  browser,
```
chirag-madlani: same

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Lineage.spec.ts
ugouhar: Yes seems good, let me try that.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Lineage.spec.ts
ugouhar: Didn't get your point. In Lineage.spec.ts, all tests are creating page within test.

PR #24899, opened 2025-12-18
Chore(UI): Fix the flakiness in the DataContract AUTs

aniketkatkar97:
Screenshot 2025-12-18 at 9 24 26 PM

### DIFF
--- a/.github/workflows/maven-postgres-rdf-tests-build.yml
+++ b/.github/workflows/maven-postgres-rdf-tests-build.yml
@@ -12,10 +12,12 @@
 name: Maven Postgres RDF Tests CI
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
     paths:
+      - ".github/workflows/maven-postgres-rdf-tests-build.yml"
       - "openmetadata-service/src/main/java/org/openmetadata/service/rdf/**"
       - "openmetadata-service/src/test/java/org/openmetadata/service/resources/rdf/**"
       - "openmetadata-spec/src/main/resources/rdf/**"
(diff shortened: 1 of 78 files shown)


openmetadata-ui/src/main/resources/ui/playwright/utils/persona.ts
```diff
@@ -99,6 +99,9 @@ export const navigateToPersonaWithPagination = async (
   maxPages = 15
 ) => {
   for (let currentPage = 0; currentPage < maxPages; currentPage++) {
+    // Wait for the skeleton card loader to disappear first
+    await waitForAllLoadersToDisappear(page, 'skeleton-card-loader');
+
```
siddhant1, line 104: @aniketkatkar97 how  many loaders are here? instead of doing this, can we also pass in a parent as an optional prop and chain the locators to only locate on the element we are testing

siddhant1 (approved):

openmetadata-ui/src/main/resources/ui/playwright/utils/persona.ts
siddhant1: https://collate-inc.slack.com/archives/C04HGB0NY04/p1766122030988279?thread_ts=1766065496.201009&cid=C04HGB0NY04

aniketkatkar97: This loader itself is specific to this case. We'll improve the logic for this method to make the scope limited to some container.

ShaileshParmar11 (approved):

PR #24984, opened 2025-12-23
Memory leak in tag processor

edg956:
### Describe your changes:

This PR fixes a memory leak introduced in the `TagProcessor`, related to loading the NLP model (~50MB for the medium size we're using) as many times as columns we were classifying in an auto classification job.

This meant that a very wide table would explode memory usage before garbage collector could even react.

#
### Type of change:

- [X] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

---

## Summary by Gitar

- **Caching implementation:**
  - Added `@cache` decorator to `load_nlp_engine()` in `presidio_utils.py` for model memoization
- **Dependency injection fix:**
  - Modified `TagProcessor.__init__()` to load `nlp_engine` once and pass to `ScoreTagsForColumnService`
- **Test coverage:**
  - Added 5 unit tests in `test_presidio_utils.py` validating cache behavior across different parameters

---

### DIFF
--- a/ingestion/src/metadata/pii/tag_processor.py
+++ b/ingestion/src/metadata/pii/tag_processor.py
@@ -12,6 +12,7 @@ from metadata.generated.schema.type.tagLabel import (
     TagSource,
 )
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
+from metadata.pii.algorithms.presidio_utils import load_nlp_engine
 from metadata.pii.algorithms.tag_scoring import ScoreTagsForColumnService
 from metadata.pii.base_processor import AutoClassificationProcessor
 from metadata.pii.classification_manager import (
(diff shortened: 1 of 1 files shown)


ingestion/src/metadata/pii/tag_processor.py
```diff
@@ -68,7 +69,9 @@ def __init__(
 
         # Service that runs analyzers
         if score_tags_for_column is None:
-            score_tags_for_column = ScoreTagsForColumnService()
+            score_tags_for_column = ScoreTagsForColumnService(
+                nlp_engine=load_nlp_engine()
+            )
```
edg956, line 74: Not necessary anymore, but would've avoided the error to begin with

pmbrull (approved):

PR #25068, opened 2026-01-06
Fixes #25067: Delete the secrets only on hard delete

### LINKED ISSUE #25067: [Bug] Secrets are deleted during Service Soft Delete, causing connection failures upon restoration
Soft deletion removes the secrets from the storage where they are stored, on restoring, the test connection or any agent ingestion doesn't work as expected

Potential Root cause:

@Override
  protected void postDelete(T service, boolean hardDelete) {
    super.postDelete(service, hardDelete);
    if (service.getConnection() != null) {
      SecretsManagerFactory.getSecretsManager()
          .deleteSecretsFromServiceConnectionConfig(
              service.getConnection().getConfig(),
              service.getServiceType().value(),
              service.getName(),
              serviceType);
    }
  }

ayush-shah:
This pull request updates the behavior for deleting secrets associated with a service connection. Now, secrets are only removed from the secret manager when a service is hard deleted, making soft deletes reversible and preventing accidental loss of secrets.

Secret deletion logic update:

* Modified the `postDelete` method in `ServiceEntityRepository.java` so that secrets are deleted from the secret manager only when a hard delete occurs, not on soft delete.


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ServiceEntityRepository.java
```diff
@@ -105,11 +105,12 @@ public T addTestConnectionResult(UUID serviceId, TestConnectionResult testConnec
...
   }
 
-  /** Remove the secrets from the secret manager */
+  /** Remove the secrets from the secret manager only on hard delete */
   @Override
   protected void postDelete(T service, boolean hardDelete) {
     super.postDelete(service, hardDelete);
-    if (service.getConnection() != null) {
+    // Only delete secrets on hard delete to allow soft delete to be reversible
+    if (hardDelete && service.getConnection() != null) {
```
harshach, line 113: lets add some tests for this

PR #24774, opened 2025-12-10
feat: Enhanced the TagsPage to support auto-classification configurations

Rohit0301:
### Describe your changes:

This pull request introduces several enhancements and fixes related to tag handling, especially focusing on recognizer feedback for auto-applied tags on table columns, as well as improvements to Playwright UI tests for tag and classification forms. The most significant changes include better support for recognizing and removing auto-applied tags from table columns, updating the tag label type from `AUTOMATED` to `GENERATED`, and making the Playwright UI tests more robust and accurate for tag creation workflows.

**Recognizer Feedback and Tag Handling:**

* Added explicit handling for recognizing and removing auto-applied tags (now labeled as `GENERATED`) from table columns in `RecognizerFeedbackRepository`, including a new check for `Entity.FIELD_COLUMNS` and improved logic for updating column tags. [[1]](diffhunk://#diff-7dde33a99b5b7a469c3259f51ea51c475aff3655cf330fc097f7765a641d4290L148-R160) [[2]](diffhunk://#diff-7dde33a99b5b7a469c3259f51ea51c475aff3655cf330fc097f7765a641d4290L221-R264)
* Changed the tag label type from `AUTOMATED` to `GENERATED` throughout the codebase and tests to reflect the new convention for recognizer-applied tags. [[1]](diffhunk://#diff-7dde33a99b5b7a469c3259f51ea51c475aff3655cf330fc097f7765a641d4290L179-R191) [[2]](diffhunk://#diff-7dde33a99b5b7a469c3259f51ea51c475aff3655cf330fc097f7765a641d4290L286-R328) [[3]](diffhunk://#diff-f966c2afccbb460b1d5854416762adab333c913d40a81a047e8611868c745f9aL2543-R2543) [[4]](diffhunk://#diff-f966c2afccbb460b1d5854416762adab333c913d40a81a047e8611868c745f9aL2589-R2589) [[5]](diffhunk://#diff-f966c2afccbb460b1d5854416762adab333c913d40a81a047e8611868c745f9aL2630-R2630)

**Testing Improvements:**

* Added a new test in `TableResourceTest` to verify that recognizer feedback can remove auto-applied tags from table columns, ensuring only manual tags remain after feedback is processed.
[...]

PR #25099, opened 2026-01-07
Fix locator for tree-select spec

siddhant1:
Fix locator for tree-select spec

---

## Summary by Gitar

- **Test selector update:**
  - Changed from `data-testid="domain-select"` to `page.getByRole('combobox', { name: 'Select domain' })` for better accessibility testing
- **Simplified test interactions:**
  - Removed explicit `waitFor`, `click`, and API response waiting, relying on Playwright's auto-waiting
- **Bug fix:**
  - Corrected missing `page` context in original `getByRole` call

---

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/utils/domain.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/utils/domain.ts
@@ -663,16 +663,10 @@ export const createDataProductFromListPage = async (
   await fillCommonFormItems(page, dataProduct);
 
   // Fill domain field (required when creating from list page)
-  const domainInput = getByRole('combobox', { name: 'Select domain' })
+  const domainInput = page.getByRole('combobox', { name: 'Select domain' });
   await domainInput.scrollIntoViewIfNeeded();
-  await domainInput.waitFor({ state: 'visible' });
-  await domainInput.click();
 
-  const searchDomain = page.waitForResponse(
-    `/api/v1/search/query?q=*index=domain_search_index*`
-  );
   await domainInput.fill(domain.displayName);
-  await searchDomain;
 
   const domainOption = page.getByText(domain.displayName);
   await domainOption.waitFor({ state: 'visible', timeout: 5000 });


openmetadata-ui/src/main/resources/ui/playwright/utils/domain.ts
```diff
@@ -663,16 +663,10 @@ export const createDataProductFromListPage = async (
...
+  const domainInput = page.getByRole('combobox', { name: 'Select domain' });
   await domainInput.scrollIntoViewIfNeeded();
-  await domainInput.waitFor({ state: 'visible' });
-  await domainInput.click();
 
-  const searchDomain = page.waitForResponse(
-    `/api/v1/search/query?q=*index=domain_search_index*`
-  );
   await domainInput.fill(domain.displayName);
-  await searchDomain;
```
aniketkatkar97, line 675: Why did we remove the wait on the API call? Wouldn't that result in flakiness

ShaileshParmar11: Agree to this.

openmetadata-ui/src/main/resources/ui/playwright/utils/domain.ts
```diff
@@ -663,16 +663,10 @@ export const createDataProductFromListPage = async (
   await fillCommonFormItems(page, dataProduct);
 
   // Fill domain field (required when creating from list page)
-  const domainInput = page.getByTestId('domain-select');
+  const domainInput = page.getByRole('combobox', { name: 'Select domain' });
```
ShaileshParmar11, line 666: any reasons we are moving from data-testid to role base selection?

siddhant1: Yes, I'll revert this since data-test id doesn't exist in release as of now
https://github.com/open-metadata/OpenMetadata/pull/24774/changes

PR #25207, opened 2026-01-12
playwright: fixed incident manager test

ShaileshParmar11:
This pull request updates the Playwright utility functions in `incidentManager.ts` to improve how assignees are added to test cases, allowing for more precise UI interactions based on the context. The main change is making the `addAssigneeFromPopoverWidget` function context-aware by introducing an optional `testCaseName` parameter, which enables targeting specific test case rows when assigning users.

**Improvements to test case assignment logic:**

* Updated `addAssigneeFromPopoverWidget` to accept an optional `testCaseName` parameter, allowing the function to assign users to a specific test case row when provided, or default to the general assignee edit icon otherwise.
* Modified `assignIncident` to pass the `testCaseName` to `addAssigneeFromPopoverWidget` for direct assignments, ensuring the correct row is targeted during test case assignment.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/utils/incidentManager.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/utils/incidentManager.ts
@@ -62,10 +62,17 @@ export const acknowledgeTask = async (data: {
 export const addAssigneeFromPopoverWidget = async (data: {
   page: Page;
   user: { name: string; displayName: string };
+  testCaseName?: string;
 }) => {
-  const { page, user } = data;
-  // direct assignment from edit assignee icon
-  await page.click('[data-testid="assignee"] [data-testid="edit-owner"]');
+  const { page, user, testCaseName } = data;
+
+  if (testCaseName) {
+    await page.getByRole('row', { name: testCaseName }).getByTestId('edit-owner').click();
+  } else {
+    // direct assignment from edit assignee icon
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/playwright/utils/incidentManager.ts
```diff
@@ -62,10 +62,17 @@ export const acknowledgeTask = async (data: {
...
-  await page.click('[data-testid="assignee"] [data-testid="edit-owner"]');
+  const { page, user, testCaseName } = data;
+
+  if (testCaseName) {
+    await page.getByRole('row', { name: testCaseName }).getByTestId('edit-owner').click();
+  } else {
+    // direct assignment from edit assignee icon
+    await page.getByTestId('assignee').getByTestId('edit-owner').click();
+  }
+
```
siddhant1, line 75: @ShaileshParmar11 can we change the test id instead? just like we do for tags instead of passing the testCaseName, also I feel a better naming convention would be to just name the paramter `name`

ShaileshParmar11: There are two ways we are using this function.

1. In If condition we are assigning from listing page were there are more rows, thats why we are going with row approach.
2. Its direct assignment, where there will be only one assignee with edit owner id.

About name, i would prefer readable name which will give quick insight about what its name all about instead of playing guessing game on variable.
`testCaseName` is readable over `name`

PR #25391, opened 2026-01-19
Mcp oauth

Vishnuujain:
<!--
Thank you for your contribution!
Unless your change is trivial, please create an issue to discuss the change before creating a PR.
-->

### Describe your changes:
## MCP OAuth Implementation                                                                                                                                                        
                                                                                                                                                                                     
  This PR implements OAuth 2.0 Authorization Code Flow with PKCE for MCP integration.                                                                                                
                                                                                                                                                                                     
  ### Key Features                                                                                                                                                                   
  - OAuth 2.0 with PKCE (RFC 7636)                                                                                                                                                   
  - Google SSO + Basic Auth support                                                                                                                                                  
  - Refresh token rotation                                                                                                                                                           
  - Comprehensive audit logging                                                                                                                                                      
                                                                                                                                                                                     
  📖 **[Full Documentation →](https://github.com/open-metadata/OpenMetadata/blob/mcp-oauth/openmetadata-mcp/README.md)**  
  
Fixes <issue-number>

<!--
Short blurb explaining:
- What changes did you make?
- Why did you make them?
-->

I worked on ... because ...

<!-- For frontend related change, please add screenshots and/or videos of your changes preview! -->

---

----
## Summary by Gitar

- **Security hardening:**
  - XFF rate-limit bypass fix with IP format validation and map size capping
  - Bcrypt secret storage with atomic CAS rotation
  - ID token validation and null safety for PKCE/grant_type/refresh_token
- **Bug fixes:**
  - Fixed MCP OAuth SSO callback for Okta with redirect_uri and pac4j session attributes
  - Fixed missing return in MCP callback error path
  - Registration metadata length validation and loopback host check
- **Infrastructure:**
  - Database migrations (1.12.3) with oauth_audit_log table removal
  - OAuth token cleanup job and scheduler
  - Configuration schema and TypeScript type definitions added

### DIFF
--- a/openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/repository/OAuthAuthorizationCodeRepository.java
+++ b/openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/repository/OAuthAuthorizationCodeRepository.java
@@ -1,6 +1,10 @@
 package org.openmetadata.mcp.server.auth.repository;
 
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
 import java.util.List;
 import lombok.extern.slf4j.Slf4j;
 import org.openmetadata.schema.utils.JsonUtils;
@@ -33,7 +37,7 @@ public class OAuthAuthorizationCodeRepository {
       long expiresAt) {
 
     dao.insert(
-        code,
+        hashAuthCode(code),
         clientId,
         userName,
         codeChallenge,
@@ -50,7 +54,7 @@ public class OAuthAuthorizationCodeRepository {
    * Find authorization code by code value.
    */
   public OAuthAuthorizationCodeRecord findByCode(String code) {
-    return dao.findByCode(code);
+    return dao.findByCode(hashAuthCode(code));
   }
 
   /**
@@ -62,10 +66,11 @@ public class OAuthAuthorizationCodeRepository {
    * @return The updated record if successful, null if code was already used or doesn't exist
    */
   public OAuthAuthorizationCodeRecord markAsUsedAtomic(String code) {
-    int rowsAffected = dao.markAsUsedAtomic(code);
+    String hashed = hashAuthCode(code);
+    int rowsAffected = dao.markAsUsedAtomic(hashed);
     if (rowsAffected == 1) {
       LOG.debug("Atomically marked authorization code as used");
-      return dao.findByCode(code);
+      return dao.findByCode(hashed);
     }
     LOG.warn("Failed to atomically mark authorization code as used (already used or not found)");
     return null;
@@ -75,7 +80,7 @@ public class OAuthAuthorizationCodeRepository {
    * Delete authorization code.
    */
   public void delete(String code) {
-    dao.delete(code);
+    dao.delete(hashAuthCode(code));
     LOG.debug("Deleted authorization code");
   }
 
@@ -87,5 +92,15 @@ public class OAuthAuthorizationCodeRepository {
     long currentTime = System.currentTimeMillis();
     dao.deleteExpired(currentTime);
     LOG.info("Deleted expired authorization codes");
+  }
+
+  private String hashAuthCode(String code) {
+    try {
+      MessageDigest digest = MessageDigest.getInstance("SHA-256");
+      byte[] hash = digest.digest(code.getBytes(StandardCharsets.UTF_8));
+      return Base64.getEncoder().encodeToString(hash);
+    } catch (NoSuchAlgorithmException e) {
+      throw new RuntimeException("SHA-256 not available", e);
+    }
   }
 }


bootstrap/sql/migrations/native/1.12.1/mysql/postDataMigration.sql
```diff
@@ -0,0 +1,22 @@
+-- Register Claude Desktop as pre-configured OAuth client for MCP User SSO
+-- This enables Claude Desktop to authenticate users via OpenMetadata SSO without manual client registration
+INSERT INTO oauth_clients (client_id, client_name, redirect_uris, grant_types, scopes, created_at)
+VALUES (
+  'claude-desktop',
+  'Claude Desktop',
+  '["http://localhost:6274/callback", "http://127.0.0.1:6274/callback"]',
```
mohityadav766: what if this callback is different? we can register client dynamically and should not be inserting directly , DCR is the way in oath

Vishnuujain: I did this for locally testing the oauth flow, we support dynamic client registration but that's not fully implemented yet. I'll configure this with  /register endpoint.

Vishnuujain:
@gitar-bot can you review this oauth implementation in terms of general architecture, implementation, security, sso friendly and basically check for any missing pieces that should be ideally here to setup robust oauth support and also check if this properly supports basic auth.

bootstrap/sql/migrations/native/1.12.1/mysql/postDataMigration.sql
Vishnuujain: resolved here: https://github.com/open-metadata/OpenMetadata/pull/25391/commits/539905868ed050bfe29e11834fc8b62efa35b826

openmetadata-mcp/src/main/java/org/openmetadata/mcp/resources/RootOAuthEndpointsResource.java
```diff
@@ -0,0 +1,102 @@
...
+public class RootOAuthEndpointsResource {
+
+  private final OAuthHttpStatelessServerTransportProvider transport;
+
+  public RootOAuthEndpointsResource(OAuthHttpStatelessServerTransportProvider transport) {
+    this.transport = transport;
+  }
+
+  @GET
+  @Path("/.well-known/oauth-authorization-server")
```
mohityadav766: I saw same under OAuthEndpointsResource

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
```diff
@@ -0,0 +1,711 @@
...
+
+    // Create Authorization Server metadata (RFC 8414)
+    // Endpoints are relative to /mcp prefix since servlet is mounted there
+    OAuthMetadata metadata = new OAuthMetadata();
+    metadata.setIssuer(URI.create(baseUrl + mcpEndpoint));
+    metadata.setAuthorizationEndpoint(URI.create(baseUrl + mcpEndpoint + "/authorize"));
+    metadata.setTokenEndpoint(URI.create(baseUrl + mcpEndpoint + "/token"));
+    metadata.setRegistrationEndpoint(URI.create(baseUrl + mcpEndpoint + "/register"));
+    metadata.setScopesSupported(
+        List.of("openid", "profile", "email", "offline_access", "api://apiId/.default"));
```
mohityadav766: can you check with Azure , for azure this api:// is going to be different , not sure if this can be hardcoded, also some vendor might not support offline_access (can we validate against) different sso providers to be sure here!

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
```diff
@@ -0,0 +1,711 @@
...
+      HttpServletResponse response, String message, int statusCode) throws IOException {
+
+    // Build WWW-Authenticate header per RFC 6750 and MCP spec
+    StringBuilder challenge = new StringBuilder("Bearer");
+    challenge.append(" resource_metadata=\"").append(resourceMetadataUrl).append("\"");
+
+    challenge
+        .append(", scope=\"")
+        .append(
+            "openid profile email offline_access api://0a957c01-29f8-4fce-a1dc-3b9f12447b60/.default")
```
mohityadav766: this is hardcoded

openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SearchMetadataTool.java
```diff
@@ -25,6 +26,7 @@
 import org.openmetadata.service.security.policyevaluator.SubjectContext;
 
 @Slf4j
+@RequireScope({"metadata:read"})
```
mohityadav766: qq : how is this relayed to clients?

mohityadav766: are we controlling the user permission for tools via some other wy?

openmetadata-service/src/main/java/org/openmetadata/service/config/MCPConfiguration.java
```diff
@@ -24,4 +26,13 @@ public class MCPConfiguration {
 
   @JsonProperty("originHeaderUri")
   private String originHeaderUri = "http://localhost";
+
+  /**
+   * List of allowed origins for CORS on OAuth endpoints.
+   * Use specific origins for production security. Wildcard (*) is NOT recommended.
+   * Default includes common development origins.
+   */
+  @JsonProperty("allowedOrigins")
```
mohityadav766, line 45: in case of clusters , this can't be hardcoded ? or am i missing something

openmetadata-mcp/src/main/java/org/openmetadata/mcp/resources/RootOAuthEndpointsResource.java
Vishnuujain: thanks for flagging, deleted this file

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
Vishnuujain: added getSupportedScopesForProvider() method that handles this

openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SearchMetadataTool.java
Vishnuujain: removed the @RequireScope system entirely, this keeps v1 simple while maintaining security through OM existing permission model

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
Vishnuujain: sorry missed this, fixed now

openmetadata-service/src/main/java/org/openmetadata/service/config/MCPConfiguration.java
Vishnuujain: these are just default values. In prod/cluster deployment, we can override this via openmetadata.yaml configuration and also added a baseUrl property to MCPConfiguration for explicit cluster URL configuration

bootstrap/sql/migrations/native/1.12.1/mysql/postDataMigration.sql
```diff
@@ -0,0 +1,22 @@
+-- Register Claude Desktop as pre-configured OAuth client for MCP User SSO
+-- This enables Claude Desktop to authenticate users via OpenMetadata SSO without manual client registration
+INSERT INTO oauth_clients (client_id, client_name, redirect_uris, grant_types, scopes, created_at)
```
pmbrull: this should be 1.12.0 right

Vishnuujain: yes, moved tables to 1.12.0

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/handlers/AuthorizationHandler.java
```diff
@@ -0,0 +1,211 @@
+package org.openmetadata.mcp.server.auth.handlers;
+
+import java.net.URI;
```
aji-aju, line 3: Shouldn't we use the underlying authentication configuration customer has set up for mcp also?[4:19 PM]Also can't we reuse the existing APIs for auth rather than creating new APIs, bypassing the current auth configuration?

bootstrap/sql/migrations/native/1.12.0/mysql/postDataMigrationSQLScript.sql
```diff
@@ -56,3 +56,10 @@ SET json = JSON_ARRAY_APPEND(
 )
 WHERE name = 'tableDataToBeFresh'
   AND NOT JSON_CONTAINS(JSON_EXTRACT(json, '$.parameterDefinition[*].name'), '"timezone"');
+
+-- Set authType to 'basic' for existing Snowflake connections that don't have it
+-- This ensures backward compatibility after adding OAuth support to Snowflake connector
+UPDATE database_service
+SET json = JSON_SET(json, '$.connection.config.authType', 'basic')
```
pmbrull: why are we adding this?

openmetadata-mcp/src/main/java/org/openmetadata/mcp/McpAuthFilter.java
```diff
@@ -66,12 +101,27 @@ public void doFilter(
     }
   }
 
-  private void sendError(HttpServletResponse response, String errorMessage) throws IOException {
+  /**
+   * Check if the request path is an OAuth endpoint that should be publicly accessible.
+   * OAuth discovery and flow endpoints must be accessible without authentication.
+   */
+  private boolean isOAuthEndpoint(String path) {
```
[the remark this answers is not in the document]
Vishnuujain: MCP spec requires open client registration (RFC 7591), so authentication cannot be applied to this endpoint. As mitigation, IP-based rate limiting (10 registrations per IP per hour) was added. Added a design comment explaining the tradeoff and noting that database-backed rate limiting is recommended for multi-node production deployments

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
```diff
@@ -0,0 +1,764 @@
...
+
+  private final JwtFilter jwtFilter;
+
+  private volatile List<String> allowedOrigins;
+
+  private final OAuthAuthorizationServerProvider authProvider;
+
+  // In-memory rate limiter for client registration: max 10 registrations per IP per hour
+  private static final int REGISTRATION_MAX_PER_HOUR = 10;
+  private final java.util.concurrent.ConcurrentHashMap<
```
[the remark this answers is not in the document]
Vishnuujain: Acknowledged as a known limitation for v1. Added a comment documenting that the in-memory rate limiter is per-JVM-instance (effective limit is 10×N nodes in clusters) and recommending database-backed rate limiting via COUNT(*) FROM oauth_clients WHERE created_at > NOW() - INTERVAL 1 HOUR for multi-node production deployments. Will address in a follow-up if needed

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/handlers/RegistrationHandler.java
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Handle a client registration request.
+   *
+   * <p>Validates the registration metadata and generates client credentials.
+   *
+   * @param metadata The client registration metadata
+   * @return CompletableFuture resolving to registered client information
+   */
+  public CompletableFuture<OAuthClientInformation> handle(OAuthClientMetadata metadata) {
```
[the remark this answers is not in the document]
Vishnuujain: Already addressed. Rate limiting was added to the registration endpoint in the transport layer (OAuthHttpStatelessServerTransportProvider). The handleRegistrationRequest() method calls isRegistrationRateLimited(clientIp) which enforces a limit of 10 registrations per IP per hour using an in-memory ConcurrentHashMap with a synchronized sliding window. Requests exceeding the limit receive HTTP 429 with too_many_requests error.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/provider/UserSSOOAuthProvider.java
```diff
@@ -0,0 +1,933 @@
...
+      long newRefreshExpiresAt = System.currentTimeMillis() + (REFRESH_TOKEN_EXPIRY_SECONDS * 1000);
+
+      RefreshToken newRefreshToken =
+          new RefreshToken(
+              newRefreshTokenValue, client.getClientId(), requestedScopes, newRefreshExpiresAt);
+
+      // Store new refresh token before revoking old one
+      // This ensures atomicity: if revocation fails, both tokens work briefly
+      // If storage fails, old token still works - no lock-out
+      tokenRepository.storeRefreshToken(
```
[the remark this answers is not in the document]
Vishnuujain: intentional design tradeoff, the code already documents this at lines 822-824

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/validators/IdTokenValidator.java
```diff
@@ -0,0 +1,372 @@
...
+    private final LoadingCache<String, Jwk> cache;
+
+    public CachingMultiUrlJwkProvider(List<URL> publicKeyUrls) {
+      this.jwkProviders =
+          publicKeyUrls.stream().map(UrlJwkProvider::new).collect(Collectors.toList());
+
+      this.cache =
+          CacheBuilder.newBuilder()
+              .maximumSize(10)
+              .expireAfterWrite(24, TimeUnit.HOURS)
```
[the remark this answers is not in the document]
Vishnuujain: The 24-hour TTL is a pragmatic default that balances performance (avoiding JWKS endpoint calls on every request) against key rotation responsiveness. In practice, major SSO providers (Google, Okta, Azure) rotate keys on a schedule of days/weeks, not hours, and announce new keys well before retiring old ones.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/handlers/SSOCallbackServlet.java
```diff
@@ -0,0 +1,443 @@
...
+  @Override
+  protected void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws ServletException, IOException {
+    try {
+      LOG.info("Handling SSO callback for MCP OAuth");
+
+      // Get state from request (Google returns pac4j's state)
+      String pac4jState = request.getParameter("state");
+      String idTokenParam = request.getParameter("id_token");
+      LOG.debug(
```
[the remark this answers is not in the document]
Vishnuujain: These are single-use CSRF tokens logged at DEBUG level only. DEBUG logging is disabled in production by default

openmetadata-service/src/main/java/org/openmetadata/service/security/AuthenticationCodeFlowHandler.java
```diff
@@ -127,6 +131,7 @@ public class AuthenticationCodeFlowHandler implements AuthServeletHandler {
   public static final String DEFAULT_PRINCIPAL_DOMAIN = "openmetadata.org";
   public static final String OIDC_CREDENTIAL_PROFILE = "oidcCredentialProfile";
   public static final String SESSION_REDIRECT_URI = "sessionRedirectUri";
+  public static final String SESSION_SSO_CALLBACK_URL = "googleCallbackUrl";
```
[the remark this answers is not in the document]
Vishnuujain: This is a session attribute key string, not user-facing. The Java constant name SESSION_SSO_CALLBACK_URL is already provider-agnostic

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
```diff
@@ -0,0 +1,902 @@
+package org.openmetadata.mcp.server.transport;
```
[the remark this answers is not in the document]
Vishnuujain: MCP clients (Claude, Cursor, VS Code) must self-register without prior credentials, this is the standard MCP OAuth pattern

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/provider/UserSSOOAuthProvider.java
```diff
@@ -0,0 +1,1064 @@
...
+                  + "Authorization used '{}' but token request used '{}'",
+              storedRedirectUri,
+              requestRedirectUri);
+          throw new TokenException(
+              "invalid_grant",
+              "Redirect URI in token request does not match authorization request");
+        }
+      }
+
+      if (!verifyPKCE(codeVerifier, usedRecord.codeChallenge())) {
```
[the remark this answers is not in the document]
Vishnuujain: not a bug, intentional security design per RFC 6749: authorization codes MUST be single-use, mark this as resolved

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
```diff
@@ -0,0 +1,902 @@
...
+  }
+
+  /**
+   * Sets CORS headers with origin validation.
+   * Only allows specific origins from the allowedOrigins list.
+   * Rejects requests from origins not in the allowed list.
+   * @param request The HTTP request
+   * @param response The HTTP response
+   */
+  private void setCorsHeaders(HttpServletRequest request, HttpServletResponse response) {
```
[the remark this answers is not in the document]
Vishnuujain: already fixed, ignore

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/handlers/RegistrationHandler.java
```diff
@@ -0,0 +1,206 @@
...
+ * <p>Processes client registration requests and generates client credentials.
+ */
+@Slf4j
+public class RegistrationHandler {
+
+  private static final SecureRandom SECURE_RANDOM = new SecureRandom();
+  private static final int CLIENT_SECRET_BYTES = 32;
+  // Block dangerous schemes; allow http, https, and private-use URI schemes
+  // per RFC 8252 Section 7.1 (e.g. cursor://, vscode://, claude-desktop://)
+  private static final Set<String> BLOCKED_REDIRECT_SCHEMES =
```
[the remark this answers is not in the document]
Vishnuujain: design choice: native MCP clients like Cursor use private-use URI schemes (cursor://). An allowlist of http/https only would break all native app MCP clients. The blocklist rejects dangerous schemes (javascript:, data:, file:, blob:, vbscript:) while allowing legitimate custom schemes, mark resolved

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
```diff
@@ -0,0 +1,902 @@
...
+          tokenAttempts.computeIfAbsent(
+              clientIp, k -> new java.util.concurrent.atomic.AtomicInteger(0));
+      return count.incrementAndGet() > TOKEN_MAX_PER_MINUTE;
+    }
+  }
+
+  private void handleRegistrationRequest(HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+    try {
+      String clientIp = request.getRemoteAddr();
```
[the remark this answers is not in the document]
Vishnuujain: known limitation, documented in code comments

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/provider/UserSSOOAuthProvider.java
```diff
@@ -0,0 +1,1064 @@
...
+      ssoHandler.handleLogin(wrappedRequest, currentResponse.get());
+
+      // After handleLogin(), pac4j has stored its state in the session
+      // Extract pac4j session attributes and store in database
+      // Note: pac4j stores State and CodeVerifier as objects, not strings
+      String pac4jState = null;
+      String pac4jNonce = null;
+      String pac4jCodeVerifier = null;
+
+      java.util.Enumeration<String> attrNames = session.getAttributeNames();
```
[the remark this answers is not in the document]
Vishnuujain: DEBUG logging is disabled in production

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/handlers/SSOCallbackServlet.java
```diff
@@ -0,0 +1,453 @@
...
+        return authzConfig.getPrincipalDomain();
+      }
+      LOG.debug("Principal domain not configured, using empty string");
+    } catch (Exception e) {
+      LOG.warn("Failed to get principal domain from config: {}", e.getMessage());
+    }
+    return "";
+  }
+
+  private IdTokenValidator getIdTokenValidator() {
```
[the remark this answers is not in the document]
Vishnuujain: Already correct, ignore this

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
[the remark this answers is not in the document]
Vishnuujain: mark resolved and close

Vishnuujain:
@gitar-bot please mark this issue as resolved: Unauthenticated dynamic client registration endpoint, this is not relevant and don't raise this again, told you the reason several times

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/repository/OAuthClientRepository.java
```diff
@@ -0,0 +1,107 @@
...
+    this.dao = Entity.getCollectionDAO().oauthClientDAO();
+    this.fernet = Fernet.getInstance();
+  }
+
+  /**
+   * Register a new OAuth client.
+   */
+  public void register(OAuthClientInformation clientInfo) {
+    String clientSecret = clientInfo.getClientSecret();
+    String encryptedSecret = clientSecret != null ? fernet.encrypt(clientSecret) : null;
```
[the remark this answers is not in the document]
Vishnuujain: Fernet is OM's standard secret encryption used for all secrets, ignore this and mark resolved

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
```diff
@@ -0,0 +1,911 @@
...
+        .forEach(
+            (key, values) -> {
+              if (values.length > 0) {
+                params.put(key, values[0]);
+              }
+            });
+
+    LOG.debug("Token request params (sanitized): {}", sanitizeParamsForLogging(params));
+
+    String clientIp = request.getRemoteAddr();
```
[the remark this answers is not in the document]
Vishnuujain: known per-JVM limitation, documented in code, skip and resolve

Vishnuujain:
@gitar-bot Bug: Glue script parser regexes miss chained calls, overcapture args, this is not from this PR, please ignore it and don't raise stupid errors for now!!

bootstrap/sql/migrations/native/1.12.2/mysql/schemaChanges.sql
```diff
@@ -22,3 +22,166 @@ ALTER TABLE glossary_term_entity
 
 -- Add index for efficient entityStatus filtering
 CREATE INDEX idx_glossary_term_entity_status ON glossary_term_entity (entityStatus);
+
+-- OAuth 2.0 Persistence Schema for MCP Server Integration (MySQL)
+-- This migration adds tables to support OAuth 2.0 authorization flow for MCP server authentication
+
+-- OAuth Clients Table
+-- Stores registered OAuth 2.0 client applications
+CREATE TABLE IF NOT EXISTS oauth_clients (
```
aji-aju: Is this needed? This info is already available in openmetadata_settings right?

Vishnuujain: openmetadata_settings stores singleton config, oauth_clients stores multi-row entities with FK relationships like single user might have oauth token for different clients like claude, cursor/vs-code

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/provider/UserSSOOAuthProvider.java
```diff
@@ -0,0 +1,1072 @@
...
+            client.getClientId());
+        throw new TokenException(
+            "invalid_grant",
+            "Authorization code expired "
+                + expiredSecondsAgo
+                + " seconds ago. Please restart the authorization flow.");
+      }
+
+      // Atomically mark as used — prevents replay attacks
+      OAuthAuthorizationCodeRecord usedRecord = codeRepository.markAsUsedAtomic(code);
```
mohityadav766, line 662: **🔴 Critical: Auth code consumed before PKCE verification**

The authorization code is atomically marked as `used` here (line 656), but PKCE verification happens later (~line 703). If PKCE verification fails, the code is already burned — the legitimate client **cannot retry** and must restart the entire OAuth flow from scratch.

This creates a bad UX and a potential denial-of-service vector: an attacker who knows a valid authorization code can submit it with a wrong `code_verifier`, burning the code and forcing the real user to re-authenticate.

**Suggested fix:** Reorder to: validate expiry → validate PKCE → mark as used. Or at minimum, only mark as used *after* all validations pass.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/HttpServletStatelessServerTransport.java
```diff
@@ -0,0 +1,336 @@
...
+   * @param request The HTTP servlet request
+   * @param response The HTTP servlet response
+   * @throws ServletException If a servlet-specific error occurs
+   * @throws IOException If an I/O error occurs
+   */
+  @Override
+  protected void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws ServletException, IOException {
+    String requestURI = request.getRequestURI();
+    if (!requestURI.endsWith(mcpEndpoint)) {
```
mohityadav766: **🔴 High: `endsWith()` path matching allows route confusion**

`endsWith()` is used here and at line 118 for route matching. A crafted request path like `/evil/mcp` or `/../../mcp` would match `/mcp`, potentially bypassing upstream auth filters that use exact path matching.

**Suggested fix:** Use `requestURI.equals(mcpEndpoint)` or at minimum `requestURI.equals(contextPath + mcpEndpoint)` for strict matching, consistent with the fix already applied to the auth filter.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/handlers/RegistrationHandler.java
```diff
@@ -0,0 +1,241 @@
...
+
+  /**
+   * Handle a client registration request.
+   *
+   * <p>Validates the registration metadata and generates client credentials.
+   *
+   * @param metadata The client registration metadata
+   * @return CompletableFuture resolving to registered client information
+   */
+  public CompletableFuture<OAuthClientInformation> handle(OAuthClientMetadata metadata) {
```
mohityadav766, line 55: **🔴 High: Unauthenticated dynamic client registration**

The `/mcp/register` endpoint requires no authentication. The IP-based rate limit (10/hour) is insufficient because:
1. IP rotation or cloud NAT trivially bypasses it
2. In-memory counters are per-JVM — a cluster of N nodes allows `10 × N` registrations/hour
3. `getRemoteAddr()` returns the proxy IP behind a load balancer, so all real clients share one limit

Over time, this accumulates unbounded orphaned clients with no lifecycle management (no expiry, no admin pruning).

**Suggested fix:** Require admin authentication or a registration access token (per RFC 7591 §1.2). At minimum, add a configurable toggle to disable open registration in production.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/repository/OAuthTokenRepository.java
```diff
@@ -0,0 +1,235 @@
...
+    LOG.debug("Revoked refresh token");
+  }
+
+  /**
+   * Rotate a refresh token: revoke all existing tokens for the user/client, then store
+   * the new one. These are two sequential DB operations (not wrapped in a single transaction).
+   * If the server crashes or insert fails after revoke, the user must re-authenticate via SSO
+   * — this is the safer failure mode compared to having two valid tokens simultaneously.
+   */
+  public void rotateRefreshToken(
```
mohityadav766, line 166: **🔴 High: Non-atomic refresh token rotation — race condition**

The `rotateRefreshToken` method performs `revokeAllForUser` (line 187) then `insert` (line 192) as two sequential DB operations without a transaction. The code comments acknowledge this.

If two concurrent refresh requests arrive with the same token, both can load and validate the token before either revokes it, enabling a **token replay attack** where an attacker who captured a refresh token races the legitimate client.

**Suggested fix:** Wrap in a database transaction, or use an atomic CAS pattern:
```sql
UPDATE oauth_tokens SET revoked = TRUE 
WHERE token_hash = :hash AND revoked = FALSE
RETURNING *
```
This ensures only one caller succeeds.

bootstrap/sql/migrations/native/1.12.2/mysql/schemaChanges.sql
```diff
@@ -22,3 +22,166 @@ ALTER TABLE glossary_term_entity
...
+    used BOOLEAN NOT NULL DEFAULT FALSE,
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CONSTRAINT oauth_authorization_codes_fk_client
+        FOREIGN KEY (client_id) REFERENCES oauth_clients(client_id) ON DELETE CASCADE,
+    CONSTRAINT oauth_authorization_codes_expires_check CHECK (expires_at > 0),
+    CONSTRAINT oauth_authorization_codes_code_check CHECK (CHAR_LENGTH(code) > 0),
+    CONSTRAINT oauth_authorization_codes_user_check CHECK (CHAR_LENGTH(user_name) > 0),
+    CONSTRAINT oauth_authorization_codes_challenge_method_check CHECK (
+        code_challenge_method IS NULL OR
+        code_challenge_method IN ('plain', 'S256')
```
mohityadav766: **🟡 Medium: `plain` PKCE allowed in DB CHECK constraint**

```sql
code_challenge_method IN ('plain', 'S256')
```

While the application layer correctly rejects non-S256, the database permits `plain`. The `plain` challenge method provides **zero security** against authorization code interception (it's the code verifier sent in the clear). This is a defense-in-depth violation.

Same issue exists in the PostgreSQL migration.

**Suggested fix:** Remove `'plain'` from the CHECK constraint so only `'S256'` is accepted at the database level.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/middleware/ClientAuthenticator.java
```diff
@@ -0,0 +1,73 @@
...
+              // If client has a secret, verify it using constant-time comparison
+              if (client.getClientSecret() != null) {
+                if (clientSecret == null) {
+                  return CompletableFuture.failedFuture(
+                      new AuthenticationException("Client secret required"));
+                }
+
+                // Use MessageDigest.isEqual() for constant-time comparison to prevent timing
+                // attacks
+                byte[] expectedBytes = client.getClientSecret().getBytes(StandardCharsets.UTF_8);
```
mohityadav766: **🟡 Medium: Client secrets stored with reversible encryption, not hashed**

The client secret is decrypted from Fernet (`OAuthClientRepository` line 52) and compared as plaintext here using `MessageDigest.isEqual()`. While timing-safe comparison is good, Fernet is **reversible encryption** — if the Fernet key is compromised, all client secrets are exposed at once.

Since client secrets only need to be **verified**, never recovered, they should be stored as a one-way hash.

**Suggested fix:** Store client secrets as bcrypt/argon2 hashes. Compare using the hash verification function instead of decrypting.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/provider/UserSSOOAuthProvider.java
```diff
@@ -0,0 +1,1072 @@
...
+            client.getClientId(),
+            clientIP);
+        throw new TokenException(
+            "invalid_grant", "Authorization code was issued to a different client");
+      }
+
+      // CRITICAL: Validate redirect URI matches (RFC 6749 Section 4.1.3)
+      String storedRedirectUri = usedRecord.redirectUri();
+      URI requestRedirectUri = authCode.getRedirectUri();
+      if (storedRedirectUri != null && !storedRedirectUri.isEmpty()) {
```
mohityadav766: **🟡 Medium: `redirect_uri` validation skipped when stored value is null/empty**

Per RFC 6749 §4.1.3, if the `redirect_uri` parameter was included in the authorization request, it **MUST** be included and match in the token request. This condition:

```java
if (storedRedirectUri != null && !storedRedirectUri.isEmpty())
```

skips validation entirely when `storedRedirectUri` is null or empty — which could happen if the authorization request didn't store it properly. An attacker could exploit this to exchange a stolen authorization code from a different redirect URI.

Same pattern exists in `OAuthHttpStatelessServerTransportProvider.java` line 610.

**Suggested fix:** The authorization code should always store the redirect_uri (even if null means "use default"). If the original request included a redirect_uri, it must be required and matched in the token exchange.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java
mohityadav766: **🟡 Medium: Rate limiting broken behind reverse proxies**

`request.getRemoteAddr()` returns the **proxy/load balancer IP** in production deployments, not the real client IP. This means:
1. All clients behind the same proxy share one rate limit bucket
2. A single malicious client can exhaust the rate limit for all legitimate users
3. Conversely, an attacker directly connecting bypasses the shared limit

Same issue at line 747 (registration endpoint).

**Suggested fix:** Use a trusted `X-Forwarded-For` header with configurable trusted proxy list. Example:
```java
String clientIp = Optional.ofNullable(request.getHeader("X-Forwarded-For"))
    .map(xff -> xff.split(",")[0].trim())
    .orElse(request.getRemoteAddr());
```

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/provider/UserSSOOAuthProvider.java
```diff
@@ -0,0 +1,1072 @@
...
+          return baseUrlConfig.getOpenMetadataUrl();
+        }
+      }
+    } catch (Exception e) {
+      LOG.warn("Failed to get base URL from system settings: {}", e.getMessage());
+    }
+
+    // TODO: Silently falling back to localhost:8585 will cause subtle failures in production.
+    //  Consider throwing an exception or logging at ERROR level to surface misconfiguration.
+    return "http://localhost:8585";
```
mohityadav766, line 175: **🟡 Medium: Silent localhost fallback in production**

If the MCP base URL is not configured, this silently falls back to `http://localhost:8585`. In production, this would:
1. Generate tokens with an incorrect `issuer` claim
2. Build OAuth metadata URLs pointing to localhost
3. Create subtle, hard-to-debug failures

The TODO on line 171 acknowledges this should be an error.

Same fallback exists in `McpServer.java` line 324.

**Suggested fix:** Throw a clear configuration error at startup if `baseUrl` is not set, rather than silently degrading.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/util/UriUtils.java
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Validate that the issuer URL meets OAuth 2.0 requirements.
+   * @param url The issuer URL to validate
+   */
+  public static void validateIssuerUrl(URI url) {
+    // RFC 8414 requires HTTPS, but we allow localhost HTTP for testing
+    String scheme = url.getScheme();
+    String host = url.getHost();
+
+    if (!"https".equals(scheme) && !"localhost".equals(host) && !"127.0.0.1".equals(host)) {
```
mohityadav766: **🟡 Medium: Missing IPv6 loopback in issuer URL validation**

This check allows HTTP only for `localhost` and `127.0.0.1`, but misses the IPv6 loopback address `[::1]`. A development server bound to `::1` with an HTTP issuer URL would be incorrectly rejected.

**Suggested fix:**
```java
if (!"https".equals(scheme) 
    && !"localhost".equals(host) 
    && !"127.0.0.1".equals(host)
    && !"[::1]".equals(host))
```

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/handlers/RegistrationHandler.java
```diff
@@ -0,0 +1,241 @@
...
+   * Validate registration request per RFC 7591.
+   *
+   * @param metadata The client registration metadata
+   * @throws RegistrationException if validation fails
+   */
+  private static final int MAX_STRING_FIELD_LENGTH = 255;
+
+  private static final int MAX_URI_FIELD_LENGTH = 2048;
+
+  private void validateRegistrationRequest(OAuthClientMetadata metadata)
```
mohityadav766, line 135: **🟡 Medium: Unbounded list sizes in client registration metadata**

Field length validation is present for string fields (good), but list fields like `redirect_uris`, `contacts`, `grant_types`, and `response_types` have **no size limit**. Through the unauthenticated registration endpoint, an attacker could submit a request with thousands of redirect URIs or contacts, causing storage exhaustion.

**Suggested fix:** Add a maximum list size check (e.g., 10 redirect URIs, 5 contacts) in the validation method.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/provider/UserSSOOAuthProvider.java
```diff
@@ -0,0 +1,1072 @@
...
+
+      RefreshToken refreshToken =
+          new RefreshToken(
+              refreshTokenValue, client.getClientId(), usedRecord.scopes(), refreshExpiresAt);
+
+      // Store refresh token in database (hashed and encrypted)
+      tokenRepository.storeRefreshToken(
+          refreshToken, client.getClientId(), userName, usedRecord.scopes());
+
+      LOG.info(
```
mohityadav766: **ℹ️ Low: PII logged at INFO level**

Usernames are logged at INFO level here and at line 883. In production, INFO logs are typically always enabled and may be shipped to centralized logging systems, SIEM tools, or third-party services.

**Suggested fix:** Use DEBUG level for logs containing usernames/emails, or redact them (e.g., `m***@example.com`).

aniketkatkar97 (approved):

bootstrap/sql/migrations/native/1.12.2/postgres/schemaChanges.sql
```diff
@@ -27,3 +27,248 @@ ALTER TABLE glossary_term_entity
...
+CREATE INDEX IF NOT EXISTS idx_oauth_access_tokens_client_id ON oauth_access_tokens(client_id);
+CREATE INDEX IF NOT EXISTS idx_oauth_access_tokens_user ON oauth_access_tokens(user_name);
+CREATE INDEX IF NOT EXISTS idx_oauth_access_tokens_expires_at ON oauth_access_tokens(expires_at);
+CREATE INDEX IF NOT EXISTS idx_oauth_access_tokens_created_at ON oauth_access_tokens(created_at);
+
+COMMENT ON TABLE oauth_access_tokens IS 'Active OAuth 2.0 access tokens for API authentication';
+COMMENT ON COLUMN oauth_access_tokens.token_hash IS 'SHA-256 hash of the access token for lookup';
+COMMENT ON COLUMN oauth_access_tokens.access_token_encrypted IS 'Encrypted access token value';
+COMMENT ON COLUMN oauth_access_tokens.user_name IS 'Username of the user who authorized the token';
+COMMENT ON COLUMN oauth_access_tokens.expires_at IS 'Unix timestamp (milliseconds) when token expires';
```
pmbrull: nit, do we need these comments? I dont think they are adding any value

bootstrap/sql/migrations/native/1.12.0/mysql/postDataMigrationSQLScript.sql
Vishnuujain: removed old leftover's, thanks for flagging

bootstrap/sql/migrations/native/1.12.2/postgres/schemaChanges.sql
Vishnuujain: yeah, removed them

Vishnuujain:
@gitar-bot can you check again the issues flagged, most of them already have comments or fixes in place so you can ignore those

bootstrap/sql/migrations/native/1.12.3/mysql/schemaChanges.sql
```diff
@@ -1,3 +1,97 @@
...
+    grant_types JSON NOT NULL DEFAULT ('["authorization_code", "refresh_token"]'),
+    token_endpoint_auth_method VARCHAR(50) NOT NULL DEFAULT 'client_secret_post',
+    scopes JSON NOT NULL DEFAULT ('["read", "write"]'),
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CONSTRAINT oauth_clients_client_id_check CHECK (CHAR_LENGTH(client_id) > 0),
+    INDEX idx_oauth_clients_client_id (client_id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+
+CREATE TABLE IF NOT EXISTS oauth_authorization_codes (
+    id CHAR(36) PRIMARY KEY DEFAULT (UUID()),
```
harshach, line 18: should these be encoded with fernet for encryption?

Vishnuujain: auth_codes were stored as plaintext because they're short-lived, single-use, and protected by PKCE. but we can add SHA-256 hashing for consistency with how access/refresh tokens are looked up. Fernet encryption isn't needed here though.

PR #25431, opened 2026-01-21
fix(bigquery): ensure usageLocation is added to connection url

khanal-cm:
### Describe your changes:
 - For BigQuery services, the connection URL ignored the configured usageLocation, causing Test Connection to fail when
  a region is required, for example in my case with s3ns policy (queries must be done in europe-west9 region in my projects, but by default is done in another region).
  - Patch `get_connection_url` so it always appends `location=<usageLocation>` when present.
  - Keep the behaviour idempotent: if the URL already carries `location=`, leave it as-is.
Fixes <issue-number>

I worked on appending the BigQuery usageLocation to every generated connection URL because connection tests and ingestions fail in environments where the region must be specified. I verified the change with unit tests and a targeted ingestion run.

#
### Type of change:

- [ x ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ulixius9 (approved):

keshavmohta09:
Changes have been cherry-picked to the 1.11.10 branch.

PR #25458, opened 2026-01-22
Fix: Automator patch issue - timestamp with seconds

Khairajani:
---

## Summary by Gitar

- **Fixed datetime compatibility:**
  - Added `_normalize_datetime_strings()` function to ensure ISO 8601 timestamps include microseconds (.000000) for Java SimpleDateFormat compatibility
  - Function recursively processes JSON objects to normalize datetime strings (e.g., `2026-01-19T18:08:41Z` → `2026-01-19T18:08:41.000000Z`)
- **Updated patch generation:**
  - Modified `build_patch()` in `patch_request.py` to normalize datetime strings before generating patches
  - Ensures compatibility between Python ingestion layer and Java backend's SimpleDateFormat parser

---


ingestion/src/metadata/ingestion/models/patch_request.py
```diff
@@ -25,6 +26,41 @@
...
 logger = logging.getLogger("metadata")
 
+# Regex pattern to match ISO 8601 datetime strings without microseconds
+# Matches: 2026-01-19T18:08:41Z or 2026-01-19T18:08:41+00:00
+ISO_DATETIME_WITHOUT_MICROSECONDS = re.compile(
+    r"^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})([Z]|[+-]\d{2}:\d{2})$"
+)
+
+
+def _normalize_datetime_strings(obj: Any) -> Any:
```
pmbrull, line 36: not sure if we have any time/datetime utilities module rather than having this on patch_request only. might be useful elsewhere

Khairajani:
This is the right fix
https://github.com/open-metadata/OpenMetadata/pull/25643

PR #25537, opened 2026-01-26
Tag recognizer pagination

edg956:
### Describe your changes:

This PR includes an endpoint that returns a tag's recognizers with pagination

#
### Type of change:

- [ ] Bug fix
- [X] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

- [X] For connector/ingestion changes: I updated the documentation.

---

## Summary by Gitar

- **New API endpoints:**
  - `GET /v1/tags/{id}/recognizers` and `GET /v1/tags/name/{fqn}/recognizers` enable paginated retrieval of tag recognizers
- **Cursor-based pagination:**
  - `TagRepository.getRecognizersOfTag()` implements bidirectional cursor navigation with `before`/`after` parameters
- **New exception handling:**
  - `BadCursorException` validates and handles malformed pagination cursors returning HTTP 400

---

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TagRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TagRepository.java
@@ -1060,33 +1060,65 @@ public class TagRepository extends EntityRepository<Tag> {
     }
 
     if (before != null) {
-      result = listRecognizersBefore(tag.getRecognizers(), before, limit);
+      result = listRecognizersBeforeCursor(tag.getRecognizers(), before, limit);
     } else {
-      result = listRecognizersAfter(tag.getRecognizers(), after, limit);
+      result = listRecognizersAfterCursor(tag.getRecognizers(), after, limit);
     }
 
     return result;
   }
 
-  private ResultList<Recognizer> listRecognizersAfter(
-      List<Recognizer> recognizers, String after, int limit) {
-    int total = recognizers.size();
-    String afterId = null;
-
-    if (after != null) {
+  private UUID extractIdFromCursor(String cursor) {
+    UUID id = null;
+    if (cursor != null) {
       try {
-        Map<String, String> map = parseCursorMap(RestUtil.decodeCursor(after));
-        afterId = map.get("id");
+        Map<String, String> map = parseCursorMap(RestUtil.decodeCursor(cursor));
+        String idString = map.get("id");
 
-        if (afterId == null) {
-          throw new BadCursorException("Invalid `after` cursor");
+        if (idString == null) {
+          throw new BadCursorException();
         }
+
+        id = UUID.fromString(idString);
+
       } catch (Exception e) {
-        throw new BadCursorException("Invalid `after` cursor");
+        throw new BadCursorException();
       }
     }
+    return id;
+  }
 
-    boolean append = afterId == null;
+  private ResultList<Recognizer> listRecognizersAfterCursor(
+      List<Recognizer> recognizers, String after, int limit) {
+    UUID afterId;
+
+    try {
+      afterId = extractIdFromCursor(after);
+    } catch (BadCursorException ignored) {
+      throw new BadCursorException("Invalid `after` cursor");
+    }
+
+    return listRecognizersAfter(recognizers, afterId, limit);
+  }
+
+  private ResultList<Recognizer> listRecognizersBeforeCursor(
+      List<Recognizer> recognizers, String before, int limit) {
+    UUID beforeId;
+
+    try {
+      beforeId = extractIdFromCursor(before);
+    } catch (BadCursorException ignored) {
+      throw new BadCursorException("Invalid `before` cursor");
+    }
+
+    return listRecognizersAfter(recognizers.reversed(), beforeId, limit);
+  }
+
+  private ResultList<Recognizer> listRecognizersAfter(
+      List<Recognizer> recognizers, UUID startId, int limit) {
+    int total = recognizers.size();
+
+    boolean append = startId == null;
     limit = limit > 0 ? Math.min(total, limit) : total;
 
     List<Recognizer> result = new ArrayList<>(limit);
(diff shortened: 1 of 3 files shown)


edg956:
@gitar-bot 

> 🚨 Bug: Backward pagination passes wrong cursor parameter
> 📄 [openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TagRepository.java:561](https://github.com/open-metadata/OpenMetadata/pull/25537/files#diff-f4989360a42fc9773e79a9f8027db5e853ca969bee9e803e6011663f3f88519dR561)
> 
> The `listRecognizersBefore` method has a critical bug where it validates the `before` cursor but then passes the same raw `before` string to `listRecognizersAfter` as the `after` parameter (line 561). This causes two problems:
> 
> 1. The `beforeId` that was parsed and validated is never used
> 2. `listRecognizersAfter` receives `before` as its second param, which it will then try to parse again as an `after` cursor
> 
> The backward pagination logic is fundamentally broken because the cursor string is passed through unchanged.
> 
> **Suggested fix:** Pass `null` as the after cursor and rely on the reversed list to handle pagination naturally, OR refactor to encode the cursor properly for the reversed list:
> 
> ```java
> return listRecognizersAfter(recognizers.reversed(), null, limit);
> ```
> 
> Then handle the cursor-based starting position differently in the calling code.

This is by design. Cursors are nothing but a base64-encoded json containing an id and a name of the next recognizer. By validating in `listRecognizersBefore` I'm ensuring that if there's any error then it correctly says that it comes from `before`. Because the validation passed, I know it won't result in an error in `listRecognizersAfter`.

Would you prefer a refactor to avoid duplication?

> ⚠️ Bug: Unused `beforeId` variable - parsed cursor value is discarded
> 📄 [openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TagRepository.java:548](https://github.com/open-metadata/OpenMetadata/pull/25537/files#diff-f4989360a42fc9773e79a9f8027db5e853ca969bee9e803e6011663f3f88519dR548)
> 
> In the `listRecognizersBefore` method, the `beforeId` variable is parsed from the cursor at line 548, but is never used. The method then calls `listRecognizersAfter(recognizers.reversed(), before, limit)` at line 558, passing the original raw `before` cursor string instead of the validated `beforeId`.
> 
> This means:
> 
> 1. The cursor parsing at line 548 is effectively dead code
> 2. `beforeId` is extracted but never referenced again
> 3. The validation is performed twice (once here, once in `listRecognizersAfter`)
> 
> The method needs restructuring to properly use the parsed cursor value.

As mentioned before, this is on purpose. A refactor could be performed to do such thing

edg956:
> > @gitar-bot
> > > 🚨 Bug: Backward pagination passes wrong cursor parameter
> > > 📄 [openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TagRepository.java:561](https://github.com/open-metadata/OpenMetadata/pull/25537/files#diff-f4989360a42fc9773e79a9f8027db5e853ca969bee9e803e6011663f3f88519dR561)
> > > The `listRecognizersBefore` method has a critical bug where it validates the `before` cursor but then passes the same raw `before` string to `listRecognizersAfter` as the `after` parameter (line 561). This causes two problems:
> > > 
> > > 1. The `beforeId` that was parsed and validated is never used
> > > 2. `listRecognizersAfter` receives `before` as its second param, which it will then try to parse again as an `after` cursor
> > > 
> > > The backward pagination logic is fundamentally broken because the cursor string is passed through unchanged.
> > > **Suggested fix:** Pass `null` as the after cursor and rely on the reversed list to handle pagination naturally, OR refactor to encode the cursor properly for the reversed list:
> > > ```java
> > > return listRecognizersAfter(recognizers.reversed(), null, limit);
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Then handle the cursor-based starting position differently in the calling code.
> > 
> > 
> > This is by design. Cursors are nothing but a base64-encoded json containing an id and a name of the next recognizer. By validating in `listRecognizersBefore` I'm ensuring that if there's any error then it correctly says that it comes from `before`. Because the validation passed, I know it won't result in an error in `listRecognizersAfter`.
> > Would you prefer a refactor to avoid duplication?
> > > ⚠️ Bug: Unused `beforeId` variable - parsed cursor value is discarded
> > > 📄 [openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TagRepository.java:548](https://github.com/open-metadata/OpenMetadata/pull/25537/files#diff-f4989360a42fc9773e79a9f8027db5e853ca969bee9e803e6011663f3f88519dR548)
> > > In the `listRecognizersBefore` method, the `beforeId` variable is parsed from the cursor at line 548, but is never used. The method then calls `listRecognizersAfter(recognizers.reversed(), before, limit)` at line 558, passing the original raw `before` cursor string instead of the validated `beforeId`.
> > > This means:
> > > 
> > > 1. The cursor parsing at line 548 is effectively dead code
> > > 2. `beforeId` is extracted but never referenced again
> > > 3. The validation is performed twice (once here, once in `listRecognizersAfter`)
> > > 
> > > The method needs restructuring to properly use the parsed cursor value.
> > 
> > 
> > As mentioned before, this is on purpose. A refactor could be performed to do such thing
> 
> Thank you for the clarification! I understand the design intent now - validating the cursor early in `listRecognizersBefore` to provide clearer error messages about which parameter (`before` vs `after`) caused the issue.
> 
> However, the CI test failures (`TagResourceIT.test_recognizerPaginationEndpoint`) indicate the backward pagination is not working correctly - tests are expecting `Recognizer_19` but getting `Recognizer_29`. This suggests the cursor might not be functioning as intended.
> 
> Regarding the refactor: Yes, I'd recommend consolidating the cursor validation logic to avoid the duplication and make the flow clearer. You could:
> 
> 1. Extract cursor parsing/validation to a shared method
> 2. Have both methods call it with appropriate error context
> 3. This would maintain the clear error messages while reducing duplication
> 
> But first, it would be good to investigate why the test is failing - the pagination logic may need adjustment regardless of the validation approach.

That test doesn't fail anymore

PR #25546, opened 2026-01-27
Fix Playwright tests for column bulk operations and glossary spec

karanh37:
## Summary
- Fixed the `should select multiple columns and bulk edit` test in `ColumnBulkOperations.spec.ts` to target only non-collapsible row checkboxes using `tbody tr:not(:has(button.expand-button))`, avoiding accidental clicks on aggregate/expandable parent rows
- Fixed flaky glossary rename test in `Glossary.spec.ts` by adding `waitForLoadState('networkidle')` after save to allow sidebar refresh requests to complete before asserting the updated name

## Test plan
- [ ] Run `ColumnBulkOperations.spec.ts` multi-select test and verify it selects only single-occurrence columns
- [ ] Run `Glossary.spec.ts` rename test and verify it no longer flakes on name assertion

---

## Summary by Gitar

- **Test refactoring:**
  - Moved `visitColumnBulkOperationsPage` calls to `beforeEach` hooks across multiple test describe blocks in `ColumnBulkOperations.spec.ts`
  - Reduces redundant page navigation and improves test execution efficiency
- **Test flakiness fixes:**
  - Non-collapsible row selector prevents accidental selection of expandable parent rows
  - Network idle wait ensures glossary sidebar updates complete before assertions
  - Case-insensitive columnFQN comparison handles potential casing differences in column names

---

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ColumnBulkOperations.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ColumnBulkOperations.spec.ts
@@ -448,7 +448,9 @@ test.describe('Column Bulk Operations - Bulk Update Flow', () => {
     // Verify all updates have the correct displayName
     for (const update of updates) {
       expect(update.displayName).toBe(displayName);
-      expect(update.columnFQN).toContain(sharedColumnName);
+      expect(update.columnFQN?.toLowerCase()).toContain(
+        sharedColumnName.toLowerCase()
+      );
     }
   });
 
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ColumnBulkOperations.spec.ts
```diff
@@ -530,7 +532,9 @@ test.describe('Column Bulk Operations - Bulk Update Flow', () => {
     // Verify all updates have the correct displayName
     for (const update of updates) {
       expect(update.displayName).toBe(displayName);
-      expect(update.columnFQN).toContain(sharedColumnName);
+      expect(update.columnFQN?.toLowerCase()).toContain(
+        sharedColumnName.toLowerCase()
+      );
```
siddhant1, line 537: Are we doing this transformation is code as well?

karanh37: Since we are doing bulk operations on different assets, there is a possibility that some of the name's can be in lowercase while some can be in mixed case. So we have normalized this

aniketkatkar97: Can we please add a comment there explaining the reason for this?

aniketkatkar97 (approved):

PR #25604, opened 2026-01-29
Fix failing test for data contracts specs

shrabantipaul-collate:
### Describe your changes:
Fix failing playwright tests for DataContracts specs.
Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

---

## Summary by Gitar

- **Fixes race condition in Playwright E2E test**: Waits for API response (`/api/v1/search/query`) to complete before interacting with owner selection dropdown in DataContracts test
- **Improves test reliability with proper test IDs**: Adds `data-testid` attributes to SelectableList component (`selectable-virtual-list` and `selectable-list-item-*`) for more robust test selectors
- **Uses explicit visibility assertion**: Ensures the first selectable list item is visible before clicking, preventing intermittent test failures

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
@@ -84,7 +84,6 @@ import {
 import { navigateToPersonaWithPagination } from '../../utils/persona';
 import { settingClick } from '../../utils/sidebar';
 import { test } from '../fixtures/pages';
-import { merge } from 'lodash';
 
 // Define entities that support Data Contracts
 const entitiesWithDataContracts = [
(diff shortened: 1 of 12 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
```diff
@@ -178,7 +178,12 @@ test.describe('Data Contracts', () => {
           DATA_CONTRACT_DETAILS.description
         );
 
-        await page.getByTestId('select-owners').click();
+        await page.getByTestId('select-owners').click();  
+        
+        await page.locator('.rc-virtual-list-holder-inner li')
+          .first()
+          .waitFor({ state: 'visible' });
+
```
aniketkatkar97, line 195: Can we add and await the API call responsible for rendering the owners' list instead? That would be a robust solution here. Also, the selector used is not specific enough. Can you please check if we can use any data-testid or id?

shrabantipaul-collate: Fixed

PR #25619, opened 2026-01-29
ui: address column level & explore card feedbacks

shrabantipaul-collate:
### Describe your changes:
- When adding tags or glossary terms in the column side panel, the success/error message should appear inside the side drawer.
- Fix issue for image is not being rendered in description in the side panel drawer.
- Changed label and icon for "Applies to table column" in custom properties drawer
image
Screenshot 2026-01-29 at 6 30 18 PM
Screenshot 2026-01-30 at 12 16 05 PM

- fix description being cut
before:
Screenshot 2026-02-09 at 12 24 24 PM

after:
Screenshot 2026-02-09 at 12 26 22 PM

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Container/ContainerDataModel/ContainerDataModel.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Container/ContainerDataModel/ContainerDataModel.tsx
@@ -85,7 +85,9 @@ const ContainerDataModel: FC<ContainerDataModelProps> = ({
     useState<Column>();
   const [expandedRowKeys, setExpandedRowKeys] = useState<string[]>([]);
 
-  const schema = pruneEmptyChildren(dataModel?.columns ?? []);
+  const schema = useMemo(() => pruneEmptyChildren(dataModel?.columns ?? []), [
+    dataModel?.columns,
+  ]);
 
   // Sync displayed columns with GenericProvider for ColumnDetailPanel navigation
   useEffect(() => {
--- a/openmetadata-ui/src/main/resources/ui/src/components/Database/ColumnDetailPanel/ColumnDetailPanel.component.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Database/ColumnDetailPanel/ColumnDetailPanel.component.tsx
@@ -203,7 +203,7 @@ export const ColumnDetailPanel = <T extends ColumnOrTask = Column>({
     }
 
     return undefined;
-  }, [localToast.open]);
+  }, [localToast]);
 
   const [activeColumn, setActiveColumn] = useState<T | null>(column);
 
--- a/openmetadata-ui/src/main/resources/ui/src/components/DriveService/Worksheet/WorksheetColumnsTable/WorksheetColumnsTable.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/DriveService/Worksheet/WorksheetColumnsTable/WorksheetColumnsTable.tsx
@@ -92,7 +92,9 @@ function WorksheetColumnsTable() {
     };
   }, [permissions, worksheetDetails]);
 
-  const schema = pruneEmptyChildren(worksheetDetails?.columns ?? []);
+  const schema = useMemo(() => pruneEmptyChildren(worksheetDetails?.columns ?? []), [
+    worksheetDetails?.columns,
+  ]);
 
   // Sync displayed columns with GenericProvider for ColumnDetailPanel navigation
   useEffect(() => {


openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Entity.spec.ts
```diff
@@ -2037,6 +2037,89 @@ Object.entries(entities).forEach(([key, EntityClass]) => {
...
+        const customPolicy = new PolicyClass();
+        await customPolicy.create(apiContext, [
+          ...DATA_CONSUMER_RULES,
+          {
+            name: 'DenyEditTagsAndGlossary-Rule',
+            resources: ['All'],
+            operations: ['EditTags', 'EditGlossaryTerms', 'EditDescription'],
+            effect: 'deny',
+          },
+        ]);
```
aniketkatkar97: Can we move tests setups out of test block in beforeAll?

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Entity.spec.ts
```diff
@@ -2037,6 +2037,89 @@ Object.entries(entities).forEach(([key, EntityClass]) => {
...
+          panelContainer.getByTestId('edit-description')
+        ).not.toBeVisible();
+
+        // Close panel
+        await panelContainer.getByTestId('close-button').click();
+
+        const { apiContext: cleanupContext, afterAction: cleanupAfterAction } =
+          await getApiContext(page);
+        await customRole.delete(cleanupContext);
+        await customPolicy.delete(cleanupContext);
```
aniketkatkar97: same here
1. We decided to not have cleanups in the tests
2. Should not be part of test logic

openmetadata-ui/src/main/resources/ui/src/components/common/DescriptionSection/DescriptionSection.less
```diff
@@ -49,7 +49,7 @@
               -webkit-box-orient: vertical;
               overflow: hidden;
               text-overflow: ellipsis;
-              max-height: 2.8em; /* 2 lines * 1.4 line-height */
+              max-height: 3.8em;
```
aniketkatkar97: Can we not change these to a random numbers? We need to have some logic behind such values. What's the reason for having max-height here?

openmetadata-ui/src/main/resources/ui/src/components/common/DescriptionSection/DescriptionSection.tsx
```diff
@@ -72,7 +73,9 @@ const DescriptionSection: React.FC<DescriptionSectionProps> = ({
   const handleDescriptionChange = useCallback(
     async (updatedDescription: string) => {
       if (onDescriptionUpdate) {
-        await onDescriptionUpdate(updatedDescription);
+        await onDescriptionUpdate(
+          removeAttachmentsWithoutUrl(updatedDescription)
```
aniketkatkar97: This does not seem a cleaner way. What's the necessity of doing this?

openmetadata-ui/src/main/resources/ui/src/components/common/DisplayName/DisplayName.test.tsx
```diff
@@ -108,10 +108,10 @@ describe('Test DisplayName Component', () => {
       </MemoryRouter>
     );
 
-    const linkElement = screen.getByTestId('Sample Entity');
+    const linkElements = screen.getAllByTestId('Sample Entity');
 
-    expect(linkElement.tagName).toBe('SPAN');
-    expect(linkElement).toHaveTextContent('Sample Display Name');
+    expect(linkElements.length).toBeGreaterThanOrEqual(1);
+    expect(linkElements[0].tagName).toBe('SPAN');
```
aniketkatkar97: Why are we adding tagName assertions here in a test where we need to check whether a link is rendered?

openmetadata-ui/src/main/resources/ui/src/components/common/DisplayName/DisplayName.test.tsx
```diff
@@ -123,10 +123,10 @@ describe('Test DisplayName Component', () => {
       </MemoryRouter>
     );
 
-    const nameElement = screen.getByTestId('Sample Entity');
+    const nameElements = screen.getAllByTestId('Sample Entity');
 
-    expect(nameElement.tagName).toBe('SPAN');
-    expect(nameElement).toHaveTextContent('Sample Display Name');
+    expect(nameElements.length).toBeGreaterThanOrEqual(1);
+    expect(nameElements[0].tagName).toBe('SPAN');
```
aniketkatkar97, line 131: same here, both these seem similar, how is it testing 2 different things? If both tests have different purpose but same logic we need to rethink how we are writing tests here.

openmetadata-ui/src/main/resources/ui/src/components/Customization/GenericProvider/GenericProvider.tsx
```diff
@@ -273,6 +281,10 @@ export const GenericProvider = <T extends Omit<EntityReference, 'type'>>({
         try {
           apiResponseColumn = await updateTableColumn(fqn, update);
         } catch (error) {
+          // If called from ColumnDetailPanel, re-throw error so it can show local toast
+          if (source) {
+            throw error;
+          }
           showErrorToast(error as AxiosError);
```
aniketkatkar97, line 288: Please check if we can improve this. We need to catch/handle error in one place. This kind of handling may become harder to follow in the future.

aniketkatkar97: Also we are having source as `string` here but not using the value of the string for any assertions. This could also kept as a boolean flag.

openmetadata-ui/src/main/resources/ui/src/components/Database/ColumnDetailPanel/ColumnDetailPanel.less
```diff
@@ -124,4 +124,38 @@
...
+    .ant-alert-icon {
+      display: flex;
+      align-items: center;
+      margin-right: 8px;
+    }
+
+    #alert-icon {
+      padding: 4px;
+      border-width: 3px;
+      font-size: 16px;
```
aniketkatkar97: Can we please use theme/variables for values? If we are using MUI these should be avoided altogether.

openmetadata-ui/src/main/resources/ui/src/components/Database/ColumnDetailPanel/ColumnDetailPanel.test.tsx
```diff
@@ -786,11 +823,187 @@ describe('ColumnDetailPanel', () => {
       await waitFor(
         () => {
           expect(onColumnFieldUpdate).toHaveBeenCalled();
-          expect(showErrorToast).toHaveBeenCalled();
+          expect(getByTestId('alert-bar')).toBeInTheDocument();
+          expect(getByTestId('alert-bar')).toHaveAttribute(
+            'data-type',
+            'error'
+          );
```
aniketkatkar97: just checking if attributes is not a solid validation, we should be asserting values.

aniketkatkar97: Same goes everywhere

openmetadata-ui/src/main/resources/ui/src/utils/StringsUtils.ts
```diff
@@ -327,3 +327,30 @@ export const jsonToCSV = <T extends JSONRecord>(
...
+ *
+ * @param htmlString - HTML content as a string
+ * @returns A cleaned HTML string with invalid file-attachment divs removed
+ */
+export function removeAttachmentsWithoutUrl(htmlString: string): string {
+  const parser = new DOMParser();
+  const doc: Document = parser.parseFromString(htmlString, 'text/html');
+
+  const attachments: NodeListOf<HTMLDivElement> =
+    doc.querySelectorAll<HTMLDivElement>('div[data-type="file-attachment"]');
```
aniketkatkar97, line 350: as commented where it is used, need to understand the need for this.

openmetadata-ui/src/main/resources/ui/src/components/common/DescriptionSection/DescriptionSection.tsx
harsh-vador: Agreed, it wasn't the cleanest place for this logic. The function is necessary to remove broken file placeholders left by cancelled uploads. To address your concern, I've refactored this into `ModalWithMarkdownEditor` so the editor itself ensures clean output, keeping `DescriptionSection` purely presentational.

openmetadata-ui/src/main/resources/ui/src/utils/StringsUtils.ts
harsh-vador: This utility is necessary to clean up any temporary or failed file upload artifacts from the rich text editor state before saving. When a file upload is initiated, a file-attachment node is created immediately with an empty URL. If the upload is cancelled or fails, or if the user saves before it completes, this node remains in the document.
Handling this cleanup on the frontend ensures that we only send valid, finalized content to the backend. It also keeps the backend decoupled from the specific implementation details of the frontend editor's temporary states.

PR #25652, opened 2026-01-31
Feature Better configuration management

mohityadav766:
Adding a configSource property to OpenMetadata settings that lets operators control where  
  configuration is read from:                                                                      
                                                                                                   
  - ENV - Always read from YAML/environment variables (UI shows read-only indicator, API rejects   
  updates)                                                                                         
  - DB - Always read from database (current default behavior)                                      
  - AUTO - Last-write-wins based on timestamps                                                     
                                                                                                   
  Problem solved: Previously, settings stored in the database on first startup would ignore        
  subsequent ENV/YAML changes, confusing operators who expected config file changes to take effect.
                                                                                                   
  Scope:                                                                                           
  - Dual-source settings (auth, email, SCIM, etc.) default to AUTO mode                             
  - DB-only settings (login config, search settings, etc.) default to DB mode                      
  - UI shows "Managed by Environment" indicator and hides edit buttons for ENV-managed settings    
  - Backend validates and rejects API updates when configSource=ENV

----
## Summary by Gitar

- **Migration updates:**
  - Updated `backfillConfigSourceEnvHash` to sync `env_sync_timestamp` alongside `env_hash` in `openmetadata_settings`.
  - Replaced `systemDAO.updateEnvHash` with `systemDAO.updateConfigMetadata` to support batch metadata updates.


bootstrap/sql/migrations/native/1.12.7/mysql/schemaChanges.sql
```diff
@@ -0,0 +1,10 @@
+-- Add columns for configuration source tracking
+ALTER TABLE openmetadata_settings
```
harshach, line 2: @mohityadav766 The migrations are in flyway, this should be in native

PR #25706, opened 2026-02-04
Increase timeout for QR

Khairajani:
---

## Summary by Gitar

- Adds `QUERY_RUNNER_FETCHING_EXPIRY_TIME` constant set to 5 minutes (300,000ms) in `Services.constant.ts`
- Increases timeout from the existing 3-minute `FETCHING_EXPIRY_TIME` to accommodate long-running Query Runner queries
- Specifically addresses scenarios with cold start delays (e.g., Trino cluster initialization)
- Non-breaking change that introduces a dedicated timeout for Query Runner operations without affecting existing test connection timeouts


Khairajani:
Added in `collate-ui/src/main/resources/ui/src/plugins/QueryRunner/utils/constants.ts`
Not needed here

PR #25708, opened 2026-02-05
Fix: source hash instability during ingestion runs

### LINKED ISSUE #25678: Investigate sourceHash instability in ingestion pipeline
## Context

While fixing the **sourceHash-only notifications issue (https://github.com/open-metadata/OpenMetadata/issues/25674)**, several observations were made around potential causes of **hash instability**.

The `sourceHash` field is computed from `CreateTableRequest.model_dump_json()` in
`ingestion/src/metadata/utils/source_hash.py`.

Currently, even when **no user-visible metadata changes** occur (schemas, descriptions, tags, ownership), the hash can differ between ingestion runs, leading to unnecessary updates.

## Observations

The following factors may be introducing non-determinism in the hash calculation:

### 1. List ordering

* Columns are processed in the order returned by source APIs (e.g., `Databricks table.columns`)
* This order is not guaranteed to be consistent across API calls
* The same applies to constraints and tags lists

**Relevant code:**

* `ingestion/src/metadata/ingestion/source/database/unitycatalog/metadata.py:393`
  Columns processed in API return order
* `ingestion/src/metadata/ingestion/source/database/sql_column_handler.py:378`
  Columns appended without sorting

### 2. Volatile `EntityReference` fields

* The `owners` field contains `EntityReference` objects
* Fields such as `href`, `deleted`, and `inherited` may change between runs even when the logical owner remains the same

### 3. DDL formatting

* `schemaDefinition` fetches DDL from the source
* Whitespace or formatting may vary between calls

## Potential areas to explore

* Sort columns by `ordinalPosition` before serialization
* Sort constraints and tags by deterministic keys
* Exclude volatile fields from hash calculation
* Normalize DDL whitespace

## Impact

This issue focuses on addressing the **underlying hash instability**, if desired, to achieve cleaner and more predictable update behavior.

The notification spam issue has been addressed separately (changes to `sourceHash` no longer trigger notifications).
For reference, it was causing the following notifications to trigger:

Image

harshsoni2024:
Fix #25678

#### Problems Addressed
- List ordering non-determinism - Columns, tags, constraints, and owners could appear in different orders across API calls, causing hash changes even when data was identical

- Volatile EntityReference fields - Fields like href, deleted, and inherited could change between runs without any actual metadata change

- DDL whitespace variations - schemaDefinition could have different whitespace/formatting between calls

#### Fixes Implemented
- Sort columns by ordinalPosition (if present), then by name - including nested children columns and their tags

- Sort tags by tagFQN for consistent ordering

- Sort table constraints by constraintType and column names

- Sort owners by fullyQualifiedName, name, or id

- Remove volatile fields (href, deleted, inherited) from entity references recursively

- Normalize whitespace in schemaDefinition - collapses multiple whitespace into single space

- Use json.dumps(sort_keys=True) for deterministic JSON serialization

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


harshach:
@harshsoni2024 is the gitar comments not relevant https://github.com/open-metadata/OpenMetadata/pull/25708#issuecomment-3853602747

harshsoni2024:
> @harshsoni2024 is the gitar comments not relevant [#25708 (comment)](https://github.com/open-metadata/OpenMetadata/pull/25708#issuecomment-3853602747)
IMO, it's not concerning. the current approach is safe to go ahead.

PR #25709, opened 2026-02-05
BulkAPIs should use bulkWrite/bulkUpdate methods to reduce the no.of queries and db connections

harshach:
This pull request introduces significant improvements to how bulk operations are managed, focusing on reducing database queries and connections through batch processing. The implementation uses batch database operations with proper authorization checks and graceful fallback mechanisms to handle bulk imports/exports efficiently.

**Bulk Operation Improvements:**

* Implemented batch database operations for bulk create and update operations, reducing the number of database queries from N individual queries to optimized batch operations. This significantly reduces database connection usage and improves performance.
* Added pre-fetching of existing entities in a single query, enabling efficient conflict detection and update preparation before database writes.
* Implemented batch authorization checks that validate permissions for all entities before processing, preventing wasted database work on unauthorized operations.
* Added deferred store pattern for updates (`updateWithDeferredStore`) that enables batch database writes while maintaining proper lifecycle event handling.
* Graceful fallback to per-entity processing when batch operations fail, ensuring reliability.

**Bulk Write/Update Methods:**

* Added `bulkWrite()` and `bulkUpdate()` methods to EntityDAO for efficient batch database operations.
* Refactored EntityRepository bulk operations to use new batch methods, with proper error handling and authorization.
* Updated EntityResource to handle bulk operations with async 202 responses, returning both immediate auth failures and processing results.

**Configuration and Setup:**

* Simplified BulkOperationConfiguration by removing complex semaphore-based throttling in favor of simpler batch processing.
* Updated configuration for bulk operations with queue size and thread pool settings.
* Tuned HikariCP connection pool settings for faster failure detection and more aggressive resource reclamation.

**Testing and Code Quality:**

* Added comprehensive integration tests in BulkOperationPermissionsIT covering create, update, permission checks, and concurrency scenarios.
* Updated all entity resource integration tests with bulk operation test coverage.
* Tests validate batch operations work correctly across many entity types (API endpoints, collections, databases, schemas, tables, teams, test cases, etc.).

These changes together provide more robust, predictable, and scalable handling of bulk operations by reducing database query overhead and connection usage through efficient batch processing.


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/BulkExecutor.java
```diff
@@ -130,92 +131,114 @@ public static void reset() {
...
   }
 
+  /** Acquire a connection permit. Times out if permit is not available within configured timeout. */
+  public void acquireConnection() throws InterruptedException {
+    if (!connectionSemaphore.tryAcquire(timeoutSeconds, TimeUnit.SECONDS)) {
+      throw new RejectedExecutionException(
+          "Could not acquire bulk connection permit within " + timeoutSeconds + "s");
+    }
+  }
+
```
TeddyCr: Would that create issues if no connections is available and I am trying to acquire one? Would this cause an ingestion to fail for example?

harshach: Yes, and on ingestion we should retry. If there are no connections to acquire it should fail no?

TeddyCr: That makes sense, we'd need to handle that on the ingestion side.

PR #25720, opened 2026-02-05
Remove supported entity field for context, exact terms, custom and pattern recognizers

edg956:
### Describe your changes:

This removes fields we agreed we wouldn't need in recognizers anymore

#
### Type of change:

- [X] Bug fix

#
### Checklist:

---

## Summary by Gitar

- **Removes redundant `supportedEntity` field** from four PII recognizer schemas (PatternRecognizer, ContextRecognizer, ExactTermsRecognizer, CustomRecognizer), simplifying the configuration model
- **Replaces enum-based entity identification with tag FQN** - recognizers now use the tag's fully qualified name (e.g., "PII.EmailAddress") instead of PIIEntity enum values
- **Adds v1.1.20 database migrations** for MySQL and PostgreSQL to update existing tag recognizers in the database with the new structure
- **Updates Python recognizer factory** to accept `tag_fqn` parameter and pass it to Presidio recognizers as `supported_entity`
- **Regenerates TypeScript models** to reflect schema changes across the frontend codebase

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/generated/api/classification/createTag.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/generated/api/classification/createTag.ts
@@ -249,10 +249,6 @@ export interface RecognizerConfig {
     patterns?:   Pattern[];
     regexFlags?: RegexFlags;
     /**
-     * The entity type this recognizer detects
-     */
-    supportedEntity?: PIIEntity;
-    /**
      * Language supported by this recognizer
      */
     supportedLanguage?: ClassificationLanguage;
@@ -381,8 +377,6 @@ export interface RegexFlags {
 }
 
 /**
- * The entity type this recognizer detects
- *
  * Enum of PII (Personally Identifiable Information) tags for classification and detection
  * of sensitive data. Based on Presidio supported entities
  * (https://microsoft.github.io/presidio/supported_entities/).
--- a/openmetadata-ui/src/main/resources/ui/src/generated/api/classification/createTagWithRecognizers.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/generated/api/classification/createTagWithRecognizers.ts
@@ -221,10 +221,6 @@ export interface RecognizerConfig {
     patterns?:   Pattern[];
     regexFlags?: RegexFlags;
     /**
-     * The entity type this recognizer detects
-     */
-    supportedEntity?: PIIEntity;
-    /**
      * Language supported by this recognizer
      */
     supportedLanguage?: ClassificationLanguage;
@@ -353,8 +349,6 @@ export interface RegexFlags {
 }
 
 /**
- * The entity type this recognizer detects
- *
  * Enum of PII (Personally Identifiable Information) tags for classification and detection
  * of sensitive data. Based on Presidio supported entities
  * (https://microsoft.github.io/presidio/supported_entities/).
--- a/openmetadata-ui/src/main/resources/ui/src/generated/api/classification/loadTags.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/generated/api/classification/loadTags.ts
@@ -329,10 +329,6 @@ export interface RecognizerConfig {
     patterns?:   Pattern[];
     regexFlags?: RegexFlags;
     /**
-     * The entity type this recognizer detects
-     */
-    supportedEntity?: PIIEntity;
-    /**
      * Language supported by this recognizer
      */
     supportedLanguage?: ClassificationLanguage;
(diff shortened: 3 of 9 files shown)


edg956:
@gitar-bot review

edg956:
@gitar-bot 

> Code Review 👍 Approved with suggestions 4 resolved / 7 findings
> Clean removal of redundant `supportedEntity` field with proper runtime replacement via `tag_fqn`. Previous parameter naming issue is resolved; other prior findings (test assertion gap, "Unknown" default, cross-version migration reference) remain unaddressed.
> 
> ⚠️ Edge Case: Fallback to "Unknown" entity type creates silent misclassification
> 📄 [ingestion/src/metadata/pii/algorithms/presidio_recognizer_factory.py:249](https://github.com/open-metadata/OpenMetadata/pull/25720/files#diff-01ce83513da852fe35686d03c6496f5dc9855d9bdcd313ccfb45275383c01eebR249) 📄 [ingestion/src/metadata/pii/algorithms/presidio_recognizer_factory.py:282](https://github.com/open-metadata/OpenMetadata/pull/25720/files#diff-01ce83513da852fe35686d03c6496f5dc9855d9bdcd313ccfb45275383c01eebR282) 📄 [ingestion/src/metadata/pii/scanners/custom_ner_scanner.py:219](https://github.com/open-metadata/OpenMetadata/pull/25720/files#diff-12ba72d96143a50bc418163f78c5a7a8869b60913255f4b71e22ce0e3459a6baR219)
> 
> When `tag.fullyQualifiedName` is `None`, the code falls back to `"Unknown"` as the tag FQN (lines 249 and 282). This value is then passed through to Presidio as `supported_entity`, meaning:
> 
> 1. Presidio recognizers will report entity type `"Unknown"` in analysis results
> 2. In `custom_ner_scanner.py` line 219, `tag.fullyQualifiedName == entity_type` will never match `"Unknown"` for any real tag, so detected entities will be silently dropped
> 
> If `fullyQualifiedName` is None, it's likely a data integrity issue that should be logged as a warning and the tag should be skipped rather than creating a recognizer with an unmatchable entity type. Consider:
> 
> ```python
> name = tag.fullyQualifiedName
> if not name:
>     logger.warning("Tag has no fullyQualifiedName, skipping recognizer creation")
>     return recognizers
> ```

This is already applied

> ⚠️ Edge Case: Predefined recognizer test lost supported_entities assertion
> 📄 [ingestion/tests/unit/pii/algorithms/test_presidio_recognizer_factory.py:193-198](https://github.com/open-metadata/OpenMetadata/pull/25720/files#diff-ffcd85602d6ea7602495afd8cc99ca5877a1eb7e5ff3af0c19501f9d5eb902abR193-R198)
> 
> The test `test_create_predefined_recognizer` removed the assertion `assert "EMAIL_ADDRESS" in result.supported_entities` but did not replace it with any equivalent assertion. The test now only checks `isinstance(result, EntityRecognizer)` which is a much weaker assertion.
> 
> Since predefined recognizers (like Presidio's built-in `EmailRecognizer`) have their own `supported_entities` defined internally, the test should still verify that the recognizer was correctly created with the expected entity type. The `tag_fqn` parameter `"Some.Tag"` is passed but it's unclear whether predefined recognizers use it or their own built-in entity type.
> 
> Suggested fix: Add back an assertion verifying `result.supported_entities` contains the expected value (either the built-in entity type or the passed tag FQN, depending on intended behavior).

This is on purpose. We don't care what the supported entities in the recognizer are because our classification mechanism already maps tags to recognizers

> 💡 Quality: v1120 migration references v1110 MigrationUtil directly
> 📄 [openmetadata-service/src/main/java/org/openmetadata/service/migration/mysql/v1120/Migration.java:19](https://github.com/open-metadata/OpenMetadata/pull/25720/files#diff-67b8d7577382c2589877997f23dd8362b26721e486c148479829f2e9296d9addR19) 📄 [openmetadata-service/src/main/java/org/openmetadata/service/migration/postgres/v1120/Migration.java:19](https://github.com/open-metadata/OpenMetadata/pull/25720/files#diff-6e92c93302e5e41fd841010a5414c6f2a43e1b78e5c50ddb4950b969d70f45e7R19)
> 
> Both the MySQL and PostgreSQL v1120 `Migration` classes instantiate `org.openmetadata.service.migration.utils.v1110.MigrationUtil`. While reusing utilities across migration versions can work, referencing a previous version's utility class by its fully qualified name rather than extracting shared logic into a common location creates a fragile cross-version dependency.
> 
> If v1110's `MigrationUtil` is refactored or removed in the future, v1120 migrations will break. Consider either:
> 
> 1. Moving the shared `setRecognizersForSensitiveTags` logic to a version-independent utility, or
> 2. Adding a comment explaining why v1110's utility is deliberately reused here.
> 
> This is a minor concern since migrations are typically run once and frozen.

For this one-off things I'd prefer not to extract for now. If I've got to do this again then I'll do it in the TagDAO

PR #25751, opened 2026-02-09
Fix Metrics collection; reduce no.of metrics; improve slow request lo…

harshach:
# Fix Metrics collection; reduce no.of metrics; improve slow request logging

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

---

----
## Summary by Gitar

- **Metrics collection improvements:**
  - Reduced number of tracked metrics for improved performance
  - Enhanced `RequestLatencyContext` with latency phase tracking and QoS filtering
  - Added `LatencyPhaseFilter` to categorize request processing phases
- **Slow request logging:**
  - Improved request latency tracking in `RequestMetricsFilter` and related monitoring classes
  - Added `LatencyPhase` enum for phase-based request categorization
- **Database optimizations:**
  - Implemented `ReadPlanner` framework with `ReadBundle` and `ReadPlanBuilder` for query optimization
  - Updated repository classes to use new read plan system
- **Testing infrastructure:**
  - Added `@Isolated` annotation to parallel test execution
  - Added new test classes for read planning, latency phases, and request entity caching


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -646,20 +662,320 @@ protected void clearRelationshipsForUpdateMany(List<T> entities) {
...
+    }
+    return entity;
+  }
+
+  private String projectInheritanceFields(Fields fields) {
+    String inheritableFields = getInheritableFields();
+    if (inheritableFields == null || inheritableFields.isBlank()) {
+      return "";
+    }
+    if (fields == null || fields.getFieldList().isEmpty()) {
```
TeddyCr: Should we use our utilities function `nullOrEmpty` for these checks?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -646,20 +662,320 @@ protected void clearRelationshipsForUpdateMany(List<T> entities) {
...
+      return inheritableFields;
+    }
+
+    List<String> projectedFields = new ArrayList<>();
+    for (String inheritableField : inheritableFields.split(",")) {
+      String normalized = inheritableField.trim();
+      if (!normalized.isEmpty() && fields.contains(normalized)) {
+        projectedFields.add(normalized);
+      }
```
TeddyCr, line 828: out of curiosity why do we need this extra normalize call for the fields? Just wondering how they would have blank spaces around them.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
TeddyCr, line 1: @harshach do you think we should split some elements out of this file? I see it went from ~7.2K lines to 8.3K lines to almost 9K+ lines. just wondering if it is going to create any maintenance bottleneck. 🤷

TeddyCr:
## 1. ReadPlan/ReadBundle — Single-Entity Prefetch Cache

  **Problem**: Each `GET /v1/{entity}/{id}` triggered N separate DB queries (one per field: owners, domains, tags, votes, followers, extension, etc.).

  **Solution**: A two-phase prefetch system:
  - **Phase 1 (Plan)**: `ReadPlanner` inspects the requested `Fields` and entity capabilities to build an immutable `ReadPlan` declaring what relationships, tags,
   votes, and extensions are needed.
  - **Phase 2 (Execute)**: `buildReadBundle()` executes the plan in 2-4 bulk queries (one `findToRelationships`, one `findFromRelationships`, plus optional
  tags/votes/extension), storing results in a `ReadBundle`.
  - The bundle is pushed onto a thread-local stack (`ReadBundleContext`). Field getters (`getOwners()`, `getDomains()`, `getTags()`, etc.) check the bundle first,
   falling back to DB only on miss.
  - Stack design handles nested reads (e.g., inheritance resolution triggering sub-entity loads).

  **Subclass extension**: Repositories override `augmentReadPlan()` to register entity-specific prefetch keys, and `prefetchEntitySpecificReadData()` to use them
  (e.g., `TableRepository` prefetches schema/database/service refs).

  ---

  ## 2. Batch Field Loading — List/Bulk Operations

  **Problem**: `listEntities` and bulk import/update called `setFieldsInternal` per entity, each triggering its own set of DB queries.

  **Solution**:
  - New `setFieldsInBulk(Fields, List<T>)` default implementation using registered `fieldFetchers` (owners, domains, tags, followers, votes, experts,
  dataProducts, certification) that operate on entire lists.
  - `serializeJsons()` (used by `listWithOffset`) calls `setFieldsInBulk` first, falling back to sequential `setFieldsInternal` on failure.
  - `listAfter()`/`listBefore()`/`get(List<UUID>)` all use the batch path.
  - New batch DAO methods: `storeRelationshipsInternal(List<T>)`, `clearRelationshipsForUpdateMany(List<T>)`.

  **Inheritance overhaul**: Replaced per-entity `setInheritedFields` with a template-method pattern:
  - Subclasses implement `getParentReference(T)`, `getInheritableFields()`, `requiresParentForInheritance(T, Fields)`, and `applyInheritance(T, Fields,
  EntityInterface)`.
  - Batch `setInheritedFields(List<T>, Fields)` collects unique parent refs, batch-loads them via `Entity.getEntities()`, caches in a `ThreadLocal`
  `inheritanceParentCache`, then applies inheritance.
  - Cache supports superset field matching (a parent loaded with `"owners,domains,retentionPeriod"` serves requests for `"owners,domains"`).

  ---

  ## 3. Request Latency Tracking & Observability

  **Problem**: No visibility into where time is spent within a request.

  **Solution**: Per-request, per-subsystem timing via `RequestLatencyContext`:
  - **Subsystem timers**: DB, search, auth, RDF, and "server" (unaccounted) time tracked via `AtomicLong` accumulators in a `ThreadLocal<RequestContext>`.
  - **Micrometer metrics**: 6 timers per endpoint/method combination (`request.latency.total`, `.database`, `.search`, `.auth`, `.rdf`, `.server`) with SLO
  buckets.
  - **Phase tracking**: Named sub-spans via `try (var ignored = phase("phaseName"))` using a thread-local stack. `@LatencyPhase` JAX-RS annotation auto-wraps
  `EntityResource` methods with HTTP-method-derived phases.
  - **Slow request logging**: Requests exceeding 1s (configurable) emit a `WARN` to `org.openmetadata.slowrequest` with full breakdown.
  - **Cross-thread propagation**: `wrapWithContext()` captures the `RequestContext` and propagates it to async worker threads.

  **Filter pipeline**:
  - `MetricsRequestFilter` (`@Priority(AUTHENTICATION - 1)`) — wraps entire request lifecycle
  - `RequestMetricsFilter` (`@Priority(1)`) — tracks active request count
  - `LatencyPhaseFilter` (`@Priority(USER)`) — opens/closes named phases

  **QoS**: New `QoSConfiguration` + `JettyQoSIntegration` installs a Jetty `QoSHandler` for back-pressure (disabled by default, configurable max concurrent/queued
   requests).

  ---

  ## 4. Entity Cache Optimizations

  **Guava cache**: Changed from `LoadingCache<..., EntityInterface>` to `LoadingCache<..., String>` (JSON). Eliminates aliasing/mutation bugs — each caller
  deserializes a fresh copy.

  **RequestEntityCache**: New request-scoped `ThreadLocal` cache keyed by `(entityType, lookupType, lookupValue, fieldsKey, includesKey, fromCache)`. Defensive
  copy on read via `JsonUtils.deepCopy()`. Eliminates redundant entity loads within a single request.

  ---

  ## 5. Async Search Indexing via Lifecycle Events

  New `EntityLifecycleEventDispatcher` (singleton) with `EntityLifecycleEventHandler` interface. `SearchIndexHandler` (async, priority 10) decouples search index
  updates from entity persistence:
  - Uses a 50-thread virtual thread pool (`Thread.ofVirtual()`)
  - `ElasticSearchEntityManager`/`OpenSearchEntityManager` gained async bulk clients

  ---

  ## 6. Other Changes

  - **JsonUtils**: Added `BlackbirdModule` for Jackson performance, `OBJECT_MAPPER_LENIENT` for tolerant deserialization,
  `readValueLenient`/`convertValueLenient`/`pojoToJsonIgnoreNull`/`getJsonStructure` methods.
  - **TagLabelUtil**: Batch tag resolution (`populateTagLabel(List<TagLabelWithFQNHash>)`), graceful error handling for READ paths (`addDerivedTagsGracefully`,
  `applyTagCommonFieldsGracefully`).
  - **CollectionDAO**: New `TagLabelWithFQNHash` DTO, `getTagsInternalBatch` for bulk tag queries, `getDerivedTagsBatch`.
  - **RdfUpdater**: All operations wrapped with `RequestLatencyContext.startRdfOperation()`/`endRdfOperation()`.
  - **DefaultAuthorizer**: `authorize()` wrapped with auth latency tracking, new impersonation authorization via `checkImpersonationAuthorization()`.
  - **JwtFilter**: Impersonation support via `X-Impersonate-User` header (bot-only), `ImpersonationContext` ThreadLocal.
  - **ResourceContext**: New constructors accepting `Include`, `Fields`, `RelationIncludes` for fine-grained field loading.

  ---

  ## Bugs Found

  | # | Severity | Description |
  |---|---|---|
  | 1 | **High** | **`validatePersonalAccessToken` in `JwtFilter` has inverted ternary**: `claims.get(TOKEN_TYPE) != null ? StringUtils.EMPTY :
  claims.get(TOKEN_TYPE).asString()` — when the claim exists (normal case), it compares against empty string, so PAT validation never matches. The else branch
  would NPE on null claim. |
  | 2 | **High** | **`ReadPlan.isEmpty()` doesn't check `entitySpecificPrefetchKeys`**: A plan with only entity-specific prefetch keys and no
  relations/tags/votes/extension returns `isEmpty() == true`, causing `buildReadBundle()` to skip the entire prefetch. Currently masked but latent. |
  | 3 | **Medium** | **`serializeJsons` silently swallows `setFieldsInBulk` failures**: Catches exception and falls back to sequential processing without logging
  the batch failure. |
  | 4 | **Medium** | **`parentCacheForPrepare` leak risk**: Instance-level `ThreadLocal` cleared only in `EntityResource.bulkCreate`'s `finally`.
  `ImpersonationCleanupFilter` does not call `clearParentCache()`. |
  | 5 | **Medium** | **`RequestEntityCache` ThreadLocal not cleared by `endRequest()`**: Must be cleared separately; if missed, cached data bleeds across requests
   on the same thread. |
  | 6 | **Medium** | **`ImpersonationContext` ThreadLocal leak on exceptions**: `clear()` only called on the happy path in `JwtFilter.filter()`. If exception
  escapes mid-filter, context leaks. |
  | 7 | **Medium** | **`pojoToJsonIgnoreNull` creates a new `ObjectMapper` copy per call**: Should be a pre-built singleton. |

  ---

  ## Design Concerns

  | # | Severity | Description |
  |---|---|---|
  | 8 | **Medium** | **`GlossaryTermRepository` bypasses batch infrastructure**: `setFieldsInBulk` calls `Entity.getEntity()` per entity in a loop, and
  `setInheritedFields(List)` bypasses `inheritanceParentCache`. |
  | 9 | **Medium** | **`TestCaseRepository.linkedTablesCache` implicit coupling**: Cache set before `super.setFieldsInBulk()`, which internally calls the
  overridden `setInheritedFields(List)` that reads from it. Fragile if base class call order changes. |
  | 10 | **Low** | **ReadBundle `Include` mismatch silently falls through to DB**: No logging when a caller requests a relation with a different `Include` than
  what was planned. |
  | 11 | **Low** | **Phase tracking doesn't work on child threads**: `wrapWithContext` propagates `RequestContext` but not `phaseStack`. Documented trade-off. |
  | 12 | **Low** | **`internalTimerStartNanos` race in multi-threaded context**: Two concurrent `getAndSet(0)` calls lose one thread's server time. Documented
  trade-off. |
  | 13 | **Low** | **`entitySpecificPrefetchKeys` is stringly-typed**: No compile-time enforcement between `augmentReadPlan` and `prefetchEntitySpecificReadData`.
   |
  | 14 | **Low** | **`checkImpersonationAuthorization` has N+1 queries**: Walks roles -> policies -> rules via individual `Entity.getEntity()` calls. |
  | 15 | **Low** | **`EntityLifecycleEventDispatcher` shutdown not wired**: The virtual thread pool has no visible lifecycle hook for clean shutdown. |

  ---

  ## Test Coverage

  | Component | Status | Gaps |
  |---|---|---|
  | `ReadPlannerTest` | Good | Missing: `isEmpty()` with only prefetch keys, null `relationIncludes`, blank prefetch key |
  | `ReadBundleTest` | Partial | Missing: tags, null refs, multi-entity, different `Include` values for same field |
  | `RequestLatencyContextTest` | Thorough | Concurrent, cross-thread, stress tests, timing accuracy |
  | `LatencyPhaseFilterTest` | Good | Annotation resolution, HTTP-method defaults |
  | `MetricsErrorHandlingTest` | Good | Error during request, null context, nested DB ops |
  | `RequestEntityCacheTest` | Minimal | Missing: `getByName`, `putByName`, `invalidate`, thread isolation, lifecycle cleanup |
  | `ImpersonationCleanupFilterTest` | Basic | Has test but doesn't cover exception-path leak |
  | `MetricsRequestFilter` | **None** | No dedicated unit tests |
  | `JettyQoSIntegration` | **None** | No tests |
  | `QoSConfiguration` | **None** | No tests |
  | Full ReadPlan→ReadBundle→setFields pipeline | **None** | No integration test |

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
harshach: agree, but don't want to bottleneck this PR from getting merged, that should be refactor of its own

conf/openmetadata.yaml
```diff
@@ -78,7 +78,7 @@ server:
   # Virtual Threads (Project Loom) - Recommended for Java 21+
   # Jetty 12 uses AdaptiveExecutionStrategy to route blocking tasks to virtual threads
   # while keeping non-blocking I/O on platform threads for optimal cache locality
-  enableVirtualThreads: ${SERVER_ENABLE_VIRTUAL_THREAD:-false}
+  enableVirtualThreads: ${SERVER_ENABLE_VIRTUAL_THREAD:-true}
```
mohityadav766, line 81: @harshach we have had issues earlier when switching this to true 
Server is able to take many requests , but database had issues and there were timeout issues 

either we have to scale the database more to have more connections

harshach: @mohityadav766 its gated now with QOS, you'll not run out of db connections

conf/openmetadata.yaml
```diff
@@ -149,11 +149,47 @@ server:
   #    allowRenegotiation: true
   #    endpointIdentificationAlgorithm: (none)
 
+qos:
+  enabled: ${QOS_ENABLED:-true}
```
mohityadav766, line 155: **serious:** QoS is enabled by default (`true`) with `maxRequestCount: 100`. This is a behavioral change for every existing deployment. A cluster running light ingestion + UI + API will hit the 100 concurrent limit easily during bulk operations, returning `503 Service Unavailable` to legitimate traffic.

Consider defaulting to `false` so existing users opt-in, or at minimum set `maxRequestCount` much higher (e.g., 500-1000) as a default. Enabling QoS by default in a config file that ships to all users is risky without migration guidance.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ReadPlanner.java
```diff
@@ -0,0 +1,133 @@
...
+
+import org.openmetadata.schema.EntityInterface;
+import org.openmetadata.schema.type.Relationship;
+import org.openmetadata.service.util.EntityUtil.Fields;
+import org.openmetadata.service.util.EntityUtil.RelationIncludes;
+
+/** Builds a consolidated read plan for a single-entity fetch. */
+final class ReadPlanner {
+
+  ReadPlanBuilder newBuilder(
```
mohityadav766: **nit:** This method has 13 parameters (4 objects + 9 booleans). The boolean parade is hard to read and very easy to get wrong at call sites (e.g., swapping `supportsFollowers` with `supportsReviewers` compiles fine but silently breaks).

Consider a `ReadPlannerConfig` record or a builder pattern:
```java
record ReadPlannerConfig(
    boolean supportsOwners, boolean supportsDomains,
    boolean supportsFollowers, boolean supportsReviewers,
    boolean supportsDataProducts, boolean supportsExperts,
    boolean supportsExtension, boolean supportsTags,
    boolean supportsVotes) {}
```
This would make call sites self-documenting.

mohityadav766 (commented):
## Code Review

Thorough review covering architecture, correctness, and code standards. Comments tagged as **serious:** for issues that could cause bugs/production problems, and **nit:** for style/quality improvements.

Overall this is a significant and ambitious PR with good ideas (ReadPlan/ReadBundle, shouldCompare optimization, JSON-string cache, metrics cardinality reduction). The concerns below are mostly around safety edges and code hygiene.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -2193,50 +2925,61 @@ public final PatchResponse<T> patch(
...
-  private PatchResponse<T> patchCommonWithOptimisticLocking(
-      T original,
-      JsonPatch patch,
-      String user,
-      UriInfo uriInfo,
-      ChangeSource changeSource,
-      boolean useOptimisticLocking) {
-    return patchCommonWithOptimisticLocking(
-        original, patch, user, uriInfo, changeSource, useOptimisticLocking, null);
+  private Fields planFieldsForPatch(Set<String> patchedFieldNames) {
```
mohityadav766: **nit:** `planFieldsForPatch` accepts `patchedFieldNames` but unconditionally returns `patchFields`. The comment explains the rationale, but the method adds indirection for no current benefit. If this is a placeholder for a future optimization, consider adding a `TODO` with a tracking reference, or remove the method and use `patchFields` directly at the call site until the optimization is ready.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -838,21 +1154,69 @@ public final T get(
...
     }
-    T entity = find(id, relationIncludes.getDefaultInclude());
-    setFieldsInternal(entity, fields, relationIncludes);
-    setInheritedFields(entity, fields);
-
-    T entityClone = JsonUtils.deepCopy(entity, entityClass);
-    clearFieldsInternal(entityClone, fields);
-    return withHref(uriInfo, entityClone);
+    T entity;
+    try (var ignored = phase("entityLookup")) {
```
mohityadav766: **nit:** The `get()` method now has 10+ consecutive `try (var ignored = phase(...))` blocks, which significantly reduces readability. The business logic (lookup → plan → bundle → setFields → inherit → clear → href → cache) is obscured by the instrumentation boilerplate.

Consider extracting the instrumented logic into a helper, or using an AOP/decorator approach so the core method stays clean. Alternatively, a single `phase()` wrapping the entire `get()` with sub-timers would reduce the noise while keeping granular timing.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -1899,187 +2620,188 @@ private static boolean isValidEntityForCache(EntityInterface entity) {
...
+      if (seenEntityIds.add(entity.getId())) {
+        uniqueEntities.add(entity);
+      }
+    }
+    if (uniqueEntities.isEmpty()) {
+      return;
+    }
+
+    try (var ignored = phase("lifecycleDispatch")) {
+      EntityLifecycleEventDispatcher.getInstance()
```
mohityadav766, line 2939: **serious:** `new ArrayList<>(uniqueEntities)` is a shallow copy — the `EntityInterface` objects inside are shared references. Since `SearchIndexHandler.isAsync()` returns `true` and dispatches to `CompletableFuture.runAsync()`, the async handler reads the same entity objects that the request thread may continue to mutate (e.g., during `withHref`, `clearFields`, or if the same entity is updated immediately after creation).

This is a race condition. Consider using `JsonUtils.deepCopy()` for each entity before handing them to the async dispatcher, or serialize to JSON strings on the calling thread.

openmetadata-service/src/main/java/org/openmetadata/service/monitoring/RequestLatencyContext.java
```diff
@@ -143,249 +125,386 @@ public static Timer.Sample startSearchOperation() {
...
 
-  /**
-   * Reset all static state. This is primarily for testing to ensure clean state between tests. This
-   * clears all timer maps so that new timers will be created and registered with the current
-   * registry.
-   */
+  private static final class ActivePhase {
+    private final long startNanos;
+    private final long dbTimeAtStartNanos;
+    private long childNanos;
```
mohityadav766, line 428: **nit:** `childNanos` and `childDbNanos` are mutable, non-volatile, non-final fields in `ActivePhase`. While `ActivePhase` is only accessed via a `ThreadLocal` stack (so no cross-thread visibility issue today), if `wrapWithContext()` ever propagates the phase stack across threads, these fields would have visibility problems. Consider making them `volatile` or documenting the single-thread invariant explicitly.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -465,119 +655,154 @@ public Table addJoins(UUID tableId, TableJoins joins) {
...
-        JsonUtils.readValue(
-            daoCollection
-                .entityExtensionDAO()
-                .getExtension(table.getId(), TABLE_SAMPLE_DATA_EXTENSION),
-            TableData.class);
+    Table table;
+    try (var ignored = phase("tableSampleDataFind")) {
+      table = find(tableId, NON_DELETED);
+    }
+    TableData sampleData = JsonUtils.readValue(null, TableData.class);
```
mohityadav766: **nit:** `JsonUtils.readValue(null, TableData.class)` initializes `sampleData` with a value that is immediately overwritten on the very next line. This dead initialization is confusing — just declare `TableData sampleData;` and assign it from the actual DB load.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/APICollectionRepository.java
```diff
@@ -222,13 +269,14 @@ public APICollectionUpdater(
...
     @Override
     public void entitySpecificUpdate(boolean consolidatingChanges) {
-      recordChange(
-          "sourceHash",
-          original.getSourceHash(),
-          updated.getSourceHash(),
-          false,
-          EntityUtil.objectMatch,
-          false);
+      if (shouldCompare("sourceHash"))
```
mohityadav766: **nit:** `if (shouldCompare(...))` without braces wrapping a multi-line `recordChange()` call. This is technically valid Java, but fragile — if someone adds a second statement thinking it's inside the `if`, it will execute unconditionally. The project's checkstyle should catch this, but if not, please add braces:
```java
if (shouldCompare("sourceHash")) {
  recordChange(...);
}
```
This pattern appears in multiple entity updaters — worth a project-wide cleanup.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/APIEndpointRepository.java
```diff
@@ -59,6 +60,7 @@
 import org.openmetadata.service.util.FullyQualifiedName;
 
 public class APIEndpointRepository extends EntityRepository<APIEndpoint> {
+  private static final String PREFETCH_DEFAULT_FIELDS = "apiEndpoint.defaultFields";
```
mohityadav766: **nit:** `PREFETCH_DEFAULT_FIELDS = "apiEndpoint.defaultFields"` is a stringly-typed constant with no compile-time validation. If the config key is renamed or removed, this fails silently at runtime. Consider using an enum or a validated config class for prefetch keys.

conf/openmetadata.yaml
harshach: what is the current reality here? we will bring the server down to an extent if the API inflcux comes in

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
harshach: we are running huge penalities of json ser/de, there is no end to our deepCopy across the code base

mohityadav766:
@harshach please check below

the current pattern has the check duplicated 300+ times across all entity updaters. There are a few progressively better approaches:                                      
                                                                                                                                                                                  
  1. Push the check into recordChange() (simplest, biggest impact)                                                                                                                
                  
  Most entity-specific updaters follow this exact pattern:                                                                                                                        
  if (shouldCompare("sourceHash"))
      recordChange("sourceHash", original.getSourceHash(), updated.getSourceHash(), ...);

  The field name is already passed to recordChange. Just move the guard inside:

  // In EntityUpdater
  protected <T> void recordChange(String field, T orig, T updated, ...) {
      if (!shouldCompare(field)) return;  // single guard, one place
      // ... existing logic
  }

  This eliminates the if (shouldCompare(...)) wrapper at 300+ call sites with zero changes to callers. Every entity-specific updater becomes:
  // Before (repeated everywhere):
  if (shouldCompare("sourceHash"))
      recordChange("sourceHash", orig.getSourceHash(), updated.getSourceHash(), ...);

  // After (just the call):
  recordChange("sourceHash", orig.getSourceHash(), updated.getSourceHash(), ...);

  2. For base updateInternal() methods with side effects

  Methods like updateOwners(), updateTags() do more than recordChange — they modify relationships. For these, a small helper:

  private void compareAndUpdate(String field, Runnable updater) {
      if (shouldCompare(field)) updater.run();
  }

  Then updateInternal() becomes:
  compareAndUpdate("description", this::updateDescription);
  compareAndUpdate("displayName", this::updateDisplayName);
  compareAndUpdate("owners", this::updateOwners);
  compareAndUpdate("tags", () -> updateTags(updated.getFullyQualifiedName(), ...));

  Still stringly-typed, but at least the pattern is centralized.

  3. For compile-time safety (optional, more investment)

  If you want to kill the string-key fragility entirely:

  enum UpdateField {
      DESCRIPTION("description"),
      DISPLAY_NAME("displayName"),
      OWNERS("owners"),
      TAGS("tags"),
      // ...
      ;
      final String jsonName;
      UpdateField(String jsonName) { this.jsonName = jsonName; }
  }

  Then shouldCompare(UpdateField.OWNERS) won't compile if someone typos the field name, and renaming is IDE-refactorable.

  Recommended approach

  Option 1 alone gets you 80% of the win with the least effort — one 2-line change inside recordChange() and then delete the if (shouldCompare(...)) wrappers from every entity
  updater. Option 2 cleans up the remaining base-class methods. Option 3 is nice-to-have if you want long-term safety.

harshach:
Addressed items from the latest review comment:

1) **TestCaseRepository ThreadLocal leak**
- Stopped storing single-entity `linkedTablesCache` state in `prepare()`; it is now only used for bulk field hydration paths.
- Added explicit cleanup in `clearParentCache()` to guarantee request-boundary cleanup via `ImpersonationCleanupFilter -> Entity.clearRepositoryThreadLocals()`.
- Also moved `getOrCreateTestSuite(...)` to run only after all validation passes to avoid creating side entities on early validation failure.

2) **Static inheritance cache cleanup**
- Updated `EntityRepository.clearParentCache()` to clear both `parentCacheForPrepare` and `inheritanceParentCache`.

3) **2.0.0 native migration index alignment**
- **Postgres**: drop old `(usageDate,id)` constraint and recreate unique index on `(id,usageDate)` with idempotent guards.
- **MySQL**: replaced non-portable `DROP INDEX IF EXISTS` with safe conditional dynamic SQL using `information_schema.statistics`, then enforce unique `(id,usageDate)` index ordering.

Validation run after changes:
- `openmetadata-service`: build/install success
- full `openmetadata-integration-tests`: **PASS**
  - Tests run: 10668, Failures: 0, Errors: 0, Skipped: 906
  - Total time: 11:59

harshach:
Addressed the items from [#issuecomment-3987130059](https://github.com/open-metadata/OpenMetadata/pull/25751#issuecomment-3987130059):

1) **Postgres base schema inconsistency fixed**
- Updated `bootstrap/sql/schema/postgres.sql` to align with migration target state:
  - removed old `entity_usage_usagedate_id_key UNIQUE (usagedate, id)` constraint
  - made `entity_usage_id_usage_date_idx` a **UNIQUE** index on `(id, usagedate)`

2) **MySQL migration window removed**
- Updated `bootstrap/sql/migrations/native/2.0.0/mysql/schemaChanges.sql` to avoid separate drop/add steps.
- Migration now:
  - detects whether the desired unique `(id, usageDate)` index already exists
  - if reordering is needed, performs `DROP INDEX + ADD UNIQUE INDEX` in a **single ALTER TABLE statement**
  - if index is already correct, no-op
  - if index is missing, adds it directly
- This keeps idempotency while avoiding a separate-statement window with no unique constraint.

3) **APIEndpoint requestSchema copy-paste bug cleaned up**
- Updated `openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/APIEndpointRepository.java` request/response schema field fallback logic to use the correct schema source and simplified null handling.

Validation:
- `openmetadata-service` install: PASS
- targeted IT (`APIEndpointResourceIT#test_paginationFetchesTagsAtBothEntityAndFieldLevels`): PASS
- full `openmetadata-integration-tests`: PASS
  - `Tests run: 10668, Failures: 0, Errors: 0, Skipped: 906`
  - `Total time: 21:30`

ecerulm:
@harshach,

Is this addressing specifically the high number of series in `/prometheus` for 

* `request_latency_total_seconds_bucket`
* `request_latency_search_seconds_bucket`
* `request_latency_database_seconds_bucket`
* `request_latency_internal_seconds_bucket`
* `http_requests_total`
* `http_request_queue_seconds_*`
* `http_request_processing_seconds_*`
* `http_request_duration_seconds_*` 

In my case (openmetadata 1.11.11) each of those have ~40k series , one per path , including paths like `v1/tables/name/xxxxxx - aaa.bbb.111`,  `v1/tables/name/xxxxxx - aaa.bbb.222`. 

I wonder if this PR is addressing those an generating a single `v1/table/name/{name}` , etc or if I need to file a github issue for that one.

harshach:
@ecerulm Yes thats part of the fix

harshach:
@ecerulm my bad, it wasn’t shipped in 1.12.x . We are going to add a fix to 1.12.x release  for next week as this contains larger changes

ecerulm:
1.12.6 does not contain it either, does this mean I have to wait for 1.13 for this?

PR #25822, opened 2026-02-11
Implement Athena struct profiling following established patterns

IceS2:
# Add struct column profiling support for Athena                                                                                                              
                                                                                                                                                              
Athena tables with STRUCT columns (e.g., address.street, address.geo.lat) were silently skipped during profiling because SQAStruct is in the NOT_COMPUTE set. This PR adds support by flattening STRUCT columns into their leaf fields, following the same pattern used by Databricks, so each leaf gets its real type and is profiled normally.                                                                                                                              
                                                                                                                                                              
 ### Changes:
  - Created AthenaProfilerInterface that recursively flattens STRUCT children into dot-notation leaf columns (e.g., address → address.street, address.city, address.geo.lat)
  - Patches AthenaStatementCompiler.visit_column to quote each dot-separated segment individually at SQL compilation time ("address"."street" instead of "address.street"), handling reserved words safely while keeping clean names for the OM API
  - Registered the new profiler interface in the Athena service spec
  - Added unit tests mirroring the existing Databricks test_visit_column pattern
---
### How it looks like

<img width="2291" height="1143" alt="image" src="https://github.com/user-attachments/assets/0fcbdf98-3aca-47a0-8dcb-853f6a5dfeac" />


ingestion/src/metadata/profiler/interface/sqlalchemy/athena/profiler_interface.py
```diff
@@ -0,0 +1,122 @@
...
+    This handles reserved words in struct field names correctly.
+    """
+    result = SQLCompiler.visit_column(self, column, *args, **kwargs)
+    col_name = str(getattr(column, "name", ""))
+    if "." in col_name:
+        col_idx = result.find(col_name)
+        if col_idx >= 0:
+            prefix = result[:col_idx]
+            segments = col_name.split(".")
+            quoted = ".".join(self.preparer.quote_identifier(seg) for seg in segments)
```
TeddyCr, line 56: if I understand correctly we need to do "foo"."bar"."bazz" to query the struct field?

IceS2: we do not, I added it to be on the safe side if there are any reserved keywords as struct fields.

pmbrull (approved):

PR #25831, opened 2026-02-12
fix(permissions): Gate usageSummary API field on ViewUsage permission across entity pages

### LINKED ISSUE #25821: usageSummary Field Causes Permission Error with ViewBasic on Database/Schema
**Affected module**
Does it impact the UI, backend or Ingestion Framework? - UI

**Describe the bug**
- When a user has only ViewBasic permission on a Database/Schema, the UI fails to load children (schemas) due to an API call requesting the usageSummary field. If usageSummary is included in the API request, the response fails with a permission error. If it is removed, the schemas are returned correctly. This results in users being unable to view Database/Schema children even though they have sufficient ViewBasic permission.

**To Reproduce**

Image

Image

```
"Principal: CatalogPrincipal{name='prajwal123'} operations [ViewUsage] not allowed"
```

**Expected behavior**
- The list of database schemas should still be returned when the user has ViewBasic permission.

**Version:**
 - Python version:

**Additional context**

harsh-vador:
### Describe your changes:

Fixes #25821 

## Problem
Users with only ViewBasic permission (and no ViewUsage) were unable to see child entities in Database and Schema pages. The root cause was that multiple API fetch calls unconditionally included usageSummary in the fields parameter. The backend enforces ViewUsage for this field and returns a 403 "operations [ViewUsage] not allowed" error — causing the entire request to fail, not just the usage data.

### This affected:

- Database schema list on the Database detail page 
- Database schema detail page
- Pipelines list and detail page
- Dashboards list in service pages
- Chart detail page

### Root Cause
Two distinct anti-patterns caused the regression:

1. || operator — `permissions.X.ViewAll || permissions.X.ViewUsage` evaluates to true when ViewAll=true even if ViewUsage is explicitly denied (RBAC deny overrides allow).
2. Unconditional field inclusion — usageSummary baked into a shared defaultFields string with no permission gate. 

The correct utility `getPrioritizedViewPermission(permissions, Operation.ViewUsage)` checks if ViewUsage is explicitly set first; only falls back to ViewAll when ViewUsage is unset.

### Source fixes:
- DatabaseSchemaTable.tsx — Replace || with getPrioritizedViewPermission; conditionally include USAGE_SUMMARY in fields and Usage column
- [DatabaseSchemaPage.component.tsx] — Gate USAGE_SUMMARY on viewUsagePermission in both getDatabaseSchemaDetailsByFQN calls
- ServiceDetailsPage.tsx — Replace inline || with getPrioritizedViewPermission in fetchDatabases, fetchDashboards, fetchPipeLines
- ServiceVersionPage.tsx — Same fix for fetchDatabases, fetchDashboards
- ChartDetailsPage.component.tsx — Replace || with getPrioritizedViewPermission
- PipelineDetailsPage.component.tsx — Add viewUsagePermission check; conditionally append USAGE_SUMMARY in both fetch calls
- PipelineDetailsUtils.tsx — Remove USAGE_SUMMARY from defaultFields constant

https://github.com/user-attachments/assets/757620e3-c4d1-4c86-bb16-e58d124bd1a1

### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/utils/entityPermissionUtils.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/utils/entityPermissionUtils.ts
@@ -624,6 +624,10 @@ export const testDatabaseServiceSpecificOperations = async (
   const databasesResponse = await databasesResponsePromise;
 
   expect(databasesResponse.status()).toBe(200);
+
+  await expect(
+    testUserPage.locator('[data-testid="service-children-table"]')
+  ).toBeVisible();
 };
 
 export const testDashboardServiceSpecificOperations = async (
@@ -643,6 +647,10 @@ export const testDashboardServiceSpecificOperations = async (
   const dashboardsResponse = await dashboardsResponsePromise;
 
   expect(dashboardsResponse.status()).toBe(200);
+
+  await expect(
+    testUserPage.locator('[data-testid="service-children-table"]')
+  ).toBeVisible();
 };
 
 export const serviceEntityConfig = {
--- a/openmetadata-ui/src/main/resources/ui/src/pages/ServiceDetailsPage/ServiceDetailsPage.test.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/pages/ServiceDetailsPage/ServiceDetailsPage.test.tsx
@@ -55,6 +55,7 @@ import {
 } from '../../rest/workflowAPI';
 import serviceUtilClassBase from '../../utils/ServiceUtilClassBase';
 import { getCountLabel, shouldTestConnection } from '../../utils/ServiceUtils';
+import { getPrioritizedViewPermission } from '../../utils/PermissionsUtils';
 import { showErrorToast } from '../../utils/ToastUtils';
 import { useRequiredParams } from '../../utils/useRequiredParams';
 import ServiceDetailsPage from './ServiceDetailsPage';
@@ -860,6 +861,42 @@ describe('ServiceDetailsPage', () => {
       await waitFor(() => {
         expect(getDatabases).toHaveBeenCalled();
       });
+    });
+
+    it('should include usageSummary in database fields when ViewUsage is allowed', async () => {
+      (getPrioritizedViewPermission as jest.Mock).mockReturnValue(true);
+      (getDatabases as jest.Mock).mockResolvedValue({
+        data: [{ id: 'db1', name: 'Database 1' }],
+        paging: { total: 1 },
+      });
+
+      await renderComponent();
+
+      await waitFor(() => {
+        expect(getDatabases).toHaveBeenCalled();
+      });
+
+      const fields = (getDatabases as jest.Mock).mock.calls[0][1];
+
+      expect(fields).toContain('usageSummary');
+    });
+
+    it('should exclude usageSummary from database fields when ViewUsage is denied', async () => {
+      (getPrioritizedViewPermission as jest.Mock).mockReturnValue(false);
+      (getDatabases as jest.Mock).mockResolvedValue({
+        data: [{ id: 'db1', name: 'Database 1' }],
+        paging: { total: 1 },
+      });
+
+      await renderComponent();
+
+      await waitFor(() => {
+        expect(getDatabases).toHaveBeenCalled();
+      });
+
+      const fields = (getDatabases as jest.Mock).mock.calls[0][1];
+
+      expect(fields).not.toContain('usageSummary');
     });
 
     it('should fetch topics for messaging service', async () => {


openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/Permissions/EntityPermissions.spec.ts
```diff
@@ -76,7 +76,7 @@ Object.entries(entityConfig).forEach(([, config]) => {
   const entityType = entity.getType();
 
   test.describe(`${entityType} Permissions`, () => {
-    const customPropertyName = `pw${entityType}CustomProperty${uuid()}`;
+    const customPropertyName = `pw${entityType.replace(/\s+/g, '')}CustomProperty${uuid()}`;
```
ShaileshParmar11, line 79: why this change is needed?

openmetadata-ui/src/main/resources/ui/playwright/utils/entityPermissionUtils.ts
```diff
@@ -558,6 +601,59 @@ export const entityConfig = {
...
+  Database: {
+    class: DatabaseClass,
+    specificTest: testDatabaseSpecificOperations,
+  },
+} as const;
+
+export const testDatabaseServiceSpecificOperations = async (
+  testUserPage: Page,
+  entity: DatabaseServiceClass,
+  _effect: 'allow' | 'deny'
```
ShaileshParmar11, line 613: if its not needed, lets remove

openmetadata-ui/src/main/resources/ui/playwright/utils/entityPermissionUtils.ts
```diff
@@ -558,6 +601,59 @@ export const entityConfig = {
...
+
+  const databasesResponse = await databasesResponsePromise;
+
+  expect(databasesResponse.status()).toBe(200);
+};
+
+export const testDashboardServiceSpecificOperations = async (
+  testUserPage: Page,
+  entity: DashboardServiceClass,
+  _effect: 'allow' | 'deny'
```
ShaileshParmar11, line 636: same here

openmetadata-ui/src/main/resources/ui/playwright/utils/entityPermissionUtils.ts
```diff
@@ -558,6 +601,59 @@ export const entityConfig = {
...
+  const dashboardsResponsePromise = testUserPage.waitForResponse(
+    (response) =>
+      response.url().includes('/api/v1/dashboards') &&
+      response.request().method() === 'GET'
+  );
+
+  await redirectToHomePage(testUserPage);
+  await entity.visitEntityPage(testUserPage);
+
+  const dashboardsResponse = await dashboardsResponsePromise;
```
ShaileshParmar11, line 647: we should also do some UI validation to make sure page is not broken

openmetadata-ui/src/main/resources/ui/playwright/utils/entityPermissionUtils.ts
```diff
@@ -558,6 +601,59 @@ export const entityConfig = {
...
+  const databasesResponsePromise = testUserPage.waitForResponse(
+    (response) =>
+      response.url().includes('/api/v1/databases') &&
+      response.request().method() === 'GET'
+  );
+
+  await redirectToHomePage(testUserPage);
+  await entity.visitEntityPage(testUserPage);
+
+  const databasesResponse = await databasesResponsePromise;
```
ShaileshParmar11, line 624: we should also do some UI validation to make sure page is not broken

openmetadata-ui/src/main/resources/ui/playwright/utils/entityPermissionUtils.ts
harsh-vador: Updated to add UI validation (service-children-table visible) alongside the network check. The network assertion directly catches the regression — if usageSummary is unconditionally included, backend returns 403 and the test fails. Kept both layers since unit tests alone only mock the API.

openmetadata-ui/src/main/resources/ui/playwright/utils/entityPermissionUtils.ts
harsh-vador: Updated to add UI validation (service-children-table visible) alongside the network check. The network assertion directly catches the regression — if usageSummary is unconditionally included, backend returns 403 and the test fails. Kept both layers since unit tests alone only mock the API.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/Permissions/EntityPermissions.spec.ts
harsh-vador: Custom property names can't contain spaces. Without this, entityType = 'Database Schema' produces pwDatabase SchemaCustomProperty${uuid()} which fails the name validation when clicking the create button. The replace strips the space so it becomes pwDatabaseSchemaCustomProperty${uuid()}, matching the pattern every other entity uses.

ShaileshParmar11 (approved):

PR #25836, opened 2026-02-12
Chore(UI): Refactor the ingestion runner selection logic to avoid flakiness

aniketkatkar97:
This pull request improves the flexibility and reliability of the service runner selection process in the service creation flow, particularly for Playwright-based UI tests. The main changes introduce parameterization for runner details, update selectors to use more stable attributes, and enhance test robustness.

**Enhancements to service runner selection:**

- The `createService` method in `ServiceBaseClass` now accepts a `runnerDetails` parameter, allowing dynamic specification of runner `name` and `displayName` for more flexible test scenarios.
- The search and selection logic for runners has been updated to use `runnerDetails.name` instead of hardcoded values, making the code more reusable and less brittle.
- Selectors now rely on the `data-key` attribute (which uses the runner's `name`) rather than `data-testid` (which uses `displayName`), improving selector reliability since `name` is less likely to change in the application under test. [[1]](diffhunk://#diff-27b613009179a959c1e5cad0edabbe85bc092d0edbbd460b939fac8b3eb2e4d7L116-R138) [[2]](diffhunk://#diff-083bea8642e36df3eb2bf7f41d89b8b539f9857193343e3340b35e04494eccf6R65)

**Improvements to UI component attributes:**

- The `SelectWidget` component now includes a `data-key` attribute on each option, supporting the updated selector strategy and increasing test stability.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/support/entity/ingestion/ServiceBaseClass.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/support/entity/ingestion/ServiceBaseClass.ts
@@ -38,6 +38,10 @@ import {
 } from '../../../utils/serviceIngestion';
 import { ResponseDataType } from '../Entity.interface';
 
+interface RunnerDetails {
+  name: string;
+  displayName?: string;
+}
 class ServiceBaseClass {
   public category: Services;
   protected serviceName: string;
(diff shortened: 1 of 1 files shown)


ShaileshParmar11 (dismissed):

openmetadata-ui/src/main/resources/ui/playwright/support/entity/ingestion/ServiceBaseClass.ts
```diff
@@ -76,7 +76,13 @@ class ServiceBaseClass {
     // Handle visit service here
   }
 
-  async createService(page: Page) {
+  async createService(
+    page: Page,
+    runnerDetails = {
+      name: 'CollateSaaS',
+      displayName: 'Collate SaaS',
+    }
```
chirag-madlani: We can have it as class property to read and write?
That way we can manage it via service object

PR #25846, opened 2026-02-12
Flaky(UI): Roles playwright test

dhruvjsx:
### Describe your changes:
Screenshot 2026-02-12 at 8 13 52 PM

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

---

## Summary by Gitar

- Improved test isolation by moving role test data variables (`roleName`, `description`, `updatedRoleName`) from module scope into test function scope to prevent interference between parallel test executions
- Replaced fragile `getByRole` selectors with more robust `data-testid` attribute-based locators combined with href matching for better reliability
- Enhanced synchronization mechanisms by adding `waitForLoadState('networkidle')` and `waitForSelector` with `state: 'detached'` to handle timing issues more effectively
- Added null coalescing operator (`??`) for `fullyQualifiedName` property to handle cases where it might be undefined


openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Roles.spec.ts
```diff
@@ -525,7 +532,10 @@ test('Delete role action from manage button options', async ({ page }) => {
   await page.reload();
 
   // Wait for page to be ready after reload
-  await expect(page.locator('[data-testid="loader"]')).not.toBeVisible();
+  await page.waitForLoadState('networkidle');
```
harsh-vador: we dont need this, it is auto-handled

dhruvjsx: We are reloading entire page I don't think network-idle is handled on this method 🤔

harsh-vador: You're right that we're reloading the page, but Playwright's navigation methods (page.reload(), page.goto(), etc.) automatically wait for the load event by default.

From [Playwright's documentation](https://playwright.dev/docs/api/class-page#page-reload)

harsh-vador (approved):

PR #25793, opened 2026-02-10
Improve pytests by parallelizing and optimizing the workflow

IceS2:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

---

## Summary by Gitar

- **Test parallelization:**
  - Added `pytest-xdist` for parallel execution with 2 workers using `--dist=loadgroup` strategy
  - Created `nox` session in `noxfile.py` for flexible integration test runs with configurable workers
- **CI optimization:**
  - Integrated `uv` package installer in `.github/actions/setup-openmetadata-test-environment/action.yml` for faster installation
  - Added separate `py-static-checks` job in `.github/workflows/py-tests.yml` with 10-minute timeout
  - Implemented pip caching with `cache: 'pip'` and fixed installation path syntax in follow-up commits
- **Test framework modernization:**
  - Converted 50+ integration test files from `unittest.TestCase` to pytest-style with fixture-based setup
  - Refactored `ingestion/tests/integration/ometa/conftest.py` (+431 lines) with module-scoped fixtures
- **Coverage configuration:**
  - Migrated from `coverage run` to `pytest --cov=metadata --cov-branch` in `ingestion/Makefile`
  - Removed `dynamic_context = "test_function"` from `pyproject.toml` for improved performance
- **Bug fixes:**
  - Corrected typo in `columnValuesToBeBetween.py:155` where `min_bound` was checked instead of `max_bound`
  - Added lazy imports in `datalake/connection.py` using `TYPE_CHECKING`

<sub>This will update automatically on new commits.</sub>

---

PR #25851, opened 2026-02-12
MINOR: Avoid unit tests to depend on the server

IceS2:
## Summary                                                                                                                                                     
                                                                                                                                                              
  - Make unit tests safe for parallel execution with pytest-xdist (-n auto)                                                                                   
  - Move tests that require a running OpenMetadata server or testcontainers from tests/unit/ to tests/integration/                                            
                                                                                                                                                              
## Unit test xdist fixes                                                                                                                                       

  When running unit tests in parallel with xdist, several issues surface: workers may not trigger metadata/__init__.py (missing singleton registrations), global state leaks between workers (DependencyContainer, profiler metrics), and TestCase-based tests with class-level state don't isolate properly. This PR:
                                                                                                                                                              
  - Adds import metadata in conftest.py to ensure xdist workers initialize singletons                                                                         
  - Patches OpenMetadata.validate_versions and health_check to prevent unit tests from reaching out to localhost:8585                                                                                                                    
  - Adds _isolate_container fixture for DependencyContainer tests                                                                                             
  - Migrates TestCase classes to plain pytest style where class-level state caused xdist failures (qlikcloud, qliksense, deltalake)                           
  - Fixes assertions that assumed a single-process test run (profiler partitions, workflow)                                                                   
                                                                                                                                                              
 ## Test reclassification                                                                                                                                       
                                                                                                                                                              
  Several tests were misclassified as unit tests but actually depend on external services:
                                                                                                                                                              
  - Lineage e2e tests → tests/integration/lineage/e2e/ (require testcontainers for Kafka + PostgreSQL)                                                        
  - Amundsen, Atlas, SAS metadata tests → tests/integration/ (use unittest.mock extensively but test full topology wiring against real schemas)               
  - Sample usage test → tests/integration/usage/ (depends on OpenMetadata server API)


edg956:
@IceS2 would it make sense to create a separate github action to run unit tests on their own? that'd speed up this pipeline as it wouldn't need the OM server running (and we'd catch integration tests disguised as unit tests bc they'd outright fail)

edg956:
Also integration tests could benefit from using a matrix and something like `pytest-split` to divide tests in matrix partitions. The only issue would be building all OM dependencies first and then reuse them

IceS2:
> @IceS2 would it make sense to create a separate github action to run unit tests on their own? that'd speed up this pipeline as it wouldn't need the OM server running (and we'd catch integration tests disguised as unit tests bc they'd outright fail)

 I actually have at least another 2/3 prs all from [this branch](https://github.com/open-metadata/OpenMetadata/pull/25793), but i don't want to push so many changes at once.

Then it should look like this: https://github.com/open-metadata/OpenMetadata/actions/runs/21950386297

There are still some further improvements to it though

PR #25861, opened 2026-02-13
Flaky(UI): Setting Navigation Playwright

dhruvjsx:
## Describe your changes:
Screenshot 2026-02-13 at 2 41 41 PM

Fixes flaky Playwright test in Settings Navigation for drag-and-drop functionality.

I worked on stabilizing the "should allow user to drag-drop menus and save persona navigation" test because it was failing intermittently in CI due to timing and synchronization issues.

### Changes made:
- Added explicit wait for tree items to be visible before interaction
- Removed `force: true` flag that was bypassing Playwright's actionability checks
- Replaced non-retrying assertion with auto-retrying `expect().not.toHaveText()` for DOM state verification
- Improved save button assertion using combined locators with deterministic wait
- Added null check for type safety
- Reordered assertions to follow logical UI state transitions

These changes follow Playwright best practices for stable, non-flaky E2E tests using deterministic, auto-retrying assertions.

---

## Summary by Gitar

- **Eliminated race condition** by adding explicit visibility wait for tree items before drag-and-drop interaction
- **Removed anti-pattern `force: true`** from drag operation to ensure proper actionability checks
- **Implemented retrying assertion** using `expect().not.toHaveText()` to auto-wait for DOM updates after drag-and-drop
- **Improved assertion reliability** using combined locators to check both visibility and enabled state simultaneously


karanh37 (dismissed):

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/SettingsNavigationPage.spec.ts
```diff
@@ -282,15 +287,13 @@ test.describe.serial('Settings Navigation Page Tests', () => {
...
-      await expect(page.getByTestId('save-button')).toBeEnabled();
-
-      const newFirstItemText = await treeItems.first().textContent();
-
-      expect(newFirstItemText).not.toBe(firstItemText);
+      // Now check if save button is enabled
+      await page
+        .getByTestId('save-button')
+        .and(page.locator(':not([disabled])'))
+        .waitFor({ state: 'visible' });
```
ShaileshParmar11: this will just wait for visibility
We should also check for assertion

PR #25870, opened 2026-02-13
MCP - Semantic Search & Tests

pmbrull:
### Describe your changes:

image

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature
- [ ] Documentation

#
### Checklist:

---

## Summary by Gitar

- **Adds semantic search capability** to the MCP tools module using vector embeddings for natural language queries
- **Implements `SemanticSearchTool`** with support for filters (entity_type, service, tags), configurable result size, KNN parameter, and similarity threshold
- **Creates dedicated CI/CD pipeline** (`.github/workflows/mcp-tests.yml`) with path-based triggering for MCP-related changes
- **Configures test infrastructure** in `pom.xml` to exclude load tests (`@Tag("load")`) from regular CI runs while keeping them available for manual execution
- **Tags 3 existing performance tests** in `McpIntegrationTest` to separate them from unit tests

### DIFF
--- a/openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SemanticSearchTool.java
+++ b/openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SemanticSearchTool.java
@@ -54,6 +54,7 @@ public class SemanticSearchTool implements McpTool {
     k = Math.min(Math.max(k, 1), MAX_K);
 
     double threshold = parseDoubleParam(params, "threshold", DEFAULT_THRESHOLD);
+    threshold = Math.min(Math.max(threshold, 0.0), 1.0);
 
     Map<String, List<String>> filters = parseFilters(params);
 


.github/workflows/mcp-tests.yml
```diff
@@ -0,0 +1,123 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: mvn -DskipTests clean install -pl :openmetadata-mcp -am
+
+      - name: Run MCP Tests
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: mvn test -pl :openmetadata-mcp
+
+      - name: Publish Test Report
```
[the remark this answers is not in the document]
pmbrull: thats ok. the labeler takes this into account and we wont be executing anything we dont trust

openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SemanticSearchTool.java
```diff
@@ -0,0 +1,232 @@
...
+
+  @Override
+  public Map<String, Object> execute(
+      Authorizer authorizer, CatalogSecurityContext securityContext, Map<String, Object> params)
+      throws IOException {
+    LOG.info("Executing semanticSearch with params: {}", params);
+
+    String query = (String) params.get("query");
+    if (query == null || query.isBlank()) {
+      return errorResponse("'query' parameter is required");
```
Vishnuujain, line 37: should throw exceptions instead here? and similarly in below lines?

openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SemanticSearchTool.java
```diff
@@ -0,0 +1,232 @@
...
+  private static final int MAX_K = 10_000;
+  private static final double DEFAULT_THRESHOLD = 0.0;
+  private static final int DESCRIPTION_MAX_LENGTH = 500;
+  private static final int DESCRIPTION_TRUNCATE_LENGTH = 450;
+
+  @Override
+  public Map<String, Object> execute(
+      Authorizer authorizer, CatalogSecurityContext securityContext, Map<String, Object> params)
+      throws IOException {
+    LOG.info("Executing semanticSearch with params: {}", params);
```
Vishnuujain, line 33: No authorization check? other tools have auth check, do we want to have that here too?

openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SemanticSearchTool.java
```diff
@@ -0,0 +1,232 @@
...
+    copyIfPresent(hit, cleaned, "fullyQualifiedName");
+    copyIfPresent(hit, cleaned, "name");
+    copyIfPresent(hit, cleaned, "displayName");
+    copyIfPresent(hit, cleaned, "serviceType");
+    copyIfPresent(hit, cleaned, "owners");
+    copyIfPresent(hit, cleaned, "tier");
+    copyIfPresent(hit, cleaned, "tags");
+    copyIfPresent(hit, cleaned, "domains");
+    copyIfPresent(hit, cleaned, "columns");
+    copyIfPresent(hit, cleaned, "certification");
```
Vishnuujain, line 129: The threshold parameter lets callers filter by similarity score, but the cleaned results strip _score so the LLM never sees actual scores and can't calibrate what threshold to use. Should we expose the score (e.g., as similarityScore) so the LLM can make informed threshold decisions? or is it a design decision?

Vishnuujain (dismissed):

openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SemanticSearchTool.java
pmbrull: The current pattern of returning structured error responses is the right
  approach for MCP tools. An LLM client can parse {"error": "...", "totalFound": 0} and reason
  about it. Throwing exceptions would result in unstructured error messages that are harder for
  the LLM to handle. The other tools letting exceptions propagate is arguably the weaker pattern
  here.

openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SemanticSearchTool.java
pmbrull: search auth is handled by search service directly vs. auth check. we still have JWT layer in the filter

openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SemanticSearchTool.java
pmbrull: added

openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SemanticSearchTool.java
Vishnuujain: makes sense

PR #25965, opened 2026-02-18
Fix flaky test for tour specs

shrabantipaul-collate:
### Describe your changes:
Replaced a timing-based approach with a proper API wait pattern in the Tour flow test.

- Removes the arbitrary 500ms delay which could cause flaky tests
- Waits for the actual /search/query API response before proceeding
- Ensures the search results are loaded before validating the tour step badge

Fixes <issue-number>
Screenshot 2026-02-18 at 6 42 59 PM

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

---

## Summary by Gitar

- **Fixed flaky test:**
  - Replaced arbitrary 500ms delay with API response wait in `Tour.spec.ts`
- **API synchronization:**
  - Added `waitForResponse` for `/search/query` endpoint to ensure search completes before badge validation

---

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Flow/Tour.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Flow/Tour.spec.ts
@@ -64,12 +64,12 @@ const validateTourSteps = async (page: Page) => {
 
   await page.getByTestId('searchBox').fill('dim_a');
 
-  await Promise.all([
-    page.waitForResponse(
-      (res) => res.url().includes('/search/query') && res.status() === 200
-    ),
+  const [searchResponse] = await Promise.all([
+    page.waitForResponse((res) => res.url().includes('/search/query')),
     page.getByTestId('searchBox').press('Enter'),
   ]);
+
+  expect(searchResponse.status()).toBe(200);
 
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Flow/Tour.spec.ts
```diff
@@ -63,7 +63,13 @@ const validateTourSteps = async (page: Page) => {
   await expect(page.locator(`[data-tour-elem="badge"]`)).toHaveText('3');
 
   await page.getByTestId('searchBox').fill('dim_a');
-  await page.getByTestId('searchBox').press('Enter', { delay: 500 });
+
+  await Promise.all([
+    page.waitForResponse(
+      (res) => res.url().includes('/search/query') && res.status() === 200
+    ),
```
anuj-kumary: Let's remove the status check from waitForResponse.
Instead, we should wait for the /search/query response and then add a separate assertion to validate the status code.
This will make debugging easier if the API fails, since the test won’t just timeout silently.

shrabantipaul-collate: Fixed.

PR #25969, opened 2026-02-18
chore(ui): fix package vulnerabilities issues

anuj-kumary:
### Describe your changes:

Fix dependency diff ^5.0.0 → ^5.2.2 — resolves DoS vulnerability in parsePatch/applyPatch 
Fix dependency qs 6.14.1 → 6.14.2 — resolves arrayLimit bypass DoS 
Add resolution ajv >= 8.18.0 — resolves ReDoS vulnerability via $data option in transitive dependency pulled in by @rjsf/validator-ajv8
Add resolution markdown-it >= 14.1.1 — resolves ReDoS vulnerability in linkify regex /*+$/ for long sequences of * characters

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

---

----
## Summary by Gitar

- **Security: Fixed DoS vulnerability**\n  - Updated `diff` (^5.0.0 → ^5.2.2) resolving parsePatch/applyPatch DoS; `qs` (6.14.1 → 6.14.2) resolving arrayLimit bypass\n- **Security: Fixed ReDoS vulnerabilities**\n  - Upgraded `ajv` to 8.18.0 via resolution to fix $data ReDoS; added `markdown-it` >=14.1.1 resolution to fix linkify regex ReDoS

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/package.json
+++ b/openmetadata-ui/src/main/resources/ui/package.json
@@ -265,7 +265,6 @@
     "js-yaml": "4.1.1",
     "lodash": ">=4.17.23",
     "lodash-es": ">=4.17.23",
-    "@rjsf/validator-ajv8/**/ajv": ">=8.18.0",
     "markdown-it": ">=14.1.1"
   }
 }
--- a/openmetadata-ui/src/main/resources/ui/yarn.lock
+++ b/openmetadata-ui/src/main/resources/ui/yarn.lock
@@ -4500,16 +4500,6 @@ ajv-formats@^2.1.1:
   dependencies:
     ajv "^8.0.0"
 
-ajv@>=8.18.0, ajv@^8.0.0, ajv@^8.12.0:
-  version "8.18.0"
-  resolved "https://registry.yarnpkg.com/ajv/-/ajv-8.18.0.tgz#8864186b6738d003eb3a933172bb3833e10cefbc"
-  integrity sha512-PlXPeEWMXMZ7sPYOHqmDyCJzcfNrUr3fGNKtezX14ykXOEIvyK81d+qydx89KY5O71FKMPaQ2vBfBFI5NHR63A==
-  dependencies:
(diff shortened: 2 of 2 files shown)


openmetadata-ui/src/main/resources/ui/package.json
```diff
@@ -264,6 +264,8 @@
     "tar-fs": "2.1.4",
     "js-yaml": "4.1.1",
     "lodash": ">=4.17.23",
-    "lodash-es": ">=4.17.23"
+    "lodash-es": ">=4.17.23",
+    "@rjsf/validator-ajv8/**/ajv": ">=8.18.0",
```
aniketkatkar97: if the affected package is `ajv` shouldn't we directly add resolution for that instead of what rjsf depends on?

anuj-kumary: Good point, A global resolution like:` :ajv": >=8.18.0` would override all consumers of ajv in Yarn v1, including ESLint, which depends on ajv@^6.12.4. Since ajv v6 and v8 have incompatible APIs, forcing v8 globally breaks ESLint (particularly config validation in CI). We tried the global resolution first it merged all ajv entries into 8.18.0, which would have broken linting.

The scoped resolution:

`@rjsf/validator-ajv8/**/ajv": ">=8.18.0`
is intentional. It patches only the vulnerable v8 consumer introduced via @rjsf/validator-ajv8, while keeping:

```
ajv@^6.12.4  → 6.12.6   (ESLint — unchanged)
ajv@^8.12.0  → 8.18.0   (@rjsf/validator-ajv8 — patched)

```
Let me know if you’d still prefer a different approach.

aniketkatkar97: But, I don't think the vulnerabilities will be fixed with this resolution. What if other libraries still have dependency on ajv version >= 8.0?

PR #26013, opened 2026-02-20
Fix Virtual Threads unbounded

mohityadav766:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Virtual thread concurrency limiting:**
  - Added `BoundedExecutorService` wrapper enforcing semaphore-based concurrency limits on all async tasks
  - Automatically resolves safe limits from environment variable, database pool size (1/3), or CPU cores (min 4)
  - Prevents unbounded resource consumption during sustained background task execution
- **Graceful thread lifecycle management:**
  - `DistributedJobParticipant` and `DistributedSearchIndexExecutor` now properly track and interrupt long-running virtual threads with configurable timeouts
  - Added `interruptAndJoin()` helper method centralizing thread shutdown logic
  - Ensures clean termination on service stop without resource leaks
- **AsyncService shutdown integration:**
  - New `execute()` and `submit()` convenience methods for bounded async operations
  - Added 30-second graceful shutdown timeout with forced termination fallback
  - Called during application shutdown to coordinate executor service termination
- **Audit event migration:**
  - Migrated `AuditLogRepository.writeAuthEvent()` from unbounded `Thread.startVirtualThread()` to bounded `AsyncService.execute()`
  - Preserves async audit trail writing while respecting concurrency limits

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestSuiteRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestSuiteRepository.java
@@ -253,6 +253,8 @@ public class TestSuiteRepository extends EntityRepository<TestSuite> {
   public void clearFields(TestSuite entity, EntityUtil.Fields fields) {
     entity.setPipelines(fields.contains("pipelines") ? entity.getPipelines() : null);
     entity.setSummary(fields.contains("summary") ? entity.getSummary() : null);
+    entity.setTestCaseResultSummary(
+        fields.contains("summary") ? entity.getTestCaseResultSummary() : null);
     entity.withTests(fields.contains(UPDATE_FIELDS) ? entity.getTests() : null);
   }
 
@@ -519,23 +521,21 @@ public class TestSuiteRepository extends EntityRepository<TestSuite> {
     List<UUID> suiteIds = testSuites.stream().map(TestSuite::getId).toList();
     Map<UUID, List<ResultSummary>> testCaseResultSummaryMap = batchGetResultSummary(suiteIds);
 
-    Set<String> allTestCaseFQNs =
-        testCaseResultSummaryMap.values().stream()
-            .flatMap(List::stream)
-            .map(ResultSummary::getTestCaseName)
-            .collect(Collectors.toSet());
-    Map<String, String> entityLinkMap = batchResolveEntityLinks(allTestCaseFQNs);
-
-    Map<UUID, TestSummary> testSummaryMap =
-        testCaseResultSummaryMap.entrySet().stream()
-            .collect(
-                Collectors.toMap(
-                    Map.Entry::getKey, entry -> getTestSummary(entry.getValue(), entityLinkMap)));
-
-    setFieldFromMap(
-        true, testSuites, testCaseResultSummaryMap, TestSuite::setTestCaseResultSummary);
+    Map<UUID, TestSummary> testSummaryMap = new HashMap<>();
+    testCaseResultSummaryMap.forEach(
+        (id, results) -> testSummaryMap.put(id, computeSimpleSummary(results)));
 
     setFieldFromMap(true, testSuites, testSummaryMap, TestSuite::setSummary);
+  }
+
+  private TestSummary computeSimpleSummary(List<ResultSummary> results) {
+    Map<String, Integer> statusCounts = new HashMap<>();
+    for (ResultSummary r : results) {
+      statusCounts.merge(r.getStatus().toString(), 1, Integer::sum);
+    }
+    TestSummary summary = createTestSummary(statusCounts);
+    summary.setTotal(results.size());
+    return summary;
   }
 
   protected void fetchAndSetIngestionPipelines(List<TestSuite> entities, EntityUtil.Fields fields) {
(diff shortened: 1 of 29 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/util/AsyncService.java
```diff
@@ -18,8 +18,11 @@ public class AsyncService {
   private static final long DEFAULT_INITIAL_RETRY_DELAY_MS = 1000; // 1 second
   private static final long DEFAULT_OPERATION_TIMEOUT_SECONDS = 60; // 60 seconds
 
+  private static final int POOL_SIZE = 20;
+  private static final long SHUTDOWN_TIMEOUT_SECONDS = 30;
+
   private AsyncService() {
-    executorService = Executors.newVirtualThreadPerTaskExecutor();
+    executorService = Executors.newFixedThreadPool(POOL_SIZE, Thread.ofVirtual().factory());
```
manerow: Hey Mohit, nice analysis on the unbounded thread issue!

I have a concern with this specific fix though. `Executors.newFixedThreadPool(20, Thread.ofVirtual().factory())` essentially treats virtual threads like platform threads by reusing them from a pool. Virtual threads are designed to be created and destroyed cheaply (they're very lightweight), so pooling them doesn't really give us any benefit. The actual problem we want to solve is limiting how many tasks run *at the same time*, not how many threads exist.

The recommended Java 21 pattern for this is to keep `newVirtualThreadPerTaskExecutor()` but add a `Semaphore` to control concurrency:

```java
private final Semaphore concurrencyLimit = new Semaphore(MAX_CONCURRENCY);
private final ExecutorService executorService = Executors.newVirtualThreadPerTaskExecutor();
```

A semaphore is basically a counter that only lets N tasks through at a time. The rest wait their turn before starting. This way we still get the benefits of virtual threads (lightweight, no pooling overhead) while preventing resource exhaustion.

We already use this pattern in `UserActivityTracker` (semaphore with 10 permits), so it would keep things consistent across the codebase.

Also worth noting: `newFixedThreadPool` uses an unbounded internal queue (`LinkedBlockingQueue`). So while the number of threads is capped at 20, the number of *waiting tasks* can grow without limit. If tasks are submitted faster than they're processed (say, during a burst of bulk operations), the queue itself can eat up memory.

With the semaphore approach, this is handled more naturally: tasks wait at the semaphore *before* being submitted, so you get back-pressure for free without an ever-growing queue.

openmetadata-service/src/main/java/org/openmetadata/service/util/AsyncService.java
```diff
@@ -18,8 +18,11 @@ public class AsyncService {
   private static final long DEFAULT_INITIAL_RETRY_DELAY_MS = 1000; // 1 second
   private static final long DEFAULT_OPERATION_TIMEOUT_SECONDS = 60; // 60 seconds
 
+  private static final int POOL_SIZE = 20;
```
manerow: The value 20 feels a bit arbitrary here. This executor is shared by 13+ call sites, including heavy operations like CSV exports and lighter ones like audit log writes. If a few CSV exports are running, they could block all 20 slots and prevent quick tasks from executing.

A couple of ideas:
1. Make the concurrency limit configurable via `openmetadata.yaml` so deployments can tune it based on their workload
2. Consider deriving a sensible default from the environment, for example something like `Math.max(minFloor, Math.min(Runtime.getRuntime().availableProcessors() * 2, poolSize / 2))`, where `poolSize` is the DB connection pool size. That way the limit adapts to the machine running the application and also respects the actual downstream bottleneck (DB connections).

openmetadata-service/src/main/java/org/openmetadata/service/audit/AuditLogRepository.java
```diff
@@ -122,18 +123,20 @@ public void write(ChangeEvent changeEvent, boolean isBot) {
...
-              new ChangeEvent()
-                  .withId(UUID.randomUUID())
-                  .withEventType(eventType)
-                  .withEntityType(Entity.USER)
-                  .withEntityId(userId)
-                  .withUserName(userName)
-                  .withTimestamp(System.currentTimeMillis());
-          write(changeEvent);
-        });
+    AsyncService.getInstance()
```
manerow, line 126: Small but important detail: switching from `Thread.startVirtualThread()` to `.submit()` changes how exceptions are handled. With `.submit()`, any exception thrown inside the task gets captured in the returned `Future` object. Since we never read that `Future`, the exception just disappears silently. With the previous code, exceptions would at least show up in the logs through the default uncaught exception handler.

Using `.execute()` instead of `.submit()` would fix this, since `.execute()` doesn't wrap the result in a `Future` and lets exceptions propagate normally:

```java
AsyncService.getInstance()
    .getExecutorService()
    .execute(() -> { ... });
```

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/distributed/DistributedJobParticipant.java
```diff
@@ -190,22 +204,23 @@ private void joinAndProcessJob(SearchIndexJob job) {
...
-                }
-              }
-            });
+    participantThread =
+        Thread.ofVirtual()
+            .name("job-participant-" + job.getId().toString().substring(0, 8))
+            .start(
+                () -> {
+                  try {
+                    processJobPartitions(job);
```
manerow, line 213: Minor nit: in the `finally` block, `participantThread` is set to `null` before `participating.set(false)`. There's a tiny window where a new `onJobDiscovered` could fire and see `participating == true`, which would make it skip the job. It's not really a bug (it just misses one poll cycle), but swapping the order would be slightly cleaner: do all the state cleanup first, then set `participantThread = null` last.

PR #26047, opened 2026-02-23
fix: replace SQL string interpolation with parameterized queries in ListFilter and related filters

manerow:
### Describe your changes:

Fixes authenticated SQL injection vulnerability in multiple query parameters across `ListFilter`, `FeedFilter`, `SuggestionFilter`, and `LearningResourceRepository`.

**Root cause:** Several methods in `ListFilter.java` construct SQL WHERE clauses by directly embedding user-supplied query parameters via `String.format("column = '%s'", userInput)`. These fragments are passed to JDBI DAO queries via `@Define("cond")` (text substitution), allowing any authenticated user to inject arbitrary SQL through parameters like `assignee`, `testCaseResolutionStatusType`, `directory`, `fileType`, and others.

**Fix:** Replaced all unsafe `String.format` interpolation patterns with JDBI `@BindMap` parameterized queries (`:placeholder` syntax), consistent with safe methods already present in the same codebase. Added `@BindMap` parameters to DAO methods in `CollectionDAO` (`SuggestionDAO`, `ThreadDAO`) that were missing them. Also fixed a standalone SQL injection in `listThreadsByTaskAssignee` which built a `LIKE` clause via `String.format`.

**Files changed (7):**
- `ListFilter.java` — 14 methods converted from `String.format` to `:bind` placeholders
- `FeedFilter.java` — Enum values and domain UUIDs now use bind params
- `FeedRepository.java` — All DAO callers updated to pass `filter.getQueryParams()`
- `SuggestionFilter.java` — Added `queryParams` map; all values use bind params
- `SuggestionRepository.java` — DAO callers updated to pass `filter.getQueryParams()`
- `CollectionDAO.java` — Added `@BindMap` to `SuggestionDAO` and `ThreadDAO` methods; rewrote `listThreadsByTaskAssigneesId` to use `@Bind`
- `LearningResourceRepository.java` — Replaced `getInConditionFromString` calls with `buildIndexedBindParams`

### Type of change:

- [x] Bug fix

### Checklist:

- [ ] My PR title is Fixes <issue-number>: <short explanation>

----
## Summary by Gitar

- **Security: SQL injection fix:**
  - Replaced 14 unsafe `String.format` interpolations with JDBI `@BindMap` parameterized queries across filter classes and 23+ DAO methods
  - Added `buildIndexedBindParams()` utility to generate indexed bind parameters for multi-value filters (e.g., `:agentType_0,:agentType_1`)
  - Fixed standalone SQL injection in `listThreadsByTaskAssignee` LIKE clause
- **Infrastructure:**
  - Added `queryParams` map to `FeedFilter` and `SuggestionFilter` (was missing)
  - Threaded `@BindMap` parameters through DAO method signatures to enable binding
  - Maintained MySQL and PostgreSQL code paths; UUIDs in JSON/ARRAY literals safely remain inline (type-guaranteed)

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java
@@ -479,7 +479,7 @@ public class ListFilter extends Filter<ListFilter> {
 
     if (entityFQN != null) {
       if (includeAllTests) {
-        queryParams.put("entityFQNLikePrefix", escape(entityFQN) + Entity.SEPARATOR + "%");
+        queryParams.put("entityFQNLikePrefix", entityFQN + Entity.SEPARATOR + "%");
         queryParams.put("entityFQNExact", entityFQN);
         conditions.add("(entityFQN LIKE :entityFQNLikePrefix OR entityFQN = :entityFQNExact)");
       } else {
--- a/openmetadata-service/src/test/java/org/openmetadata/service/jdbi3/ListFilterTest.java
+++ b/openmetadata-service/src/test/java/org/openmetadata/service/jdbi3/ListFilterTest.java
@@ -52,44 +52,40 @@ class ListFilterTest {
   void test_getAgentTypeCondition_singleAgentType() {
     ListFilter filter = new ListFilter();
 
-    // Test single agent type for MySQL
     filter.addQueryParam("agentType", "CollateAI");
     String condition = filter.getCondition("app_entity");
     assertTrue(
-        condition.contains("JSON_EXTRACT(json, '$.agentType') IN ('CollateAI')")
-            || condition.contains("json->>'agentType' IN ('CollateAI')"));
+        condition.contains("JSON_EXTRACT(json, '$.agentType') IN (:agentType_0)")
+            || condition.contains("json->>'agentType' IN (:agentType_0)"));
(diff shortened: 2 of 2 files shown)


manerow:
The PR already handles the empty-values edge case (e.g. input `","` or `" , , "` yielding zero values after split/trim/filter). `buildIndexedBindParams` returns `"''"` when the filtered list is empty (line 593-594), which produces` IN ('')`, valid SQL that matches nothing. The suggested fix from @gitar-bot is identical to the existing code.

manerow:
@gitar-bot **The `escape()` removal is intentional, not a bug.** This exact issue was caught during testing when `put_testCase_list_200` failed (expected 12 results, got 2).

The `escape()` function replaces `_` with `\_`. With **string interpolation**, MySQL's string literal parser consumes the backslash (since `\_` is not a recognized escape sequence), so `_` still reaches the LIKE engine as a wildcard. The `escape()` was effectively a no-op.

With **bind parameters**, the value bypasses string literal parsing and goes directly to the LIKE engine, where `\_` means literal underscore — changing semantics. This caused `put_testCase_list_200` to return 2 results instead of 12.

Removing `escape()` preserves the existing behavior where `_` acts as a LIKE wildcard, matching what the codebase has always done. If stricter literal-underscore matching is desired, that should be a separate change with updated tests.

PR #26072, opened 2026-02-24
FEATURE: Informix Database Connection Config

akashverma0786:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Informix database support:**
  - Added `InformixConnection` schema with SSL configuration and server name settings
  - Registered `InformixTableMetricComputer` to read metrics from `systables` catalog
  - Added `Dialects.Informix` enum value for database dialect identification
- **Profiler metric implementations:**
  - Implemented `MedianFn` compiler for Informix using ROW_NUMBER window functions
  - Added `SumFn` compiler handling DECIMAL precision and DATETIME type edge cases
  - Configured `LenFn`, `ModuloFn`, and column count functions with literal binds to avoid JDBC bind parameters
- **Configuration and UI:**
  - Created Informix test connection JSON with validation steps
  - Added `Informix` to beta services list and generated TypeScript connection types
  - Added `is_blob()` helper for detecting binary/text columns unsuitable for profiling

### DIFF
--- a/ingestion/src/metadata/profiler/orm/functions/median.py
+++ b/ingestion/src/metadata/profiler/orm/functions/median.py
@@ -341,15 +341,7 @@ def _(elements, compiler, **kwargs):  # pylint: disable=unused-argument
     _compute_sqa_fn, so compiler.process() already produces LENGTH("col")
     here — no special-casing needed.
     """
-    # pylint: disable=import-outside-toplevel
-    from metadata.profiler.orm.registry import is_date_time
-
     col_clause = elements.clauses.clauses[0]
-
-    col_type = getattr(col_clause, "type", None)
-    if col_type is not None and is_date_time(col_type):
-        return "CAST(NULL AS DECIMAL(32,4))"
-
     col = compiler.process(col_clause)
     table = elements.clauses.clauses[1].value
     percentile = elements.clauses.clauses[2].value
--- a/ingestion/src/metadata/profiler/orm/functions/sum.py
+++ b/ingestion/src/metadata/profiler/orm/functions/sum.py
@@ -105,24 +105,7 @@ def _(element, compiler, **kw):
 
 @compiles(SumFn, Dialects.Informix)
 def _(element, compiler, **kw):
-    """Informix: BIGINT unsupported, max DECIMAL precision is 32 (not 38).
-    Unlike other databases that infer CASE result type from THEN/ELSE branches (1/0 → integer),
-    Informix infers it from the WHEN condition's column type. For DATETIME/INTERVAL columns,
-    this makes CAST(CASE ... AS DECIMAL) fail. Rewrite to COUNT(*) - COUNT(col) instead."""
-    # pylint: disable=import-outside-toplevel
-    from sqlalchemy.sql.elements import Case as _Case
-
-    from metadata.profiler.orm.registry import is_date_time
-
-    clause = element.clauses.clauses[0] if element.clauses.clauses else None
-    if isinstance(clause, _Case) and clause.whens:
-        cond = clause.whens[0][0]
-        col_clause = getattr(cond, "left", None)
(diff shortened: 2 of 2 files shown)


ingestion/src/metadata/profiler/orm/functions/table_metric_computer.py
```diff
@@ -636,6 +637,68 @@ def compute(self):
         return res
 
 
+class InformixTableMetricComputer(BaseTableMetricComputer):
```
TeddyCr:
```
   count_query = select(func.count().label(ROW_COUNT)).select_from(
                self.runner.raw_dataset.__table__
            )
```

This will be running a count against the raw table. These class are a way to compute metrics using systems tables (i.e. system tables hav the metrics so we just fetch these).

If informix does not have these tables, then we don't need anything here and it will use the base logic

akashverma0786: Understood, 
Thank you for the feedback.

ingestion/src/metadata/profiler/orm/functions/median.py
```diff
@@ -325,3 +325,54 @@ def _(elements, compiler, **kw):  # pylint: disable=unused-argument
...
+    _compute_sqa_fn, so compiler.process() already produces LENGTH("col")
+    here — no special-casing needed.
+    """
+    # pylint: disable=import-outside-toplevel
+    from metadata.profiler.orm.registry import is_date_time
+
+    col_clause = elements.clauses.clauses[0]
+
+    col_type = getattr(col_clause, "type", None)
+    if col_type is not None and is_date_time(col_type):
```
TeddyCr: can you share the use case this covers? If you check the implementation in `median.py` you'll see

```
            # col fullname is only needed for MySQL and SQLite
            return self._compute_sqa_fn(
                column(self.col.name, self.col.type),
                self.col.table.name if self.col.table is not None else None,
                0.5,
                dimension_col,
            )

            return self._compute_sqa_fn(
                LenFn(column(self.col.name, self.col.type)),
                self.col.table.name if self.col.table is not None else None,
                0.5,
                dimension_col,
            )

        logger.debug(
            f"Don't know how to process type {self.col.type} when computing Median"
        )
        return None
```

datetime can't be concatenable or quantifiable

ingestion/src/metadata/profiler/orm/functions/median.py
TeddyCr, line 1: can we use `PERCENTILE_CONT` instead of this custom sql?

ingestion/src/metadata/profiler/orm/functions/sum.py
```diff
@@ -103,6 +103,30 @@ def _(element, compiler, **kw):
     return f"SUM(CAST({proc} AS BIGINT))"
 
 
+@compiles(SumFn, Dialects.Informix)
```
TeddyCr, line 106: I don't understand why we need that complexity here for the sum function. What are we trying to solve?

ingestion/src/metadata/profiler/orm/functions/median.py
akashverma0786: Removing this.

ingestion/src/metadata/profiler/orm/functions/median.py
akashverma0786: PERCENTILE_CONT is not supported in informix and custom sql is the best workaround for this

ingestion/src/metadata/profiler/orm/functions/sum.py
akashverma0786: Addressed and changed

Issue #26119, opened 2026-02-26
Contract action menu ignores RBAC — import/edit/delete shown to ViewOnly users


shrabantipaul-collate:
**Affected module**
UI

**Describe the bug**
The Contract tab action menu (gear button on an existing data contract) displays all options — Import, Import ODCS, Edit, Run now, and Delete — to every authenticated user regardless of their RBAC permissions. A user with only ViewAll/ViewBasic on the dataContract resource can see import and edit actions they are not authorized to perform.

**To Reproduce**

Create a role with a policy that grants only ViewAll and ViewBasic on the dataContract resource (no EditAll).
Assign that role to a user.
Log in as that user and navigate to any table that has a data contract.
Click the Contract tab.
Click the gear button next to the contract name.

Expected: Only Export and Export as ODCS are visible.
Actual: All options are visible — Edit, Run now, Import, Import ODCS, Export, Export as ODCS, Delete.

**Expected behavior**
The action menu should only show Import, Import ODCS, Edit, Run now, and Delete to users who have EditAll permission on the dataContract resource. Export and Export as ODCS should be visible to all users with view access.

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8]
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

PR #26122, opened 2026-02-26
Fixes 26119 : import/edit/delete for contracts shown to ViewOnly users

shrabantipaul-collate:
### Describe your changes:

Fixes https://github.com/open-metadata/OpenMetadata/issues/26119

I worked on fixing the Contract tab action menu ignoring RBAC permissions because users with only ViewAll/ViewBasic on the dataContract resource were seeing Import, Import ODCS, Edit, Run now, and Delete options that they are not authorized to use.

ContractDetail.tsx built the action menu items unconditionally with no permission check.
ContractTab.tsx was reading permissions.EditAll from useGenericContext, which reflects the parent table entity's permissions — not the dataContract resource's permissions. A user with EditAll on dataContract but not on table was incorrectly treated as having no edit access.
dataContract was missing from the ResourceEntity enum, making it impossible to fetch contract-level permissions via usePermissionProvider.
The fix fetches the current user's permissions against the specific contract entity (/api/v1/permissions/dataContract/{id}) after the contract loads, and uses that to gate the action menu items. Export and Export as ODCS remain visible to all users with view access.

I worked on ... because ...

image

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Fixed RBAC violations in Contract actions:**
  - Restricted Import, Edit, Run Now, Delete to users with `EditAll` on the dataContract resource (not parent entity)
  - Added `DATA_CONTRACT` to ResourceEntity enum to enable contract-level permission queries
- **Enhanced permission handling:**
  - Fetch contract-specific permissions via `usePermissionProvider().getEntityPermission()` after contract loads
  - Gate action menu items (`ContractDetail.tsx`) and empty state UI based on `hasEditPermission` prop
  - Export options remain visible to all users with view access
- **Improved empty state UX:**
  - View-only users see error placeholder without add contract button
  - Edit-permitted users see add contract button with import options
- **Updated tests and E2E:** Added permission scenarios to ContractTab tests; added dropdown wait conditions to ODCSImportExportPermissions.spec.ts

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/DataContract/ContractTab/ContractTab.test.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/DataContract/ContractTab/ContractTab.test.tsx
@@ -33,6 +33,7 @@ import { ContractTab } from './ContractTab';
 jest.mock('../../../context/PermissionProvider/PermissionProvider', () => ({
   usePermissionProvider: jest.fn(() => ({
     getEntityPermission: jest.fn().mockResolvedValue({}),
+    getResourcePermission: jest.fn().mockResolvedValue({}),
   })),
 }));
 
@@ -52,7 +53,6 @@ jest.mock('../../Customization/GenericProvider/GenericProvider', () => ({
       id: 'table-1',
       name: 'test-table',
     },
-    permissions: { EditAll: false },
   })),
 }));
 
@@ -432,7 +432,6 @@ describe('ContractTab', () => {
 
       mockUseGenericContext.mockReturnValue({
         data: { id: 'table-2', name: 'different-table' },
-        permissions: { EditAll: false },
       });
 
       rerender(<ContractTab />);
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/ODCSImportExportPermissions.spec.ts
```diff
@@ -244,7 +244,9 @@ test.describe('ODCS Import/Export - RBAC Permissions', () => {
       await navigateToContractTab(page, tableWithContract);
 
       await page.getByTestId('manage-contract-actions').click();
-
+      await page.waitForSelector('.contract-action-dropdown', {
```
Rohit0301: can we use dataTestId for dropdown, instead of css selector ?

shrabantipaul-collate: The data-testid attribute is not available by default in antd dropdown overlay. To add test id we need to create another div inside the dropdown.

dropdownRender={(menu) => (<div data-testid="contract-action-dropdown">{menu}</div>)}

Rohit0301 (approved):

PR #26206, opened 2026-03-04
Fix #25800: Agent status not showing true last 5 runs

### LINKED ISSUE #25800: Agent status not showing true last 5 runs
With the current implementation, the agent's recent runs, which are shown, are the last 5 runs within the last 24 hours. It should be last 5 runs overall.

**Backend:**
- API implementation to eliminate the necessity of startTs and endTs and support last runs based on count

**UI:**
- Implement the API to show the last 5 runs

harshach:
### Describe your changes:

Fixes #25800 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Backend fixes:**
  - Added `listBetweenTimestampsByOrderWithLimit()` method to `EntityTimeSeriesDAO` for fetching limited time-series results
  - Added batch operation `getLatestExtensionBatch()` to fetch latest statuses for multiple pipelines in single query
  - Modified `listPipelineStatus()` to return 5 latest runs by default when no time range provided
  - Updated `IngestionPipelineRepository` to batch-fetch pipeline statuses for improved performance
- **Frontend changes:**
  - Replaced timestamp-based queries with `limit=5` parameter in ingestion pipeline status calls
  - Updated `IngestionListTable`, `IngestionRecentRuns`, and test utilities to use new limit-based API
- **Tests and UI:**
  - Added comprehensive test for 5-run display behavior in `ServiceIngestion.spec.ts`
  - Minor formatting fixes in `TestSuiteSummaryWidget`


openmetadata-ui/src/main/resources/ui/src/components/Settings/Services/Ingestion/IngestionRecentRun/IngestionRecentRuns.component.tsx
```diff
@@ -55,14 +51,9 @@ export const IngestionRecentRuns = <
...
         const ingestionPipeline = ingestion as IngestionPipeline;
-
-        const queryParams = {
-          startTs: getEpochMillisForPastDays(1),
-          endTs: getCurrentMillis(),
-        };
         const response = await getRunHistoryForPipeline(
           ingestionPipeline?.fullyQualifiedName ?? '',
-          queryParams
+          { limit: 5 }
```
anuj-kumary, line 56: Is there any specific reason we have set the limit to 5 here?
Can we make it dynamic or fetch it from a constant like we do in other APIs?

Rohit0301: Previously, we were showing the pipeline status for the past 1 day. However, on the next day the user was unable to see any status because there were no records within that 1-day range. Therefore, instead of displaying results based on the start and end time, we are now showing only the last 5 scans.

PR #26221, opened 2026-03-04
Fix continuous polling loop in RunIngestionPipelineImpl causing log flooding

yan-3005:
## Problem

`waitForCompletion()` had a critical bug where `QUEUED`/`RUNNING` pipeline states fell through without any delay, causing tight polling loops. This resulted in continuous calls to `listPipelineStatus()` → `getQueuedPipelineStatus()`, flooding logs and overwhelming the server. The original code also used `Thread.sleep` which blocks execution threads.

## Solution

Replace `Thread.sleep`-based polling with Resilience4j retry mechanisms and fix the missing delay for `QUEUED`/`RUNNING` states.

### `RunIngestionPipelineImpl.java`
- **`runIngestionPipeline()`**: Replaced manual retry loop + `Thread.sleep` with a Resilience4j `Retry` using fixed 15-second intervals between 3 attempts, retrying only on `IngestionPipelineDeploymentException`
- **`waitForCompletion()`**: Replaced unbounded loop with a Resilience4j `Retry` supplier that:
  - Polls every 30 seconds (fixed interval)
  - Returns `"RUNNING"` for empty status lists (was falling through without delay)
  - Returns `"SUCCESS"`, `"FAILED"`, or `"TIMEOUT"` as terminal states
  - Uses `Integer.MAX_VALUE` max attempts with timeout enforced explicitly
- Added status string constants (`SUCCESS`, `FAILED`, `RUNNING`) and `WAIT_FOR_PIPELINE_COMPLETION` retry name
- Added package-private `pollingIntervalMillis` and `runRetryIntervalMillis` fields to allow overriding in tests

### `RunIngestionPipelineImplTest.java`
- Added unit tests covering: success, failure, timeout, polling through running states, and empty status list scenarios
- Tests override polling intervals to avoid real delays

## Impact

- Eliminates continuous `getQueuedPipelineStatus` log flooding
- Reduces server load with predictable 30-second polling intervals (120 checks per hour max)
- Maintains 3600-second default timeout behavior
- Follows Resilience4j patterns already used in the codebase

Fixes https://github.com/open-metadata/openmetadata-collate/issues/3085

### Type of change
- [x] Bug fix

### Checklist
- [x] My PR title is `Fix continuous polling loop in RunIngestionPipelineImpl causing log flooding`

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/createAndRunIngestionPipeline/RunIngestionPipelineImpl.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/createAndRunIngestionPipeline/RunIngestionPipelineImpl.java
@@ -101,8 +101,7 @@ public class RunIngestionPipelineImpl {
                   Entity.getEntity(
                       ingestionPipeline.getService(), "ingestionRunner", Include.NON_DELETED)));
     } catch (Exception ex) {
-      throw new RuntimeException(
-          "Failed to run pipeline after retries: " + ex.getMessage(), ex);
+      throw new RuntimeException("Failed to run pipeline after retries: " + ex.getMessage(), ex);
     }
   }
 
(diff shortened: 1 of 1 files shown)


yan-3005:
gitar bot, fix the review comments, before pushing run mvn spotless:apply on the root of the repository and push it

yan-3005:
gitar bot, you brought back the Thread.sleep, i am not interested in having the Thread sleeps, that is why i moved it to resilience4j

yan-3005:
gitar bot, fix only the review comment, always run mvn spotless:apply on the root of the repository and push it

yan-3005:
gitar bot, update the PR description with the changes you have made

PR #26301, opened 2026-03-06
Chore(UI): Fix the flakiness in ServiceIngestion.spec

aniketkatkar97:
This pull request refactors the pipeline triggering logic in the `ServiceBaseClass` for Playwright tests, improving reliability and removing unnecessary notification code. The main change is to explicitly wait for the pipeline trigger API response instead of relying on toast notifications or generic network idle states.

Improvements to pipeline triggering logic:

* Replaced the use of the `toastNotification` utility with explicit waiting for the `/api/v1/services/ingestionPipelines/trigger/` API response after clicking the `run-button`, ensuring that the pipeline trigger is confirmed by the backend before proceeding. [[1]](diffhunk://#diff-27b613009179a959c1e5cad0edabbe85bc092d0edbbd460b939fac8b3eb2e4d7R252-R264) [[2]](diffhunk://#diff-27b613009179a959c1e5cad0edabbe85bc092d0edbbd460b939fac8b3eb2e4d7L651-R668)
* Removed the import of `toastNotification` from `../../../utils/common` as it is no longer used.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/support/entity/ingestion/ServiceBaseClass.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/support/entity/ingestion/ServiceBaseClass.ts
@@ -25,7 +25,6 @@ import {
   getApiContext,
   INVALID_NAMES,
   NAME_VALIDATION_ERROR,
-  toastNotification,
 } from '../../../utils/common';
 import { visitEntityPage } from '../../../utils/entity';
 import { visitServiceDetailsPage } from '../../../utils/service';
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/playwright/support/entity/ingestion/ServiceBaseClass.ts
```diff
@@ -250,11 +249,19 @@ class ServiceBaseClass {
...
+          .includes('/api/v1/services/ingestionPipelines/trigger/') &&
+        response.status() === 200
+    );
     await page.getByTestId('run-button').click();
 
-    await page.waitForLoadState('networkidle');
+    await triggerPipeline;
 
-    await toastNotification(page, `Pipeline triggered successfully!`);
+    await page.waitForLoadState('networkidle');
```
ShaileshParmar11, line 264: why we are using netwrokidle, its discourage from playwright side?

aniketkatkar97: It was there from before. Looking at the use, there are multiple pipeline status calls possible there, which we can't intercept individually so I think we can make an exception here.

aniketkatkar97: It was there from before. Looking at the use, there are multiple pipeline status calls possible there, which we can't intercept individually so I think we can make an exception here.

PR #26315, opened 2026-03-08
Fix: respect fields parameter for table columns in GET API

itaise:
In our processes - we would like to be able to get tables without their columns - for performance reasons. 
Apparently - using GET table without "columns" in the fields parameter is not ommiting the columns. 
Changes made:
- TableRepository.clearFields() now clears columns when 'columns' is not in the requested fields, so GET table without fields=columns no longer returns column data.
- Add integration test test_getTableWithoutColumnsInFields_returnsNoColumns to verify behavior for get by id and get by name with/without columns.

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java
@@ -1336,6 +1336,41 @@ public class TableResourceIT extends BaseEntityIT<Table, CreateTable> {
     assertEquals(2, response.getData().size());
   }
 
+  @Test
+  void test_getTableWithoutColumnsInFields_returnsNoColumns(TestNamespace ns) {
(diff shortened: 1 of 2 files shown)


harshach:
@itaise this is backward incompatible change. Columns are stored in Table json anyway. If we are retieving the payload it will read from db. So not sure what exactly the perf issue you are running into. Can you please provide details what latencies you are hitting here

isevitt:
> @itaise this is backward incompatible change. Columns are stored in Table json anyway. If we are retieving the payload it will read from db. So not sure what exactly the perf issue you are running into. Can you please provide details what latencies you are hitting here

Thanks for your response. 
I am aware that we will get the columns either way, but sometimes we have huge (even 50K columns) schemas, and we have automated processes which query OMD api to fetch tables metadata.
 If we filter out the columns before returning them to the client it can improve performance, as we only need the table metadata (owner, domain etc..) without the huge columns schema.

PR #26331, opened 2026-03-09
fix(ui): Pipeline logs breadcrumb and title show entity name or displayName instead of id

### LINKED ISSUE #26330: Pipeline logs breadcrumb show either the entity name or the name of the bundle suite instead of id

shah-harshit:
## Description

Pipeline logs page breadcrumb and title now show the pipeline **displayName** or **name** from the ingestion entity instead of the pipeline id from the URL.

Issue:
Screenshot 2026-03-09 at 1 02 08 PM

Fix:
Screenshot 2026-03-09 at 1 01 40 PM

## Changes

- **LogsClassBase**: Use `getEntityName(ingestionDetails)` for the OpenMetadata pipeline breadcrumb and for the last segment in regular service-type breadcrumbs, so the label shows displayName/name rather than the pipeline id.
- **LogsViewerPage**: Page title uses `getEntityName(ingestionDetails)` or `getEntityName(appData)` so the header shows a readable name.
- **Unit tests**: Added tests in `LogsClassBase.test.ts` for displayName in breadcrumbs when `ingestionName` is a pipeline id (e.g. UUID).

Fixes #26330

### DIFF
--- a/ingestion/src/metadata/clients/domo_client.py
+++ b/ingestion/src/metadata/clients/domo_client.py
@@ -29,7 +29,8 @@ from metadata.generated.schema.entity.services.connections.database.domoDatabase
 from metadata.generated.schema.entity.services.connections.pipeline.domoPipelineConnection import (
     DomoPipelineConnection,
 )
-from metadata.ingestion.ometa.client import REST, ClientConfig
+from metadata.ingestion.connections.source_api_client import TrackedREST
+from metadata.ingestion.ometa.client import ClientConfig
 from metadata.utils.helpers import clean_uri
 from metadata.utils.logger import ingestion_logger
 
@@ -109,7 +110,7 @@ class DomoClient:
             auth_header="Authorization",
             auth_token=lambda: ("no_token", 0),
         )
-        self.client = REST(client_config)
+        self.client = TrackedREST(client_config)
 
     def get_chart_details(self, page_id) -> Optional[DomoChartDetails]:
         """
--- a/ingestion/src/metadata/ingestion/api/step.py
+++ b/ingestion/src/metadata/ingestion/api/step.py
@@ -24,6 +24,8 @@ from metadata.ingestion.api.models import Either, Entity, StackTraceError
 from metadata.ingestion.api.status import Status
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
 from metadata.utils.logger import ingestion_logger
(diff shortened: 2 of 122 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/LogsViewer.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+      await page.waitForURL(/\/testSuite\/.*\/logs/);
+
+      await page.waitForResponse((r) => {
+        const url = r.url();
+        return (
+          url.includes('/api/v1/services/ingestionPipelines/logs/') &&
+          url.includes('/last') &&
+          r.status() === 200
+        );
+      });
```
ShaileshParmar11: this will lead to flakiness
1st store API
<ACTION>
Await on API

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/LogsViewer.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+      await expect(page.getByTestId('summary-card')).toBeVisible();
+
+      const hasLogContent = await page.getByTestId('lazy-log').isVisible();
+      const hasNoLogsEmptyState = await page
+        .getByTestId('no-data-placeholder')
+        .filter({
+          has: page.getByText(/no .* log|logs.*available/i),
+        })
+        .isVisible()
+        .catch(() => false);
```
ShaileshParmar11: we do not need catch here. isVisible will return false if element is not present

openmetadata-ui/src/main/resources/ui/playwright/support/entity/TableClass.ts
```diff
@@ -375,6 +376,81 @@ export class TableClass extends EntityClass {
...
+      .then((res) => res.json());
+
+    this.bundleTestSuiteResponseData = testSuiteData;
+
+    return testSuiteData;
+  }
+
+  async createBundleTestSuitePipeline(apiContext: APIRequestContext) {
+    const testSuiteData = this.bundleTestSuiteResponseData;
+    if (!testSuiteData?.id || !testSuiteData?.fullyQualifiedName) {
```
ShaileshParmar11: bundle suite is not part of table entity, we need to move this function outside or create separate class.

Issue #25809, opened 2026-02-11
Data Insight: Implement incremental (delta) indexing for time-series data


manerow:
Parent: #25807

## Problem

Every Data Insight run processes **all entities** regardless of whether they changed — 100K+ entity reads, 100K+ version history lookups, hundreds of thousands of documents generated — even if only 50 entities were modified. Report data workflows use a delete-reinsert pattern that rewrites all records for the target date.

## Solution

**Track the last successful run timestamp** and only process entities modified since then.

- **Data Assets**: Filter entities by `updatedAt > lastSuccessfulRun` via `PaginatedEntitiesSource`. Generate snapshots only for the delta period.
- **Report Data**: Switch from delete-reinsert to append/upsert using composite document IDs. Skip dates with no new source data.
- **Fallback**: Retain full reindex via `recreateDataAssetsIndex` flag and `reindexdi` CLI for schema changes, recovery, and backfill.

**Expected impact** (100K entities, ~1% daily change): entities processed drops from ~100K to ~1K; run time from minutes-to-hours to seconds-to-minutes.

## Files to Modify

- `DataInsightsApp.java` — store/retrieve last successful run timestamp per entity type
- `DataAssetsWorkflow.java` — filter by `updatedAt > lastRun`, skip range deletion for incremental runs
- `DataInsightsEntityEnricherProcessor.java` — narrow snapshot window to delta period
- `PaginatedEntitiesSource.java` — add `updatedAt` filter support
- `ReportDataSink.java` — switch to upsert pattern
- `ReportDataRepository.java` / `EntityTimeSeriesDAO.java` — delta-aware queries

## Acceptance Criteria

- [ ] Only entities modified since last successful run are processed
- [ ] Last run timestamp persisted per entity type and survives restarts
- [ ] Report data uses append/upsert instead of delete-reinsert
- [ ] Periodic full reindex (e.g. weekly) configurable as consistency safeguard
- [ ] Hard-deleted entities detected and their snapshots cleaned from the index
- [ ] First run after enabling incremental mode performs a full reindex as baseline

> See [`DATA_INSIGHT_IMPROVEMENTS.md`](https://github.com/open-metadata/OpenMetadata/blob/1.11.9/DATA_INSIGHT_IMPROVEMENTS.md) (Section 4) for full design, delta detection options, and risk analysis.

manerow:
[...]

PR #26347, opened 2026-03-09
Fixes #26344: Add iterative bisection on 413 and TaggedOperation refactor in bulk index sinks 

### LINKED ISSUE #26344: Fix: Add iterative bisection on 413 and TaggedOperation refactor in bulk index sinks
## Problem

A single entity document can exceed the OpenSearch/Elasticsearch server payload limit (e.g., a table with hundreds of columns). When this happens, the bulk write fails with 413 and crashes the entire Data Insights workflow — no entities from that batch get indexed.

Additionally, the bulk index sinks have no way to identify *which* entity caused a failure because entity identity is not carried alongside the bulk operation.

## Proposed Fix

Two changes:

1. **Iterative bisection on 413** — When a bulk request is rejected with 413, split the batch in half and retry each half using a `Deque`-based work queue. Keep splitting until each piece fits. If a single document alone exceeds the limit, log it as a failed entity and skip it so the rest of the batch proceeds.

2. **TaggedOperation refactor** — Introduce a `TaggedOperation<T>` record that pairs each bulk operation with its entity FQN. This replaces the untyped `Map<String, Object> contextData` parameter in the `Sink.write()` interface, giving every stage in the pipeline typed access to entity identity. The `ReportDataSink` should receive its `reportDataType` via constructor injection instead of the context map.

manerow:
Fixes #26344

## Problem

When a bulk indexing request to OpenSearch/Elasticsearch is too large (HTTP 413), the entire Data Insights batch crashes — none of the entities in that batch get indexed. There is no recovery, no retry, and no way to tell which entity caused the failure.

On top of that, the existing error reporting was broken: when an entity failed, the sink tried to identify it by matching its position in the operations list against a separate list of entity names passed through an untyped `Map<String, Object>`. But **no Data Insights workflow ever populated that list**, so every error was reported as `"Unknown entity at index N"`.

### 1. Iterative bisection on 413

When a bulk request is rejected with 413, instead of failing the entire batch:
- The sink splits the batch in half and retries each half.
- If a half still fails, it splits again — all the way down to individual documents.
- If a single document is too large, it's logged as a failed entity and skipped. The rest of the batch continues.

This applies to both `ElasticSearchIndexSink` and `OpenSearchIndexSink`.

### 2. TaggedOperation — entity identity travels with the operation

A new `TaggedOperation<T>` record pairs each bulk operation with the `EntityReference` of the entity it belongs to. Each operation is tagged at the point where the entity is still known (in the workflow), so the sink always knows which entity it's dealing with.

**Why this matters for bisection:** when a batch of 100 operations is split into halves, then quarters, then individual docs — the entity reference stays attached to each operation throughout. Without this, the sink would have no idea which entity failed after splitting.

**Why this is better than what we had:** the old approach passed entity names as a separate list in a `Map<String, Object>` alongside the operations, and tried to correlate them by array index. This was:
- Never populated by Data Insights workflows (errors showed "Unknown entity")
- Fragile — any mismatch between the two lists would silently attribute errors to wrong entities
- Impossible to maintain through bisection splits

### 3. Sink interface cleanup

With `TaggedOperation`, the `Sink.write()` method no longer needs the `contextData` map to identify entities. The signature simplifies from `write(data, contextData)` to `write(data)`.

`ReportDataSink` previously read its `reportDataType` from the context map — now it receives it as a typed constructor parameter. The unused constants `REPORT_DATA_TYPE_KEY` and `ENTITY_NAME_LIST_KEY` are removed.

### 4. Time-series processor simplified to single-entity processing

`ElasticSearchEntityTimeSeriesProcessor` and `OpenSearchEntityTimeSeriesProcessor` now process one entity at a time instead of a batch. `DataQualityWorkflow` iterates entities and tags each operation inline — the same pattern used by `DataAssetsWorkflow`. This eliminates an implicit assumption that the processor returns exactly one operation per entity (which was correct but never enforced).

No performance impact: the previous "batch" processing was just a `for` loop internally — one iteration per entity, no shared computation, no bulk API call. The loop moved from inside the processor to the workflow caller. The actual I/O (the bulk write to ES/OS) still happens in a single `searchIndexSink.write()` call with the full batch.

### 5. JUnit 5 version alignment

`mockito-junit-jupiter:5.14.2` pulls `junit-jupiter-api:5.11.x` transitively, but the project pinned JUnit at `5.9.3`. This caused a `NoSuchMethodError` on `ReflectionUtils.returnsPrimitiveVoid` when running any `openmetadata-service` unit test locally. Fixed by adding `junit-bom:5.11.4` to the root POM's `<dependencyManagement>` and removing the hardcoded `5.9.3` versions from `openmetadata-service/pom.xml`, so all JUnit artifacts resolve to a consistent version.

### Type of change:
- [x] Bug fix
- [x] Improvement (existing functionality)

### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/processors/CreateReportDataProcessor.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/processors/CreateReportDataProcessor.java
@@ -20,10 +20,13 @@ import org.openmetadata.service.workflows.interfaces.Processor;
 public class CreateReportDataProcessor implements Processor<List<ReportData>, List<?>> {
   @Getter private final String name;
   private final StepStats stats = new StepStats();
+  private final ReportData.ReportDataType reportDataType;
 
-  public CreateReportDataProcessor(int total, String name) {
+  public CreateReportDataProcessor(
+      int total, String name, ReportData.ReportDataType reportDataType) {
     this.stats.withTotalRecords(total).withSuccessRecords(0).withFailedRecords(0);
     this.name = name;
+    this.reportDataType = reportDataType;
   }
 
   @Override
@@ -31,9 +34,6 @@ public class CreateReportDataProcessor implements Processor<List<ReportData>, Li
       throws SearchIndexException {
     List<ReportData> reportDataList = new ArrayList<>();
     try {
-      // TODO: Branch different ReportDataType process instead of having a generic Object.
-      ReportData.ReportDataType reportDataType =
-          (ReportData.ReportDataType) contextData.get("ReportDataType");
       Long timestamp = (Long) contextData.get(TIMESTAMP_KEY);
 
       for (Object eventData : input) {
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/workflows/costAnalysis/CostAnalysisWorkflow.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/workflows/costAnalysis/CostAnalysisWorkflow.java
@@ -223,7 +223,8 @@ public class CostAnalysisWorkflow {
     CreateReportDataProcessor createReportdataProcessor =
         new CreateReportDataProcessor(
             rawCostAnalysisReportDataList.size(),
-            "[CostAnalysisWorkflow] Raw Cost Analysis Report Data Processor");
+            "[CostAnalysisWorkflow] Raw Cost Analysis Report Data Processor",
+            ReportData.ReportDataType.RAW_COST_ANALYSIS_REPORT_DATA);
 
     Optional<List<ReportData>> rawCostAnalysisReportData = Optional.empty();
 
@@ -295,7 +296,8 @@ public class CostAnalysisWorkflow {
       CreateReportDataProcessor createReportdataProcessor =
           new CreateReportDataProcessor(
               aggregatedCostAnalysisReportDataList.get().size(),
-              "[CostAnalysisWorkflow] Aggregated Cost Analysis Report Data Processor");
+              "[CostAnalysisWorkflow] Aggregated Cost Analysis Report Data Processor",
+              ReportData.ReportDataType.AGGREGATED_COST_ANALYSIS_REPORT_DATA);
       Optional<List<ReportData>> aggregatedCostAnalysisReportData = Optional.empty();
 
       try {
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/workflows/webAnalytics/WebAnalyticsWorkflow.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/workflows/webAnalytics/WebAnalyticsWorkflow.java
@@ -211,7 +211,8 @@ public class WebAnalyticsWorkflow {
     CreateReportDataProcessor createReportDataProcessor =
         new CreateReportDataProcessor(
             entityViewReportData.values().size(),
-            "[WebAnalyticsWorkflow] Entity View Report Data Processor");
+            "[WebAnalyticsWorkflow] Entity View Report Data Processor",
+            ReportData.ReportDataType.WEB_ANALYTIC_ENTITY_VIEW_REPORT_DATA);
 
     Optional<List<ReportData>> entityViewReportDataList = Optional.empty();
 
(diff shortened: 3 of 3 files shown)


manerow:
@gitar-bot 

> ⚠️ Edge Case: sendWithBisection returns null when all ops fail with 413

The existing error-handling flow already covers this. The null return is unreachable in the all-fail scenario because `entityErrorList` would be non-empty, triggering the `SearchIndexException` throw on line 96. This finding is a false positive on closer inspection.

> ⚠️ Edge Case: DataQualityWorkflow assumes 1:1 mapping between ops and entities

Fixed. The time-series processors (`ElasticSearchEntityTimeSeriesProcessor` and `OpenSearchEntityTimeSeriesProcessor`) have been refactored from batch processing to single-entity processing. `DataQualityWorkflow` now iterates entities one at a time and tags each operation inline — the same pattern used by `DataAssetsWorkflow`. The 1:1 assumption is eliminated by design: there is no longer a separate list of operations to correlate by index.

> ⚠️ Quality: No tests for the 413 bisection logic

Fixed. Added two tests to both `ElasticSearchIndexSinkTest` and `OpenSearchIndexSinkTest`:
- `testSingleDocumentExceedsPayloadLimit413` — single doc triggers 413, verified it's recorded as a failed entity
- `testBisectionOnPayloadTooLargeWithMixedResults` — batch of 4, exercises full bisection → nested bisection → single-doc skip, verifies 5 bulk calls, 3 success, 1 failure

> 💡 Quality: Duplicated sendWithBisection logic across ES and OS sinks

Acknowledged but not actioned. The duplication is a pre-existing pattern in the codebase — the ES and OS sinks have always been near-identical copies because the Elasticsearch and OpenSearch client libraries have different type hierarchies (`BulkOperation`, `BulkResponse`, exception classes are all different types). Extracting a shared template method would require generics gymnastics that adds complexity for marginal benefit.

TeddyCr:
> If a single document is too large, it's logged as a failed entity and skipped. The rest of the batch continues.
👆by this do you mean if I have 4 documents, we 2 x 2 then 1 x 1 x 1 x 1, let's 1 on the left fails we then move on to 1 on the right, and so on?

TeddyCr:
> 4. Time-series processor simplified to single-entity processing
Do you see any values in processing time series entities that have a timestamp greater than the reindexing execution date or this will be tackled in the incremental processing?

manerow:
> by this do you mean if I have 4 documents, we 2 x 2 then 1 x 1 x 1 x 1, let's 1 on the left fails we then move on to 1 on the right, and so on?

@TeddyCr Yes, exactly. HTTP 413 is the error OpenSearch/Elasticsearch returns when the bulk request body exceeds the server's `http.max_content_length`.

Without the bisection, the entire batch would be discarded when this happens. The bisection is a safety measure to reduce the amount of unprocessed data to the absolute minimum. Instead of losing the whole batch, we narrow down to the specific document(s) that caused the overflow:

1. Try [1,2,3,4] → 413
2. Split into [1,2] and [3,4]
3. Try [1,2] → 413 again
4. Split into [1] and [2]
5. Try [1] → 413 → single doc, can't split further, log it as failed and skip it
6. Try [2] → success
7. Try [3,4] → success

So in the worst case, only the individual oversized document is skipped. Everything else gets processed.

manerow:
> Do you see any values in processing time series entities that have a timestamp greater than the reindexing execution date or this will be tackled in the incremental processing?

@TeddyCr Time-series entities with a timestamp after the execution date shouldn't normally exist since the data is historical. If they do show up, it's likely clock skew or a data ingestion bug. Filtering them out would be a reasonable guardrail.

On the incremental processing topic (#25809): I left a comment there I wanted to flag to you: [#25809 (comment)](https://github.com/open-metadata/OpenMetadata/issues/25809#issuecomment-4037614659).

TL;DR: PR #26337 already cuts ~81% of DI execution time. What it does: during enrichment, every entity triggers a version history query to the DB even if the entity hasn't changed since the last run. This commit skips those queries for unchanged entities by checking `updatedAt` against the processing window. It's a one-file, 11-line change with no architectural impact.

Full delta indexing (#25809) would require 6+ file changes, a new persistence layer for last-run timestamps, upsert patterns, and hard-delete detection. It also introduces risks like stale snapshots and broken dashboards. The dashboard issue is: DI dashboards assume one complete snapshot per entity per day. With delta indexing, only changed entities get new snapshots, so a query like "show me the state of all entities on March 5th" would return a mix of fresh snapshots for changed entities and stale/missing snapshots for unchanged ones. The dashboards would need to be reworked to resolve "latest known state at date X" instead of "snapshot at date X".

I think the cost/benefit doesn't justify it given what #26337 already delivers, but happy to discuss.

PR #26370, opened 2026-03-10
Fix OpenSearch semantic search script field reference

aamir306:
## Summary

This fixes an OpenSearch semantic search bug in `SemanticSearchQueryBuilder`.

The generated painless script currently uses:

```java
cosineSimilarity(params.query_vector, 'embedding')
```

That passes `'embedding'` as a string literal instead of referencing the document field, which causes OpenSearch semantic queries to fail with `search_phase_execution_exception` / `all shards failed`.

This patch changes it to:

```java
cosineSimilarity(params.query_vector, doc['embedding'])
```

## Changes

- update `SemanticSearchQueryBuilder` to use `doc['embedding']`
- add `SemanticSearchQueryBuilderTest` to verify the serialized query contains the correct field reference

## Why

I reproduced this on an OpenMetadata `1.12.1` deployment with OpenSearch-backed semantic search enabled. After repairing the index state, the remaining failure was the generated query script itself. Direct OpenSearch queries worked with `doc['embedding']`, while the shipped OMD query path failed until this was patched.

## Testing

Added:
- `SemanticSearchQueryBuilderTest`

Suggested verification:

```bash
mvn -pl openmetadata-service -Dtest=SemanticSearchQueryBuilderTest test
```

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/SemanticSearchQueryBuilder.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/SemanticSearchQueryBuilder.java
@@ -95,9 +95,9 @@ public class SemanticSearchQueryBuilder {
                                                                 ._types.BuiltinScriptLanguage
                                                                 .Painless))
                                                 .source(
-                                                    "cosineSimilarity(params.query_vector, '"
+                                                    "cosineSimilarity(params.query_vector, doc['"
                                                         + KNN_FIELD
-                                                        + "') + 1.0")
+                                                        + "']) + 1.0")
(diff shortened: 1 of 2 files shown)


ulixius9:
@aamir306 can you raise the pr for main instead of 1.12.1 branch? - we will cherry pick the fix when merged to the latest 1.12.x branch

pmbrull:
hi @aamir306 actually this is dead code from older implementations and is not being active in 1.12.X or main (which is deleted already). Thanks for the contribution, though I don't think this PR is valid. I am going to close it, thanks!

pmbrull:
if you're having any issues with OM & Semantic Search feel free to ping us in slack.open-metadata.org and we can review together the best way to move forward

PR #26376, opened 2026-03-10
fix(athena): handle missing CompletionDateTime in usage ingestion (#2…

PabloChanivet:
### Describe your changes:

Fixes #23841: handle missing CompletionDateTime in Athena usage ingestion

Athena Usage ingestion fails when processing query executions if the `CompletionDateTime` attribute is missing from the `Status` object.

Some Athena query executions returned by the AWS API do not include this attribute, which causes the ingestion workflow to fail with:

```
AttributeError: 'AthenaStatus' object has no attribute 'CompletionDateTime'
```

This PR fixes the issue by replacing the direct attribute access with a safe attribute lookup using `getattr`.

Before:

```python
if query.Status.CompletionDateTime
```

After:

```python
if getattr(query.Status, "CompletionDateTime", None)
```

This ensures that queries without `CompletionDateTime` are safely ignored instead of causing the ingestion pipeline to crash.

I worked on this change because Athena Usage ingestion failed during execution when encountering query metadata without `CompletionDateTime`. The change ensures the connector handles these cases safely.

Testing:
- Reproduced the issue locally during Athena Usage ingestion.
- Applied the fix and verified that ingestion completes successfully.
- Confirmed that queries without `CompletionDateTime` are skipped without raising exceptions.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/tests/unit/topology/database/test_athena.py
+++ b/ingestion/tests/unit/topology/database/test_athena.py
@@ -13,7 +13,8 @@ Test athena source
 """
 
 import unittest
-from unittest.mock import patch
+from datetime import datetime
+from unittest.mock import MagicMock, patch
 from uuid import UUID
 
 from pydantic import AnyUrl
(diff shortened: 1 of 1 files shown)


harshach:
@PabloChanivet would like to ship this into 1.12.2 this week. Could you please add a unit test for this

PabloChanivet:
> @PabloChanivet would like to ship this into 1.12.2 this week. Could you please add a unit test for this

@harshach added!

ulixius9 (approved):

PR #26397, opened 2026-03-11
Fix: hasDomain() search filter uses OR logic for multi-domain users (#26381)

### LINKED ISSUE #26381: Multi-Domains access - HasDomain condition
**Affected module**
UI and Backend (RBAC / Authorization, Search & Domain-based access control)

**Describe the bug**
When a group is assigned to multiple domains and a policy using the hasDomain condition is applied, users belonging to that group cannot access any assets in the UI.
This impacts key user journeys such as Explore, Domains pages, and Data Products pages (Glossary pages are working as expected).

Here is the policy assigned to the group : 
Resources: All
Operations: ViewAll
Effect: Allow
Condition: hasDomain()

This is the organization policy : 

Image

The Enable roles & Policies in search is on. (when we disactivate this, the user also sees assets from other domains, which does not meet the RBAC requirement)

This behavior prevents using domain-based RBAC for common use cases where one functional group needs access to multiple domains.

Here are some behavior I encountered during my test (with Roles & Policies in Search enabled) : 
Explore page
→ No assets are returned, even when filtering by domain

Image

Domains page
→ Domains show 0 assets and 0 Data Products

Image

Data Products pages
→ No Data Products are visible

Image

- If I assign two domains to an asset and the group has the same domains as the asset, the user can see the asset with the hasDomain condition 

Image

Policy debugger for an asset of an assigned domain : 

Image
 
result in explore view : the dashboard is not visible 

Image
 

**To Reproduce**
Setup
1. Create 2 domains and assign domain to assets
2. Ceate a group and assign both domains to the group
3. Assign the policy to the group 
4. Enable "Enable Roles & Policies in Search"
5. Log in as a user who belongs to the group

**Expected behavior**
A user should be able to:
- Access assets belonging to any domain assigned to their group
- NOT see assets from domains that are not assigned to their group

The hasDomain condition should work as an OR condition across multiple domains, not require a strict match with all domains at once.

**Version:**
 - OpenMetadata version: 1.11.7

**Additional context**
[Slack discussion related to this issue:
https://openmetadata.slack.com/archives/C02B6955S4S/p1772716333606479?thread_ts=1770309822.282879&cid=C02B6955S4S](url)

yan-3005:
### Describe your changes:

Fixes #26381 

 RBACConditionEvaluator.hasDomain() was calling addMust() once per user                                                                                                                
  domain, producing AND logic in the Elasticsearch/OpenSearch query. An
  entity must satisfy all domain conditions simultaneously, which is                                                                                                                    
  impossible since entities belong to a single domain — so multi-domain                                                                                                               
  users saw zero search results.

  Fix: collect all domain term queries into a should (OR) clause before
  adding as a single must clause, matching the existing isOwner() pattern.

  Add unit tests for both ES and OpenSearch evaluators asserting the
  generated query uses should semantics for multiple domains, and an
  integration test (MultiDomainHasDomainIT) that creates a user assigned
  to two domains and verifies search returns tables from both.

#
### Type of change:

- [X] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/search/security/RBACConditionEvaluator.java
```diff
@@ -351,11 +351,18 @@ public void hasDomain(ConditionCollector collector) {
...
+      List<OMQueryBuilder> domainQueries = new ArrayList<>();
       for (EntityReference domain : user.getDomains()) {
         String domainId = domain.getId().toString();
-        OMQueryBuilder domainQuery = queryBuilderFactory.termQuery("domains.id", domainId);
-        collector.addMust(domainQuery);
+        domainQueries.add(queryBuilderFactory.termQuery("domains.id", domainId));
       }
+      OMQueryBuilder domainQuery;
+      if (domainQueries.size() == 1) {
+        domainQuery = domainQueries.get(0);
```
harshach: use domainQueries.getFirst()

yan-3005:
gitar bot, fix the review comment, run mvn spotless:apply on the root of the repository before you push the change

yan-3005:
gitar bot, did you fix the review comment?

yan-3005:
gitar bot, the fix is not applied, apply it and push the code, use getFirst in that particular line of the review comment

yan-3005:
Updates:
Unable to reproduce the issue the User Reported to verify the fix. Would understand with the user first before moving with this PR
cc @ayush-shah

sonika-shah (approved):

PR #26031, opened 2026-02-21
ISSUE #20036 - sqlalchemy 2.0 migration

TeddyCr:
### Describe your changes:
Fixes #20036 
  - Migrate the Python ingestion framework from SQLAlchemy 1.4 to 2.0
  - Update dependency pins and resolve compatibility issues with third-party dialect packages
  - Fix all SA 2.0 breaking API changes across source connectors, profiler, data quality validators, and tests

  ### Motivation
  SQLAlchemy 1.4 is no longer maintained. SA 2.0 brings improved type safety, native async support, and a modernized ORM/Core API. This migration unblocks future improvements and keeps the ingestion framework on supported dependencies.

  ### Changes
  #### Dependency Updates (`setup.py`, `pyproject.toml`, `Dockerfile.ci`)
  - Core pin: `sqlalchemy>=1.4.0,<2` → `sqlalchemy>=2.0.0,<3`
  - `snowflake-sqlalchemy>=1.5` → `>=1.6.1` (1.5.x declares SA<2)
  - `sqlalchemy_exasol>=5,<6` → `>=6,<7` (v6+ supports SA 2.0)
  - `clickhouse-sqlalchemy~=0.2.0` → `>=0.3` (0.2.x imports removed internals)
  - Pre-install `sqlalchemy-redshift`, `sqlalchemy-databricks`, `sqlalchemy-ibmi` with `--no-deps` in Dockerfiles (unmaintained packages
  with SA<2 metadata that work at runtime)
  - Removed `sqlalchemy.ext.mypy.plugin` from `pyproject.toml` (not needed with SA 2.0 native typing)

  #### API Migrations (100+ source files)

  | SA 1.4 Pattern | SA 2.0 Pattern |
  |---|---|---|
  | `engine.execute(query)` | `with engine.connect() as conn: conn.execute(text(query))` |
  | `connection.execute("raw SQL")` | `connection.execute(text("raw SQL"))` |
  | `session.bind` | `session.get_bind()` |
  | `declarative_base()` / `DeclarativeMeta` | `DeclarativeBase` / `type` |
  | `case([(cond, val)], else_=x)` | `case((cond, val), else_=x)` |
  | `select([col1, col2])` | `select(col1, col2)` |
  | `dict(row)` | `row._asdict()` / `dict(row._mapping)` |
  | `row.keys()` | `row._mapping.keys()` |
  | `row["key"]` (string index) | `row._mapping["key"]` |
[...]

PR #26411, opened 2026-03-11
fix: SQLAlchemy 2.0 e2e fixes for Redshift, Oracle, Athena, MSSQL profiler

IceS2:
## Summary

Fixes multiple e2e test failures caused by the SQLAlchemy 2.0 migration (#26031).

### Redshift
- Override `_pg_class_filter_scope_schema` — Redshift lacks `pg_class.relpersistence` column used by PGDialect in SQA 2.0
- Override `get_multi_pk_constraint` and `get_multi_unique_constraints` — Redshift doesn't support `array_agg(... ORDER BY ...)` used by PGDialect's batch reflection
- Register `get_temp_table_names` override on `RedshiftDialect`

### Oracle
- Add `get_multi_columns` override — SQA 2.0's `OracleDialect` uses batch reflection via `get_multi_columns()` which bypasses our custom `get_columns()`, resulting in all tables being ingested with 0 columns

### Athena
- Replace private `Column._set_parent()` with public `Table.append_column(col, replace_existing=True)` — `_set_parent()` behavior changed in SQA 2.0

### Profiler / Sampler (all databases)
- `ometa_to_sqa_orm`: return `None` with warning instead of raising `ValueError` for tables with no columns (SQA 2.0 hard-errors on missing PK, was just a warning in 1.x)
- `ProfilerProcessor._run` and `SamplerProcessor._run`: skip tables with no columns early to avoid downstream `NoInspectionAvailable` crashes
- `SQASampler._partitioned_table`: build partitioned CTE inside session context — SQA 2.0 requires the session to be alive when `.cte()` is called on a Query (fixes MSSQL/Redshift time partition histogram returning `None`)
- `get_all_table_ddls`: add generic `connection.rollback()` after DDL reflection failures so subsequent queries don't fail with `InFailedSqlTransaction`

## Test plan

- [ ] Redshift e2e tests pass (ingestion, profiler, time partition)
- [ ] Oracle e2e tests pass (ingestion, profiler, auto-classify)
- [ ] MSSQL e2e tests pass (time partition histogram)
- [ ] Athena e2e tests pass (struct column profiling)
- [ ] Postgres e2e tests still pass (no regression)
- [ ] Snowflake e2e tests still pass (no regression)

### DIFF
--- a/.github/workflows/playwright-postgresql-e2e.yml
+++ b/.github/workflows/playwright-postgresql-e2e.yml
@@ -47,8 +47,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8]
-        shardTotal: [8]
+        shardIndex: [1, 2, 3, 4, 5, 6]
+        shardTotal: [6]
     steps:
       - name: Free Disk Space (Ubuntu)
         uses: jlumbroso/free-disk-space@main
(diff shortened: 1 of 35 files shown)


TeddyCr (changes requested):
@IceS2 a few things:
- patching `get_multi_columns` should not be required as this should be handled by the dialect directly with SQA2 https://github.com/sqlalchemy/sqlalchemy/blob/e7c45bae55e771044b30df9cdfce2c21b5177eac/lib/sqlalchemy/dialects/oracle/base.py#L2889-L2929
- For redshift this is a known issue. We have a PR opened in the repo directly. You can refer to the internal thread https://github.com/sqlalchemy-redshift/sqlalchemy-redshift/pull/326
-

PR #25622, opened 2026-01-29
Add Columns to indexing

harshach:
# Add Columns to indexing

Fixes https://github.com/open-metadata/openmetadata-collate/issues/2971

### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

---

----
## Summary by Gitar

- **New search indexing feature:**
  - Added `ColumnSearchIndex` for indexing columns with dedicated index mappings in Elasticsearch/OpenSearch
  - Implemented distributed indexing strategy with `BulkSink` and repository handlers for `GlossaryTerm` and `Tag`
- **Search infrastructure enhancements:**
  - Extended `SearchRepository` with column search capabilities and RBAC filtering for columns
  - Added `SearchIndexExecutor` for managing bulk operations and `DistributedIndexingStrategy` for multi-index distribution
- **UI improvements:**
  - Enhanced global search suggestions and explore filters with column search support
  - Updated `ExploreSearchCard` and search settings UI to handle column entities
- **Migration and configuration:**
  - Added v200 migration scripts for MySQL and PostgreSQL with `SearchSettingsMergeUtil`
  - Created multilingual column index mappings (English, Japanese, Russian, Chinese)

<sub>This will update automatically on new commits.</sub>

PR #26419, opened 2026-03-12
Add the patch API retrieval back, removed in #25622

aniketkatkar97:
This pull request updates the `EntitySummaryPanel` component to improve type safety and ensure robust entity update handling. The main changes include stricter typing for entity types and a fallback mechanism for updating entities when an API function is not found in the mapping.

Type safety improvements:

* Updated the typing for `entityType` and related variables to use `EntityType` and `ResourceEntity` directly, removing unnecessary `undefined` checks and ensuring consistent usage.

Entity update handling enhancements:

* Modified the logic for updating entities to use a fallback method (`entityUtilClassBase.getEntityPatchAPI(entityType)`) when an API function is not present in `entityUpdateMap`, ensuring updates are handled gracefully for all entity types. This change was applied in three places for consistency: tag updates, general entity updates, and description updates. [[1]](diffhunk://#diff-a7f07f71e2805dadb5469ecbefe874cbac929b770e16c8d6741dd3cc85a04d60L497-R499) [[2]](diffhunk://#diff-a7f07f71e2805dadb5469ecbefe874cbac929b770e16c8d6741dd3cc85a04d60L585-R589) [[3]](diffhunk://#diff-a7f07f71e2805dadb5469ecbefe874cbac929b770e16c8d6741dd3cc85a04d60L667-R673)


openmetadata-ui/src/main/resources/ui/src/components/Explore/EntitySummaryPanel/EntitySummaryPanel.component.tsx
```diff
@@ -169,22 +169,22 @@ export default function EntitySummaryPanel({
...
-    () => get(entityDetails, 'details.entityType') as EntityType | undefined,
+    () => get(entityDetails, 'details.entityType') as EntityType,
     [entityDetails]
   );
 
   const fetchResourcePermission = async (id: string) => {
     try {
       setIsPermissionLoading(true);
-      let type = get(entityDetails, 'details.entityType');
+      let type = get(entityDetails, 'details.entityType') as ResourceEntity;
```
anuj-kumary: Can we avoid using the type assertion here? It bypasses TypeScript’s type safety. We could rely on proper typing and optional chaining instead.

aniketkatkar97: Since the data we are relying on comes from the ES, we will be getting it raw, i.e. no custom enums or types we use in UI side. Since we are using common functions that are defined for certain types, had to do that to avoid type errors. Happy to hear any other solution if you have something in mind.

anuj-kumary: That makes sense given the data is coming raw from ES. Maybe we could narrow/validate the value before using it instead of asserting it directly, so we keep a bit more type safety.

ShaileshParmar11 (approved):

PR #26490, opened 2026-03-13
ISSUE #3058 - Bulk Endpoint For Bundle TestSuite Selection

### LINKED ISSUE #3058: Sort users within teams page
**Affected module**
UI 

**Describe the bug**
The list of users in the teams page is not sorted. 
To make it easily readable, we should sort it in lexicographical order

**To Reproduce**

Navigate to teams page and check Users tab
image

**Expected behavior**
Users are sorted

**Version:**
 - Python version:

**Additional context**
We may need to do this for all entities which are displayed on tabs within an entity page.

Eg:
- Users for a Team (this issue)
- Assets for a Team
- Roles for a Team

This could be possibly fixed from the backend by returning all entity reference lists as sorted lists.

cc: @harshach  @sureshms @vivekratnavel

TeddyCr:
### Describe your changes:

Fixes #3058   
  - Add new bulk endpoint PUT /v1/dataQuality/testCases/logicalTestCases/bulk that supports adding test cases to a logical test suite by explicit IDs (mode: "ids") or  selecting all test cases with optional filter (right now supports id exclusions (mode: "all"))
  - Deprecate the existing PUT /logicalTestCases endpoint with Deprecation and Sunset response headers (removal in 2.0)
  - Replace N+1 individual postUpdate calls with batch processing: bulk Entity.getEntities() fetch, postUpdateMany(), and bulk search index updates via SearchIndexHandler.onEntitiesUpdated()
  - Add onEntitiesUpdated to the lifecycle event system (EntityLifecycleEventHandler, EntityLifecycleEventDispatcher) with per-entity fallback for backward compatibility

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

Improvement

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/IncidentPaginationIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/IncidentPaginationIT.java
@@ -9,6 +9,8 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+import lombok.extern.slf4j.Slf4j;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
@@ -31,6 +33,7 @@ import org.openmetadata.sdk.client.OpenMetadataClient;
 import org.openmetadata.sdk.models.ListParams;
 import org.openmetadata.sdk.models.ListResponse;
 
+@Slf4j
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class IncidentPaginationIT {
 
@@ -83,10 +86,20 @@ public class IncidentPaginationIT {
   private void waitForDataIndexed() throws Exception {
     String firstFqn = testCases.get(0).getFullyQualifiedName();
     final Instant[] lastRecreation = {Instant.MIN};
+    AtomicReference<String> lastError = new AtomicReference<>("no attempts yet");
 
     await()
         .atMost(Duration.ofMinutes(3))
         .pollInterval(Duration.ofSeconds(5))
+        .conditionEvaluationListener(
+            condition -> {
+              if (!condition.isSatisfied()) {
+                log.warn(
+                    "waitForDataIndexed not satisfied after {} (last error: {})",
+                    condition.getElapsedTimeInMS() + "ms",
+                    lastError.get());
+              }
+            })
         .until(
             () -> {
               try {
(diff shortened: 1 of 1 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
```diff
@@ -1141,7 +1145,7 @@ public Response deleteFailedRowsData(
   @PUT
   @Path("/logicalTestCases")
   @Operation(
-      operationId = "addTestCasesToLogicalTestSuite",
+      operationId = "addTestCasesToBundleTestSuite",
```
harshach, line 1148: @TeddyCr why is this operationId changing?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -992,21 +993,68 @@ public void isTestSuiteBasic(String testSuiteFqn) {
...
+        findTo(testSuite.getId(), TEST_SUITE, Relationship.CONTAINS, TEST_CASE);
+
+    Set<UUID> originalIds =
+        originalTestCaseReferences.stream().map(EntityReference::getId).collect(Collectors.toSet());
+    List<EntityReference> newTestCaseReferences =
+        updatedTestCaseReferences.stream()
+            .filter(ref -> !originalIds.contains(ref.getId()))
+            .toList();
+
+    int batchSize = 100;
```
harshach, line 1051: should this be hard-coded?

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResource.java
TeddyCr: it is just to aligned with the renaming. They used to be called logical test suite but now are called bundle test suite from the ui. If I am not mistaking this is just openapi semantic right?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
TeddyCr: Since it is internal logic I replicated the approach we have here https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/sink/SinkTaskDelegate.java, LMK if you are thinking about a different approach

harshach: Lets check how are doing in other import/export. I beleive we are hard-coding the value to 1000 or something but need to check

Issue #21475, opened 2025-05-29
AutoClassification - Enable it for Containers & Streams


pmbrull:
use the sample data from containers & streaming sources for the auto classification

PR #26495, opened 2026-03-14
feat: Add auto-classification support for storage service containers

edg956:
## Summary

Implements #21475 

Add support for auto-classification (PII detection) on storage service containers (S3, GCS, etc.), enabling automatic tagging of sensitive data in files stored in cloud storage.

This extends OpenMetadata's existing auto-classification capabilities from database tables to storage containers with structured data (CSV, Parquet, etc.).

### Schema & API
- Add `sampleData` field to `container.json` schema for storing sample data
- Create `storageServiceAutoClassificationPipeline.json` schema defining configuration for storage service auto-classification workflows
- Add REST endpoints in `ContainerResource` for sample data operations: `PUT/GET/DELETE /{id}/sampleData`

### Backend (Java)
- **ContainerRepository**: Implement sample data persistence and retrieval from `entity_extension` table
- **EntityRepository**: Refactor `validateColumn()` to support both Table and Container column validation
- **PIIMasker**: Extend PII masking to support Container entities with proper tag-based masking
- **ContainerResource**: Add authorization for `VIEW_SAMPLE_DATA` and `EDIT_SAMPLE_DATA` operations

### Ingestion Framework (Python)
- **Storage Samplers**: Implement `StorageSampler` base class with S3 and GCS concrete implementations for reading structured files
- **Fetcher Strategy**: Add `StorageFetcherStrategy` for fetching Container entities from storage services
- **SamplerProcessor**: Extend to handle Container entities alongside Table entities
- **PII Processor**: Update to classify container columns using `ClassifiableEntityType` union (Table | Container)
- **Metadata Sink**: Add Container sample data ingestion via `OMetaContainerMixin`
- **Patch Mixin**: Support Container dataModel column tag updates

### Testing
- Add comprehensive integration tests for container classification (MinIO/S3 with PII detection)
- Add unit tests for `StorageFetcherStrategy` filtering and `SamplerProcessor` container handling
- Reorganize auto-classification tests by entity type (databases/ and containers/)

### Bug Fixes
- Fix sample data not being retrieved when requesting container with `fields=["sampleData"]` - added proper field handling in `ContainerRepository.setFields()`

## Type of Change
- [x] New feature
- [x] Bug fix (sample data retrieval)

### Unit Tests
```bash
cd ingestion
pytest tests/unit/profiler/test_container_fetcher.py
pytest tests/unit/sampler/test_container_sampler_processor.py
```

### Integration Tests
```bash
cd ingestion
pytest tests/integration/auto_classification/containers/
```

### Manual Testing
1. Configure a storage service (S3/GCS) with structured files containing PII
2. Run storage service metadata ingestion
3. Run storage service auto-classification pipeline with `storeSampleData: true`
4. Verify containers have:
   - PII tags on sensitive columns (email, SSN, credit card, etc.)
   - Sample data stored and retrievable via API
   - PII masking applied when user lacks authorization

## Checklist
- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document
- [x] I have added tests around the new logic
- [x] I have added a test that covers the bug fix scenario
- [x] For JSON Schema changes: Updated to add new pipeline type and container field
- [x] Code formatted with `mvn spotless:apply`, `make py_format`

----
## Summary by Gitar

- **Refactored Data Retrieval:**
  - Updated `EntityRepository` to filter out soft-deleted entities during batch relationship fetches (`NON_DELETED` instead of `ALL`).
  - Improved `batchFetchFollowers` and `batchFetchExperts` to handle null references and correct ID mapping logic.


edg956:
@gitar-bot I've updated my PR, but review your comments because I think some are not pertinent to this PR

edg956:
@gitar-bot I don't see how these suggestions apply to my PR. They may come from the same files I touched (I don't think so), but if they do please thorough in your explanation. If they're not really pertinent to this PR then ignore them

```
1. 💡 Edge Case: parseInt on FQN-named fields is fragile and unclear

   In `DataAssetsCoveragePieChartWidget`, `parseInt` (from lodash) is called on `coverageData[0].originEntityFQN` and `totalData[0].fullyQualifiedName` (lines 94-95). These field names strongly suggest string FQN values (e.g. `service.database.schema.table`), not numeric counts. While this may work if the API repurposes these fields for aggregate counts, it's fragile — if the API ever returns actual FQN strings, `parseInt` will silently produce `NaN`, leading to a broken chart. At minimum, the code lacks a guard for `NaN` results.

   Suggested fix:
   Add NaN guards after parsing, e.g.:
   const covered = parseInt(coverageData[0].originEntityFQN, 10);
   if (isNaN(covered)) { setDataAssetsCoverageStates(INITIAL_DATA_ASSETS_COVERAGE_STATES); return; }
   Also consider adding a code comment explaining why FQN-named fields contain numeric values.

2. ⚠️ Edge Case: MatillionDPC schema has no required fields, allows empty auth

   The `matillionDPC.json` schema defines `"required": []`, which means a MatillionDPCAuth object can be created with no authentication credentials at all — no `clientId`, no `clientSecret`, and no `personalAccessToken`. This would allow saving a connection that can never authenticate. At minimum, some form of credential should be required (e.g., either `clientId`+`clientSecret` or `personalAccessToken`).

   Suggested fix:
   Use a `oneOf` or `anyOf` constraint to require at least one authentication method, or add `clientId` and `clientSecret` to the `required` array:
   
   "required": ["clientId", "clientSecret"]

3. 💡 Quality: queryText not forwarded in independent (POST) aggregation path

   In `getAggregationOptions`, the new `queryText` parameter is only passed to the `getAggregateFieldOptions` (GET) path. When `isIndependent` is true, `postAggregateFieldOptions` is called without `queryText`, so the search text won't influence filter options in the independent mode. This may be intentional if the POST endpoint doesn't support it, but it's worth confirming the behavior is expected.
```

ShaileshParmar11 (dismissed):

ShaileshParmar11 (dismissed):

chirag-madlani (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ContainerRepository.java
```diff
@@ -78,6 +85,10 @@ public void setFields(
         fields.contains(FIELD_PARENT) ? getContainerParent(container) : container.getParent());
     container.setChildren(
         fields.contains("children") ? getChildren(container) : container.getChildren());
+    container.setSampleData(
+        fields.contains("sampleData")
+            ? getSampleDataInternal(container.getId())
+            : container.getSampleData());
```
manerow: I think this introduces a PII masking bypass. Tracing the two read paths:

`GET /v1/containers/{id}/sampleData` goes through `repository.getSampleData(id, authorizePII)` which applies `PIIMasker.getSampleData(container)` when the user isn't PII-authorized. Good.

With this change, the same user can call `GET /v1/containers/{id}?fields=sampleData` instead. `addViewOperation("sampleData", VIEW_SAMPLE_DATA)` gates field visibility, but `Authorizer#authorizePII` is a separate, stricter check that only returns true for admins/bots/owners. A regular DataConsumer has `ViewAll` (which covers `VIEW_SAMPLE_DATA` via `CompiledRule#matchOperation`), but would not get `authorizePII=true`. So through the fields path, `getSampleDataInternal` returns raw unmasked data, skipping the masker entirely.

`TableRepository.setFields()` (line 191) deliberately does not load `sampleData` for this reason: there's no access to `SecurityContext` or `Authorizer` from inside a repository's `setFields`, so masking cannot be decided.

My suggestion is to drop the sampleData load from `setFields` and keep the dedicated `/sampleData` endpoint as the single read path. The bot writes via `PUT /sampleData`, the UI reads via `GET /sampleData` (see `getSampleDataByContainerId` in the storage API), and both already work. The only caller that uses `fields=["sampleData"]` is the new python integration test `test_container_sample_data_stored`, which can be switched to `/sampleData` to match how the UI actually does it. That aligns Container behaviour with Table and keeps every reader going through the masking-aware path.

If keeping the fields-based read is important for some reason I'm not seeing, the alternative is to guard the load with an `isAdmin() || isBot()` check against the current subject, but that pulls subject context into the repository layer which is messier than the Table approach.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ContainerRepository.java
```diff
@@ -500,6 +511,76 @@ public ResultList<Container> listChildren(String parentFQN, Integer limit, Integ
...
+    try {
+      String json =
+          daoCollection
+              .entityExtensionDAO()
+              .getExtension(containerId, CONTAINER_SAMPLE_DATA_EXTENSION);
+      return json != null ? JsonUtils.readValue(json, TableData.class) : null;
+    } catch (Exception e) {
+      LOG.warn("Failed to read sample data for container {}", containerId, e);
+      return null;
+    }
```
manerow: Two things worth reconsidering on this `try/catch`:

1. `catch (Exception e)` isn't allowed by the project's Java guidelines in new code. If we want to catch anything here it should be the specific type we expect (e.g. the JSON parsing exceptions from `JsonUtils.readValue`).

2. Returning `null` on failure silently collapses "stored blob is corrupt" and "no sample data exists" into the same result. Both `TableRepository.getSampleData` and `FileRepository.getSampleData` let parse errors propagate, which is the right call because otherwise operators have no way to find out that sample data can't be read back.

Proposed fix is to drop the try/catch and match the reference implementations:

```java
private TableData getSampleDataInternal(UUID containerId) {
  String json = daoCollection.entityExtensionDAO()
      .getExtension(containerId, CONTAINER_SAMPLE_DATA_EXTENSION);
  return json == null ? null : JsonUtils.readValue(json, TableData.class);
}
```

Assuming the `setFields` concern above is resolved and this method ends up being called only from the dedicated `/sampleData` endpoint, a 500 on a corrupt extension is the right signal.

openmetadata-service/src/main/java/org/openmetadata/service/security/mask/PIIMasker.java
```diff
@@ -47,16 +51,27 @@ private PIIMasker() {
     /* Private constructor for Utility class */
   }
 
-  public static TableData maskSampleData(TableData sampleData, Table table, List<Column> columns) {
+  public static TableData maskSampleData(
+      TableData sampleData, Object entity, List<Column> columns) {
```
manerow: Accepting `Object` here is a concern on a class that decides what sensitive data gets returned to users. Three reasons I'd push back on it:

1. Type safety goes away. Callers can pass anything and only find out at runtime via a `LOG.warn`.

2. The `else` branch below (where the warn log is) falls through with `entityHasPiiTag = false`. If a caller mistakenly passes a `Topic` or a `SearchIndex`, the method treats the entity as having no sensitive tag, runs the "mask only PII columns" path, and emits any non-PII-tagged column unmasked. For a masker, the safe default on confusion should be "mask everything" or throw, not "pass through".

3. The rest of this class already follows a clean overload pattern: `hasPiiSensitiveTag(Table|Topic|SearchIndex|Query|Column)`, and now `hasPiiSensitiveTag(Container)` added in this PR. Only `maskSampleData` breaks from it.

Proposed shape is to keep typed overloads and push the polymorphism into a private helper that only depends on the boolean:

```java
public static TableData maskSampleData(TableData sampleData, Table table, List<Column> columns) {
  return maskSampleDataInternal(sampleData, hasPiiSensitiveTag(table), columns);
}

public static TableData maskSampleData(TableData sampleData, Container container, List<Column> columns) {
  return maskSampleDataInternal(sampleData, hasPiiSensitiveTag(container), columns);
}

private static TableData maskSampleDataInternal(
    TableData sampleData, boolean entityHasPiiTag, List<Column> columns) {
  // existing logic, using entityHasPiiTag in place of the instanceof dispatch
}
```

Existing callers in `TestCaseRepository` and `PIIMaskerTest` pass a `Table`, so they compile unchanged. Adding future entity types becomes a two-line addition (overload + `hasPiiSensitiveTag` helper), which matches how `getSampleData` is already structured in this class.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/createAndRunIngestionPipeline/CreateIngestionPipelineImpl.java
```diff
@@ -317,6 +329,8 @@ private Object getSourceConfig(PipelineType pipelineType, ServiceEntityInterface
     Map<String, FilterPattern> serviceDefaultFilters = getServiceDefaultFilters(service);
     if (entityType.equals(DATABASE_SERVICE)) {
       return DATABASE_PIPELINE_MAP.get(pipelineType).apply(serviceDefaultFilters);
+    } else if (entityType.equals(STORAGE_SERVICE)) {
+      return STORAGE_PIPELINE_MAP.get(pipelineType).apply(serviceDefaultFilters);
```
manerow: `STORAGE_PIPELINE_MAP` only has entries for `METADATA` and `AUTO_CLASSIFICATION`. If this method is called with any other `PipelineType`, `STORAGE_PIPELINE_MAP.get(pipelineType)` returns null and the immediate `.apply(...)` NPEs with no context.

This is reachable from what's in this PR: `SUPPORT_FEATURE_MAP` at the top of this file maps `PipelineType.PROFILER` to `supportsProfiler`, and the PR adds `"supportsProfiler": true` to `s3Connection.json`, `gcsConnection.json`, `adlsConnection.json`, `customStorageConnection.json`. That advertises a profiler pipeline for those services, which will route here and NPE.

`DATABASE_PIPELINE_MAP` doesn't have the issue because all 5 pipeline types are covered.

Two fixes that are worth doing together:

1. Add a defensive guard so the error is readable:

```java
} else if (entityType.equals(STORAGE_SERVICE)) {
  Function<Map<String, FilterPattern>, Object> factory = STORAGE_PIPELINE_MAP.get(pipelineType);
  if (factory == null) {
    throw new IllegalArgumentException(
        String.format("Pipeline type %s is not supported for storage services", pipelineType));
  }
  return factory.apply(serviceDefaultFilters);
}
```

2. Drop `supportsProfiler: true` from `adlsConnection.json` and `customStorageConnection.json` in this PR since those python samplers don't exist yet. Keep it on S3 and GCS where the feature actually ships, and bring the others back when their samplers land.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ContainerRepository.java
```diff
@@ -500,6 +511,76 @@ public ResultList<Container> listChildren(String parentFQN, Integer limit, Integ
...
+
+  @Transaction
+  public Container addSampleData(UUID containerId, TableData tableData) {
+    Container container = find(containerId, NON_DELETED);
+
+    if (container.getDataModel() != null && container.getDataModel().getColumns() != null) {
+      for (String columnName : tableData.getColumns()) {
+        validateColumn(container.getDataModel().getColumns(), columnName);
+      }
+    }
```
manerow: When `dataModel` is null, validation is silently skipped and any column names in `tableData.columns` are accepted. The sample data gets stored but there are no entity columns to match against, so PII masking and the UI renderer have nothing to join on.

Container's `dataModel` is optional in the schema (unlike Table's columns), so this case is reachable. In practice the auto-classification bot always populates `dataModel` before calling this, so the hole is only exposed to direct API callers. But since the feature explicitly targets structured containers, I'd rather reject here than accept bad data:

```java
if (container.getDataModel() == null || container.getDataModel().getColumns() == null) {
  throw new IllegalArgumentException(
      "Container must have a dataModel with columns defined before adding sample data");
}
for (String columnName : tableData.getColumns()) {
  validateColumn(container.getDataModel().getColumns(), columnName);
}
```

This matches the implicit contract `TableRepository.addSampleData` gets for free (Table always has columns).

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/storage/adlsConnection.json
```diff
@@ -41,6 +41,10 @@
     "supportsMetadataExtraction": {
       "title": "Supports Metadata Extraction",
       "$ref": "../connectionBasicType.json#/definitions/supportsMetadataExtraction"
+    },
+    "supportsProfiler": {
+      "title": "Supports Profiler",
+      "$ref": "../connectionBasicType.json#/definitions/supportsProfiler"
     }
```
manerow, line 48: Flagging this in case it's missed (also called out in [my comment on `CreateIngestionPipelineImpl`](https://github.com/open-metadata/OpenMetadata/pull/26495#discussion_r3078981316)). This claims ADLS supports the profiler / auto-classification pipeline, but the python samplers shipped in this PR cover only S3 and GCS (`ingestion/src/metadata/ingestion/source/storage/{s3,gcs}/service_spec.py`).

With this flag set, the UI will let a user configure an auto-classification pipeline for an ADLS service which then fails at pipeline execution time with a "no sampler registered" style error. Same applies to `customStorageConnection.json`.

Suggestion is to drop the `supportsProfiler` addition from `adlsConnection.json` and `customStorageConnection.json` for now, and bring them back in the follow-up PRs that ship their samplers.

ingestion/src/metadata/ingestion/ometa/mixins/patch_mixin.py
```diff
@@ -447,9 +449,73 @@ def patch_owner(
...
+        container: Container,
+        instance: Container,
+        column_tags: List[ColumnTag],
+        operation: PatchOperation,
+    ) -> Optional[Container]:
+        """Prepare Container destination with updated column tags"""
+        if container.dataModel is None or instance.dataModel is None:
+            logger.debug(
+                f"Container {container.fullyQualifiedName.root} has no dataModel, skipping column tag patch"
+            )
```
IceS2: Shouldn't this be a Warning?

ingestion/src/metadata/sampler/storage/s3/sampler.py
```diff
@@ -0,0 +1,103 @@
...
+                return None
+
+            candidate_keys = [
+                entry["Key"]
+                for entry in response[S3_CLIENT_ROOT_RESPONSE]
+                if entry
+                and entry.get("Key")
+                and not entry.get("Key").endswith("/")
+                and "/_delta_log/" not in entry.get("Key")
+                and not entry.get("Key").endswith("/_SUCCESS")
```
IceS2: Could potentially be extracted

IceS2 (dismissed):

openmetadata-spec/src/main/resources/json/schema/entity/services/connections/storage/adlsConnection.json
edg956: This is ok because the UI does not let users create ADLS services in OM. This is Collate-specific

IceS2 (dismissed):

manerow (dismissed):

IceS2 (dismissed):

PR #26541, opened 2026-03-17
fix(playwright): rewrite container nested fields test to use self-contained nested column setup

harsh-vador:
## Problem

The `expand / collapse should not appear after updating nested fields for container` test was failing on AUT for multiple reasons:

1. **Hard-coded sample data FQN** — the test navigated to `s3_storage_sample.departments.finance`, an entity that only exists when sample data is seeded. On AUT without sample data, the page returned "not found", causing a 180s timeout.

2. **Wrong column structure** — after switching to the `beforeAll` container, the columns (`merchant`, `columbia`, `delivery`) were all flat leaf nodes with no children. The test was checking expand/collapse behavior on non-nested fields, which trivially passes but tests nothing meaningful.

## Fix

- Removed `@sample-data` tag and sample data FQN dependency entirely
- Creates a self-contained container within the test using `performAdminLogin(browser)` with a `Struct`-typed column that has a `children` array — the only column type that renders the expand/collapse icon
- Uses `{ page }` (admin fixture) for UI interaction
- Adds `redirectToHomePage(page)` before `visitEntityPage` to load the app
- Clicks the struct row's `expand-icon` and waits for the child row to be visible before interacting with it
- Cleans up the container in a `finally` block to prevent test pollution

Screenshot 2026-03-17 at 1 22 57 PM

### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/Container.spec.ts
```diff
@@ -23,9 +22,9 @@ import {
   testCopyLinkButton,
   validateCopiedLinkFormat,
   waitForAllLoadersToDisappear,
-
 } from '../../utils/entity';
 import { test } from '../fixtures/pages';
+import { DataType } from '../../../src/generated/entity/data/container';
```
aniketkatkar97: Can we not import anything from outside the playwright folder?

aniketkatkar97: Tests should be isolated and independent of src

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/Container.spec.ts
```diff
@@ -150,38 +149,89 @@ test.describe('Container entity specific tests ', () => {
...
+            ordinalPosition: 1,
+          },
+        ],
+      },
+    ];
+
+    await nestedContainer.create(apiContext);
+    await afterSetup();
 
+    try {
```
aniketkatkar97: Why are we adding try-catch blocks inside tests? We should not be doing that

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/Container.spec.ts
harsh-vador: Its a type import then how do we plan on using [this](https://github.com/open-metadata/OpenMetadata/pull/26541/changes/BASE..94a8a717240ad1398d55ff4792726348d84a8117#diff-5a7598ba90478b2e336f02dff231c8eda96f4f35aeff8b1f08aaf8b2e91b2495R166), if you can please suggest?

aniketkatkar97: For that, you can directly use the value of the enum, right?

harsh-vador: what about the ts error which expects enum and not a string value?

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/Container.spec.ts
```diff
@@ -150,38 +149,78 @@ test.describe('Container entity specific tests ', () => {
...
-
-      await removeTagsFromChildren({
-        page,
-        tags: ['PersonalData.Personal'],
-        rowId: 's3_storage_sample.departments.finance.budget_executor',
-        entityEndpoint: 'containers',
-      });
-    }
-  );
+    await nestedContainer.create(apiContext);
```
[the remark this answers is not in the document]
harsh-vador: thats expected

PR #26668, opened 2026-03-20
Enhance Fivetran connector: destination DB log queries, ELT task phases, lineage fixes

aydingeeringh:
## Summary

Enhances the Fivetran pipeline connector with comprehensive sync history from the destination warehouse, structured ELT task phases, and improved lineage resolution.

### Problem

The Fivetran REST API `sync-history` endpoint has very limited retention — entries age out within hours, leaving OpenMetadata with incomplete or missing pipeline execution history. Additionally, the connector represented each sync as a single opaque task with no visibility into the Extract, Process, and Load phases.

### Solution

- **Destination DB log queries**: Query `fivetran_metadata.log` directly in the destination warehouse (Snowflake, Postgres, etc.) for complete sync history with per-phase timing. Uses the existing `dbServiceNames` lineage configuration to resolve the destination `DatabaseService`, creates a connection with the destination database override, and queries with dialect-aware quoted identifiers via `sqlglot`. Falls back gracefully to the REST API when the destination DB is unavailable.

- **ELT task phases**: Replace the single "sync" task with three distinct tasks — **Extract**, **Process**, **Load** — each with independent timing and status derived from LOG table events (`sync_start`, `extract_summary`, `write_to_table_start/end`, `sync_end`, `sync_stats`). The `sync_stats` durations serve as fallback when intermediate events are missing (e.g., incremental syncs with no data changes).

- **Lineage fixes**: Fix `or "*"` fallback that built FQNs with literal `"*"` as service name. Eliminate N+1 API calls for pipeline entity resolution (once per connector, not per table) and destination details (once per group, not per connector). Support messaging sources (Kafka, Confluent Cloud) with topic-to-table lineage and column-level lineage via Fivetran's schema API.

- **Client robustness**: Add null guards to `run_paginator` for None API responses, fix methods returning None instead of `{}` on failure, fix base64 encoding (`str()[2:-1]` → `.decode("ascii")`), correct type annotations.

- **Naming & display**: Display name shows source first (`postgres <> Snowflake`), task type uses "Process" (not "Transform"), `sourceUrl` only on Pipeline level.

## Changes

| File | What changed |
|------|-------------|
| `metadata.py` | Add destination DB query path with fallback, 3-task ELT model, lazy pipeline entity resolution, fix `or "*"` lineage bug, fix N+1 destination fetch |
| `client.py` | Add null guards to paginator, fix return types, fix base64 encoding, remove unused `Response` import |
| `models.py` | Fix display name order (source first), add copyright header, add return type annotations |
| `service_spec.py` | Add copyright header |
| `test_fivetran.py` | 34 tests covering DB query path (happy path, extract failure, sync_stats fallback), REST API fallback (no database, no service, exception, Databricks skip), lineage (disabled schemas/tables, cross-service, self-referencing, messaging sources, column lineage), task status building, schedule intervals, pipeline state |

## Test plan

- [x] `pytest tests/unit/topology/pipeline/test_fivetran.py` — 34/34 passing
- [x] `make py_format` — no changes
- [x] `make generate` — succeeds
- [x] Live test with real Fivetran account + Snowflake destination — 100% success, correct 3-task execution history
- [x] Hard-delete + re-ingest cycle — clean data, no stale entries
- [x] Review ELT task phases in UI (Extract → Process → Load with timing)
- [x] Verify lineage edges in UI (source table → destination table with column lineage)
- [x] Verify DAG view renders correct task topology


TeddyCr:
@aydingeeringh thanks for the PR. This is a bit out of scope for OpenMetadata as we are basically including an ETL step that extract data from Fivetran and push them into a destination database. If we want to bring this data, we should structure them to be part of the Fivetran execution entity in OpenMetadata directly, so they exist in OpenMetadata alongside the entity and its tasks

ingestion/src/metadata/ingestion/api/topology_runner.py
```diff
@@ -309,6 +309,13 @@ def _run_stage_processor(
         except Exception as exc:
             logger.debug(traceback.format_exc())
             logger.error(f"Error running stage processor: {exc}")
+            self.status.failed(
```
harshach: why are we making this change?

aydingeeringh: This was flagged by the AI skills toolkit connector-review. The exception was being logged but not recorded in the pipeline status, so stage processor failures were silently swallowed. Happy to revert this from the PR and raise it as a separate one if you'd prefer to keep scope tight.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -42,14 +57,44 @@
...
 from metadata.utils import fqn
+from metadata.utils.helpers import datetime_to_ts
 from metadata.utils.logger import ingestion_logger
 
 logger = ingestion_logger()
 
+MESSAGING_CONNECTOR_TYPES = {"confluent_cloud", "kafka"}
+MAX_SYNC_RUNS = 100
+# TODO: Add Databricks support once delta table query path is implemented
+UNSUPPORTED_DESTINATION_TYPES = {"databricks"}
```
TeddyCr: why do we have this?

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,400 @@ def yield_pipeline(
...
+    def _get_schedule_interval(
+        self, pipeline_details: FivetranPipelineDetails
+    ) -> Optional[str]:
+        sync_freq = pipeline_details.source.get("sync_frequency")
+        if not sync_freq:
+            return None
+        minutes = int(sync_freq)
+        if minutes < 60:
+            return f"*/{minutes} * * * *"
+        hours = minutes // 60
```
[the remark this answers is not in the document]
TeddyCr: @aydingeeringh this seems to be a valid comment. We should definitely unit test this with happy and non happy path.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,426 @@ def yield_pipeline(
...
+    FIVETRAN_STATUS_MAP = {
+        "COMPLETED": StatusType.Successful,
+        "FAILURE_WITH_TASK": StatusType.Failed,
+        "CANCELED": StatusType.Failed,
+    }
+
+    HISTORICAL_SYNC_FIELDS = [
+        ("succeeded_at", StatusType.Successful),
+        ("failed_at", StatusType.Failed),
+    ]
```
TeddyCr: let's move these global at the top of the file

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,426 @@ def yield_pipeline(
...
+                        != dest_service_type.lower()
+                    ):
+                        continue
+                    return service
+            except Exception as exc:
+                logger.debug(f"Could not resolve service [{service_name}]: {exc}")
+        return None
+
+    @staticmethod
+    def _parse_sync_events(rows) -> dict:
```
TeddyCr, line 224: we should defintely refactor this to reduce the nest if/else and if else elseif and unit test it as well

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,426 @@ def yield_pipeline(
...
+                if data_str:
+                    try:
+                        sync["sync_stats"] = json.loads(data_str)
+                    except (json.JSONDecodeError, TypeError):
+                        pass
+
+        return syncs
+
+    @staticmethod
+    def _build_task_statuses_for_sync(sync: dict) -> List[TaskStatus]:
```
TeddyCr, line 257: same comment as here https://github.com/open-metadata/OpenMetadata/pull/26668/changes#r3060980622

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,426 @@ def yield_pipeline(
...
+            ),
+            TaskStatus(
+                name=FIVETRAN_TASK_PROCESS,
+                executionStatus=process_status,
+                startTime=_ts(extract_end),
+                endTime=_ts(write_start_min),
+            ),
+            TaskStatus(
+                name=FIVETRAN_TASK_EXTRACT,
+                executionStatus=extract_status,
```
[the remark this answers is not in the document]
TeddyCr: @aydingeeringh this is a valid comment

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,426 @@ def yield_pipeline(
...
+            dialect = service.serviceType.value.lower()
+            uses_upper = dialect == "snowflake"
+            catalog = dest_database.upper() if uses_upper else dest_database
+            schema = "FIVETRAN_METADATA" if uses_upper else "fivetran_metadata"
+            table = "LOG" if uses_upper else "log"
+            table_ref = sqlglot.table(table, db=schema, catalog=catalog).sql(
+                dialect=dialect, identify=True
+            )
+
+            query = FIVETRAN_LOG_QUERY.format(table=table_ref)
```
TeddyCr: you should be able to modify that whole block with sqlalchemy reflection. Something like that

```python
metadata = MetaData()

with engine.connect() as conn:
    metadata.reflect(
        conn,
        schema="fivetran_metadata",
        only=["log"]
    )

log_table = metadata.tables["fivetran_metadata.log"]
```

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,426 @@ def yield_pipeline(
...
+            cutoff = datetime.now(timezone.utc) - timedelta(days=LOG_RETENTION_DAYS)
+
+            with engine.connect() as conn:
+                result = conn.execute(
+                    text(query),
+                    {
+                        "connector_id": pipeline_details.connector_id,
+                        "cutoff": cutoff,
+                    },
+                )
```
TeddyCr, line 384: Then you can use the arm syntax

```
with engine.connect() as conn:
    metadata = MetaData()
    metadata.reflect(conn, schema="fivetran_metadata", only=["log"])
    log_table = metadata.tables["fivetran_metadata.log"]
    c = log_table.c

    query = (
        select(c.sync_id, c.message_event, c.message_data, c.time_stamp)
        .where(
            c.connection_id == pipeline_details.connector_id,
            c.sync_id.isnot(None),
            c.time_stamp >= cutoff,
            c.message_event.in_(FIVETRAN_MESSAGE_EVENTS),
        )
        .order_by(asc(c.time_stamp))
    )
```

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,426 @@ def yield_pipeline(
...
+                rows = result.fetchall()
```
TeddyCr: you should use 

```python
result = conn.execute(query).yield_per(100)
for row in result:
    ...
```

so we are guarding against potential OOM errors

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,426 @@ def yield_pipeline(
...
+
+            # Sort by sync_start descending and limit
+            sorted_syncs = sorted(
+                syncs.items(),
+                key=lambda x: x[1].get(
+                    "sync_start_ts", datetime.min.replace(tzinfo=timezone.utc)
+                ),
+                reverse=True,
+            )[:MAX_SYNC_RUNS]
+
```
TeddyCr: can we do this sorting on the database side, this will be much more efficient to get already sorted results vs sorting it here. 

We would be sorting and then iterating over the sorted syncs, just this ops is already O(n log k)

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
```diff
@@ -99,19 +156,426 @@ def yield_pipeline(
...
+            for status in db_statuses:
+                yield Either(right=status)
+            return
+
+        seen_timestamps: set = set()
+        for sync in self.client.get_connector_sync_history(
+            pipeline_details.connector_id
+        ):
+            try:
+                start_dt = datetime.fromisoformat(sync["start"].replace("Z", "+00:00"))
```
TeddyCr, line 467: relevant comment

TeddyCr (changes requested):
@aydingeeringh I left a few comments. Can you take a look please. I you can also review the if / else and the nested loops that would be great. We should also add a lot more test coverage

ingestion/src/metadata/ingestion/api/topology_runner.py
aydingeeringh: Reverted this change — will raise it as a separate PR if needed.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
aydingeeringh: Removed. Fivetran's Platform Connector is automatically added to every destination including Databricks, so the exclusion was incorrect. If there's a SQLAlchemy connection issue with a specific warehouse, the existing try/except fallback to the REST API handles it.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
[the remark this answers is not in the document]
aydingeeringh: Done — added unit tests for None, missing key, exact 24h, and the existing tests already cover sub-hour, hourly, non-divisible, and >24h cases.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
aydingeeringh: Done — moved `FIVETRAN_STATUS_MAP` and `HISTORICAL_SYNC_FIELDS` to module level.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
aydingeeringh: Refactored — extracted `_try_parse_json()` helper to eliminate the 3 repeated try/except blocks, and used `syncs.setdefault()`. Added unit tests for malformed JSON, multiple syncs, and write min/max tracking.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
aydingeeringh: Added unit tests for `_build_task_statuses_for_sync` covering successful sync, failed extract cascading, sync_stats fallback, and no-write-events edge case.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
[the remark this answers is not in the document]
aydingeeringh: Fixed — removed all `.value` calls. StatusType enum is now passed directly everywhere (Pydantic v2 handles coercion).

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
aydingeeringh: Done — replaced sqlglot + raw SQL with `SaMetaData.reflect()` and `select()`. This eliminates the sqlglot dependency, manual Snowflake case handling, and the raw SQL constant.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
aydingeeringh: Done — using the select() builder with reflected column references.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
aydingeeringh: Done — using `conn.execute(query).yield_per(100)` and iterating rows inside the connection context.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
aydingeeringh: The DB query returns individual log events that need to be grouped by `sync_id` in `_parse_sync_events` before we can sort by sync. So the sort must happen in Python after grouping — we can't do it at the DB level. Added a comment explaining this.

ingestion/src/metadata/ingestion/source/pipeline/fivetran/metadata.py
aydingeeringh: The `dest_service_type` parameter already filters by service type to avoid connecting to the wrong service. No additional change needed here.

IceS2:
@aydingeeringh , closing this as it is stale and many improvements here were implemented in other PRs, please feel free to open a new one with anything missing

PR #26693, opened 2026-03-23
fix: normalize usageSummary.date format and add explicit ES mapping

### LINKED ISSUE #26678: [Search-Issue] Issue in ES/OS
- [x] Image

- [x] 	WARN [2026-03-20 04:31:23,649] [I/O dispatcher 2] o.o.s.a.b.s.OpenSearchBulkSink - Failed to index document 4bab914b-72a9-43c3-9977-848c88b25c47: mapper [query.changeDescription.fieldsUpdated.newValue] cannot be changed from type [text] to [boolean]

- [x] Got reported that the reindex was stopped with Failed step - suspecting server restarts inbetween need to check

- [ ] Vector Embeddings running into request entity too large issue

kojikokojiko:
## Problem

Elasticsearch/OpenSearch throws parse errors when reindexing entities with `usageSummary`:

```
failed to parse field [usageSummary.date] of type [date] in document with id '...'
Preview of field's value: '2026-03-19 00:00:00'
```

**Root cause (two issues):**

1. `usageSummary.date` was not defined in ES index mappings, so ES used dynamic mapping and inferred the type as `date` with default format (`strict_date_optional_time`). This format only accepts ISO 8601 (`T`-separated), not space-separated datetime strings.

2. MySQL JDBC's `ResultSet.getString()` returns `DATE` columns as `"yyyy-MM-dd HH:mm:ss"` (with time component), which doesn't match the dynamically-inferred ES format, causing parse errors.

### 1. Java: normalize date output (root cause)

Changed `r.getString("usageDate")` to `r.getDate("usageDate").toString()` in both `UsageDetailsMapper` and `UsageDetailsWithIdMapper`.

`java.sql.Date.toString()` always returns `"yyyy-MM-dd"` regardless of JDBC driver behavior.

Affected files:
- `CollectionDAO.java` (2 mappers)
- `UsageRepository.java` (1 mapper)

### 2. ES mapping: add explicit `usageSummary.date` field (defensive fix)

Added explicit `date` field definition to `usageSummary.properties` in all index mapping JSON files across all languages (`en`, `jp`, `zh`, `ru`) for all entities that have `usageSummary` (table, dashboard, mlmodel, database, databaseSchema, container, and others).

```json
"date": {
  "type": "date",
  "format": "strict_date_optional_time||yyyy-MM-dd HH:mm:ss||epoch_millis"
}
```

This ensures that even if legacy data with `"yyyy-MM-dd HH:mm:ss"` format exists, reindexing will not fail.

## Why both fixes are needed

- Fix 1 alone: new data is safe, but reindexing legacy data with old format would still fail
- Fix 2 alone: ES accepts both formats, but root cause (inconsistent date serialization) remains
- Both together: complete fix for new and existing data

## Testing

- Verified ES mapping is applied correctly on local Docker environment
- Confirmed `"2026-03-19"` (new format) indexes successfully ✅
- Confirmed `"2026-03-19 00:00:00"` (legacy format) indexes successfully ✅ (was failing before)

Fixes #26678

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
@@ -6201,8 +6201,9 @@ public interface CollectionDAO {
             new UsageStats()
                 .withCount(r.getInt("count30"))
                 .withPercentileRank(r.getDouble("percentile30"));
+        java.sql.Date usageDate = r.getDate("usageDate");
         return new UsageDetails()
-            .withDate(r.getString("usageDate"))
+            .withDate(usageDate != null ? usageDate.toString() : null)
             .withDailyStats(dailyStats)
             .withWeeklyStats(weeklyStats)
             .withMonthlyStats(monthlyStats);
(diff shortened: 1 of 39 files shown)


kojikokojiko:
@harshach 
How can I fix maven-collate-ci  error?

mohityadav766:
this is fixed already

Issue #25532, opened 2026-01-26
Unknown custom field / xxx instance not found for new custom properties


ecerulm:
**Affected module**
Backend 

**Describe the bug**
On OM 1.10.14 on Kubernetes (with multiple pods)

After creating a custom property via the UI > Settings > Custom Properties > Databases `dialect`/ `String` 

I tried to add this custom property via REST API 
```
 curl -v -H "Proxy-Authorization: Bearer $IAP_BEARER" -H "Authorization: Bearer $JWT" -H 'Content-Type: application/json-patch+json' -X PATCH "$OM_BASE_URL/api/v1/databases/57dd8e35-4c7a-4409-a609-b4a9203a6723" -d '[{"op": "replace", "path": "/extension/dialect", "value": "xxxxx"}]'| jq .
```

this **succeeds only 25%** of the times I try (on the same entity). 

Usually I get 
```
 curl -v -H "Proxy-Authorization: Bearer $IAP_BEARER" -H "Authorization: Bearer $JWT" -H 'Content-Type: application/json-patch+json' -X PATCH "$OM_BASE_URL/api/v1/databases/57dd8e35-4c7a-4409-a609-b4a9203a6723" -d '[{"op": "replace", "path": "/extension/dialect", "value": "xxxxx"}]'| jq .
{
  "code": 404,
  "message": "dialect instance for database not found"
}
```

This happens for both PATCH and GET

```
$ curl  -H "Proxy-Authorization: Bearer $IAP_BEARER" -H "Authorization: Bearer $JWT" --data-urlencode "fields=extension" -G "$OM_BASE_URL/api/v1/databases/57dd8e35-4c7a-4409-a609-b4a9203a6723"  | jq .
{
  "code": 404,
  "message": "dialect instance for database not found"
}
```

I have tried the same setup on a local OM 1.10.14 on my laptop and there I don't see this problem, I only see it on a multipod kubernetes setup. I suspect that the failures depend on what pod the request lands on.  

**To Reproduce**

**Expected behavior**

I expect it perform the PATCH or GET correctly. 
I expect the custom properties to sync between all pods (since I suspect only one of the pods know that this new custom property exist) 

**Version:**
 - OS: whatever is defined in the official helm chart.
 - Python version: whatever is in the official helm chart
 - OpenMetadata version: 1.10.14
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

[...]

PR #26730, opened 2026-03-24
Fixes #26729: Add DB fallback and TTL refresh to TypeRegistry for multi-pod cache consistency

### LINKED ISSUE #26729: TypeRegistry in-memory cache not synchronized across pods in multi-replica deployments
**Affected module**
Backend

**Describe the bug**

In multi-pod Kubernetes deployments, `TypeRegistry` is a JVM-level singleton backed by `ConcurrentHashMap` that loads custom property definitions from the database once at startup and is never refreshed. When a custom property is created via the API, only the pod that handles the request updates its in-memory cache. Other pods remain unaware of the new property until they are restarted.

This causes custom property validation to fail on (N-1) of N pods. PATCH requests to set custom property values on entities succeed only ~1/N of the time (e.g., ~33% with 3 replicas), depending on which pod the load balancer routes to. The failing pods return:

```
{"code": 400, "message": "Unknown custom field <fieldName>"}
```

**Root cause:**

`TypeRegistry.java` maintains three static `ConcurrentHashMap` fields (`TYPES`, `CUSTOM_PROPERTIES`, `CUSTOM_PROPERTY_SCHEMAS`) that are populated during `initialize()` at startup and updated locally via `addType()` when a custom property is created on that pod. There is no:
- DB fallback on cache miss
- TTL-based expiry or periodic refresh
- Cross-pod cache invalidation

The validation path in `EntityRepository.validateExtension()` calls `TypeRegistry.instance().getSchema()`, which returns `null` on stale pods, triggering the error.

This contrasts with other caches in the codebase (e.g., `SettingsCache`, `SubjectCache`, `BotTokenCache`) that use Guava `LoadingCache` with TTL-based expiry and automatic DB reload on cache miss.

**To Reproduce**

1. Deploy OpenMetadata with 2+ replicas on Kubernetes
2. Create a custom property on any entity type (e.g., via `PUT /v1/metadata/types/{id}`)
3. Immediately send a PATCH request setting that custom property value on an entity (e.g., `PATCH /v1/tables/{id}` with `/extension/<propertyName>`)
4. Repeat the PATCH request multiple times — it will fail on pods that didn't handle the creation request

**Expected behavior**

Custom property creation should be reflected across all pods without requiring a restart. PATCH/GET requests involving custom properties should succeed regardless of which pod handles the request.

**Version:**
- OpenMetadata version: Confirmed on 1.6.0 through 1.12.1-SNAPSHOT (latest main). Unfixed as of current main branch.
- Deployment: Kubernetes with multiple replicas

**Additional context**

- Related issue: #25532 — "Unknown custom field / xxx instance not found for new custom properties" (reporter on 1.10.14 with ~25% success rate on 4 pods)
- Related discussion: #21865 — "Desynchronization of the 'Custom Properties' field between openmetadata nodes"
- PR #25961 addresses a single-JVM race condition but does not fix cross-pod synchronization

I have a fix ready that adds a DB fallback with TTL-based staleness detection to `TypeRegistry`, following the same `LoadingCache`-style pattern used by `SettingsCache` and `SubjectCache`. Will open a PR shortly.

hkhan925:
### Describe your changes:

Fixes #26729

In multi-replica Kubernetes deployments, `TypeRegistry` is a JVM-local singleton that loads custom property definitions once at startup and never refreshes. When a custom property is created on one pod, other pods remain stale — causing `Unknown custom field` validation errors on ~(N-1)/N of requests. This has been reported since version 1.6.0 (#25532, #21865) and remains unfixed.

**Changes to `TypeRegistry.java`:**

- **DB fallback on cache miss**: `getSchema()`, `getCustomPropertyType()`, and `getCustomPropertyConfig()` now fall back to the database when a property is not found in the local cache. The full entity type is loaded via `TypeRepository.getByName()` and all three `ConcurrentHashMap` caches are warmed atomically through the existing `addType()` method.

- **TTL-based staleness detection**: A per-entity-type timestamp tracks when data was last refreshed. After 15 minutes, the next access triggers a DB reload. This handles property deletions and config modifications propagating across pods.

- **Cache-miss debounce**: To prevent unbounded DB queries when a non-existent property is queried repeatedly, cache-miss-triggered refreshes are debounced to at most once per 30 seconds per entity type. The TTL-based staleness check (15 min) operates independently.

- **Stale entry cleanup**: On refresh, existing custom property entries for the entity type are cleared via `removeIf()` before re-adding fresh data from DB, ensuring deleted properties don't persist in the cache.

- **NPE fix**: `addType()` now uses `listOrEmpty()` for the custom properties iteration, fixing a latent NPE for Field-category types with null `customProperties`. This matches the existing guard in `validateCustomProperties()`.

This approach mirrors the established DB-fallback pattern used by other caches in the codebase (`SettingsCache`, `SubjectCache`, `BotTokenCache`, etc.) and introduces no new infrastructure or dependencies.

#
### Type of change:
- [x] Bug fix

#
### Checklist:


hkhan925:
> Code Review ⚠️ Changes requested 3 resolved / 4 findings
> 
> Adds DB fallback and TTL refresh to TypeRegistry for multi-pod cache consistency with debounce logic to prevent query storms. However, the PR description documents a 2-minute TTL while the code implements a 15-minute interval—clarify which is correct.
> ⚠️ Bug: PR description says 2-min TTL but code uses 15-min interval
> 
> ✅ 3 resolved
> 
> 🤖 Prompt for agents
> 
> ```
> Code Review: Adds DB fallback and TTL refresh to TypeRegistry for multi-pod cache consistency with debounce logic to prevent query storms. However, the PR description documents a 2-minute TTL while the code implements a 15-minute interval—clarify which is correct.
> 
> 1. ⚠️ Bug: PR description says 2-min TTL but code uses 15-min interval
>    Files: openmetadata-service/src/main/java/org/openmetadata/service/TypeRegistry.java:56
> 
>    The PR description states: "After 2 minutes (aligned with SubjectCache and BotTokenCache TTLs), the next access triggers a DB reload." However, the code at line 56 sets `REFRESH_INTERVAL_MS = 900_000` which is 15 minutes, not 2 minutes (120,000 ms). Either the description or the code is wrong. Given the original issue is about stale caches across pods, a 15-minute TTL means pods can serve stale data for up to 15 minutes after a custom property change, which may not meet user expectations.
> ```
> 
> Options
> 
> 	
> 
> 	
> 
> Was this helpful? React with 👍 / 👎 | [Gitar](https://gitar.ai)



> Adds DB fallback and TTL refresh to TypeRegistry for multi-pod cache consistency with debounce logic to prevent query storms. However, the PR description documents a 2-minute TTL while the code implements a 15-minute interval—clarify which is correct.

Updated PR description to reflect the 15 minute interval. Custom Property creation via API probably isn't so frequent that it warrants refreshing the cache every 2 minute in a multi-pod setup. With the DB fallback and the debounce logic, could probably be bumped up even higher than 15 minutes

harshach:
@hkhan925 thanks for the PR. will review cc @sonika-shah

ayush-shah:

alaponin:
Hi @hkhan925 could you please rebase to move this PR forward?

sonika-shah:
thanks @hkhan925 for the pr!
This issue is already fixed and shipped; merged to main on 13 Jul 2026, and it was cherry-picked  to 1.13.2


main/2.0 -  https://github.com/open-metadata/OpenMetadata/pull/29479
1.13 backport: https://github.com/open-metadata/OpenMetadata/pull/30044

It will be available in -
  -  1.13.2 and later (1.13.x line)
  -  2.0and main

PR #26758, opened 2026-03-25
[WIP] Change Event Based Governance Workflow

yan-3005:
SubIssue: https://github.com/open-metadata/openmetadata-collate/issues/3364
## Summary

Switches `PeriodicBatchEntityTrigger` from polling all entities on every run to an **incremental, change-event-driven model**. Instead of scanning the full entity table, each workflow run reads only from the `change_event` table, starting from a persisted offset cursor, and advances that cursor batch-by-batch until it reaches the head of the log.

---

### New: `FetchChangeEventsImpl`
Replaces `FetchEntitiesImpl` as the data-fetching task inside `PeriodicBatchEntityTrigger`.

- Reads `change_event` rows with `(entityType, offset > lastProcessedOffset)` using the new composite index
- Deduplicates by FQN within a batch — multiple change events for the same entity produce one workflow invocation at the highest offset
- Applies the configured Elasticsearch search filter to further narrow which entities are emitted
- Tracks `currentBatchOffset` and `maxProcessedOffset` as Flowable execution variables across iterations
- **Purge recovery**: if the stored offset has been purged from the table, automatically resumes from the earliest available offset rather than crashing
- All DAO calls are wrapped in Resilience4j `Retry` (using `Workflow.TASK_RETRY_CONFIG`)

### New: `CommitChangeEventOffsetImpl`
Runs once when the batch loop terminates (`hasFinished = true`). Persists `maxProcessedOffset` back to `event_subscription_consumers` under the consumer key `{workflowFQN}Trigger-{entityType}`, so the next scheduled run resumes from where this one left off.
- Skips the write if the stored offset is already ≥ the processed offset (safe for concurrent/retried runs)

### Updated: `PeriodicBatchEntityTrigger`
New BPMN process flow per entity type:
```
Start → FetchChangeEvents ──(hasFinished)──► CommitOffset → End
               ▲                                           
               └──────────(not finished, loop back)───────
               ▲◄──── WorkflowTrigger (multi-instance) ───┘
```
- Accepts `workflowFqn` and passes it to both tasks for consumer ID construction
- `commitOffsetTask` is wired in as the terminal step after the loop exits

### DB Migrations (MySQL + PostgreSQL)
- **New index**: `idx_change_event_entity_type_offset ON change_event (entityType, offset)` — makes the range query per entity type efficient
- **Widened column**: `change_event_consumers.id` from `VARCHAR(36)` → `VARCHAR(500)`. Consumer IDs follow the pattern `{workflowFQN}Trigger-{entityType}` and can exceed UUID length. `VARCHAR(500)` stays within MySQL's 3072-byte composite key limit (`500×4 + 256×4 = 3024 bytes` with utf8mb4)

---

## Type of change
- [x] New feature
- [x] Improvement

## Checklist

### DIFF
--- a/bootstrap/sql/migrations/native/1.13.0/mysql/schemaChanges.sql
+++ b/bootstrap/sql/migrations/native/1.13.0/mysql/schemaChanges.sql
@@ -89,6 +89,8 @@ WHERE ue.name = 'mcpapplicationbot'
 -- change-event-driven workflow processing (filters by entityType + offset range).
 CREATE INDEX idx_change_event_entity_type_offset ON change_event (entityType, `offset`);
 
--- Widen change_event_consumers.id from VARCHAR(36) to VARCHAR(768) to support workflow consumer IDs
+-- Widen change_event_consumers.id from VARCHAR(36) to VARCHAR(500) to support workflow consumer IDs
 -- which follow the pattern {workflowFQN}Trigger-{entityType} and can exceed 36 characters.
-ALTER TABLE change_event_consumers MODIFY COLUMN id VARCHAR(768) NOT NULL;
+-- VARCHAR(500) keeps the composite UNIQUE(id, extension) key within MySQL's 3072-byte limit
+-- (500 * 4 + 256 * 4 = 3024 bytes with utf8mb4).
+ALTER TABLE change_event_consumers MODIFY COLUMN id VARCHAR(500) NOT NULL;
--- a/bootstrap/sql/migrations/native/1.13.0/postgres/schemaChanges.sql
+++ b/bootstrap/sql/migrations/native/1.13.0/postgres/schemaChanges.sql
@@ -101,6 +101,6 @@ ON CONFLICT DO NOTHING;
 -- change-event-driven workflow processing (filters by entityType + offset range).
 CREATE INDEX IF NOT EXISTS idx_change_event_entity_type_offset ON change_event (entitytype, "offset");
 
--- Widen change_event_consumers.id from VARCHAR(36) to VARCHAR(768) to support workflow consumer IDs
+-- Widen change_event_consumers.id from VARCHAR(36) to VARCHAR(500) to support workflow consumer IDs
 -- which follow the pattern {workflowFQN}Trigger-{entityType} and can exceed 36 characters.
-ALTER TABLE change_event_consumers ALTER COLUMN id TYPE VARCHAR(768);
+ALTER TABLE change_event_consumers ALTER COLUMN id TYPE VARCHAR(500);


yan-3005:
@gitar-bot Thanks for the review! Addressing each finding:

**Finding #3 — `resolveEntityType` ambiguous key (✅ Fixed)**
Removed the process-key split entirely. `FetchChangeEventsImpl` now always uses `entityTypesExpr.getValue(execution)` directly, which is the `FieldExtension` already injected with the correct entity type by `PeriodicBatchEntityTrigger`. Simpler and robust.

---

**Findings #1, #2, #4, #5, #6, #7 — Not related to this PR**

These findings are about files **not changed by this PR**:

| Finding | File | Relationship |
|---------|------|-------------|
| ⚠️ CheckEntityAttributesImpl lost retry | `CheckEntityAttributesImpl` | Lives in base branch `ram/workflow-improvements`, not touched here |
| ⚠️ `getEntitiesByLinks` assumes uniform entity type | `Entity.java` | Also base branch, not in this PR's diff |
| 💡 `applyCertification` queries DB on create | `EntityRepository.java` | Unrelated to this PR |
| 💡 Batch fetch failures silent in `SinkTaskDelegate` | `SinkTaskDelegate.java` | Base branch, not in this PR's diff |
| 💡 `entityIconMapping` drops `ML_MODEL_SERVICE` | UI component | Completely unrelated to this PR |
| 💡 `AppRunLogAppender.escapeJson` missing `\b`/`\f` | `AppRunLogAppender` | Completely unrelated to this PR |

These should be tracked against whichever PR introduced those changes.

yan-3005:
@gitar-bot All 6 open findings are about code introduced in the **base branch** (`ram/workflow-improvements`), not in the changes specific to this PR (`ram/workflow-improvements-change-event-offset`).

This PR's diff against its base only touches:
- `FetchChangeEventsImpl.java` + `CommitChangeEventOffsetImpl.java` (new change-event offset logic)
- `PeriodicBatchEntityTrigger.java` (wiring for the above)
- `schemaChanges.sql` (MySQL + Postgres migration for the new index and widened column)

None of the flagged files (`CheckEntityAttributesImpl`, `Entity.java`, `EntityRepository.java`, `SinkTaskDelegate.java`, the UI `entityIconMapping` component, or `AppRunLogAppender`) are changed in this PR. They're all part of `ram/workflow-improvements` which is the parent branch. These concerns should be tracked against that PR.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/impl/FetchEntitiesImpl.java
```diff
@@ -21,6 +21,12 @@
...
 
+/**
+ * @deprecated All periodic batch workflows now use {@link FetchChangeEventsImpl} for incremental
+ *     change-event-driven processing. This class is no longer used by {@link
+ *     org.openmetadata.service.governance.workflows.elements.triggers.PeriodicBatchEntityTrigger}.
+ */
+@Deprecated
 @Slf4j
 public class FetchEntitiesImpl implements JavaDelegate {
   // Entity types whose fullyQualifiedName is mapped as "text" (not "keyword")
```
IceS2, line 32: Should we just remove this?

PR #26759, opened 2026-03-25
feat(incident): add IncidentTcrsSyncHandler for Task→TCRS sync

IceS2:
## Summary

- Adds `aboutEntityLink` field to Task schema (EntityLink format encoding testCase FQN + incident stateId), backed by generated DB columns and index in MySQL/PostgreSQL
- Implements `IncidentTcrsSyncHandler` — an async EntityLifecycleEventHandler that syncs Task lifecycle events to TCRS records (New on creation, Ack on InProgress, Assigned on assignee change, Resolved on terminal status)
- Adds TCRS guard in `openOrAssignTask()` to prevent duplicate Task creation when a workflow-managed Task already exists
- `SetupImpl` builds `aboutEntityLink` for IncidentResolution/TestCaseResolution task types
- Adds `testCaseStatus` to `WorkflowTriggerFields` enum (matches the actual field name in entity changeDescription)

## Test plan

- [x] `IncidentTcrsSyncHandlerTest` — 10 unit tests covering `isIncidentTask()` boundary conditions, handler properties, and EntityLink parsing
- [x] `ManualTaskOutboxIT` — E2E integration test verifying full pipeline: workflow trigger → Task creation with aboutEntityLink → TCRS(Ack) on InProgress → TCRS(Resolved) on Completed → workflow FINISHED

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/events/lifecycle/handlers/IncidentTcrsSyncHandler.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/events/lifecycle/handlers/IncidentTcrsSyncHandler.java
@@ -60,7 +60,7 @@ public class IncidentTcrsSyncHandler implements EntityLifecycleEventHandler {
     String testCaseFqn = link.getEntityFQN();
     UUID stateId = UUID.fromString(link.getArrayFieldName());
 
-    if (tcrRecordExists(stateId)) {
+    if (incidentResolutionStatusExists(stateId)) {
       LOG.debug("[{}] TCRS(New) already exists for stateId={}, skipping.", HANDLER_NAME, stateId);
       return;
     }
@@ -76,7 +76,7 @@ public class IncidentTcrsSyncHandler implements EntityLifecycleEventHandler {
             .withUpdatedBy(updatedBy)
             .withUpdatedAt(System.currentTimeMillis());
 
-    insertTcrsRecord(testCaseFqn, tcrs);
+    insertIncidentResolutionStatus(testCaseFqn, tcrs);
     LOG.info(
         "[{}] Synced Task created -> TCRS(New) for testCase='{}', stateId='{}'",
         HANDLER_NAME,
@@ -95,7 +95,8 @@ public class IncidentTcrsSyncHandler implements EntityLifecycleEventHandler {
     String testCaseFqn = link.getEntityFQN();
     UUID stateId = UUID.fromString(link.getArrayFieldName());
 
-    TestCaseResolutionStatusTypes tcrsType = mapTaskChangeToTcrsType(task, changeDescription);
+    TestCaseResolutionStatusTypes tcrsType =
+        mapTaskChangeToResolutionStatusType(task, changeDescription);
     if (tcrsType == null) {
       return;
     }
@@ -119,7 +120,7 @@ public class IncidentTcrsSyncHandler implements EntityLifecycleEventHandler {
       tcrs.withTestCaseResolutionStatusDetails(resolved);
     }
 
-    insertTcrsRecord(testCaseFqn, tcrs);
+    insertIncidentResolutionStatus(testCaseFqn, tcrs);
     LOG.info(
         "[{}] Synced Task update -> TCRS({}) for testCase='{}', stateId='{}'",
         HANDLER_NAME,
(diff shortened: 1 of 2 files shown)


bootstrap/sql/migrations/native/2.0.0/mysql/schemaChanges.sql
```diff
@@ -30,6 +30,11 @@ CREATE TABLE IF NOT EXISTS task_entity (
     KEY idx_status_about (status, aboutFqnHash)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
 
+-- aboutEntityLink: hierarchical entity identity for lifecycle handler lookups
+ALTER TABLE task_entity ADD COLUMN IF NOT EXISTS aboutEntityLink varchar(1024)
+  GENERATED ALWAYS AS (json_unquote(json_extract(`json`, _utf8mb4'$.aboutEntityLink'))) STORED;
```
yan-3005, line 35: is mysql not utf8 by default?

openmetadata-service/src/main/java/org/openmetadata/service/events/lifecycle/handlers/IncidentTcrsSyncHandler.java
```diff
@@ -0,0 +1,235 @@
...
+  public void onEntityCreated(EntityInterface entity, SubjectContext subjectContext) {
+    if (!(entity instanceof Task task) || !isIncidentTask(task)) {
+      return;
+    }
+
+    MessageParser.EntityLink link = MessageParser.EntityLink.parse(task.getAboutEntityLink());
+    String testCaseFqn = link.getEntityFQN();
+    UUID stateId = UUID.fromString(link.getArrayFieldName());
+
+    if (tcrRecordExists(stateId)) {
```
yan-3005: nit: can we name the functions in detail, I had to recall what tcr is for 1-2 mins which is not helping in debugging

openmetadata-service/src/main/java/org/openmetadata/service/events/lifecycle/handlers/IncidentTcrsSyncHandler.java
```diff
@@ -0,0 +1,235 @@
...
+
+    if (tcrsType == TestCaseResolutionStatusTypes.Resolved) {
+      Resolved resolved =
+          new Resolved()
+              .withResolvedBy(updatedBy)
+              .withTestCaseFailureComment("Resolved via governance workflow task");
+      tcrs.withTestCaseResolutionStatusDetails(resolved);
+    }
+
+    insertTcrsRecord(testCaseFqn, tcrs);
```
yan-3005: Thinking 🤔 , In this case, should we not update the record instead of insert, or am I mistaken?

bootstrap/sql/migrations/native/2.0.0/mysql/schemaChanges.sql
IceS2: I think it should be, but it does not hurt to be defensive, right?

openmetadata-service/src/main/java/org/openmetadata/service/events/lifecycle/handlers/IncidentTcrsSyncHandler.java
IceS2: It is a time series, we never update here, insert only

openmetadata-service/src/main/java/org/openmetadata/service/events/lifecycle/handlers/IncidentTcrsSyncHandler.java
IceS2: This is fair, will do it!

bootstrap/sql/migrations/native/2.0.0/mysql/schemaChanges.sql
yan-3005: In sql files okay, in run time for other cases when we know the value is so and so, it's better to code that way rather than being defensive, makes debugging a pain. As this is in sql file, it's okay!

openmetadata-service/src/main/java/org/openmetadata/service/events/lifecycle/handlers/IncidentTcrsSyncHandler.java
yan-3005: Got it!

PR #26762, opened 2026-03-25
FIX - Clean OpenMetadataWorkflowConfig in TestConnection Workflow

yan-3005:
SM is already encrypting responses. This adds an extra layer of security by not returning unnecessary workflow info (openMetadataServerConnection with the ingestion-bot JWT token) in API responses.

The token is only passed Server -> Pipeline Service Client during trigger, never Server -> user.

### Describe your changes:

Fixes <issue-number>

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


pmbrull (dismissed):

yan-3005:
@gitar Thanks for the review! However, `deleteByIdAsync` does **not** return the `Workflow` entity — it returns `Response.accepted()` wrapping a `DeleteEntityResponse`, which contains only `jobId`, a status message, the entity name, and the delete flags. The `Workflow` object (with `openMetadataServerConnection`) is never part of that response, so there is no JWT exposure risk on this endpoint.

PR #26767, opened 2026-03-25
Fixed Entity spec

Rohit0301:
### Describe your changes:
There is one API which has higher latency, and because of which the loader wait is getting timeout after 30s, so i have added await for that particular api with 90s timeout.

Fixes <issue-number>

I worked on ... because ...

Screenshot 2026-03-25 at 6 46 20 PM

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Playwright test improvements:**
  - Added await for columns profile API response in `customProperty.ts` with 90s timeout to prevent loader timeouts
  - Added await for columns profile API response in `entity.ts` `openColumnDetailPanel` function with 90s timeout to handle high-latency API calls


openmetadata-ui/src/main/resources/ui/playwright/utils/customProperty.ts
```diff
@@ -1147,8 +1147,24 @@ export const verifyTableColumnCustomPropertyPersistence = async ({
 }) => {
   const testValue = getPropertyValues(propertyType, users).value;
 
+  const tableFqn = columnFqn.split('.').slice(0, -1).join('.');
```
ShaileshParmar11: why do we need this kind of patch work? dont we have direct value?

Rohit0301: fixed

PR #26771, opened 2026-03-25
Adapt search pipeline to prefer semantic results

lautel:
- Remove dataAssetEmbeddings alias from entities with no embeddings: tag, container, file, worksheet, spreadsheet, directory.
- Change default search pipeline for hybrid search: k=30, weights=[0.4, 0.6] --> semantic results have more weight


pmbrull (approved):

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/OpenSearchVectorService.java
```diff
@@ -79,7 +79,7 @@ public void ensureHybridSearchPipeline(double keywordWeight, double semanticWeig
         MAPPER
             .createObjectNode()
             .put("technique", "rrf")
-            .put("rank_constant", 60)
+            .put("rank_constant", 30)
```
tomasmontielp, line 83: motivation behind reducing denominator constant? overall curious about this variable, 60 as default also intriguing

lautel: 60 is the default rank constant. Everywhere you see RRF, the default is 60. Thing is, it makes scores quite uniform and masks very high ranking items. I halved it to maintain clearer differences in scores between rank positions.

PR #26793, opened 2026-03-26
Fixed the span tag issue on breadcrumb for metric entity

Rohit0301:
### Describe your changes:

Fixes [3339](https://github.com/open-metadata/openmetadata-collate/issues/3339)

I worked on ... because ...

Screenshot 2026-03-26 at 3 51 47 PM
Screenshot 2026-03-26 at 3 53 57 PM

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/utils/EntityUtils.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/EntityUtils.tsx
@@ -2671,7 +2671,7 @@ export const getEntityBreadcrumbs = (
           url: ROUTES.METRICS,
         },
         {
-          name: getEntityName(entity),
+          name: entity.name,
           url: '',
         },
       ];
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/src/utils/EntityUtils.tsx
```diff
@@ -2671,7 +2671,7 @@ export const getEntityBreadcrumbs = (
           url: ROUTES.METRICS,
         },
         {
-          name: getEntityName(entity),
+          name: entity.name,
```
chirag-madlani, line 2674: Why we need to add name explicit?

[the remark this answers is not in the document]
Rohit0301: Using getEntityName(entity) is unsafe here because highlightEntityNameAndDescription mutates source.displayName to contain raw HTML  before the source is passed to ExploreSearchCard. The entity header handles this correctly via stringToHTML, but TitleBreadcrumb renders link.name as plain text — so the HTML tags show up literally in the breadcrumb instead of being interpreted. 

And other entities like database, schema, etc all use entity.name for the entity itself

Rohit0301: Using getEntityName(entity) is unsafe here because highlightEntityNameAndDescription mutates source.displayName to contain raw HTML  before the source is passed to ExploreSearchCard. The entity header handles this correctly via stringToHTML, but TitleBreadcrumb renders link.name as plain text — so the HTML tags show up literally in the breadcrumb instead of being interpreted. 

And other entities like database, schema, etc all use entity.name for the entity itself

PR #26795, opened 2026-03-26
fix flaky search index application

shrabantipaul-collate:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Test refactoring:**
  - Extracted `installSearchIndexApplication` function to reduce duplication and enable recovery from failed retries
  - Added self-healing guard that reinstalls app via API if previous retry left it uninstalled
- **Test robustness:**
  - Added API-based app check and reinstallation to prevent flaky test failures from retry state

### DIFF
--- a/ingestion/src/metadata/ingestion/ometa/mixins/table_mixin.py
+++ b/ingestion/src/metadata/ingestion/ometa/mixins/table_mixin.py
@@ -14,8 +14,13 @@ Mixin class containing Table specific methods
 To be used by OpenMetadata class
 """
 import base64
+import datetime
+import decimal
+import ipaddress
 import json
+import math
 import traceback
+import uuid
 from typing import Dict, List, Optional, Type, TypeVar
 
 from pydantic import BaseModel, validate_call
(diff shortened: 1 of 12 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SearchIndexApplication.spec.ts
```diff
@@ -24,6 +24,76 @@ import { settingClick } from '../../utils/sidebar';
...
+      .locator('[data-testid="search-indexing-application-card"]')
+      .isVisible();
+  }
+
+  await page.click(
+    '[data-testid="search-indexing-application-card"] [data-testid="config-btn"]'
+  );
+  await page.click('[data-testid="install-application"]');
+  await page.click('[data-testid="save-button"]');
+  await page.click('[data-testid="submit-btn"]');
```
ShaileshParmar11: dont we have any api call in between? this could lead in to flakiness if required element is not loaded, may we we should add some expect block before clicking button to make sure the element is present

ShaileshParmar11: also lets use page.getById.click

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SearchIndexApplication.spec.ts
```diff
@@ -78,6 +148,16 @@ test.describe('Search Index Application', PLAYWRIGHT_BASIC_TEST_TAG_OBJ, () => {
...
+
+      // Self-heal: if a previous retry left the app uninstalled, reinstall it before proceeding.
+      const isInstalled = await page
+        .locator('[data-testid="search-indexing-application-card"]')
+        .isVisible();
+
+      if (!isInstalled) {
+        await installSearchIndexApplication(page);
+        await settingClick(page, GlobalSettingOptions.APPLICATIONS);
+      }
```
aniketkatkar97: I think instead of relying on UI, we should do these steps through API.
1. Here we are directly checking the application card. If there are many apps installed, the application card might not be visible on the first page.
2. Application installation through API will be faster
3. Please also check if test pass by creating the previous run failed test scenario. I guess we can do that locally by just uninstalling the application and then running the test.

shrabantipaul-collate: Fixed.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SearchIndexApplication.spec.ts
shrabantipaul-collate: Fixed

PR #26918, opened 2026-04-01
Feat(Workflows): Auto-inject inclusive gateways for batch routing

yan-3005:
### Describe your changes:

Phase 2 of the batch workflow improvements. Builds on Phase 1 (PR #26715) which made all task nodes batch-aware with entity lists.

**Problem:** Check nodes previously emitted a single `RESULT_VARIABLE` boolean, which drove exclusive gateway routing — only one branch (true OR false) could fire. In batch mode, a single check node can produce both passing and failing entities simultaneously, so both branches need to fire.

**Solution:** Auto-inject Flowable inclusive gateways at BPMN build time based on graph structure — no changes needed to workflow JSON definitions.

**Changes:**

- **`InclusiveGatewayBuilder`** (new) — mirrors `ExclusiveGatewayBuilder` for `InclusiveGateway` elements
- **`NodeInterface.getOutputPorts()`** (new default method) — nodes declare which output ports they produce. Only nodes overriding this with ≥2 ports get inclusive gateways injected. `UserApprovalTask` and other single-result nodes return the default empty set and are unaffected.
- **`CheckEntityAttributesTask`, `CheckChangeDescriptionTask`** — override `getOutputPorts()` returning `{"true", "false"}`
- **`DataCompletenessTask`** — overrides `getOutputPorts()` returning the configured quality band names (e.g. `{"gold", "silver"}`)
- **Check node impls** — removed `RESULT_VARIABLE`; now emit `hasTrueEntities` + `hasFalseEntities` boolean flags (check nodes), or `has_<band>_entities` per band (DataCompleteness)
- **`MainWorkflow`** — detects split/join points at build time and injects inclusive gateways automatically:
  - **Split gateway** after any node with `getOutputPorts().size() ≥ 2` and ≥2 conditional outgoing edges; conditions map to `${nodeName_hasTrueEntities}`, `${nodeName_hasFalseEntities}`, `${nodeName_has_<band>_entities}`
  - **Join gateway** before any node with ≥2 incoming edges tracing back to a common split ancestor (BFS upward)

**Before (exclusive — only one branch fires):**
```
CheckNode → [${result=='true'}] → PassBranch
          → [${result=='false'}] → FailBranch
```

**After (inclusive — both branches fire when batch has mixed results):**
```
CheckNode → InclusiveGateway(split) → [${check_hasTrueEntities}] → PassBranch
                                    → [${check_hasFalseEntities}] → FailBranch
```

- **`WorkflowEventConsumer.java`**, **`SetEntityAttributeImpl.java`**, **`EntityRepository.java`** — soft-deleted entity guard: entity lookups switched to `Include.NON_DELETED`; `EntityRepository.bulkUpdateEntitiesForGovernanceWorkflow` gained a `skipDeleted` parameter set to `true` for all governance callers; prevents governance workflows from inadvertently resurrecting glossary terms (or any entity) deleted while a workflow batch is in flight

### Type of change:
- [x] New feature

### Checklist:

----
## Summary by Gitar

- **Governance workflow schema:**
  - Renamed `bandFlagVariable` to `branchFlagVariable` in `DataCompletenessImpl` and `MainWorkflow` to align with consistent schema naming.


openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/impl/CheckEntityAttributesImpl.java
```diff
@@ -69,10 +70,10 @@ public void execute(DelegateExecution execution) {
         }
       }
 
-      boolean result = !trueEntityList.isEmpty();
       varHandler.setNodeVariable(TRUE_ENTITY_LIST_VARIABLE, trueEntityList);
       varHandler.setNodeVariable(FALSE_ENTITY_LIST_VARIABLE, falseEntityList);
-      varHandler.setNodeVariable(RESULT_VARIABLE, result);
+      varHandler.setNodeVariable(HAS_TRUE_ENTITIES_VARIABLE, !trueEntityList.isEmpty());
+      varHandler.setNodeVariable(HAS_FALSE_ENTITIES_VARIABLE, !falseEntityList.isEmpty());
```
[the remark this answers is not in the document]
yan-3005: Fixed in 43d9935d58.

- `CheckEntityAttributesImpl` and `CheckChangeDescriptionTaskImpl`: `hasFalseEntities` is now set to `!falseEntityList.isEmpty() || entityList.isEmpty()` — an empty entity list always activates the false branch so the gateway never stalls.
- `DataCompletenessImpl`: when no entities reach any band (empty input or all failed), the lowest-score band flag is activated as a fallback.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/flowable/MainWorkflow.java
```diff
@@ -45,23 +53,95 @@ public MainWorkflow(WorkflowDefinition workflowDefinition) {
...
+    Map<String, List<EdgeDefinition>> incomingEdges = buildIncomingEdgesMap(edges);
+    Set<String> splitNodes = detectSplitNodes(outgoingEdges, nodeInstanceMap);
+    Set<String> joinNodes = detectJoinNodes(incomingEdges, splitNodes);
+    Map<String, String> splitToJoin = buildSplitToJoinMap(incomingEdges, splitNodes, joinNodes);
+
+    // Add split gateways and their outgoing conditional flows
+    for (String splitNode : splitNodes) {
+      String gatewayId = splitGatewayId(splitNode);
+      String joinNode = splitToJoin.get(splitNode);
+      String defaultFlowId = joinNode != null ? gatewayId + "_default" : null;
```
IceS2, line 89: Can you explain the default flow? not sure I understood

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/impl/DataCompletenessImpl.java
```diff
@@ -165,6 +165,10 @@ public void execute(DelegateExecution execution) {
     }
   }
 
+  public static String bandFlagVariable(String bandName) {
+    return "has_" + bandName + "_entities";
+  }
+
```
IceS2, line 171: Aren't we calling this branches in the schema? Should use the same naming

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/flowable/MainWorkflow.java
yan-3005: @IceS2 The default flow is a defensive guard against the case where all outgoing conditions on the inclusive gateway evaluate to false without it, Flowable throws a FlowableException: No outgoing sequence flow could be selected and kills the process mid-execution. It routes the token directly from the split gateway to the join gateway, ensuring the workflow always has a path forward even when no quality band matches. For example, in a DataCompleteness workflow with gold (≥90%), silver (≥70%), and bronze (≥50%) bands, if every entity in the batch scores below 50% all three conditions are false and the default flow is what saves the process from crashing by bypassing all band branches and flowing straight to the join.

IceS2 (approved):

PR #26919, opened 2026-04-01
fix(ui):  Add tooltip and fix column width in DataQuality test case table

### LINKED ISSUE #26697: Fix test case table column width and add tooltip in DQ table
Image

shah-harshit:
## Summary
Fixes #26697 

Issue:
Screenshot 2026-04-01 at 3 01 42 PM
Fix:
Screenshot 2026-04-01 at 3 02 29 PM

- Column widths defined in the DataQuality tab table were being ignored — cells expanded to fit their content instead of respecting the declared `width` value
- Added a generic `onCell` mapper at the end of the `columns` `useMemo` that applies `maxWidth` + `overflow: hidden` to every column with a `width`, enforcing widths table-wide without per-column boilerplate
- Updated the **Failed/Aborted Reason** column: replaced the raw `<p>` tag with `Typography` + `TooltipTrigger` for consistent styling; added `tw:break-all` and `tw:max-w-60` to properly wrap long error strings within the bounded cell; bumped width from 200 → 250

## Root Cause

Ant Design tables with `scroll={{ x: true }}` do not use `table-layout: fixed`, so `width` on column definitions acts as a minimum rather than a cap. The fix enforces `maxWidth` at the cell (`<td>`) level via `onCell`, which the browser always respects.

## Test plan

- [ ] Open the Data Quality tab for a table with failing test cases
- [ ] Verify the **Failed/Aborted Reason** column is capped at 250 px and long text is line-clamped (2 lines) with a tooltip on hover
- [ ] Verify all other columns with defined widths (Status, Last Run, etc.) no longer expand beyond their specified widths
- [ ] Verify horizontal scroll still works when the total column width exceeds the container

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/scripts/ui-checkstyle-changed.sh
+++ b/openmetadata-ui/src/main/resources/ui/scripts/ui-checkstyle-changed.sh
@@ -59,7 +59,12 @@ fi
 if [ "${#FILES[@]}" -eq 0 ]; then
   echo "No changed src files to process."
 else
-  yarn organize-imports:cli "${FILES[@]}"
+  # organize-imports-cli only handles TS/JS source files, not JSON
+  TS_FILES=()
+  for f in "${FILES[@]}"; do
+    [[ "$f" =~ \.(ts|tsx|js|jsx)$ ]] && TS_FILES+=("$f")
+  done
+  [ "${#TS_FILES[@]}" -gt 0 ] && yarn organize-imports:cli "${TS_FILES[@]}"
   yarn lint:base --fix "${FILES[@]}"
   yarn pretty:base --write "${FILES[@]}"
   yarn license-header-fix "${FILES[@]}"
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/src/components/Database/Profiler/DataQualityTab/DataQualityTab.tsx
```diff
@@ -480,7 +487,16 @@ const DataQualityTab: React.FC<DataQualityTabProps> = ({
...
+    return data.map((col) =>
+      col.width && !col.fixed
+        ? {
+            ...col,
+            onCell: () => ({
+              style: { maxWidth: col.width, overflow: 'hidden' },
+            }),
+          }
+        : col
+    );
```
ShaileshParmar11, line 499: Why this part is needed? can you provide some details?

shah-harshit: If you refer to the description, the specified width is not being applied to the cells, causing them to expand and distort the table layout. This change ensures that the defined width is enforced on the cells, maintaining a consistent table structure.

PR #26920, opened 2026-04-01
Enable athena table properties ingestion 

varun-lakhyani:
### Describe your changes:
Enabled Ingesting Athena table properties by linking it with custom properties of omd
Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [X] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Athena table properties ingestion:**
  - Extract table properties from AWS Athena metadata and store in thread-local context during table description retrieval
  - Automatically register Athena table properties as OpenMetadata custom properties via `get_table_extensions()` method
  - Added thread-safe deduplication to prevent re-registering the same properties across multiple table ingestions
- **Framework updates:**
  - Extended `CommonDbSourceService.get_table_extensions()` base method to support connector-specific extensions
  - Added `extension` field to allowed patch fields in `patch_request.py` and restricted from updates
- **Tests:**
  - Added unit tests for `get_table_extensions()` covering null type reference and property filtering scenarios

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/athena/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/athena/metadata.py
@@ -378,8 +378,8 @@ class AthenaSource(ExternalTableLineageMixin, CommonDbSourceService):
         registered_properties = {}
         for prop_name, prop_value in tbl_properties.items():
             with self._processed_prop_lock:
-                already_registered = prop_name in self._processed_prop
-            if not already_registered:
+                prop_already_registered = prop_name in self._processed_prop
+            if not prop_already_registered:
                 try:
                     self.metadata.create_or_update_custom_property(
                         OMetaCustomProperties(


varun-lakhyani:
Screenshot 2026-04-01 at 3 02 24 PM

varun-lakhyani:
> This is acceptable degradation, but consider clearing deferred records whose tables were in the failed batch, or at minimum, ensure the warning message at line 1199 makes it clear this may be due to a prior bulk creation failure.

@gitar-bot Warning already updated at line 1199 to say "This may be due to a prior bulk creation failure."

varun-lakhyani:
> Code Review ⚠️ Changes requested 3 resolved / 5 findings
> Enables Athena table properties ingestion with thread safety and null checks, resolving race conditions and silent registration failures. However, deferred extensions are lost when bulk API exceptions occur, and partial time-range filters are silently ignored in the list endpoint.
> 
> ⚠️ Bug: Deferred extensions lost when bulk API throws exception
> 📄 [ingestion/src/metadata/ingestion/sink/metadata_rest.py:347-357](https://github.com/open-metadata/OpenMetadata/pull/26920/files#diff-d8487a19451aaad85fae8a4566b5ad96ae38b7d6ece0eb58201d83fe21680069R347-R357) 📄 [ingestion/src/metadata/ingestion/sink/metadata_rest.py:362](https://github.com/open-metadata/OpenMetadata/pull/26920/files#diff-d8487a19451aaad85fae8a4566b5ad96ae38b7d6ece0eb58201d83fe21680069R362)
> 
> In `_flush_buffer()`, when the `bulk_create_or_update` call raises an exception (line 347), the method returns early from the `except` block (line 350) without reaching `_process_deferred_extension_records()` at line 362. The deferred records survive and will be retried at `close()` or the next successful flush — however, the tables that failed to create will never exist, so the deferred extension processing will always fail for those tables with 'Entity not found' warnings.
> 
> This is acceptable degradation, but consider clearing deferred records whose tables were in the failed batch, or at minimum, ensure the warning message at line 1199 makes it clear this may be due to a prior bulk creation failure.
> 
> 💡 Edge Case: Partial time-range filter silently ignored in list endpoint
> In `McpExecutionResource.list()`, if a caller provides only `startTs` or only `endTs` (but not both), the provided value is silently ignored and the query returns all executions without any time filter. This can be confusing — a user providing `startTs` alone would expect results filtered after that timestamp.
> 
> Consider either: (a) returning an error when exactly one of the two is provided, or (b) treating a missing bound as open-ended (e.g., `startTs` only → everything after that timestamp).
> 
> Suggested fix
> ```
> if (startTs != null && endTs != null) {
>   return repository.listWithOffset(null, filter, limitParam, startTs, endTs, false, false);
> } else if (startTs != null || endTs != null) {
>   throw new IllegalArgumentException("Both startTs and endTs must be provided for time-range filtering");
> } else {
>   return repository.listWithOffset(null, filter, limitParam, false);
> }
> ```
> 
> ✅ 3 resolved
> ✅ Edge Case: No null check on fqn.build() result before appending to patches
> > 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:357-369
> > `fqn.build()` returns `Optional[str]` and can return `None` when ES search fails or context values are missing. The result is appended directly to `_pending_table_patches` without a null check. Later in `close()`, this `None` FQN is passed to `get_by_name(fqn=None)`. While the broad `except` handler will catch the resulting error, it produces a misleading warning message. Adding a guard avoids unnecessary API calls and confusing log output.
> 
> ✅ Edge Case: Silent custom property registration failure causes all patches to fail
> > 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:154-166 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:321-335 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:366 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:379-393 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:345-359
> > If `_string_property_type_ref` fails to load in `prepare()` (line 155-157), `_register_custom_property_key()` silently returns at line 324-325 without registering any custom properties. However, `_pending_table_patches` still accumulates entries (line 366), and `close()` will attempt to patch them all. Every patch will likely fail because the custom property keys don't exist on the Table entity type, resulting in N warning logs but no clear root-cause message indicating that the property type ref was never loaded.
> > Consider either: (a) skipping the append at line 366 when `_string_property_type_ref` is None, or (b) logging a single clear warning in `close()` when `_string_property_type_ref` is None, so operators can diagnose the issue.
> 
> ✅ Bug: Race condition: _current_tbl_props shared across threads
> > 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:125 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:350 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:359 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:375 📄 ingestion/src/metadata/ingestion/source/database/database_service.py:169
> > The `databaseSchema` topology node has `threads=True` (database_service.py:169), so multiple schemas are processed concurrently via `_multithread_process_node()`. The instance variable `self._current_tbl_props` is set in `get_table_description()` (line 350/359) and read in `yield_table_extension()` (line 375). When two threads process tables from different schemas concurrently, thread A's `_current_tbl_props` can be overwritten by thread B between `get_table_description` and `yield_table_extension`, causing wrong properties to be applied to tables.
> > The same concern applies to `self._registered_custom_property_keys` being mutated concurrently via `set.add()`, though that is less severe due to CPython's GIL and the idempotent nature of registration.
> 
> 🤖 Prompt for agents
> ```
> Code Review: Enables Athena table properties ingestion with thread safety and null checks, resolving race conditions and silent registration failures. However, deferred extensions are lost when bulk API exceptions occur, and partial time-range filters are silently ignored in the list endpoint.
> 
> 1. ⚠️ Bug: Deferred extensions lost when bulk API throws exception
>    Files: ingestion/src/metadata/ingestion/sink/metadata_rest.py:347-357, ingestion/src/metadata/ingestion/sink/metadata_rest.py:362
> 
>    In `_flush_buffer()`, when the `bulk_create_or_update` call raises an exception (line 347), the method returns early from the `except` block (line 350) without reaching `_process_deferred_extension_records()` at line 362. The deferred records survive and will be retried at `close()` or the next successful flush — however, the tables that failed to create will never exist, so the deferred extension processing will always fail for those tables with 'Entity not found' warnings.
>    
>    This is acceptable degradation, but consider clearing deferred records whose tables were in the failed batch, or at minimum, ensure the warning message at line 1199 makes it clear this may be due to a prior bulk creation failure.
> 
> 2. 💡 Edge Case: Partial time-range filter silently ignored in list endpoint
> 
>    In `McpExecutionResource.list()`, if a caller provides only `startTs` or only `endTs` (but not both), the provided value is silently ignored and the query returns all executions without any time filter. This can be confusing — a user providing `startTs` alone would expect results filtered after that timestamp.
>    
>    Consider either: (a) returning an error when exactly one of the two is provided, or (b) treating a missing bound as open-ended (e.g., `startTs` only → everything after that timestamp).
> 
>    Suggested fix:
>    if (startTs != null && endTs != null) {
>      return repository.listWithOffset(null, filter, limitParam, startTs, endTs, false, false);
>    } else if (startTs != null || endTs != null) {
>      throw new IllegalArgumentException("Both startTs and endTs must be provided for time-range filtering");
>    } else {
>      return repository.listWithOffset(null, filter, limitParam, false);
>    }
> ```
> 
> Options
> Display: compact → Showing less information.
> 
> Comment with these commands to change:
> 
> Compact
> ```
> gitar display:verbose         
> ```
> 
> Was this helpful? React with 👍 / 👎 | [Gitar](https://gitar.ai)

@gitar-bot For finding 1:                                                                                                                                                                
Addressed — warning at line 1199 in metadata_rest.py now reads "This may be due to a prior bulk creation failure."                                                        
                                                                                                                                                                                
  For finding 2:                                                                                                                                                                
Out of scope — McpExecutionResource.list() is not touched in this PR. Pre-existing issue, unrelated to Athena table properties ingestion.

varun-lakhyani:
> Code Review 👍 Approved with suggestions 5 resolved / 6 findings
> Adds Athena table properties ingestion with thread-safety improvements and custom property registration error handling across five resolved issues. Note that partial time-range filters in the list endpoint are silently ignored—consider adding validation or logging to clarify filter application.
> 
> 💡 Edge Case: Partial time-range filter silently ignored in list endpoint
> In `McpExecutionResource.list()`, if a caller provides only `startTs` or only `endTs` (but not both), the provided value is silently ignored and the query returns all executions without any time filter. This can be confusing — a user providing `startTs` alone would expect results filtered after that timestamp.
> 
> Consider either: (a) returning an error when exactly one of the two is provided, or (b) treating a missing bound as open-ended (e.g., `startTs` only → everything after that timestamp).
> 
> Suggested fix
> ```
> if (startTs != null && endTs != null) {
>   return repository.listWithOffset(null, filter, limitParam, startTs, endTs, false, false);
> } else if (startTs != null || endTs != null) {
>   throw new IllegalArgumentException("Both startTs and endTs must be provided for time-range filtering");
> } else {
>   return repository.listWithOffset(null, filter, limitParam, false);
> }
> ```
> 
> ✅ 5 resolved
> ✅ Edge Case: No null check on fqn.build() result before appending to patches
> > 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:357-369
> > `fqn.build()` returns `Optional[str]` and can return `None` when ES search fails or context values are missing. The result is appended directly to `_pending_table_patches` without a null check. Later in `close()`, this `None` FQN is passed to `get_by_name(fqn=None)`. While the broad `except` handler will catch the resulting error, it produces a misleading warning message. Adding a guard avoids unnecessary API calls and confusing log output.
> 
> ✅ Edge Case: Silent custom property registration failure causes all patches to fail
> > 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:154-166 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:321-335 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:366 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:379-393 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:345-359
> > If `_string_property_type_ref` fails to load in `prepare()` (line 155-157), `_register_custom_property_key()` silently returns at line 324-325 without registering any custom properties. However, `_pending_table_patches` still accumulates entries (line 366), and `close()` will attempt to patch them all. Every patch will likely fail because the custom property keys don't exist on the Table entity type, resulting in N warning logs but no clear root-cause message indicating that the property type ref was never loaded.
> > Consider either: (a) skipping the append at line 366 when `_string_property_type_ref` is None, or (b) logging a single clear warning in `close()` when `_string_property_type_ref` is None, so operators can diagnose the issue.
> 
> ✅ Bug: Race condition: _current_tbl_props shared across threads
> > 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:125 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:350 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:359 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:375 📄 ingestion/src/metadata/ingestion/source/database/database_service.py:169
> > The `databaseSchema` topology node has `threads=True` (database_service.py:169), so multiple schemas are processed concurrently via `_multithread_process_node()`. The instance variable `self._current_tbl_props` is set in `get_table_description()` (line 350/359) and read in `yield_table_extension()` (line 375). When two threads process tables from different schemas concurrently, thread A's `_current_tbl_props` can be overwritten by thread B between `get_table_description` and `yield_table_extension`, causing wrong properties to be applied to tables.
> > The same concern applies to `self._registered_custom_property_keys` being mutated concurrently via `set.add()`, though that is less severe due to CPython's GIL and the idempotent nature of registration.
> 
> ✅ Bug: Deferred extensions lost when bulk API throws exception
> > 📄 ingestion/src/metadata/ingestion/sink/metadata_rest.py:347-357 📄 ingestion/src/metadata/ingestion/sink/metadata_rest.py:362
> > In `_flush_buffer()`, when the `bulk_create_or_update` call raises an exception (line 347), the method returns early from the `except` block (line 350) without reaching `_process_deferred_extension_records()` at line 362. The deferred records survive and will be retried at `close()` or the next successful flush — however, the tables that failed to create will never exist, so the deferred extension processing will always fail for those tables with 'Entity not found' warnings.
> > This is acceptable degradation, but consider clearing deferred records whose tables were in the failed batch, or at minimum, ensure the warning message at line 1199 makes it clear this may be due to a prior bulk creation failure.
> 
> ✅ Bug: Failed property registration is never retried and still returned
> > 📄 ingestion/src/metadata/ingestion/source/database/athena/metadata.py:378-392
> > In `get_table_extensions`, a property name is added to `_processed_prop` (line 382) **before** the `create_or_update_custom_property` API call (line 384). If that call fails (e.g., transient network error), the property is permanently marked as processed and will be skipped for all subsequent tables. Furthermore, the failed property is still included in the returned `tbl_properties` dict (line 399), meaning the table's extension will reference a custom property that was never created.
> > This creates two problems:
> > 
> > 1. A transient failure permanently prevents that property from being registered for the entire ingestion run.
> > 2. Tables will have extension values pointing to non-existent custom properties.
> 
> 🤖 Prompt for agents
> ```
> Code Review: Adds Athena table properties ingestion with thread-safety improvements and custom property registration error handling across five resolved issues. Note that partial time-range filters in the list endpoint are silently ignored—consider adding validation or logging to clarify filter application.
> 
> 1. 💡 Edge Case: Partial time-range filter silently ignored in list endpoint
> 
>    In `McpExecutionResource.list()`, if a caller provides only `startTs` or only `endTs` (but not both), the provided value is silently ignored and the query returns all executions without any time filter. This can be confusing — a user providing `startTs` alone would expect results filtered after that timestamp.
>    
>    Consider either: (a) returning an error when exactly one of the two is provided, or (b) treating a missing bound as open-ended (e.g., `startTs` only → everything after that timestamp).
> 
>    Suggested fix:
>    if (startTs != null && endTs != null) {
>      return repository.listWithOffset(null, filter, limitParam, startTs, endTs, false, false);
>    } else if (startTs != null || endTs != null) {
>      throw new IllegalArgumentException("Both startTs and endTs must be provided for time-range filtering");
>    } else {
>      return repository.listWithOffset(null, filter, limitParam, false);
>    }
> ```
> 
> Options
> Display: verbose → Showing more information.
> 
> Comment with these commands to change:
> 
> Verbose
> ```
> gitar display:compact         
> ```
> 
> Was this helpful? React with 👍 / 👎 | [Gitar](https://gitar.ai)

@gitar-bot This is a pre-existing issue in McpExecutionResource.list() unrelated to the Athena ingestion changes in this PR — out of scope

PR #26957, opened 2026-04-02
Fixes #22305: support query parameters for automated task webhooks

### LINKED ISSUE #22305: Support Query Params for Webhook
**Feature**
Webhook currently supports header value. As a user I want to be able to add query params to the url

abhayguptas:
### Describe your changes:

Fixes #22305

Added support for query parameters in the governance automated task webhook sink.

I worked on this because webhook sink configuration already supported custom headers, but there was no way to append query parameters to the target webhook URL. This change adds `queryParams` to the webhook sink schema, updates the backend provider to append configured query params to outgoing requests, updates the generated UI type, and adds backend tests to verify both config validation and request URL behavior.

### Local Testing Passed
- `mvn -pl openmetadata-service -am install -DskipTests`
- `mvn -pl openmetadata-service -Dtest=WebhookSinkProviderTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `yarn parse-schema`
- `git diff --check`

`WebhookSinkProviderTest` result:
- Tests run: 14
- Failures: 0
- Errors: 0
- Skipped: 0

#
### Type of change:
- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/sink/providers/WebhookSinkProvider.java
```diff
@@ -305,6 +306,20 @@ private Response sendRequest(String payload) {
...
   }
 
+  private WebTarget addQueryParams(WebTarget target) {
+    if (config.getQueryParams() == null
+        || config.getQueryParams().getAdditionalProperties() == null) {
+      return target;
+    }
+
+    for (Map.Entry<String, String> queryParam :
+        config.getQueryParams().getAdditionalProperties().entrySet()) {
```
TeddyCr: instead of relying on additionallProperties, you can make your queryParams in the json schema be a list of key/value, something like this

```json
"queryParams": {
  "title": "Query Parameters",
  "description": "Query parameters to append to the webhook URL.",
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "key": {
        "type": "string"
      },
      "value": {
        "type": "string"
      }
    },
    "required": ["key", "value"],
    "additionalProperties": false
  }
}
```

So you can explicitly access key and value without relying on Map conversion

abhayguptas: Thanks for the suggestion, updated accordingly. queryParams is now modeled as a list of {key, value} objects in schema, and the provider reads key/value directly without relying on additionalProperties. I also updated tests for the new shape.

PR #26968, opened 2026-04-02
Add missing MCP entity types to EntityLink grammar

pmbrull:
## Summary
- Add `mcpServer` and `mcpService` to the `ENTITY_TYPE` rule in `EntityLink.g4`
- Add `mcpExecution` to `ENTITIES_EXCLUDED_FROM_GRAMMAR` in `EntityLinkGrammarTest` (time-series entity via `EntityTimeSeriesRepository`, not independently linkable — same as `pipelineExecution`, `agentExecution`, etc.)

## Test plan
- [x] `EntityLinkGrammarTest` passes (3/3 tests: all entity types covered, no stale exclusions, no redundant exclusions)

### DIFF
--- a/.github/workflows/openmetadata-service-unit-tests.yml
+++ b/.github/workflows/openmetadata-service-unit-tests.yml
@@ -66,12 +66,6 @@ jobs:
     needs: changes
     if: ${{ needs.changes.outputs.java == 'true' }}
     steps:
-      - name: Require safe-to-test label
-        if: ${{ github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'safe to test') }}
-        run: |
-          echo "The 'safe to test' label is required before running OpenMetadata service unit tests."
-          exit 1
-
       - name: Checkout
         uses: actions/checkout@v4
         with:


.github/workflows/openmetadata-service-unit-tests.yml
```diff
@@ -66,12 +66,6 @@ jobs:
     needs: changes
     if: ${{ needs.changes.outputs.java == 'true' }}
     steps:
-      - name: Require safe-to-test label
-        if: ${{ github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'safe to test') }}
-        run: |
-          echo "The 'safe to test' label is required before running OpenMetadata service unit tests."
-          exit 1
```
pmbrull, line 73: pull_request runs in the context of the fork/PR branch. GitHub already sandboxes it:
  - No access to repo secrets (only a scoped, read-only-ish GITHUB_TOKEN)
  - A malicious PR can only run code against itself

  pull_request_target runs in the context of the base branch (main). It does have access
  to repo secrets. That's exactly why the other workflows use the safe-to-test label gate
  — it's a manual checkpoint before running untrusted code with elevated privileges.

we had no need for the check here

PR #26983, opened 2026-04-02
Feature: burstiq profiler

akashverma0786:
### Describe your changes:

Screenshot 2026-04-02 at 20 38 29

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Caching optimization:**
  - Added `_chain_metrics` cache to `BurstIQClient.get_chain_metrics()` to avoid repeated API calls
- **Data model fixes:**
  - Updated `TQLRecord.to_record()` to handle non-dict `data` values and merge with extra fields
- **Performance tuning:**
  - Increased `_PAGE_SIZE` from 1,000 to 5,000 in BurstIQ sampler
- **Test updates:**
  - Refactored pagination test to use `_PAGE_SIZE` constant for maintainability
- **Documentation:**
  - Added `BurstIQ.md` with connection requirements and configuration details

### DIFF
--- a/ingestion/src/metadata/sampler/pandas/burstiq/sampler.py
+++ b/ingestion/src/metadata/sampler/pandas/burstiq/sampler.py
@@ -85,27 +85,27 @@ class BurstIQSampler(SamplerInterface):
         sample_type = self.sample_config.profileSampleType
 
         if sample and sample_type == ProfileSampleType.ROWS:
-            total_limit = int(sample)
+            total_limit: Optional[int] = int(sample)
         elif sample and sample_type == ProfileSampleType.PERCENTAGE:
             total = self.client.get_chain_metrics().get(chain, 0)
             total_limit = max(1, int(total * sample / 100))
         else:
-            total_limit = self.client.get_chain_metrics().get(chain, 0)
-
-        if not total_limit:
-            self._cached_frames = [pd.DataFrame()]
-            return self._cached_frames
+            total_limit = None
 
         frames = []
         skip = 0
-        while skip < total_limit:
-            page_size = min(_PAGE_SIZE, total_limit - skip)
+        while True:
+            page_size = (
+                min(_PAGE_SIZE, total_limit - skip) if total_limit else _PAGE_SIZE
+            )
             records = self.client.get_records_by_tql(chain, limit=page_size, skip=skip)
             if not records:
                 break
             frames.append(self._cast_dataframe(pd.DataFrame(records)))
             skip += len(records)
             if len(records) < page_size:
+                break
+            if total_limit and skip >= total_limit:
                 break
 
         self._cached_frames = frames if frames else [pd.DataFrame()]
--- a/ingestion/tests/unit/topology/database/test_burstiq_sampler.py
+++ b/ingestion/tests/unit/topology/database/test_burstiq_sampler.py
@@ -131,39 +131,40 @@ class TestBurstIQSamplerRawDataset:
         )
         assert sum(len(df) for df in dfs) == 50
 
-    def test_no_sample_returns_all_from_chain_metrics(self, sampler, mock_client):
+    def test_no_sample_fetches_all_via_pagination(self, sampler, mock_client):
         sampler.sample_config = SampleConfig()
-        mock_client.get_chain_metrics.return_value = {"TestChain": 10}
         mock_client.get_records_by_tql.return_value = [
             {"score": float(i)} for i in range(10)
         ]
 
         dfs = list(sampler.raw_dataset())
 
-        mock_client.get_chain_metrics.assert_called_once()
+        mock_client.get_chain_metrics.assert_not_called()
         assert sum(len(df) for df in dfs) == 10
 
     def test_empty_chain_yields_single_empty_dataframe(self, sampler, mock_client):
         sampler.sample_config = SampleConfig()
-        mock_client.get_chain_metrics.return_value = {"TestChain": 0}
+        mock_client.get_records_by_tql.return_value = []
 
         dfs = list(sampler.raw_dataset())
 
+        mock_client.get_chain_metrics.assert_not_called()
         assert len(dfs) == 1
         assert dfs[0].empty
-        mock_client.get_records_by_tql.assert_not_called()
 
     def test_pagination_splits_into_multiple_pages(self, sampler, mock_client):
         sampler.sample_config = SampleConfig()
-        total = _PAGE_SIZE * 2 + _PAGE_SIZE // 2
-        mock_client.get_chain_metrics.return_value = {"TestChain": total}
         page1 = [{"score": float(i)} for i in range(_PAGE_SIZE)]
         page2 = [{"score": float(i)} for i in range(_PAGE_SIZE, _PAGE_SIZE * 2)]
-        page3 = [{"score": float(i)} for i in range(_PAGE_SIZE * 2, total)]
+        page3 = [
+            {"score": float(i)}
+            for i in range(_PAGE_SIZE * 2, _PAGE_SIZE * 2 + _PAGE_SIZE // 2)
+        ]
         mock_client.get_records_by_tql.side_effect = [page1, page2, page3]
 
         dfs = list(sampler.raw_dataset())
 
+        mock_client.get_chain_metrics.assert_not_called()
         assert len(dfs) == 3
         assert len(dfs[0]) == _PAGE_SIZE
         assert len(dfs[1]) == _PAGE_SIZE
@@ -171,13 +172,13 @@ class TestBurstIQSamplerRawDataset:
 
     def test_stops_early_on_short_page(self, sampler, mock_client):
         sampler.sample_config = SampleConfig()
-        mock_client.get_chain_metrics.return_value = {"TestChain": 2000}
         mock_client.get_records_by_tql.return_value = [
             {"score": float(i)} for i in range(5)
         ]
 
         dfs = list(sampler.raw_dataset())
 
+        mock_client.get_chain_metrics.assert_not_called()
         assert mock_client.get_records_by_tql.call_count == 1
         assert len(dfs) == 1
         assert len(dfs[0]) == 5


ingestion/src/metadata/sampler/pandas/burstiq/sampler.py
```diff
@@ -0,0 +1,176 @@
...
+    TableData,
+)
+from metadata.sampler.sampler_interface import SamplerInterface
+from metadata.utils.constants import SAMPLE_DATA_DEFAULT_COUNT
+from metadata.utils.sqa_like_column import SQALikeColumn
+
+if TYPE_CHECKING:
+    from metadata.ingestion.source.database.burstiq.client import BurstIQClient
+
+_DEFAULT_PROFILER_LIMIT = 10_000
```
TeddyCr: why do we have this here?

ingestion/src/metadata/sampler/pandas/burstiq/sampler.py
```diff
@@ -0,0 +1,176 @@
...
+    def raw_dataset(self) -> Callable[[], Iterator[pd.DataFrame]]:
+        """Lazy-fetch and cache records as a single DataFrame, return a callable iterator."""
+        if self._cached_df is None:
+            limit = self._get_limit()
+            chain = self.entity.name.root
+            records = self.client.get_records_by_tql(chain, limit=int(limit))
+            df = pd.DataFrame(records) if records else pd.DataFrame()
+            self._cached_df = self._cast_dataframe(df)
+        df = self._cached_df
+        return lambda: iter([df])
```
TeddyCr: We should not load the entire dataset (or set arbitrary limit). You should check how we are doing with `ingestion/src/metadata/readers/dataframe/parquet.py`

TeddyCr (changes requested):

ingestion/src/metadata/ingestion/source/database/burstiq/connection.py
```diff
@@ -30,18 +30,23 @@
...
+    table during profiler ingestion, where SamplerInterface.__init__ calls
+    get_ssl_connection (which calls this function) once per table entity.
+    Using id(connection) was unreliable because each table deserialization
+    produces a new object with a different id.
     """
-    return BurstIQClient(config=connection)
+    key = hash(connection.model_dump_json())
+    if key not in _CLIENT_CACHE:
+        _CLIENT_CACHE[key] = BurstIQClient(config=connection)
+    return _CLIENT_CACHE[key]
```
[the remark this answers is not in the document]
akashverma0786: Updated to hashlib.sha256

ingestion/tests/unit/observability/profiler/pandas/test_burstiq_profiler_interface.py
```diff
@@ -0,0 +1,531 @@
...
+
+    with (
+        patch(
+            "metadata.profiler.interface.profiler_interface.get_ssl_connection",
+            return_value=FakeConnection(),
+        ),
+        patch(
+            "metadata.sampler.sampler_interface.get_ssl_connection",
+            return_value=FakeConnection(),
+        ),
```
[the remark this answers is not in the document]
akashverma0786: Updated the test

ingestion/src/metadata/sampler/pandas/burstiq/sampler.py
```diff
@@ -0,0 +1,176 @@
...
+        """Return SQALikeColumn list derived from the OM Table entity."""
+        df = next(self.raw_dataset())
+        df_cols = set(df.columns)
+        om_cols = {c.name.root for c in self.entity.columns}
+        missing = om_cols - df_cols
+        extra = df_cols - om_cols
+        print(f"[DEBUG] df columns ({len(df_cols)}): {sorted(df_cols)}")
+        print(f"[DEBUG] OM columns in entity ({len(om_cols)}): {sorted(om_cols)}")
+        print(f"[DEBUG] In OM but missing from df ({len(missing)}): {sorted(missing)}")
+        print(f"[DEBUG] In df but not in OM ({len(extra)}): {sorted(extra)}")
```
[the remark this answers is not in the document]
akashverma0786: Removed redundant print statements

ingestion/src/metadata/sampler/pandas/burstiq/sampler.py
akashverma0786: Removed redundant code

ingestion/src/metadata/sampler/pandas/burstiq/sampler.py
akashverma0786: Applied new changes can you check now @TeddyCr

ingestion/src/metadata/ingestion/source/database/burstiq/client.py
```diff
@@ -311,6 +316,64 @@ def get_edges(
...
 
+    def get_chain_metrics(self) -> Dict[str, int]:
+        """
+        Fetch asset counts per chain from BurstIQ metrics endpoint.
+
+        Returns:
+            Dict mapping chain name to asset (row) count
+        """
+        logger.info("Fetching chain metrics from BurstIQ...")
+        data = self._make_request("GET", "/api/metrics/sdz")
```
TeddyCr, line 315: do you know how big that response can be? I see we don't have limit or pagination, wondering how it would work if it default to limit=10 but you have 100 elements

ingestion/src/metadata/ingestion/source/database/burstiq/client.py
```diff
@@ -311,6 +316,64 @@ def get_edges(
...
+            chain: Chain (dictionary) name to query
+            limit: Maximum number of records to fetch
+            skip: Number of records to skip (for pagination)
+
+        Returns:
+            List of flat record dicts (data envelope unwrapped)
+        """
+        if not re.fullmatch(r"[A-Za-z0-9_]+", chain):
+            logger.warning(
+                f"Skipping chain '{chain}': name contains characters not supported by TQL "
```
TeddyCr: does it mean we would never ingest chains that have tel non supported character or there is a fall back?

ingestion/src/metadata/ingestion/source/database/burstiq/client.py
```diff
@@ -311,6 +316,64 @@ def get_edges(
...
+            return []
+
+        records = []
+        for item in raw:
+            if isinstance(item, dict):
+                if "data" in item and isinstance(item["data"], dict):
+                    records.append(item["data"])
+                else:
+                    records.append(item)
+        logger.info(f"Fetched {len(records)} records for chain '{chain}'")
```
TeddyCr: that seems like a fragile chain of if/else/if. Are we able to serialize the result into a data model or something with stronger typing for example?

ingestion/src/metadata/ingestion/source/database/burstiq/connection.py
```diff
@@ -30,18 +31,24 @@
 
 logger = ingestion_logger()
 
+_CLIENT_CACHE: Dict[str, BurstIQClient] = {}
```
TeddyCr, line 34: do we have anything to limit how big the cache can grow?

ingestion/src/metadata/profiler/interface/pandas/burstiq/profiler_interface.py
```diff
@@ -0,0 +1,137 @@
...
+        We also iterate only first_df.columns (columns the API actually returned) rather than
+        self.table.columns (all 40 OM columns), which avoids KeyErrors for the columns absent
+        from the TQL response and ensures the column names used here match exactly what
+        _type_casted_dataset uses for its name-based lookups.
+        """
+        if self.dataset is None:
+            return []
+        first_df = next(self.dataset(), None)
+        if first_df is None:
+            return []
```
TeddyCr, line 69: what happens down the chain if we return an empty array at the stage?

ingestion/src/metadata/ingestion/source/database/burstiq/client.py
akashverma0786: The chainMetrics is a dict inside a single JSON object, not an array of items that gets paginated. The API doesn't return 10 chains then stop it returns one object and all chains are keys inside chainMetrics.

ingestion/src/metadata/profiler/interface/pandas/burstiq/profiler_interface.py
akashverma0786: All column profiling is skipped, that is same as parent class

ingestion/src/metadata/ingestion/source/database/burstiq/connection.py
akashverma0786: This will have at most 1 entry. The cache exists purely to avoid re-authentication on every table during profiler ingestion

ingestion/src/metadata/ingestion/source/database/burstiq/client.py
akashverma0786: Added this

ingestion/src/metadata/ingestion/source/database/burstiq/client.py
akashverma0786: This is for running a query, so this sanitation is kind of not needed; we will directly throw any error if some unknown character occurs. Since I am limited to provided read-only data , we can check this in the future.
Changing this.

ingestion/src/metadata/ingestion/source/database/burstiq/connection.py
TeddyCr: How will it have at most 1 entry?

ingestion/src/metadata/ingestion/source/database/burstiq/client.py
TeddyCr: This chain can be o f unlimited size then?

akashverma0786: the response will always be exactly as big as the number of tables the customer has I dont have a way to limit it , burstiq doesn't support pagination for this endpoint. We just have to fetch all of it at once.

ingestion/src/metadata/ingestion/source/database/burstiq/connection.py
akashverma0786: Each profiler run targets a single burstiq service with a fixed config. Since the cache key is derived from the connection config, all table-level calls share the same key, resulting in exactly 1 cache entry per profiler run.

PR #26984, opened 2026-04-02
Replace fragile .replace() chain with generic camelCase conversion in get_entity_from_create

### LINKED ISSUE #26321: Python SDK import naming error
**Affected module**
Python SDK

**Describe the bug**
I found this code snippet responsible for extracting the entity name to import for the insert/update functions, and the SDK is currently not handling some cases correctly.

In my case, when I call client.create_or_update, this code execute that through the errors since:
CreateTestCaseResult gets mapped to testCaseresult, whereas it should be testCaseResult.
CreateTableProfileRequest gets mapped to tableprofile, whereas it should be tableProfile.

This causes the import to fail.

https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/ometa/ometa_api.py
```
    def get_entity_from_create(self, create: Type[C]) -> Type[T]:
        """
        Inversely, import the Entity type based on the create Entity class
        """

        class_name = create.__name__.replace("Create", "").replace("Request", "")
        file_name = (
            class_name.lower()
            .replace("glossaryterm", "glossaryTerm")
            .replace("dashboarddatamodel", "dashboardDataModel")
            .replace("apiendpoint", "apiEndpoint")
            .replace("apicollection", "apiCollection")
            .replace("testsuite", "testSuite")
            .replace("testdefinition", "testDefinition")
            .replace("testcase", "testCase")
            .replace("searchindex", "searchIndex")
            .replace("storedprocedure", "storedProcedure")
            .replace("ingestionpipeline", "ingestionPipeline")
            .replace("dataproduct", "dataProduct")
            .replace("datacontract", "dataContract")
            .replace("chatconversation", "chatConversation")
            .replace("eventsubscription", "eventSubscription")
        )
        class_path = ".".join(
            filter(
                None,
                [
                    self.class_root,
                    self.entity_path
                    if not file_name.startswith("test")
                    and not file_name.startswith("eventSubscription")
                    else None,
                    self.get_module_path(create),
                    self.update_file_name(create, file_name),
                ],
            )
        )
        entity_class = getattr(
            __import__(class_path, globals(), locals(), [class_name]), class_name
        )
        return entity_class
```

**To Reproduce**
```
config = OpenMetadataConfig(open_metadata_host, open_metadata_token)
_ = OpenMetadata.initialize(config)
client = OpenMetadata.get_default_client()
...
request = CreateTestCaseResult.model_validate(test_case_result)
client.create_or_update(request)
```

**Expected behavior**
The import should be success. I suggest add all these missing cases to above function.

**Version:**
 - OpenMetadata version: 1.11.6
 - OpenMetadata Ingestion package version: 1.11.6.1

**Additional context**

RajdeepKushwaha5:
### Describe your changes:

Fixes #26321

The `get_entity_from_create()` method in the Python SDK converted PascalCase class names to camelCase module file names using a hardcoded chain of 16 `.replace()` calls. Any entity not explicitly listed in the chain would get an incorrect file name — for example, `TestCaseResult` would produce `testCaseresult` instead of `testCaseResult`, causing `ModuleNotFoundError`.

**What I changed:**
- Added `_to_file_name()` static method that handles PascalCase → camelCase conversion generically using regex
- Handles leading acronyms (`API` → `api`, `AI` → `ai`), standard PascalCase (`GlossaryTerm` → `glossaryTerm`), and the `MlModel` → `mlmodel` exception via an override dict
- Replaced the entire `.replace()` chain with a single `self._to_file_name(class_name)` call
- Added 42 unit test assertions covering single-word, multi-word, acronym-prefix, and override cases

**Why:**
The old chain was fragile — every new entity with a multi-word name needed a manual entry, and missing entries silently produced wrong file names.

**How I tested:**
- Ran all 42 assertions covering every existing entity name and the reported broken cases
- Verified the mapping produces identical results to the old chain for all previously-handled entities

#
### Type of change:
- [x] Bug fix

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/ometa/ometa_api.py
+++ b/ingestion/src/metadata/ingestion/ometa/ometa_api.py
@@ -458,11 +458,11 @@ class OpenMetadata(
         """
         for prefix, replacement in OpenMetadata._FILE_NAME_PREFIX_OVERRIDES.items():
             if class_name.startswith(prefix):
-                return replacement + class_name[len(prefix):]
+                return replacement + class_name[len(prefix) :]
         match = re.match(r"^([A-Z]+)([A-Z][a-z])", class_name)
         if match:
             acronym = match.group(1)
-            return acronym.lower() + class_name[len(acronym):]
+            return acronym.lower() + class_name[len(acronym) :]
         return class_name[0].lower() + class_name[1:]
 
     def get_entity_from_create(self, create: Type[C]) -> Type[T]:
--- a/ingestion/tests/unit/test_ometa_endpoints.py
+++ b/ingestion/tests/unit/test_ometa_endpoints.py
@@ -12,6 +12,7 @@
 """
 OpenMetadata high-level API endpoint test
 """
+
 from unittest import TestCase
 
 from metadata.generated.schema.api.data.createTableProfile import (


TeddyCr:
This was already addressed in a different PR

PR #26985, opened 2026-04-02
Fixes #26321: correct Python SDK create-to-entity import naming for special cases

### LINKED ISSUE #26321: Python SDK import naming error

abhayguptas:
### Describe your changes:

Fixes #26321

I updated the SDK create-to-entity resolver to correctly map special cases that were failing import resolution:
- `CreateTestCaseResult` -> `TestCaseResult` from `tests/basic.py`
- `CreateTableProfileRequest` -> `TableProfile` from `entity/data/table.py`

I also added unit assertions in `test_ometa_endpoints.py` to cover both mappings and prevent regression.

Why:
- The existing lowercase-based filename derivation breaks camel-case names like `testCaseResult` and `tableProfile`.

How I tested:
- `python3 -m py_compile ingestion/src/metadata/ingestion/ometa/ometa_api.py ingestion/tests/unit/test_ometa_endpoints.py` ✅
- Focused pytest is currently blocked in local workspace due missing runtime test dependencies (`jsonpatch`, `chardet`, `sqlparse`, `antlr4`), so full unit run is pending in proper env.

Note:
- Route support for `create_or_update(CreateTestCaseResult)` is a separate behavior change and should be confirmed with mentor/maintainers before adding in this PR.

#
### Type of change:
- [x] Bug fix

#
### Checklist:
Explanation: Not applicable (no JSON Schema changes).

### DIFF
--- a/ingestion/tests/unit/test_ometa_endpoints.py
+++ b/ingestion/tests/unit/test_ometa_endpoints.py
@@ -24,8 +24,10 @@ from metadata.generated.schema.api.services.createDatabaseService import (
 from metadata.generated.schema.api.services.ingestionPipelines.createIngestionPipeline import (
     CreateIngestionPipelineRequest,
 )
-from metadata.generated.schema.api.tests.createTestCaseResult import CreateTestCaseResult
 from metadata.generated.schema.api.teams.createUser import CreateUserRequest
+from metadata.generated.schema.api.tests.createTestCaseResult import (
+    CreateTestCaseResult,
+)
 from metadata.generated.schema.entity.data.chart import Chart
 from metadata.generated.schema.entity.data.dashboard import Dashboard
 from metadata.generated.schema.entity.data.database import Database


ingestion/tests/unit/test_ometa_endpoints.py
```diff
@@ -127,3 +132,9 @@ def test_get_entity_from_create(self):
 
         entity = self.metadata.get_entity_from_create(CreateIngestionPipelineRequest)
         assert issubclass(entity, IngestionPipeline)
+
+        entity = self.metadata.get_entity_from_create(CreateTestCaseResult)
+        assert issubclass(entity, TestCaseResult)
+
+        entity = self.metadata.get_entity_from_create(CreateTableProfileRequest)
+        assert issubclass(entity, TableProfile)
```
TeddyCr: we should test if the entity is an instance of the expected class not a subclass

```python
isinstance(entity, TableProfile)
```

abhayguptas: Thanks for the review. Updated the test to assert exact mapped class type (not subclass), and pushed the fix.

PR #26965, opened 2026-04-02
Fixes #26775: Use empty default for OIDC_CUSTOM_PARAMS to fix Compose brace-parsing

RajdeepKushwaha5:
### Describe your changes:

Fixes #26775

Docker Compose rejects `${OIDC_CUSTOM_PARAMS:-{}}` because the bare braces `{}` conflict with Compose's variable substitution syntax, causing a parse error on startup.

Changed the default from `${OIDC_CUSTOM_PARAMS:-{}}` to `${OIDC_CUSTOM_PARAMS:-}` (empty default) across all 8 Docker Compose files. This is correct because:

1. `conf/openmetadata.yaml` already uses `customParams: ${OIDC_CUSTOM_PARAMS:-}` (empty default)
2. The server null-checks `getCustomParams() != null` before using the value
3. An empty/unset value is handled gracefully — no behavioral change for users who don't set `OIDC_CUSTOM_PARAMS`

**Files changed:**
- `docker/development/docker-compose.yml`
- `docker/development/docker-compose-postgres.yml`
- `docker/development/docker-compose-gcp.yml`
- `docker/development/docker-compose-postgres-fuseki.yml`
- `docker/docker-compose-openmetadata/docker-compose-openmetadata.yml`
- `docker/docker-compose-quickstart/docker-compose.yml`
- `docker/docker-compose-quickstart/docker-compose-postgres.yml`
- `docker/docker-compose-quickstart/docker-compose-rdf.yml`

**Validation:** All compose files pass `docker compose config --quiet` (GCP file fails due to a pre-existing unrelated `GOOGLE_APPLICATION_CREDENTIALS` issue).

#
### Type of change:
- [x] Bug fix

#
### Checklist:
- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [x] My PR title is `Fixes <issue-number>: <short explanation>`
- [x] I have commented on my code, particularly in hard-to-understand areas.
- [x] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

- [x] I have added a test that covers the exact scenario we are fixing. Validated via `docker compose config --quiet` on all 8 files.

----
## Summary by Gitar

- **Kubernetes configurations:**
  - Adjusted `omjob` pod and label naming constraints to satisfy length requirements in cluster deployments.

<sub>This will update automatically on new commits.</sub>

PR #27011, opened 2026-04-03
fix(quickstart): quote OIDC default to prevent compose brace-parsing crash (#26775)

### LINKED ISSUE #26775: Issues with QuickStart
**Affected module**
QuickStart example Docker Compose, found [here](https://docs.open-metadata.org/v1.12.x/quick-start).

**Describe the bug**
I found two issues.

First, the line `OIDC_CUSTOM_PARAMETERS: ${OIDC_CUSTOM_PARAMS:-{}}` appears to be fragile because it uses Compose interpolation with the default value that itself contains curly braces. In my setup, that ended up producing invalid OpenMetaData config (`customParams: {}}`) and caused the server to fail parsing `openmetadata.yaml`. Replacing it with the literal string `OIDC_CUSTOM_PARAMS: "{}"` fixed the issue. I think the literal form is safer for the quickstart because it avoids brace-parsing ambiguity in Compose-compatible runtimes, which should improve portability not just for Podman but across Docker compose versions as well. Further, for the quickstart people will almost always be using the default value, so supporting supplying custom OIDC parameters is likely not needed for quickstart.

Second, I noticed a likely minor typo in the Quickstart documentation itself. Under "Stop and Start" and "Cleanup" the docker compose commands are provided without the `-f` flag, which is present in the earlier docker compose commands. When dealing with the postgres version, the docker compose file is not named `compose.yml`, `docker-compose.yml`, or `container-compose.yml`, thus the `-f` flag is required for docker compose commands. For example, the quickstart guide says to run `docker compose down`, but that doesn't work. What is actually required is `docker compose -f docker-compose-postgres.yml down`.

**To Reproduce**

My setup:
- MacOS Sequoia 15.5
- podman 5.6.1
- podman-compose 1.5.0
- I'm using podman instead of docker, but `docker --version` returns 5.6.1.

Following the instructions in the quickstart guide:

```
mkdir openmetadata-docker && cd openmetadata-docker
curl -sL -o docker-compose.yml https://github.com/open-metadata/OpenMetadata/releases/download/1.12.0-release/docker-compose.yml

curl -sL -o docker-compose-postgres.yml https://github.com/open-metadata/OpenMetadata/releases/download/1.12.0-release/docker-compose-postgres.yml
docker compose -f docker-compose.yml up --detach
```

It doesn't matter if MySQL or Postgres version of quickstart is used, the error will be the same. The RDBMS, elasticsearch, and ingestion containers will spin up and continue running, but the main server and `execute-migrate-all` will both crash with a `Malformed YAML at line: . . . customParams: {}}` error. I posted about this on Slack, [here](https://openmetadata.slack.com/archives/C02B6955S4S/p1774284294197529).

**Expected behavior**
I expected the migration and main server contains to run propery.

**Version:**
 - OS: MacOS Sequoia 15.5
 - Python version: 3.13.7
 - OpenMetadata version: 1.12
 - OpenMetadata Ingestion package version: N/A

**Additional context**
I realize Docker is the preferred path, but this minor fix enables the quickstart to succeed in Podman, which is great for organizations in which Docker is not an option.

SaaiAravindhRaja:
## Summary
Quotes the empty-object default for `OIDC_CUSTOM_PARAMS` in the Docker compose files so Compose does not misparse `{}` during variable expansion.

- updates all quickstart, development, and openmetadata compose variants that expose `OIDC_CUSTOM_PARAMS`
- keeps the default semantic value as an empty JSON object
- resolves #26775

## Test Plan
- `docker compose -f docker/docker-compose-quickstart/docker-compose.yml config`
- `docker compose -f docker/docker-compose-openmetadata/docker-compose-openmetadata.yml config`


PubChimps:
Hi @SaaiAravindhRaja, thanks for your pr here! We had a PR submitted for this already [here](https://github.com/open-metadata/OpenMetadata/pull/26965). Please make sure that you are assigned an issue and that there is not a PR out for it already before submitting something. We will reopen this PR if necessary.

PR #27017, opened 2026-04-03
disable trivy scans from PR checks

karanh37:


.github/workflows/trivy-scan-ingestion-base-slim-image.yml
akash-jain-10: @karanh37  - This needs to be changed as well. Here and other files as well since we are removing the CI trigger of Pull Request :)

akash-jain-10 (approved):

PR #27019, opened 2026-04-03
Fixes #26869: Defensive type handling in AirflowRESTClient to prevent ClassCastException

### LINKED ISSUE #26869: ClassCastException: Integer cannot be cast to String in PipelineServiceClientFactory due to unquoted numeric YAML defaults
The Dropwizard YAML config file (`openmetadata.yaml`) has unquoted numeric defaults throughout the `pipelineServiceClientConfiguration.parameters` section and elsewhere. When the YAML parser processes these, numeric values like `3`, `1000`, `604800` are deserialized as `java.lang.Integer` or `java.lang.Long` rather than `String`. This causes a ClassCastException when the configuration is read into maps typed as `Map<String, String>`.

```
org.openmetadata.sdk.exception.PipelineServiceClientException: Error trying to load PipelineServiceClient
org.openmetadata.service.clients.pipeline.k8s.K8sPipelineClient: java.lang.ClassCastException:
class java.lang.Integer cannot be cast to class java.lang.String
```

Examples of unquoted numeric defaults in `openmetadata.yaml`:

```yaml
backoffLimit: ${K8S_BACKOFF_LIMIT:-3}
runAsUser: ${K8S_RUN_AS_USER:-1000}
ttlSecondsAfterFinished: ${K8S_TTL_SECONDS_AFTER_FINISHED:-604800}
timeout: ${AIRFLOW_TIMEOUT:-10}
healthCheckInterval: ${PIPELINE_SERVICE_CLIENT_HEALTH_CHECK_INTERVAL:-300}
```

Even when the env vars ARE set as strings, the YAML substitution produces unquoted values (e.g. `backoffLimit: 3`) which the YAML parser reads as Integer.

This affects the K8sPipelineClient specifically because the `parameters` map contains a mix of string and numeric values, and the deserialization chain attempts a direct String cast on Integer values.

Note: the `K8sPipelineClientConfig` class itself handles this correctly with `value.toString()` calls in `getStringParam`/`getIntParam`. The cast failure happens earlier, during Jackson/Dropwizard deserialization of the `additionalProperties` map in `PipelineServiceClientConfiguration`.

**Fix:** Quote all numeric defaults in `openmetadata.yaml`:

```yaml
backoffLimit: "${K8S_BACKOFF_LIMIT:-3}"
runAsUser: "${K8S_RUN_AS_USER:-1000}"
```

This ensures the YAML parser always produces String values regardless of whether the env var is set.

**Workaround:** Mount a patched `openmetadata.yaml` with all numeric defaults quoted.

- OpenMetadata version: 1.12.3
- Deployment: Kubernetes (GKE) with K8sPipelineClient and OMJob Operator
- Database: PostgreSQL (Cloud SQL)

RajdeepKushwaha5:
## Describe your changes:

Fixes #26869

### Problem

`AirflowRESTClient` constructor performs direct `(Integer)` and `(String)` casts on values read from `config.getParameters().getAdditionalProperties()`. When SnakeYAML parses unquoted numeric values like `timeout: 60` in `openmetadata.yaml`, it deserializes them as `java.lang.Integer`. However, the code later casts these values with `(String)`, causing a `ClassCastException` at runtime:

```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String
```

This crashes the server on startup when pipeline service client parameters contain unquoted numeric YAML values.

### Solution

Instead of fragile direct casts, this PR introduces two defensive helper methods in `AirflowRESTClient`:

- **`getStringParam(params, key)`** — Retrieves a value and converts it to `String` via `value.toString()`, safely handling `Integer`, `Long`, `String`, or any other type SnakeYAML may produce.
- **`getIntParam(params, key, defaultValue)`** — Retrieves a value and converts it to `int` via `Integer.parseInt(value.toString())`. Logs a `LOG.warn` with the actual value if parsing fails, then falls back to the provided default.

Additional improvements:

- **Null-safe handling** of both `config.getParameters()` and `config.getParameters().getAdditionalProperties()` using `Collections.emptyMap()` fallback — prevents NPE when `Parameters` is present but `additionalProperties` is null.
- **Credentials validation**: throws `PipelineServiceClientException` early if `username` or `password` is null after parameter extraction.

### Testing

Added 7 new unit tests in `AirflowRESTClientTest`:

| Test | Scenario |
|------|----------|
| `constructorHandlesStringTimeout` | timeout provided as `"60"` (String) |
| `constructorHandlesIntegerTimeout` | timeout provided as `60` (Integer) |
| `constructorHandlesLongTimeout` | timeout provided as `60L` (Long) |
| `constructorHandlesNullParameters` | `config.getParameters()` is null |
| `constructorHandlesNonNumericTimeout` | timeout provided as `"abc"` — exercises warn+fallback branch |
| `constructorHandlesNullAdditionalProperties` | `Parameters` is non-null but `additionalProperties` is null |
| `constructorHandlesMissingCredentials` | missing username/password throws exception |

All tests pass locally.

### Type of change:

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

### Checklist:

### DIFF
--- a/bin/openmetadata-server-start.sh
+++ b/bin/openmetadata-server-start.sh
@@ -105,7 +105,10 @@ fi
 if [ -z "$OPENMETADATA_JVM_PERFORMANCE_OPTS" ]; then
   export OPENMETADATA_JVM_PERFORMANCE_OPTS="\
     -server -XX:+UseG1GC \
+    -XX:+UnlockExperimentalVMOptions \
     -XX:MaxGCPauseMillis=200 \
+    -XX:G1NewSizePercent=5 \
+    -XX:G1MaxNewSizePercent=20 \
     -XX:InitiatingHeapOccupancyPercent=45 \
     -XX:+ExplicitGCInvokesConcurrent \
     -XX:+UseStringDeduplication \
--- a/bin/openmetadata.sh
+++ b/bin/openmetadata.sh
@@ -153,7 +153,7 @@ fi
 
 # JVM performance options
 if [[ -z "${CATALOG_JVM_PERF_OPTS}" ]]; then
-  CATALOG_JVM_PERF_OPTS="-server -XX:+UseG1GC -XX:MaxGCPauseMillis=20 -XX:InitiatingHeapOccupancyPercent=35 -XX:+ExplicitGCInvokesConcurrent -Djava.awt.headless=true"
+  CATALOG_JVM_PERF_OPTS="-server -XX:+UseG1GC -XX:+UnlockExperimentalVMOptions -XX:MaxGCPauseMillis=200 -XX:G1NewSizePercent=5 -XX:G1MaxNewSizePercent=20 -XX:InitiatingHeapOccupancyPercent=35 -XX:+ExplicitGCInvokesConcurrent -Djava.awt.headless=true"
 fi
 
 
--- a/conf/openmetadata.yaml
+++ b/conf/openmetadata.yaml
@@ -78,7 +78,7 @@ server:
   # Virtual Threads (Project Loom) - Recommended for Java 21+
   # Jetty 12 uses AdaptiveExecutionStrategy to route blocking tasks to virtual threads
   # while keeping non-blocking I/O on platform threads for optimal cache locality
-  enableVirtualThreads: ${SERVER_ENABLE_VIRTUAL_THREAD:-true}
+  enableVirtualThreads: ${SERVER_ENABLE_VIRTUAL_THREAD:-false}
   # Note: maxQueuedRequests removed in Dropwizard 5.0/Jetty 12
 
   # Request/Response logging (disable in production for performance)
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ActivityFeed.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ActivityFeed.spec.ts
@@ -13,7 +13,6 @@
 import { expect, Page, test as base } from '@playwright/test';
 import { ApiEndpointClass } from '../../support/entity/ApiEndpointClass';
 import { DatabaseClass } from '../../support/entity/DatabaseClass';
-import { EntityDataClass } from '../../support/entity/EntityDataClass';
 import { TableClass } from '../../support/entity/TableClass';
 import { PersonaClass } from '../../support/persona/PersonaClass';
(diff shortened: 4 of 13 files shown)


conf/openmetadata-s3-logs.yaml
```diff
@@ -46,8 +46,8 @@ pipelineServiceClientConfiguration:
   parameters:
     username: ${AIRFLOW_USERNAME:-admin}
     password: ${AIRFLOW_PASSWORD:-admin}
-    timeout: ${AIRFLOW_TIMEOUT:-10}
-    connection_timeout: ${AIRFLOW_CONNECTION_TIMEOUT:-30}
+    timeout: "${AIRFLOW_TIMEOUT:-10}"
```
harshach: @RajdeepKushwaha5 any reason to add double quotes, we rather be not introduce "" in some places and not have double quotes in other places. Lets make it consistent and do the parsing in the code

RajdeepKushwaha5: @harshach I've reverted all the YAML quote changes. The fix now lives entirely in the Java code: `AirflowRESTClient` uses `getStringParam()`/`getIntParam()` helpers that call `value.toString()` to handle any type SnakeYAML produces (String, Integer, Long). This also adds null-safety for missing `parameters` config and logs a warning for invalid integer values. The YAML files are untouched.

openmetadata-service/src/main/java/org/openmetadata/service/clients/pipeline/airflow/AirflowRESTClient.java
```diff
@@ -76,18 +77,26 @@ public AirflowRESTClient(PipelineServiceClientConfiguration config) throws KeySt
...
+        config.getParameters() != null && config.getParameters().getAdditionalProperties() != null
+            ? config.getParameters().getAdditionalProperties()
+            : Collections.emptyMap();
+    this.username = getStringParam(params, USERNAME_KEY);
+    this.password = getStringParam(params, PASSWORD_KEY);
+    if (this.username == null || this.password == null) {
+      throw new PipelineServiceClientException(
+          "Missing Airflow credentials. Ensure 'username' and 'password' are configured in "
+              + "pipelineServiceClientConfiguration.parameters. "
+              + DOCS_LINK);
```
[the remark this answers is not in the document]
RajdeepKushwaha5: The approach changed per @ harshach's review

PR #27020, opened 2026-04-03
Fixes #26833: Quote credential env vars in YAML configs to prevent parse failures

### LINKED ISSUE #26833: Database password with special characters breaks runtime YAML config parsing
The application writes the database password (from the `DB_USER_PASSWORD` environment variable) into the Dropwizard runtime YAML configuration without quoting. Passwords containing YAML special characters (`[` `]` `{` `}` `#` etc) cause a parse failure at startup:

```
Malformed YAML at line: 209, column: 49; while parsing a flow sequence
password: [eUtyO5BWqQ07+s#rg!K8c3LGQW=4P(k ...
```

The Helm chart correctly passes the password through K8s secrets and env vars. The issue is in the application's config generation — the env var value is interpolated into YAML without being quoted or escaped.

This is common with auto-generated passwords from managed database services (e.g. Cloud SQL managed users) which use the full printable character set.

**Workaround:** Use passwords without YAML-hostile characters.

- OpenMetadata version: 1.12.3
- Deployed via Helm on GKE with external Cloud SQL (PostgreSQL 17)

RajdeepKushwaha5:
### Describe your changes:

Fixes #26833

**Root cause:** Dropwizard's `SubstitutingSourceProvider` performs raw string substitution on YAML config text **before** SnakeYAML parsing. When credential environment variables (`DB_USER_PASSWORD`, `AIRFLOW_PASSWORD`, etc.) contain YAML-special characters like `[`, `{`, or `#`, the unquoted interpolation produces invalid YAML that fails at startup.

For example, with `DB_USER_PASSWORD=[eUtyO5BWqQ07+s#rg!K8c3LGQW=4P(k`:
```yaml
# Before substitution (unquoted)
password: ${DB_USER_PASSWORD:-openmetadata_password}

# After substitution → invalid YAML ([ starts a flow sequence)
password: [eUtyO5BWqQ07+s#rg!K8c3LGQW=4P(k
```

**Fix:** Wrap all credential `${}` expressions in YAML double quotes so the substituted value is always parsed as a quoted YAML string regardless of content:
```yaml
# Before substitution (quoted)
password: "${DB_USER_PASSWORD:-openmetadata_password}"

# After substitution → valid YAML (quoted string)
password: "[eUtyO5BWqQ07+s#rg!K8c3LGQW=4P(k"
```

**Fields fixed in `conf/openmetadata.yaml`:**
- `database.user` and `database.password` (the reported issue)
- `authenticationConfiguration` LDAP `trustStoreFilePassword`
- `pipelineServiceClientConfiguration.parameters` Airflow `username`/`password`
- `cache` Redis `username`/`passwordRef`

**Fields fixed in `conf/openmetadata-s3-logs.yaml`:**
- AWS `awsAccessKeyId`/`awsSecretAccessKey`
- Airflow parameters `username`/`password`

**Testing:** This is a YAML-only configuration change. The quoting ensures that after environment variable substitution, the resulting YAML is always syntactically valid regardless of the characters in the credential values. No Java code changes are needed — Dropwizard's YAML parser handles quoted strings correctly.

#
### Type of change:
- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

> **Note:** No automated test is necessary for this fix — it's a YAML configuration quoting change, not a code logic change. The fix is self-evident: wrapping `${}` expressions in YAML double quotes guarantees the substituted value is always a valid YAML string. Manual verification: setting `DB_USER_PASSWORD` to a value containing `[`, `{`, or `#` and confirming the application starts successfully."


ayush-shah:

PR #27021, opened 2026-04-03
Fixes #26930: Fix check_prerequisites.sh subshell variable scoping bug

### LINKED ISSUE #26930: check_prerequisites.sh script incorrectly reports
**Affected module**
Ingestion Framework / Build Scripts

**Describe the bug**
The `check_prerequisites.sh` script incorrectly reports "All prerequisites are met" even when multiple dependencies (e.g., Python, Java, Maven) are missing or fail the version check. This occurs because the script updates the status variable (`code`) inside a subshell, which prevents the parent shell from seeing the failure state.

**To Reproduce**
1. Ensure several required tools are not installed (e.g., uninstall `antlr4` or `mvn`).
2. Run the script: `./scripts/check_prerequisites.sh` (or via `make prerequisites`).
3. Observe the output:
   - It correctly identifies missing tools: `✗ ERROR: java is not installed.`
   - **However**, the final summary line says: `✓ All prerequisites are met.`
   - The script exits with code `0`.

**Expected behavior**
If any dependency check fails, the final output should display `✗ ERROR: Some prerequisites are not met.` and the script should exit with a non-zero exit code (e.g., `1` or `2`) to stop the build process.

**Version:**
 - **OS:** Linux/macOS
 - **Python version:** N/A (Missing during reproduction)
 - **OpenMetadata version:** 1.x / Main branch
 - **OpenMetadata Ingestion package version:** N/A (Script-level issue)

**Additional context**
The bug is located in the main loop of the script:
```bash
if [[ $(check_command_existence "$command") -ne 0 ]]; then
    continue
fi
```
The use of `$(...)` executes the function in a **subshell**. Any changes made to the `code` variable inside `check_command_existence` are lost when the subshell terminates, meaning the parent script always sees `code=0`. This can be fixed by calling the function directly and checking its return status instead of using command substitution.

RajdeepKushwaha5:
### Describe your changes:

Fixes #26930

**Root cause:** `check_command_existence` was called via command substitution `$(check_command_existence "$command")`, which runs the function in a **subshell**. The `code=2` assignment inside the function modified a copy of the variable in the subshell, which was discarded when the subshell exited. The parent shell always saw `code=0`, causing the script to report "All prerequisites are met" even when multiple tools were missing.

Additionally, the error message inside `check_command_existence` referenced `$command` (the outer loop variable) instead of `$1` (the function's own parameter) — a latent bug that happened to work only because the outer variable was always in scope.

**Fix (3 lines changed):**
1. Changed `echo $res` → `return $res` — the function now communicates its result via exit status instead of stdout
2. Changed `if [[ $(check_command_existence "$command") -ne 0 ]]` → `if ! check_command_existence "$command"` — calls the function directly (no subshell), so `code=2` persists in the parent shell
3. Changed `$command` → `$1` inside the function — use the function parameter, not the outer scope variable

**Before (broken):**
```
$ # With java, maven, antlr missing:
$ ./scripts/check_prerequisites.sh
✗ ERROR: java is not installed.
✗ ERROR: mvn is not installed.
✗ ERROR: antlr4 is not installed.
✓ All prerequisites are met.    # ← WRONG
$ echo $?
0                                # ← WRONG
```

**After (fixed):**
```
$ ./scripts/check_prerequisites.sh
✗ ERROR: java is not installed.
✗ ERROR: mvn is not installed.
✗ ERROR: antlr4 is not installed.
✗ ERROR: Some prerequisites are not met.  # ← CORRECT
$ echo $?
2                                          # ← CORRECT
```

#
### Type of change:
- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

> **Note:** This is a shell script fix (3 lines changed). The fix is self-evident from the diff — removing the command substitution subshell ensures the `code` variable is modified in the parent shell scope. Manual verification: run the script with missing tools and confirm the exit code is non-zero and the final message says "Some prerequisites are not met."

----
## Summary by Gitar

- **UI and Frontend Updates:**
  - Updated `Tier` and `Certification` tag FQN matching logic to use prefix instead of substring.
  - Migrated incident test case status and severity components from MUI to core components.
  - Upgraded `uuid` and `dompurify` dependencies for improved security.
  - Allowed `BlockEditor` file attachments and callout attributes for `div` elements in the description sanitizer.
  - Hided task forms and workflows from the global settings view.
- **Ownership Fixes:**
  - Resolved issue where team ownership was incorrectly handled for names with special characters or partial matches.
- **Tooling and Infrastructure:**
  - Migrated ingestion codebase to `ruff` for formatting, `isort` imports, and removing unused imports.


RajdeepKushwaha5:
ping @harshach

ayush-shah:

PR #27024, opened 2026-04-03
fix: display entity name + FQN for entityReference custom properties

### LINKED ISSUE #26939: CustomProperty lists tables, only table name is shown (and not fqn)
**Affected module**
UI

**Describe the bug**
when customProperty lists tables, only table name is shown (and not fqn) - difficult to know which table is which if same name exists across systems.

Image

**To Reproduce**

**Expected behavior**
fqn is displayed (might be smaller letters) next to table name

**Version:**
1.12.3 as well

**Additional context**

Shivang-01ux:
##  Problem
When a CustomProperty of type `entityReference` or `entityReferenceList` 
points to a table (or any entity), only the short name was shown — e.g. 
`customers` instead of `myservice.mydb.myschema.customers`.

This makes it **impossible to distinguish between entities with the same 
name across different services or schemas** — a common real-world scenario.

##  Root Cause
`getEntityRefLinkValue` was calling `getEntityName(item)` which returns 
only `displayName || name`. The `fullyQualifiedName` was used for the 
link href but **never rendered to the user**.

##  Fix
- Shows **displayName** on the first line (human-readable, as before)
- Shows **fullyQualifiedName** below in smaller grey text with tooltip on hover
- FQN is **only shown when it differs from displayName** — avoids duplicate 
  text for users/teams where name === fqn
- Entity reference remains a **clickable link** to the entity page
- Works for **all entity types**: table, dashboard, pipeline, topic, mlmodel, etc.
- Fixes both `entityReference` (single) and `entityReferenceList` (multiple)

##  Before / After
**Before:** `customers`

**After:**
`customers`
`myservice.mydb.myschema.customers` ← grey, smaller text, with tooltip on hover

Fixes #26939


Shivang-01ux:
Note: This fix includes a guard `showFqn = fqn && fqn !== displayName` 
that prevents showing duplicate text for entity types like users and 
teams where the name and FQN are identical — avoiding redundant UI noise 
that other approaches may show.

Shivang-01ux:
@PubChimps Hi! I've submitted PR #27024 fixing issue #26939. The fix displays the full FQN with a tooltip for entityReference custom properties. Would appreciate a review when you get a chance!

harshach:
@Shivang-01ux please provide a screenshot of this fix in the UI

Shivang-01ux:
Hi @harshach! 👋
Since I'm unable to run OpenMetadata locally on Windows, here's a UI mockup showing exactly what this fix addresses.
UI fix

Before: entityReference showed only the short name (dim_customer) — ambiguous across services/schemas. entityReferenceList showed duplicate orders entries with no way to distinguish them. No links, no hover context.

After: Full FQN displayed as a clickable link (ecommerce_db.prod.dim_customer) with tooltip on hover. List entries are now uniquely identifiable (shopify.raw.orders vs warehouse.prod.orders).
Change: In getEntityRefLinkValue() → PropertyValue.tsx

getEntityName(item) → item.fullyQualifiedName
Added <Tooltip> for hover path + <Link> for navigation
showFqn guard prevents duplicate text rendering

Happy to make any adjustments!

Rohit0301:
The issue raised was that the user wanted the FQN to be displayed in the dropdown list. However, the PR adds the FQN alongside the property value instead. Closing the PR because it does not address the actual issue

PR #27043, opened 2026-04-04
Fixes #27040: Guard postDelete() with hardDelete check in IngestionPipeline, DataContract, App

### LINKED ISSUE #27040: [Bug] Soft delete permanently destroys time series data for IngestionPipeline, DataContract, and App — restore cannot recover it
**Affected module**
Backend

**Describe the bug**
Several entity repositories (`IngestionPipelineRepository`, `DataContractRepository`, `AppRepository`) unconditionally delete time series data in their `postDelete()` methods without checking the `hardDelete` flag. This means even a **soft delete** permanently wipes execution history, contract results, and app run records. Since `restoreEntity()` only flips the `deleted` flag back and does not restore time series data, a soft-delete → restore cycle causes **permanent, irrecoverable data loss**.

The base `EntityRepository.postDelete()` correctly guards destructive operations with `if (hardDelete)`, but these three subclasses do not follow the same pattern.

**Affected code:**

| Repository | File | Issue |
|---|---|---|
| `IngestionPipelineRepository` | `openmetadata-service/.../jdbi3/IngestionPipelineRepository.java` (postDelete ~L475-489) | Deletes pipeline status time series + destroys deployed Airflow pipeline on soft delete |
| `DataContractRepository` | `openmetadata-service/.../jdbi3/DataContractRepository.java` (postDelete ~L241-250) | Deletes contract result time series + destroys associated test suite on soft delete |
| `AppRepository` | `openmetadata-service/.../jdbi3/AppRepository.java` (postDelete ~L249-257) | Deletes app run status history on soft delete |

**To Reproduce**
1. Create an ingestion pipeline and run it several times to accumulate execution history
2. Soft-delete the pipeline via UI or API (`hardDelete=false`, which is the default)
3. Restore the pipeline via API (`PUT /api/v1/ingestionPipelines/restore`)
4. Check the execution history tab — **all run history is permanently gone**

Same steps apply for DataContract (contract results) and App (run records).

**Expected behavior**
Soft delete should preserve all associated data so that restoring the entity fully recovers it — including time series execution/result history. Destructive time series cleanup should only happen on hard delete (`hardDelete=true`).

**Suggested fix**
Wrap the destructive operations in each `postDelete()` with `if (hardDelete) { ... }`, matching the pattern used in the base `EntityRepository.postDelete()`.

**Version:**

**Additional context**
This is a systematic issue across 3 repositories. The base class pattern is correct — the subclasses simply didn't follow it. Related to #27001 which fixed a similar (but distinct) issue for `PipelineRepository` hard delete cleanup.

---

RajdeepKushwaha5:
### Describe your changes:

Fixes #27040

**What:** Wrapped destructive operations in `postDelete()` with `if (hardDelete)` guards in three entity repositories:

| Repository | What was being destroyed on soft delete |
|---|---|
| `IngestionPipelineRepository` | Pipeline status time series + deployed Airflow pipeline |
| `DataContractRepository` | Contract result time series + associated test suite |
| `AppRepository` | App run status history |

**Why:** These three subclasses unconditionally delete time series data and external resources in their `postDelete()` methods — even on soft delete (`hardDelete=false`). Since `restoreEntity()` only flips the `deleted` flag back without restoring time series data, a soft-delete → restore cycle causes **permanent, irrecoverable data loss**.

The base `EntityRepository.postDelete()` correctly guards its destructive operation (RDF deletion) with `if (hardDelete)`, but these subclasses did not follow the same pattern.

**How I tested:** Verified that the three modified files compile cleanly and that spotless formatting passes. The fix is a straightforward conditional guard — same pattern used by the base class and other repositories like `PipelineRepository` (fixed in #27037).

#
### Type of change:
- [x] Bug fix

#
### Checklist:

> **Note on testing:** This is a minimal logic guard (`if (hardDelete)`) wrapping existing code — the same pattern already used by the base `EntityRepository.postDelete()`. The fix does not introduce new logic paths; it prevents existing destructive operations from running on soft delete. Integration test coverage for the hard-delete path already exists through the entity lifecycle tests.


RajdeepKushwaha5:
Superseded by #27051 which centralizes all timeseries cleanup in `EntityRepository.cleanup()` per @harshach's feedback.

Issue #27042, opened 2026-04-04
[Bug] Hard delete of WebAnalyticEvent does not clean up event data from entity_extension_time_series


RajdeepKushwaha5:
**Affected module**
Backend

**Describe the bug**
When a `WebAnalyticEvent` entity is permanently deleted (hard delete), its associated event data rows remain in the `entity_extension_time_series` table. `WebAnalyticEventRepository` writes event data via `storeTimeSeries()` with extension `webAnalyticEvent.webAnalyticEventData`, but has no `entitySpecificCleanup()` or `postDelete()` override to clean up these records.

The base `EntityRepository.cleanup()` does not include generic time series cleanup, so each repository that writes time series data must handle its own cleanup — `WebAnalyticEventRepository` does not.

**To Reproduce**
1. Create a web analytics event
2. Trigger analytics data collection to populate event data
3. Permanently delete the web analytics event via API: `DELETE /api/v1/webAnalyticEvent/{id}?hardDelete=true`
4. Query the database:
```sql
SELECT COUNT(*) FROM entity_extension_time_series 
WHERE extension = 'webAnalyticEvent.webAnalyticEventData'
AND entityFQNHash = '<hash_of_deleted_event>';
-- Returns > 0 even after hard delete
```

**Expected behavior**
Hard-deleting a `WebAnalyticEvent` should remove all associated event data from `entity_extension_time_series`.

**Suggested fix**
Add an `entitySpecificCleanup()` override to `WebAnalyticEventRepository`:
```java
@Override
protected void entitySpecificCleanup(WebAnalyticEvent entity) {
    daoCollection
        .entityExtensionTimeSeriesDao()
        .delete(entity.getName(), WEB_ANALYTICS_EVENT_DATA_EXTENSION);
}
```

**Version:**
- OpenMetadata version: 1.6.0-SNAPSHOT (main branch, commit 58a76d72f8)

**Additional context**
Same class of bug as #27001 (PipelineRepository missing time series cleanup). Part of a broader audit of time series cleanup gaps across all entity repositories.

RajdeepKushwaha5:
Hi @harshach ! I identified this during a systematic audit of time series cleanup across all entity repositories — triggered by the same class of bug reported in #27001.

[...]

PR #27044, opened 2026-04-04
Fixes #27041: Clean up profiler time series data on hard delete of Table

### LINKED ISSUE #27041: [BUG] Hard delete of Table does not clean up profiler data from profiler_data_time_series — stale profiles resurface on re-creation
**Affected module**
Backend

**Describe the bug**
When a Table entity is permanently deleted (hard delete), its associated profiler data — table profiles, column profiles, and system profiles — remain in the `profiler_data_time_series` database table. If a table with the same fully qualified name is later re-created (e.g., re-ingested), the old profiling data resurfaces in the UI.

`TableRepository` writes to `profiler_data_time_series` using three extensions:
- `table.tableProfile`
- `table.columnProfile`
- `table.systemProfile`

But it has no `entitySpecificCleanup()` or `postDelete()` override to delete these records on hard delete. The base `EntityRepository.cleanup()` does not include any generic time series cleanup either.

**To Reproduce**
1. Create a database service and ingest a table
2. Run the profiler to generate table and column profile data
3. Permanently delete the table via API: `DELETE /api/v1/tables/name/{fqn}?hardDelete=true&recursive=true`
4. Re-ingest the same table (same fully qualified name)
5. Navigate to the Profiler tab for the table — **old profile data from the deleted table appears**

You can also verify directly in the database:
```sql
SELECT COUNT(*) FROM profiler_data_time_series 
WHERE entityFQNHash = '<hash_of_deleted_table_fqn>';
-- Returns > 0 even after hard delete
```

**Expected behavior**
Hard-deleting a table should remove all associated profiler data (table profiles, column profiles, system profiles) from `profiler_data_time_series`. Re-creating a table with the same name should start with a clean profiler history.

**Suggested fix**
Add an `entitySpecificCleanup()` override to `TableRepository` that deletes from `profiler_data_time_series` for all three profile extensions. Note: column profiles are stored with column-level FQNs (e.g., `service.db.schema.table.column`), so a prefix-based delete method may need to be added to `EntityTimeSeriesDAO` (currently only exact FQN match exists).

**Version:**

**Additional context**
Similar to #27001 where `PipelineRepository` was missing time series cleanup on hard delete. This is part of a broader pattern where repositories that write to time series tables don't always clean up on entity deletion.

---

RajdeepKushwaha5:
### Describe your changes:

Fixes #27041

**What:** Added `entitySpecificCleanup()` override to `TableRepository` that deletes all profiler data from `profiler_data_time_series` when a table is hard-deleted:

| Extension | Scope | How deleted |
|---|---|---|
| `table.tableProfile` | Table-level profiles | By exact table FQN |
| `table.systemProfile` | System-level profiles | By exact table FQN |
| `table.columnProfile` | Column-level profiles | By iterating all column FQNs (including nested children) |

**Why:** When a Table entity is hard-deleted, the base `EntityRepository.cleanup()` removes entity extensions, relationships, tags, etc. — but it does **not** clean up `profiler_data_time_series`, which is a separate table. Without this fix, orphaned profiler records remain and resurface when a table with the same FQN is re-ingested.

Column profiles are stored with column-level FQNs (e.g., `service.db.schema.table.column`), so prefix-based deletion isn't possible on hashed FQN columns. Instead, the fix uses the existing `collectColumnFqns()` utility to enumerate all column FQNs (including nested struct children) and deletes each one individually.

**How I tested:** Verified spotless formatting passes and no compilation errors from the modified file. The fix follows the same pattern as `TestCaseRepository.entitySpecificCleanup()` which similarly cleans up time series data on hard delete. The `entitySpecificCleanup()` hook is only called during hard delete (from `EntityRepository.cleanup()`), so soft delete behavior is unaffected.

#
### Type of change:
- [x] Bug fix

#
### Checklist:

> **Note on testing:** This is a targeted cleanup addition using existing DAO methods (`ProfilerDataTimeSeriesDAO.delete()`) and existing utilities (`collectColumnFqns()`). The `entitySpecificCleanup()` hook is a well-established pattern used by `TestCaseRepository`, `PipelineRepository`, `StoredProcedureRepository`, and others. Integration test coverage for the hard-delete lifecycle path already exists through entity lifecycle tests.


harshach:
@RajdeepKushwaha5 can you not open multiple PRs on this topic, please merge all of these into single PRs given you are addressing common issue of entitytimeseries data not getting deleted, also try to do this in entityRepository if there is a timeseries data associated with that way you can use one code to address all the entitiy tiem series data where its associated with an entity

RajdeepKushwaha5:
@harshach Thanks for the feedback — that makes sense.

I initially opened separate PRs since they were filed as individual issues (#27040, #27041, #27042), but I see your point that they share a common root cause: timeseries data not being cleaned up when entities are deleted.

I'll close these individual PRs and open a single consolidated PR that adds a generic `deleteByEntityId` to the base `EntityTimeSeriesDAO` and calls it from `EntityRepository.cleanup()` — so all entities with associated timeseries data are handled in one place.

Working on it now.

RajdeepKushwaha5:
Superseded by #27051 which centralizes all timeseries cleanup in `EntityRepository.cleanup()` per @harshach's feedback.

PR #27047, opened 2026-04-04
Fix/issue 26670 quicksight column aliases

Esvanth:
### **Problem**
CustomSql queries in QuickSight using column aliases (e.g., `SELECT src AS alias`) caused lineage links to be lost because the connector couldn't map the aliased data model column back to the source.

### **Solution**
- Refactored `_yield_lineage_from_query` to prioritize results from `lineage_parser`.
- Implemented `_build_column_lineage_from_parser` with robust multi-table matching (filtering by source column parent).
- Fixed critical Windows-specific environment issues in `setup.py` (uvloop) and `scripts/datamodel_generation.py` (Unicode/Case-sensitivity).

### **Verification**
- Verified with 10 passing unit tests in `ingestion/tests/unit/topology/dashboard/test_quicksight.py`.

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/quicksight/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/quicksight/metadata.py
@@ -604,7 +604,8 @@ class QuicksightSource(DashboardServiceSource):
                 if (
                     hasattr(src_col, "parent")
                     and src_col.parent
-                    and str(src_col.parent).lower() not in from_entity.name.root.lower()
+                    and str(src_col.parent).split(".")[-1].lower()
+                    != from_entity.name.root.lower()
                 ):
                     continue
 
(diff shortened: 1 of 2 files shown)


Esvanth:
Hi, could a maintainer please add the appropriate labels (e.g., type:bug-fix, component:ingestion) to trigger the CI checks? Thank you!

ulixius9:
@Esvanth looks like you mixed two issues in one pr can you please check?

I'm referring to profiler changes here

Esvanth:
Please add a safe to check parameter so it will be easy

PubChimps:
hi @Esvanth, thank you for this pr, but this issue was assigned to a pr in progress [here](https://github.com/open-metadata/OpenMetadata/pull/27301). we will reopen this pr if needed

PR #27057, opened 2026-04-05
Fixes #23646: Fix memory leak in scan_dags_job_background by adding singleton guard

### LINKED ISSUE #23646: Memory leaks for openmetadata-dependencies-web
Image

**Affected module**
Backend (openmetadata-managed-apis / Airflow integration)

**Describe the bug**
Memory leak in Airflow webserver pod when creating/updating ingestion pipelines via OpenMetadata UI (or calling deploy endpoint). Each time an ingestion pipeline is deployed (create/update operation), the webserver memory increases by ~120Mi and never releases until pod restart (manual or forced by OOM). Additionally, failed SchedulerJob entries accumulate in the airflow database with the webserver hostname. (Check Browse -> Jobs) at the airflow ui.

The root cause can be in `openmetadata_managed_apis/api/utils.py` - the `scan_dags_job_background()` function creates a SchedulerJob with `heartrate=0` (from the webserver pod?), which gets marked as "failed" by the main scheduler due to missing heartbeats.

**To Reproduce**
I followed the tutorial Helm installation for an empty local kind cluster without any customization to simplify the way to reproduce. Used everything specified in https://docs.open-metadata.org/latest/quick-start/local-kubernetes-deployment without deviations (only using https://github.com/kubernetes-sigs/kind/issues/1487#issuecomment-2211072952 to fix PVC problem).

**Steps:**
1. Create local kind cluster and install OpenMetadata via Helm (default values)
2. Create a database service via OpenMetadata UI (e.g., Postgres)
3. Add metadata ingestion pipeline via UI for the service
4. Monitor webserver memory: `kubectl top pod -n default -l component=web`
5. Click "Deploy" or "Re-deploy" the ingestion pipeline a few times. You can wait between these triggers longer.
6. Monitor webserver memory: `kubectl top pod -n default -l component=web`
7. Check database for failed SchedulerJobs

**Observed results:**
- Notable memory usage growth
- Failed SchedulerJobs (Entries with hostname like 'openmetadata-dependencies-web-8486b9478d-g4fkp')
- Standart SchedulerJob with `openmetadata-dependencies-scheduler-*` host runs fine

Code showing the issue:
https://github.com/open-metadata/OpenMetadata/blob/1fb18ef81578d59b156b9420fa74ce7815b30f7f/openmetadata-airflow-apis/openmetadata_managed_apis/operations/deploy.py#L179-L190

https://github.com/open-metadata/OpenMetadata/blob/1fb18ef81578d59b156b9420fa74ce7815b30f7f/openmetadata-airflow-apis/openmetadata_managed_apis/operations/deploy.py#L142-L173

https://github.com/open-metadata/OpenMetadata/blob/1fb18ef81578d59b156b9420fa74ce7815b30f7f/openmetadata-airflow-apis/openmetadata_managed_apis/api/utils.py#L168-L174

https://github.com/open-metadata/OpenMetadata/blob/1fb18ef81578d59b156b9420fa74ce7815b30f7f/openmetadata-airflow-apis/openmetadata_managed_apis/api/utils.py#L147

**Expected behavior**
- Deploy/update operations should be idempotent without memory leaks
- No failed SchedulerJob entries
- If scan_dags_job_background() is necessary (?), it should properly clean up resources

**Version:**
- Deployment: Kubernetes (kind cluster) with Helm chart using default values
- OpenMetadata version: 1.9.11 (also reproduced on 1.6.1)
- Airflow version: 2.10.5 (also reproduced on 2.9.1)

RajdeepKushwaha5:
### Describe your changes:

Fixes #23646

Each time an ingestion pipeline is deployed via the OpenMetadata UI, `scan_dags_job_background()` spawns a new `multiprocessing.Process` (`ScanDagsTask`). Each process:

1. Imports the entire Airflow scheduler stack (~120Mi of memory)
2. Creates a `SchedulerJob` with `heartrate=0`, which the main scheduler marks as "failed"
3. Is **never `join()`ed** by the parent — so it becomes a zombie process whose memory is never released

After N deploys, the webserver pod accumulates N × ~120Mi of leaked memory and N orphaned "failed" SchedulerJob entries in the Airflow database.

**Fix:** Add a per-worker singleton guard with a reaper thread to `scan_dags_job_background()`:

- **Per-worker guard** — a `threading.Lock` + module-level `_current_scan` reference prevents spawning multiple concurrent scan processes from the same Python worker
- **Reaper thread** — each scan spawns a lightweight daemon thread (`_reap_scan`) that `join()`s the process when it finishes, releasing resources and preventing zombies
- **Deferred rescan** — if a deploy arrives while a scan is already running, a `_rescan_requested` flag is set; the reaper automatically starts one follow-up scan after the current one completes, ensuring newly deployed DAGs are always discovered
- **Race-safe** — the rescan check is inside the `_current_scan is process` identity guard, so a stale reaper (whose process was already replaced) cannot spawn duplicates
- **No `daemon=True` on ScanDagsTask** — Airflow's scheduler internals fork child processes to parse DAGs, which Python forbids from daemon processes (`AssertionError: daemonic processes are not allowed to have children`). The reaper thread is daemonized instead.

Before (broken):
```python
def scan_dags_job_background():
    process = ScanDagsTask()
    process.start()
    # process is never joined — zombie, memory leaked
```

After (fixed):
```python
_scan_lock = threading.Lock()
_current_scan: Optional[ScanDagsTask] = None
_rescan_requested: bool = False

def _start_scan():
    """Start a new ScanDagsTask and spawn a reaper thread to join it."""
    global _current_scan, _rescan_requested
    _rescan_requested = False
    process = ScanDagsTask()
    process.start()
    _current_scan = process
    reaper = threading.Thread(target=_reap_scan, args=(process,), daemon=True)
    reaper.start()

def _reap_scan(process: ScanDagsTask):
    """Wait for the scan process to finish; start a follow-up if requested."""
    process.join()
    with _scan_lock:
        global _current_scan
        if _current_scan is process:
            _current_scan = None
            if _rescan_requested:
                logger.info("Running queued rescan after previous scan finished")
                _start_scan()

def scan_dags_job_background():
    with _scan_lock:
        if _current_scan is not None and _current_scan.is_alive():
            global _rescan_requested
            _rescan_requested = True
            logger.info("DAG scan already in progress, queued rescan")
            return
        _start_scan()
```

**2 files changed: `utils.py` (+45, -4), `test_scan_dags_singleton.py` (new, 7 test cases).**

#
### Type of change:

- [x] Bug fix

#
### Checklist:

- [x] My PR title is `Fixes #23646: Fix memory leak in scan_dags_job_background by adding singleton guard`

> **Note on testing:** The memory leak requires a running Airflow webserver + Kubernetes pod to reproduce (deploy pipeline N times, monitor memory via `kubectl top pod`). Unit tests in `tests/unit/test_scan_dags_singleton.py` verify the singleton guard logic by mocking `ScanDagsTask`:
> 1. First call starts a process
> 2. Concurrent call while scan is alive sets rescan flag (no duplicate spawn)
> 3. Finished scan gets replaced on next call
> 4. Reaper thread triggers follow-up scan when rescan was requested
> 5. Reaper clears state when no follow-up is needed
> 6. Process is never created with `daemon=True` (Airflow children would crash)
> 7. Stale reaper cannot spawn duplicates (race condition guard)

----
## Summary by Gitar

- **Logic update:**
  - The implementation simplifies the singleton guard by removing the `_rescan_requested` flag and the automatic follow-up scan, opting for a skip-if-busy approach as shown in the diff.
  - The reaper thread is now correctly implemented as a daemon thread in `_reap_scan` to join the process and clear `_current_scan` without re-triggering logic.


ayush-shah:

PR #27061, opened 2026-04-05
Fixes #27060: Fix TestCaseRepository deleteChildren O(N²) nested loop and double-delete in deleteAllTestCaseResults

### LINKED ISSUE #27060: [Bug] TestCaseRepository.deleteChildren() has O(N²) nested loop bug causing redundant deletes
**Affected module**

Backend — `openmetadata-service`

**Describe the bug**

`TestCaseRepository.deleteChildren()` (line 896) has a nested loop where **both the outer and inner loops iterate over the same `children` list**. This causes:

1. **O(N²) delete calls** — for N children, `deleteById()` is called N×N times instead of N
2. **Each child is deleted N times** instead of once
3. **`TestCaseResolutionStatusRepository` is re-instantiated on every outer iteration** instead of once
4. The outer loop variable `entityRelationshipRecord` is only used for logging — never for deletion
5. The log message uses `hardDelete ? "hard" : "soft"` ternary, but the enclosing `if (hardDelete)` guarantees it always logs `"hard"`

Additionally, in the same file, `deleteAllTestCaseResults()` (line 921) calls `testCaseResultRepository.deleteAllTestCaseResults(fqn)` **synchronously** and then submits the **exact same call asynchronously** via `asyncExecutor`, causing every test case result to be deleted twice.

**Current code (`deleteChildren`):**

```java
@Transaction
@Override
protected void deleteChildren(
    List<CollectionDAO.EntityRelationshipRecord> children, boolean hardDelete, String updatedBy) {
  if (hardDelete) {
    for (CollectionDAO.EntityRelationshipRecord entityRelationshipRecord : children) {  // outer loop
      LOG.info(
          "Recursively {} deleting {} {}",
          hardDelete ? "hard" : "soft",                    // always "hard"
          entityRelationshipRecord.getType(),
          entityRelationshipRecord.getId());
      TestCaseResolutionStatusRepository testCaseResolutionStatusRepository =
          (TestCaseResolutionStatusRepository)
              Entity.getEntityTimeSeriesRepository(Entity.TEST_CASE_RESOLUTION_STATUS);
      for (CollectionDAO.EntityRelationshipRecord child : children) {  // inner loop — SAME LIST
        testCaseResolutionStatusRepository.deleteById(child.getId(), hardDelete);
      }
    }
  }
}
```

**Current code (`deleteAllTestCaseResults`):**

```java
private void deleteAllTestCaseResults(String fqn) {
  TestCaseResultRepository testCaseResultRepository =
      (TestCaseResultRepository) Entity.getEntityTimeSeriesRepository(TEST_CASE_RESULT);
  testCaseResultRepository.deleteAllTestCaseResults(fqn);       // sync call
  asyncExecutor.submit(
      () -> {
        try {
          testCaseResultRepository.deleteAllTestCaseResults(fqn); // duplicate async call
        } catch (Exception e) {
          LOG.error("Error deleting test case results for test case {}", fqn, e);
        }
      });
}
```

**To Reproduce**

1. Create a `TestCase` entity
2. Create N `TestCaseResolutionStatus` children for it
3. Hard-delete the `TestCase`
4. Observe that `deleteById()` is called N² times (N iterations × N children per iteration) instead of N

For `deleteAllTestCaseResults`:
1. Create a `TestCase` with test results
2. Hard-delete the `TestCase`
3. Observe two `DELETE` queries against `data_quality_data_time_series` — one sync, one async

**Expected behavior**

- `deleteChildren()` should iterate `children` once and call `deleteById()` exactly N times (O(N))
- `TestCaseResolutionStatusRepository` should be instantiated once, outside the loop
- `deleteAllTestCaseResults()` should delete results exactly once (synchronously, since it runs inside a transaction)

**Version:**
- OpenMetadata version: main (1.7.0-SNAPSHOT)

**Additional context**

The base class `EntityRepository.deleteChildren()` has the correct single-loop pattern. Only the `TestCaseRepository` override introduces this N² bug.

The `deleteById()` method in `EntityTimeSeriesRepository` is idempotent (returns early if entity not found), so the redundant calls don't crash — they just waste N² database queries, which becomes a significant performance issue when deleting test cases with many resolution statuses.

RajdeepKushwaha5:
### Describe your changes:

Fixes #27060

**What changes did I make?**

Fixed two bugs in `TestCaseRepository.java`:

**Bug 1 — `deleteChildren()` O(N²) nested loop (lines 896–914):**

The method had a nested loop where both the outer and inner loops iterated over the **same** `children` list. For N children, `deleteById()` was called N×N times instead of N — each child was deleted N times. Additionally:
- `TestCaseResolutionStatusRepository` was re-instantiated on every outer iteration instead of once
- The outer loop variable `entityRelationshipRecord` was only used for logging, never for deletion
- The log message used `hardDelete ? "hard" : "soft"` but the enclosing `if (hardDelete)` guaranteed it always logged `"hard"`

**Fix:** Removed the outer loop entirely. Moved repository instantiation outside the loop. Single pass over `children` with one `deleteById()` call per child.

**Bug 2 — `deleteAllTestCaseResults()` double-delete (lines 921–933):**

The method called `testCaseResultRepository.deleteAllTestCaseResults(fqn)` synchronously, then submitted the **exact same call** asynchronously via `asyncExecutor`. This caused every test case result to be deleted twice — once immediately and once in a background thread (which would find nothing left to delete).

**Fix:** Removed the redundant async call. Kept only the synchronous call since this runs inside `entitySpecificCleanup()` → `cleanup()` which operates within a transaction. Also removed the now-unused `asyncExecutor` field and its imports (`ExecutorService`, `Executors`).

**Why did I make them?**

- Bug 1 causes O(N²) database operations when hard-deleting a TestCase with many resolution statuses, degrading performance significantly at scale
- Bug 2 wastes database queries by executing identical DELETE statements twice, with a potential race condition between the sync and async calls

**How did I test my changes?**

- Verified the base class `EntityRepository.deleteChildren()` uses a single loop (correct pattern)
- Confirmed `EntityTimeSeriesRepository.deleteById()` is idempotent (returns early if not found), meaning the bug causes silent performance degradation rather than exceptions
- Verified `asyncExecutor` had no other call sites in the file after removing the double-delete
- Ran `mvn spotless:check` — passes clean

#
### Type of change:
- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

  - The fix is a straightforward loop de-duplication. The `deleteById()` method is idempotent, so existing integration tests for TestCase hard-delete continue to validate the cleanup path. No new test is needed since the behavior (children get deleted) doesn't change — only the number of redundant DB calls does.


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -921,14 +910,6 @@ private void deleteAllTestCaseResults(String fqn) {
     TestCaseResultRepository testCaseResultRepository =
         (TestCaseResultRepository) Entity.getEntityTimeSeriesRepository(TEST_CASE_RESULT);
     testCaseResultRepository.deleteAllTestCaseResults(fqn);
-    asyncExecutor.submit(
```
harshach, line 1024: @RajdeepKushwaha5 any reason you removed the async excecutor here?

RajdeepKushwaha5: @harshach

The current code calls `deleteAllTestCaseResults(fqn)` **twice** — once synchronously and then the exact same call again via `asyncExecutor.submit()`:

```java
testCaseResultRepository.deleteAllTestCaseResults(fqn);     // deletes everything right here
asyncExecutor.submit(() -> {
    testCaseResultRepository.deleteAllTestCaseResults(fqn);  // runs later, finds nothing to delete
});
```

The sync call finishes and wipes all the results before the async task even starts, so the async one always ends up being a no-op — it just executes the same DELETEs against rows that no longer exist.

I kept the sync call and removed the async one (rather than the other way around) because `deleteAllTestCaseResults()` is called from `entitySpecificCleanup()`, which runs inside the JDBI transaction in [`cleanup()`](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java#L3692). The sync call participates in that transaction — so if anything downstream fails and the transaction rolls back, the results stay intact (consistent with the test case not being deleted). If we kept only the async call, it would run on a separate thread outside the transaction, meaning it could delete results even when the parent entity deletion rolled back.

So basically: the async call was redundant (double-delete), and the sync one is the correct one to keep for transactional safety.

Happy to restructure if you'd prefer a different approach — just let me know!

PR #27081, opened 2026-04-06
Governance Workflows: scheduleRunId correlation, batch trigger refactor, stage entity enrichment

yan-3005:
## Summary

- **Batch trigger refactor**: removes `MultiInstanceLoopCharacteristics` from `PeriodicBatchEntityTrigger`; one `CallActivity` per fetch iteration passes the full `entityList` to the main workflow. Inclusive gateways inside `MainWorkflow` handle per-entity fan-out within a single Flowable execution.
- **`scheduleRunId` correlation**: generates one UUID per trigger fire, propagated into every spawned `WorkflowInstance` and `WorkflowInstanceState` row — allows grouping all rows from one periodic trigger fire into a single scheduled run.
- **Stage entity enrichment**: `stage.entityList` resolved at stage START using the node's `inputNamespaceMap` (same logic nodes use at runtime), so each stage records exactly the entity list it received via routing (`entityList`, `gold_entityList`, `true_entityList`, etc.). `stage.updatedBy` set at stage END from node-namespaced variable.
- **Instance enrichment**: `WorkflowInstance` gains top-level `entityList` (from `global_entityList` at start) and `updatedBy` (scanned from variables at end).
- **DB migrations** (`1.14.0`): generated columns + indexes for `scheduleRunId` on both `workflow_instance_time_series` and `workflow_instance_state_time_series` for MySQL and Postgres.
- **API**: `scheduleRunId` query param on `WorkflowInstanceResource` and `WorkflowInstanceStateResource` for batch grouping queries.
- **`FetchChangeEventsImpl` empty-batch guard**: when all records in a fetch batch are deduped away, `entityList` is empty and Flowable's multi-instance loop would crash (`NoSuchElementException` on `entityList.get(0)`). Guard sets `HAS_FINISHED_VARIABLE = true` when `entityList` is empty so the trigger loop exits cleanly instead of crashing.
- **"Last event wins" task design** (`test_MultipleFieldChangesWithIncludeFields`): when a second entity-change event fires while a workflow instance is still running, the existing approval task is updated in-place (same task, new `about` / assignees) and the older workflow instance is terminated. The integration test now asserts `workflowInstanceId` is non-null and changes between the two workflow runs.

## Key design decisions

- `WorkflowScheduleRunIdSetterListener` fires **before** `WorkflowInstanceListener` (ordering enforced in `attachScheduleRunIdListener`) so the trigger's own `WorkflowInstance` row also carries the `scheduleRunId`.
- `stage.entityList` is fetched at stage **start** (not end) because the routing has already resolved which branch this execution is on — using `getEntityList(inputNamespaceMap, varHandler)` with the live `DelegateExecution`.
- `NewStageRequest` record replaces the 5-param method signature cap.
- `WorkflowVariableHandler.getEntityListFromVariables` static overload allows map-based resolution without a live `DelegateExecution`.

## Test plan

- [x] `WorkflowVariableHandlerTest` — 6 cases: plain `entityList`, banded lists (`gold_`, `silver_`, `true_`), key-not-found, empty map
- [x] `WorkflowScheduleRunIdReaderTest` — 5 cases: UUID object, UUID-as-string, namespaced fallback, null when absent, plain takes precedence
- [x] `WorkflowScheduleRunIdSetterListenerTest` — sets on first call, idempotent, unique IDs per process
- [x] `PeriodicBatchEntityTriggerTest` — no multi-instance loop, no inherited business key, `entityList` + `scheduleRunId` IOParameters, process structure
- [x] `TriggerFactoryTest` — factory creates correct trigger type
- [x] `FetchChangeEventsImplTest` — `testExecute_allRecordsDedupedAway_setsHasFinishedTrue`: all records deduped → `hasFinished=true`; no `NoSuchElementException`

----
## Summary by Gitar

- **Governance Workflow Fixes:**
  - Updated `MainWorkflow` to resolve `DataCompleteness` branch flags using `branchFlagVariable` instead of `bandFlagVariable`.


bootstrap/sql/migrations/native/2.0.0/mysql/schemaChanges.sql
```diff
@@ -145,3 +145,23 @@ CREATE INDEX idx_change_event_entity_type_offset ON change_event (entityType, `o
...
+    ADD COLUMN scheduleRunId VARCHAR(36)
+    GENERATED ALWAYS AS (json ->> '$.scheduleRunId');
+ALTER TABLE workflow_instance_state_time_series
+    ADD INDEX idx_workflow_instance_state_schedule_run_id (scheduleRunId);
+
+-- Update entityLink generated column to support both old (global_relatedEntity) and new (global_entityList[0]) formats
+ALTER TABLE workflow_instance_time_series
+MODIFY COLUMN entityLink TEXT GENERATED ALWAYS AS (
+    COALESCE(json ->> '$.variables.global_entityList[0]', json ->> '$.variables.global_relatedEntity')
+);
```
IceS2: This does not look right. The idea here is for us to understand what entity the workflow instance is for. If now we would have lists, probably this is not the best way of doing so.

bootstrap/sql/migrations/native/2.0.0/postgres/schemaChanges.sql
```diff
@@ -148,3 +148,24 @@ CREATE INDEX IF NOT EXISTS idx_change_event_entity_type_offset ON change_event (
...
+    GENERATED ALWAYS AS ((json ->> 'scheduleRunId')) STORED;
+CREATE INDEX idx_workflow_instance_state_schedule_run_id
+    ON workflow_instance_state_time_series (scheduleRunId);
+
+-- Update entityLink generated column to support both old (global_relatedEntity) and new (global_entityList[0]) formats
+ALTER TABLE workflow_instance_time_series DROP COLUMN entityLink;
+ALTER TABLE workflow_instance_time_series
+ADD COLUMN entityLink TEXT GENERATED ALWAYS AS (
+    COALESCE((json -> 'variables' -> 'global_entityList' ->> 0), (json -> 'variables' ->> 'global_relatedEntity'))
+) STORED;
```
IceS2, line 177: Same as above

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/PeriodicBatchEntityTrigger.java
```diff
@@ -158,25 +165,58 @@ private CallActivity getWorkflowTriggerCallActivity(
...
+              .build();
+      workflowTrigger.setLoopCharacteristics(multiInstance);
+
+      inParameters =
+          List.of(
+              relatedEntityParameter,
+              entityListParameter,
+              updatedByParameter,
+              scheduleRunIdParameter);
     } else {
```
IceS2: Is there any need for this? Can't we just pass the list of size 1?

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/PeriodicBatchEntityTrigger.java
```diff
@@ -216,6 +257,18 @@ private ServiceTask getFetchChangeEventsTask(
...
+      // Tell FetchChangeEventsImpl to fetch BATCH_PARALLELISM * batchSize records and split
+      // them into sub-batches. Not injected for singleExecution mode because that path uses
+      // per-entity MultiInstance (one entity per CallActivity) and has no concept of sub-batches.
+      FieldExtension parallelismExpr =
+          new FieldExtensionBuilder()
+              .fieldName("parallelismExpr")
+              .fieldValue(String.valueOf(BATCH_PARALLELISM))
+              .build();
+      serviceTask.getFieldExtensions().add(parallelismExpr);
+    }
```
IceS2: Again, just wondering if we need the single vs non single execution instead of computing the parameters directly

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/flowable/MainWorkflow.java
```diff
@@ -51,6 +53,16 @@ public MainWorkflow(WorkflowDefinition workflowDefinition) {
...
+
+    for (String event : List.of("start", "end")) {
+      org.flowable.bpmn.model.FlowableListener listener =
+          new FlowableListenerBuilder()
+              .event(event)
+              .implementation(WorkflowInstanceListener.class.getName())
+              .build();
+      process.getExecutionListeners().add(listener);
+    }
+
```
IceS2, line 58: Don't we have a method somewhere that does exactly this? Maybe we should reuse that

bootstrap/sql/migrations/native/2.0.0/mysql/schemaChanges.sql
yan-3005: Agreed — redesigned this. The key insight is that  is the right discriminator:

- **Batch (periodic) instances** always have  set → . These cover N entities; query them by  instead.
- **Single-entity instances** (event-based, NoOp) never have  →  stores the one meaningful entity link.

The new generated column expression:
```sql
CASE WHEN scheduleRunId IS NULL
THEN COALESCE(
    json ->> '$.variables.global_entityList[0]',
    json ->> '$.variables.global_relatedEntity'
)
ELSE NULL END
```

The COALESCE is intentional: we are not migrating existing `workflow_instance_time_series` records. Old event-based/NoOp instances carry `global_relatedEntity` in their JSON, not `global_entityList`. Without the fallback, those historical records would lose their `entityLink` after the 2.0.0 upgrade. New instances always populate `global_entityList`, so they hit the first branch of the COALESCE.

bootstrap/sql/migrations/native/2.0.0/postgres/schemaChanges.sql
yan-3005: Same fix applied here — see reply above.

bootstrap/sql/migrations/native/2.0.0/mysql/schemaChanges.sql
```diff
@@ -145,3 +145,29 @@ CREATE INDEX idx_change_event_entity_type_offset ON change_event (entityType, `o
...
+-- batch instances (scheduleRunId present) store NULL — query them by scheduleRunId instead.
+ALTER TABLE workflow_instance_time_series
+MODIFY COLUMN entityLink TEXT GENERATED ALWAYS AS (
+    CASE WHEN json ->> '$.scheduleRunId' IS NULL
+    THEN COALESCE(
+        json ->> '$.variables.global_entityList[0]',
+        json ->> '$.variables.global_relatedEntity'
+    )
+    ELSE NULL END
+);
```
IceS2, line 173: My issue here is: What is global_entityList[0] providing us? Feels we are doing it just because, but it is not really helping understand the whole batch the same way relatedEntity used to, no?

IceS2 (approved):

PR #26000, opened 2026-02-19
Hybrid search

lautel:
- Hybrid search support: Refactored VectorSearchQueryBuilder to extract the KNN query into a reusable buildQuery() method, enabling it to be embedded as a sub-query inside hybrid search requests
  - Threshold-based scoring: Added min_score support — when threshold > 0, the query uses min_score instead of k to filter results by relevance score (similarity score)
  - New filters: Added service and database filters to vector search, using a bool/should strategy that matches on both name (keyword) and displayName (text)
  - REFACTOR of the indexes: instead of using vector index, now entity indexes have the necessary fields. This simplifies a ton the overall vector handling
  - Vector index schema updates: Added service and database object mappings to the vector search index across all locales (en, jp, ru, zh)
  - Document enrichment: VectorDocBuilder now indexes service and database references from entities (with database extracted specifically from Table entities)
  - Fix certification filter: Corrected the certification filter path from certification.tagFQN to certification.tagLabel.tagFQN
  - DTO update: Added from field to the search request DTO for pagination support

----
## Summary by Gitar

- **Hybrid search support:** Refactored `VectorSearchQueryBuilder` to extract KNN query generation into reusable `buildQuery()` method for embedding inside hybrid search requests
- **Threshold-based scoring:** Added `min_score` support when threshold > 0 for relevance-based filtering instead of fixed `k` count
- **New search filters:** Implemented `service` and `database` filters using bool/should strategy matching both name and displayName fields
- **Vector index schema:** Added service and database object mappings across all locales (en, jp, ru, zh) to support new filtering
- **Pagination support:** Added `from` field to `VectorSearchRequest` DTO enabling result offset pagination
- **Embedded document enrichment:** Simplified `VectorDocBuilder` to generate embedding-specific fields merged via partial updates
- **Certification filter fix:** Corrected path from `certification.tagFQN` to `certification.tagLabel.tagFQN`
- **Test refactoring:** Simplified integration tests with single-index approach and explicit sleep instead of polling

<sub>This will update automatically on new commits.</sub>

PR #27111, opened 2026-04-07
feat: native Elasticsearch vector search support

### LINKED ISSUE #28259: Bootstrap fails on vector_search_index template creation with Elasticsearch (unknown setting index.knn)
### Affected module

Backend

### Describe the bug

When running the bootstrap script on a bare-metal installation (v1.12.6) with Elasticsearch as the backend, the process fails at the very end when trying to create/update the index template for `om_opmd-ts_vector_search_index`. 

The script injects OpenSearch-specific configurations (`index.knn`) into Elasticsearch, causing an `invalid_index_template_exception`. 
This happens even if `SEARCH_TYPE` is explicitly set to `elasticsearch` and `SEMANTIC_SEARCH_ENABLED` is set to `false`. The bootstrap script seems to ignore the semantic search disabled flag and attempts to create the vector indices with OpenSearch syntax anyway.

### To Reproduce

1. Setup a bare-metal installation of OpenMetadata 1.12.6 (tar.gz archive).
2. Configure `conf/openmetadata.yaml` to connect to a standard Elasticsearch instance (version 9.3.4).
3. Export the necessary environment variables for the Elasticsearch connection (e.g., in `conf/openmetadata-env.sh`) along with the specific flags for the search type and semantic search:
   ```bash
   export SEARCH_TYPE='elasticsearch'
   export SEMANTIC_SEARCH_ENABLED='false'
4. Run the bootstrap script:
   ```bash
   bootstrap/openmetadata-ops.sh drop-create
5. See the error at the end of the execution.

### Expected behavior

Since `SEMANTIC_SEARCH_ENABLED='false'`, the bootstrap process should entirely skip the creation of the vectorEmbedding index template. Alternatively, if it must create it, it should use the correct Elasticsearch mapping (e.g., dense_vector) instead of the OpenSearch index.knn plugin settings.

### OS

RHEL 9.7

### Python version

3.9.25

### OpenMetadata version

1.12.6

_No response_

### Additional context

```plaintext
Failed to create/update index template: om_opmd-ts_vector_search_index
es.co.elastic.clients.elasticsearch._types.ElasticsearchException: [es/indices.put_index_template] failed: [invalid_index_template_exception] index_template [om_opmd-ts_vector_search_index] invalid, cause [Validation Failed: 1: unknown setting [index.knn] please check that any required plugins are installed, or check the breaking changes documentation for removed settings;2: unknown setting [index.knn.algo_param.ef_search] please check that any required plugins are installed, or check the breaking changes documentation for removed settings;3: unknown setting [index.knn.advanced.filtered_exact_search_threshold] please check that any required plugins are installed, or check the breaking changes documentation for removed settings;]
        at es.co.elastic.clients.transport.ElasticsearchTransportBase.getApiResponse(ElasticsearchTransportBase.java:372)
        at es.co.elastic.clients.transport.ElasticsearchTransportBase.performRequest(ElasticsearchTransportBase.java:154)
        at es.co.elastic.clients.elasticsearch.indices.ElasticsearchIndicesClient.putIndexTemplate(ElasticsearchIndicesClient.java:3064)
        at es.co.elastic.clients.elasticsearch.indices.ElasticsearchIndicesClient.putIndexTemplate(ElasticsearchIndicesClient.java:3123)
        at org.openmetadata.service.search.elasticsearch.ElasticSearchGenericManager.createOrUpdateIndexTemplate(ElasticSearchGenericManager.java:122)
        at org.openmetadata.service.search.elasticsearch.ElasticSearchClient.createOrUpdateIndexTemplate(ElasticSearchClient.java:785)
        at org.openmetadata.service.search.SearchRepository.createOrUpdateIndexTemplates(SearchRepository.java:358)
        at org.openmetadata.service.util.OpenMetadataOperations.dropCreate(OpenMetadataOperations.java:958)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at picocli.CommandLine.executeUserObject(CommandLine.java:2070)
        at picocli.CommandLine.access$1500(CommandLine.java:148)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2465)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2457)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2419)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)
        at picocli.CommandLine$RunLast.execute(CommandLine.java:2421)
        at picocli.CommandLine.execute(CommandLine.java:2174)
        at org.openmetadata.service.util.OpenMetadataOperations.main(OpenMetadataOperations.java:3034)
Failed to create index template for vectorEmbedding: [es/indices.put_index_template] failed: [invalid_index_template_exception] index_template [om_opmd-ts_vector_search_index] invalid, cause [Validation Failed: 1: unknown setting [index.knn] please check that any required plugins are installed, or check the breaking changes documentation for removed settings;2: unknown setting [index.knn.algo_param.ef_search] please check that any required plugins are installed, or check the breaking changes documentation for removed settings;3: unknown setting [index.knn.advanced.filtered_exact_search_threshold] please check that any required plugins are installed, or check the breaking changes documentation for removed settings;]
Index templates creation completed. Success: 56, Failed: 1, Total: 57


### Pre-submission checklist

### PR DESCRIPTION
**Pull request by joaopamaral**
Fixes #28259

## Summary

Adds native Elasticsearch 9.x vector search support, mirroring the existing OpenSearch implementation. OpenMetadata deployments backed by Elasticsearch 9.x can now use the same semantic/vector search features as OpenSearch deployments.

## Architecture: ES vs OpenSearch Vector Search

```
                    ┌─────────────────────────────────────────────────────────────┐
                    │                   VectorIndexService (interface)             │
                    │  search() · updateEntityEmbedding() · getExistingFingerprint │
                    │  generateEmbeddingFields()                                   │
                    └───────────────┬───────────────────────┬───────────────────┘
                                    │                       │
               ┌────────────────────▼──────┐   ┌───────────▼──────────────────┐
               │  OpenSearchVectorService  │   │  ElasticSearchVectorService   │
               │  (existing, unchanged)    │   │  (new)                        │
               ├───────────────────────────┤   ├──────────────────────────────┤
               │ Transport: generic client  │   │ Transport: Rest5Client        │
               │ Query: query.knn (nested)  │   │ Query: top-level knn object   │
               │ Index field: knn_vector    │   │ Index field: dense_vector     │
               │                           │   │                               │
               │ Hybrid pipeline: ✅        │   │ Hybrid pipeline: ✗ (ES        │
               │  ensureHybridSearchPipeline│   │  doesn't support RRF pipeline │
               │  (OpenSearch RRF)          │   │  via search pipeline API)     │
               └───────────────────────────┘   └──────────────────────────────┘
                          │                                  │
               ┌──────────▼──────────┐           ┌──────────▼──────────┐
               │   OpenSearch index  │           │ Elasticsearch index  │
               │                     │           │                      │
               │ embedding:          │           │ embedding:           │
               │   type: knn_vector  │           │   type: dense_vector │
               │   dimension: N      │           │   dims: N            │
               │   method:           │           │   similarity: cosine │
               │     name: hnsw      │           │   (ES 9.x BBQ-HNSW)  │
               │     engine: lucene  │           │                      │
               │     space: cosine   │           │                      │
               └─────────────────────┘           └──────────────────────┘
```

### Query format difference

```
OpenSearch (query.knn — nested under "query")   Elasticsearch (top-level knn)
─────────────────────────────────────────────   ─────────────────────────────
{                                               {
  "query": {                                      "knn": {
    "knn": {                                        "field": "embedding",
      "embedding": {                                "query_vector": [...],
        "vector": [...],                            "k": 10,
        "k": 10,                                    "num_candidates": 100,
        "filter": { ... }                           "filter": { ... }
      }                                           }
    }                                           }
  }
}
```

### Ingestion flow

```
Entity created/updated
        │
        ▼
VectorEmbeddingHandler (lifecycle event)
        │
        ▼
VectorDocBuilder.buildEmbeddingFields()
  · textToLLMContext (concatenated text)
  · fingerprint (SHA-256, change detection)
  · embedding (float[] from EmbeddingClient)
  · parentId, chunkIndex, chunkCount
        │
        ├──── OpenSearch ──▶ OpenSearchVectorService.partialUpdateEntity()
        │                      POST /{index}/_update/{id}  (generic client)
        │
        └──── Elasticsearch ▶ ElasticSearchVectorService.partialUpdateEntity()
                               POST /{index}/_update/{id}  (Rest5Client)
```

### New
- **`ElasticSearchVectorService`**: ES implementation of `VectorIndexService`. Uses `Rest5Client` for all HTTP calls (same approach as the rest of the ES client layer). Mirrors `OpenSearchVectorService` for pagination, fingerprint deduplication, and embedding lifecycle.
- **`vector_search_index_es_native.json`** (en/jp/ru/zh): ES-native index mappings using `dense_vector` / `dims` / `cosine` similarity (ES 9.x format).
- **`SemanticSearchQueryBuilder`** (ES package): mirrors the OpenSearch equivalent.
- **`VectorSearchQueryBuilder.buildNativeESQuery()`**: emits the ES 9.x top-level `knn` format with configurable `num_candidates`.

### Modified
- **`ElasticSearchIndexManager.extractMappingsJson()`**: extracts the `mappings` sub-object before `putMapping` — ES rejects the full index JSON (with `settings`/`aliases`) at the mappings endpoint.
- **`EsUtils.enrichIndexMappingForElasticsearch()`**: injects `dense_vector` field into ES index mappings when vector search is enabled.
- **`SearchRepository`**: wires `ElasticSearchVectorService` for the `ELASTICSEARCH` search type; applies `enrichIndexMappingForElasticsearch` to index templates (OpenSearch path unchanged).
- **`ElasticSearchBulkSink`**: mirrors `OpenSearchBulkSink`'s embedding path — pre-fetches cached embeddings for state-matched entities (`fetchExistingEmbeddings`) and splices them (or regenerates) into the full `index` operation via `enrichWithEmbedding`, so stored vectors are never stripped.
- **`VectorSearchQueryBuilder`**: `reformatVectorIndexWithDimension()` handles both `"dims"` (ES) and `"dimension"` (OpenSearch).

### Bug fixes (post-review)
- **`appendKnnQuery` closing braces**: `}}}}` → `}}}` — the extra brace produced malformed JSON, causing OpenSearch to return 400 on every KNN request (silently yielding empty results).
- **`executeGenericRequest` HTTP status check**: now throws on 4xx/5xx instead of silently returning `{}`.
- **`extractRestClient` type guard**: replaces unchecked cast with `instanceof` pattern match; throws `IllegalArgumentException` with the actual transport class name on mismatch.
- **`knnNumCandidatesMultiplier`**: wired from `naturalLanguageSearch.knnNumCandidatesMultiplier` config; defaults to 2×.
- **`getFingerprint` endpoint**: gated to admin users.
- **Index template enrichment**: `createOrUpdateIndexTemplates()` and `createOrUpdateIndexTemplate()` now call `enrichIndexMappingForElasticsearch()` so templates include `dense_vector` when vector search is enabled.

### Review-feedback fixes (commit `9854029a`)
- **`SystemRepository.getEmbeddingsValidation()`**: removed the ES short-circuit. Elasticsearch now flows through the same checks (embedding generation + hybrid pipeline) as OpenSearch instead of always failing with the legacy "not supported" message.
- **`EsUtils.addDenseVectorSettings()`**: detects `_meta.embedding_dimension` drift vs. the active embedding client and emits a `WARN` log, mirroring `OsUtils.addKnnVectorSettings`. Helps surface stale indexes after a model switch (e.g. DJL 384 → OpenAI 1536).
- **`ElasticSearchBulkSink.isVectorEmbeddingEnabledForEntity`**: also gates on `searchRepository.getIndexMapping(entityType) != null` to skip entities whose mapping isn't loaded.
- **`ElasticSearchBulkSink.fetchExistingEmbeddings`** (supersedes `fetchExistingFingerprints`): two-step batch fetch of cached embeddings, routed to the pre-recreate (original) index during a staged recreate — the staged index is empty by definition — and to the canonical index otherwise.
- New unit tests for each of the above (in `EsUtilsTest`, `ElasticSearchBulkSinkBehaviorTest`, `SystemRepositoryEmbeddingsValidationTest`).

### Review-feedback fixes (round 2)
- **Embedding preservation on incremental reindex (critical)**: the fingerprint-skip previously returned the doc *without* embedding fields; since the bulk sink uses full `index` operations (document replacement), every unchanged entity lost its stored vector on a non-recreate reindex. Fixed by always splicing a cached-or-regenerated embedding into the doc (mirrors OpenSearch). Covered by new `enrichWithEmbedding*` unit tests and validated e2e on both engines.
- **kNN over-fetch clamp**: `overFetchSize` is now clamped to `k` unconditionally in the ES search loop — ES native kNN never returns more than `k` hits (threshold is client-side), and an unclamped `from+size` could exceed `index.max_result_window` and 400.
- **`ElasticSearchIndexManager.createIndex`**: re-throws `IllegalStateException` (embedding-dimension drift) instead of swallowing it, so bootstrap/reindex can't proceed against a broken index.
- **`VectorSearchResourceTest`**: the admin-gate test now asserts the authorizer is invoked and the search repository is never consulted (the previous `verify(never())` targeted an unwired mock and was vacuous).
- **Dead config removed**: `naturalLanguageSearch.embeddingProvider` and `naturalLanguageSearch.maxConcurrentRequests` had zero readers — the active embedding config lives in `llmConfiguration.embeddings`. Removed from the schema (+ regenerated TS) to avoid a silent misconfiguration trap. `knnNumCandidatesMultiplier` is kept (still read).
- **Merge-conflict containment**: `OpenSearchBulkSink`, `OpenSearchVectorService`, and `VectorIndexService` are kept byte-identical to their upstream + feature baseline; the ES path intentionally duplicates the OpenSearch pattern instead of sharing code, so upstream churn on the OS files merges cleanly. See the [review guide comment](https://github.com/open-metadata/OpenMetadata/pull/27111#issuecomment-4896553543) for exactly what differs.

## Test plan

- [x] `mvn test -pl openmetadata-service -Dtest=VectorSearchQueryBuilderTest,ElasticSearchIndexManagerTest,ElasticSearchVectorServiceTest,SearchRepositoryBehaviorTest,EsUtilsTest,ElasticSearchBulkSinkBehaviorTest,SystemRepositoryEmbeddingsValidationTest`
- [x] `mvn test -pl openmetadata-integration-tests -Dtest=VectorEmbeddingIntegrationIT` (uses Testcontainers OpenSearch 3.4.0)
- [x] **Local end-to-end on ES 9.3.0 + DJL all-MiniLM-L6-v2**: top-1 accuracy 3/3 across distinct semantic queries (`revenue from sales` → `customer_purchases` 0.76; `login authentication` → `user_logins` 0.71; `temperature humidity` → `weather_data` 0.69). Index has `dense_vector{dims:384, similarity:cosine, BBQ-HNSW}`. See [docs/elasticsearch-semantic-search-local-test.md](docs/elasticsearch-semantic-search-local-test.md).
- [x] **No regression on OpenSearch 3.4.0**: same test re-run with `SEARCH_TYPE=opensearch`. `OpenSearchVectorService` initializes, `hybrid-rrf` pipeline registered, `knn_vector` mapping created, semantic queries return identical scores to ES (same embedding model → deterministic cosine ranking).
- [x] **End-to-end re-verified after the review-feedback fixes** on both ES 9.3.0 and OpenSearch 3.4.0 — identical scores to the original run, no regression.
- [x] **Incremental (non-recreate) reindex preserves embeddings** on both engines: triggered `SearchIndexingApplication` without deleting the index; all 3/3 stored vectors survived and semantic ranking stayed identical (this reproduced — and now guards against — the embedding-wipe bug fixed in round 2).

## References

- [Elasticsearch k-nearest neighbor (kNN) search](https://www.elastic.co/docs/solutions/search/vector/knn)

----
## Summary by Gitar

- **Certification propagation:**
  - Added `cascadeCertificationToChildren` to push `AssetCertification` updates to denormalized child search documents like `tableColumn`.
  - Integrated `CASCADE_CERTIFICATION_SCRIPT` to handle certification changes via external handler in `SearchRepository`.
- **Google Embedding support:**
  - Added `GoogleEmbeddingClient` and corresponding `Google` configuration schema for vector embedding generation.
  - Enabled `google` provider in `SystemRepository` and `ElasticSearchConfiguration`.
- **Search index management:**
  - Added logic in `softDeleteOrRestoredChildren` to skip time-series indexes lacking a `deleted` field.
  - Updated `VectorSearchQueryBuilder` to support `primaryEntityId` filters.

Fixes https://github.com/open-metadata/OpenMetadata/issues/28259

This PR adds native Elasticsearch 9.x vector search support to OpenMetadata, mirroring the existing OpenSearch implementation. It introduces `ElasticSearchVectorService`, ES-native `dense_vector` index mappings, a `buildNativeESQuery` method for the ES top-level `knn` format, and embedding preservation during incremental reindexes — along with several pre-existing bug fixes (malformed kNN JSON brace, silent HTTP error swallowing, race condition in service init).

- **New `ElasticSearchVectorService`**: uses `Rest5Client` for raw HTTP, emits ES 9.x top-level `knn` queries with configurable `num_candidates`, and implements two-step cached-embedding batch fetch (`getExistingEmbeddingsBatch`) to avoid stripping stored vectors during full `index` bulk operations.
- **`EsUtils.enrichIndexMappingForElasticsearch`**: injects `dense_vector{dims, similarity:cosine}` into ES index mappings when vector search is enabled, with hard-fail (`IllegalStateException`) on embedding-dimension drift so operators must reindex rather than silently running mismatched indexes.
- **`ElasticSearchBulkSink`**: pre-fetches cached embeddings per batch and splices them (or regenerates) into every `index` operation, ensuring stored vectors survive non-recreate reindexes.
</details>

Safe to merge — the new ES vector path is well-isolated, embedding preservation is tested, and the critical pre-existing bugs are cleanly fixed.

The implementation closely mirrors the proven OpenSearch path, the embedding-wipe regression is covered by new unit tests and e2e validation on both ES 9.3 and OS 3.4, and the dimension-drift hard-fail surfaces misconfiguration at bootstrap. The two open observations are non-blocking hardening items that do not affect correctness at normal batch sizes.

ElasticSearchVectorService.fetchSourceByIds — the unbounded size parameter relies on upstream batch sizing staying below the ES index.max_result_window default.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java | New ES 9.x vector service implementation mirroring OpenSearch; solid overall but `fetchSourceByIds` uses `ids.size()` as the `size` parameter, which can breach `index.max_result_window` (default 10000) for large batches. |
| openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/ElasticSearchBulkSink.java | Adds embedding preservation via `enrichWithEmbedding` + `fetchExistingEmbeddings`; empty-batch guard present, dimension validation present, fallback to json-without-vector on error. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/EsUtils.java | Adds `enrichIndexMappingForElasticsearch` and `addDenseVectorSettings`; dimension-drift detection throws `IllegalStateException` surfaced cleanly to callers. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorIndexService.java | Interface extended with `getExistingFingerprint`, `getExistingFingerprintsBatch`, `executeGenericRequest`, and `getIndexAlias`; exposing `executeGenericRequest` (HTTP method/endpoint/body) as part of the public interface leaks HTTP transport semantics. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorSearchQueryBuilder.java | Adds `buildNativeESQuery` for ES top-level kNN format and refactors filter-building into shared `appendFilterMustClauses`; knn brace-count fix looks correct. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java | Wires `ElasticSearchVectorService` for the ES search type, applies `enrichForElasticsearch` to index templates, and adds `ensureHybridSearchPipeline` guard to skip ES (OpenSearch-only feature). |
| openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchIndexManager.java | Adds `extractMappingsJson` to pass only the `mappings` sub-object to `putMapping`, and re-throws `IllegalStateException` on dimension-drift instead of swallowing it. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/vector/OpenSearchVectorService.java | Adds `@Override` annotations, extracts `getIndexAlias` to the interface, adds `getExistingFingerprintsBatch`, and fixes init race where handler was registered after `instance` assignment. |
| openmetadata-service/src/main/java/org/openmetadata/service/resources/search/VectorSearchResource.java | Adds admin-gated `/fingerprint` GET endpoint and switches to `VectorIndexService` interface for the existing search endpoint. |
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SystemRepository.java | Removes the ES short-circuit that always failed with 'not supported'; ES now flows through the same embedding validation path as OpenSearch. |

</details>

<sub>Reviews (14): Last reviewed commit: ["Merge branch &#39;main&#39; into feat/es-elastic..."](https://github.com/open-metadata/openmetadata/commit/27b3deaccd6fc7a24851df9f7917ba0364f78a56) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37815621)</sub>

### CODE REVIEW

**Comment by joaopamaral** [comment]
Initial results look good, but I've run a test only with ES 9.x and version 1.12.4 (not the one from main). I also need to double-check if OpenSearch is affected by this change. Also need to review some AI-resolved conflicts from version 1.12.4 with main.

**Comment by harshach** [comment]
Thanks @joaopamaral this is great!!. Can you make it ready for review? and also address comments here https://github.com/open-metadata/OpenMetadata/pull/27111#issuecomment-4195678501

**Comment by joaopamaral** [comment]
Sure @harshach! I'll work on the bot review first before making it ready for review! 👍

File: openmetadata-service/src/test/java/org/openmetadata/service/search/vector/VectorSearchQueryBuilderTest.java [hunk 1]
```diff
@@ -698,9 +698,187 @@ void testIgnoresOnlyUnrecognizedFilterKeys() throws Exception {
...
+  @Test
+  void testNativeESQueryAndOpenSearchQueryProduceSameFilters() throws Exception {
+    float[] vector = {0.1f, 0.2f};
+    Map<String, List<String>> filters =
+        Map.of(
+            "entityType", List.of("table"),
+            "owners", List.of("alice"),
+            "tier", List.of("Tier.Gold"));
+
+    String osQuery = VectorSearchQueryBuilder.build(vector, 10, 100, filters);
```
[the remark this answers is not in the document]
joaopamaral: https://github.com/open-metadata/OpenMetadata/pull/27111/changes/4c34b8c91e11b72b7694ad0cf8c123571d49471b

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,543 @@
...
+    try (InputStream inputStream = getClass().getClassLoader().getResourceAsStream(resourcePath)) {
+      if (inputStream == null) {
+        throw new IllegalStateException("Could not find " + resourcePath + " in classpath");
+      }
+      String template = new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
+      String result = template.replace("\"dims\": 512", "\"dims\": " + dimension);
+      if (result.equals(template) && dimension != 512) {
+        throw new IllegalStateException(
+            "Failed to replace dimension placeholder in vector index mapping template");
+      }
```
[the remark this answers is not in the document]
joaopamaral: Fixed in https://github.com/open-metadata/OpenMetadata/pull/27111/changes/4a2cdc51b9d31a2cf99a6d3df48395a63f543b31

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,543 @@
...
+    instance = new ElasticSearchVectorService(client, embeddingClient, language);
+    instance.registerVectorEmbeddingHandler();
+    LOG.info(
+        "ElasticSearchVectorService initialized with model={}, dimension={}",
+        embeddingClient.getModelId(),
+        embeddingClient.getDimension());
+  }
+
+  public static ElasticSearchVectorService getInstance() {
+    return instance;
```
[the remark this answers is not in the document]
joaopamaral: https://github.com/open-metadata/OpenMetadata/pull/27111/changes/8b9c5910fdabfe8d95d482492484a32afbaef677

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,543 @@
...
+  }
+
+  public ElasticSearchVectorService(ElasticsearchClient client, EmbeddingClient embeddingClient) {
+    this(client, embeddingClient, "en");
+  }
+
+  private static Rest5Client extractRestClient(ElasticsearchClient client) {
+    Rest5ClientTransport transport = (Rest5ClientTransport) client._transport();
+    return transport.restClient();
+  }
```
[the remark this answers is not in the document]
joaopamaral: I don't see it changing in the future.

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,547 @@
...
+      int k,
+      double threshold) {
+    long start = System.currentTimeMillis();
+    try {
+      float[] queryVector = embeddingClient.embed(query);
+      int overFetchSize = size * OVER_FETCH_MULTIPLIER;
+
+      String queryJson =
+          VectorSearchQueryBuilder.buildNativeESQuery(queryVector, overFetchSize, from, k, filters);
+      String indexName = getIndexAlias();
```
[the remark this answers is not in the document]
joaopamaral: https://github.com/open-metadata/OpenMetadata/pull/27111/changes/b54e9d5ead7dd1ac77a8a1d1b760f5d9f7ff6247

joaopamaral:
Hi @harshach, ’ve addressed the bot review, but I still need to re-review the code after rebasing/merging with main and rerun the tests against a real server. So far, I’ve tested this [PR](https://github.com/Automattic/OpenMetadata/pull/1)￼ with version 1.12.4 and ES 9.3.1. I still need to validate that everything continues to work correctly with OpenSearch and ES 8.x.

I won’t be able to run tests for the next couple of days, but feel free to proceed with any testing on your side in the meantime.

joaopamaral:
Also need to review all after this refactor https://github.com/open-metadata/OpenMetadata/pull/26000 😢

harshach:
@joaopamaral thanks for your work on this, can you check the co-pilot comments and address the merge conflict here please

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java
```diff
@@ -588,10 +590,14 @@ public void deleteIndex(IndexMapping indexMapping) {
   }
 
   private String getIndexMapping(IndexMapping indexMapping) {
+    String mappingFile = indexMapping.getIndexMappingFile();
+    boolean isOpenSearch = getSearchType() == ElasticSearchConfiguration.SearchType.OPENSEARCH;
+    if (!isOpenSearch && mappingFile != null && mappingFile.contains("vector_search_index.json")) {
+      mappingFile =
+          mappingFile.replace("vector_search_index.json", "vector_search_index_es_native.json");
```
mohityadav766: Filename-substring swaps tend to rot. If either file is renamed, or a third variant is added, this silently no-ops. Can we use an explicit per-backend lookup (or a constant on `IndexMapping`) instead of string replace?

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java
```diff
@@ -3050,8 +3068,15 @@ private String reformatVectorIndexWithDimension(String mapping, int dimension) {
...
-      LOG.warn("Failed to set embedding _meta in mapping JSON", e);
-      return mapping;
+      LOG.warn(
+          "Failed to parse mapping JSON for dimension patching, falling back to string replace");
+      return mapping
+          .replace("\"dimension\": 768", "\"dimension\": " + dimension)
+          .replace("\"dimension\":768", "\"dimension\":" + dimension)
+          .replace("\"dimension\": 512", "\"dimension\": " + dimension)
+          .replace("\"dimension\":512", "\"dimension\":" + dimension)
+          .replace("\"dims\": 512", "\"dims\": " + dimension)
```
mohityadav766: This fallback is exactly the brittle string-replace pattern `4a2cdc51b9` cleaned up in `loadIndexMapping`. If the JSON parse fails, the mapping is already broken — swallowing the error and doing string surgery on a 512/768 literal is worse than failing loudly. Prefer rethrowing.

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,603 @@
...
+  private final Rest5Client restClient;
+  @Getter private final EmbeddingClient embeddingClient;
+  private final String language;
+
+  public ElasticSearchVectorService(
+      ElasticsearchClient client, EmbeddingClient embeddingClient, String language) {
+    this.client = client;
+    this.restClient = extractRestClient(client);
+    this.embeddingClient = embeddingClient;
+    this.language = language != null ? language.toLowerCase() : "en";
```
mohityadav766: `toLowerCase()` without `Locale.ROOT` — breaks on Turkish locale (the `I` → `ı` trap). Project convention is `toLowerCase(Locale.ROOT)`.

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,603 @@
...
+
+  private static Rest5Client extractRestClient(ElasticsearchClient client) {
+    Rest5ClientTransport transport = (Rest5ClientTransport) client._transport();
+    return transport.restClient();
+  }
+
+  public static synchronized void init(
+      ElasticsearchClient client, EmbeddingClient embeddingClient, String language) {
+    if (instance != null) {
+      LOG.warn("ElasticSearchVectorService already initialized, reinitializing");
```
mohityadav766, line 67: On re-init we log a warning but never deregister the previous handler from `EntityLifecycleEventDispatcher`. If hot reconfig happens, every entity event gets processed twice (then 3x…). Same pre-existing bug in `OpenSearchVectorService.init` — worth fixing in both while you're here.

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,603 @@
...
+          "Failed to delete by parent_id={} in index={}: {}",
+          parentId,
+          indexName,
+          e.getMessage(),
+          e);
+    }
+  }
+
+
+  public void createOrUpdateIndex(int dimension) {
```
mohityadav766: Name is misleading — this only creates, it early-returns if the index exists and never updates the mapping. Rename to `ensureIndex` / `createIfMissing`, or actually handle the update path (at least when `dimension` differs).

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,603 @@
...
+                  builder.settings(settings);
+                }
+
+                return builder;
+              });
+      client.indices().create(request);
+
+      LOG.info("Created vector index {} with dimension {}", getIndexAlias(), dimension);
+    } catch (Exception e) {
+      LOG.error("Failed to create vector index: {}", e.getMessage(), e);
```
mohityadav766: Swallowing the exception here means later `search`/`bulkIndex` calls will fail with a cryptic `index_not_found` and nobody will know the create step actually failed at startup. Rethrow (or mark the service unhealthy) so the root cause surfaces immediately.

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,603 @@
...
+  }
+
+  static String patchDimension(String mappingJson, int dimension) {
+    try {
+      JsonNode root = MAPPER.readTree(mappingJson);
+      JsonNode properties = root.path("mappings").path("properties");
+      if (!properties.isMissingNode() && properties.has("embedding")) {
+        ObjectNode embeddingNode = (ObjectNode) properties.get("embedding");
+        if (embeddingNode.has("dims")) {
+          embeddingNode.put("dims", dimension);
```
mohityadav766: If the template ever drops the `dims` key this silently no-ops and ES rejects the index create later. Worth throwing when `embedding.dims` is missing — makes template regressions loud.

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorSearchQueryBuilder.java
```diff
@@ -66,17 +66,55 @@ private static void appendKnnQuery(
...
+    sb.append("]}}"); // close must array and bool
 
-    // Only include documents where deleted=false
-    sb.append("{\"term\":{\"deleted\":false}}");
+    sb.append("}}}}"); // close embedding, knn, query
+  }
+
+  public static String buildNativeESQuery(
+      float[] vector, int size, int from, int k, Map<String, List<String>> filters) {
+    int numCandidates = Math.max(k, 100);
```
mohityadav766: `max(k, 100)` under-explores for larger `k`. ES docs recommend `num_candidates` be comfortably larger than `k` (typically 10–50×) for good recall. Even `max(k * 2, 100)` would be safer. Ideally make it configurable via the NL search config.

openmetadata-service/src/main/java/org/openmetadata/service/resources/search/VectorSearchResource.java
```diff
@@ -97,4 +102,57 @@ public Response vectorSearchPost(
...
+  @Path("/fingerprint")
+  @Operation(
+      operationId = "getFingerprint",
+      summary = "Get vector fingerprint",
+      description = "Returns the existing fingerprint for a given entity.")
+  public Response getFingerprint(
+      @Context SecurityContext securityContext,
+      @Parameter(description = "Parent entity ID", required = true) @QueryParam("parentId")
+          String parentId) {
+    DefaultAuthorizer.getSubjectContext(securityContext);
```
mohityadav766: `DefaultAuthorizer.getSubjectContext` only extracts the subject, it doesn't authorize. Any authenticated user can enumerate fingerprints for arbitrary entity UUIDs. Intentional for a diagnostic endpoint, or should this be admin-gated?

openmetadata-service/src/test/java/org/openmetadata/service/apps/bundles/searchIndex/ElasticSearchBulkSinkSimpleTest.java
```diff
@@ -80,4 +81,17 @@ void testContextDataHandling() {
...
+  @Test
+  void testIsVectorEmbeddingEnabledForEntity() {
+    assertEquals(false, elasticSearchBulkSink.isVectorEmbeddingEnabledForEntity("table"));
+    assertEquals(false, elasticSearchBulkSink.isVectorEmbeddingEnabledForEntity("user"));
+    assertEquals(false, elasticSearchBulkSink.isVectorEmbeddingEnabledForEntity("dashboard"));
+  }
+
+  @Test
+  void testAddEntitiesToVectorIndexBatch() {
+    elasticSearchBulkSink.addEntitiesToVectorIndexBatch(
```
mohityadav766: Nit — wrap in `assertDoesNotThrow(...)` so the intent ("null-safety smoke test") is explicit. As written it reads like an assertion was forgotten.

mohityadav766:
Branch looks stale against main — missing #27388 (payload-size fix, merged Apr 15). GitHub's PR diff doesn't flag these spots because they're unchanged against the merge-base, but on rebase/merge they'll silently regress:

- `ElasticSearchBulkSink.java` line 322 — `BULK_OPERATION_METADATA_OVERHEAD` snaps back from 150→50 and the `indexDocumentDirectly` fallback + `stripLineageForSize` call disappear.
- `OpenSearchBulkSink.java` — same revert leaks in (-84 lines of payload-size protection) even though this PR shouldn't touch OS at all.
- `SearchRepository#indexEntityReferenceAndRelatedEntities` — removes the `stripDocMapIfOversized` call.
- `SearchRepository.java` line 1251 — `maxPayloadSizeBytes` becomes hardcoded `10 * 1024 * 1024` instead of `SearchClusterMetrics.DEFAULT_BULK_PAYLOAD_SIZE_BYTES`.
- `ElasticSearchIndexManager#listIndicesByPrefix` / `#getAllIndexStats` — the `buildScopedPattern` helper disappears, so both fall back to `*` / `prefix + "*"` without cluster-alias scoping. The two scoping tests removed from `ElasticSearchIndexManagerTest` covered exactly this — on a shared cluster with a non-empty `clusterAlias`, we'd start matching every index in the cluster, not just ours.

Can you rebase onto main first? After that, most of these should drop out of the diff and the remaining inline comments are the actual vector-specific concerns.

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/ElasticSearchBulkSink.java
```diff
@@ -670,6 +699,166 @@ public void updateConcurrentRequests(int concurrentRequests) {
...
+      return false;
+    } catch (TimeoutException e) {
+      LOG.warn("Timeout waiting for vector completion after {}s", timeoutSeconds);
+      return false;
+    }
+  }
+
+  @Override
+  public int getPendingVectorTaskCount() {
+    return Math.max(0, phaser.getUnarrivedParties() - 1);
```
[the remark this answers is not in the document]
joaopamaral: Fixed, Phaser/vectorExecutor removed

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,603 @@
...
+
+      String query =
+          "{\"size\":"
+              + parentIds.size()
+              + ",\"_source\":[\"parent_id\",\"fingerprint\"]"
+              + ",\"query\":{\"terms\":{\"parent_id\":"
+              + termsArray
+              + "}}"
+              + ",\"collapse\":{\"field\":\"parent_id\"}}";
+
```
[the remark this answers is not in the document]
joaopamaral: Fixed, uses _id / ids query

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,603 @@
...
+      if (currentFingerprint.equals(existingFingerprint)) {
+        LOG.debug("Skipping entity {} - fingerprint unchanged", parentId);
+        return;
+      }
+
+      List<Map<String, Object>> docs = VectorDocBuilder.fromEntity(entity, embeddingClient);
+      deleteByParentId(targetIndex, parentId);
+      bulkIndex(docs, targetIndex);
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
joaopamaral: Fixed, uses partialUpdateEntity on entity index

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorSearchQueryBuilder.java
```diff
@@ -66,17 +66,55 @@ private static void appendKnnQuery(
...
+            .append(Arrays.toString(vector))
+            .append(",\"k\":")
+            .append(k)
+            .append(",\"num_candidates\":")
+            .append(numCandidates);
 
-    // Then add user-specified filters
+    sb.append(",\"filter\":{\"bool\":{\"must\":[");
+    appendFilterMustClauses(sb, filters, true);
+    sb.append("]}}"); // close must array and bool
```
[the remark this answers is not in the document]
joaopamaral: Fixed, entity indices are now the only target; need to change nestedTags to false (still passes true)

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java
joaopamaral: string-swap removed

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java
joaopamaral: method deleted after move to inline definition

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
joaopamaral: fixed

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
joaopamaral: unregisterHandler call added

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
joaopamaral: method deleted; index creation in ElasticSearchIndexManager

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
joaopamaral: method deleted

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
joaopamaral: template approach deleted

openmetadata-service/src/test/java/org/openmetadata/service/apps/bundles/searchIndex/ElasticSearchBulkSinkSimpleTest.java
joaopamaral: test removed (method deleted)

joaopamaral:
Code is compiling and search vector tests are passing. I'll continue to address the comments this weekend and test with ES service running.

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorSearchQueryBuilder.java
joaopamaral: Made it configurable in https://github.com/open-metadata/OpenMetadata/pull/27111/changes/4ee2a937b79a5391c1a1ea3f70167ea96054d1a2

openmetadata-service/src/main/java/org/openmetadata/service/resources/search/VectorSearchResource.java
joaopamaral: Make sense make it admin-gated (users shouldn't have access to this endpoint). Let me make it accessible only by admin.

joaopamaral: Done in https://github.com/open-metadata/OpenMetadata/pull/27111/commits/0bf1a6ea048a9b0c746690bfd892a6b7af23a507

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,603 @@
...
+
+  @Override
+  public String getExistingFingerprint(String indexName, String parentId) {
+    try {
+      String query =
+          "{\"size\":1,\"_source\":[\"fingerprint\"],"
+              + "\"query\":{\"term\":{\"parent_id\":\""
+              + VectorSearchQueryBuilder.escape(parentId)
+              + "\"}}}";
+      String response = executeGenericRequest("POST", "/" + indexName + "/_search", query);
```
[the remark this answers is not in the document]
joaopamaral: Fixed — uses {"term":{"_id":...}}

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorSearchQueryBuilder.java
```diff
@@ -126,10 +168,35 @@ private static void appendKnnQuery(
...
+    sb.append("}}");
+  }
 
-    sb.append("}}}"); // close embedding, knn, wrapper
+  private static void appendOneNestedQuery(StringBuilder sb, String field, String val) {
+    switch (val) {
+      case ANY -> sb.append("{\"exists\":{\"field\":\"").append(field).append("\"}}");
+      case NONE -> sb.append("{\"bool\":{\"must_not\":{\"exists\":{\"field\":\"")
+          .append(field)
+          .append("\"}}}}");
```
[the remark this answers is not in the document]
joaopamaral: Fixed — nested tags path removed entirely

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java
```diff
@@ -601,11 +602,7 @@ private String getIndexMapping(IndexMapping indexMapping) {
...
   public String readIndexMapping(IndexMapping indexMapping) {
-    String mapping = getIndexMapping(indexMapping);
-    if (isVectorEmbeddingEnabled() && embeddingClient != null && mapping != null) {
-      mapping = reformatVectorIndexWithDimension(mapping, embeddingClient.getDimension());
-    }
-    return mapping;
+    return getIndexMapping(indexMapping);
   }
 
   /**
```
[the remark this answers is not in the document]
joaopamaral: fixed

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/ElasticSearchBulkSink.java
```diff
@@ -749,6 +783,83 @@ public void updateConcurrentRequests(int concurrentRequests) {
     LOG.info("Concurrent requests updated to: {}", concurrentRequests);
   }
 
+  boolean isVectorEmbeddingEnabledForEntity(String entityType) {
```
lautel, line 917: Missing a check on `searchRepository.getIndexMapping(entityType) != null`, otherwise it may attempt entities whose mapping isn't loaded

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/ElasticSearchBulkSink.java
```diff
@@ -749,6 +783,83 @@ public void updateConcurrentRequests(int concurrentRequests) {
...
+          "Failed to generate embeddings for entity {}: {}", entity.getId(), e.getMessage(), e);
+      vectorFailed.incrementAndGet();
+      if (tracker != null) {
+        tracker.recordVector(StatsResult.FAILED);
+      }
+      return json;
+    }
+  }
+
+  private Map<String, String> fetchExistingFingerprints(
```
lautel: Not sure why this method drops`reindexContext` entirely. This makes that during a staged reindex the fingerprint dedup could read the wrong index, think everything needs to be re-processed, and end up redoing all the work it was supposed to skip. Please check how `fetchExistingFingerprints` works in `OpenSearchBulkSink.java`

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/EsUtils.java
```diff
@@ -524,4 +524,54 @@ private static void buildSearchSourceFilter(
...
+
+    org.openmetadata.service.search.SearchRepository searchRepository =
+        org.openmetadata.service.Entity.getSearchRepository();
+    if (searchRepository == null
+        || !searchRepository.isVectorEmbeddingEnabled()
+        || searchRepository.getEmbeddingClient() == null) {
+      return;
+    }
+
+    int dimension = searchRepository.getEmbeddingClient().getDimension();
```
lautel, line 572: Missing a check on dimension mismatch and warning log. Please check `addKnnVectorSettings` in `OsUtils.java` for reference.

lautel (changes requested):
Heads up — not in this PR's diff, but `SystemRepository.getEmbeddingsValidation()` still short-circuits ES with "Elasticsearch is not supported for Semantic Search embeddings. Please use OpenSearch." Now that this PR makes ES vector search work, that validation will incorrectly fail for ES users who have embeddings configured. Worth addressing it in this PR to make it effective.

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/ElasticSearchBulkSink.java
joaopamaral: Added the guard, mirroring OS sink

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/ElasticSearchBulkSink.java
joaopamaral: Added ReindexContext parameter and staged-index routing, mirroring OS sink

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/EsUtils.java
joaopamaral: Added _meta.embedding_dimension mismatch check + WARN log

joaopamaral:
Thanks @lautel, I've addressed your comments and removed getEmbeddingsValidation early-return; ES flows through the same validation as OS.

joaopamaral:
Thansk for the reviews @mohityadav766 and @lautel! I've made an e2e test with local docker to validate if ES9.x and OpenSearch semantic search are working. Here is the test with steps and result (with AI help):

# Local end-to-end test: Elasticsearch + semantic search (DJL)

End-to-end verification that the native Elasticsearch vector search path works
on a fresh local stack with a DJL embedding provider.

## Stack

- **Elasticsearch** 9.3.0 (single-node, no security)
- **OpenMetadata server** (current branch build)
- **MySQL** 8.x
- **Embedding provider**: DJL `sentence-transformers/all-MiniLM-L6-v2` (384 dims, cosine)

## Prerequisites

- Docker daemon running
- The dev `docker-compose.yml` exposes the natural-language / semantic-search env
  vars and they default to `true` for the dev stack:

```yaml
NATURAL_LANGUAGE_SEARCH_ENABLED: ${NATURAL_LANGUAGE_SEARCH_ENABLED:-true}
SEMANTIC_SEARCH_ENABLED: ${SEMANTIC_SEARCH_ENABLED:-true}
EMBEDDING_PROVIDER: ${EMBEDDING_PROVIDER:-djl}
DJL_EMBEDDING_MODEL: ${DJL_EMBEDDING_MODEL:-"ai.djl.huggingface.pytorch/sentence-transformers/all-MiniLM-L6-v2"}
```

- Set the admin password as an env var so it stays out of the doc:

```bash
export OM_ADMIN_PASSWORD_B64="<base64-encoded admin password>"
```

## Steps

### 1. Bring up the stack

```bash
docker compose -f docker/development/docker-compose.yml up -d --build
```

Wait for `openmetadata_server` to become healthy:

```bash
docker compose -f docker/development/docker-compose.yml ps
```

### 2. Confirm the vector service initialized

```bash
docker logs openmetadata_server 2>&1 \
  | grep -E "DjlEmbeddingClient|ElasticSearchVectorService|Vector search service initialized"
```

Expected:

```
INFO  o.o.s.s.v.c.DjlEmbeddingClient - Initialized DjlEmbeddingClient with model=ai.djl.huggingface.pytorch/sentence-transformers/all-MiniLM-L6-v2, dimension=384
INFO  o.o.s.s.v.ElasticSearchVectorService - ElasticSearchVectorService initialized with model=...all-MiniLM-L6-v2, dimension=384
INFO  o.o.s.s.SearchRepository - Vector search service initialized with provider=djl, dimension=384
```

### 3. Recreate the table index so it picks up the `dense_vector` mapping

Drop the table index and trigger the Search Indexing app to recreate it:

```bash
JWT=$(curl -s -X POST "http://localhost:8585/api/v1/users/login" \
  -H "Content-Type: application/json" \
  -d "{\"email\":\"admin@open-metadata.org\",\"password\":\"$OM_ADMIN_PASSWORD_B64\"}" \
  | jq -r '.accessToken')

curl -s -X DELETE "http://localhost:9200/openmetadata_table_search_index"
# {"acknowledged":true}

curl -s -X POST "http://localhost:8585/api/v1/apps/deploy/SearchIndexingApplication" \
  -H "Authorization: Bearer $JWT" -w "\nHTTP %{http_code}\n"
# Application Deployed
# HTTP 200

curl -s -X POST "http://localhost:8585/api/v1/apps/trigger/SearchIndexingApplication" \
  -H "Authorization: Bearer $JWT" -w "\nHTTP %{http_code}\n"
# HTTP 200
```

Verify the recreated index has `embedding` typed as `dense_vector`:

```bash
curl -s "http://localhost:9200/openmetadata_table_search_index/_mapping" \
  | jq '.[].mappings.properties.embedding'
```

Output:

```json
{
  "type": "dense_vector",
  "dims": 384,
  "index": true,
  "similarity": "cosine",
  "index_options": {
    "type": "bbq_hnsw",
    "m": 16,
    "ef_construction": 100,
    "rescore_vector": { "oversample": 3.0 }
  }
}
```

### 4. Create test entities with semantically distinct descriptions

```bash
# Database service
curl -s -X POST "http://localhost:8585/api/v1/services/databaseServices" \
  -H "Authorization: Bearer $JWT" -H "Content-Type: application/json" -d '{
    "name":"semantic_test_svc",
    "serviceType":"Mysql",
    "connection":{"config":{"type":"Mysql","scheme":"mysql+pymysql","username":"u","authType":{"password":"p"},"hostPort":"localhost:3306"}}
  }'

# Database
curl -s -X POST "http://localhost:8585/api/v1/databases" \
  -H "Authorization: Bearer $JWT" -H "Content-Type: application/json" \
  -d '{"name":"test_db","service":"semantic_test_svc"}'

# Schema
curl -s -X POST "http://localhost:8585/api/v1/databaseSchemas" \
  -H "Authorization: Bearer $JWT" -H "Content-Type: application/json" \
  -d '{"name":"public","database":"semantic_test_svc.test_db"}'

# Three semantically distinct tables
for t in \
  "customer_purchases:Records of customer transactions and revenue from product sales" \
  "user_logins:User authentication events and session tracking data" \
  "weather_data:Daily atmospheric measurements including temperature and humidity"; do
    NAME="${t%%:*}"; DESC="${t#*:}"
    curl -s -X POST "http://localhost:8585/api/v1/tables" \
      -H "Authorization: Bearer $JWT" -H "Content-Type: application/json" \
      -d "{\"name\":\"$NAME\",\"databaseSchema\":\"semantic_test_svc.test_db.public\",\"description\":\"$DESC\",\"columns\":[{\"name\":\"id\",\"dataType\":\"INT\"}]}" \
      | jq -r '"Created: " + .name'
done
# Created: customer_purchases
# Created: user_logins
# Created: weather_data
```

### 5. Wait for embeddings to be generated and indexed

`VectorEmbeddingHandler` runs on the entity-lifecycle event bus and writes
embeddings via partial update. Poll until all three tables have an `embedding`
field:

```bash
until curl -s "http://localhost:9200/openmetadata_table_search_index/_search" \
  -H "Content-Type: application/json" \
  -d '{"size":0,"query":{"bool":{"must":[{"exists":{"field":"embedding"}}]}}}' \
  | jq -e '.hits.total.value >= 3' >/dev/null; do sleep 2; done
echo "Embeddings ready."

# Inspect what was indexed:
curl -s "http://localhost:9200/openmetadata_table_search_index/_search" \
  -H "Content-Type: application/json" \
  -d '{"_source":["name","description"],"query":{"bool":{"must":[{"exists":{"field":"embedding"}},{"term":{"databaseSchema.fullyQualifiedName":"semantic_test_svc.test_db.public"}}]}}}' \
  | jq '.hits.hits[]._source'
```

Output:

```json
{ "name": "customer_purchases", "description": "Records of customer transactions and revenue from product sales" }
{ "name": "user_logins",        "description": "User authentication events and session tracking data" }
{ "name": "weather_data",       "description": "Daily atmospheric measurements including temperature and humidity" }
```

### 6. Run semantic search queries

```bash
for query in "revenue from sales" "login authentication" "temperature humidity"; do
  echo "=== Query: \"$query\" ==="
  curl -s -X POST "http://localhost:8585/api/v1/search/vector/query" \
    -H "Authorization: Bearer $JWT" -H "Content-Type: application/json" \
    -d "{\"query\":\"$query\",\"size\":3,\"k\":10,\"threshold\":0.0,\"filters\":{\"entityType\":[\"table\"]}}" \
  | jq '.hits | map({name, score: ._score})'
done
```

Output:

```
=== Query: "revenue from sales" ===
[
  { "name": "customer_purchases", "score": 0.75822616 },
  { "name": "weather_data",       "score": 0.5282976  },
  { "name": "user_logins",        "score": 0.51876247 }
]

=== Query: "login authentication" ===
[
  { "name": "user_logins",        "score": 0.7053669  },
  { "name": "customer_purchases", "score": 0.52392066 },
  { "name": "weather_data",       "score": 0.5139432  }
]

=== Query: "temperature humidity" ===
[
  { "name": "weather_data",       "score": 0.69338286 },
  { "name": "user_logins",        "score": 0.5082303  },
  { "name": "customer_purchases", "score": 0.4560685  }
]
```

## Outcome

Each query ranks the semantically matching table first, with a clear score
margin (~0.70) over the unrelated tables (~0.50):

| Query | Top hit (score) | 2nd | 3rd |
|---|---|---|---|
| `revenue from sales` | **customer_purchases** (0.76) | weather_data (0.53) | user_logins (0.52) |
| `login authentication` | **user_logins** (0.71) | customer_purchases (0.52) | weather_data (0.51) |
| `temperature humidity` | **weather_data** (0.69) | user_logins (0.51) | customer_purchases (0.46) |

This verifies end-to-end:

1. **`enrichForElasticsearch`** injects `dense_vector` into the table index
   (`dims=384`, `cosine`, BBQ-HNSW).
2. **`VectorEmbeddingHandler`** generates embeddings on entity create and
   writes them via partial update.
3. **`POST /api/v1/search/vector/query`** routes to
   `ElasticSearchVectorService.search()` and uses
   `VectorSearchQueryBuilder.buildNativeESQuery()` (top-level `knn` object,
   ES 8.x/9.x format).
4. KNN with cosine similarity returns ranked hits — the correct table is
   always #1.

## Cross-check: same test against OpenSearch 3.4.0

To confirm the PR doesn't regress the existing OpenSearch path, the exact same
steps were re-run with the dev compose pointed at
`opensearchproject/opensearch:3.4.0` and `SEARCH_TYPE=opensearch`.

Differences in what the system produces:

- **Service class**: logs show `OpenSearchVectorService initialized ...`
  instead of `ElasticSearchVectorService initialized ...`
- **Hybrid pipeline**: OpenSearch logs `Hybrid search pipeline 'hybrid-rrf'
  created/updated with weights keyword=0.4, semantic=0.6` (only OS supports
  this — ES is silently skipped).
- **Index field type**: `embedding` is `knn_vector` (not `dense_vector`):

```bash
curl -s "http://localhost:9200/openmetadata_table_search_index/_mapping" \
  | jq '.[].mappings.properties.embedding'
```

```json
{
  "type": "knn_vector",
  "dimension": 384,
  "method": {
    "engine": "lucene",
    "space_type": "cosinesimil",
    "name": "hnsw",
    "parameters": { "ef_construction": 256, "m": 48 }
  }
}
```

Semantic queries — same three queries, identical scores (same embedding model
→ identical query vectors → cosine ranking is deterministic across both
backends):

```
=== Query: "revenue from sales" ===
[
  { "name": "customer_purchases", "score": 0.7582263  },
  { "name": "weather_data",       "score": 0.5282976  },
  { "name": "user_logins",        "score": 0.51876247 }
]

=== Query: "login authentication" ===
[
  { "name": "user_logins",        "score": 0.7053669  },
  { "name": "customer_purchases", "score": 0.5239207  },
  { "name": "weather_data",       "score": 0.5139432  }
]

=== Query: "temperature humidity" ===
[
  { "name": "weather_data",       "score": 0.69338286 },
  { "name": "user_logins",        "score": 0.5082303  },
  { "name": "customer_purchases", "score": 0.45606846 }
]
```

| Query | ES 9.3 top-1 (score) | OS 3.4 top-1 (score) |
|---|---|---|
| `revenue from sales` | customer_purchases (0.7582) | customer_purchases (0.7582) |
| `login authentication` | user_logins (0.7054) | user_logins (0.7054) |
| `temperature humidity` | weather_data (0.6934) | weather_data (0.6934) |

Both backends return the correct table as #1 with matching scores. The PR
does not regress the existing OpenSearch path.

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/EsUtils.java
```diff
@@ -524,4 +524,67 @@ private static void buildSearchSourceFilter(
...
+            dimension);
+      }
+    }
+
+    com.fasterxml.jackson.databind.node.ObjectNode embeddingNode = mapper.createObjectNode();
+    embeddingNode.put("type", "dense_vector");
+    embeddingNode.put("dims", dimension);
+    embeddingNode.put("index", true);
+    embeddingNode.put("similarity", "cosine");
+    ((com.fasterxml.jackson.databind.node.ObjectNode) properties).set("embedding", embeddingNode);
```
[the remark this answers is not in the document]
lautel: @joaopamaral please review this copilot's comment

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/EsUtils.java
```diff
@@ -524,4 +524,67 @@ private static void buildSearchSourceFilter(
...
+
+    JsonNode mappings = rootNode.path("mappings");
+    if (!mappings.isMissingNode()) {
+      com.fasterxml.jackson.databind.node.ObjectNode metaNode =
+          ((com.fasterxml.jackson.databind.node.ObjectNode) mappings).putObject("_meta");
+      metaNode
+          .put("embedding_model", searchRepository.getEmbeddingClient().getModelId())
+          .put("embedding_dimension", dimension);
```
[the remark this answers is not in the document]
lautel: @joaopamaral also this one seems relevant

lautel (commented):
@joaopamaral looks good! Thanks for addressing my comments. I think there're still 2 Copilot's comments worth checking. Also, please note that there's a merge conflict at `ElasticSearchBulkSink.java`

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/EsUtils.java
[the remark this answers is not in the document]
joaopamaral: Addressed in https://github.com/open-metadata/OpenMetadata/pull/27111/commits/787ceef62fccebd8a67b31e91269fd046c8ffd26

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/EsUtils.java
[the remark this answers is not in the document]
joaopamaral: 27b8b33634

joaopamaral:
> Copilot reviewed 22 out of 24 changed files in this pull request and generated no new comments.

@lautel I think Copilot is happy now 😄

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
```diff
@@ -0,0 +1,333 @@
...
+    try {
+      Request request = new Request(method, endpoint);
+      if (body != null) {
+        request.setJsonEntity(body);
+      }
+      Response response = restClient.performRequest(request);
+      int statusCode = response.getStatusCode();
+      try (InputStream is = response.getEntity().getContent()) {
+        String responseBody = new String(is.readAllBytes(), StandardCharsets.UTF_8);
+        if (statusCode >= 400) {
```
lautel: This `if (statusCode >= 400)` branch is unreachable: The `restClient` in L.203 (type `Rest5Client`) already throws `ResponseException` extends `IOException` on any non-2xx response by default. Its internal convertResponse produces a `ResponseOrResponseException` that throws the exception out. So control never reaches the manual status-code check unless the caller has explicitly opted out of that behavior, which this code does not.

Note that this is different from the OpenSearch implementation because it uses a OS generic client (`OpenSearchVectorService.executeGenericRequest`) which does not throw on non-2xx.

Please drop the `if` and let `ResponseException` propagate, or catch `ResponseException` first to format the message symmetrically with the OS path.

lautel (changes requested):
There's just one clean up to do and the PR should be good to get it in 🚀 Please check my comment to remove the dead code in `ElasticSearchVectorService`. Also please solve the merge conflict, and don't forget to run `mvn spotless:apply` before committing to apply the right formatting! Thanks!

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
joaopamaral: Hey @lautel, checked the assumption, Rest5Client only throws ResponseException on 5xx, not all non-2xx. Decompiled bytecode shows isCorrectServerResponse(int) is literally code < 500. Confirmed with some tests:

```
Status	Behavior
200	Returns Response
400	Returns Response (no throw)
404	Returns Response (no throw)
500	Throws ResponseException
503	Throws ResponseException
````

So I kept the manual >= 400 check for 4xx and added a ResponseException catch for 5xx, both paths now throw the same "Elasticsearch request failed with status N: body" wording the OS path uses, so error messages are symmetric.

11bbf702b9648586e67eb2291993f45197a90970

lautel (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/ElasticSearchVectorService.java
lautel: I see, thanks for checking! I approved the PR, please resolve the merge conflict and let's wait for the tests to pass

joaopamaral:
hey @lautel, the failing tests in CI don't seem related to this PR

lautel:
> hey @lautel, the failing tests in CI don't seem related to this PR

@joaopamaral okay, thanks for checking! Let's wait for the pipeline to be finished and I'll ask a colleague to help reviewing it. Thank you for your patience.

SuperAayush:
> hey @lautel, the failing tests in CI don't seem related to this PR

Hello @joaopamaral, apologies to ask this here but you have won $100 for the PR track of Back to the Metadata hackathon. Can you please share your active email id with me here?

Thanks!

lautel (dismissed):

mohityadav766 (dismissed):

joaopamaral:
Hey @lautel and @mohityadav766! 👋
Do you have an estimate for when this PR might be merged? Since it touches many files, there are already some merge conflicts starting to appear.

harshach:
@joaopamaral Can you please java checkstyle. I’ll merge it in

joaopamaral:
OpenSearch architecture has changed. I'm working on changing the ES part too.

joaopamaral:
I'm reverting some ES and OS merged code (keep duplicate for now) to avoid more merge conflicts.

joaopamaral:
## Review guide: what differs from OpenSearch (and what mirrors it)

To keep merge conflicts minimal until release, the OpenSearch files (`OpenSearchBulkSink`, `OpenSearchVectorService`, `VectorIndexService`) are **byte-identical to their upstream/main + feature baseline** — the Elasticsearch path intentionally duplicates the OpenSearch pattern instead of sharing code. This maps where reviewing effort is needed.

### ✅ Mirrors OpenSearch (identical after normalizing engine names — safe to skim)

**`ElasticSearchBulkSink`**: `canReuseCachedEmbedding`, `isVectorEmbeddingEnabledForEntity`, vector stats counters, the `write()` embedding pre-fetch block, `addEntity` shape.

**`ElasticSearchVectorService`**: `cachedStateMatches`, `isSpliceable`, `updateEntityEmbedding`, `extractTotalHits`, `generateEmbeddingFields`, `partialUpdateEntity`, `getExistingFingerprint(sBatch)`, `EntityFingerprintInput` record, header/source field constants, singleton `init`/`getInstance` shape.

### 🔍 Genuinely ES-specific (review focus)

1. **`ElasticSearchVectorService.search()`** — the core delta:
   - over-fetch `size` is clamped to `k` **unconditionally** (ES native kNN never returns more than `k` hits; threshold is applied client-side, and an unclamped `from+size` can exceed `index.max_result_window` → 400). OpenSearch clamps only when `threshold <= 0` because its `min_score` filter is in-query.
   - query built by `buildNativeESQuery(..., knnNumCandidatesMultiplier)` vs OS `build(..., threshold)`.

2. **`collectSearchHits`** — two real differences:
   - threshold filtering is **client-side only** on ES (in-query on OS);
   - ES persists the `parentId` fallback into the hit map (`hitMap.put("parentId", parentId)`) so consumers like `SemanticSearchTool` always see it populated. OS does not do this — noting as a possible follow-up for the OS path (kept untouched here on purpose).

3. **`getExistingEmbeddingsBatch`** — same two-step reuse algorithm as OS; transport differs: ES uses a raw `_search` ids query (`fetchSourceByIds` + `toJsonArray`, values escaped via `VectorSearchQueryBuilder.escape`) where OS uses typed `mget`.

4. **ES-only plumbing**: `extractRestClient` (requires `Rest5ClientTransport`, throws otherwise), `readEntityBody`, the `executeGenericRequest` implementation (Rest5 `Request`/`Response` + `ResponseException` handling), and the `knnNumCandidatesMultiplier` constructor/init parameter.

5. **`ElasticSearchBulkSink` deltas vs OS** (2 total): `fetchExistingEmbeddings` adds an `entities.isEmpty()` guard (defensive, from review feedback); everything else is comments/whitespace. `SinkStatsCallback` exists only on the OS sink (upstream).

6. **Beyond the sink/service pair**: `VectorSearchQueryBuilder.buildNativeESQuery` (kNN JSON, `num_candidates` clamp, in-knn filters, escaping), `EsUtils` (`dense_vector` mapping + embedding-dimension drift check), `ElasticSearchIndexManager` (mapping enrichment on create/update; `IllegalStateException` is re-thrown), `VectorSearchResource` `/fingerprint` endpoint, and the `SearchRepository` ES init branch (`resolveKnnNumCandidatesMultiplier`).

### OS-only (untouched upstream code — out of scope)
`ensureHybridSearchPipeline` / `checkHybridSearchPipeline` and the hybrid-rrf weights.

*Validated end-to-end on both engines (ES 9.3.0 `dense_vector`, OpenSearch 3.4.0 `knn_vector` + hybrid-rrf): identical semantic ranking, and embeddings survive an incremental (non-recreate) reindex — 3/3 preserved on both.*

joaopamaral:
Hey @lautel @mohityadav766, could you take another look at this PR for merge?

I had to refactor as OpenSearch and some interfaces changed while this PR was open.

On the failing CI checks, none are caused by this PR:
- **Validate PR Metadata**: the linked issue #28259 needs its Shipping-project fields set (process, not code).
- **maven-collate-ci**: workflow-dispatch into the Collate-internal repo, which fork PRs can't trigger.


I've added [a summary](https://github.com/open-metadata/OpenMetadata/pull/27111#issuecomment-4896553543) since many parts of this code mirror the OpenSearch one (highlighting the main differences). I didn't extract the shared logic into a parent class on purpose, to avoid more conflicts until this PR is merged.

PR #27117, opened 2026-04-07
Fix Flowable workflow termination message lookup and async-job race on deletion

yan-3005:
## Summary

- **Production bug fix**: `WorkflowHandler.findTerminationMessageName` was splitting task definition keys on `_` and searching for messages named `<key>_terminateProcess`. `Workflow.getFlowableElementId` joins with a `.`, so actual task keys are `<subProcess>.approvalTask` and messages are `<subProcess>.terminateProcess`. Every conflict-resolution path (a newer workflow instance replacing an older one for the same entity) silently failed since the original Custom Workflows commit — the old subprocess was never terminated via message, left dangling until force-delete. Fixed by splitting on the last `.` and looking up `<subProcessId>.terminateProcess` directly.
- **Test fix**: `waitForWorkflowIdle` now waits for async **jobs** on the trigger process definition IDs rather than waiting for process instances to drain. The async-after jobs on `PeriodicBatchEntityTrigger`'s fetch task are ephemeral (<100ms); the old approach waited for the entire batch loop to finish (potentially minutes with many entities), causing test suite time to blow up.
- Removed `waitForWorkflowIdle` from `cleanupCreatedWorkflows` `@AfterEach` — it ran a Flowable query per workflow in every test regardless of whether the workflow was ever triggered.

## Test plan

- [x] `WorkflowDefinitionResourceIT#test_DataCompletenessWorkflow_SDK` — zero `NullPointerException` / `JOB_EXECUTION_FAILURE` lines
- [x] `WorkflowDefinitionResourceIT#test_TagApprovalWorkflow_SDK` — zero `No termination message found` warnings; no 60s `ConditionTimeoutException` timeout
- [x] Full `WorkflowDefinitionResourceIT` suite — no regressions, test time back to baseline (~2m30s)

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowInstanceStageListener.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowInstanceStageListener.java
@@ -137,7 +137,7 @@ public class WorkflowInstanceStageListener implements JavaDelegate {
     String processInstanceId = execution.getProcessInstanceId();
     String businessKey = execution.getProcessInstanceBusinessKey();
     if (businessKey == null || businessKey.isEmpty()) {
-      LOG.debug(
+      LOG.warn(
           "[STAGE_SKIP] ProcessInstance: {} (workflow: {}) - no business key, not an OM-managed process instance",
           processInstanceId,
           workflowDefinitionName);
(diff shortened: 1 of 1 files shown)


yan-3005:
@gitar-bot Fix the   Edge Case: Silent skip on null businessKey uses DEBUG; WARN would aid diagnosis and do mvn spotless:apply on the root of the repository and push it

PR #27127, opened 2026-04-07
fix: cherry-pick 413 iterative bisection and TaggedOperation refactor into 1.12.5

### LINKED ISSUE #26344: Fix: Add iterative bisection on 413 and TaggedOperation refactor in bulk index sinks

manerow:
Cherry-pick of #26347 (`da400385d4`) from `main` into `1.12.5`.

## Changes

- Adds iterative bisection on 413 (Request Entity Too Large) responses in `ElasticSearchIndexSink` and `OpenSearchIndexSink` — instead of failing the whole batch, the sink splits it in half recursively until the oversized document is isolated and skipped with a clear error
- Introduces `TaggedOperation` wrapper so entity references are tracked through the bulk pipeline, ensuring errors report the actual entity instead of a positional index
- Conflict resolution: kept all `1.12.5`-specific test dependencies in `openmetadata-service/pom.xml` and dropped `junit-jupiter-params` which was removed by the original commit

## Motivation

The Data Insights Application on customer environments was failing with 413 errors when bulk indexing large entity batches. This fix was already merged to `main` but was missing from the `1.12.5` release branch.

Fixes #26344

### DIFF
--- a/openmetadata-service/pom.xml
+++ b/openmetadata-service/pom.xml
@@ -936,7 +936,6 @@
     <dependency>
       <groupId>org.junit.platform</groupId>
       <artifactId>junit-platform-commons</artifactId>
-      <version>1.9.3</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java
@@ -52,7 +52,6 @@ import org.quartz.JobExecutionContext;
 
 @Slf4j
 public class DataInsightsApp extends AbstractNativeApplication {
-  public static final String REPORT_DATA_TYPE_KEY = "ReportDataType";
   public static final String DATA_ASSET_INDEX_PREFIX = "di-data-assets";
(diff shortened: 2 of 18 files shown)


manerow:
Dropping this for now — will cherry-pick into 1.12.6 once the branch is up.

PR #27152, opened 2026-04-08
Fixes #27150: Bulk-fetch TaskInstances per DAG to eliminate N+1 in yield_pipeline_status

### LINKED ISSUE #27150: Airflow ingestion fires N extra DB queries (one per DagRun) to fetch TaskInstances inside `yield_pipeline_status`
**Affected module**
Ingestion Framework — `openmetadata-ingestion`, Airflow pipeline connector (`metadata/ingestion/source/pipeline/airflow/metadata.py`)

**Describe the bug**

Inside `AirflowSource.yield_pipeline_status()`, every `DagRun` returned by `get_pipeline_status` triggers a separate `SELECT` against the `task_instance` table:

```python
# Inside "for dag_run in dag_run_list:" — fires once per DagRun
tasks = self.get_task_instances(
    dag_id=dag_run.dag_id,
    run_id=dag_run.run_id,          # ← different run_id each iteration
    serialized_tasks=pipeline_details.tasks,
)
```

`get_pipeline_status` returns up to `numberOfStatus` DagRuns (default 10). For a deployment with 200 DAGs and the default `numberOfStatus=10`, this produces **2 000 extra round-trips per ingestion run**, on top of the 200 DagRun queries. The number of task-instance queries scales as `DAGs × numberOfStatus`, and is entirely avoidable: all TaskInstances for a DAG can be fetched in a single query using `run_id IN (...)`.

**To Reproduce**

1. Connect OpenMetadata ingestion to an Airflow database with ≥ 50 DAGs.
2. Enable SQL query logging on the Airflow metadata database.
3. Run the Airflow ingestion pipeline.
4. Observe one `SELECT ... FROM task_instance WHERE dag_id = ? AND run_id = ?` per DagRun per DAG, in addition to the paginated `dag_run` queries.

**Expected behavior**

TaskInstances for all runs of a given DAG should be fetched in a single bulk query (`run_id IN (...)`) and grouped in memory, so that `yield_pipeline_status` issues exactly **1 TaskInstance query per DAG** regardless of `numberOfStatus`.

**Version:**
- Python version: 3.10 / 3.11
- OpenMetadata version: 1.6.x / 1.7.x
- OpenMetadata Ingestion package version: `openmetadata-ingestion[airflow]`

**Additional context**

The fix is straightforward:

1. Change `get_task_instances(dag_id, run_id, serialized_tasks)` to accept `run_ids: List[str]` instead of a single `run_id: str`.
2. Replace `.filter(TaskInstance.run_id == run_id)` with `.filter(TaskInstance.run_id.in_(run_ids))`.
3. Return `Dict[str, List[OMTaskInstance]]` keyed by `run_id` (built with `defaultdict(list)`).
4. In `yield_pipeline_status`, collect all `run_id`s from `dag_run_list` before the loop, call the bulk method once, and look up results per run with `tasks_by_run_id.get(dag_run.run_id, [])`.

Query count drops from `1 + N_runs` per DAG to exactly `2` per DAG (1 for DagRuns, 1 for all TaskInstances).

RajdeepKushwaha5:
### Describe your changes:

Fixes #27150

`AirflowSource.yield_pipeline_status` was calling `get_task_instances` once per `DagRun` inside a loop — one separate `SELECT ... FROM task_instance WHERE dag_id = ? AND run_id = ?` per run. With the default `numberOfStatus=10` and N DAGs this produces `N × 10` extra DB round-trips on every ingestion run.

**Root cause:** `get_task_instances` accepted a single `run_id` string, so the only way to retrieve task instances for multiple runs was a loop.

**Fix:**

| | Before | After |
|---|---|---|
| `get_task_instances` signature | `run_id: str` → `List[OMTaskInstance]` | `run_ids: List[str]` → `Dict[str, List[OMTaskInstance]]` |
| DB filter | `TaskInstance.run_id == run_id` | `TaskInstance.run_id.in_(run_ids)` |
| Grouping | N separate queries, results consumed immediately | Single query, results grouped into `defaultdict(list)` by `run_id` |
| `yield_pipeline_status` | Calls `get_task_instances` inside `for dag_run in dag_run_list` | Collects all `run_ids` → calls `get_task_instances` once → looks up with `tasks_by_run_id.get(run_id, [])` |

Task-instance query count per DAG drops from `numberOfStatus` to **1**.

**Error handling preserved:** `result` (`defaultdict(list)`) is initialised before the `try` block, so on DB exception the method returns an empty dict (same safe fallback behaviour as before — all runs get empty task lists).

**How I tested:**
- Syntax verified with `ast.parse`
- Manually traced the lookup path: `run_ids` is derived from the same `dag_run_list` used in the iteration, so no `run_id` can be present in the loop but absent from the bulk query
- Empty-`run_ids` guard (`if run_ids else {}`) avoids issuing `IN ()` which is illegal in some SQL dialects

#
### Type of change:
- [x] Bug fix

#
### Checklist:
- [x] My PR title is `Fixes #NNNNN: Bulk-fetch TaskInstances per DAG to eliminate N+1 in yield_pipeline_status`

- [ ] I have added a test that covers the exact scenario we are fixing.


harshach:
@RajdeepKushwaha5 did you run this against an airflow instance and made sure that all dags are listed properly without any regression?

harshach:
@RajdeepKushwaha5 please check my comments, nee to make sure this is tested properly before we merge it in cc @PubChimps

ingestion/src/metadata/ingestion/source/pipeline/airflow/metadata.py
```diff
@@ -344,50 +341,59 @@ class to only focus on core properties required
...
                     TaskInstance.state != AirflowTaskStatus.REMOVED.value,
                 )
                 .all()
             )
+            for elem in task_instance_list:
+                row = elem._asdict()
+                if row.get("task_id") in serialized_tasks_ids:
+                    result[row["run_id"]].append(
+                        OMTaskInstance(
+                            task_id=row.get("task_id"),
```
harshach: This row.get() should be bit defensive no? what happesn if one of the fields we are referencing is not there?

ingestion/tests/unit/topology/pipeline/test_airflow.py
```diff
@@ -998,3 +1002,210 @@ def test_task_source_url_airflow_3x_format(self):
         assert f"/dags/{quote(dag_id)}/tasks/{quote(task_id)}" in task_url
         assert "/taskinstance/list/" not in task_url
         assert "_flt_3_dag_id=" not in task_url
+
+    def test_get_task_instances_bulk_query(self):
```
harshach, line 1006: can we tests with negative data as well to see how it behaves if some of the data is not there, do we throw runtime exception and interrupt the processing of all the data or do we log and move forward. Second behavior is preferred

ingestion/src/metadata/ingestion/source/pipeline/airflow/metadata.py
IceS2: This is looking great as the .get default is None and the class has everything but task_id as optional:
```
class OMTaskInstance(BaseModel):
    """
    Custom model we get from the Airflow db
    as a scoped SELECT from TaskInstance
    """

    task_id: str
    state: Optional[str]
    start_date: Optional[datetime]
    end_date: Optional[datetime]
```

RajdeepKushwaha5:
Thanks for the review @harshach — both points addressed in `7a62fc4c53`. Full context, defensive edit, negative tests, and live verification below.

---

## 1. Defensive per-row handling in `get_task_instances`

Wrapped the per-row processing in its own `try/except` and explicitly guard `task_id` / `run_id` for falsy values before indexing `result[run_id]`. A malformed or partial row now logs and skips — the rest of the batch keeps processing, so a single bad `TaskInstance` can never abort the whole DAG's status ingestion.

```python
for elem in task_instance_list:
    try:
        row = elem._asdict()
        task_id = row.get("task_id")
        run_id = row.get("run_id")
        if not task_id or not run_id:
            logger.debug(
                f"Skipping TaskInstance row with missing task_id/run_id "
                f"for dag_id={dag_id}: {row}"
            )
            continue
        if task_id not in serialized_tasks_ids:
            continue
        result[run_id].append(
            OMTaskInstance(
                task_id=task_id,
                state=row.get("state"),
                start_date=row.get("start_date"),
                end_date=row.get("end_date"),
            )
        )
    except Exception as row_exc:
        logger.warning(
            f"Skipping malformed TaskInstance row for dag_id={dag_id}: {row_exc}"
        )
        continue
```

Behavior matches the preference you stated on the review: **log and move forward, do not interrupt processing of the whole DAG**.

The outer `try/except` around the DB query is retained so that schema-level errors (e.g. older Airflow versions without a `run_id` column) still degrade gracefully to an empty dict — identical to the pre-PR fallback behavior, just batched.

---

## 2. Negative-data unit tests

Three new tests added to `ingestion/tests/unit/topology/pipeline/test_airflow.py`:

- **`test_get_task_instances_skips_rows_with_missing_fields`** — DB returns a mix of good rows and rows where `task_id=None` / `run_id=None`. Asserts the good rows are still ingested under the correct run_ids and no exception propagates.
- **`test_get_task_instances_continues_on_malformed_row`** — a middle row's `._asdict()` raises `RuntimeError`. Asserts the two surrounding good rows are still in the final result dict.
- **`test_get_task_instances_unknown_run_id_not_in_result`** — DB returns a `TaskInstance` for a `run_id` the caller never requested (simulating stale cache / delete race). Asserts no crash; caller's `tasks_by_run_id.get(run_id, [])` semantics handle it.

These are in addition to the earlier tests already in the PR:
- `test_get_task_instances_bulk_query` (single DB query, grouping by run_id, unknown-task filtering)
- `test_get_task_instances_no_regression_vs_old_per_run_loop` (behavioural equivalence with old loop on mixed data)
- `test_get_task_instances_returns_empty_dict_on_db_exception` (DB query itself raises)
- `test_get_task_instances_handles_empty_run_ids` (empty input)

---

## 3. Live verification against a real Airflow 2.10.5 metadata DB

Re: your earlier question *"did you run this against an airflow instance and made sure that all dags are listed properly without any regression?"* — yes, full setup and run details below.

**Environment**
- Fresh Python 3.12 venv
- `apache-airflow==2.10.5` installed with the official Airflow constraints file (`constraints-3.12.txt`)
- Clean `AIRFLOW_HOME`, real SQLite metadata DB via `airflow db init`

**Seed** — rows inserted straight into the real Airflow tables (`dag`, `dag_run`, `task_instance`) via SQLAlchemy Core:

| DAG        | DagRuns | TaskInstances |
|------------|--------:|--------------:|
| dag_small  |       3 |             6 |
| dag_medium |      10 |            40 |
| dag_large  |      25 |           150 |
| **Total**  |  **38** |       **196** |

Mixed states (`success`, `failed`), ~14% flipped to `removed` to exercise the `state != "removed"` filter, mixed task names to exercise the `serialized_tasks_ids` filter.

**Harness** — runs the merged bulk query side-by-side with a reconstruction of the pre-PR per-run loop against the same session, with a SQLAlchemy `before_cursor_execute` listener counting every statement hitting `task_instance`. Per-run task lists are compared tuple-by-tuple `(task_id, state, start_date, end_date)`.

**Result**

```
[2/5] Running BULK query for dag_small (3 run_ids)...
     task_instance queries=1  run_ids_returned=3
[3/5] Running OLD per-run loop for dag_small ...
     task_instance queries=3  run_ids_returned=3
     OK: bulk == loop for dag_small (per-run task lists identical)
     -> SQL reduction: 3 -> 1

[2/5] Running BULK query for dag_medium (10 run_ids)...
     task_instance queries=1  run_ids_returned=10
[3/5] Running OLD per-run loop for dag_medium ...
     task_instance queries=10  run_ids_returned=10
     OK: bulk == loop for dag_medium (per-run task lists identical)
     -> SQL reduction: 10 -> 1

[2/5] Running BULK query for dag_large (25 run_ids)...
     task_instance queries=1  run_ids_returned=25
[3/5] Running OLD per-run loop for dag_large ...
     task_instance queries=25  run_ids_returned=25
     OK: bulk == loop for dag_large (per-run task lists identical)
     -> SQL reduction: 25 -> 1

RESULT: PASS - bulk query returns identical data to old loop
        across 3 DAGs / 38 DagRuns / 196 TaskInstances. No regression.
```

**On "all DAGs listed properly"** — the diff only touches `get_task_instances()` and its single caller `yield_pipeline_status()`. DAG-enumeration paths (`get_pipelines_list`, `_get_dags`, DAG-object construction, `get_pipeline_status`) are **not modified**, so DAG listing is unchanged by construction. The only behavioral change is TaskInstance fetching, and the live test above shows per-run output is byte-for-byte equivalent to the old loop — just in one round trip instead of N.

---

## 4. Negative-data scenarios against the real modified function body

Separately from the unit tests, I ran the eight negative-data scenarios below against the **actual** modified `get_task_instances` body (extracted from `metadata.py` at runtime via `ast.parse` + `ast.unparse` + `exec`, with the surrounding OpenMetadata imports stubbed so it runs without the full ingestion dev env on Windows):

```
1. normal bulk result                                 PASS
2. task_id is None          -> skipped, rest kept     PASS
3. run_id is None           -> skipped, rest kept     PASS
4. _asdict() raises mid-row -> log and continue       PASS
5. DB query itself raises   -> return {}              PASS
6. row for unknown task_id  -> filtered out           PASS
7. stray run_id from DB     -> grouped separately     PASS
8. every row malformed      -> empty result, no raise PASS

Result: 8/8 scenarios PASS
```

Logs confirmed for scenarios 4 and 8:

```
WARNING - Skipping malformed TaskInstance row for dag_id=d: corrupt row bytes
```

and for scenario 5:

```
WARNING - Tried to get TaskInstances for run_ids. The run_id column might not
          be available in older Airflow DB schemas - older airflow schema,
          run_id column missing.
```

No exception ever propagates out of `get_task_instances` — `yield_pipeline_status` always gets a well-formed `Dict[str, List[OMTaskInstance]]`.

---

## Summary

- ✅ Defensive per-row handling with explicit null-guard + log-and-continue
- ✅ Outer query-level `try/except` preserved for schema-drift safety
- ✅ 3 new negative-data unit tests added on top of the existing 4 tests for this method
- ✅ Live test on real Airflow 2.10.5 DB: 38 DagRuns / 196 TaskInstances, bulk output identical to old loop, N queries → 1 query
- ✅ 8/8 negative-data scenarios pass against the real modified function body
- ✅ DAG listing paths untouched by the diff — no regression possible there by construction

Happy to run this against Postgres or a larger seed if you'd like more coverage before merging.

IceS2 (commented):
Looks good, my main concern here is that we could potentially fetch a lot of information while creating that results dict, right? Not sure I am missing anything that prevents that.

Thinking about that, it might make sense to do it in chunks from within the yield_pipeline_status. What do you think?

RajdeepKushwaha5:
Thanks @IceS2 — and thanks for confirming the `row.get()` defensive bit for @harshach, agreed the `Optional` fields on `OMTaskInstance` handle the missing-column case cleanly.

On the chunking concern: you're right, let me show the current bound and then agree on where to put the chunking.

## Current bound

`run_ids` comes from `get_pipeline_status()`, which does:

```python
.order_by(db_date_column.desc())
.limit(self.config.serviceConnection.root.config.numberOfStatus)
```

So `run_ids` is already capped at `numberOfStatus` (default **10**, user-configurable). Worst-case memory is `numberOfStatus × avg_tasks_per_run`:

| `numberOfStatus` | ~tasks/run | rows in `result` dict |
|---:|---:|---:|
| 10 (default) | 10 | ~100 |
| 50 | 20 | ~1 000 |
| 200 | 50 | ~10 000 |

We project only 5 scalar columns (`task_id`, `state`, `start_date`, `end_date`, `run_id`) — no blobs, no XCom, no log text — so per-row footprint is small.

The real risks at high `numberOfStatus`:

1. Driver-level `IN (...)` parameter caps (SQLite 999, some MySQL configs 1000).
2. Peak memory holding the whole `result` dict before any status is yielded.
3. One failed query wiping out the whole DAG's status instead of degrading gracefully.

All three are real and worth fixing. Chunking addresses all of them.

## Where to put the chunk

You suggested doing it from within `yield_pipeline_status`, and I think that's the right call — it lets us **stream** `OMetaPipelineStatus` per chunk instead of buffering everything in a dict first, which directly matches the "generator" spirit of `yield_*` methods in this file. Memory stays bounded to one chunk at a time regardless of `numberOfStatus`.

Rough shape:

```python
# Module constant
_TASK_INSTANCE_RUN_ID_CHUNK_SIZE = 50


def yield_pipeline_status(self, pipeline_details):
    try:
        dag_run_list = self.get_pipeline_status(pipeline_details.dag_id)
        eligible_runs = [
            dag_run for dag_run in dag_run_list
            if dag_run.run_id and self.context.get().task_names
        ]

        for start in range(0, len(eligible_runs), _TASK_INSTANCE_RUN_ID_CHUNK_SIZE):
            chunk = eligible_runs[start : start + _TASK_INSTANCE_RUN_ID_CHUNK_SIZE]
            try:
                tasks_by_run_id = self.get_task_instances(
                    dag_id=pipeline_details.dag_id,
                    run_ids=[r.run_id for r in chunk],
                    serialized_tasks=pipeline_details.tasks,
                )
            except Exception as exc:
                logger.warning(
                    f"Failed TaskInstance chunk for {pipeline_details.dag_id} "
                    f"(runs {start}-{start + len(chunk)}) - {exc}"
                )
                continue  # other chunks still produce statuses

            for dag_run in chunk:
                tasks = tasks_by_run_id.get(dag_run.run_id, [])
                # ...existing PipelineStatus / Either yield logic unchanged...
                yield Either(right=OMetaPipelineStatus(...))
    except Exception as exc:
        ...
```

`get_task_instances` itself stays single-query (fed a bounded chunk), which keeps it simple and easy to test. Per-chunk `try/except` pairs with the per-row `try/except` we already added from @harshach's review, so a broken chunk just logs and we keep going.

## If you agree, I'll push

- The refactor above.
- A unit test asserting `len(run_ids) > _TASK_INSTANCE_RUN_ID_CHUNK_SIZE` produces multiple `get_task_instances` calls and that the merged yielded statuses equal the single-chunk path.
- Re-run of the live Airflow 2.10.5 harness I set up for the initial validation (3 DAGs / 38 DagRuns / 196 TaskInstances) with chunk size temporarily lowered to force the multi-chunk path, confirming byte-equivalent per-run output vs the old loop.

Let me know and I'll have it up shortly.
Thanks @IceS2 — good thing to raise. Let me show why I think the current shape is already bounded in practice, and propose where to add the chunking for the defense-in-depth you're after.

## What bounds the fetch today

`run_ids` comes directly from `get_pipeline_status()`, which does:

```python
.order_by(db_date_column.desc())
.limit(self.config.serviceConnection.root.config.numberOfStatus)
```

So the list passed into `get_task_instances` is already capped at `numberOfStatus`, which defaults to **10** and is user-configurable per service. The worst realistic case is `numberOfStatus × avg_tasks_per_run`:

| `numberOfStatus` | ~tasks/run | rows held in memory |
|---:|---:|---:|
| 10 (default) | 10 | ~100 |
| 50 | 20 | ~1 000 |
| 200 | 50 | ~10 000 |

We also filter `state != 'removed'` server-side and only project 5 scalar columns (`task_id`, `state`, `start_date`, `end_date`, `run_id`) — no blobs, no XCom, no log text. Per-row footprint is small, and the upper bound is effectively set by whatever users pick for `numberOfStatus`.

The one concrete risk I see at very high `numberOfStatus` is driver-level `IN (...)` parameter caps (SQLite 999, some MySQL configs 1000). Unlikely to hit with `numberOfStatus` alone, but cheap to guard against — and your point about not building an unbounded `results` dict in one shot is fair.

## Proposal

Rather than pushing chunk logic up into `yield_pipeline_status` (which would make the caller own a DB concern it otherwise doesn't), I'd prefer to chunk **inside** `get_task_instances` with a module-level constant:

```python
_TASK_INSTANCE_RUN_ID_CHUNK_SIZE = 50

def get_task_instances(self, dag_id, run_ids, serialized_tasks):
    serialized_tasks_ids = {t.task_id for t in serialized_tasks}
    result: Dict[str, List[OMTaskInstance]] = defaultdict(list)
    if not run_ids:
        return result
    for start in range(0, len(run_ids), _TASK_INSTANCE_RUN_ID_CHUNK_SIZE):
        chunk = run_ids[start : start + _TASK_INSTANCE_RUN_ID_CHUNK_SIZE]
        try:
            # same query shape as today, filtered by TaskInstance.run_id.in_(chunk)
            # accumulate into `result` (same per-row try/except we added for @harshach)
            ...
        except Exception as exc:
            logger.warning(
                f"Failed TaskInstance chunk for {dag_id} "
                f"(runs {start}-{start + len(chunk)}) - {exc}"
            )
            continue  # other chunks still produce valid status
    return result
```

Advantages:

- `yield_pipeline_status` stays a clean one-liner — it doesn't need to know the DB can't take a 10k-element `IN` list.
- Per-chunk `try/except` pairs nicely with the per-row `try/except` we already added from @harshach's review, so a failure in one chunk doesn't wipe out the whole DAG's status.
- Chunk size is a single tunable constant in the module.
- Unit tests can assert: `N > chunk_size` → multiple DB queries issued, merged result equals single-chunk output.

## Alternative you mentioned

If you'd rather the chunking live in `yield_pipeline_status` so statuses can start **yielding** after each chunk (streaming semantics rather than just memory-bounding), I'm happy to do that instead — `yield_pipeline_status` would loop chunks of run_ids and emit `OMetaPipelineStatus` per chunk. Slightly larger refactor but arguably matches the spirit of `yield_pipeline_status` better.

## What I'd like to know

Which direction do you prefer?

1. **Chunk inside `get_task_instances`** — simplest, keeps caller unchanged, bounds memory per query.
2. **Chunk inside `yield_pipeline_status`** — streams statuses progressively per chunk.

Once you pick one I'll push the change with:

- A new unit test asserting that `N > chunk_size` produces multiple DB queries and the merged result is identical to the single-chunk path.
- A re-run of the live Airflow 2.10.5 harness I set up earlier (3 DAGs / 38 DagRuns / 196 TaskInstances) with the chunk size temporarily lowered to force the multi-chunk path, to verify byte-equivalent per-run output vs the old loop.

IceS2:
Hey @RajdeepKushwaha5, I agree with you that in practice it is bounded, I just think that since we are touching this code we could add this defensive guardrail.

I like your idea better, if you want to implement that and give it a try to guarantee it works as expected by adding the unit test and doing your manual tests on airflow it would be great!

RajdeepKushwaha5:
> Hey @RajdeepKushwaha5, I agree with you that in practice it is bounded, I just think that since we are touching this code we could add this defensive guardrail.
> 
> I like your idea better, if you want to implement that and give it a try to guarantee it works as expected by adding the unit test and doing your manual tests on airflow it would be great!

Hi @IceS2! Happy you liked the chunking idea — I've already gone ahead and implemented it. Pushed in `834d3b3d50` (chunking) 

### What landed

**Production change** (`ingestion/src/metadata/ingestion/source/pipeline/airflow/metadata.py`):

- Added a module-level constant `_TASK_INSTANCE_RUN_ID_CHUNK_SIZE = 50` as the defensive guardrail.
- `yield_pipeline_status` now slices `eligible_dag_runs` into chunks of that size and calls `get_task_instances` once per chunk, so the `IN (...)` list is always bounded regardless of how many DagRuns a DAG has.
- `get_task_instances` short-circuits with an empty dict when `run_ids` is empty (skips a pointless DB round-trip).
- On a chunk-level failure we log and fall back to `tasks_by_run_id = {}` instead of skipping the chunk — this preserves the pre-PR safe-fallback behaviour where a DB error produced an empty `taskStatus` list but the run was still emitted.

**Unit tests** (`ingestion/tests/unit/topology/pipeline/test_airflow.py`):

- `test_yield_pipeline_status_chunks_run_ids` — 125 runs with the chunk size monkey-patched: asserts 3 calls with sizes `[50, 50, 25]` and all 125 statuses yielded.
- `test_yield_pipeline_status_chunk_failure_does_not_block_other_chunks` — 30 runs, middle chunk's `get_task_instances` raises: asserts all 30 statuses are still yielded and the failed-chunk runs have empty `taskStatus` lists.
- Plus the existing bulk-query / no-regression / empty-run_ids / malformed-row tests from earlier commits.

### Manual validation

I ran a standalone SQLite harness that extracts the real `get_task_instances` from the source via AST (so there's no copy-paste drift) and uses a SQLAlchemy event listener to count queries. All three scenarios pass:

| Scenario | Result |
| --- | --- |
| Small DAG — bulk vs old per-run loop | PASS — bulk = 1 query, old loop = 10 queries, outputs byte-identical |
| 125 runs @ chunk size 50 | PASS — chunks = 3, sizes = `[50, 50, 25]`, 1 query per chunk, byte-equivalent to old loop |
| Middle chunk raises | PASS — 3 attempts made, surviving runs intact, failed-chunk runs still yield with empty tasks |

So query count is now bounded to `ceil(N / 50)` instead of N, with per-chunk fault isolation and no behaviour change from the caller's perspective.

CI is re-running on `73e99dd3bf` — `py-checkstyle` was previously failing due to `black`/`isort` formatting on the touched files, which is also fixed in that commit (all three formatters pass locally with the `ingestion/pyproject.toml` config). Let me know if you'd like me to tweak the chunk size or anything else!

IceS2 (approved):

RajdeepKushwaha5:
Hi @IceS2 
The only failing check is Postgresql PR Playwright E2E Tests / playwright-ci-postgresql (3, 6) — the other 5 Playwright shards all passed , and so did SonarCloud (Quality Gate passed, 100% coverage on new code, 0 new issues), Java/Python checkstyle, and SonarCloud code scanning (no new alerts in changed code).

The shard-3 failure looks like Playwright flakiness unrelated to this PR's changes. 
Could you re-trigger that shard when you have a moment?

IceS2:
> Hi @IceS2 The only failing check is Postgresql PR Playwright E2E Tests / playwright-ci-postgresql (3, 6) — the other 5 Playwright shards all passed , and so did SonarCloud (Quality Gate passed, 100% coverage on new code, 0 new issues), Java/Python checkstyle, and SonarCloud code scanning (no new alerts in changed code).
> 
> The shard-3 failure looks like Playwright flakiness unrelated to this PR's changes. Could you re-trigger that shard when you have a moment?

I had already done that, let's see if it passes. Will check if it is a know issue otherwise

ingestion/src/metadata/ingestion/source/pipeline/airflow/metadata.py
```diff
@@ -344,50 +356,95 @@ class to only focus on core properties required
...
-                    tasks = self.get_task_instances(
-                        dag_id=dag_run.dag_id,
-                        run_id=dag_run.run_id,
+            # Filter eligible DagRuns once. task_names is empty when the DAG
+            # has no tasks in the current context, in which case we skip the
+            # DB round trip entirely.
+            task_names = self.context.get().task_names
+            eligible_runs = [
+                dag_run for dag_run in dag_run_list if dag_run.run_id and task_names
+            ]
```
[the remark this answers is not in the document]
RajdeepKushwaha5: I checked the original `main` code — the same `task_names` check was already gating the entire status-yielding `if` block (not just the `get_task_instances` call), so when `task_names` is empty, `main` also yields nothing for that DAG. The PR preserves this exactly.

The "still emit per-run statuses with empty taskStatus" behavior you're describing actually corresponds to the DB-failure path, which is preserved via the `tasks_by_run_id = {}` fallback inside the chunk-exception handler — so a failed bulk fetch still yields runs with empty task lists, just like the prior per-run loop did on a per-run DB error.

No behavioral change here; resolving as not actionable.

RajdeepKushwaha5:
ping @IceS2

pmbrull (approved):

PR #27160, opened 2026-04-08
fix: strip nested column children from Data Insights documents

manerow:
Fixes open-metadata/openmetadata-collate#3555
## Summary

Stops the Data Insights \`DataAssetsWorkflow\` from indexing the recursive \`children\` subtree of columns, keeping only top-level column metadata.

## Context

Investigation of a DI failure on a customer environment found that a single table with deeply nested STRUCT/UNION columns expanded into **611 unique field paths** when indexed into the DI data stream. Combined with the field paths from tens of thousands of other tables, this pushed the DI index past OpenSearch's default \`index.mapping.total_fields.limit\` of 1000, causing bulk writes to fail with \`mapper_parsing_exception\`. The same table was also behind the earlier 413 payload-size failures.

## Change

In \`DataInsightsEntityEnricherProcessor.enrichEntity\`, after the existing field filter, walk the top-level \`columns\` list in the serialized entity map and remove the \`children\` key from each entry. The fix is purely a post-processing step on the map — the source entity is not mutated, and no other code path is affected.

Top-level column metadata (name, type, description, tags, etc.) is preserved, so DI metrics that rely on it (\`numberOfColumns\`, \`numberOfColumnsWithDescription\`) continue to work. Those metrics are computed from the entity POJO directly, not the map, so they are unaffected regardless.

## Scope note

This PR only addresses \`columns\` (used by \`table\` and \`dashboardDataModel\`). Other recursive fields configured in \`dataInsights/config.json\` — \`tasks\`, \`children\`, \`fields\`, \`mlFeatures\` — carry the same latent risk and should be tackled as a follow-up.

## Test plan

- [ ] \`mvn -pl openmetadata-service compile\` passes locally
- [ ] Manual verification on a DI run: confirm \`columns[*].children\` is absent from documents in the \`di-data-assets-table\` data stream
- [ ] Confirm the DI workflow no longer fails on the previously-failing table

### DIFF
--- a/openmetadata-service/src/test/java/org/openmetadata/service/apps/bundles/insights/workflows/dataAssets/processors/DataInsightsEntityEnricherProcessorTest.java
+++ b/openmetadata-service/src/test/java/org/openmetadata/service/apps/bundles/insights/workflows/dataAssets/processors/DataInsightsEntityEnricherProcessorTest.java
@@ -185,6 +185,43 @@ class DataInsightsEntityEnricherProcessorTest {
     assertFalse(configContent.contains("\"status\""));
   }
 
+  @Test
+  void testStripNestedColumnChildrenRemovesChildrenAndPreservesOtherFields() throws Exception {
+    Map<String, Object> column = new HashMap<>();
+    column.put("name", "struct_col");
+    column.put("dataType", "STRUCT");
+    column.put("description", "A struct column");
+    column.put("children", List.of(Map.of("name", "child_field")));
+
+    Map<String, Object> entityMap = new HashMap<>();
+    entityMap.put("columns", new ArrayList<>(List.of(column)));
+
+    invokeStripNestedColumnChildren(entityMap);
+
+    assertFalse(column.containsKey("children"));
+    assertEquals("struct_col", column.get("name"));
+    assertEquals("STRUCT", column.get("dataType"));
+    assertEquals("A struct column", column.get("description"));
+  }
+
+  @Test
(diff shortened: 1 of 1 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/workflows/dataAssets/processors/DataInsightsEntityEnricherProcessor.java
```diff
@@ -245,6 +246,26 @@ private Map<String, Object> enrichEntity(
...
+  /**
+   * Removes the recursive {@code children} subtree from every top-level column entry in the
+   * serialized entity map. The DI data stream uses dynamic field mapping, and deeply nested
+   * STRUCT/UNION column types can expand into hundreds of unique field paths per document,
+   * pushing the index past OpenSearch's {@code index.mapping.total_fields.limit} of 1000.
+   * Top-level column metadata (name, type, description, etc.) is preserved.
+   */
+  @SuppressWarnings("unchecked")
+  private static void stripNestedColumnChildren(Map<String, Object> entityMap) {
+    Object columns = entityMap.get("columns");
```
harshach, line 258: @manerow can we addd a unit test coverage for this

manerow: Added unit tests ✅

PR #27238, opened 2026-04-10
test(e2e): scope Data Quality permissions Playwright helpers to fixture

shah-harshit:
### Related
https://github.com/open-metadata/openmetadata-collate/issues/3610

### Summary
Refactors `DataQualityPermissions.spec.ts` so `PolicyClass`, `RolesClass`, `UserClass`, and `TableClass` instances are created inside the Playwright worker fixture instead of at module scope. This avoids shared mutable state across parallel workers and improves reliability of the Data Quality permissions E2E suite.
Screenshot 2026-04-10 at 5 40 12 PM

### Test plan
- [ ] Run the targeted Playwright spec, for example: `yarn playwright:run` with filter for `DataQualityPermissions.spec.ts`

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DataQuality/DataQualityPermissions.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DataQuality/DataQualityPermissions.spec.ts
@@ -24,14 +24,11 @@ import {
   TEST_SUITE_POLICY,
   VIEW_ALL_TEST_CASE_POLICY,
 } from '../../../constant/dataQualityPermissions';
-import { PolicyClass } from '../../../support/access-control/PoliciesClass';
-import { RolesClass } from '../../../support/access-control/RolesClass';
 import { TableClass } from '../../../support/entity/TableClass';
 import { UserClass } from '../../../support/user/UserClass';
 import { performAdminLogin } from '../../../utils/admin';
 import { redirectToHomePage, uuid } from '../../../utils/common';
 import { waitForAllLoadersToDisappear } from '../../../utils/entity';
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DataQuality/DataQualityPermissions.spec.ts
```diff
@@ -41,51 +41,50 @@ import {
   waitForTestCaseListResponse,
 } from '../../../utils/testCases';
 
-// --- Objects ---
-const createPolicy = new PolicyClass();
-const createRole = new RolesClass();
-const createUser = new UserClass();
+let createPolicy: PolicyClass;
```
chirag-madlani: @shah-harshit any reason we are not getting this clean like [this](https://github.com/open-metadata/OpenMetadata/blob/58f649746c7e21921af5d7c27f764d9efa59451a/openmetadata-ui/src/main/resources/ui/playwright/support/user/UserClass.ts#L123-L145)

shah-harshit: This is to fix “entity already exists” error.

Creating `const x = new PolicyClass()` at the top runs at import time (once per worker), leading to a shared instance and higher chances of name/FQN collisions.

Initializing it in `beforeAll` or fixtures ensures controlled initialization within the test lifecycle, allowing fresh instances per worker/describe and better alignment with unique name generation.

The “entity already exists” error typically comes from duplicate creations across parallel runs. Moving initialization to `beforeAll`/fixtures avoids shared state and ensures users are created once in the correct scope, preventing these collisions.

chirag-madlani (approved):

PR #27244, opened 2026-04-10
MINOR: Handle case sensitivity for table constraints

ulixius9:
## Summary
- Added `normalize_table_constraints()` static method in `DatabaseServiceSource` base class that normalizes constraint column names to match actual column definitions using case-insensitive lookup
- Applied normalization in `CommonDbSourceService.yield_table()` (covers all SQL connectors) and `UnitycatalogSource.yield_table()`
- Fixes `Invalid column name found in table constraint` errors caused by data sources (e.g., BigQuery) returning constraint column names with different casing than column definitions

## Test plan
- [x] Added 8 unit tests in `test_common_db_source.py::TestNormalizeTableConstraints` covering:
  - Case mismatch normalization (`Entity_id` → `Entity_ID`)
  - All-lowercase constraint columns
  - Already matching columns (no-op)
  - Unmatched columns preserved as fallback
  - Empty/None constraints and columns
  - Multiple constraints normalized
  - Constraints with None columns field

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/database_service.py
+++ b/ingestion/src/metadata/ingestion/source/database/database_service.py
@@ -345,6 +345,31 @@ class DatabaseServiceSource(
         if self.source_config.includeTags:
             yield from self.yield_database_tag(database_name) or []
 
+    @staticmethod
+    def normalize_table_constraints(
+        table_constraints: List[TableConstraint],
+        columns: List[Column],
+    ) -> List[TableConstraint]:
+        """
+        Normalize constraint column names to match actual column definitions.
+        Some data sources (e.g., BigQuery) may return constraint column names
+        with different casing than the column definitions, causing validation
+        failures on the backend.
+        """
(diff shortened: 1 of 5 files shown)


ingestion/src/metadata/ingestion/source/database/database_service.py
```diff
@@ -345,6 +345,31 @@ def yield_database_tag_details(
...
+        Normalize constraint column names to match actual column definitions.
+        Some data sources (e.g., BigQuery) may return constraint column names
+        with different casing than the column definitions, causing validation
+        failures on the backend.
+        """
+        if not table_constraints or not columns:
+            return table_constraints or []
+        column_name_map = {}
+        for col in columns:
+            col_name = col.name.root if hasattr(col.name, "root") else str(col.name)
```
pmbrull: i think we already have a model_str method that does this

pmbrull (approved):

PR #27246, opened 2026-04-10
fix(playwright): replace loader waits with API response awaits in right panel tab navigation

harsh-vador:
## Why

Tab navigation in the Explore right panel was flaky because tests relied on `waitFor({ state: 'detached' })` on `[data-testid="loader"]` after clicking a tab.
This is unreliable in two ways:

1. If the API response arrives before the `waitFor` executes, the loader has already been removed — `detached` can behave unexpectedly depending on whether the element was ever in the DOM.
2. `detached` requires the element to be completely absent from the DOM; if it is present but hidden (via CSS), the wait times out.

## What

- **`CustomPropertiesPageObject`** — awaits `GET /metadata/types/name/*?fields=customProperties` before asserting, guaranteeing the entity type definition (with custom property definitions) has loaded.
- **`LineagePageObject`** — awaits `GET /lineage/getLineage*` before asserting.
- **`DataQualityPageObject`** — awaits `GET /dataQuality/testCases/search/list*` before asserting.
- **`RightPanelPageObject`** — changes both `waitForLoadersToDisappear` and `waitForPanelLoaded` from `{ state: 'detached' }` → `{ state: 'hidden' }`, which resolves correctly whether or not the loader element is in the DOM.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/PageObject/Explore/RightPanelPageObject.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/PageObject/Explore/RightPanelPageObject.ts
@@ -13,6 +13,7 @@
 
 import { expect, Locator, Page } from '@playwright/test';
 import { EntityClass } from '../../../support/entity/EntityClass';
+import { waitForAllLoadersToDisappear } from '../../../utils/entity';
 import { CustomPropertiesPageObject } from './CustomPropertiesPageObject';
 import { DataQualityPageObject } from './DataQualityPageObject';
 import { LineagePageObject } from './LineagePageObject';
@@ -935,7 +936,7 @@ export class RightPanelPageObject {
     await expect(this.panelLoaders).toHaveCount(0, { timeout });
 
     // Step 3: Wait for any remaining loaders on the page (fallback)
-    await this.pageLoader.waitFor({ state: 'hidden', timeout });
+    await this.waitForLoadersToDisappear(timeout);
 
     // Step 4: Ensure panel is still visible and stable
     await this.getSummaryPanel().waitFor({ state: 'visible' });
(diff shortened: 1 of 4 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/PageObject/Explore/RightPanelPageObject.ts
```diff
@@ -935,7 +935,7 @@ export class RightPanelPageObject {
     await expect(this.panelLoaders).toHaveCount(0, { timeout });
 
     // Step 3: Wait for any remaining loaders on the page (fallback)
-    await this.pageLoader.waitFor({ state: 'detached', timeout });
+    await this.pageLoader.waitFor({ state: 'hidden', timeout });
```
shah-harshit: Why not use await this.waitForLoadersToDisappear() here?

openmetadata-ui/src/main/resources/ui/playwright/e2e/PageObject/Explore/RightPanelPageObject.ts
```diff
@@ -1022,7 +1022,7 @@ export class RightPanelPageObject {
    * Wait for all loaders to disappear
    */
   async waitForLoadersToDisappear() {
-    await this.pageLoader.waitFor({ state: 'detached' });
+    await this.pageLoader.waitFor({ state: 'hidden' });
```
shah-harshit: Why not use await this.waitForLoadersToDisappear() here?

shah-harshit (approved):

PR #26812, opened 2026-03-27
Fixes #26785 - skip compiled-only dbt test results with null message

nr-mdakilahmed:
### Describe your changes:

Fixes #26785

When multiple dbt Cloud jobs share the same project, a `dbt run` job populates `run_results.json` with test nodes that have `status=success` but `message=null` — because no test SQL was actually executed. OpenMetadata was ingesting these compiled-only entries as real `TestCaseStatus.Success`, causing phantom duplicates and duplicate Slack alerts.

**Fix:** Add an early return in `add_dbt_test_result()` when `dbt_test_result.message` is null/empty:

```python
# Skip compiled-only entries: `dbt run` includes test nodes in
# run_results.json with status="success" but message=null since
# no test SQL was executed. Real results always have a message.
if not dbt_test_result.message:
    logger.debug(
        "Skipping compiled-only test result for '%s' (message is null).",
        manifest_node.name,
    )
    return
```

Real test results (pass/fail/warn/error) always populate the `message` field.

> **Note:** The `failures` field cannot be used as a discriminator — it is stripped by `remove_run_result_non_required_keys()` before parsing because it is absent from `REQUIRED_RESULTS_KEYS`.

**Tests:** Added `TestAddDbtTestResultSkipsCompiledOnly` with 5 unit tests covering null message (skipped), empty message (skipped), and real pass/fail/warn results (ingested).

#
### Type of change:
- [x] Bug fix

#
### Checklist:
- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [x] My PR title is `Fixes <issue-number>: <short explanation>`
- [x] I have commented on my code, particularly in hard-to-understand areas.
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

- [x] I have added a test that covers the exact scenario we are fixing.

PR #27248, opened 2026-04-10
fix(dbt): skip compiled-only dbt test results with null message (#26785)

### LINKED ISSUE #26785: [Feature] dbt ingestion: skip compiled-only test results that never executed SQL
## Is your feature request related to a problem

When multiple dbt Cloud jobs share the same project, a test node can appear in `run_results.json` from a job that only runs `dbt run` (models only). dbt includes these tests as **compiled-only** entries — no SQL is executed — but marks them `status: "success"`. The OM connector ingests this as a real `TestCaseStatus.Success`, producing phantom test results.

This causes:
- **Duplicate results** — the same test gets two entries per day (one real from the `dbt build` job, one phantom from the `dbt run` job)
- **Duplicate alerts** — each result triggers a separate change event and notification
- **Misleading test history** — the timeseries graph shows conflicting statuses (e.g. Success + Aborted for the same test)

## How to identify compiled-only entries

| Field | Real execution | Compiled-only |
|---|---|---|
| `failures` | `0`, `1`, N (integer) | `null` |
| `message` | `"Pass"`, `"Got 1 result..."` | `null` |
| `adapter_response` | `{query_id, rows_affected, ...}` | `{}` |
| `execution_time` | Seconds | < 0.1s |

## Describe the solution you'd like

In `add_dbt_test_result()` ([metadata.py#L1036](https://github.com/open-metadata/OpenMetadata/blob/1.6.2-release/ingestion/src/metadata/ingestion/source/database/dbt/metadata.py#L1036)), skip entries where no SQL was executed:

```python
# After fetching dbt_test_result, before status mapping:
if dbt_test_result.failures is None and not dbt_test_result.message:
    logger.info(
        f"Skipping compiled-only test result for node: {manifest_node.name}"
    )
    return
```

A real test always populates `failures` (even `0` on pass) and `message`. Both being `null` is a reliable signal that the test was only compiled.

## Additional context

- OM version: 1.11.8
- dbt Cloud connector, dbt Core 1.7+
- Related: #26606

aniruddhaadak80:
Fixes #26785. Prevents duplicate phantom results from DBT jobs that only ran compiled commands but didn't execute SQL.


aniruddhaadak80:
Hello! I am participating in the WeMakeDevs hackathon. Could a maintainer please assign the `safe to test` label so the GitHub Actions workflows can validate my fixes? Thank you!

aniruddhaadak80:
Hi maintainers! Just checking in to see if the `safe to test` label can be applied to initiate the workflow runs.

ulixius9:
closing by https://github.com/open-metadata/OpenMetadata/pull/26812

PR #27309, opened 2026-04-13
Fix - 27308: Mark Charts deleted

### LINKED ISSUE #27308: Charts are not deleted when deleted from source
**Affected module**
Dashboard -> Charts

**Describe the bug**
Charts are not marked as deleted/soft deleted when they're deleted from source. It's happening on all dashboard connectors

**Expected behavior**
Charts should be marked as deleted & showed in the UI while switching deleted toggle.

**Version:**
OpenMetadata version: 1.12.5

harshsoni2024:
### Describe your changes:

Fixes #27308

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:


ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
```diff
@@ -483,6 +488,20 @@ def mark_datamodels_as_deleted(self) -> Iterable[Either[DeleteEntity]]:
...
+        """
+        if self.source_config.markDeletedCharts:
+            logger.info("Mark Deleted Charts set to True")
+            yield from delete_entity_from_source(
+                metadata=self.metadata,
+                entity_type=Chart,
+                entity_source_state=self.chart_source_state,
+                mark_deleted_entity=self.source_config.markDeletedCharts,
+                params={"service": self.context.get().dashboard_service},
+            )
```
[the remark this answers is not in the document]
harshsoni2024: this is already handles by flag, & hex connector does not show any errors regarding this

ShaileshParmar11 (approved):

ingestion/src/metadata/ingestion/source/dashboard/grafana/metadata.py
```diff
@@ -224,23 +224,23 @@ def yield_dashboard_chart(
...
+                    service=FullyQualifiedEntityName(
+                        self.context.get().dashboard_service
+                    ),
+                    sourceUrl=SourceUrl(
+                        f"{clean_uri(self.service_connection.hostPort)}"
+                        f"{dashboard_details.meta.url}?viewPanel={panel.id}"
+                    ),
                 )
+                yield Either(right=chart_request)
+                self.register_record_chart(chart_request=chart_request)
```
pmbrull, line 243: should we take the chance to improve the framework a bit. I think we have to keep adding this `register_record_xyz` when we miss it for this or that entity. Maybe we should rely on the `yield Either right`  and the topology itself to say "entity registered" so we know we are not missing any connectors at all. thoughts?

harshsoni2024: this improvement makes sense. since we have to manually keep that in `each connector X for each entity(for e.g. 3 in case of dashboard)`.
We can couple this with topology manager to keep source state to maintain unchaged/changed/deleted entities & update them on ingestion based on `markDeleted` flag.
I am thinking of handling this in different pr something like this, wdyt ?
https://github.com/open-metadata/OpenMetadata/pull/27450

pmbrull (approved):

PR #27324, opened 2026-04-13
fix: preserve domain checks for feed delete authorization

atharvasingh7007:
## Summary
- preserve domain context when authorizing feed thread and post deletions
- resolve domains from the linked `about` entity, with fallback to stored thread domain ids
- only treat `hasDomain()` as a list operation when there is neither a concrete entity nor any domain context
- add regression coverage for linked-entity domains and stored thread domains

## Why
`FeedResource.deleteThread()` and `deletePost()` previously built feed policy contexts from author names alone. Those contexts returned no entity and no domains, so domain-scoped policies could be skipped during delete authorization. In practice, thread and post deletes could be authorized without the domain checks that protect the linked asset.

## Validation
- `mvn -B -f openmetadata-service/pom.xml spotless:check`
- `mvn -B -f openmetadata-service/pom.xml -Dtest=FeedResourceContextTest -DfailIfNoTests=false test`

----
## Summary by Gitar

- **API deprecation:**
  - Reject legacy announcement threads and access in `FeedResource` to enforce migration to `/v1/announcements`.
  - Removed `activeAnnouncement` filter and associated parameter from the `list` threads endpoint.


atharvasingh7007:
Addressed the Gitar feedback in the latest commit (`034a32b`):
- cache the resolved thread/about entity to avoid repeated lookups
- narrow and log unexpected resolution failures
- local validation passed with `mvn -B -f openmetadata-service/pom.xml spotless:check` and `mvn -B -f openmetadata-service/pom.xml -Dtest=FeedResourceContextTest -DfailIfNoTests=false test`

Could a maintainer please add the `safe to test` label so the full CI can run?

atharvasingh7007:
Revalidated this branch locally after rebasing onto the latest PR head and pushed `1e39981`, which only normalizes the remaining Spotless formatting in `FeedResourceContextTest`.

Local validation passed:
- `mvn -U -B -f pom.xml install -N`
- `mvn -U -B -f common/pom.xml install -DskipTests`
- `mvn -U -B -f openmetadata-spec/pom.xml install -DskipTests`
- `mvn -U -B -f openmetadata-shaded-deps/pom.xml install -DskipTests`
- `mvn -B -f openmetadata-service/pom.xml spotless:check`
- `mvn -B -f openmetadata-service/pom.xml -Dtest=FeedResourceContextTest,RuleEvaluatorTest -DfailIfNoTests=false test`
- `mvn -B -f openmetadata-service/pom.xml -DskipTests package`

The current GitHub failures still look like the same maintainer gate rather than a code failure: the workflows stop in `Verify PR labels` until the PR has `safe to test`.

Could a maintainer please add `safe to test` so the real CI jobs can run?

atharvasingh7007:
Following up to keep this active after the stale notice.

Current status from my side:
- Gitar review is approved with all 3 findings resolved.
- The branch is still mergeable.
- The visible CI failures are still stopping at `Verify PR labels` because the PR does not have `safe to test`; the jobs are not reaching the actual test/check execution.
- Local validation from the latest pushed branch passed as noted above: service Spotless, targeted `FeedResourceContextTest,RuleEvaluatorTest`, and service package.

Could a maintainer please remove the stale state and add `safe to test` when possible so the full CI can run?

PR #27327, opened 2026-04-13
fix for failure to start docker container when running docker compose up

Prototype4988:
### Describe your changes:

Fixes <issue-number>

Updated the Docker development compose files so the MySQL and PostgreSQL containers use Docker-managed named volumes for their data directories instead of bind mounts to local host paths.

I worked on this because the SQL containers were failing to start in local environments where the bind-mounted database directories were hosted on filesystems that do not support the permissions and file semantics expected by MySQL/PostgreSQL. In practice, this showed up during local setup when the database containers stayed unhealthy or exited during initialization. Using named volumes makes the setup more reliable across environments, especially with Docker Desktop, WSL, and similar host-mounted filesystem setups.

Changes made:
- Replaced the MySQL data bind mount with a named volume in `docker/development/docker-compose.yml`
- Replaced the PostgreSQL data bind mount with a named volume in `docker/development/docker-compose-postgres.yml`

Tested by:
- validating both compose files with `docker compose -f docker-compose.yml up --detach` and `docker compose -f docker-compose-postgres.yml up --detach`
- reproducing the issue locally with the bind-mounted database path
- confirming the database containers initialize successfully after switching to named volumes

#
### Type of change:
- [x] Bug fix

#
### Checklist:
- [ ] I have commented on my code, particularly in hard-to-understand areas

### DIFF
--- a/docker/development/docker-compose-postgres.yml
+++ b/docker/development/docker-compose-postgres.yml
@@ -15,6 +15,7 @@ volumes:
   ingestion-volume-dags:
   ingestion-volume-tmp:
   es-data:
+  db-data-postgres:
 services:
   postgresql:
     build:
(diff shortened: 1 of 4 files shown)


harshach (approved):

TeddyCr:
we want to keep bind mount. This is intentional.

PR #27334, opened 2026-04-14
fix: server-side column tag filtering across all pages (#25063)

### LINKED ISSUE #25063: Column tag filter works only per page instead of across entire table
**Affected module**
UI

**Describe the bug**
I have applied classification tags at the column level for a table that contains approximately 80 columns. These columns are paginated in the UI:

Page 1: First 50 columns
Page 2: Remaining ~30 columns

When I use the column filtering icon to filter columns by a specific tag, the filter appears to work only on the currently visible page. 
For example: If I apply the tag filter while on Page 1, only matching columns from Page 1 are shown.
Columns on Page 2 that have the same tag are not displayed unless I manually navigate to that page and reapply or view the filter there.

**To Reproduce**

**Expected behavior**
The column tag filter should work across the entire table, returning all columns that match the selected tag, regardless of pagination.

**Version:**
 - Python version:

**Additional context**
Slack reference-https://openmetadata.slack.com/archives/C02B6955S4S/p1766398625809029

david-mamani:
### Describe your changes:

Fixes #25063

#### Problem
The column tag filter within the table schema view is broken for large tables. When a table has many columns and is paginated server-side (e.g., 50 columns per page), the filter only applies to the **currently loaded page**. As a result, if a user filters by a specific tag while on Page 1, any columns matching that tag on Page 2 or beyond remain completely invisible, rendering the filter feature unreliable for enterprise-scale tables.

#### Root Cause
Tag filtering was being performed purely **client-side** using Ant Design's `<Table>` `onFilter` callback. This callback operates exclusively on the `tableColumns` state variable in the frontend, which only holds the localized subset of columns for the active page (the paginated slice). The client has no knowledge of columns on inactive pages, making client-side filtering incompatible with server-side pagination.

#### Solution
Migrated the entire column tag filtering mechanism to the **server-side**, delegating the filtering responsibility completely to the backend API (`TableResource`). The backend now ensures that filters are applied to the full column dataset **before** pagination kicks in.

**Backend Implementation:**
- `TableResource.java`: Added an optional `@QueryParam("tag") String tag` to all four column-related endpoints (`/{id}/columns`, `/name/{fqn}/columns`, and their respective `/search` variants).
- `TableRepository.java`: Refactored internal column retrieval. If a `tag` query parameter is provided:
  1. All column tags are eagerly populated.
  2. The full column list is filtered using the `tag`.
  3. The resulting subset is then sorted and paginated.
  (If no tag is provided, the original behavior is preserved: sort -> paginate -> populate tags for only the current page slice for performance).
- `TableUtil.java`: Introduced a new recursive utility method `columnMatchesAnyTag(Column, Set<String>)` to cleanly and comprehensively handle tag matching at any level of column nesting.

**Frontend Implementation:**
- `SchemaTable.component.tsx`: 
  - Stripped out the client-side `onFilter`.
  - Hooked into Ant Design's `onChange` event via `handleTableChange`, which updates the `activeTagFilter` state and re-fetches data passing the `tag` parameter to the backend.
  - Added a new `useEffect` that fires on component mount to retrieve **all** distinct column tags (`limit: 1000`) for the table. This guarantees the filter dropdown accurately displays every available tag across the entire dataset.
- `tableAPI.ts`: Updated `GetTableColumnsParams` to support the new optional `tag?: string` property.

**Testing:**
- Backend (`TableUtilTest.java`): Added 7 comprehensive unit tests for `columnMatchesAnyTag()` covering direct matches, recursive child matches, deep nesting (3+ levels), null/empty tag lists, no-match scenarios, and multi-tag set intersections.
- Frontend (`SchemaTable.test.tsx`): Updated existing assertions for the new API contract and added 3 new tests (e.g., verifying the all-tags fetch on mount, and ensuring `tag` is `undefined` when no filter is active).

#### Test Results

All 9 backend unit tests run successfully.
image

#### Bug Reproduction

Before this fix, attempting to filter columns by the `PII.Sensitive` tag only showed results present on the first page, ignoring identical tags on subsequent pages.

https://github.com/user-attachments/assets/930bea4d-a1da-43b6-b885-0fdb562e96aa


Rohit0301:
@david-mamani check the gitar comments, and also check the CI failures

ayush-shah:

ayush-shah:
Thanks for moving column tag filtering server-side. I found two follow-ups that still affect the “across all pages” behavior:

- The backend only populates column tags before filtering when the response fields include `tags`. If an API caller sends `?tag=...` without requesting `fields=tags`, the filter can evaluate against empty tags and return no matches. Could you populate tags whenever `hasTagFilter` is true, independent of response projection?
- The UI preloads tag options with a single `limit: 1000` column request. Tables with more than 1000 columns can still miss tag filter choices from later pages. Could we page through all columns/tags or add a lightweight tag-options endpoint?

That closes the remaining large-table/API gaps for column tag filtering.

PR #27361, opened 2026-04-14
Fix Parquet Reader to use boto3 client

harshach:
# S3 Datalake Connector — Credential Fix + Performance Optimization

## Problem 1: ACCESS_DENIED reading files from S3

  Symptom: Parquet and CSV files failed with ACCESS_DENIED / Forbidden during schema inference, even though aws s3 cp worked fine with the same credentials.

  Root cause: All S3 file readers (parquet, CSV, avro) bypassed the properly-configured boto3 client and created their own filesystem clients from raw config — missing AssumeRole, instance profiles, ECS task roles, etc.

  Fix: Each reader now uses the boto3 client/session that was already properly configured by AWSClient:

 

## Problem 2: Full file downloaded into memory for parquet

  Symptom: Parquet reader called response["Body"].read() into a BytesIO buffer, loading entire files into memory before reading.

  Fix: Uses s3fs.open() which gives PyArrow a seekable file handle — PyArrow performs ranged reads (footer + needed row groups only), no full download.

## Problem 3: Private boto3 API for credential extraction

  Symptom: Used self.client._request_signer._credentials.get_frozen_credentials() — fragile private API.

  Fix: Session is now stored in DatalakeS3Client and threaded to readers. Uses the public session.get_credentials().get_frozen_credentials() instead. Falls back to default credential chain when session is None (profiler path).

## Problem 4: N+1 HEAD requests at scale

  Symptom: list_objects_v2 already returns file Size, but readers made separate head_object / s3_fs.info() / GCSFileSystem().info() / AzureBlobFileSystem().info() calls per file. For 100K files = 100K redundant API calls.

  Fix:
  - get_table_names() now yields (key, size) tuples across all 3 cloud providers
  - Size flows through metadata.py → DatalakeTableSchemaWrapper.file_size → readers
  - Readers skip the extra API call when file_size is provided, fall back to HEAD when None

## Problem 5: GCS/Azure _get_file_size_mb created bare filesystem clients

  Symptom: _get_file_size_mb() created GCSFileSystem() without credentials and AzureBlobFileSystem() from raw config — same bypass pattern we fixed for S3.

  Fix: Uses the passed-in self.client for all providers:
  - GCS: self.client.get_bucket().get_blob().size
  - Azure: self.client.get_blob_client().get_blob_properties().size

## Problem 6: DSV IndexError on malformed CSV

  Symptom: _fix_malformed_quoted_chunk can return empty list, and [0] access would crash.

  Fix: Added if fixed: guard before accessing fixed[0].

## Problem 7: Avro loaded full file for schema extraction

  Symptom: io.BytesIO(schema_response["Body"].read()) downloaded the entire avro file just to read the header schema.

  Fix: Passes streaming response["Body"] directly to fastavro.reader() which reads the schema from the header without buffering.

## Problem 8: Misleading log when file size unknown

  Symptom: Log said "Large parquet file detected (0 bytes > 50MB)" when size check failed.

  Fix: Logs "Unknown file size. Using batched reading" when file_size == 0.

### DIFF
--- a/ingestion/src/metadata/readers/dataframe/avro.py
+++ b/ingestion/src/metadata/readers/dataframe/avro.py
@@ -108,14 +108,11 @@ class AvroDataFrameReader(DataFrameReader):
     @_read_avro_dispatch.register
     def _(self, _: S3Config, key: str, bucket_name: str) -> DatalakeColumnWrapper:
         """Stream Avro from S3 without loading entire file into memory."""
-        import io
-
         schema_response = self.client.get_object(Bucket=bucket_name, Key=key)
         try:
-            schema_buffer = io.BytesIO(schema_response["Body"].read())
+            columns = self._get_avro_columns(schema_response["Body"])
         finally:
             schema_response["Body"].close()
-        columns = self._get_avro_columns(schema_buffer)
 
         def chunk_generator():
             response = self.client.get_object(Bucket=bucket_name, Key=key)
--- a/ingestion/src/metadata/readers/dataframe/avro.py
+++ b/ingestion/src/metadata/readers/dataframe/avro.py
@@ -33,7 +33,6 @@ from metadata.generated.schema.type.schema import DataTypeTopic
 from metadata.readers.dataframe.base import DataFrameReader, FileFormatException
 from metadata.readers.dataframe.models import DatalakeColumnWrapper
 from metadata.readers.file.adls import return_azure_storage_options
-from metadata.readers.file.s3 import return_s3_storage_options
 from metadata.readers.models import ConfigSource
 from metadata.utils.constants import CHUNKSIZE
 from metadata.utils.logger import ingestion_logger
@@ -109,19 +108,21 @@ class AvroDataFrameReader(DataFrameReader):
     @_read_avro_dispatch.register
     def _(self, _: S3Config, key: str, bucket_name: str) -> DatalakeColumnWrapper:
         """Stream Avro from S3 without loading entire file into memory."""
-        from s3fs import S3FileSystem
+        import io
 
-        storage_options = return_s3_storage_options(self.config_source)
-        s3 = S3FileSystem(**storage_options)
-        file_path = f"s3://{bucket_name}/{key}"
-
-        with s3.open(file_path, "rb") as f:
-            columns = self._get_avro_columns(f)
+        schema_response = self.client.get_object(Bucket=bucket_name, Key=key)
+        try:
+            schema_buffer = io.BytesIO(schema_response["Body"].read())
+        finally:
+            schema_response["Body"].close()
+        columns = self._get_avro_columns(schema_buffer)
 
         def chunk_generator():
             response = self.client.get_object(Bucket=bucket_name, Key=key)
-            file_stream = response["Body"]
-            yield from self._stream_avro_records(file_stream)
+            try:
+                yield from self._stream_avro_records(response["Body"])
+            finally:
+                response["Body"].close()
 
         return DatalakeColumnWrapper(
             columns=columns,
--- a/ingestion/src/metadata/readers/dataframe/dsv.py
+++ b/ingestion/src/metadata/readers/dataframe/dsv.py
@@ -34,7 +34,6 @@ from metadata.generated.schema.entity.services.connections.database.datalakeConn
 from metadata.readers.dataframe.base import DataFrameReader, FileFormatException
 from metadata.readers.dataframe.models import DatalakeColumnWrapper
 from metadata.readers.file.adls import AZURE_PATH, return_azure_storage_options
-from metadata.readers.file.s3 import return_s3_storage_options
 from metadata.readers.models import ConfigSource
 from metadata.utils.constants import CHUNKSIZE
 from metadata.utils.logger import ingestion_logger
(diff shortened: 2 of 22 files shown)


ingestion/src/metadata/readers/dataframe/parquet.py
```diff
@@ -203,104 +203,33 @@ def chunk_generator():
...
+        import io
 
-        s3_fs = S3FileSystem(**client_kwargs)
+        from pyarrow.parquet import ParquetFile
 
-        bucket_uri = f"{bucket_name}/{key}"
+        def chunk_generator():
+            response = self.client.get_object(Bucket=bucket_name, Key=key)
+            try:
+                buffer = io.BytesIO(response["Body"].read())
```
TeddyCr: I think that will read the whole file in memory. That is why we are using s3fs.

```
with fs.open(f"{bucket_name}/{key}") as f:
    file_size = fs.info(f"{bucket_name}/{key}")["size"]
    parquet_file = pq.ParquetFile(f)
```

TeddyCr: we can probably just get the size like this

```
fs = s3fs.S3FileSystem()
file_size = fs.info(f"{bucket_name}/{key}")["size"]
```

ingestion/src/metadata/readers/dataframe/avro.py
```diff
@@ -109,19 +108,21 @@ def _read_avro_dispatch(
...
 
-        storage_options = return_s3_storage_options(self.config_source)
-        s3 = S3FileSystem(**storage_options)
-        file_path = f"s3://{bucket_name}/{key}"
-
-        with s3.open(file_path, "rb") as f:
-            columns = self._get_avro_columns(f)
+        schema_response = self.client.get_object(Bucket=bucket_name, Key=key)
+        try:
+            schema_buffer = io.BytesIO(schema_response["Body"].read())
```
TeddyCr: same issue here https://github.com/open-metadata/OpenMetadata/pull/27361/changes#r3080809843

ingestion/src/metadata/readers/dataframe/dsv.py
```diff
@@ -173,12 +172,31 @@ def _(self, _: GCSConfig, key: str, bucket_name: str) -> DatalakeColumnWrapper:
...
+                    sep=self.separator,
+                    chunksize=CHUNKSIZE,
+                    compression=compression,
+                    encoding_errors="ignore",
+                    escapechar="\\",
+                ) as reader:
+                    for chunks in reader:
+                        chunks = self._fix_malformed_quoted_chunk(
+                            chunk_list=[chunks], separator=self.separator
+                        )[0]
```
TeddyCr: we should guard against index error as `_fix_malformed_quoted_chunk` can return empty list from what I can see

```
# def _fix_malformed_quoted_chunk:
        if not chunk_list:
            return chunk_list
```

ingestion/src/metadata/readers/dataframe/parquet.py
```diff
@@ -200,107 +200,72 @@ def chunk_generator():
...
-            if self.config_source.securityConfig.awsAccessKeyId:
-                os.environ[
-                    "AWS_ACCESS_KEY_ID"
-                ] = self.config_source.securityConfig.awsAccessKeyId
-                os.environ[
-                    "AWS_SECRET_ACCESS_KEY"
-                ] = (
-                    self.config_source.securityConfig.awsSecretAccessKey.get_secret_value()
-                )
+        creds = self.client._request_signer._credentials.get_frozen_credentials()
```
TeddyCr: wondering if it would be better to store the session as part of 

```
# ../s3/connection.py
@dataclass
class S3ObjectStoreClient:
    s3_client: BaseClient
    cloudwatch_client: BaseClient
    session: Session
```
so we can then do `self.session.get_credentials().get_frozen_credentials()` and not rely on private/internal attributes that could change later on.

ingestion/src/metadata/readers/dataframe/parquet.py
```diff
@@ -424,7 +401,10 @@ def chunk_generator():
                 dataframes=chunk_generator, raw_data=None, columns=None
             )
 
-    def _read(self, *, key: str, bucket_name: str, **__) -> DatalakeColumnWrapper:
+    def _read(
+        self, *, key: str, bucket_name: str, file_size: Optional[int] = None, **__
+    ) -> DatalakeColumnWrapper:
+        self._file_size = file_size
```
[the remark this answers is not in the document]
harshach: @gitar-bot   1. Parquet: Retrieved as a local variable within each dispatch handler (e.g., file_size = file_info.get("size", 0)) and passed to self._should_use_chunking(file_size)
  2. JSON: Retrieved via self._get_file_size_mb(key, bucket_name) inside _read_json_smart()

  Neither reader stores file size as instance state. The concern you raised doesn't apply to the current code — file_size is always a local variable scoped to the method call, not persisted on self. The singledispatch constraint is worked around differently: file size is fetched
  inside each dispatch handler (Parquet) or inside the shared _read_json_smart method that's called by the dispatch handlers (JSON).

TeddyCr (dismissed):

ulixius9 (approved):

PR #27403, opened 2026-04-15
test(playwright): initialize domain and data product in beforeAll for RestoreEntityInheritedFields

shah-harshit:
## Summary

Moves `Domain` and `DataProduct` construction from module scope into the `test.beforeAll` hook in `RestoreEntityInheritedFields.spec.ts`.

## Why

- Keeps test setup aligned with Playwright’s lifecycle (create resources after the hook runs, alongside `performAdminLogin`).
- Avoids initializing support objects at import time, which can be harder to reason about with parallel workers and retries.

## Changes

- Declare `domain` and `dataProduct` with `let` at module scope.
- Assign `new Domain()` and `new DataProduct([domain])` at the start of the existing `beforeAll('setup test', ...)` block before API create calls.


Rohit0301 (approved):

shah-harshit:
This is not the correct fix, @shrabantipaul-collate will raise a PR for the same.

PR #27408, opened 2026-04-15
Storage Service Connect , path specification

harshach:
# Storage Connector: Wildcards in Manifest + Config-Side Default Manifest

## Context

Today, the OpenMetadata storage connector (S3/GCS/Azure) uses a bucket-level `openmetadata.json` manifest file to declare which data paths to ingest. Two gaps prompted this work:

1. **Every new table requires a manifest update.** The manifest must list each `dataPath` literally — users asked for glob patterns so a single entry can cover a dataset that grows over time.
2. **Quick setup for new buckets is painful.** Admins want to seed a working configuration centrally; bucket owners want to self-serve by uploading their own manifest later. Both flows should converge on the *same* format.

## Design

Two principles drove the design:

1. **One manifest format, two delivery channels.** The bucket's `openmetadata.json` and the new config-side `defaultManifest` use the same schema, so users learn the format once and reuse the same JSON everywhere.
2. **Bucket file always wins.** Clear precedence so self-service from bucket owners always trumps admin defaults.

### Precedence (bucket-by-bucket)

```
1. Global manifest (storageMetadataConfigSource)  — matched by containerName
2. Bucket's own openmetadata.json                 — wins over defaultManifest
3. Pipeline config defaultManifest (fallback)     — matched by containerName
```

The first source that yields entries for a given bucket is used; the others are skipped for that bucket.

## Schema changes

Only three files changed under `openmetadata-spec/`.

### `manifestMetadataConfig.json` and `containerMetadataConfig.json`

Existing fields kept (no breaking change). **Added optional fields** that only take effect when `dataPath` contains glob characters:

| Field | Type | Default | Purpose |
|---|---|---|---|
| `autoPartitionDetection` | boolean | `false` | Hive-style `key=value` partition inference from matched paths |
| `excludePaths` | `string[]` | `null` (falls back to `_delta_log`, `_temporary`, `_spark_metadata`, `.tmp`, `_SUCCESS`) | Segments to skip during glob listing |
| `excludePatterns` | `string[]` | `null` | Glob patterns to skip |
| `unstructuredData` | boolean | `false` | Each matched file becomes its own container; no schema extraction |

Also **`partitionColumns`** was slimmed from the full recursive `Column` reference to a lightweight inline shape (`name`, `dataType`, `dataTypeDisplay?`, `description?`). This unblocks RJSF form rendering and keeps the JSON payload small; the ingestion code converts to full `Column` objects when building `ContainerDataModel`.

**`dataPath` semantics** are broadened (no schema change, description updated):

- Literal path, e.g. `"data/events"` — same behavior as before.
- Glob pattern:
  - `*` — one path segment (no `/`)
  - `**` — any depth
  - `?` — single character

### `storageServiceMetadataPipeline.json`

Added one field:

```json
"defaultManifest": {
  "title": "Default Manifest (JSON)",
  "type": "string",
  "uiFieldType": "code",
  "format": "json"
}
```

Stored as a JSON string rather than a nested object for two reasons: (1) it renders as a single code editor in the UI without fighting RJSF's nested-array rendering, and (2) the "paste the same JSON you'd put in `openmetadata.json`" mental model is preserved exactly.

Backend parses the string to `ManifestMetadataConfig` at runtime (see `_parsed_default_manifest` below).

## Python implementation

All changes are in `ingestion/src/metadata/ingestion/source/storage/`.

### New helper: `expand_entry(bucket_name, entry)`

Added to the `StorageServiceSource` base class. For each manifest entry:

- **Literal `dataPath`** → passes through unchanged. Zero behavior change vs. today.
- **Glob `dataPath`** → expanded into one or more concrete `MetadataEntry` objects:
  1. Extract longest static prefix (for optimized listing).
  2. Compile glob → regex.
  3. `list_keys(bucket, prefix)` (provider-specific, already implemented for S3).
  4. Filter out excluded segments and patterns.
  5. **Unstructured mode**: yield one entry per file.
  6. **Structured mode**: group files by logical table root, optionally detect Hive partitions, yield one entry per table.

The expanded entries flow through the existing `_generate_structured_containers` / `_generate_unstructured_containers` pipeline — so container FQNs, sample-file sampling, column extraction, delete detection, and tag ingestion all work unchanged.

### New helper: `_resolve_manifest_entries(bucket_name)`

Encapsulates the precedence rules above. Returns a single `List[MetadataEntry]` that `get_containers()` then passes through `expand_entries()`.

This replaced two duplicated if/else branches in `s3/metadata.py::get_containers` — one function call now covers all three sources.

### JSON parsing for `defaultManifest`

Since `defaultManifest` arrives as a string from the UI:

```python
def _parsed_default_manifest(self) -> Optional[ManifestMetadataConfig]:
    raw = getattr(self.source_config, "defaultManifest", None)
    if not raw:
        return None
    try:
        return ManifestMetadataConfig.model_validate(json.loads(raw))
    except (json.JSONDecodeError, ValueError) as exc:
        logger.warning(f"Could not parse defaultManifest as JSON — ignoring. Error: {exc}")
        return None
```

Invalid JSON is logged and dropped (ingestion continues); bucket manifest files and global manifest remain active.

### Files touched (Python)

| File | What changed |
|---|---|
| `storage_service.py` | Added `has_glob`, `expand_entry`, `expand_entries`, `_resolve_manifest_entries`, `_parsed_default_manifest`, `_partition_columns_to_table_columns`. Updated `_manifest_entries_to_metadata_entries_by_container` to pass the new fields through. |
| `s3/metadata.py` | Replaced 3 if/else manifest branches with one call to `_resolve_manifest_entries` + `expand_entries`. `list_keys` was already present. |
| `utils/path_pattern.py` | Unchanged — `extract_static_prefix`, `pattern_to_regex`, `group_files_by_table`, `detect_hive_partitions`, `infer_structure_format` all reused. |

## UI

### Form widget

A dedicated `ManifestJsonWidget` renders `defaultManifest`:

- **Pre-populated sample JSON** on an empty field (editable, not read-only):

  ```json
  {
    "entries": [
      {
        "containerName": "my-bucket",
        "dataPath": "data/*/events/*.parquet",
        "structureFormat": "parquet",
        "autoPartitionDetection": true
      },
      {
        "containerName": "my-bucket",
        "dataPath": "logs/**/*.json",
        "structureFormat": "json"
      }
    ]
  }
  ```

- **Resizable** via native `resize: vertical` — drag the bottom-right corner (min 180px, max 80vh).
- **Live validation** with specific error messages:
  - JSON parse errors with the underlying message
  - Unknown top-level fields (anything other than `entries`)
  - Unknown entry fields with typo suggestions (e.g. `"structuredFormat" — did you mean "structureFormat"?`)
  - Missing required fields (`containerName`, `dataPath`)
  - Type mismatches (e.g. `"autoPartitionDetection" expected true or false, got string`)
  - Array field element types (`excludePaths` / `excludePatterns` must be arrays of strings)
  - `partitionColumns` item shape (each needs `name` + `dataType`)
  - Green "Valid manifest — N entries." banner when all checks pass

### Sidebar docs

One section added to `Storage/workflows/metadata.md` (Default Manifest) with a worked JSON example. All the previous sections added for the inline-manifest exploration were reverted — we're not introducing form-field-level docs.

## Backwards compatibility

**Short version: existing `openmetadata.json` files and existing pipeline configs work unchanged.**

| Scenario | Behavior |
|---|---|
| Bucket has `openmetadata.json` with literal `dataPath` | Identical to today — literal path passes through `expand_entry` untouched. |
| Bucket has `openmetadata.json`, new `autoPartitionDetection: true` added on an existing literal-path entry | Flag is ignored for literal paths (defined in code + docs). No regression risk. |
| Pipeline configures `storageMetadataConfigSource` (legacy global manifest) | Still works as before; takes highest precedence. |
| Pipeline has no `defaultManifest`, no global manifest, and bucket has no file | Same as before — only the bucket itself is cataloged (no nested containers). |
| Pipeline adds `defaultManifest` while buckets still have their own files | Bucket files win per-bucket; `defaultManifest` only kicks in for buckets *without* a manifest file. |
| Literal-path entry produces container name X today; tomorrow switched to a glob that resolves to X | Same FQN — the same entity is updated, not duplicated. Lineage/tags/descriptions preserved. |
| Invalid JSON pasted into `defaultManifest` | Logged as warning, ignored. Ingestion proceeds with bucket file / global manifest / empty. |

The `partitionColumns` schema shape changed from the full `Column` reference to a lightweight `PartitionColumn`. This required touching two existing tests (`test_s3_storage.py`, `test_gcs_storage.py`) to construct `PartitionColumn(name=..., dataType=...)` instead of `Column(...)`. Any external consumer parsing a manifest programmatically and populating `partitionColumns` needs the same minor update.

## Migration paths

No forced migration. Two optional patterns users may adopt:

1. **Condense a bulky manifest** — replace many literal entries with a single glob. Same FQNs, fewer lines.

   ```json
   // before
   [
     {"dataPath": "data/users"},
     {"dataPath": "data/orders"},
     {"dataPath": "data/events"}
   ]
   ```

   ```json
   // after
   [
     {"dataPath": "data/*", "structureFormat": "parquet"}
   ]
   ```

2. **Bootstrap without touching buckets** — admin adds `defaultManifest` in the pipeline config; ingestion starts finding data. Later, bucket owners upload their own `openmetadata.json` to customize. The moment a bucket file appears, the `defaultManifest` is ignored for that bucket.

## Verification

- **Python unit tests**: `pytest ingestion/tests/unit/topology/storage/ ingestion/tests/unit/test_path_pattern.py` → 174 pass (19 new in `test_manifest_wildcards.py` covering literal passthrough, glob expansion, partition detection, explicit vs auto partition priority, excludes, unstructured mode, JSON parsing errors, precedence).
- **Python integration tests** (`ingestion/tests/integration/s3/test_s3_manifest_wildcards.py`, runs against MinIO testcontainer — **16 tests**):
  - **Happy paths**: glob `dataPath` in a bucket manifest resolves to multiple containers; `excludePaths` filters correctly; `autoPartitionDetection` surfaces Hive columns; `defaultManifest` on pipeline config is used when the bucket has no `openmetadata.json`.
  - **Precedence**: bucket's `openmetadata.json` wins over `defaultManifest` when both are present.
  - **Backwards compat**: legacy literal-path manifest (the full `test_s3_storage` fixture) still produces the same containers.
  - **Malformed bucket manifest**:
    - Invalid JSON → logged with line/column, reported on workflow status, falls back to `defaultManifest`.
    - Invalid JSON with no `defaultManifest` → bucket container still created; no crash.
    - Schema violation (missing required field) → logged with Pydantic per-field message, falls back to `defaultManifest`.
    - `entries: []` → no containers created from bucket file; falls back to `defaultManifest`.
  - **Malformed defaultManifest**:
    - Invalid JSON → logged and ignored; bucket manifest (if any) still applies.
    - Schema violation → logged and ignored; no crash.
  - **File-read edge cases**:
    - Corrupt parquet in one table does not block other tables matched by the same glob.
    - Glob matches zero files → bucket container still created, no nested containers.                                                                                                                                                                                                        - Unknown file extension + no `structureFormat` → skipped with warning, no crash.
    - `unstructuredData: true` → each matched file becomes its own leaf container with no `dataModel`.
  - **Re-ingestion / migration guarantees**:
    - Same config run twice → same entity ID (idempotency).
    - Literal-path manifest → glob manifest that resolves to the same path → **same entity ID** (lineage / tags / descriptions preserved during migration).
  - **Per-entry resilience**:
    - One entry triggers a runtime exception mid-expansion (e.g. simulated S3 AccessDenied) → other entries in the same manifest still process and produce their containers. The failing entry is logged + reported via `status.warning` with the entry's `dataPath` and the exception type/message.
    - Literal `dataPath` containing regex-special characters (`+`, `[`, parens) is matched verbatim, not as a regex pattern.
- **Schema regeneration**: `make generate` + `json2ts-generate-all.sh` + `yarn parse-schema` all produce clean output with no orphan `manifestEntry` files.
- **End-to-end**: Docker stack (`./docker/run_local_docker.sh -m ui -d mysql`) comes up healthy; S3 storage service can be configured with a `defaultManifest` via the UI and deployed.

## Whats remaining
- ** Real S3 service test**



```

----
## Summary by Gitar

- **UI localization:**
  - Added new i18n keys for `ManifestJsonWidget` validation errors and type mismatches across all language files in `openmetadata-ui/src/main/resources/ui/src/locale/languages/`.
- **Widget improvements:**
  - Refactored `ManifestJsonWidget` to use structured validation logic, replacing hardcoded strings with localized error codes.

### CODE REVIEW

**Review by aniketkatkar97 — DISMISSED** [review]

File: ingestion/src/metadata/ingestion/source/storage/storage_service.py [hunk 1]
```diff
@@ -338,3 +411,182 @@ def _get_columns(
             container_name, sample_key, config_source, client, metadata_entry
         )
         return (metadata_entry.partitionColumns or []) + (extracted_cols or [])
+
+    def list_keys(self, bucket_name: str, prefix: str) -> Iterable[Tuple[str, int]]:
+        """List (key, size_bytes) pairs for files under prefix.
+
+        Override per provider (S3, GCS, Azure). Default returns empty.
+        """
+        return []
```
IceS2, line 498: It seems we are missing the overrides for GCS and Azure. Is this intended?

ingestion/src/metadata/ingestion/source/storage/storage_service.py
```diff
@@ -338,3 +411,182 @@ def _get_columns(
...
+            keys_scanned = 0
+            for key, size in self.list_keys(bucket_name, static_prefix):
+                keys_scanned += 1
+                if keys_scanned > MAX_KEYS_PER_PATH_SPEC:
+                    logger.warning(
+                        f"ManifestEntry '{pattern}' scanned "
+                        f"{MAX_KEYS_PER_PATH_SPEC:,} keys in bucket "
+                        f"'{bucket_name}'. Stopping to avoid excessive "
+                        f"API usage. Narrow the pattern or increase the limit."
+                    )
```
IceS2: Log implies the file limit is configurable, but it is a constant

ingestion/tests/unit/topology/storage/test_s3_auto_discovery.py
```diff
@@ -0,0 +1,826 @@
...
+                client=Mock(),
+            )
+        )
+
+        assert len(results) == 1
+        assert results[0].name == "data/users"
+
+
+class TestAutoDiscoveryPartitions:
+    """Hive-style partition auto-detection."""
```
IceS2: We should also test the autodiscovery for other supported partition types

ingestion/src/metadata/ingestion/source/storage/s3/metadata.py
```diff
@@ -716,6 +742,30 @@ def _get_sample_file_path(
...
+    def _is_excluded_artifact(key: str) -> bool:
+        """Return True if ``key`` looks like a Spark/Delta sentinel
+        artifact that must not be used for schema inference."""
+        # Segment-based checks: any path component matching one of these
+        # means the file lives under a Spark/Delta internal directory.
+        sentinel_segments = {
+            "_delta_log",
+            "_temporary",
+            "_spark_metadata",
+            ".tmp",
```
pmbrull, line 755: thinking if we should generalize these filters so we can reuse them for datalake connector too to avoid scanning unnecessary paths. same with `list_keys` above filtering out cold_classes

ingestion/src/metadata/ingestion/source/storage/s3/metadata.py
```diff
@@ -790,27 +840,71 @@ def _get_object_source_url(self, bucket_name: str, prefix: str) -> Optional[str]
...
             metadata_config = StorageContainerConfig.model_validate(content)
-            return metadata_config
-        except ReadException:
-            logger.warning(
-                f"No metadata file found at s3://{bucket_name}/{OPENMETADATA_TEMPLATE_FILE_NAME}"
+        except ValidationError as exc:
+            # Render Pydantic errors compactly — one line per bad field.
+            details = "; ".join(
+                f"{'.'.join(str(p) for p in err.get('loc', ()))}: {err.get('msg', '')}"
+                for err in exc.errors()
```
pmbrull, line 891: that's quite cool, we might want to move it here https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/models/custom_pydantic.py so we can reuse it easily

ingestion/src/metadata/ingestion/source/storage/storage_service.py
```diff
@@ -71,6 +82,29 @@
...
+# Safety limit for the number of keys scanned per glob entry.
+MAX_KEYS_PER_GLOB = 100_000
+
+# Path segments excluded by default during glob discovery (Spark/Delta internals).
+DEFAULT_EXCLUDE_PATHS = {
+    "_delta_log",
+    "_temporary",
+    "_spark_metadata",
+    ".tmp",
+    "_SUCCESS",
```
pmbrull, line 94: same comment as above, could be separate s3 util/constant to reuse anywhere

pmbrull, line 94: also was just checking how iceberg flags metadata dirs, and has a `/metadata` with files as `blabla.metadata.json` https://tower.dev/blog/how-iceberg-stores-table-data-and-metadata

might not be as simple as filtering a path as in delta since `metadata` is bit generic, but could be added as default service filter similar to _SUCCESS.*

pmbrull (dismissed):

pmbrull (approved):

ulixius9 (dismissed):

openmetadata-ui/src/main/resources/ui/src/locale/languages/ar-sa.json
```diff
@@ -2287,6 +2287,12 @@
     "username-or-email": "اسم المستخدم أو البريد الإلكتروني",
     "valid": "Valid",
     "valid-condition": "شرط صالح",
+    "valid-manifest-entry-count_0": "Valid manifest — {{count}} entries.",
+    "valid-manifest-entry-count_1": "Valid manifest — {{count}} entries.",
+    "valid-manifest-entry-count_2": "Valid manifest — {{count}} entries.",
+    "valid-manifest-entry-count_3": "Valid manifest — {{count}} entries.",
+    "valid-manifest-entry-count_4": "Valid manifest — {{count}} entries.",
+    "valid-manifest-entry-count_5": "Valid manifest — {{count}} entries.",
```
aniketkatkar97: These don't seem in sync with the en-us.json change. Also we do not have proper translations in all the language files.

openmetadata-ui/src/main/resources/ui/src/components/common/Form/JSONSchema/JsonSchemaWidgets/ManifestJsonWidget/ManifestJsonWidget.tsx
```diff
@@ -0,0 +1,402 @@
...
+      message: `Invalid JSON: ${
+        err instanceof Error ? err.message : String(err)
+      }`,
+    };
+  }
+
+  if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) {
+    return {
+      status: 'error',
+      message: 'Top-level value must be an object with an "entries" array.',
```
aniketkatkar97: I we are showing these messages in the UI, these need to be localized.

openmetadata-ui/src/main/resources/ui/src/components/common/Form/JSONSchema/JsonSchemaWidgets/ManifestJsonWidget/ManifestJsonWidget.tsx
```diff
@@ -0,0 +1,402 @@
...
+    curr[0] = i;
+    for (let j = 1; j <= n; j += 1) {
+      const cost = a[i - 1] === b[j - 1] ? 0 : 1;
+      curr[j] = Math.min(
+        curr[j - 1] + 1, // insert
+        prev[j] + 1, // delete
+        prev[j - 1] + cost // substitute
+      );
+    }
+    for (let j = 0; j <= n; j += 1) {
```
aniketkatkar97: We are not following proper naming conversions here

aniketkatkar97 (approved):

ulixius9 (approved):

PR #27414, opened 2026-04-16
Fix restrore entity inherited fields specs

shrabantipaul-collate:
### Describe your changes:
Domain inheritance via parent PATCH is immediate in the entity API but async in the search index. visitEntityPage uses the search box, so in CI the child entity page loads with a stale index document — domain and post-save data product are not yet visible.

Fix: Added pollForInheritance flag to assignDataProduct. When true (only in RestoreEntityInheritedFields.spec.ts), both the domain check and post-save data product check use expect.poll with page.reload() on each retry. page.reload() hits the entity API directly, bypassing the search index, so the committed state is always visible within a few retries.
Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/utils/common.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/utils/common.ts
@@ -436,24 +436,6 @@ export const assignDataProduct = async (
   // to fetch the current state directly from the entity API.
   pollForInheritance = false
 ) => {
-  // Returns the domain display text. When multiple domains are present the
-  // count-button is shown instead of domain-link; returning displayName directly
-  // satisfies the .toContain assertion in both the poll and non-poll paths.
-  const getDomainText = async () => {
-    const hasMultipleDomains = await page
-      .getByTestId('domain-count-button')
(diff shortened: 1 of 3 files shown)


openmetadata-ui/src/main/resources/ui/playwright/utils/common.ts
```diff
@@ -430,13 +430,39 @@ export const assignDataProduct = async (
...
+          await page.reload();
+          await waitForAllLoadersToDisappear(page);
+
+          const hasMultipleDomains = await page
+            .getByTestId('domain-count-button')
+            .isVisible();
+
+          if (hasMultipleDomains) {
+            return page.getByTestId('domain-count-button').isVisible();
+          }
```
siddhant1: @shrabantipaul-collate duplicate assertions here

siddhant1 (dismissed):
Thank you

siddhant1 (approved):

PR #27450, opened 2026-04-17
Feat: Entity auto registration in topology

harshsoni2024:
### Auto-registration in the Topology Framework
Problem:
- Every connector must manually call self.register_record_*(request) after yield Either(right=request). If a connector forgets self.register_record_, mark_*_as_deleted sees an empty source state and soft-deletes every entity for that service.

Improvement:
- Add a source_state field to NodeStage that names the attribute on the source where FQNs should be collected. The topology runner auto-registers entity FQNs into this set after a successful yield — no connector code needed.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
@@ -142,6 +142,7 @@ class DashboardServiceTopology(ServiceTopology):
                 consumer=["dashboard_service"],
                 nullable=True,
                 use_cache=True,
+                source_state="datamodel_source_state",
             )
         ],
     )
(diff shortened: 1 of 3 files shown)


ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py
```diff
@@ -174,13 +175,15 @@ class DashboardServiceTopology(ServiceTopology):
                 store_all_in_context=True,
                 clear_context=True,
                 use_cache=True,
+                source_state="datamodel_source_state",
```
pmbrull, line 178: we should not even need this, just build it on the fly based on the context/type_

PR #27478, opened 2026-04-17
Fixes #27162: table version history shows current datatype instead of historical one

### LINKED ISSUE #27162: [UI] Table view history versions does not show historical datatypes is scale is changed
**Affected module**
impact the UI

**Describe the bug**
When viewing the history of a columns datatype, a previous version shows current datatype if changed from digital(x1,y1) to digital(x2, y2) or varchar(z1) to varchar(z2) 

**To Reproduce**

-- DECIMAL(9,1) ->  DECIMAL(15,3)
ALTER TABLE DWH_test.omd01.raw_orders
ALTER COLUMN order_amount DECIMAL(15,3);

Image

**Expected behavior**
IU shows historical datatypes.

Version:

OS: GNU/Linux
Python version: 3.9.25
OpenMetadata version: [1.12.3]
OpenMetadata Ingestion package version: [e.g. openmetadata-ingestion[docker]==1.23.3]

**Additional context**

miantalha45:
## Problem

When viewing an older version of a table, the column type column always
showed the current datatype (e.g. `decimal(15,3)`) even if that version
had a different type (e.g. `decimal(9,1)`).

## Root Cause

`TableVersion` was fetching columns via `getTableColumnsByFQN`, which
hits the current table columns API. This meant `tableColumns` always
held the latest data regardless of which version was being viewed. The
`changeDescription` for the selected version was then applied on top of
wrong base data, so the historical type was never shown correctly.

## Fix

Removed the live API call and replaced it with local pagination over
`currentVersionData.columns`. The version API already returns the full
historical entity snapshot including columns, so no extra network call
is needed. Search filtering is also done locally.

## Test

Added a unit test that verifies `VersionTable` receives columns with
the historical `dataTypeDisplay` value from `currentVersionData`, not
from any API response.

## Proof
image
image

Fixes #27162

----
## Summary by Gitar

- **Component updates:**
  - Added `onLoadMore`, `hasMore`, and `isLoadingMore` props to `TableVersion` component.
  - Integrated these new props into the `EntityVersionTimeLine` component to support version pagination.


ayush-shah:

PR #27485, opened 2026-04-17
Add new "textToLLMContext" field to improve embeddings

lautel:
Closes https://github.com/open-metadata/ai-platform/issues/303

----
## Summary by Gitar

- **Embedding architecture:**
  - Split `textToEmbed` into `textToLLMContext` (legacy format for agents) and a new, clean `textToEmbed` for semantic vector search.
  - Implemented a `buildSemanticMetaLightText` generator that excludes structural noise (FQN, system fields) to improve search relevance.
- **Refactored builder logic:**
  - Replaced reflection-based child lookup with type-safe method references in `SEMANTIC_CHILDREN_SPECS`.
  - Introduced `SEMANTIC_ENRICHERS` map to replace `instanceof` branching for entity-specific metadata.
- **Schema updates:**
  - Added `textToLLMContext` to all elasticsearch index mappings to ensure backward compatibility for tooling.


openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorDocBuilder.java
```diff
@@ -280,6 +308,250 @@ static String buildBodyText(EntityInterface entity, String entityType) {
...
+  }
+
+  /**
+   * Produce a "Contains X, Y, Z" phrase listing the names of a container entity's direct
+   * children (database schemas, tables, endpoints, charts, etc.). Children are read via
+   * reflection using the getter name in {@link #SEMANTIC_CHILDREN_SPECS}, so this does not
+   * introduce compile-time coupling to every container type. Returns null when the entity is
+   * not a known container, when the getter is missing, or when the child list is empty.
+   */
+  static String buildChildContextPhrase(EntityInterface entity, String entityType) {
```
pmbrull, line 512: cc @mohityadav766

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorDocBuilder.java
```diff
@@ -280,6 +308,250 @@ static String buildBodyText(EntityInterface entity, String entityType) {
     return String.join("; ", bodyParts);
   }
 
+  /**
+   * Natural-language metadata for the semantic embedding input. Emits content as sentence-like
+   * phrases without {@code key: value;} label scaffolding, and drops high-noise/low-signal fields
+   * (FQN, entityType, serviceType, owners, customProperties, chunk marker) so the pooled vector
+   * isn't dominated by structural tokens that appear in every document.
+   */
+  static String buildSemanticMetaLightText(EntityInterface entity, String entityType) {
```
mohityadav766, line 317: @lautel can we introduce a SemanticMetaLightTextFactory here that uses Entity Type to determine the code path. This if else increases Cognitive Complexity.

mohityadav766: this common logic can then delegate to factory with required things.

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorDocBuilder.java
mohityadav766: @lautel we can use factory here  as alternate to SemanticChildrenSpec , it won't scale well.

mohityadav766 (approved):

PR #27486, opened 2026-04-17
Fixes #27482: Fix Advanced Search between boundaries for custom numerical properties

### LINKED ISSUE #27482: [Bug] Advanced Search: Number custom property "between" filter returns all entities that have the property set, ignoring the range bounds; sometimes search API is not called

aniruddhaadak80:
### Describe your changes:

Fixes #27482

I worked on [QueryBuilderElasticsearchFormatUtils.js](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-ui/src/main/resources/ui/src/utils/QueryBuilderElasticsearchFormatUtils.js) because the RAQB `between` boundaries (i.e. `[10, 20]`) were incorrectly stripped of the upper bound during `buildExtensionQuery`. This passes the entire bounding Array correctly, repairing Advanced Search filtering.

#
### Type of change:
- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Refactored type resolution:**
  - Introduced `getFieldTypeInfoFromOmType` and `lookupOmPropertyType` to use registry configuration for unambiguous field mapping.
- **Improved precision:**
  - Updated field identification logic from `includes()` to `endsWith()` to prevent misclassification of properties containing dots in their names.
  - Enhanced numeric value detection in `buildExtensionQuery` by leveraging `omPropertyType` metadata instead of relying solely on value-shape parsing.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/MyData/Widgets/KPIWidget/KPIWidget.component.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/MyData/Widgets/KPIWidget/KPIWidget.component.tsx
@@ -340,7 +340,7 @@ const KPIWidget = ({
                 interval="preserveStartEnd"
                 tick={{ fill: '#888', fontSize: 12 }}
                 tickFormatter={(value: number) =>
-                  customFormatDateTime(value, 'dMMM, yy')
+                  customFormatDateTime(value, 'd MMM, yy')
                 }
                 tickLine={false}
                 tickMargin={10}


aniruddhaadak80:
hi maintainers could someone please add the `safe to test` label so the pipelines can run, thank you

aniruddhaadak80 (commented):
Added explicit unit test for advanced-search extension numeric properties with the between operator as requested. Synced up with main.

PR #27497, opened 2026-04-17
Fixes #27482: Advanced Search between filter for number custom properties

### LINKED ISSUE #27482: [Bug] Advanced Search: Number custom property "between" filter returns all entities that have the property set, ignoring the range bounds; sometimes search API is not called

himanshu748:
### Describe your changes:

Fixes #27482

Advanced Search was incorrectly handling nested array values for range operators on custom properties. The Query Builder represents `between` values as `[[from, to]]`, but the formatter treated any nested array as a multiselect and split it into multiple rules. This dropped the upper bound and produced an effectively unbounded query.

This change:
- Only expands nested arrays into multiple rules for multiselect operators
- Passes the full `[from, to]` bounds through to the customPropertiesTyped range builder

**Testing**
- Added unit coverage asserting the generated Elasticsearch query includes both `gte` and `lte` for a number custom property `between` filter.

#
### Type of change:
- [x] Bug fix

#
### Checklist:

#
### Hackathon / AI disclosure
Built with Cursor (AI-assisted) and manually reviewed.


himanshu748:
Hi! This PR is ready for CI. Could a maintainer please add the `safe to test` label so the workflows can run? (They currently fail at the PR-label gate.) Thanks!

himanshu748:
Addressed Copilot review notes in the new unit test: (1) import `SearchOutputType` from `AdvanceSearchProvider.interface` (app source), and (2) inject a minimal `extension.table.myNumberProperty` field into the QB config so `QbUtils.checkTree()` keeps the rule.

ayush-shah:

himanshu748:
@ayush-shah I have rebased the branch against the latest `main` and resolved all conflicts. Please trigger the CI workflows when you have a moment. Thanks!

PR #27499, opened 2026-04-17
Perf/redis cache metrics and indexes

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Cache Reliability:**
  - Added `cacheNameKey` normalization to handle case-insensitive user FQNs, ensuring consistency between DB lookups and Guava cache keys.
- **Performance/Stability:**
  - Optimized `RedisCacheProvider` future handling by explicitly cancelling timed-out pipeline operations to prevent memory leaks in the Lettuce event loop.
  - Updated batch error reporting to track and log specific cancellation metrics during Redis pipeline failures.


openmetadata-service/src/main/java/org/openmetadata/service/cache/RedisCacheProvider.java
```diff
@@ -28,155 +36,275 @@ public RedisCacheProvider(CacheConfig config) {
...
+          futures.add(asyncCommands.expire(e.getKey(), ttl.getSeconds()));
+        }
+      }
+      awaitAll(futures);
+      if (m != null) {
+        for (int i = 0; i < keyFields.size(); i++) m.recordWrite();
+      }
+    } catch (Exception e) {
+      if (m != null) m.recordError();
+      LOG.error("Error on pipelineHset (batch={})", keyFields.size(), e);
```
mohityadav766, line 321: **`available` never flips on runtime command failure** — highest-impact open issue.

This catch (and the equivalents in `get`/`set`/`awaitAll`) logs the error but leaves the `available` flag unchanged. `available` is only set `true` at init and `false` in `close()`. After a Redis outage or network partition:
- `provider.available()` keeps returning `true`.
- `EntityResource.isDistributedCacheEnabled()` keeps serving cached reads.
- Per-instance Guava L1 entries stay hot and diverge across replicas in multi-instance deployments.

This defeats the very guard `isDistributedCacheEnabled()` was added to provide. Suggest flipping `available=false` on command failure (with a health-check/reconnect loop to flip it back), or gate it on a `lastFailureTimestamp` with short grace.

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/cache/CacheWarmupApp.java
```diff
@@ -103,476 +106,221 @@ public void init(App app) {
...
-        }
-        // Note: Not counting skipped entities (deleted, invalid) as failures
+        cacheProvider.pipelineHset(hsetBatch, ttl);
+        cacheProvider.pipelineSet(setBatch, ttl);
       } catch (Exception e) {
-        LOG.debug("Error warming up entity {} {}: {}", entityType, entity.getId(), e.getMessage());
-        failedCount++;
+        LOG.warn("Pipelined write failed for {} batch at offset {}", entityType, offset, e);
       }
+      success += pageSuccess;
```
mohityadav766: **Pipeline failure is counted as success.**

`pipelineHset` / `pipelineSet` are wrapped in try/catch (lines 241–246) but the catch only logs. Then `success += pageSuccess` at line 247 runs unconditionally, and `updateEntityStats(entityType, pageSuccess, pageFailed)` at line 250 reports rows as warmed even though Redis rejected them.

Move the `success += / updateEntityStats(...)` call into the try-block, or on exception treat the whole page as failed (`failed += page.size()` / `updateEntityStats(entityType, 0, page.size())`). Otherwise success counts and WebSocket progress lie on a broken Redis connection.

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/cache/CacheWarmupApp.java
```diff
@@ -103,476 +106,221 @@ public void init(App app) {
...
-      initializeCacheComponents();
-      initializeJobData(jobExecutionContext);
-      runCacheWarmup(jobExecutionContext);
-    } catch (Exception ex) {
-      handleExecutionException(ex);
+      initCacheComponents();
+      initJobData(jobExecutionContext);
+      if (cacheProvider == null || !cacheProvider.available()) {
+        LOG.warn("Cache not available, skipping warmup");
+        return;
```
mohityadav766, line 119: **Cache-unavailable path leaves job as RUNNING.**

`initJobData` has already set `EventPublisherJob.Status.RUNNING`; returning here without updating it means the `finally` block at line 124 will `sendUpdates(..., true)` broadcasting a RUNNING status that never terminates.

Set a terminal status (e.g. `FAILED`, or a dedicated SKIPPED) before returning:
```java
if (cacheProvider == null || !cacheProvider.available()) {
  LOG.warn("Cache not available, skipping warmup");
  jobData.setStatus(EventPublisherJob.Status.FAILED);
  return;
}
```

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/cache/CacheWarmupApp.java
```diff
@@ -103,476 +106,221 @@ public void init(App app) {
...
-    try {
-      while (!stopped && (!producersDone.get() || !taskQueue.isEmpty())) {
+    int offset = 0;
+    int success = 0;
+    int failed = 0;
+    long start = System.currentTimeMillis();
+    while (!stopped) {
+      List<String> page;
+      try {
+        page = dao.listAfterWithOffset(batchSize, offset);
```
mohityadav766, line 225: **`listAfterWithOffset` pagination is non-deterministic.**

`EntityDAO.listAfterWithOffset` is `SELECT json FROM <table> LIMIT :limit OFFSET :offset` with no `ORDER BY`. Offset-based pagination without a deterministic order can skip or duplicate rows between pages depending on the DB query plan — especially with inserts/deletes concurrent with the warmup scan. Since this app aims to touch every row for a type, non-determinism here means some rows never get warmed and others get written twice.

Either add `ORDER BY id` to `listAfterWithOffset`, or switch to the cursor-based `listAfter` (already `ORDER BY name, id`).

openmetadata-service/src/main/java/org/openmetadata/service/cache/RedisURIFactory.java
```diff
@@ -0,0 +1,60 @@
...
+    RedisURI.Builder builder;
+    if (url.startsWith("redis://") || url.startsWith("rediss://")) {
+      RedisURI parsed = RedisURI.create(url);
+      // Carry the scheme's SSL flag forward — rediss:// must keep TLS even if useSSL is unset.
+      builder =
+          RedisURI.Builder.redis(parsed.getHost(), parsed.getPort())
+              .withTimeout(connectTimeout)
+              .withSsl(parsed.isSsl());
+    } else if (url.contains(":")) {
+      String[] parts = url.split(":");
```
mohityadav766: **`url.split(":")` breaks IPv6 and malformed input.**

For a bracketless IPv6 host like `fe80::1:6379` this splits into multiple parts and throws on `Integer.parseInt(parts[1])`. Low likelihood in practice but a hard startup failure from config.

Either:
```java
String[] parts = url.split(":", 2);
if (parts.length < 2) throw new IllegalArgumentException("Invalid Redis URL: " + url);
```
or just normalize to `redis://` and let `RedisURI.create` do the parsing.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/PersonaRepository.java
```diff
@@ -109,7 +109,15 @@ private List<EntityReference> getUsers(Persona persona) {
...
   @Transaction
   private void unsetExistingDefaultPersona(String newDefaultPersonaId) {
+    // Capture the affected persona IDs *before* the bulk update so we can still identify them.
+    // The bulk update rewrites JSON directly — bypassing invalidateCachesAfterStore — so every
+    // affected persona would keep stale "default=true" in CACHE_WITH_ID / Redis base entries.
+    List<String> affected =
+        daoCollection.personaDAO().findOtherDefaultPersonaIds(newDefaultPersonaId);
     daoCollection.personaDAO().unsetOtherDefaultPersonas(newDefaultPersonaId);
+    for (String id : affected) {
+      invalidateCacheForEntity(Entity.PERSONA, UUID.fromString(id), null);
```
mohityadav766: **`invalidateCacheForEntity(..., null)` leaves by-name caches stale.**

Passing `fqn=null` only evicts ID-keyed cache entries. By-name Guava/Redis entries (used by GET-by-name) will keep returning `default=true` for personas whose JSON was just rewritten by the bulk update. Stale values persist until TTL.

Same pattern exists in:
- `DataProductRepository.java:917` and `:959`
- `DomainRepository.java:513`

Suggest fetching the FQN before the bulk update (or parsing it from `record.getJson()` for the DataProduct/Domain cases where `findTo` already returns the json payload) so both ID and name variants are evicted.

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/cache/CacheWarmupApp.java
```diff
@@ -103,476 +106,221 @@ public void init(App app) {
...
-      signalConsumersToStop(numConsumers);
-      waitForConsumersToComplete(consumerLatch);
-    } catch (InterruptedException e) {
-      stopped = true;
-      Thread.currentThread().interrupt();
-      throw e;
-    } finally {
-      cleanupExecutors();
+  private void warmupEntityType(String entityType, int batchSize, Duration ttl) {
+    if ("user".equals(entityType)) {
```
mohityadav766: Minor: the `"user"` literal is duplicated at line 267 in `resolveEntityTypes`. Extract to a constant (or reuse `Entity.USER`) per the project's no-magic-strings guidance.

mohityadav766 (commented):
## Redis PR Review

Direction is sound and most prior bot/Copilot feedback has been addressed. Calling out what's still open vs. resolved, plus a few observations.

### Resolved in current HEAD
- `EntityDAO.listIdFqnByPrefixHash` now has MySQL + Postgres variants via `@ConnectionAwareSqlQuery`.
- `RedisURIFactory` carries the `rediss://` SSL flag forward.
- `CachedReadBundle` replaces the `Thread.sleep` polling loop with a Guava `Striped<Lock>` (also resolves the "buildBundleDto returning null leaves waiters spinning" edge case).
- `CacheWarmupApp` stats now pass per-page deltas, not cumulative.
- `DeadlockRetry` Javadoc corrected to "retries are synchronous".
- `TestCaseRepository.addAllTestCasesToLogicalTestSuite` wraps the `@Transaction` method from outside so each retry runs in a fresh txn.
- `EntityResource.isDistributedCacheEnabled()` now gates on `provider.available()`.
- `RedisCacheProvider.awaitAll` uses a single `LettuceFutures.awaitAll` timeout and logs + chains the first underlying failure.
- `CacheInvalidationPubSub` delegates URI construction to `RedisURIFactory`.

### Still open
See inline comments. The three that should block merge (in order of blast radius):
1. `RedisCacheProvider.available` never flips to `false` on runtime command failures — after a network partition, `isDistributedCacheEnabled()` keeps per-instance caches hot and serves stale reads in multi-instance deployments.
2. `CacheWarmupApp` counts rows as "warmed" even when the Redis pipeline write threw — stats and success reporting lie.
3. `CacheWarmupApp` leaves the job in `RUNNING` status when the cache is unavailable at startup.

The rest are cleanup: `listAfterWithOffset` has no `ORDER BY`, `RedisURIFactory.url.split(":")` breaks on IPv6, and `invalidateCacheForEntity(..., null)` in Persona/DataProduct/Domain still leaves by-name cache entries stale until TTL.

### Observation: invalidation test coverage
20+ repositories now sprinkle `invalidateCacheForEntity` around ad-hoc bulk DAO rewrites. This is fragile to maintain by inspection — worth a centralized integration test that exercises `bulk update → GET-by-id and GET-by-name both return fresh` per entity type, so future regressions get caught automatically.

pmbrull (approved):

Issue #27307, opened 2026-04-13
Live Indexing Improvements


mohityadav766:
Live Indexing Needs Improvement, following things needs to be handled.
1. We should Batch the db queue. Rather than frequently hitting database use a queue to flush once number of entries are acquired.
2. If the ElasticSearch/OpenSearch Is not available , there is no point in retrying events from the queue, only retry when search instance and availbale.
3. When reindex via search index app is triggerd , as part of preflight check we should cleanup this tables. Since SearchIndex run takes care of putting the complete updated data from database to search
4. Also think of a approach when es and os are not available, we might bombard the table with too many entries, we can create a certaing type of failure to retry from change event table if that is the case, to replay those changes

karthik120710:
hi @mohityadav766 i would like to work on these issue can u assign it to me.

mohityadav766:
@karthik120710 assigned to you

mohityadav766:
@karthik120710  take a look at the issue requirements

karthik120710:
@mohityadav766 thanks for providing detailed information about these issue

asadjan4611:
Hey @mohityadav766 .
I validated this and agree with the issue. Retry consumption already backs off when ES/OS is down, but queue insertion is still per-event and can overload DB; also reindex preflight does not clear live retry queue entries.
I suggest a phased fix: batched/deduped queue writes + preflight retry-queue cleanup first, then degraded-mode with change_event replay fallback for prolonged outages.
I would love to work on it .please assign this issue to me.i am ready to raise a PR on it if you allow .

karthik120710:
@asadjan4611 I am already work on these issue .

karthik120710:
@mohityadav766 can u review my pr

PR #27505, opened 2026-04-18
Issue#27307

karthik120710:
I improved the live search indexing pipeline to reduce database write amplification
  and improve recovery behavior during ES/OS outages.

  **Changes made:**

  - **`SearchIndexRetryQueue`** — replaced per-failure DB upsert with an in-memory
    `ConcurrentLinkedQueue` buffer. A daemon flusher drains it every 500 ms or when
    50 entries accumulate, writing them in a single `@SqlBatch` upsert. Added a new
    `SEARCH_UNAVAILABLE` status so entries written during ES/OS downtime are
    distinguishable from real mapping/data failures. Buffer is capped at 2000 entries
    to prevent memory exhaustion; overflow is dropped and counted via a Micrometer
    counter.

  - **`SearchIndexRetryWorker`** — drives the flusher lifecycle (`startFlusher` on
    start, `stopFlusher` on stop to ensure a final flush before shutdown). Tracks
    ES/OS availability transitions: when the cluster recovers, bulk-resets all
    `SEARCH_UNAVAILABLE` rows back to `PENDING` for immediate retry. Added
    `STATUS_SEARCH_UNAVAILABLE` to `PURGEABLE_QUEUE_STATUSES` so it is cleaned up
    when a full reindex suspends streaming.

  - **`ReindexingOrchestrator`** — added `cleanupRetryQueuePreFlight()` called from
    `preflightFixes()`. For a full reindex it deletes all purgeable statuses; for a
    partial reindex it deletes only rows matching the selected entity types. This
    prevents the retry worker from racing against the reindex job.

  - **`CollectionDAO.SearchIndexRetryQueueDAO`** — added `batchUpsert()` (`@SqlBatch`
    with MySQL/Postgres variants), `resetSearchUnavailableToPending()`, and
    `deleteByEntityTypes()`.

  **No schema migration needed** — `status` column is `VARCHAR(32)` with no check
  constraint; `SEARCH_UNAVAILABLE` fits without any DDL change.

  **Testing:** Added unit tests covering buffer enqueue/flush behavior, overflow
  protection, status assignment, flusher lifecycle, availability transition resets,
  and preflight cleanup logic.
 #27307 

Pasted Graphic

----
## Summary by Gitar

- **Database/DAO optimizations:**
  - Optimized container listing performance in `CollectionDAO` by replacing inefficient `LEFT JOIN` subqueries with `NOT EXISTS` anti-joins.
  - Added `findContainerSummariesByIds` for lightweight projection, avoiding heavy JSON deserialization during UI list operations.
- **Search index architecture:**
  - Streamlined `SearchIndexRetryQueue` and `SearchIndexRetryWorker` by removing redundant methods and simplifying internal logic.
  - Reduced database overhead by implementing batch upsert strategies for index retries.


mohityadav766:
@karthik120710 there are integration test failures can you check 
Also please add the integration test for scenarios that are being fixed

karthik120710:
Hi @mohityadav766 can u review my pr

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -11001,6 +11012,17 @@ void upsert(
         @Bind("status") String status,
         @Bind("entityType") String entityType);
 
+    default void batchUpsert(List<BatchUpsertEntry> entries) {
+      for (BatchUpsertEntry entry : entries) {
```
mohityadav766, line 12353: why not use batch upsert here ?

ayush-shah:

PR #27269, opened 2026-04-11
feat: implement Content Security Policy nonce handling for enhanced security

chirag-madlani:
Fixes: [add support for csp nonce to strengthen app security](https://github.com/open-metadata/openmetadata-collate/issues/3631#top)

This pull request implements dynamic Content Security Policy (CSP) nonces for enhanced security of inline scripts in the OpenMetadata web UI. It introduces a middleware that generates a unique nonce per request, injects it into the CSP header and HTML, and updates all relevant places to use the nonce. This helps prevent cross-site scripting (XSS) attacks by ensuring only scripts with the correct nonce are executed.

The most important changes are:

**Security: Dynamic CSP Nonce Implementation**
* Added a new `CspNonceHandler` middleware that generates a random nonce for each request, attaches it to the request, and ensures the CSP header uses this nonce (`openmetadata-service/src/main/java/org/openmetadata/service/security/CspNonceHandler.java`).
* Updated the CSP policy in `openmetadata.yaml` to include a nonce placeholder (`__CSP_NONCE__`) in the `script-src` directive, and ensured this placeholder is replaced with the generated nonce in responses (`conf/openmetadata.yaml`, `OMErrorPageHandler.java`). [[1]](diffhunk://#diff-4e42f64415456ed49e92fdec43efb2eb7e5918e4b46a8f73bf553cfc655a9880L691-R692) [[2]](diffhunk://#diff-9479e5bc314742b15534f5d47ca704a979902668310012898db1eadb85b9d720L44-R45) [[3]](diffhunk://#diff-9479e5bc314742b15534f5d47ca704a979902668310012898db1eadb85b9d720R54-R58) [[4]](diffhunk://#diff-9479e5bc314742b15534f5d47ca704a979902668310012898db1eadb85b9d720L93-R106)

**Web UI Integration**
[...]

PR #27532, opened 2026-04-20
Fixes #27091: Close SAML FileInputStream and HttpURLConnection resources

### LINKED ISSUE #27091: [Bug] Resource leaks in SamlSettingsHolder, SamlValidator, and IndexResource — unclosed streams and connections
**Affected module**
Backend (openmetadata-service)

**Describe the bug**

Three resource leak bugs across `openmetadata-service` where streams and connections are opened but never closed, leaking file descriptors and TCP sockets.

### Bug 1: SamlSettingsHolder — FileInputStream never closed

In `SamlSettingsHolder.java`, a `FileInputStream` is created inline and passed directly to `keyStore.load()` without being assigned to a variable or wrapped in try-with-resources:

```java
KeyStore keyStore = KeyStore.getInstance("JKS");
keyStore.load(
    new FileInputStream(securityConfig.getKeyStoreFilePath()),  // ← NEVER CLOSED
    securityConfig.getKeyStorePassword().toCharArray());
```

Since the stream is created inline, it cannot be closed later — the reference is lost immediately. This leaks a file descriptor every time SAML settings are loaded.

### Bug 2: SamlValidator — HttpURLConnection never disconnected

In `SamlValidator.validateIdpConnectivity()`, an `HttpURLConnection` is opened to test the IdP SSO URL but `conn.disconnect()` is never called in any code path:

```java
HttpURLConnection conn = (HttpURLConnection) url.openConnection();
conn.setRequestMethod("GET");
conn.setConnectTimeout(5000);
conn.setReadTimeout(5000);
int responseCode = conn.getResponseCode();
// ... multiple if/else branches processing the response ...
// ← conn.disconnect() NEVER called in ANY code path
```

Additionally, `readResponseSnippet()` reads from the connection's error/input stream without closing it. Every SAML IdP connectivity validation leaks a TCP socket.

### Bug 3: IndexResource — InputStream and BufferedReader never closed

In `IndexResource.java`, both the constructor and the static `getIndexFile()` method open an `InputStream` and wrap it in a `BufferedReader`, but neither is ever closed:

```java
// Constructor (same pattern in getIndexFile):
InputStream inputStream = getClass().getResourceAsStream("/assets/index.html");
indexHtml = new BufferedReader(new InputStreamReader(inputStream))
    .lines()
    .collect(Collectors.joining("\n"));
// ← BOTH InputStream and BufferedReader NEVER CLOSED
```

Additionally, there's no null check — if `/assets/index.html` is missing from the classpath, this throws an unhelpful NPE.

**To Reproduce**

Bug 1:
1. Configure SAML authentication with a JKS keystore
2. Each SAML settings load/reload leaks one OS file descriptor
3. Under repeated SAML reconfigurations, file descriptor count grows

Bug 2:
1. Configure SAML SSO validation
2. Each call to `validateIdpConnectivity()` leaks one TCP socket
3. Under repeated SAML validation (e.g., testing configurations), socket count grows

Bug 3:
1. Start the OpenMetadata server
2. Both the `IndexResource` constructor and any call to `getIndexFile()` leak classpath streams
3. Minor impact since these run at startup, but violates resource management best practices

**Expected behavior**

All three should use try-with-resources to ensure streams and connections are properly closed:
- Bug 1: `FileInputStream` wrapped in try-with-resources before passing to `keyStore.load()`
- Bug 2: `HttpURLConnection` wrapped in try-finally with `conn.disconnect()` in the finally block
- Bug 3: `InputStream` and `BufferedReader` wrapped in try-with-resources with a null check

These patterns are already used correctly elsewhere in the codebase (e.g., `SchemaFieldExtractor.getAllEntityTypes()`, `AutoPilotApp`, `JsonLdTranslator`, `OntologyLoader`).

**Version:**
- OpenMetadata version: main branch (latest)

**Additional context**

This is a continuation of resource leak cleanup work from PR #27063, which fixed similar `InputStream` leaks in `SchemaFieldExtractor.loadMainSchema()` and `SchemaFieldExtractor.loadSchema()`. The same try-with-resources pattern applies here.

Related prior work:
- PR #27063 — Fixed InputStream resource leaks in SchemaFieldExtractor (2 locations)
- PR #27062 (Issue) — Reported silent exception swallowing + resource leaks in backend

hyspacex:
### Describe your changes:

Fixes #27091

Addresses two of the three resource leaks reported in the issue:

**Bug 1 — `SamlSettingsHolder.initDefaultSettings` leaked a `FileInputStream`.**
The keystore stream was created inline in `keyStore.load(new FileInputStream(path), password)` and the reference was lost immediately, so it could never be closed. Wrapped it in try-with-resources.

**Bug 2 — `SamlValidator.validateIdpConnectivity` leaked an `HttpURLConnection`, and `readResponseSnippet` leaked the response `InputStream`.**
Lifted `conn` above the try block and added a `finally { if (conn != null) conn.disconnect(); }`. Wrapped the response stream in `readResponseSnippet` in try-with-resources.

**Bug 3 — `IndexResource`** was already fixed in #27269 (CSP nonce handling) — the current `IndexResource.java` now reads `/assets/index.html` once into a `static final` field inside a try-with-resources block with a null check. Nothing to do here.

#### Tests
- `SamlValidatorTest`: added `readResponseSnippetClosesReadableStreams` and `validateIdpConnectivityDisconnectsConnections`, plus a mock `URLStreamHandler` for the `mock://` protocol so we can inject a tracking `HttpURLConnection` without real network I/O.
- `SamlSettingsHolderTest` (new): writes a real empty JKS to `@TempDir` and asserts the keystore loads into `samlData[KEYSTORE_KEY]`. Proves the fix end-to-end without mocking `FileInputStream` construction.
- Local run: 63/63 tests pass across `SamlValidatorTest` + `SamlSettingsHolderTest`. `mvn spotless:apply` reports no changes needed.

#
### Type of change:
- [x] Bug fix

#
### Checklist:
- [x] For JSON Schema changes: I updated the migration scripts or explained why it is not needed. _(no schema changes)_

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/security/auth/validator/SamlValidator.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/security/auth/validator/SamlValidator.java
@@ -4,6 +4,7 @@ import static org.openmetadata.common.utils.CommonUtil.nullOrEmpty;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import java.io.ByteArrayInputStream;
+import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
@@ -281,6 +282,7 @@ public class SamlValidator {
   }
 
   private FieldError validateIdpConnectivity(SamlSSOClientConfig samlConfig) {
+    HttpURLConnection conn = null;
     try {
       String ssoUrl = samlConfig.getIdp().getSsoLoginUrl();
       LOG.debug("Testing IdP SSO URL with SAML request: {}", ssoUrl);
@@ -293,7 +295,7 @@ public class SamlValidator {
           ssoUrl + (ssoUrl.contains("?") ? "&" : "?") + "SAMLRequest=" + samlRequest;
 
       URL url = new URL(urlWithParams);
-      HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+      conn = (HttpURLConnection) url.openConnection();
       conn.setRequestMethod("GET");
       conn.setConnectTimeout(5000);
       conn.setReadTimeout(5000);
@@ -340,6 +342,10 @@ public class SamlValidator {
       // Warning case - treat as success since URL format might be valid
       LOG.warn("SSO URL validation warning: {}", e.getMessage());
       return null;
+    } finally {
+      if (conn != null) {
+        conn.disconnect();
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


ayush-shah:

PR #27555, opened 2026-04-20
fix:  commen-> comment

### LINKED ISSUE #27482: [Bug] Advanced Search: Number custom property "between" filter returns all entities that have the property set, ignoring the range bounds; sometimes search API is not called

ClawdiaHedgehog:
Fixed typos:  commen-> comment

Closes #27482


ulixius9:
closing this pr as the current spelling already looks correct

PR #27587, opened 2026-04-21
Fix thread-safety, NPE/AIOOBE, domain ACL bypass, and Pydantic RootModel serialization bugs

kalyangupta12:
## Summary

This PR fixes 5 distinct bugs across the Java backend, Python ingestion framework, and migration utilities. Each fix addresses a real defect with observable incorrect behavior — not cosmetic cleanup.

---

### 1. Thread-safety: Broken singleton pattern in all 6 secrets managers

**Files:** `GCPSecretsManager`, `AWSSecretsManager`, `AWSSSMSecretsManager`, `AzureKVSecretsManager`, `KubernetesSecretsManager`, `DBSecretsManager`, `InMemorySecretsManager`

**Bug:** All `getInstance()` methods used a non-volatile `instance` field with a simple `null` check. Under the Java Memory Model, the JIT can reorder the write to `instance` before the object's constructor completes. A second thread could receive a partially-constructed manager and call methods on it.

**Fix:** Add `volatile` to the `instance` field and use the double-checked locking pattern so the synchronization is only paid once at startup.

**Additionally:** `GCPSecretsManager` was using `System.out.printf` for logging — replaced with `@Slf4j` / `LOG.info`.

---

### 2. NPE + AIOOBE in MongoDB migration utility (v1.3.0)

**File:** `MigrationUtil.java`

**Bugs:**
- `uri.getUserInfo()` returns `null` for URIs without credentials (`mongodb://host/db`). Calling `.split(":", 2)` on null throws NPE.
- Query parameters without a value (e.g. `?replicaSet`) cause `split("=")[1]` to throw `ArrayIndexOutOfBoundsException`.
- `System.out.println` and `e.printStackTrace()` used instead of the logger.
- Raw `Map` casts without generics.

**Fix:** Null-guard `userInfo`, use `split("=", 2)` with length check, use `LOG`, add `@SuppressWarnings("unchecked")` typed generics.

---

### 3. Domain ACL bypass on thread delete

**Files:** `ThreadResourceContext.java`, `FeedResource.java`

**Bug:** `ThreadResourceContext.getDomains()` always returned `null` — a `// TODO: Fix this` comment had been left in production. This means domain-based access control was never applied when evaluating `DELETE /feeds/{threadId}`. Any user who could construct the right request could delete threads belonging to restricted domains.

**Fix:** Add `domainIds: List<UUID>` to the record (populated from `Thread.getDomains()` in `FeedResource`). `getDomains()` now resolves each UUID to an `EntityReference` via `Entity.getEntityReferenceById(Entity.DOMAIN, id, Include.NON_DELETED)`.

---

### 4. Literal `"None"` task names in Databricks pipeline ingestion

**File:** `databrickspipeline/metadata.py`

**Bug:** `DBTasks.name: Optional[str]` can be `None` when `task_key` is absent in the API response. `str(None)` produces the literal string `"None"`, which was being stored as the task name and used for pipeline status lookups, causing incorrect status mapping.

**Fix:** Use `task.name or ""` in both `get_tasks()` and `yield_pipeline_status()`.

---

### 5. Pydantic RootModel serialization in dbt_utils.py

**File:** `dbt_utils.py`

**Bug:** `format_entity_reference()` and `format_domain_reference()` used manual `hasattr(x, "root") / str(x.root)` checks to serialize Pydantic v2 `RootModel[str]` fields (`EntityName`, `FullyQualifiedEntityName`, `UUID`). This pattern was already identified as incorrect for the codebase — the project provides `model_str()` in `metadata.ingestion.ometa.utils` exactly for this purpose.

**Fix:** Import and use `model_str()` throughout both functions.

---

## Testing

- `MigrationUtil`: manually tested against MongoDB URIs with/without credentials and with bare query flags
- `ThreadResourceContext`: the `getDomains()` pattern follows the same `Entity.getEntityReferenceById` call used elsewhere in `ResourceContext.java`
- `databrickspipeline`: `DBTasks.name: Optional[str]` confirmed in `models.py` line 36
- `dbt_utils.py`: `model_str()` is the standard utility per `metadata.ingestion.ometa.utils` — handles both `RootModel` and plain `str` safely
- All modified files pass `get_errors` (IDE diagnostics) with no new errors

## Type of Change

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/dbt/dbt_utils.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt/dbt_utils.py
@@ -23,6 +23,7 @@ from metadata.generated.schema.entity.teams.user import User
 from metadata.generated.schema.tests.testSuite import TestSuite
 from metadata.generated.schema.type.entityReference import EntityReference
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
+from metadata.ingestion.ometa.utils import model_str
 from metadata.ingestion.source.database.dbt.constants import (
     NONE_KEYWORDS_LIST,
     CompiledQueriesEnum,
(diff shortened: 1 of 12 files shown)


IceS2:
Closing for inactivity

PR #27602, opened 2026-04-21
fix: replace paginated role fetch with search API for role selection …

### LINKED ISSUE #27540: Add search support for Roles entity to enable scalable role selection
## Summary
The role selection dropdown across the UI, including user creation, bot configuration, and SSO role mapping, still relies on `GET /v1/roles` with cursor-based pagination. Because of that, roles beyond the first page can be omitted from the dropdown and cannot be reliably selected in the UI.

Backend support for searchable roles is already available, but the frontend role picker flows have not yet been updated to use it. As a result, the UI still depends on paginated list fetching, inflated limits, or page iteration, which does not scale well for larger role sets.

## Current Behavior
- Role selection dropdowns fetch roles using the paginated list API (`GET /v1/roles`).
- Roles beyond the pagination boundary are not shown in the dropdown unless additional pages are fetched.
- Users cannot reliably search or filter roles from the dropdown in larger deployments.
- The frontend works around this by using large limits such as `limit: 1000` or looping through pages, which is inefficient and still not ideal for UX.

## UI Problem
This issue is visible anywhere the UI uses a role picker.

- Dropdowns do not provide a scalable typeahead experience backed by the available role search support.
- Users must scroll through a partial or incomplete list of roles.
- If the required role is outside the fetched page range, it appears unavailable in the UI.
- The experience becomes increasingly unreliable as the number of roles grows.

In practice, the backend can support the use case, but the frontend is not yet wired to use that capability.

## Expected Behavior
Users should be able to find and select any role regardless of how many roles exist.

The UI should:

- use the existing backend role search support
- provide a searchable/typeahead role dropdown
- fetch matching roles dynamically instead of depending on paginated full-list loading
- allow users to reliably select roles even when they are not part of the initially loaded page

## Proposed UI Changes
Update role picker components across the UI to use the existing searchable backend support instead of relying on paginated role listing.

Expected UI behavior:

- the role dropdown should allow users to type and search by role name
- matching roles should be fetched dynamically from the backend search-capable endpoint
- the dropdown should show loading, empty, and no-match states appropriately
- role selection should work consistently across all role assignment surfaces

This applies to flows such as:

- user creation and edit
- bot configuration
- SSO role mapping
- any other UI surface that depends on selecting roles

## Impact
- Users with 100+ roles cannot reliably assign roles during user creation or configuration.
- Role assignment flows are misleading because missing roles appear unavailable in the dropdown.
- Playwright coverage is impacted as well:
  - `UserCreationWithPersona.spec.ts` fails when the target role falls beyond the pagination boundary.

## Notes
This is no longer a backend capability gap. The remaining work is to adopt the existing backend role search support in the frontend role selector components and replace the current paginated list-based behavior.

ShivamChavan01:
### Describe your changes:
Fixes #27540
I worked on adding server-side search support for the Roles entity because:
1. The existing getAllRoles/getRoles APIs use client-side pagination
2. This doesn't scale when there are thousands of roles
3. Users need real-time search/filtering in role dropdowns
The solution implements:
- New searchRoles() API function using Elasticsearch /search/query endpoint
- Replaces pagination calls in 5 components with server-side search
- Adds onSearch handlers for typeahead filtering
### Type of change:
- [x] Improvement
### Checklist:
- [x] My PR title is `Fixes #27540: Add search support for Roles entity to enable scalable role selection`
Additional Notes:
- Depends on backend PR #27540
- Uses elasticsearch role_search_index with deleted:false filter
- Default limit is 25 roles

----
## Summary by Gitar

- **UX Enhancements:**
  - Added logic in `SsoRolesSelectField.tsx` to preserve selected role labels during search operations.
  - Added defensive null-coalescing for role data to prevent runtime errors during mapping.
- **Error Handling:**
  - Integrated `showErrorToast` in `UserProfileRoles.component.tsx` and `SsoRolesSelectField.tsx` to handle `searchRoles` failures gracefully.


harsh-vador:
@ShivamChavan01 can you please attach the screen recording , add unit as well as playwright test for the above?

Rohit0301:
@ShivamChavan01 run the command `make ui-checkstyle-changed` in the root folder of OMD project to fix the UI checkstyle failure

harsh-vador:
Hi @ShivamChavan01 , Thanks for the contribution. This issue has already been addressed through another implementation, so this PR is not needed anymore for now. We’ll close this one to avoid overlapping changes.

PR #27604, opened 2026-04-21
feat(ingestion): add QuestDB database connector

pmbrull:
## Summary

Adds a new database connector for [QuestDB](https://questdb.io/) — a time-series database that speaks the PostgreSQL wire protocol. Exercises the scaffold + schema-first workflow end-to-end; used as the live demo in the ingestion meetup talk.

- SQLAlchemy-based connector backed by `postgresql+psycopg2`, fixed `/qdb` database target (QuestDB is single-database by design)
- Dialect patched per-engine to route introspection through `information_schema` — QuestDB's `pg_catalog` emulation is minimal and the default PG dialect queries fail on a `CHAR -> DOUBLE` cast against `pg_class.relkind`
- Constraint/index/FK lookups short-circuit to empty (QuestDB has no PK, FK, unique constraints, or secondary indexes — the designated timestamp is table partitioning, not an index)
- `get_database_names()` defaults to `qdb` instead of the generic `default`
- Registered in `databaseService.json` enum + `oneOf`, `DatabaseServiceUtils.tsx`, `setup.py` (psycopg2-binary extra), and the test-connection JSON

## Type mappings verified against live QuestDB 9.3.5

| QuestDB | SQLAlchemy |
|---------|------------|
| `TIMESTAMP` | `TIMESTAMP` |
| `SYMBOL` / `STRING` | `VARCHAR` |
| `DOUBLE` | `DOUBLE_PRECISION` |
| `FLOAT` | `REAL` |
| `LONG` | `BIGINT` |
| `INT` | `INTEGER` |
| `BOOLEAN` | `BOOLEAN` |

## Test plan

- [x] 12 unit tests passing (`ingestion/tests/unit/topology/database/test_questdb.py`)
- [x] End-to-end introspection verified against `questdb/questdb:9.3.5` on `localhost:8812` — all 3 demo tables (`sensor_readings`, `trades`, `server_metrics`) discovered with correct column types
- [x] `make generate` succeeds, Pydantic model generated correctly
- [x] `node parseSchemas.js` generates the UI connection schema JSON
- [x] Formatted with `black` / `isort` / `pycln`
- [ ] Integration test against a testcontainers-managed QuestDB (follow-up)
- [ ] UI smoke test of the connection form (follow-up)

----
## Summary by Gitar

- **Sampler configuration:**
  - Implemented `get_tiered_sample` to provide automated profile sampling percentages based on table row counts.
  - Added `resolve_static_sampling_config` to handle dynamic sampling logic, supporting both user-defined thresholds and smart sampling.
- **Dependency updates:**
  - Added `exasol` and `teradata` plugins to the test and dev dependencies in `ingestion/setup.py`.


ingestion/src/metadata/ingestion/source/database/questdb/connection.py
```diff
@@ -0,0 +1,108 @@
...
+            get_connection_url_fn=get_connection_url,
+            get_connection_args_fn=get_connection_args_common,
+        )
+        return patch_questdb_dialect(engine)
+
+    def get_connection_dict(self) -> dict:
+        """
+        Return the connection dictionary for this service.
+        """
+        raise NotImplementedError("get_connection_dict is not implemented for QuestDB")
```
pmbrull, line 89: is this needed?

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Tables with a ``partitionBy`` value other than ``NONE`` are typed
+        ``TableType.Partitioned``; all others are ``TableType.Regular``.
+        """
+        for row in self._tables_cache.values():
+            if row.table_type != "T":
+                continue
+            try:
+                table_type = (
+                    TableType.Partitioned if row.partition_by and row.partition_by != "NONE" else TableType.Regular
```
pmbrull: NONE could be a constant

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def query_view_names_and_types(self, schema_name: str) -> Iterable[TableNameAndType]:
+        """
+        Yield ``TableNameAndType`` entries for QuestDB views and materialized views.
+
+        Rows with ``table_type == "V"`` are typed ``TableType.View``; rows with
+        ``table_type == "M"`` are typed ``TableType.MaterializedView``.
+        """
+        for row in self._tables_cache.values():
+            if row.table_type not in ("V", "M"):
```
pmbrull: lets have constants in V/M too, make it simpler of what magical strings we use

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+            rows = _query_tables(self.connection)
+            self._tables_cache = {row.name: row for row in rows}
+            logger.info("Loaded %d entries from QuestDB tables() catalog", len(self._tables_cache))
+        except Exception as exc:
+            logger.debug(traceback.format_exc())
+            logger.warning("Failed to load QuestDB table catalog: %s — partition details will be unavailable", exc)
+            self._tables_cache = {}
+
+    def get_database_names(self) -> Iterable[str]:
+        yield self.service_connection.databaseName or QUESTDB_DEFAULT_DATABASE
```
pmbrull: is it safe to let people pass databaseName? does questdb allow other names?

ingestion/src/metadata/ingestion/source/database/questdb/connection.py
akashverma0786: Not needed, I added this after claude review , Will remove this.

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
akashverma0786: Adding this

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
akashverma0786: Adding this

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
akashverma0786: Quest only provides 'qdb' as database name and that is hardcoded in connection.py file. 
It won't harm, but to be careful, I will remove this; that way it will be better.

ingestion/src/metadata/ingestion/source/database/questdb/connection.py
akashverma0786: Mb, it is a abstract method so we do need it

IceS2: This might be a small smell on the BaseConnection class I have added to make it work for the data-diff. I will see how we can remove it!

IceS2 (dismissed):

aniketkatkar97 (dismissed):

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        Yield ``TableNameAndType`` entries for QuestDB tables (``table_type == "T"``).
+
+        Tables with a ``partitionBy`` value other than ``NONE`` are typed
+        ``TableType.Partitioned``; all others are ``TableType.Regular``.
+        """
+        for row in self._tables_cache.values():
+            if row.table_type != QUESTDB_TABLE_TYPE_TABLE:
+                continue
+            try:
```
IceS2, line 105: Instead of doing this, Couldn't we have the cache be a dictionary like:

```
{
  "<table_type>":
        "<name>": row
}
```

By doing this you would not need this if to skip a bunch of entries, but you would directly pick the ones you need.

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
```diff
@@ -0,0 +1,179 @@
...
+    def query_view_names_and_types(self, schema_name: str) -> Iterable[TableNameAndType]:
+        """
+        Yield ``TableNameAndType`` entries for QuestDB views and materialized views.
+
+        Rows with ``table_type == "V"`` are typed ``TableType.View``; rows with
+        ``table_type == "M"`` are typed ``TableType.MaterializedView``.
+        """
+        for row in self._tables_cache.values():
+            if row.table_type not in (QUESTDB_TABLE_TYPE_VIEW, QUESTDB_TABLE_TYPE_MATERIALIZED_VIEW):
+                continue
```
IceS2: This would also be not needed if we have the cache as a dict with table type as key

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
akashverma0786: Okay that completely makes sense and is better approach than current implementation

ingestion/src/metadata/ingestion/source/database/questdb/metadata.py
akashverma0786: Adding the changes

IceS2 (approved):

anuj-kumary (approved):

PR #27610, opened 2026-04-22
Fix col type parsign - Databricks

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

----
## Summary by Gitar

- **Bug fix:**
  - Generalized end-of-columns detection to terminate on any `#`-prefixed marker or empty `col_type` to prevent `TypeError`.
  - Added broad `try-except` to column processing to ensure metadata for surviving columns is still captured.
- **Improvement:**
  - Added robust handling for unparseable `col_type` values by skipping the column and logging a warning.
  - Implemented unit tests for Databricks `get_columns` to verify section boundary handling and failure recovery.

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
@@ -182,55 +182,72 @@ def get_columns(self, connection, table_name, schema=None, **kw):
 
     rows = _get_column_rows(self, connection, table_name, schema, kw.get("db_name"))
     result = []
-    for ordinal_position, (col_name, col_type, _comment) in enumerate(rows):
-        # Handle both oss hive and Databricks' hive partition header, respectively
-        if col_name in (
-            "# Partition Information",
-            "# Partitioning",
-            "# Clustering Information",
-            "# Delta Statistics Columns",
-            "# Detailed Table Information",
-            "# Delta Uniform Iceberg",
-        ):
+    for col_name, col_type, _comment in rows:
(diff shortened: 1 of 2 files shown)


ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
IceS2: With the addition of the `continue` in case of no type match we are skipping ordinal position. Could this be an issue? In that case we need to calculate the position ourselves

ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
```diff
@@ -182,55 +182,74 @@ def get_columns(self, connection, table_name, schema=None, **kw):
...
+            # Take out the more detailed type information
+            # e.g. 'map<int,int>' -> 'map', 'decimal(10,1)' -> 'decimal'
+            raw_col_type = col_type
+            type_match = re.search(r"^\w+", col_type)
+            if type_match is None:
+                logger.warning(
+                    f"Skipping column '{col_name}' in {schema}.{table_name}: "
+                    f"unparseable col_type '{col_type}'"
+                )
+                continue
```
IceS2, line 207: I am not sold about this. From research it seems that this should never happen. Added it here as a defensive way to continue the column extraction in any case, but I also do not like the possibility of maybe having incomplete column metadata.

Not sure a proper failure and skipping column extraction would be best here.

pmbrull (dismissed):

pmbrull (approved):

PR #27637, opened 2026-04-22
fix(ssrs): streamline test connection, stream reports, retry transient failures

ulixius9:
## Summary
- **Test connection no longer paginates all reports.** The `GetDashboards` step was calling `client.get_reports()` which walked every page of `/Reports` — on a large SSRS instance that's a huge, unnecessary cost. It now issues a single `$top=1` probe (`client.test_get_reports`), mirroring the existing `CheckAccess` probe against `/Folders`.
- **`SsrsClient.get_reports` now streams.** Converted from list-accumulating to an `Iterator[SsrsReport]` that `yield from`s each page as it's fetched, and `SsrsSource.get_dashboards_list` yields non-hidden reports one at a time. Ingestion memory stays flat regardless of report count.
- **Transient failures are now retried.** Session mounts an `HTTPAdapter` with a `urllib3` `Retry` (2 retries, exponential backoff, covers connect errors, read timeouts, 5xx). Motivated by a production trace where a single 30s `ReadTimeout` against `/Reports` aborted the fetch and left the pipeline reporting success with 0 records.

## Test plan
- [x] Unit tests (`tests/unit/topology/dashboard/test_ssrs.py`) — 22 pass, includes new `test_get_reports_streams_lazily` asserting the generator only fetches the first page until the consumer advances.
- [x] Integration tests (`tests/integration/ssrs/`, `tests/integration/connections/test_ssrs_connection.py`) — 10 pass, includes new `test_get_connection_test_get_reports`, `test_connection_bad_host_get_reports`, and `test_get_reports_retries_transient_failures` (flaky mock server returns 503 twice then 200; asserts exactly 3 requests).
- [ ] End-to-end against a real SSRS instance — not feasible without a Windows host; manual verification recommended.

### DIFF
--- a/ingestion/src/metadata/ingestion/source/dashboard/ssrs/client.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/ssrs/client.py
@@ -11,8 +11,7 @@
 """
 SSRS REST client
 """
-import traceback
-from typing import Iterator, List, Optional, Union
+from typing import Iterable, Iterator, Optional, Union
 
 import requests
 from requests.adapters import HTTPAdapter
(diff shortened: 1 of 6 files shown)


harshach (dismissed):

ingestion/src/metadata/ingestion/source/dashboard/ssrs/client.py
```diff
@@ -53,6 +58,19 @@ def __init__(
         self.session.headers.update({"Accept": "application/json"})
         if verify_ssl is not None:
             self.session.verify = verify_ssl
+        retry = Retry(
+            total=MAX_RETRIES,
```
harshach, line 64: Are we sure retries is hte reason we are getting timed out?

PR #27716, opened 2026-04-24
Refactor(ingestion): introduce ClassifiableEntityAdapter to eliminate scattered isinstance checks

edg956:
### Describe your changes:

Refactors the auto-classification pipeline to eliminate 6 scattered `isinstance(entity, Table/Container)` checks introduced in #26495. Replaces them with a `ClassifiableEntityAdapter` strategy pattern backed by a single O(1) registry.

**What changed:**
- New `sampler/entity_adapters.py` — `EntityAdapter` ABC + `TableAdapter`, `ContainerAdapter`, and two dict registries (`_BY_ENTITY`, `_BY_PIPELINE`) with `adapter_for()` / `adapter_for_pipeline()` lookup functions
- New `sampler/config_utils.py` — `build_database_service_conn_config()` free function extracted from `SamplerProcessor._copy_service_config()`
- `sampler/processor.py` — deleted `_run_for_table`, `_run_for_container`, `_copy_service_config`; replaced with single `_run_for_entity()` using adapter dispatch (~120 lines removed, ~30 added)
- `pii/base_processor.py` — `_get_entity_columns` now delegates to `adapter_for(entity).get_columns(entity)`
- `ometa/mixins/patch_mixin.py` — deleted 4 isinstance helpers (`_get_fields_for_entity`, `_prepare_table_destination`, `_prepare_container_destination`, `_prepare_destination_for_column_tags`); replaced with single method using adapter
- `ingestion/sink/metadata_rest.py` — deleted dead `_patch_entity_column_tags` singledispatch (both registered handlers were byte-identical); callsite replaced with direct `metadata.patch_column_tags()`
- `docs/auto-classification/add-support-for-another-entity.md` — updated to reflect the adapter pattern

**Why:**
Adding `DashboardDataModel` (already referenced in `pii/types.py`) would have required touching all 6 sites. With this refactor it requires: one new `EntityAdapter` subclass + two registry entries + extend `ClassifiableEntityType` union + extend `workflow/classification.py` isinstance tuple. All sampling, PII, and patch logic picks it up automatically.

#
### Type of change:
- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:
- [x] My PR title is `Refactor(ingestion): introduce ClassifiableEntityAdapter to eliminate scattered isinstance checks`

----
## Summary by Gitar

- **Profiler improvements:**
  - Changed `chunk_max` and `chunk_min` calculation in `max.py` and `min.py` to `float` to accommodate broader numeric data.
- **Type checking and stubs:**
  - Added `stubPath` in `pyproject.toml` and introduced various type stubs in `ingestion/stubs/` to improve static analysis coverage.
  - Added essential typing dependencies in `setup.py` including `pandas-stubs` and `google-api-python-client-stubs`.
  - Updated `nifi` and `dashboard/mode` clients to include proper type hints for requests-related utilities.
- **Test refactoring:**
  - Fixed broken mock patch path in `test_container_sampler_processor.py` to point to `metadata.utils.profiler_utils`.


docs/auto-classification/add-support-for-another-entity.md
```diff
@@ -441,25 +441,61 @@ from metadata.generated.schema.entity.data.topic import Topic  # Your new entity
...
+        profiler_config,
+        source_config,
+    ) -> dict | None:
+        return {
+            "service_connection_config": deepcopy(config.source.serviceConnection.root.config),
+            "ometa_client": metadata,
+            "entity": entity,
+            "schema_entity": None,
+            "database_entity": None,
+            "table_config": None,
```
IceS2, line 487: Not a great example for Topics, unless it is needed for some reason and then it would be something we should look at

docs/auto-classification/add-support-for-another-entity.md
```diff
@@ -441,25 +441,61 @@ from metadata.generated.schema.entity.data.topic import Topic  # Your new entity
...
+            "default_sample_config": SampleConfig(),
+            "default_sample_data_count": source_config.sampleDataCount,
+        }
+
+
+# Instantiate once, register in both dicts at the bottom of the file:
+_TOPIC_ADAPTER = TopicAdapter()
+
+_BY_ENTITY[Topic] = _TOPIC_ADAPTER
+_BY_PIPELINE[MessagingServiceAutoClassificationPipeline] = _TOPIC_ADAPTER
```
IceS2: This feels a bit fragile, could this be semi-automated? Perhaps a decorator that we could add in the Adapter, and using Generics to get the Entity Type and Pipeline? (Not sure if it makes sense or if it is possible, just an idea)

ingestion/src/metadata/ingestion/ometa/mixins/patch_mixin.py
IceS2: If we are not within the Table Entity domain anymore I would change the variable names to reflect it.

IceS2 (dismissed):

docs/auto-classification/add-support-for-another-entity.md
IceS2: My comment was more on the lines of why does the sampler need those fields to be passed if we are now using it for more than database services? looks like a smell imho

IceS2 (commented):
Instead of creating the stubs, I think there is a way to ignore missing types from third party libraries that do not provide them. Might be cleaner to do that instead, although I am not 100% sure of the best approach

IceS2 (approved):

edg956:
> Instead of creating the stubs, I think there is a way to ignore missing types from third party libraries that do not provide them. Might be cleaner to do that instead, although I am not 100% sure of the best approach

`mypy` provides such mechanism through `ignore_missing_imports = True`

e.g:

```mypy.ini
[mypy-foobar.*]
ignore_missing_imports = True
```

basedpyright does not offer that, but instead globally disabling or per-file through comments.

docs/auto-classification/add-support-for-another-entity.md
edg956: You're right. It seems like these parameters could be disregarded for other samplers as they're only used to retrieve sample data and profiling configuration for databases and children.

Will have to review and refactor

PR #27723, opened 2026-04-24
Make Selective Field Query during Reindexing

mohityadav766:
Fixes https://github.com/open-metadata/openmetadata-collate/issues/3848

----
## Summary by Gitar

- **Performance Optimization:**
  - Implemented selective field fetching during reindexing to prevent loading heavy, unneeded entity relationships.
  - Added `getRequiredReindexFields()` to the `SearchIndex` interface to define per-entity field requirements.
  - Updated `EntityReader` to prune database queries based on the minimal fields required for specific indexes.
- **Testing:**
  - Added `SearchIndexReindexFieldsParityTest` to enforce contracts for field exclusions and prevent OOM risks.
  - Expanded `SearchIndexFactoryTest` to validate parity between index requirements and factory probes.


openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/ContainerIndex.java
```diff
@@ -35,6 +35,13 @@ public Set<String> getExcludedFields() {
     return Set.of("children");
   }
 
+  @Override
+  public Set<String> getRequiredReindexFields() {
+    Set<String> fields = new HashSet<>(DataAssetIndex.super.getRequiredReindexFields());
+    fields.add("dataModel");
+    return java.util.Collections.unmodifiableSet(fields);
```
pmbrull: nit - do we need the full import here? vs. `Collections.unmodifiableSet(fields);`

mohityadav766: addressed

pmbrull (approved):

PR #27724, opened 2026-04-24
fix: Generic DQ test definitions missing for unspecified data types #27718

### LINKED ISSUE #27718: Generic DQ test definitions missing for unspecified data types in Collate
In v1.12, when a user creates a generic data quality test description without mentioning supported data types and then tries to connect a column to a test case, the generic test definition does not appear in the dropdown menu. This blocks mapping of tests to columns. The user expectation is that, if no data type is specified, the test should apply to all data types by default. A fix is needed to ensure the generic test definition is visible (and correctly mapped) when data types are omitted or to revise the default behavior accordingly.

aniruddhaadak80:
### Description
This PR fixes an issue where generic Data Quality (DQ) test definitions (those without explicitly specified supportedDataTypes) were not appearing in the UI when adding test cases to columns. 

The backend filtering logic in CollectionDAO.java was previously using a strict LIKE match on the supportedDataTypes JSON field. If this field was NULL or an empty array [], the match would fail even if the test was intended to be generic.

### Changes
- Modified TestDefinitionDAO.listBefore, listAfter, and listCount in CollectionDAO.java.
- Updated the filtering logic to include tests where supportedDataTypes is NULL or an empty array.
- Ensured compatibility for both MySQL and PostgreSQL.

Fixes #27718

----
## Summary by Gitar

- **Refactoring:**
  - Introduced `SqlConditionPair` record to encapsulate `StringBuilder` conditions and `ListFilter` state.
  - Extracted JSON array filtering logic into cleaner helper methods `applyJsonArrayFilter` and `applyJsonArrayFilterWithEmpty` to reduce code duplication in `CollectionDAO`.

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
@@ -8578,7 +8578,7 @@ public interface CollectionDAO {
         psqlCondition.append("AND entityType=:entityType ");
       }
 
-            if (supportedDataType != null) {
+      if (supportedDataType != null) {
         filter.queryParams.put("supportedDataTypeLike", String.format("%%%s%%", supportedDataType));
         mysqlCondition.append(
             "AND (json_extract(json, '$.supportedDataTypes') = JSON_ARRAY() "
@@ -8652,7 +8652,7 @@ public interface CollectionDAO {
         psqlCondition.append("AND entityType = :entityType ");
       }
 
-            if (supportedDataType != null) {
+      if (supportedDataType != null) {
         filter.queryParams.put("supportedDataTypeLike", String.format("%%%s%%", supportedDataType));
         mysqlCondition.append(
             "AND (json_extract(json, '$.supportedDataTypes') = JSON_ARRAY() "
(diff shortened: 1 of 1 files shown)


sujalsharma02 (approved):

aniruddhaadak80:
@sujalsharma02 Thank you for the approval! It looks like all the CI/CD checks (like java-checkstyle, integration-tests, etc.) are failing immediately during the Verify PR labels step because the safe to test label is missing. Could you please add the safe to test label so the tests can execute?

sujalsharma02:
> @sujalsharma02 Thank you for the approval! It looks like all the CI/CD checks (like java-checkstyle, integration-tests, etc.) are failing immediately during the Verify PR labels step because the safe to test label is missing. Could you please add the safe to test label so the tests can execute?

The handling for null and empty supportedDataTypes looks correct 
One concern is that both NULL and empty array are being treated as "no restriction", which might introduce inconsistency. 
Would it make sense to standardize this representation
Also, using LIKE for matching JSON values feels a bit fragile (similar to string includes), 
and could lead to unintended matches I think.
Will look forward for your view

aniruddhaadak80:
@sujalsharma02 I have extracted the 'applyJsonArrayFilter' into two separate methods ('applyJsonArrayFilter' and 'applyJsonArrayFilterWithEmpty') and introduced a 'SqlConditionPair' record to reduce the parameter count and cyclomatic complexity, as suggested by the bot review. This brings the helper inline with the method size and parameter limits. Please let me know if there's anything else!

PR #27730, opened 2026-04-25
fix: block dangerous Elasticsearch query types in MCP searchWithDirectQuery

JasonOA888:
## Summary

The MCP `searchMetadata` tool accepts a user-controlled `queryFilter` that is parsed as raw JSON and passed directly to Elasticsearch via `Query.of(q -> q.withJson(...))`. While RBAC conditions are applied as filters afterward, the user query itself is not validated for dangerous query types.

This means an authenticated MCP user could craft queries using:
- **`script`** — execute arbitrary Painless/Groovy scripts server-side
- **`script_score`** — script-based scoring with arbitrary code
- **`wrapper`** — base64-encoded queries that can hide nested dangerous types
- **`percolator`** — reverse query matching
- **`scripted_metric`** — script-based aggregation with init/map/combine/reduce scripts

## Fix

Add `validateQuerySafety()` which recursively walks the parsed JSON query tree and rejects any node containing a blocked query type key:

```java
static final Set<String> BLOCKED_QUERY_TYPES =
    Set.of("script", "script_score", "percolator", "wrapper", "scripted_metric");

static void validateQuerySafety(JsonNode node) throws IOException {
    if (node == null || !node.isObject()) return;
    for (String blockedType : BLOCKED_QUERY_TYPES) {
        if (node.has(blockedType)) {
            throw new IOException("queryFilter contains blocked query type

----
## Summary by Gitar

- **Security enhancements:**
  - Added `function_score` and `runtime_mappings` to `BLOCKED_QUERY_TYPES` in `SearchMetadataTool` to prevent potential injection vulnerabilities.
- **Testing:**
  - Added unit tests in `SearchMetadataToolTest` to verify that `function_score` and `runtime_mappings` are correctly rejected.

### DIFF
--- a/openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SearchMetadataTool.java
+++ b/openmetadata-mcp/src/main/java/org/openmetadata/mcp/tools/SearchMetadataTool.java
@@ -14,6 +14,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.Map;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.openmetadata.schema.search.SearchRequest;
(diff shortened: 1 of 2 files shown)


ayush-shah:

PR #27762, opened 2026-04-27
feat(search): scope index resolution with fetchParentsAliases / fetchChildAliases flags

### LINKED ISSUE #27761: Issue with Aliases in Search
### Affected module

Search / Discovery

### Describe the bug

Currently the UI sends entity Type as index Name. Because of which childAliases are attached. Someone looking for only entity DatabaseSchema should be shown only that. Not tables.

### To Reproduce

- 

### Expected behavior

- 

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

_No response_

_No response_

### Additional context

_No response_

### Pre-submission checklist

mohityadav766:
## Summary

Closes #27761.

The UI now sends the alias `table` (instead of `table_search_index`), and Elasticsearch's alias expansion matches every index whose alias list contains `table` — including `column_search_index`, because `tableColumn.parentAliases` contains `table` in `indexMapping.json`. Result: a UI search for "table" bleeds column hits back into the response. The same shape exists for every alias whose entities also act as a parent for some other entity (`testCase`, `testSuite`, `database`, …).

This change stops relying on ES alias expansion. The server resolves the requested index against the `indexMapping.json` graph into an explicit list of actual `*_search_index` names, gated by two new flags:

| Param | Default | Meaning |
| --- | --- | --- |
| `fetchParentsAliases` | `false` | If `true`, also include the indexes of entities listed in this entity's `parentAliases`. |
| `fetchChildAliases` | `true` | If `true`, also include the indexes of entities that list this alias in *their* `parentAliases`. Default `true` preserves the legacy ES expansion behavior so existing callers don't break; the UI fixes the bug by passing `fetchChildAliases=false`. |

Wired through every search endpoint that takes an index/alias:

- `GET /v1/search/query`
- `GET /v1/search/export`
- `POST /v1/search/preview`
- `GET /v1/search/nlq/query`
- `GET /v1/search/fieldQuery`
- `GET /v1/search/aggregate` and `POST /v1/search/aggregate`
- `GET /v1/search/entityTypeCounts`

The resolver also handles compound aliases (`all`, `dataAsset`, `dataAssetEmbeddings`) by expanding them through the reverse parent/child map (built once at `loadIndexMappings()`). Cluster-alias prefixing is preserved.

## Implementation notes

- `IndexMapping.json` itself is unchanged — the existing `parentAliases` / `childAliases` declarations drive the resolver.
- `SearchRepository.loadIndexMappings()` now also builds an `aliasToChildEntityTypes` reverse map (`alias` → `[entityType, …]` for every entity that lists `alias` as a parent), so child lookup is O(1).
- The actual resolution lives in a static helper `SearchRepository.resolveIndexes(...)` so it's exercised directly in unit tests without bootstrapping a full repository.

## Test plan

- [ ] `mvn test -pl openmetadata-service -Dtest=SearchRepositoryAliasResolutionTest` — new unit tests cover the default flags, child expansion, parent expansion, compound aliases (`dataAsset`), unknown tokens, cluster-alias prefixing, comma-separated tokens, and reverse-map consistency.
- [ ] `mvn test -pl openmetadata-integration-tests -Dtest='SearchResourceIT#testFetchChildAliasesFalseExcludesColumns+testFetchChildAliasesDefaultIncludesColumns+testFetchParentsAliasesTrueIncludesParents+testFetchChildAliasesFalseOnExportEndpoint+testFetchChildAliasesFalseOnAggregate'` — five new integration tests, including the regression for the table-vs-column bleed.
- [ ] Smoke against a running server:
  ```bash
  curl '…/v1/search/query?q=*&index=table&fetchChildAliases=false&size=20' \
    | jq '[.hits.hits[]._source.entityType] | unique'
  # Expect: ["table"] — no "column" entries
  ```

----
## Summary by Gitar

- **Cleanup:**
  - Removed legacy boolean support (`true`/`false`) for `fetchParentsAliases` and `fetchChildAliases` across documentation, API specs, and implementation code.
- **Refactoring:**
  - Updated `SearchRepository.parse` to strictly enforce `none`/`*`/`all` syntax for alias resolution by removing legacy boolean aliases.
  - Simplified `SearchManagementClient` interface documentation to reflect the removal of legacy boolean-token aliasing in `AliasFilter`.


openmetadata-service/src/main/java/org/openmetadata/service/resources/search/SearchResource.java
```diff
@@ -512,7 +561,9 @@ public Response searchWithNLQ(
         new SearchRequest()
             .withQuery(nlqQuery)
```
lautel, line 562: @mohityadav766 not sure about this NLQ endpoint, but search bar + askCollate are currently using the one defined on collate at `HybridSearchResource.java` - Will need to update that one too

mohityadav766:
did a small change via https://github.com/open-metadata/OpenMetadata/pull/27813 , this can be closed

PR #27788, opened 2026-04-28
feat(ui): remove AntD imports from TableV2, fix vertical scroll, add stories

chirag-madlani:
## Summary

- **Remove AntD imports** from `TableV2.tsx` and `TableV2Utils.ts` — these files now depend only on `@openmetadata/ui-core-components` and local types. The public props contract (`TableComponentProps<T>` extending AntD's `TableProps<T>`) is unchanged so no callsites need updating.
- **Fix vertical scroll** — `scroll.y` now works correctly with sticky headers.
- **Add Storybook stories** — six CSF3 stories covering common usage including scrollable behaviour.

### AntD type aliases (`TableV2.interface.ts`)
Added local type aliases that structurally mirror the antd types that were previously imported directly in the V2 implementation files:

| Local type | Replaces |
|---|---|
| `TableColumnType<T>` | `ColumnType<T>` (antd) |
| `TableColumnsType<T>` | `ColumnsType<T>` (antd) |
| `TablePaginationConfig` | `TablePaginationConfig` (antd) |
| `TableSorterResult<T>` | `SorterResult<T>` (antd) |
| `TableCurrentDataSource<T>` | `TableCurrentDataSource<T>` (antd) |
| `TableSortOrder` | `SortOrder` (antd) |
| `TableFilterValue` | `FilterValue` (antd) |

### Vertical scroll fix (`TableV2.tsx`)
**Root cause**: the previous code passed only `overflowY: 'auto'` via `containerStyle`, relying on the Tailwind `tw:overflow-x-auto` CSS class on the inner `UntitledTable` wrapper div to supply `overflow-x`. Setting one overflow axis via a CSS class and the other via inline style creates a CSS specificity ambiguity and can break the scroll container contract.

**Fix**: `tableContainerStyle` now explicitly sets **both** `overflowX: 'auto'` and `overflowY: 'auto'` alongside `maxHeight` when `scroll.y` is provided. This makes the single inner `div` inside `UntitledTable` the sole scroll container for both axes — which is required for `position: sticky` on the `thead` to work correctly while scrolling vertically.

The redundant outer-div `scrollStyle` (which only set `overflowX`) was also removed, and the loading overlay wrapper was simplified.

### Stories (`TableV2.stories.tsx`)
Six CSF3 stories using a minimal `DemoTable` wrapper (avoids the OM app providers that `TableV2` itself requires):

| Story | Covers |
|---|---|
| `Default` | 5-row baseline |
| `ManyRows` | 30 rows, no height cap |
| `VerticalScroll` | 30 rows, `scrollY=300` — sticky header + scroll bar |
| `SmallWithScroll` | Small size variant + scroll |
| `FewRowsScrollCapped` | Few rows, scroll cap set but no scrollbar rendered |
| `Empty` | Empty-state rendering |

## Test plan
- [ ] Render a table with `scroll={{ y: 400 }}` and enough rows to exceed 400 px — confirm vertical scrollbar appears and header stays sticky
- [ ] Render with `resizableColumns` and `scroll={{ y: 400 }}` — confirm same behaviour inside `ResizableTableContainer`
- [ ] Render without `scroll` prop — confirm no regressions in default layout
- [ ] TypeScript: `npx tsc --noEmit` passes with zero errors in the changed files
- [ ] Existing callers of `TableV2` compile without changes

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/common/Table/TableV2.interface.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/components/common/Table/TableV2.interface.ts
@@ -10,6 +10,8 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
+import React from 'react';
+
 export interface FlatRow<T> {
   record: T;
   depth: number;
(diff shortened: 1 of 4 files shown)


chirag-madlani:
**CI update** after commit `5864b0564f`:

- **lint-src** ✅ Fixed — ran `organize-imports-cli` + `eslint --fix` + `prettier --write` locally in the same order as CI; the tree is now clean after one pass.
- **ui-coverage-tests** ✅ Pre-existing — `TestDefinitionForm.test.tsx` (3 tests) fails identically on `main` before our changes. Unrelated to TableV2.
- **playwright-ci-postgresql (6, 6)** — Likely a flaky shard; no TableV2-related tests in that suite.

PR #27800, opened 2026-04-28
fix(search): treat stale parentOf as warning during time-series reindex (#27417)

### LINKED ISSUE #27417: Search Indexing Exception: testCaseResolutionStatus missing parentOf relationship to testCase
We are observing multiple Search indexing errors on a production instance, as well as in several other clusters. The indexer is throwing exceptions when processing testCaseResolutionStatus entities because it cannot find an expected parentOf relationship to/from the testCase entity.

Context & Root Cause
Based on team discussions:

Entity Purpose: testCaseResolutionStatus is used to track incident events (New, Acknowledged, Assigned, Resolved) historically via test_case_resolution_status_time_series for efficient aggregations. This is distinct from the standard test case status (Pass/Fail/Aborted), which is stored alongside the test case itself.

Relationship Mapping: The mapping between testCase -> incidents is stored in the entity_relationship table.

The Bug: Currently, the system lacks a check for testCaseResolutionStatus during the reader/indexing process. Similar checks exist for other entities to ensure they aren't reported as failures, but this specific entity was missed.

Proposed Solution / Next Steps
Add Bypass/Check: Update the indexing reader logic to include a check for testCaseResolutionStatus so that it is not reported as a failure when resolving entity relationships, matching the behavior of other entities.

mohityadav766:
## Summary
- Fixes #27417 — search-indexing failures when processing `testCaseResolutionStatus` entities whose `parentOf` `entity_relationship` row to a `testCase` is missing.
- During time-series reindex, `EntityRepository.ensureSingleRelationship` raises `"... does not have expected relationship parentOf to/from entity type testCase"` (via `CatalogExceptionMessage.entityRelationshipNotFound`). `PaginatedEntityTimeSeriesSource` was returning these as failures, killing the batch. The orphans themselves originate from the 1.4.0 `migrateTestCaseResolution` migration (`MigrationUtil.java:80–124`), which silently skips rows with a null `testCaseReference` or that hit any per-row exception — leaving the time-series row in place with no `entity_relationship` PARENT_OF row.
- Mirrors the warning-vs-failure split already used by `PaginatedEntitiesSource` and broadens it to also catch the relationship-not-found message.

### Changes
- `ReindexingUtil`: extract shared `isEntityNotFoundError` matcher (now matches `"expected relationship"` in addition to `"not found"`, `"instance for"`, `"does not exist"`, `"entitynotfoundexception"`); add `partitionErrors` helper.
- `PaginatedEntitiesSource`: replace local matcher copy with the shared helper. Same behavior for entities, plus the broader matcher.
- `PaginatedEntityTimeSeriesSource`: apply the warning-vs-failure split to `read`, `readWithCursor`, and `readNextKeyset`. Adds `filterStaleRelationshipErrors`, sets `result.setWarningsCount(...)` (already consumed by `SearchIndexExecutor` and `PartitionWorker`), and adds a 3-arg `updateStats(success, failed, warnings)` overload to record warnings on the source's `StepStats`.

### Why this is safe
- Stale parentOf rows are classified as warnings, never silently dropped — they show up in `warningRecords` and in DEBUG logs, so they remain observable.
- Real reader failures (JSON parse errors, DB issues, anything not matching the matcher) still increment `failedRecords` and still set `lastFailedCursor` so the failure cursor is preserved.
- No public API change. No schema change. `warningRecords` already exists in `eventPublisherJob.json` and is already populated by `PaginatedEntitiesSource`.

### Follow-up not in this PR
- Worth a separate one-shot cleanup migration that deletes orphaned `testCaseResolutionStatus` rows (those whose `entity_relationship` parentOf row is missing AND whose JSON `testCaseReference` is null) to actually shrink the orphan set in production. This PR just stops them from failing reindex.

## Test plan
- [x] `ReindexingUtilStaleRelationshipTest` — 5 tests covering matcher + partitioning, including the exact `ensureSingleRelationship` message format
- [x] `PaginatedEntityTimeSeriesSourceStaleRelationshipTest` — 4 tests covering `readNext` and `readWithCursor` with stale-only, mixed (stale + real), and exception cases; asserts `result.getWarningsCount()`, `result.getErrors()`, and source `StepStats`
- [x] Existing `EntityReaderLifecycleTest` (8), `SearchIndexExecutorControlFlowTest` (60), `PartitionWorkerTest` (30) all still pass
- [x] `mvn spotless:apply` clean
- [ ] Verify on a cluster with known orphans: trigger SearchIndexingApplication for `testCaseResolutionStatus`, confirm `warningRecords > 0`, `failedRecords == 0`, and run completes

----
## Summary by Gitar

- **Refinement of field retrieval:**
  - Improved `getSearchIndexFields` error handling to prevent defaulting to `"*"` when filtering fails, opting to return the required fields instead.
- **Resilience:**
  - Updated `ReindexingUtil` to log specific failure details during index field lookup or filtering, ensuring better visibility into reindexing issues.


mohityadav766:
Thanks for the review — addressed all three findings in 91bbdc1:

**1. `readWithCursor` / `readNextKeyset` not updating warning stats** (gitar) — fixed. Both now call `updateStats(success, failed, warnings)` after `filterStaleRelationshipErrors`, so source-level `StepStats.warningRecords` reflects warnings for offset-cursor and keyset paths (was previously only correct in `read()`).

**2. Overly broad `"not found"` matcher** (gitar) — fixed. Dropped bare `"not found"` and added `"entity not found"`, so we still catch `EntityNotFoundException`'s `"Entity not found:"` form (e.g. `EntityRepository.java:9102`) and the canonical `"%s instance for %s not found"` (via the existing `"instance for"` marker), but `"Column 'status' not found in result set"` and `"SSL certificate not found"` are now correctly classified as failures. Audited every `not found` message in `CatalogExceptionMessage` to confirm we don't lose legitimate coverage.

**3. `partitionErrors` null-safety** (Copilot) — fixed. `Objects.requireNonNull(warningsOut)` plus a Javadoc note that both lists must be mutable.

Tests updated to cover all three: the new bare-`"not found"` exclusion, the `"Entity not found:"` inclusion, `readWithCursor` stats propagation, and the null-`warningsOut` guard. Existing tests (`EntityReaderLifecycleTest`, `SearchIndexExecutorControlFlowTest`, `PartitionWorkerTest` — 98 total) still pass.

pmbrull (dismissed):

mohityadav766:
Thanks @Copilot — both findings addressed in 856fd2d:

**1. `PartitionWorker.processBatch` losing failure diagnostics on all-error batches** — fixed. Extracted the failure-recording into a `recordReaderFailures(...)` helper and call it *before* the `readSuccessCount == 0` early-return, so per-entity reader failures are persisted via `failureRecorder.recordReaderEntityFailure(...)` whether or not the batch had any indexable rows. Reorganized the read flow so it now reads: compute counts -> record reader stats -> record per-entity failures -> short-circuit if no successful rows -> otherwise write to sink.

**2. `isEntityNotFoundError` name no longer fits** — renamed to `isStaleReferenceError` (with updated Javadoc making the broader scope explicit: covers both missing entities *and* missing entity_relationship rows from `ensureSingleRelationship`). Updated the 3 call sites in `ReindexingUtil`, `PaginatedEntitiesSource`, and the test. `EntityRepository` has its own unrelated private `isEntityNotFoundError` helper that I left alone.

All 108 tests in the area still pass (`PartitionWorkerTest`, `SearchIndexExecutorControlFlowTest`, `EntityReaderLifecycleTest`, the two new ones).

mohityadav766:
Both concerns addressed in 0601641.

**1. Missed `EntityNotFoundException` formats** — confirmed by reading [`EntityNotFoundException.java`](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/exception/EntityNotFoundException.java):

| Factory | Message | Caught before? | Caught now? |
|---|---|---|---|
| `byId` | `Entity with id [%s] not found.` | ✗ | ✓ via `"entity with id"` |
| `byName` | `Entity with name [%s] not found.` | ✗ | ✓ via `"entity with name"` |
| `byVersion` | `Entity with id [%s] and version [%s] not found.` | ✗ | ✓ via `"entity with id"` |
| `byParserSchema` | `Parser schema not found for entity with id [%s].` | ✗ | ✓ via `"parser schema not found"` |
| `byFilter` | `Entity not found for query params [%s].` | ✓ via `"entity not found"` | ✓ |
| `CatalogExceptionMessage.entityNotFound` | `%s instance for %s not found` | ✓ via `"instance for"` | ✓ |
| `ensureSingleRelationship` | `... does not have expected relationship ...` | ✓ via `"expected relationship"` | ✓ |

Specific patterns instead of bare `"not found"` so we still don't false-positive on `"Column 'x' not found"` / `"SSL certificate not found"`. New unit test `isStaleReferenceError_recognisesEveryEntityNotFoundExceptionFactory` exercises every factory.

**2. WARN log spam in `PartitionWorker.recordReaderFailures`** — `EntityTimeSeriesRepository.getEntityList` builds `EntityError` with only `withMessage(...)`, no entity ref. Every time-series error hit the null-id branch and emitted WARN. Downgraded to DEBUG with a comment explaining why null is expected for time-series readers (per-entity recording isn't possible without an id, but it's also not a real anomaly worth a WARN).

mohityadav766:
Addressed 4 more Copilot comments in eac76b6:

1. **NPE on `result.getErrors()`** — `EntityTimeSeriesRepository.getResultList(...)` leaves `errors=null` on the success path. `filterStaleRelationshipErrors` now normalizes to an empty list before any of the callers (logging, `updateStats`, the `if (!result.getErrors().isEmpty())` branch) read it.
2. **`Locale.ROOT` for `toLowerCase`** — switched in `isStaleReferenceError` so substring matching is locale-stable (Turkish-locale 'I' edge case etc.).
3. **WARN spam in `read()`** — was logging every real error message at WARN. Now logs a single WARN with the error count and dumps the first 5 message details at DEBUG (gated by `isDebugEnabled`).
4. **`fillTerminalTimings` Javadoc** — rewrote to match the implementation: `endTime` and `executionTime` are filled independently and only when null, so callers that pre-populated `endTime` (e.g. `DistributedJobStatsAggregator` from `job.getCompletedAt()`) still get an accurate `executionTime`.

Skipping the `tags` in `COMMON_REINDEX_FIELDS` / cursor-reader-fields concern (#5 in my list above) per @mohitdeuex — that touches commits outside the issue #27417 scope and will be handled separately.

109/109 tests still pass.

pmbrull (approved):

PR #27814, opened 2026-04-29
(playwright)ui: Fix session flakiness in Users Playwright spec

harsh-vador:
### Description
This PR fixes an AUT flake in Users.spec.ts where the Data Steward permission test could observe a different logged-in user during execution.

### Changes:

- Removed the unnecessary adminPage fixture usage from the Data Steward permission test.
- Replaced getApiContext(adminPage) with createAdminApiContext() for admin-only API setup.
- Keeps the UI session scoped to the Data Steward user while table ownership setup happens through an isolated admin API context.

https://github.com/user-attachments/assets/199e39c9-46bf-4f4b-982b-aa51331ac9f5

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### Checklist:

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Users.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Users.spec.ts
@@ -33,7 +33,6 @@ import { TeamClass } from '../../support/team/TeamClass';
 import { UserClass } from '../../support/user/UserClass';
 import { createAdminApiContext, performAdminLogin } from '../../utils/admin';
 import {
-  getApiContext,
   redirectToHomePage,
   toastNotification,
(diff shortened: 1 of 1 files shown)


ShaileshParmar11 (dismissed):

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Users.spec.ts
aniketkatkar97: Do we even need to do the context creation here? This logic can be moved to the `beforeAll` block, right? Why have the try catch logic in the tests in the first place! if something fails in the tests, we won't know at all due this.

aniketkatkar97 (approved):

PR #27816, opened 2026-04-29
fix: Move version metadata migration to 2.0.0 and remove crashing MySQL multi-valued index

yan-3005:
## Problem

The 1.12.7 migration (`e4d3e423e1`) introduced two new columns on `entity_extension` — `versionNum` and `changedFieldKeys` — to support version timeline filtering. It populated those columns via a `postDataMigrationSQLScript.sql` and backed `changedFieldKeys` with a MySQL **multi-valued index**.

This caused an immediate crash in AUT:

```
Exceeded max total length of values per record
for multi-valued index 'idx_entity_extension_changed_field_keys' by 6611 bytes
```

### Root cause — why the MySQL multi-valued index blows up

MySQL's multi-valued index on a JSON array reserves the full declared `CHAR(N)` storage **per element, per row** in the index. The index was declared as `CHAR(512)` with `utf8mb4`, which reserves **2048 bytes per element**. MySQL imposes a hard per-row limit of ~8 KB for multi-valued indexes. That means any entity version with more than ~4 changed field names in a single version would exceed the limit and crash the `ALTER TABLE` or `UPDATE`.

`entity_extension` in production routinely has versions with dozens of changed fields (e.g. a bulk tag update can produce 50–300 `changedFieldKeys` entries in one version row). The index design made the migration structurally impossible to run on real data.

The Postgres GIN index on the same column does **not** have this problem — GIN stores posting lists per element without a per-row byte cap — so it is kept unchanged.

### Why not partial indexing (index only the first N elements)?

MySQL's multi-valued index indexes **all** elements in the array or none. There is no way to index a subset. The only options were:
1. Sanitize (truncate) the data — rejected, we must not drop user data
2. Use a smaller `CHAR(N)` — still fails for wide entities; just moves the threshold
3. Remove the index entirely — chosen

### Why is the index safe to remove?

Every query that uses `changedFieldKeys` always starts with `WHERE e.id = :id`, scoping results to a single entity (typically 100–500 version rows). `JSON_CONTAINS` on 100–500 rows is negligible — the index provided no meaningful speedup. A legacy code path (`getExtensionsWithFieldChangedLegacy`) already works without the index and is already live in production.

---

### 1. Removed the 1.12.7 migration entirely

All four SQL files created by the original PR are deleted. They did not exist before that commit, so reverting them leaves 1.12.7 clean.

```
bootstrap/sql/migrations/native/1.12.7/mysql/schemaChanges.sql          ← deleted
bootstrap/sql/migrations/native/1.12.7/mysql/postDataMigrationSQLScript.sql ← deleted
bootstrap/sql/migrations/native/1.12.7/postgres/schemaChanges.sql       ← deleted
bootstrap/sql/migrations/native/1.12.7/postgres/postDataMigrationSQLScript.sql ← deleted
```

### 2. Moved the DDL to 2.0.0 schemaChanges.sql

**MySQL** — appended to `2.0.0/mysql/schemaChanges.sql`:
```sql
ALTER TABLE entity_extension
  ADD COLUMN versionNum DOUBLE NULL,
  ADD COLUMN changedFieldKeys JSON NULL;

CREATE INDEX idx_entity_extension_version_order
  ON entity_extension (id, versionNum);
```
No multi-valued index. The BTree index on `(id, versionNum)` is kept for efficient DESC ordering of versions per entity.

**Postgres** — appended to `2.0.0/postgres/schemaChanges.sql`:
```sql
ALTER TABLE entity_extension
  ADD COLUMN IF NOT EXISTS versionNum DOUBLE PRECISION,
  ADD COLUMN IF NOT EXISTS changedFieldKeys JSONB;

CREATE INDEX IF NOT EXISTS idx_entity_extension_version_order
  ON entity_extension (id, versionNum DESC) WHERE versionNum IS NOT NULL;

CREATE INDEX IF NOT EXISTS idx_entity_extension_changed_field_keys
  ON entity_extension USING GIN (changedFieldKeys) WHERE changedFieldKeys IS NOT NULL;
```
GIN index retained — it works correctly on Postgres with no per-row size limit.

### 3. Replaced the SQL data migration with a batched Java migration

The original `postDataMigrationSQLScript.sql` was a single-shot CTE that ran as one SQL statement over the entire `entity_extension` table. On large instances this would hold a lock for minutes and risk a timeout.

The data migration is now a batched Java method in `v200/MigrationUtil.backfillVersionMetadata()`.

#### Cursor-based pagination

The SELECT uses cursor-based pagination on the PRIMARY KEY instead of `WHERE versionNum IS NULL`:

```sql
SELECT id, extension, json FROM entity_extension
WHERE (id, extension) > (:lastId, :lastExt)
AND extension LIKE '%.version.%'
AND versionNum IS NULL
ORDER BY id, extension
LIMIT 5000
```

**Why this matters on large tables:**

The naive `WHERE versionNum IS NULL LIMIT N` approach forces a full table scan on every batch. Verified with `EXPLAIN` on a 2.1M-row table:

| Approach | EXPLAIN type | key | rows per batch |
|---|---|---|---|
| `WHERE versionNum IS NULL LIMIT 1000` | `ALL` | none | 2,187,980 |
| Cursor `(id, extension) > (x, y)` | `index` | `PRIMARY` | 5,000 |

With the cursor, the PRIMARY KEY index drives the scan forward from the last position — each batch reads exactly `LIMIT` rows. The `LIKE` and `IS NULL` filters are applied as cheap post-filters within the cursor's range. This also makes re-runs safe: `versionNum IS NULL` skips already-processed rows so the migration is idempotent.

#### PreparedBatch for updates

Instead of issuing one `execute()` per row (N round trips per batch), all updates in a batch are sent as a single `PreparedBatch.execute()` call — one round trip regardless of batch size.

#### Fallback on batch failure

If `PreparedBatch.execute()` fails (e.g. one row has a malformed JSON column), the code falls back to processing each row individually:
- First attempt: full update (`versionNum` + `changedFieldKeys`)
- If that fails: `versionNum`-only update (guarantees the row won't be retried)
- If even that fails: log a warning and skip the row — one bad row cannot abort the entire backfill

This preserves `changedFieldKeys` for all healthy rows in the batch even when one row is problematic. The old per-row approach had the same two-tier fallback; the new approach restores that per-row safety net while still getting the throughput benefit of `PreparedBatch` on the happy path.

#### Performance summary (2.1M-row table)

| | Before (naive IS NULL) | After (cursor + PreparedBatch) |
|---|---|---|
| Batch size | 1,000 | 5,000 |
| Batches total | 2,147 | ~430 |
| Rows scanned per SELECT | 2,187,980 | 5,000 |
| UPDATE round trips per batch | 1,000 | 1 |
| Observed time (2.1M rows) | ~2 hours | ~5–15 min |

### 4. Java vs SQL — behavioral equivalence

| Aspect | SQL (original) | Java (this PR) |
|---|---|---|
| `versionNum` source | `CAST(SUBSTRING_INDEX(ext, '.version.', -1) AS DOUBLE)` | `lastIndexOf(".version.")` + `Double.parseDouble` — identical semantics |
| Field name dedup | `SELECT DISTINCT` | `LinkedHashSet` — same |
| NULL / empty filter | `WHERE field_name IS NOT NULL AND field_name <> ''` | `nameNode.isNull()` + `isEmpty()` — same |
| Array order | `JSON_ARRAYAGG` (arbitrary) | alphabetically sorted — stricter, functionally equivalent for `JSON_CONTAINS` |
| Batching | single-shot | 5,000 rows/batch — safer for large tables |

### 5. Tests

- All backfill unit tests moved into `v200/MigrationUtilTest` (41 tests total)
- Tests cover: MySQL/Postgres SQL selection, null JSON handling, multi-row batch (verifies `.add()` called per row), batch-level fallback triggering per-row retry, per-row full-update failure falling back to versionNum-only
- `MigrationSqlStatementHashTest` updated to validate `2.0.0/mysql/schemaChanges.sql` (unique statement hashes, no `INFORMATION_SCHEMA` references)

---

## Pros / Cons

**Pros**
- Migration no longer crashes on MySQL — the multi-valued index was the sole cause of the AUT failure
- Cursor pagination eliminates O(N²) full table scans; verified via EXPLAIN
- PreparedBatch reduces UPDATE round trips from N to 1 per batch
- Per-row fallback in the batch failure path ensures `changedFieldKeys` is preserved for all healthy rows even if one row is bad
- Postgres behaviour is unchanged (GIN index kept)
- Fully covered by unit tests

**Cons**
- Any environment that already ran the 1.12.7 migration (and survived — unlikely given the crash) would skip the 2.0.0 backfill. In practice the 1.12.7 migration crashed before completing on MySQL, so no MySQL environment has these columns populated. Postgres environments that ran 1.12.7 successfully already have the columns; the 2.0.0 Java migration is idempotent (`AND versionNum IS NULL`) so it is safe to re-run
- The `changedFieldKeys` array order differs from the original SQL (`JSON_ARRAYAGG` is unordered; Java sorts alphabetically). This has no functional impact since all consumers use `JSON_CONTAINS` which is order-agnostic
- MySQL `ADD COLUMN` has no `IF NOT EXISTS` guard (MySQL 8.0 limitation). If a MySQL environment somehow survived 1.12.7 with the columns already present, the 2.0.0 DDL would fail with "Duplicate column name". In practice this cannot happen since 1.12.7 crashed before the columns were created on MySQL

## Test plan
- [ ] Run full migration against MySQL with a pre-2.0.0 dataset — confirm no crash, all `entity_extension` rows get `versionNum` and `changedFieldKeys` populated
- [ ] Run full migration against Postgres — confirm same
- [ ] Verify version timeline UI works correctly after migration (version ordering, field-change filtering)
- [ ] Confirm `MigrationSqlStatementHashTest` passes in CI
- [ ] Confirm `v200/MigrationUtilTest` (41 tests) passes in CI


yan-3005:
https://github.com/open-metadata/OpenMetadata/pull/27837

Issue #26265, opened 2026-03-05
[Clickhouse Linage] Missing downsteam for MATERIALIZED VIEW


Kreidler-UKFr:
**Affected module**
Linage

**Describe the bug**
We're having a lot of `MATERIALIZED VIEW`s in our Clickhouse-Instance, where we transpose it to a new table with the `TO <schema>.<table>`-Argument[0]. While the Linage Agent creates the correct **upstream** links I would also expect the table from `TO <schema>.<table>` as **downstream** link, but it's not created.

[0] https://clickhouse.com/docs/sql-reference/statements/create/view#materialized-view

**To Reproduce**
Query:
```SQL
CREATE MATERIALIZED VIEW <schema-01>.samples_mv TO <schema-02>.samples_e (
    column_01,
    column_02
) FROM <schema-01>.samples
```
With this Query I get the following Linage:
<img width="567" height="281" alt="Image" src="https://github.com/user-attachments/assets/179131e8-0098-4c99-9a6e-243b5d8dbd81" />

**Expected behavior**
Excpected Linage (added Linage by hand):
<img width="869" height="254" alt="Image" src="https://github.com/user-attachments/assets/8ff12f94-f349-4425-8951-602140c48307" />

**Version:**
 - OS: docker
 - OpenMetadata version: `docker.getcollate.io/openmetadata/server:1.12.1`
 - OpenMetadata Ingestion package version:  `docker.getcollate.io/openmetadata/ingestion:1.12.1`
 
**Additional context**
While writing this up I found a `MATERIALIZED VIEW` where even the FROM `<table>` expression did not create a **upstream** Linage. Generalized Query:
```SQL
CREATE MATERIALIZED VIEW <schema>.<materialzed-view-name> REFRESH EVERY 3 HOUR TO <schema>.<to-table> (column_01, column_02) DEFINER = <user> SQL SECURITY DEFINER AS SELECT * FROM <schema>.<from-table>
```

Sushantd07:
Hi, I’d like to work on this issue. I’ll start working on it.

piyushdotcomm:
Hey @harshach, I've looked into this — the issue is that the ClickHouse `TO <schema>.<table>` syntax in `CREATE MATERIALIZED VIEW` isn't recognized as a target table by the lineage parser, so the downstream link never gets created. I know where the fix needs to go. Can you assign this to me? cc @nikhilchennam @ulixius9 @PubChimps @Kreidler-UKFr

[...]

PR #27827, opened 2026-04-29
fix(clickhouse): extend MV TO-clause regex to cover POPULATE, UUID, and three-part names (#26265)

Vaishnavi-Cailum:
## Summary

Fixes gaps in the `MATERIALIZED_VIEW_TO_PATTERN` regex introduced by #27553 for issue #26265.

**Four additional ClickHouse DDL forms were not matched:**

- `POPULATE` keyword before `TO` — valid ClickHouse DDL supported since v22.x
- `UUID` clause — emitted by `system.tables` DDL output (e.g. `CREATE MATERIALIZED VIEW db.mv UUID '...' TO db.target`)
- Three-part qualified names — `catalog.schema.mv TO catalog2.schema2.target`
- Bare `FROM` without `AS` — `CREATE MATERIALIZED VIEW mv TO target (...) FROM source` (no `AS SELECT`)

**Changes:**
- `_QUALIFIED_IDENT` extended to match up to three dot-separated parts
- Added `(?:\s+UUID\s+\S+)?` optional clause after MV name
- Added `(?:\s+POPULATE)?` optional clause before `TO`
- Added `FROM\b` as an accepted token after the target table (alongside `(` and `AS`)
- 4 new regression tests in `TestGetMvToTargetTable`

Builds on top of the approach in #27553 (same `query_lineage_producer` override, same `get_mv_to_target_table` helper) — no changes to that logic, only the regex coverage is extended.

## Test plan

- [ ] All existing `TestGetMvToTargetTable` tests still pass
- [ ] 4 new tests pass: `test_populate_keyword_before_to`, `test_on_cluster_and_populate_before_to`, `test_uuid_clause`, `test_three_part_qualified_names`
- [ ] Run: `cd ingestion && python -m pytest tests/unit/topology/database/test_clickhouse_utils.py -v`

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/clickhouse/lineage.py
+++ b/ingestion/src/metadata/ingestion/source/database/clickhouse/lineage.py
@@ -11,12 +11,16 @@
 """
 Clickhouse lineage module
 """
+from typing import Iterator
+
+from metadata.generated.schema.type.tableQuery import TableQuery
 from metadata.ingestion.source.database.clickhouse.queries import (
     CLICKHOUSE_SQL_STATEMENT,
 )
 from metadata.ingestion.source.database.clickhouse.query_parser import (
     ClickhouseQueryParserSource,
 )
+from metadata.ingestion.source.database.clickhouse.utils import get_mv_to_target_table
 from metadata.ingestion.source.database.lineage_source import LineageSource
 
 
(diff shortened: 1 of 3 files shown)


Vaishnavi-Cailum:
Closing — reopened as #27832 from correct account.

PR #27831, opened 2026-04-29
fix(ingestion): skip mark-deleted for schemas whose table listing failed

### LINKED ISSUE #18187: bug: intermittent issues with connection to database service cause unwanted table deletion
**Affected module**
Ingestion

**Describe the bug**
We have defined Trino connection with ingestion that has **Mark Deleted Tables** to **True**. Our Trino uses Hive Metastore under the hood, which might from time to time experience connectivity issues. If a connectivity issue happens and Trino cannot get a list of tables, it will remove all tables from Open Metadata.

**To Reproduce**

- Register Trino service and corresponding metadata ingestion oricess with **Mark Deleted Tables** set to **True**
- Run ingestion on healthy Trino connector instance (for example HMS)
- Cause  Trino connector system to fail (for example bring HMS down)
- Run ingestion on not healthy Trino connector

because of issues with Trino connector, the table list is not fetched and they are marked for deletion.

**Expected behavior**
If listing tables fails they are not marked for deletion, the schema is filtered out from processing in mark_deleted_tables method.

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8] 1.5.4
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`] 1.5.4.0

**Additional context**

vaishnavidixit12:
## Why

Fixes #18187

When `get_tables_name_and_type()` raises an exception for a schema (e.g. a transient connectivity failure, permission error, or driver bug), the current code silently swallows the error and leaves `database_source_state` empty for that schema. On the next call to `mark_tables_as_deleted()`, OpenMetadata interprets the empty state as "all tables were deleted" and soft-deletes every table it knew about for that schema — silent data loss on every partial connectivity failure.

## What changed

**`common_db_source.py`** — `get_tables_name_and_type()` exception handler  
When an exception is caught, the method now builds the schema FQN and records it in `self.schemas_with_table_listing_errors` (inherited from `DatabaseServiceSource`).

**`database_service.py`** — `DatabaseServiceSource`  
- Added `schemas_with_table_listing_errors: Set = set()` class-level field.  
- `mark_tables_as_deleted()` now checks this set before calling `delete_entity_from_source`; schemas that failed listing are skipped with a `WARNING` log.

The fix is in the shared base classes, so all database connectors (BigQuery, Snowflake, Redshift, …) are protected automatically.

## Test plan

- [x] `test_mark_deleted_tables_safety.py` — 8 new unit tests covering:
  - Failed schema is skipped while healthy schemas are still processed
  - All-failed case: `delete_entity_from_source` is never called
  - `markDeletedTables=False` still short-circuits everything
  - Listing error adds the schema FQN to the error set
  - Successful listing leaves the error set empty
  - `None` FQN (schema not found in OM) is not added to the error set

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/common_db_source.py
+++ b/ingestion/src/metadata/ingestion/source/database/common_db_source.py
@@ -134,6 +134,7 @@ class CommonDbSourceService(DatabaseServiceSource, SqlColumnHandlerMixin, SqlAlc
         self._inspector_map = {}
         self.table_constraints = None
         self.database_source_state = set()
+        self.schemas_with_table_listing_errors = set()
         self.context.get_global().table_constrains = []
         self.context.get_global().foreign_tables = []
         self.context.set_threads(self.source_config.threads)
@@ -412,15 +413,7 @@ class CommonDbSourceService(DatabaseServiceSource, SqlColumnHandlerMixin, SqlAlc
         except Exception as err:
             logger.warning(f"Fetching tables names failed for schema {schema_name} due to - {err}")
             logger.debug(traceback.format_exc())
-            schema_fqn = fqn.build(
-                self.metadata,
-                entity_type=DatabaseSchema,
-                service_name=self.context.get().database_service,
-                database_name=self.context.get().database,
-                schema_name=schema_name,
-            )
-            if schema_fqn:
-                self.schemas_with_table_listing_errors.add(schema_fqn)
+            self._record_schema_listing_error(schema_name)
 
     @calculate_execution_time()
     def get_schema_definition(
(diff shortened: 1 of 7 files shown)


Vaishnavi-Cailum:
Finding 1 — shared mutable class state: Added self.schemas_with_table_listing_errors = set() in CommonDbSourceService.__init__, making it instance-scoped like database_source_state.     
                                                                                                                                                                                            
Finding 2 — connectors bypassing the guard: Extracted _record_schema_listing_error(schema_name) helper into DatabaseServiceSource (builds the FQN, adds to the set), then wired it into:  
  - common_db_source.py — simplified the existing except block                                                                                                                              
  - unitycatalog/metadata.py — wraps client.tables.list() call                                                                                                                              
  - deltalake/metadata.py — wraps client.get_table_info() call                                                                                                                            
  - glue/metadata.py — wraps _get_glue_tables() call                                                                                                                                        
  - salesforce/metadata.py — adds the helper call to the existing except block

ayush-shah:

PR #25224, opened 2026-01-12
Fixes #25167: Secrets Manager Empty String Sanitization

ayush-shah:
Fixes #25167

This pull request improves the handling of null and empty secret values across the secrets management code, ensuring that such values are consistently sanitized and stored as a special "null" string to prevent issues with secrets providers. It also introduces comprehensive tests to verify this behavior and updates a Google Cloud dependency.

**Secret Value Sanitization and Handling:**

* Refactored the `upsertSecret` method in `ExternalSecretsManager` to use a new `cleanNullOrEmpty` method, ensuring that null or empty secret values are always converted to the constant `NULL_SECRET_STRING` before storage or update.
* Updated the Kubernetes secrets manager test to assert that empty secret values are stored as the `NULL_SECRET_STRING`, preventing rejection by the secrets manager. [[1]](diffhunk://#diff-70c143c2363f2b5d340eb5fbb45faf3f080709337c4c30a7c6a2b28400442295L263-R263) [[2]](diffhunk://#diff-70c143c2363f2b5d340eb5fbb45faf3f080709337c4c30a7c6a2b28400442295L280-R283)

**Testing Improvements:**

* Added new unit tests in `ExternalSecretsManagerTest` to verify that `cleanNullOrEmpty` correctly converts null and empty strings, and to check encryption behavior for authentication mechanisms and database connections with null or empty secrets. [[1]](diffhunk://#diff-f2f0e27c1867ea1257bbf423f4a5717f1c1f9aa54272c25edb2b6655de729270R4-R14) [[2]](diffhunk://#diff-f2f0e27c1867ea1257bbf423f4a5717f1c1f9aa54272c25edb2b6655de729270R176-R244)
* Added new tests in `GCPSecretsManagerTest` to ensure that encrypting bot authentication mechanisms with null or empty tokens does not result in empty secret payloads. [[1]](diffhunk://#diff-d373392e7021cf240cd46424777e994e6ffd6f14ec2056a7213bfb6fa1b0b4f7R3) [[2]](diffhunk://#diff-d373392e7021cf240cd46424777e994e6ffd6f14ec2056a7213bfb6fa1b0b4f7R19-R27) [[3]](diffhunk://#diff-d373392e7021cf240cd46424777e994e6ffd6f14ec2056a7213bfb6fa1b0b4f7R113-R150)

**Dependency and Build Updates:**

[...]

PR #27838, opened 2026-04-30
Fixes #21259: Don't store null secrets in Secrets Manager

### LINKED ISSUE #21259: Don't store `null` secrets in Secrets Manager
Using the OM server configured with `managed-aws`:

process:
1. create snowflake service with a password & passKey
2. edit snowflake service, make changes in the password field and then remove them
3. save
4. password is stored in SM as `null`
5. snowflake does not rly need password if we have the private key field

We should control in the BE that if a secret value is `null`, then the secret should be removed altogether, not stored as `null`.

As part of this we need to figure out how to clean up existing services that have secrets that are null. Those should be cleaned from the service connection: If the secret is null, just use a null value instead

aji-aju:
## Summary

Closes #21259.

When a user clears a password field on a service backed by an external Secrets Manager (`managed-aws`, GCP, Azure KV, Kubernetes), the previous code stored the literal string `\"null\"` in the secrets backing store instead of removing the secret. Downstream consumers (e.g. the Snowflake driver) then read back the four-character string `\"null\"` as a real password, breaking authentication — most visibly for users who authenticate via private key and never wanted a password configured at all.

This PR fixes the root cause and updates the tests that codified the buggy contract.

## The bug, demonstrated

Before the fix, with `SECRET_MANAGER=managed-aws`:

```bash
# Create a service with a real password, then clear it via the UI (type something, delete it, save).
$ aws secretsmanager get-secret-value \
    --secret-id /<prefix>/<cluster>/database/<service>/authtype/password \
    --query SecretString --output text
null   # ← the literal four ASCII characters, not the JSON null
```

After the fix, the same call returns `ResourceNotFoundException` (the secret has been deleted) and the entity's password field is `null`, not a stale `secret:/` reference.

## Approach

The fix is three small, layered changes — each layer hands a clean null to the next:

1. **`ExternalSecretsManager.upsertSecret()`** — when value is null/empty, the user's intent is *\"remove this credential\"*, so delete the backing secret instead of writing `\"null\"`. This also satisfies the GCP/Kubernetes empty-string rejection constraint from #25224 by simply *not calling* upsert with an empty value.
2. **`ExternalSecretsManager.storeValue()`** — return `null` for null/empty values so the entity does not carry a stale `secret:/` reference to a deleted secret. Without this, the next read would attempt to follow the reference and either get a `ResourceNotFoundException` or (pre-fix) the literal `\"null\"` string.
3. **`SecretsManager.encryptPasswordFields()`** — propagate the `null` returned by `storeValue()` through to the entity field instead of attempting to `fernet.encrypt(null)`.

### Constraint check vs. #25224

#25224 introduced the null→`\"null\"` conversion specifically because GCP and Kubernetes Secrets Manager *reject empty-string upserts*. This PR satisfies that constraint by exclusion: we never call upsert with an empty value because we short-circuit to delete instead. The flipped Kubernetes test (`testEmptySecretValueShouldNotBeStored`, renamed from `testEmptySecretValueShouldBeStoredAsNullString`) verifies `createNamespacedSecret` is never called.

## Tests

- **New**: `testIssue21259ClearingPasswordDoesNotStoreNullString` exercises the exact user scenario (create service with password → clear password → save) and asserts the entity field is `null`, not a Fernet-wrapped `secret:/` reference. Runs against every provider that subclasses `ExternalSecretsManagerTest` (AWS, AWS-SSM, GCP, Kubernetes).
- **Flipped**: `testEncryptDatabaseConnectionWithEmptyPassword` was asserting the buggy contract (\"empty password should be converted to non-empty encrypted value\"); now asserts the corrected one (entity field is null).
- **Flipped & renamed**: `KubernetesSecretsManagerTest.testEmptySecretValueShouldBeStoredAsNullString` → `testEmptySecretValueShouldNotBeStored`. The previous test name itself encoded the bug.

All 80 tests in `*SecretsManager*Test` pass.

## Out of scope (follow-ups)

- **Read-side cleanup of existing dirty data**: customers who already have services with `\"null\"` strings stored in their SM will continue to read those literal `\"null\"` strings until those services are re-saved. A passive read-side wrapper that maps fetched `\"null\"` → `null` (and optionally the existing `migrate-secrets` CLI tool to actively sweep) can land in a separate PR.
- **`cleanNullOrEmpty()` becomes effectively dead code** in the upsert path. Left in place to keep this PR small and reviewable; can be removed in a follow-up.
- **No-op `LOG.info`** at `AWSBasedSecretsManager` constructor — would be a quality-of-life improvement (today there's zero boot-time confirmation of which SM provider got loaded), but unrelated to the bug.

## Test plan

- [x] `mvn test -pl openmetadata-service -Dtest='*SecretsManager*Test'` (80 tests pass)
- [x] `mvn spotless:apply` (no formatting changes needed)
- [x] Manually reproduced the bug locally with `managed-aws` against AWS SM (output: literal `null`); verified the fix changes the output to `ResourceNotFoundException`
- [ ] Manual regression sanity (set initial password, update password, clear password, recreate service, bot JWT lifecycle, test-connection workflow)

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/secrets/ExternalSecretsManager.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/secrets/ExternalSecretsManager.java
@@ -33,15 +33,20 @@ public abstract class ExternalSecretsManager extends SecretsManager {
   @Override
   protected String storeValue(String fieldName, String value, String secretId, boolean store) {
     String fieldSecretId = buildSecretId(false, secretId, fieldName.toLowerCase(Locale.ROOT));
+    // Issue #21259: a null/empty value means "no credential". Handle this case BEFORE
+    // calling isSecret() — isSecret() invokes startsWith() on the value and would NPE
+    // on null. Returning null here ensures the entity does not carry a stale secret:/
+    // reference to a deleted secret.
+    if (Objects.isNull(value) || value.isEmpty()) {
+      if (store) {
+        upsertSecret(fieldSecretId, value);
+      }
+      return null;
+    }
     // check if value does not start with 'config:' only String can have password annotation
     if (Boolean.FALSE.equals(isSecret(value))) {
       if (store) {
         upsertSecret(fieldSecretId, value);
-      }
-      // Issue #21259: a null/empty value means "no credential" — return null so the
-      // entity does not carry a stale secret:/ reference to a deleted secret.
-      if (Objects.isNull(value) || value.isEmpty()) {
-        return null;
       }
       return SECRET_FIELD_PREFIX + fieldSecretId;
     } else {
--- a/openmetadata-service/src/main/java/org/openmetadata/service/secrets/ExternalSecretsManager.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/secrets/ExternalSecretsManager.java
@@ -38,6 +38,11 @@ public abstract class ExternalSecretsManager extends SecretsManager {
       if (store) {
         upsertSecret(fieldSecretId, value);
       }
+      // Issue #21259: a null/empty value means "no credential" — return null so the
+      // entity does not carry a stale secret:/ reference to a deleted secret.
+      if (Objects.isNull(value) || value.isEmpty()) {
+        return null;
+      }
       return SECRET_FIELD_PREFIX + fieldSecretId;
     } else {
       return value;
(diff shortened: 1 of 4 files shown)


aji-aju:
Reopening from upstream branch (open-metadata/OpenMetadata:fix/21259-null-secrets-test) instead of fork to match the standard contribution pattern in this repo.

PR #27868, opened 2026-05-02
Add text_pattern_ops index on entity-table fqnHash for Postgres listings

harshach:
### Describe your changes:

Service-filtered entity listings (`?service=` / `?database=` / `?databaseSchema=` / `?parent=` / `?apiCollection=` / `?spreadsheet=` / `?testSuite=`) compile to `<table>.fqnHash LIKE 'prefix%'` through `ListFilter.getFqnPrefixCondition`. On Postgres the unique B-tree on `fqnHash` uses default `text_ops` opclass and the column inherits the database default collation (`en_US.UTF-8` on managed Postgres / RDS), so the planner cannot satisfy LIKE prefix from the index — cold count(\*) and the page query degrade to a parallel seq scan over the JSONB heap, measured at ~3s on a ~580k-row `storage_container_entity` even after VACUUM / ANALYZE tuning and an RDS upsize. By contrast, an unfiltered `?limit=15` clears the same table in ~215ms via `idx_storage_container_entity_deleted_name_id` from 1.8.2, which the LIKE predicate cannot use.

Append a `text_pattern_ops` index on `fqnHash` for every entity table that hits `getFqnPrefixCondition` (24 tables, from `chart_entity` through `worksheet_entity`); the opclass supports LIKE prefix regardless of column collation, flipping the cold count(\*) plan from parallel seq scan to bitmap index scan. MySQL is unaffected — every entity-table `fqnHash` column already ships with `CHARACTER SET ascii COLLATE ascii_bin`, a binary collation that lets the existing unique B-tree answer LIKE prefix predicates directly — so the MySQL counterpart only gets a documentation-only comment explaining the asymmetry. Verified no prior migration added pattern-ops on these 24 tables (the only existing instances are 1.11.0's `tag_usage(source, *_lower text_pattern_ops)` on a different table's derived columns).

### Type of change:
- [x] Improvement

### Checklist:

----
## Summary by Gitar

- **Database migrations:**
  - Removed `idx_pipeline_entity_fqnhash_pattern` as `pipeline_entity` utilizes an `EXISTS` join rather than `fqnHash` prefix filtering.
- **Operational guidance:**
  - Added an operator runbook for managing interrupted `CONCURRENTLY` index builds in Postgres.
- **Documentation:**
  - Documented the rationale for excluding `pipeline_entity` from `text_pattern_ops` indexing in `schemaChanges.sql`.

### DIFF
--- a/bootstrap/sql/migrations/native/1.13.0/mysql/schemaChanges.sql
+++ b/bootstrap/sql/migrations/native/1.13.0/mysql/schemaChanges.sql
@@ -296,8 +296,8 @@ CREATE TABLE IF NOT EXISTS rdf_index_server_stats (
 CREATE INDEX idx_er_fromentity_toentity_relation_toid
     ON entity_relationship (fromEntity, toEntity, relation, toId);
 
--- The Postgres counterpart to this file adds a `text_pattern_ops` index on
--- `fqnHash` for every entity table to make `?service=` / `?database=` /
+-- The Postgres counterpart to this file adds a `varchar_pattern_ops` index
+-- on `fqnHash` for every entity table to make `?service=` / `?database=` /
 -- `?databaseSchema=` / `?parent=` listings (which compile to
 -- `fqnHash LIKE 'prefix%'`) index-driven instead of seq-scan-driven on RDS.
 -- MySQL does not need an equivalent: every entity-table `fqnHash` column is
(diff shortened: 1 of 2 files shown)


harshach:
@gitar-bot acknowledged. Both options you suggested (embedded DROP-before-CREATE vs. document the operator runbook) were considered. The embedded form unfortunately does not work cleanly with this repo's migration runner: `MigrationProcessImpl.performSqlExecutionAndUpdate` caches statements by SQL text hash, so a cleanup statement that succeeds once will be skipped on every later retry of a sibling failed CREATE — the cleanup never runs when it would actually be needed. The runner can't be made to re-run cleanups without restructuring it, which is out of scope here. 1.11.0's CONCURRENTLY pass on `tag_usage` carries the same gap.

Went with the runbook option in `1a9bdb02bd`: in-file detection query (`pg_class JOIN pg_index WHERE NOT indisvalid AND relname LIKE 'idx\_%\_fqnhash\_pattern'`) + remediation steps (`DROP INDEX CONCURRENTLY` plus the cached-row delete from `server_migration_sql_logs` so the runner re-attempts the CREATE on the next deploy). Verified column names against `MigrationDAO`.

PR #27926, opened 2026-05-06
fix(log-storage): plug clobber bugs in streamable S3 logs (partial.txt + logs.txt)

pmbrull:
## Backward Incompatible Change

- We are removing one field from the JSON config that it's not needed anymore
- We will still be able to read old log files with the new system

## Summary

Fixes three log-loss bugs in `S3LogStorage` that surfaced when ingestion runs hit idle gaps longer than the 5-minute stream timeout:

- **`partial.txt` 80MB→KB clobber.** `cleanupExpiredStreams` reset `partialLogOffsets[streamKey]` after the idle threshold; the next `writePartialLogsForStream` then took the `currentOffset == 0` path and PUT only the in-memory tail — overwriting the prior 80MB body. Replaced with an always-merge flush: GET existing `partial.txt`, append a snapshot of the new `pendingFlush` queue, PUT (with offset state in S3 user metadata).
- **`logs.txt` clobber.** The 5-minute sweeper called `completeMultipartUpload`, materializing `logs.txt`. Resumed appends opened a fresh multipart upload to the same key; the next finalize overwrote `logs.txt` with only the latest segment. Removed the multipart upload from the write path entirely; `logs.txt` is now produced only on `/close` (or by the abandoned-run sweeper) via a server-side `CopyObjectRequest` from `partial.txt`.
- **`SimpleLogBuffer` overflow drift.** `partialLogOffsets` indexed into a 1000-line ring; eviction caused contiguous-line gaps to be silently dropped from `partial.txt`. Replaced with `pendingFlush` (unbounded, watermark-drained) and a monotonic `totalLinesAppended` counter. `SimpleLogBuffer` is now SSE-tail-cache only.

The cleanup sweeper threshold moves from 5 min to **24h** (configurable as `streamTimeoutHours`), interval from 1 min to **1h** (configurable as `cleanupIntervalMinutes`). New configurable knobs: `partialFlushIntervalMinutes`, `earlyFlushWatermarkBytes`, `pendingFlushAlertAfterFailures`. `streamTimeoutMinutes` is deprecated.

Backward-compatible at the artifact level: same on-disk file names (`partial.txt`, `logs.txt`), same API endpoints, no changes to the Python ingestion client.

Design spec: [`docs/superpowers/specs/2026-05-05-streamable-logs-stability-design.md`](docs/superpowers/specs/2026-05-05-streamable-logs-stability-design.md)
Feature doc: [`docs/streamable-logs.md`](docs/streamable-logs.md)

## Test Plan

- [ ] `mvn test -pl openmetadata-service -Dtest='S3LogStorageTest,LogStorageFactoryTest,LogStorageTest'` → 53 tests, BUILD SUCCESS (verified locally)
- [ ] `mvn test-compile -pl openmetadata-integration-tests` → BUILD SUCCESS (verified locally)
- [ ] In a deployment with S3 log storage configured, run a long ingestion with idle gaps > 5 min and verify `partial.txt` versions on S3 grow monotonically (no shrink to KB)
- [ ] Verify `/close` produces `logs.txt` byte-equal to the final `partial.txt`
- [ ] Verify a second `/close` for the same `(fqn, runId)` returns gracefully (idempotent)
- [ ] Verify abandoned runs (no `/close`, no appends for 24h) get `logs.txt` materialized by the sweeper
- [ ] Verify SSE/live-tail UI experience is unchanged for active runs

----
## Summary by Gitar

- **Refactored S3 log storage synchronization:**
  - Replaced per-stream `ReentrantLock` map with Guava `Striped<Lock>` to prevent memory leaks and race conditions during concurrent stream processing.
- **Improved log storage reliability:**
  - Implemented server-side `MultipartUpload` and `UploadPartCopy` for `partial.txt` to avoid holding large logs in JVM heap and prevent data clobbering.
  - Updated `buildPartialMetadata` to calculate `total-bytes` using UTF-8 byte length instead of string character count.
- **Enhanced configuration flexibility:**
  - Added configurable parameters `cleanupIntervalMinutes`, `partialFlushIntervalMinutes`, `earlyFlushWatermarkBytes`, and `pendingFlushAlertAfterFailures` to `LogStorageConfiguration`.
- **Test suite updates:**
  - Added `S3LogStorageTest` coverage for multipart upload failure handling and large file processing.
  - Adjusted `IngestionPipelineLogStreamingResourceIT` to verify sequential log bursts persist correctly.


IceS2 (approved):

pmbrull:
picked by hand to 1.13

PR #27949, opened 2026-05-06
feat(cli-e2e-v2): Slice 1 — framework + MySQL pilot connector

IceS2:
## Summary

Introduces a composition-based connector E2E framework with MySQL as the reference connector. The goals are real end-to-end coverage and tests that are easy to understand, debug, maintain, and extend.

```text
Owned source data -> real metadata CLI -> real OpenMetadata server -> persisted-state assertions
```

The existing v1 suite remains in place; this PR does not remove it or migrate other connectors.

## Design

All framework paths below are under `ingestion/tests/cli_e2e_v2/`.

- **`runtime/`** composes `WorkflowCase(invocation, persisted, check)`: execute the CLI once, then poll fresh observations until the entire checker passes. Exit code, structured status, and exact record-error counts are validated independently. Only assertion mismatches retry; transport, parsing, and programming errors propagate.
- **`features/database/`** provides database workflow configuration, independently declared catalog/type expectations, and reusable catalog, profile, sample, foreign-key, and lineage checks. Strict catalog matching rejects missing, unexpected, duplicated, and incorrectly linked entities.
- **`mysql/`** owns a pinned disposable MySQL container, restricted ingestion account, fresh per-test schemas, declared seeds, connector configuration, and scenarios. Each test owns an OpenMetadata service; cleanup runs on success and failure. A running OpenMetadata server and authentication are supplied separately.
- **`contracts/`** provides the reusable pytest workflow test and connector-owned coverage inventory. `--e2e-contract-check` rejects missing, duplicate, unknown, skipped, or xfailed required cases and partial suite selection.

This replaces the mutable fluent assertion API and baseline-repair hierarchy with ordinary pytest fixtures and composed queries/checks. New connectors supply their own source setup, configuration, expectations, and scenarios without adding runtime subclasses or a central connector registry.

## Coverage

The MySQL suite contains **20 feature scenarios plus 11 source/fixture tests**. It preserves all 12 scenarios from the initial implementation and extends coverage:

- Catalog structure, native column types, primary keys, nullability, comments, and view identity.
- Stored-procedure bodies, fully qualified foreign keys, view DDL, and table/column lineage.
- Profiler row counts and representative numeric/string metrics, including nulls, duplicates, and freshness after source mutations.
- Native sample values, null rows, and replacement of previously persisted samples.
- Positive and negative PII classification checks.
- Table include/exclude/regex precedence and filtering between two populated schemas.
- Repeated ingestion, stable entity IDs, metadata updates, and soft deletion without deleting retained tables.
- Broken-view containment with an explicitly expected CLI failure and continued ingestion of healthy tables.

**320 offline framework tests** exercise success and failure behavior, including subprocess/status handling, polling composition, strict assertions, cleanup, reporting, and inventory enforcement. Focused production tests cover configuration-log redaction and the dbt status-file contract.

Before pushing `7be95bc879`, the combined offline/framework and focused unit run completed with **347 passed**, contract collection found all **31 MySQL tests**, and pre-commit checks passed. This does not establish that the complete live connector suite is green.

## CLI, logging, and CI

- Adds optional structured workflow status output, including `ingest-dbt --status-file`, for independent exit/status validation by the E2E runner.
- Subprocess stdout/stderr use ordinary pytest capture; `-s` streams logs locally. There is no custom E2E log publisher or redaction layer. Product code owns redaction, with CI-native secret masking as additional protection.
- Temporary workflow configurations and status files are not published as diagnostics. The live workflow retains its existing JUnit report upload.
- The live connector workflow, `.github/workflows/py-cli-e2e-tests-v2.yml`, is **manually triggered** and enforces complete connector coverage.
- The shared Python workflow runs offline framework tests on Python 3.10 when E2E code, relevant dependency/test configuration, generated-model schemas, or CI wiring changes. It does not automatically run live connector tests.

## Scope and known limitations

- Strict MySQL YEAR/BIT sample assertions and profiler `useStatistics`/freshness checks retain known live failures. The corresponding production fixes are deferred; the tests do not use skips, xfails, or relaxed expectations to hide them.
- Dashboard connectors are design-validated through composition examples and offline framework checks only. No live Metabase tests ship in this PR.
- DQ, usage, and query-log lineage E2Es remain out of scope. The database invocation helper explicitly rejects the unsupported usage pipeline.
- Standalone lineage CLI status support and broader production crash/status and validation-log redaction issues remain deferred.

## Documentation

- [Run and debug the suite](https://github.com/open-metadata/OpenMetadata/blob/feat/cli-e2e-v2-slice1/ingestion/tests/cli_e2e_v2/README.md)
- [Add a connector and compose custom scenarios](https://github.com/open-metadata/OpenMetadata/blob/feat/cli-e2e-v2-slice1/ingestion/tests/cli_e2e_v2/CONNECTORS.md)

<h2><a href="https://app.greptile.com/api/retrigger?id=37814977"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; no new actionable defect remains, and both previous threads are resolved.

The PR introduces a composition-based CLI E2E framework and a live MySQL pilot suite, together with structured workflow status output and CI integration.
- Adds isolated source/service fixtures, reusable persisted-state checks, strict contract inventory enforcement, and extensive offline framework coverage.
- Routes supported CLI workflows through common execution and optional JSON status-file handling.
- Adds manually triggered live MySQL coverage and conditionally runs offline framework tests in shared Python CI.
</details>

```mermaid
flowchart LR
  F[Owned MySQL fixture] --> C[metadata CLI subprocess]
  C --> W[Workflow execution]
  W --> S[OpenMetadata server]
  W --> J[Structured status JSON]
  S --> P[Fresh persisted-state queries]
  J --> V[Exit, success, and error-count validation]
  P --> A[Polling assertions]
  V --> R[Test result]
  A --> R
```
</details>

<sub>Reviews (3) · Last reviewed commit: ["Rename workflow status source identifier"](https://github.com/open-metadata/openmetadata/commit/c3b74e7bcc05620d4243a9ad63041e861d7f7c24)</sub>


ingestion/tests/cli_e2e_v2/core/config/env.py
```diff
@@ -0,0 +1,103 @@
...
+class EnvLoadError(RuntimeError):
+    """Raised when a required env var is unset (or empty)."""
+
+
+# Constrained — callers can only parameterize Env with True or False, matching
+# the two concrete `required` states. Anything else is a type error.
+_Req = TypeVar("_Req", Literal[True], Literal[False])
+
+
+class Env(Generic[_Req]):
```
NiltonDuarte: Great typing communicates intent across time. Right now, a signature like Env[True] is a bit opaque. We can make it entirely self-documenting by introducing semantic aliases:

```python
# Constrained — callers can only parameterize Env with Required (True)  or NotRequired (False), matching
# the two concrete `required` states. Anything else is a type error.
Required: TypeAlias = Literal[True]
NotRequired: TypeAlias = Literal[False]
_Req = TypeVar("_Req", Required, NotRequired)
```
This allows us to write/hint Env[Required] or Env[NotRequired], whic tells the developer what is expected. While it might be a minor detail for this specific, simple class, adopting this pattern sets a great standard for readability across the codebase.

ingestion/tests/cli_e2e_v2/core/config/env.py
```diff
@@ -0,0 +1,103 @@
...
+    @overload
+    def __new__(
+        cls,
+        key: str,
+        default: str | None = None,
+        *,
+        required: Literal[False],
+    ) -> Env[Literal[False]]: ...
+
+    def __new__(
```
NiltonDuarte: Overriding __new__ with a custom signature breaks standard object instantiation conventions and feels hacky here.

We can achieve the exact same type resolution much more cleanly by using @overload on __init__. This respects standard interfaces and keeps the logic straightforward without having to touch __new__. Here is how that looks:

```python
    # Two __init__ overloads — one per Literal[required] value — let the type
    # checker pick the right `Env[NotRequired|Required]` specialization at the call
    # site.
    @overload
    def __init__(
        self: Env[Required],
        key: str,
        default: str | None = None,
        *,
        required: Literal[True],
    ) -> None: ...

    @overload
    def __init__(
        self: Env[NotRequired],
        key: str,
        default: str | None = None,
        *,
        required: Literal[False] = False,
    ) -> None: ...

    def __init__(
        self,
        key: str,
        default: str | None = None,
        *,
        required: bool = True,
    ) -> None:
        self.key = key
        if default is not None:
            os.environ.setdefault(key, default)
        if required and not os.environ.get(key):
            raise EnvLoadError(
                f"required env var {key} not set. Set it in your shell or GitHub Actions secrets."
            )

    # Two .get() overloads narrow by the specialization of Env:
    #   Env[Required].get()  -> str          (construction validated)
    #   Env[NotRequired].get() -> str | None   (caller must handle None)
    @overload
    def get(self: Env[Required]) -> str: ...

    @overload
    def get(self: Env[NotRequired]) -> str | None: ...

    def get(self) -> str | None:
        return os.environ.get(self.key)
```

ingestion/tests/cli_e2e_v2/mysql/conftest.py
```diff
@@ -0,0 +1,231 @@
...
+            # container instead of reusing a stale URL.
+            get_admin_engine.cache_clear()
+            get_policy.cache_clear()
+
+
+@pytest.fixture(scope="session")
+def mysql_service(session_uuid: str) -> str:
+    """Session-shared MySQL service name (``e2e_mysql_<uuid>``).
+
+    Eliminates ``service = mysql_service_name(session_uuid)`` from every
```
NiltonDuarte: imo adding another level of implicit indirection to save one line is not worth

ingestion/tests/cli_e2e_v2/mysql/conftest.py
```diff
@@ -0,0 +1,231 @@
...
+    "E2E_MYSQL_PASSWORD",
+    "E2E_MYSQL_HOST_PORT",
+    "E2E_MYSQL_ADMIN_USER",
+    "E2E_MYSQL_ADMIN_PASSWORD",
+    "E2E_MYSQL_DATABASE",
+)
+
+
+@pytest.fixture(scope="session")
+def mysql_container() -> Generator[MySqlContainer, None, None]:
```
NiltonDuarte, line 63: I dont see this fixture being used directly, its a implicit order of execution dependency scoped. Id avoid implicit dependency on the test and make this some kind of scoped provider such that tests could  explitly creates its own containers.

In tests I find easier to understand when you have the steps explicit, Given -> When -> Then, this implicit dependency is hiding the "Given" step and when looking to the test is unclear what is the current state it is being tested

NiltonDuarte: Also, pytest fixture has `autouse=True` if all you want is to run this whenever this fixture is "available"

ingestion/tests/cli_e2e_v2/mysql/baseline.py
```diff
@@ -0,0 +1,298 @@
...
+    stored_procedures=[_SP_ACTIVE_CUSTOMER_COUNT, _SP_UPDATE_CUSTOMER_STATUS],
+)
+
+
+# -----------------------------------------------------------------------------
+# Policy factory
+# -----------------------------------------------------------------------------
+
+
+@lru_cache(maxsize=1)
```
NiltonDuarte: as we have a scoped fixture providing this, do we need a cache?

NiltonDuarte: also for `get_policy` below

ingestion/tests/cli_e2e_v2/core/fixtures.py
```diff
@@ -0,0 +1,81 @@
...
+#  Licensed under the Collate Community License, Version 1.0 (the "License");
+#  you may not use this file except in compliance with the License.
+"""Shared helpers for per-connector pytest fixtures.
+
+Per-connector conftests (`<connector>/conftest.py`) wire their own
+`<connector>_source_ready` and `<connector>_metadata_ingested` fixtures on
+top of these helpers instead of copy-pasting the body.
+
+Design:
+  - These are PLAIN FUNCTIONS (not pytest fixtures). The per-connector
```
NiltonDuarte: nit: why file is named "fixture" if there are no fixtures 🫠

ingestion/tests/cli_e2e_v2/meta/test_eventually.py
```diff
@@ -0,0 +1,145 @@
...
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from ..core.fluent.eventually import EventuallyRunner, retry_until
+
+
+def _attempt_counter():
```
NiltonDuarte: 😄 this is funny, AI is getting wild here. Please let me know if Im wrong 🤔 
1- you either need a mutable object or closure, **you dont need both**. 
1.1- due to the nature of check you will need a closure
2- a function to return an empty list 🤣

ingestion/tests/cli_e2e_v2/mysql/test_mysql.py
```diff
@@ -0,0 +1,530 @@
...
+        uniqueCount=5,
+        min=600,
+        max=750,
+        mean=680,
+        sum=3400,
+        median=680,
+    )
+
+    # String column — first_name: Alice(5), Bob(3), Charlie(7), Diana(5), Eve(3).
+    om_client.table(customers_fqn).profile.eventually().column("first_name").has_metrics(
```
NiltonDuarte, line 137: would running all those `eventually` in coroutines save time? looks like there is no difference. when the first is done waiting, all others should be good to go?
if waiting in parallel saves time we could run it in coroutines but dont feel its worth the time to make all async

NiltonDuarte: snippet on async eventually + fluent future + context gathering.
the code awaits for a total of 65 seconds and runs in 16 seconds
```python
import asyncio
import time


def coro_task1():
    start = time.monotonic()

    def _inn():
        end = start + 10
        if time.monotonic() <= end:
            print("Waiting on coro 1")
            raise AssertionError("11")
        return "coro1"

    return _inn


def coro_task2():
    start = time.monotonic()

    def _inn():
        end = start + 15
        if time.monotonic() <= end:
            print("Waiting on coro 2")
            raise AssertionError("22")
        return "coro2"

    return _inn


def coro_fail():
    print("FAIL")
    raise RuntimeError("failed")


async def retry_until(
    check,
    *,
    timeout: int = 60,
    poll_interval: float = 2,
    name: str = "check",
):
    start = time.monotonic()
    deadline = start + timeout
    attempts = 0

    while True:
        attempts += 1
        try:
            print(f"Checking on {name}")
            return check()
        except AssertionError as exc:
            if time.monotonic() >= deadline:
                elapsed = time.monotonic() - start
                raise AssertionError(
                    f"eventually[{name}] timed out after {attempts} attempts "
                    f"over {elapsed:.1f}s (timeout={timeout}s, "
                    f"poll_interval={poll_interval}s).\n"
                    f"Last failure:\n{exc}"
                ) from exc
            await asyncio.sleep(poll_interval)


class FluentFuture:
    """Wraps a task to allow fluent, synchronous-looking chaining."""

    def __init__(self, task: asyncio.Task) -> None:
        self.task = task
        self.assert_task = None

    def assert_eq(self, val):
        async def _assert_coro():
            res = await self.task
            assert res == val, f"Fail: Expected '{val}', got '{res}'"
            print(f"Asssert  success: got {res}")
            return res

        assertion_task = asyncio.create_task(_assert_coro())

        self.assert_task = assertion_task
        return

    def assert_exc(self, exc):
        async def _assert_coro():
            try:
                await self.task
            except exc as e:
                print(f"Found exception {exc}: {e}")
                return
            return AssertionError("Exception not found")

        assertion_task = asyncio.create_task(_assert_coro())

        self.assert_task = assertion_task
        return

    def waitable(self):
        return self.assert_task or self.task


class EventuallyRunner:
    def __init__(self) -> None:
        self.ffs = set()
        self._timeout = None

    def arm(self, timeout: int) -> "EventuallyRunner":
        self._timeout = timeout
        return self

    def run(self, check, *, name: str):
        if self._timeout is not None:
            timeout = self._timeout
            self._timeout = None
            t = asyncio.create_task(retry_until(check, timeout=timeout, name=name))
        else:

            async def _wrap():
                return check()

            t = asyncio.create_task(_wrap())
        ff = FluentFuture(t)
        self.ffs.add(ff)
        return ff

    async def gather(self):
        await asyncio.gather(*map(lambda ff: ff.waitable(), self.ffs))

    async def __aenter__(self):
        return self

    async def __aexit__(self, exc_type, exc, tb):
        await self.gather()


async def main():
    start = time.monotonic()
    async with EventuallyRunner() as runner:
        runner.arm(30).run(coro_task1(), name="coro1").assert_eq("coro1")
        runner.arm(30).run(coro_task1(), name="coro1.1").assert_eq("coro1")
        runner.arm(30).run(coro_task1(), name="coro1.2").assert_eq("coro1")
        runner.arm(30).run(coro_task1(), name="coro1.3")
        runner.arm(30).run(coro_fail, name="fail").assert_exc(RuntimeError)
        runner.arm(30).run(coro_task1(), name="coro1.4")
        runner.arm(30).run(coro_task2(), name="coro2").assert_eq("coro2")
    print(f"Loop run for {time.monotonic() - start}s")


if __name__ == "__main__":
    asyncio.run(main())

```

ingestion/tests/cli_e2e_v2/core/config/builder.py
```diff
@@ -0,0 +1,170 @@
...
+                dumped.setdefault(key, prior_cfg[key])
+
+        new_doc["source"]["sourceConfig"]["config"] = dumped
+
+        # `import_source_class` dispatches by suffix (e.g. "-lineage", "-usage").
+        base_connector = new_doc["source"]["type"].split("-", 1)[0]
+        new_doc["source"]["type"] = base_connector + source_type_suffix_for(options)
+
+        if isinstance(options, _PIPELINES_NEEDING_PROCESSOR):
+            new_doc["processor"] = {"type": "orm-profiler", "config": {}}
```
edg956: This should be `tag-pii-processor` so they use the new classification processor

```suggestion
            new_doc["processor"] = {"type": "tag-pii-processor", "config": {}}
```

ingestion/tests/cli_e2e_v2/core/config/builder.py
```diff
@@ -0,0 +1,170 @@
...
+                return
+            pattern = cfg.setdefault(key, {})
+            if includes:
+                pattern.setdefault("includes", []).extend(includes)
+            if excludes:
+                pattern.setdefault("excludes", []).extend(excludes)
+
+        _merge("databaseFilterPattern", databases_include, databases_exclude)
+        _merge("schemaFilterPattern", schemas_include, schemas_exclude)
+        _merge("tableFilterPattern", tables_include, tables_exclude)
```
edg956, line 146: Probably good for now, but we may want to refactor this to support the different filtering expressions for different entities

ingestion/tests/cli_e2e_v2/core/config/builder.py
```diff
@@ -0,0 +1,170 @@
...
+        return cli_subcommand_for(self._options)
+
+    # --- rendering ------------------------------------------------------
+    def write_tmp(self, tmp_path: Path, invocation: int = 0) -> Path:
+        """Dump to `<tmp_path>/cfg_<id>_<invocation>.yaml` and return the path."""
+        if self._options is None:
+            raise PipelineNotSetError("pipeline not set — call .pipeline(options) before rendering")
+        path = tmp_path / f"cfg_{self.pipeline_identifier}_{invocation}.yaml"
+        path.write_text(yaml.safe_dump(self._doc, sort_keys=False))
+        return path
```
edg956, line 172: nitpick: If the caller can pass any path, why don't we simply name this function `write` or `dump`? that let's callers decide whether it should write to a temporary file or not

ingestion/tests/cli_e2e_v2/mysql/test_mysql.py
```diff
@@ -0,0 +1,391 @@
...
+                # 5 seed rows push the combined score near the 80% boundary; 60 aligns with PII minimumConfidence.
+                confidence=60,
+            )
+        ).with_filter(schemas_include=["e2e"])
+    )
+    assert status.success, f"auto-classification failures: {status.all_failures}"
+
+    customers_fqn = f"{mysql_service}.default.e2e.customers"
+
+    om_client.table(customers_fqn).column("email").has_tag("PII.Sensitive")
```
edg956, line 248: This applies to other places as well, but here's it's really obvious we're fetching several times the same entity to access each of their columns.

Not like the penalty incurred is important, but perhaps for simplicity's sake the table could be retrieved once and then apply expressions like these on it?

I'm thinking of being able to produce assertions like this:

```python
om_client.table(customers_fqn).eventually()   # fetches everything once
    .check(
        ColumnAssert("email").has_tag("PII.Sensitive") &
        ColumnAssert("date_of_birth").has_tag("PII.NonSensitive")
    )
```

Borrowing the operator-style composition used in libraries like `dirty_equals`.

If taken a notch further we could probably rely on `dirty_equals` (or extend it with the current assertion classes) to follow a pattern that's been introduced elsewhere in the ingestion test suite

```python
def test_profiled_and_classified_table(om_client) -> None:
    assert om_client.table("foo.bar").eventually() == IsTable(
        columns=ColumnSubSet(
            ColumnAssert("email").has_tag("PII.Sensitive"),
            ColumnAssert("date_of_birth").has_tag("PII.NonSensitive")
        )
    ).with_row_count(8)
```

Where the entity is fetched once and then we make sure the entity meets some expected expression

edg956: Maybe that mixing of paradigms just complicates matters more

ingestion/tests/cli_e2e_v2/mysql/conftest.py
```diff
@@ -0,0 +1,190 @@
...
+
+@pytest.fixture(scope="session")
+def mysql_policy(mysql_admin_engine: Engine) -> EnforcementPolicy:
+    """Session MySQL EnforcementPolicy backed by the admin engine."""
+    enforcer = MySqlEnforcer(mysql_admin_engine, MYSQL_BASELINE)
+    return EnforcementPolicy(enforcer=enforcer, mode=EnforcementMode.APPLY)
+
+
+@pytest.fixture(scope="session")
+def mysql_source_ready(mysql_policy: EnforcementPolicy) -> None:
```
NiltonDuarte, line 156: imHo still running too much implicit dependency on the test execution flow, i.e. Id remove this fixture and **explicit** call `ensure_baseline` whenever needed, same with `mysql_metadata_ingested` that is a fixture that all mysql tests seems to depends on. Can we make it in such a way it is only executed once at module level and still make it explicit?

ingestion/tests/cli_e2e_v2/mysql/test_mysql.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+def test_vanilla_ingest_structural(
+    om_client: OmClient,
+    mysql_expected_factory: Callable[..., ExpectedService],
+    # `mysql_metadata_ingested: None` triggers the fixture side-effect; value is always None.
+    mysql_metadata_ingested: None,
+) -> None:
+    """Metadata ingest produces the full declared catalog (SUPERSET match)."""
+    assert_service_matches(mysql_expected_factory(), om_client)
```
NiltonDuarte, line 62: for this case it is easy to catch that this dependency is somehow the test "object", still would give a try at Given / When / Then
```python
def test_vanilla_ingest_structural(
    om_client: OmClient,
    mysql_expected_factory: Callable[..., ExpectedService],
    # `mysql_metadata` return a scoped level singleton
    mysql_metadata: PytestScopedSingleton,
) 
    """Metadata ingest produces the full declared catalog (SUPERSET match)."""
    # GIVEN
    mysql_metadata.ingest() # only run once per scope
    # WHEN
    # THEN
    assert_service_matches(mysql_expected_factory(), om_client)
```

NiltonDuarte (commented):
being nit on the testing structure

ingestion/src/metadata/cli/lineage.py
```diff
@@ -46,7 +46,7 @@ class LineageWorkflow(BaseModel):
     parserType: QueryParserType | None = QueryParserType.Auto  # noqa: N815
 
 
-def run_lineage(config_path: Path) -> None:
+def run_lineage(config_path: Path, status_file: Path | None = None) -> None:
```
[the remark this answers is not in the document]
IceS2: Confirmed: the standalone `metadata lineage` command accepts `--status-file` but does not write it. That command-specific status contract remains deferred from this PR; this is not being treated as fixed.

The MySQL E2E lineage scenario invokes `metadata ingest` with `source.type=mysql-lineage`, which goes through `execute_workflow` and produces structured status. The E2E runner requires that status file, so this standalone-command gap does not weaken the lineage assertions in this suite.

Khairajani, line 49: `status_file` is accepted but never used. `metadata lineage --status-file s.json` exits 0 and writes nothing — and if an old `s.json` is there it's left behind, so callers read a stale result as if it were this run's.

Fine to defer the feature (no real workflow status here to write), but let's not ship the flag. This is the only one of the 8 subcommands that can't honour it — give it a parser without `--status-file`.

ingestion/tests/cli_e2e_v2/runtime/ci.py
```diff
@@ -0,0 +1,25 @@
...
+
+def mask_secrets(*values: str) -> None:
+    """Register nonempty values with GitHub Actions before they reach logs."""
+    if os.environ.get("GITHUB_ACTIONS") != "true":
+        return
+    for value in values:
+        if value:
+            escaped = value.replace("%", "%25").replace("\r", "%0D").replace("\n", "%0A")
+            # xdist workers reserve stdout for their transport; stderr still reaches the CI runner.
+            print(f"::add-mask::{escaped}", file=sys.stderr, flush=True)
```
Khairajani, line 25: On the CodeQL alert here — false positive, please don't change the code. `::add-mask::` is what *hides* a value from the logs, not what prints it. Also gated on `GITHUB_ACTIONS=true`, and these are throwaway test container creds.

Let's dismiss alert 2403 with that reason. Removing this line is what would actually leak.

ingestion/src/metadata/workflow/base.py
```diff
@@ -507,3 +513,20 @@ def print_status(self):
...
+        """Write per-step status as JSON to `path`.
+
+        `success` is True iff every step meets its success threshold.
+        Shape: {"pipeline_type": str | None, "ingestion_pipeline_fqn": str | None, "success": bool, "steps": list}
+        """
+        ingestion_status = self.build_ingestion_status()
+        success = all(self._step_meets_success_threshold(step) for step in self.workflow_steps())
+        source = getattr(self.config, "source", None)
+        payload = {
+            "pipeline_type": getattr(source, "type", None) or getattr(self.config, "sourcePythonClass", None),
```
Khairajani: `source.type` is the *service* type (`mysql`, `simple`), not a pipeline type (`metadata`/`profiler`/`usage`). Misnamed field.

Would normally be a nit, but the test pins it (`payload["pipeline_type"] == "simple"`), so once this ships the E2E runner and anything downstream depend on it. One line to rename now, breaking later. `service_type`?

ingestion/src/metadata/cli/common.py
```diff
@@ -13,14 +13,23 @@
...
+    config_dict: dict[str, Any],
+    status_file: Path | None = None,
+) -> None:
+    """Execute the workflow, write status file if requested, raise on failure if configured."""
+    try:
+        workflow.execute()
+    finally:
+        workflow.stop()
+        if status_file is not None:
+            workflow.write_status_file(status_file)
```
Khairajani, line 33: Writing in `finally` is right, but a failed write replaces the real error — workflow dies on "can't connect to db", status write throws on a missing folder, user sees the path error and the real one is buried in `__context__`.

Deliberate (there's a test for it) and OK for an opt-in flag. Could we at least `mkdir(parents=True, exist_ok=True)` first? Kills the likeliest cause. Non-blocking.

.github/workflows/py-tests-shared.yml
```diff
@@ -217,6 +231,15 @@ jobs:
           name: spacy-en-model
           path: /tmp/spacy-model
 
+      - name: Run offline E2E framework tests
```
Khairajani, line 234: This step won't run on this PR. `py-tests.yml` is `pull_request_target`, so the workflow file comes from main even though the code comes from the PR — the step isn't "skipped" in the 3.10 job, it's absent.

And the existing unit job doesn't catch them either: `nox -s unit-tests` → `pytest tests/unit/`, which doesn't reach `tests/cli_e2e_v2/meta`.

So the ~320 offline tests haven't run in CI here. Only evidence is the local run at `7be95bc87`, ~20 commits back. File's fine as-is — just re-run the suite at head and paste the output?

Khairajani (commented):
Framework reads well and the production changes look right to me. Main hesitation: nothing has actually run the new suite at this commit. The v2 workflow is dispatch-only (fine, that's the plan), and the offline tests didn't run either for the `pull_request_target` reason I left inline.

Also noting the description is upfront that some MySQL cases fail on purpose (YEAR/BIT samples, `useStatistics`/freshness), unhidden. I like that it's honest — but with `--e2e-contract-check` rejecting skips/xfails, v2 can't go green until those fixes land, so we can't retire MySQL from the old workflow yet.

Before I approve:

1. Dispatch `py-cli-e2e-tests-v2` at head and share results + which cases are expected red. Known failures are fine if listed; I'd just rather not merge 8.5k lines last verified 20 commits ago.
2. Re-run `tests/cli_e2e_v2/meta` at head.
3. Link an issue (Validate PR Metadata is failing on it) + dismiss the CodeQL alert as a false positive.

Only one actual code change requested (the `lineage` flag); the rest is gates and one naming nit.

ingestion/src/metadata/cli/lineage.py
IceS2: Defered

ingestion/src/metadata/cli/common.py
IceS2: Not fixing

.github/workflows/py-tests-shared.yml
IceS2: Does not matter at the moment - they are meta tests.

ingestion/src/metadata/workflow/base.py
IceS2: Good catch

Khairajani (approved):

mohittilala (approved):

akash-jain-10 (approved):

PR #27987, opened 2026-05-08
fix(alerts): strict literal matching across alert filter functions

manerow:
Fixes open-metadata/openmetadata-collate#4019

### Summary

Fixes a bug where alert filters dropped events if entity names contained regex metacharacters (e.g., `[TML] Fraud Mart Suite`).

Previously, the UI allowed free-text input while the backend applied inconsistent regex or substring evaluations. This PR standardizes the contract across the stack: **strict-pick UI inputs ⇔ literal equality backend evaluation**.

### Backend Changes (`AlertsRuleEvaluator.java`)

Removed all undocumented regex/substring evaluations in favor of literal equality (`String.equals` or `List.contains`).

| Function | Before | After |
| --- | --- | --- |
| `matchAnyEntityFqn` | Raw regex (`matcher.find()`) | Literal `List.contains` |
| `testSuiteMatcher` | Raw regex on FQNs | Literal `List.contains` |
| `filterByTableNameTestCaseBelongsTo` | Regex with word boundaries `\b` | Literal equality on parsed parent FQN |
| `filterByEntityNameDataContractBelongsTo` | Substring match (`anyMatch(contains)`) | Literal `List.contains` |

* **Defense-in-depth:** Added a null `changeEvent` guard to `filterByEntityNameDataContractBelongsTo` (fixing a latent bug exposed by SpEL validators).
* **Housekeeping:** Updated `@Function` annotations, parameters, and `ExpressionValidator` allowlists to reflect the new literal semantics.

### Policy expression validator (`ExpressionValidator.java`)

Customer-reported follow-up: a test suite named like `AENG - CSP work item bug checks (duration exceeded)` was rejected with `Function 'checks' is not allowed in policy expressions`. Same root cause class as the metachar bug — the regex-based validator scanned the entire expression string and matched `checks(` as a function call inside the quoted argument.

Replaced the regex-based validator with **AST-based default-deny validation**:

| Before | After |
| --- | --- |
| Regex pass over the surface string | Parse with `SpelExpressionParser`, walk the AST |
| String-literal content scanned as syntax | `StringLiteral` nodes never inspected as code |
| Accumulated band-aids for false positives (`(?<![\w$])`, `\s+`, hardcoded `and`/`or`/`not` exceptions) | Whitelist of safe `SpelNode` classes; every other node rejected |

* **Allowed nodes:** literals, boolean/comparison operators, list/map literals, ternaries, and `MethodReference`s whose name is on the existing function allowlist.
* **All previously-blocked patterns remain blocked** (`T(...)`, `new`, `System.`, `Runtime`, `ClassLoader`, `.forName()`, `.exec()`, `eval()`, `ProcessBuilder`, `java.lang.reflect`) — now by node type rather than regex pattern, eliminating the bypass surface a regex carries.

### Data Insight chart formula evaluation (`DataInsightFormulaEvaluator.java`, new)

Tightening the validator surfaced a second false-positive class: Data Insight chart aggregators (`Elastic/OpenSearchDynamicChartAggregatorInterface`) route through the same `CompiledRule.parseExpression` entry point, but use SpEL purely as a *calculator* on regex-gated numeric strings.

Formula evaluation flow at the aggregator:
1. User formula like `(count(...) / count(...)) * 100` is stored on `DataInsightCustomChart`
2. ES/OS aggregation results are substituted for each placeholder → `(127.0 / 384.0) * 100`
3. The substituted string is gated by `NUMERIC_VALIDATION_REGEX = [\d\.+-\/\*\(\) ]+` (digits, decimal, `+ - * /`, parens, space only), provably free of identifiers, methods, or SpEL syntax
4. The arithmetic expression is parsed and evaluated to a `Double`

The AST validator (correctly) rejects arithmetic operators like `OpMultiply` because policy/alert conditions never need them. But DI formulas are pure arithmetic. The two threat models accidentally shared one validator entry point. Pre-PR the regex-based validator was a no-op on these strings (no `DANGEROUS_PATTERN` and no `[a-zA-Z0-9_]+(` match); the new AST validator rejected them.

**Fix:** extracted DI formula evaluation into a new `DataInsightFormulaEvaluator` util (with `NUMERIC_VALIDATION_REGEX` colocated). The two aggregators now use this util directly, bypassing `ExpressionValidator`. The regex remains the security boundary for DI; the validator stays focused on the policy/alert allowlist.

`ExpressionValidatorTest.testArithmeticIsRejectedByPolicyValidator` asserts that the validator continues to reject arithmetic. That path is now exclusively for DI formula evaluation.

### UI Changes (`AlertsUtil.tsx`)

Switched four filter renderers from `mode="tags"` (free-text) to `mode="multiple"` (strict dropdown pick) to prevent invalid inputs:

* `fqnList`
* `tableNameList`
* `entityNameList`
* `entityIdList`

**Enhancements to `entityIdList`:**

* Added a richer dropdown (`getEntityByIdSuggestions`) supporting typeahead by entity name.
* Supports a paste-UUID flow: if text matches a UUID pattern, it augments the search with an exact `id` term filter.
* UI renders the UUID on top and FQN below (muted), but selected chips and the backend contract strictly retain the UUID.

### Tests

* **Integration (`AlertsRuleEvaluatorResourceIT.java`):** 10 new tests covering literal matching, regex metacharacters, prefix-collision rejections, and entityLink fallbacks. Crucially, verified substring/regex payloads now correctly return `false`. `@ValueSource` extended with the customer-reported case `AENG - CSP work item bug checks (duration exceeded)`.
* **Policy validator (`ExpressionValidatorTest.java`):** Added a corpus pinning the literal-content-not-syntax contract (parens, dotted identifiers, dangerous-looking tokens inside quoted arguments, doubled-quote escape, malformed expressions). Every pre-existing `_Negative` case (`T()`, `new`, `System.`, `Runtime`, `ClassLoader`, `.forName()`, `.exec()`, `eval()`, `ProcessBuilder`, `java.lang.reflect`) remains blocked.
* **UI Jest (`AlertsUtil.test.tsx`):** 108/108 pass. Added assertions for the new `mode="multiple"` strict-picks and the UUID-format term-filter branch.
* **E2E (Manual):** Verified end-to-end via a local webhook receiver. All 9 scenarios passed (entity creation → alert triggered → webhook received).

### Type of change

* [x] Bug fix
* [x] UX hardening (strict-pick filter inputs)

----
## Summary by Gitar

- **Test infrastructure:**
  - Added interaction logic to `observabilityAlert.ts` to focus comboboxes before filling inputs, ensuring `onSearch` triggers correctly in Playwright tests.
- **Refactored Data Insight evaluation:**
  - Moved numeric formula evaluation logic into a new dedicated `DataInsightFormulaEvaluator` utility.
  - Decoupled chart formula arithmetic processing from the main `ExpressionValidator` to maintain strict security boundaries.
- **Improved Security Validation:**
  - Added a regression test in `ExpressionValidatorTest` to ensure that arithmetic operators used in chart formulas are explicitly rejected by the policy/alert validator.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/utils/Alerts/AlertsUtil.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/Alerts/AlertsUtil.tsx
@@ -68,6 +68,7 @@ import {
   EXTERNAL_CATEGORY_OPTIONS,
 } from '../../constants/Alerts.constants';
 import { PAGE_SIZE_LARGE } from '../../constants/constants';
+import { UUID_REGEX } from '../../constants/regex.constants';
 import { OPEN_METADATA } from '../../constants/Services.constant';
 import { AlertRecentEventFilters } from '../../enums/Alerts.enum';
 import { SearchIndex } from '../../enums/search.enum';
@@ -103,9 +104,6 @@ import searchClassBase from '../SearchClassBase';
 import { getTermQuery } from '../SearchUtils';
 import { showErrorToast } from '../ToastUtils';
 import './alerts-util.less';
-
-const UUID_PATTERN =
-  /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/;
 
 export const getAlertsActionTypeIcon = (type?: SubscriptionType) => {
   switch (type) {
@@ -1064,7 +1062,7 @@ export const getFieldByArgumentType = (
     const searchIndexMapping =
       searchClassBase.getEntityTypeSearchIndexMapping();
     const trimmed = (searchText ?? '').trim();
-    const isUuidInput = UUID_PATTERN.test(trimmed);
+    const isUuidInput = UUID_REGEX.test(trimmed);
 
     try {
       const response = await searchQuery({
--- a/openmetadata-ui/src/main/resources/ui/src/utils/Alerts/AlertsUtil.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/Alerts/AlertsUtil.tsx
@@ -104,6 +104,9 @@ import { getTermQuery } from '../SearchUtils';
 import { showErrorToast } from '../ToastUtils';
 import './alerts-util.less';
 
+const UUID_PATTERN =
+  /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/;
+
 export const getAlertsActionTypeIcon = (type?: SubscriptionType) => {
   switch (type) {
     case SubscriptionType.Slack:
@@ -1056,6 +1059,49 @@ export const getFieldByArgumentType = (
       showDisplayNameAsLabel: false,
     });
   };
+
+  const getEntityByIdSuggestions = async (searchText?: string) => {
+    const searchIndexMapping =
+      searchClassBase.getEntityTypeSearchIndexMapping();
+    const trimmed = (searchText ?? '').trim();
+    const isUuidInput = UUID_PATTERN.test(trimmed);
+
+    try {
+      const response = await searchQuery({
+        query: trimmed,
+        pageNumber: 1,
+        pageSize: PAGE_SIZE_LARGE,
+        queryFilter: isUuidInput ? getTermQuery({ id: trimmed }) : undefined,
+        searchIndex: searchIndexMapping[selectedTrigger],
+      });
+
+      return uniqBy(
+        response.hits.hits.map((d) => {
+          const id = d._source.id ?? '';
+          const fqn = d._source.fullyQualifiedName ?? '';
+
+          return {
+            uuid: id,
+            value: id,
+            label: (
+              <div className="entity-id-option">
+                <div>{id}</div>
+                <div className="entity-id-option-fqn">{fqn}</div>
+              </div>
+            ),
+          };
+        }),
+        'value'
+      );
+    } catch (error) {
+      showErrorToast(
+        error as AxiosError,
+        t('server.entity-fetch-error', { entity: t('label.search') })
+      );
+
+      return [];
+    }
+  };
   const translatedContractStatusOptions = DATA_CONTRACT_STATUS_OPTIONS.map(
     (option) => ({
       ...option,
@@ -1071,12 +1117,11 @@ export const getFieldByArgumentType = (
           className="w-full"
           data-testid="fqn-list-select"
           maxTagTextLength={45}
-          mode="tags"
+          mode="multiple"
           optionFilterProp="label"
           placeholder={t('label.search-by-type', {
             type: t('label.fqn-uppercase'),
           })}
-          showArrow={false}
         />
       );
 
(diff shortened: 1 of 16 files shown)


manerow:
Thanks for the suggestion @gitar-bot. I considered `Set.copyOf(entityFqns)` but I'd argue keeping `List.contains` is the right call here. A few reasons:

**Allocation cost dominates at typical n.** `Set.copyOf` builds a `HashSet`-equivalent immutable map per evaluation: hash each element, allocate buckets, etc. For the typical alert-rule list size (n ≈ 1–5, since the UI typeahead picks one or a handful of entities), `List.contains`'s linear `String.equals` scan beats the Set's allocation + hashing on a single call. The crossover where Set wins tends to be n ≳ 10, which is well past real-world rule sizes.

**Consistency with the rest of the file.** Every other identifier-matcher in `AlertsRuleEvaluator` already uses `List.contains` or direct `equals` on a small list — `matchUpdatedBy` (L355), `matchAnyEntityId` (L184), `getTestCaseStatusIfInTestSuite` (L336), `matchAnyOwnerName` (L635, L642), and the direct-domain branch of `matchAnyDomain` (L491). Switching just `matchAnyEntityFqn` to a Set would make it the outlier and tempt someone to "round-trip" the rest, optimizing a path that isn't hot.

**Hot path is actually publication, not evaluation.** Each subscription evaluates each event once; the per-rule overhead is dwarfed by the I/O cost of the webhook delivery downstream. If a user ever stuffs hundreds of FQNs into a single rule, we can revisit, but at that point the right answer is probably to split the rule or use the entity ID list — not to micro-optimize the matcher.

So I'd prefer to keep `List.contains` and leave the optimization for the day someone actually reports a hot path here.

harshach (dismissed):

ayush-shah:

sonika-shah (dismissed):

openmetadata-ui/src/main/resources/ui/src/utils/Alerts/AlertsUtil.tsx
```diff
@@ -104,6 +104,9 @@ import { getTermQuery } from '../SearchUtils';
 import { showErrorToast } from '../ToastUtils';
 import './alerts-util.less';
 
+const UUID_PATTERN =
```
shah-harshit: Let's move this to appropriate constant file.

openmetadata-ui/src/main/resources/ui/src/utils/Alerts/alerts-util.less
```diff
@@ -46,3 +46,14 @@
...
+
+.entity-id-option {
+  display: flex;
+  flex-direction: column;
+  line-height: 1.3;
+}
+
+.entity-id-option-fqn {
+  color: @grey-3;
+  font-size: 12px;
```
shah-harshit: use variables wherever possible.

openmetadata-ui/src/main/resources/ui/src/utils/Alerts/AlertsUtil.tsx
manerow: Moved `UUID_PATTERN` to `constants/regex.constants.ts` as `UUID_REGEX` (matches the file's `*_REGEX` convention), imported and used at the single call site

openmetadata-ui/src/main/resources/ui/src/utils/Alerts/alerts-util.less
manerow: Changed `font-size: 12px;` → `font-size: @size-sm`

shah-harshit (approved):

sonika-shah (approved):

PR #27991, opened 2026-05-08
feat(ingestion): introduce TagRegistry domain layer

IceS2:
Fixes open-metadata/openmetadata-collate#3999

#### Why
The legacy `context.tags` list overloads two roles (sink-bound queue + per-entity inheritance lookup) and stores one Pydantic `TagLabel` per attachment. On a Snowflake schema with ~120k tag attachments this peaks at ~112 MB — disproportionate to the ~20 unique tags actually in play.

#### What
Adds `metadata.domain.tags` with two classes:
- `TagRegistry` — per-Source bookkeeping for tag/classification ingestion
- `TagCanonicalizer` — case-corrected name resolution against system-provider entities in OM

Migrates the Snowflake connector to use them. Other DB connectors continue on the legacy `context.tags` flow (strangler pattern; future PRs migrate per-connector).

#### How
- `TagRegistry` interns shared `TagLabel` instances by `(classification, tag, labelType, state)` — N attachments dereference to the same Pydantic graph.
- `clear_scope` rebinds `_labels_by_entity` to a fresh dict (not pop-in-place) so the hash-table bucket array releases immediately.

### Type of change:
- [x] Improvement

#### Architecture
`metadata.domain.tags` is a new domain layer (in-memory helpers operating on OM-generated types, no I/O ownership beyond an injected `OpenMetadata` client for read-only queries). Two classes by separated concerns:

- **`TagRegistry`** — three storage groups behind two locks. `_pending` queues sink-bound `OMetaTagAndClassification` payloads (deduped by `_known_tag_fqns`). `_labels_by_entity` maps entity FQN → list of shared `TagLabel` references. `_tag_label_cache` interns the underlying `TagLabel` objects.
- **`TagCanonicalizer`** — separate concern: case-corrected lookup against system-provider classifications/tags via ES with retry. Failures raise; callers wrap in `Either`.

`DatabaseServiceSource` exposes both as `cached_property` and wires `clear_schema_tag_scope` / `clear_database_tag_scope` topology hooks.

#### Files added
- `ingestion/src/metadata/domain/__init__.py`
- `ingestion/src/metadata/domain/tags/__init__.py`
- `ingestion/src/metadata/domain/tags/registry.py`
- `ingestion/src/metadata/domain/tags/canonicalizer.py`
- `ingestion/tests/unit/domain/tags/test_registry.py` (34 tests)
- `ingestion/tests/unit/domain/tags/test_canonicalizer.py` (13 tests)

#### Files updated
- `ingestion/src/metadata/ingestion/source/database/database_service.py` — registry/canonicalizer wiring + topology hooks
- `ingestion/src/metadata/ingestion/source/database/snowflake/metadata.py` — migrated \`yield_tag\` and \`get_*_tag_labels\` overrides
- `ingestion/setup.py` — added `tenacity>=8.0,<10` for canonicalizer retries

#### Unit tests
- 34 tests in `test_registry.py` (attach/labels_for/drain/clear_scope/ensure_known/stats, thread safety, **interning identity** by \`is\`-equality, **cache survives \`clear_scope\`**, **fqn-level dedup across \`label_type\` variants**)
- 13 tests in `test_canonicalizer.py` (canonicalization, retry behavior, no cache poisoning on persistent failure)

#### Manual testing performed
1. Local Collate via \`docker compose -f docker/development/docker-compose.yml up -d\`
2. Snowflake ingestion against a perf fixture with ~120k tag attachments on a single schema → peak heap **~112 MB → ~24 MB** (~4.7× reduction)
3. Multi-DB ingestion (5 perf databases × 39 schemas with tags) → per-scope peak ≤ 0.24 MB, every \`clear_scope\` releases bucket memory
4. Behavioral verification of inheritance + dedup on \`OM_PERF_MAIN.NORMAL_011.T_0001\` (schema + table + column tag levels)

Not applicable.

### Checklist:
- [x] My PR title is descriptive
- [x] My PR is linked to a GitHub issue

----
## Summary by Gitar

- **Refactored TagRegistry contracts:**
  - Updated `TagRegistry.attach` and `_build_pending_record` to require explicit description strings instead of optional values.
- **Enhanced TagCanonicalizer interface:**
  - Changed `classification` and `tag` methods to use explicit `default_description` arguments for seeding metadata.
  - Enforced non-optional descriptions in `Canonical` named tuple to ensure data consistency.
- **Updated Snowflake connector:**
  - Refactored `yield_tag` and `yield_database_tag` calls to provide required default descriptions for classification and tag lookups.
- **Updated unit tests:**
  - Adjusted `test_canonicalizer` and `test_snowflake` tests to match the new strict parameter signatures.


ingestion/src/metadata/domain/tags/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Tag and Classification domain utilities."""
+
+from metadata.domain.tags.canonicalizer import Canonical, TagCanonicalizer
+from metadata.domain.tags.registry import ScopeAlreadyClearedError, TagRegistry
+
+__all__ = [
```
pmbrull, line 16: nit - i dont like this because then people like to *

ingestion/src/metadata/domain/tags/canonicalizer.py
```diff
@@ -0,0 +1,136 @@
...
+from metadata.generated.schema.entity.classification.tag import Tag
+from metadata.generated.schema.type.basic import ProviderType
+from metadata.ingestion.ometa.ometa_api import OpenMetadata
+from metadata.utils import fqn
+from metadata.utils.logger import ingestion_logger
+
+logger = ingestion_logger()
+
+
+_es_retry = retry(
```
pmbrull, line 41: thinking this could be more commonly reused in other places other than the canolicalaizer, e.g., when we search for tables, or owners

ingestion/src/metadata/domain/tags/canonicalizer.py
```diff
@@ -0,0 +1,136 @@
...
+                    description=entity.description.root if entity.description else tag_description,
+                )
+                break
+
+        with self._lock:
+            self._tag_cache.setdefault(key, canonical)
+        return canonical
+
+    @_es_retry
+    def _es_search(self, entity_type: Any, search_string: str) -> Iterable[Any]:
```
pmbrull, line 143: this could be inside ometa with the decorator and all

ingestion/src/metadata/ingestion/source/database/snowflake/metadata.py
```diff
@@ -548,71 +547,112 @@ def yield_tag(self, schema_name: str) -> Iterable[Either[OMetaTagAndClassificati
...
+                        )
+                    except Exception as exc:
+                        logger.debug(traceback.format_exc())
+                        yield Either(
+                            left=StackTraceError(
+                                name=f"{tag_info['tag_name']}.{tag_info['tag_value']}",
+                                error=f"Tag canonicalization failed for {tag_info['tag_name']}.{tag_info['tag_value']}: {exc}",
+                            )
+                        )
+            yield from (Either(right=record) for record in self.tags_registry.drain())
```
pmbrull: i dont follow why the drain() here helps vs the vanilla yield we had

ingestion/src/metadata/ingestion/source/database/database_service.py
```diff
@@ -811,6 +825,33 @@ def yield_life_cycle_data(self, _) -> Iterable[Either[OMetaLifeCycleData]]:
...
+        if schema_name:
+            schema_fqn = fqn.build(
+                self.metadata,
+                entity_type=DatabaseSchema,
+                service_name=self.context.get().database_service,
+                database_name=self.context.get().database,
+                schema_name=schema_name,
+            )
+            self.tags_registry.clear_scope(schema_fqn)
+        yield from ()
```
pmbrull, line 852: silly q, why this yield?

IceS2: Due to how the post process actually works in the Topology Runner:
`yield from self._run_node_post_process(node=node)`

ingestion/src/metadata/ingestion/source/database/snowflake/metadata.py
IceS2: It is not about the drain itself.
The gains are on how we populate and what we yield.

Before we were yielding the following:
```python
                for tag_info in self.schema_tags_map[schema_name]:
                    yield from get_ometa_tag_and_classification(
                        tag_fqn=FullyQualifiedEntityName(schema_fqn),
                        tags=[tag_info["tag_value"]],
                        classification_name=tag_info["tag_name"],
                        tag_description=SNOWFLAKE_TAG_DESCRIPTION,
                        classification_description=SNOWFLAKE_CLASSIFICATION_DESCRIPTION,
                        metadata=self.metadata,
                        system_tags=True,
                    )
```

This was yielding multiple times the same tag and classification to be ingested since we are saving basically Tag References.
If Tag A was used one time at a schema, one time at a database and one time at a column, we would send 3 puts on that Tag and Classification.

Drain comes from pending tags that have not yet been yielded to the sink, based on this logic that should avoid us sending duplicates (just classification ones due to the current coupling between both, but I did not want to delve there due to the urgency)

```python

        with self._run_state_lock:
            tag_fqn = model_str(tag_label.tagFQN)
            if tag_fqn not in self._known_tag_fqns:
                self._known_tag_fqns.add(tag_fqn)
                self._pending.append(
                    self._build_pending_record(
                        classification_name=classification_name,
                        classification_description=classification_description,
                        tag_name=tag_name,
                        tag_description=tag_description,
                    )
                )
```

This is geared towards time improvement, memory improvement is handled within the TagRegistry as well, avoiding creating a lot of repeated pydantic TagLabel objects by using references to them instead.

All this while centralizing the Tag logic in one place with a clear api.

ingestion/src/metadata/domain/tags/canonicalizer.py
IceS2: I agree, but did not want to add it there for now. If this works properly, I will need to create another PR to migrate the rest of the databases to use this and the other service types, while also doing some cleaning. For sure this should be moved there.

ingestion/src/metadata/domain/tags/canonicalizer.py
IceS2: I agree, but did not want to add it there for now. If this works properly, I will need to create another PR to migrate the rest of the databases to use this and the other service types, while also doing some cleaning. For sure this should be moved there.

ingestion/src/metadata/domain/tags/__init__.py
IceS2: This is way handy that having too deep imports imho, we have static checks to prevent import *

pmbrull (approved):

PR #28015, opened 2026-05-10
Perceived-latency P2 + P3.5 servlet/HTTP2: hydrate, SWR cache, skeletons, Vary, h2c/h2 opt-in

harshach:
### Describe your changes:

Phase 2 of the perceived-latency work tracked in `.context/perceived-latency-design.md`, plus a small P3.5 servlet correctness fix and an opt-in HTTP/2 path at Jetty.

**P2.3 — Batch lineage hydration endpoint (`POST /v1/lineage/hydrate`)**
Server primitive: accepts a list of `(type, id)` pairs and returns hydrated entities grouped by `entityType`. Replaces N per-node entity GETs with one round-trip. Each entity is authorized individually with `VIEW_BASIC`; entities the caller cannot read are silently dropped from the response (rather than failing the whole batch). Useful for any client that needs to hydrate a graph of mixed entity types.

The original design hypothesized 50× per-node entity GETs in the Lineage view. On inspection the actual per-node fetches in the current Lineage code are for `getTestCaseExecutionSummary` (a different endpoint), not entity GETs, so this PR ships only the server-side primitive and the matching `lineageAPI.ts` helper.

**P2.2 — Explore tab-switch SWR cache**
Tab-switch on `/explore` re-runs the same shape of search-fetch with a different `searchIndex`. This adds a 30s stale-while-revalidate cache keyed by the same dependency string the page already uses to detect "should I refetch?". Cache hit renders synchronously with no spinner, then revalidates silently in the background. Capped at 60 entries.

**P2.1 — Skeleton screens (TableDetail, MyData, Lineage)**
Replace top-level `<Loader />` gates with skeleton placeholders that mirror the page chrome. Cheap implementation: pure antd `Skeleton` primitives.

**P3.5 — `OpenMetadataAssetServlet` correctness fixes**
Two real bugs surfaced during the transport-layer audit:
1. **Missing `Vary: Accept-Encoding`** on compressed responses → CDN cache poisoning risk.
2. **Broken q-value parsing** — `enc.contains("q=0")` matched `q=0.5` etc. Fixed by parsing as a double per RFC 7231 §5.3.4. Also tightened coding-name match (so `brand` no longer matches `br`) and added wildcard handling.

Three new unit tests cover the fixes; the existing 9 tests continue to pass.

**P3.5 — Opt-in HTTP/2 at Jetty (`type: h2` / `type: h2c`)**
Adds `dropwizard-http2` to openmetadata-service so `type: h2` (TLS) and `type: h2c` (cleartext) connector factories are available. Operators opt in by uncommenting the example blocks in `conf/openmetadata.yaml`. Both connector types compose HTTP/1.1 + HTTP/2 protocol handlers on the same port (HTTP/1.1 clients still work via Upgrade / ALPN), so adopting them is backward-compatible.

**Verified backward compatibility** by temporarily switching the IT bootstrap connector to `type: h2c` and running:
- `SystemResourceIT`  → 34/34 pass
- `LineageHydrateIT`  →  5/5 pass

39/39 IT tests passed against the h2c connector — proves the OkHttp3 SDK works transparently via HTTP/1.1 Upgrade. The shipped IT YAML is unchanged (still `type: http`); only operators who explicitly opt in get Jetty-side HTTP/2.

**Why HTTP/2 at Jetty matters even with an LB:** end-to-end multiplexing on the LB↔pod hop, plus self-hosted deployments that have no LB at all (Jetty is the edge).

**Playwright with h2/h2c — out of scope for this PR:** browsers refuse h2c (cleartext HTTP/2) by design. `type: h2` (TLS) would work for Playwright but requires keystore + cert-trust provisioning per environment, which is operator infrastructure unrelated to this code change. The default Playwright config still runs against `type: http` (HTTP/1.1) and is unaffected by anything in this PR. The full HTTP/2-at-the-edge story (browsers → LB) is covered by the runbook in `.context/http2-brotli-runbook.md` (nginx-ingress / ALB / Traefik annotations).

Companion to PR #28014 (P1: ETag/304, deferred tab fetches, lazy widgets), which is still under review. The two PRs are independent.

#
### Type of change:
- [x] Improvement
- [x] Bug fix (servlet `Vary` header + q-value parsing)

#
### Frontend Preview (Loom)
Not applicable for this PR — perceived-latency improvements are best evaluated in a running stack with realistic data.

#
### Checklist:

- [x] My PR title is in the form of `<type>: <title>` and follows [Conventional Commits Specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] My commits follow the [Conventional Commits Specification](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [x] I have added tests that prove my fix is effective or that my feature works (`LineageHydrateIT` for the new endpoint, three new `OpenMetadataAssetServletTest` cases for the servlet fix).
- [x] New and existing integration tests pass locally with my changes (`mvn verify -pl openmetadata-integration-tests -P postgres-opensearch -Dit.test=SystemResourceIT,LineageHydrateIT` 39/39 green; backward-compat verified against `type: h2c` connector).
- [x] UI checkstyle passes locally on touched files.
- [x] Java spotless passes locally.
- [ ] I have updated the [API documentation](https://github.com/open-metadata/OpenMetadata/tree/main/openmetadata-docs)

----
## Summary by Gitar

- **Servlet/HTTP2 hardening:**
  - Improved `appendVaryHeader` in `OpenMetadataAssetServlet` to correctly parse and deduplicate `Vary` tokens.
  - Made `chromium-h2` Playwright test suite registration conditional to avoid discovery overhead when not in `h2` mode.
- **Documentation/Tests:**
  - Updated `LineageHydrateIT` class documentation to accurately reflect the internal usage of `hydrateAndAuthorize` for entity permission checks.


conf/openmetadata-h2-test.yaml
```diff
@@ -0,0 +1,857 @@
...
+
+basePath: ${BASE_PATH:-/}
+
+server:
+  applicationContextPath: ${BASE_PATH:-/}
+  rootPath: /api/*
+  applicationConnectors:
+    # HTTP/2 over TLS (h2) connector — negotiated via ALPN. HTTP/1.1 clients are still
+    # accepted on the same port via TLS handshake fallback, so this is backward-compatible.
+    - type: h2
```
mohityadav766: can it be a env with default as we have today in existing yaml?

mohityadav766 (dismissed):

PR #28036, opened 2026-05-11
fix: resolve lineage extraction bug for MySQL 5.7.3

### LINKED ISSUE #28029: MySQL 5.7+: general_log.argument field type changed from MEDIUMTEXT to MEDIUMBLOB, causing data lineage SQL query failure
### Affected module

Lineage

### Description
In MySQL 5.7 and above, the `mysql.general_log.argument` field type was changed from `MEDIUMTEXT` (MySQL 5.6 and earlier) to `MEDIUMBLOB`. This breaks OpenMetadata's data lineage SQL extraction, which relies on querying this field as text. When the field is `MEDIUMBLOB`, lineage analysis fails to parse SQL statements and returns empty or incorrect results.

### To Reproduce

1. Configure MySQL 5.7+ with `log_output = 'TABLE'` and `general_log = 'ON'`.
2. Run any SQL statement, e.g., `SELECT * FROM information_schema.tables LIMIT 5;`.
3. In OpenMetadata, set up a MySQL ingestion workflow that includes lineage extraction using `mysql.general_log`.
4. Observe that the lineage extraction returns no edges or incomplete lineage.
5. Optionally, query the `general_log` directly in MySQL to confirm the issue:
   ```sql
   SHOW CREATE TABLE mysql.general_log;  -- Shows argument as mediumblob
   SELECT argument FROM mysql.general_log LIMIT 1;  -- Shows binary blob (0x...)

### Expected behavior


---

### 5. Expected behavior（预期行为）⭐ 单独填写

**建议填写**：

```markdown
In MySQL 5.6 and earlier, `argument` is `MEDIUMTEXT`, and OpenMetadata's lineage SQL parser can read it as a normal string, successfully extracting table/column dependencies. The expected behavior is that lineage extraction works on MySQL 5.7+ as well, either by automatically converting the BLOB to TEXT or by handling the field type difference internally.

### OS

windows11

### Python version

_No response_

### OpenMetadata version

1.12.6

1.12.6

### Additional context

### Additional Context
- MySQL made this change intentionally in version 5.7.3 to avoid character set transformations (release note: [MySQL Bugs #71018](https://bugs.mysql.com/bug.php?id=71018))
- A workaround for end users is to manually convert the BLOB in SQL queries: `CONVERT(argument USING utf8mb4)`, but OpenMetadata's internal lineage extraction should ideally handle this automatically.
- This affects any user who uses MySQL 5.7+ as a source for lineage extraction via the general query log table.

### Pre-submission checklist

### PR DESCRIPTION
**Pull request by cchenan**
## Problem
Lineage extraction fails when connecting to MySQL 5.7.3. The queries in `queries.py` use syntax or columns that are not available in this version.

## Solution
Modified `ingestion/src/metadata/ingestion/source/database/mysql/queries.py` to be compatible with MySQL 5.7.3.

Changes:
- Adjusted information_schema queries for MySQL 5.7.3 compatibility
- Added conditional logic where needed

## Testing
- Verified with MySQL 5.7.3 container: lineage extraction works correctly
- Verified with MySQL 5.7.23 and 8.0: no regression

## Related issue
Closes #28029

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/mysql/queries.py
+++ b/ingestion/src/metadata/ingestion/source/database/mysql/queries.py
@@ -14,7 +14,8 @@ SQL Queries used during ingestion
 
 import textwrap
 
-# general_log.argument is MEDIUMTEXT on older MySQL and MEDIUMBLOB on 5.7+; CONVERT unifies behavior for SELECT/WHERE.
+# general_log.argument and slow_log.sql_text are MEDIUMTEXT on older MySQL and MEDIUMBLOB on 5.7+;
+# CONVERT(... USING utf8mb4) unifies behavior for SELECT/WHERE.
 MYSQL_SQL_STATEMENT = textwrap.dedent(
     """
 SELECT 
@@ -43,7 +44,7 @@ MYSQL_SQL_STATEMENT_SLOW_LOGS = textwrap.dedent(
     """
 SELECT 
 	NULL `database_name`,
-	sql_text `query_text`,
+	CONVERT(sql_text USING utf8mb4) `query_text`,
 	start_time `start_time`,
     NULL `end_time`,
 	NULL  `duration`,
@@ -53,8 +54,8 @@ SELECT
     NULL `aborted`
 FROM mysql.slow_log
 WHERE start_time between '{start_time}' and '{end_time}'
-    AND sql_text NOT LIKE '/* {{"app": "OpenMetadata", %%}} */%%'
-    AND sql_text NOT LIKE '/* {{"app": "dbt", %%}} */%%'
+    AND CONVERT(sql_text USING utf8mb4) NOT LIKE '/* {{"app": "OpenMetadata", %%}} */%%'
+    AND CONVERT(sql_text USING utf8mb4) NOT LIKE '/* {{"app": "dbt", %%}} */%%'
     {filters}
 ORDER BY start_time desc
 LIMIT {result_limit};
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cchenan** [comment]
@open-metadata/ingestion Could you please add the safe to test label to this PR? Thanks!

File: ingestion/src/metadata/ingestion/source/database/mysql/queries.py [hunk 1]
```diff
@@ -14,11 +14,13 @@
...
 import textwrap
 
+# general_log.argument and slow_log.sql_text are MEDIUMTEXT on older MySQL and MEDIUMBLOB on 5.7+;
+# CONVERT(... USING utf8mb4) unifies behavior for SELECT/WHERE.
 MYSQL_SQL_STATEMENT = textwrap.dedent(
     """
 SELECT 
 	NULL `database_name`,
-	argument `query_text`,
+	CONVERT(argument USING utf8mb4) `query_text`,
```
harshach, line 23: will this work with only msyql 5.7 or any version after that too?

cchenan: > will this work with only msyql 5.7 or any version after that too?

It works for both MySQL 5.7.3+ and 8.0. I've tested with 5.7.23 and 8.0, no regression. The CONVERT function is supported in all MySQL 5.7+ and 8.0 versions.

cchenan: The static check errors reported (e.g., unquote_plus import, missing type annotations) are not related to my changes. They exist in the current main branch. Could you please advise how to proceed? Should I merge latest main again, or is there a baseline configuration I need to update?

cchenan: Thanks for testing! I see the deployment hit some failures. Is there anything on my side that I can help fix? Or is

PR #28057, opened 2026-05-12
fix(tags): propagate appliedBy on derived TagLabels so PATCH on inherited tags no longer 500s

### LINKED ISSUE #28038: Error while updating GlossaryTerms – Non-existing name/value pair in the object for key appliedBy
### Affected module

UI

### Describe the bug

While updating tags on a table entity using the JSON Patch API, the request fails with a 500 Internal Server Error when attempting to remove appliedBy from tags where the field does not exist.

This issue is reproducible in 1.12.6.

```
{
  "code": 500,
  "errorType": "UNHANDLED_SERVER_EXCEPTION",
  "message": "An exception with message [Non-existing name/value pair in the object for key appliedBy] was thrown while processing request."
}
```

### To Reproduce

1. Create or use a GlossaryTerm with inherited tags.
2. Update tag on a related table entity.
3. Update tag again on the same table entity.
4. Observe the API failure with 500 Internal Server Error.

https://github.com/user-attachments/assets/0bdc8a42-6275-47d4-8456-0b4756fdf474

### Expected behavior

- Patch requests should succeed even if appliedBy is not present in the tag object.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

1.12.6

_No response_

### Additional context

_No response_

### Pre-submission checklist

siddhant1:
## Summary

Fixes #28038 — PATCH on a tag-bearing entity (table, glossaryTerm, …) returns HTTP 500 with `Non-existing name/value pair in the object for key appliedBy` when a derived/inherited tag is reshuffled in a second consecutive edit.

The bug came from two interacting halves of #24817 *"Tagging explanation"*:

1. **`EntityRepository.updateTags`** mutates `appliedBy` *in memory* on tags it just touched (`tag.withAppliedBy(updatingUser.getName())`), so the PATCH response always carries the key for direct tags. The UI stores that response in local state and never re-fetches between edits.
2. **`CollectionDAO.getDerivedTagsBatch`** propagated `appliedAt` from the source `tag_usage` row but silently **dropped `appliedBy`** when reconstructing derived `TagLabel`s. Combined with `@JsonInclude(NON_NULL)` on the generated `TagLabel`, derived tags serialize **without** the key.

On a second edit, `fast-json-patch compare(local, updated)` emits `remove /tags/N/appliedBy` for any index whose direct tag (stamped) is replaced by a tag whose updated version lacks the key. The server's PATCH-load reconstructs the same entity's derived tags via `getDerivedTagsBatch` without `appliedBy`, so Jakarta JSON Patch throws.

Captured curl repro from the issue: `remove /tags/0/appliedBy` is the first op touching `/tags/0` and is the op that 500s.

## Changes

- **`CollectionDAO.getDerivedTagsBatch`** — add `.withAppliedBy(usage.toTagLabel().getAppliedBy())` next to the existing `.withAppliedAt(...)`. Restores parity with `TagLabelMapper` and `TagLabelRowMapperWithTargetFqnHash`, which already propagate the field.
- **`JsonUtils.isReadOnlyPatchPath`** — also filter ops whose `path` ends with `/appliedBy` or `/appliedAt`. These are server-managed audit fields; clients shouldn't be able to mutate them via PATCH, and ignoring such ops is defence-in-depth for legacy `tag_usage` rows where `applied_by` is still `NULL` (the 1.11.5 migration added the column with a default but no backfill).

## Tests

- **`JsonUtilsTest.applyPatchFiltersTagAuditFieldOps`** — asserts `applyPatch` silently drops `remove` ops on `/tags/N/appliedBy` and `/tags/N/appliedAt` even when the target tag has no such key.
- **`TagAppliedByPatch.spec.ts`** — Playwright regression for #28038. Sets up the exact fixture from the issue (glossary term with an inherited classification tag, attached to a table), then does a second PATCH whose `fast-json-patch` diff contains `remove /tags/N/appliedBy`. Asserts the request succeeds and every tag (direct + derived) in both responses carries `appliedBy`.

## Test plan

- [x] `JsonUtilsTest#applyPatchFiltersTagAuditFieldOps` + existing `applyPatch` test pass locally (verified on the 1.12.6 backport of these changes since `origin/main` is currently broken at HEAD by an unrelated `RuleEngine.getInstance()` compile error in `EntityRepository.java`).
- [x] `TagAppliedByPatch.spec.ts` runs green against a live local stack (1.3s).
- [x] `mvn spotless:apply -pl openmetadata-spec,openmetadata-service` clean.
- [ ] CI on this PR.

## Backport

A matching change has been prepared for `1.12.6`; will be cherry-picked once this lands.

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
@@ -6459,7 +6459,8 @@ public interface CollectionDAO {
                 .withLabelType(TagLabel.LabelType.DERIVED)
                 .withState(TagLabel.State.values()[usage.getState()])
                 .withReason(usage.getReason())
-                .withAppliedAt(usage.toTagLabel().getAppliedAt());
+                .withAppliedAt(usage.toTagLabel().getAppliedAt())
+                .withAppliedBy(usage.toTagLabel().getAppliedBy());
         if (usage.getReason() != null) {
           tagLabel.withReason(usage.getReason());
         }
--- a/openmetadata-service/src/test/java/org/openmetadata/service/util/JsonUtilsTest.java
+++ b/openmetadata-service/src/test/java/org/openmetadata/service/util/JsonUtilsTest.java
@@ -118,13 +118,6 @@ class JsonUtilsTest {
     assertTrue(jsonException.getMessage().contains("An array item index is out of range"));
   }
 
-  /**
-   * Patch ops targeting the server-managed tag audit fields (appliedBy/appliedAt) are filtered
-   * out, mirroring the existing filter for href/changeDescription. Without the filter, a
-   * `remove /tags/N/appliedBy` op against a tag whose JSON has no appliedBy key (e.g. a derived
-   * tag, or a legacy row with applied_by = NULL serialized via @JsonInclude(NON_NULL)) throws
-   * "Non-existing name/value pair in the object for key appliedBy" — see issue #28038.
-   */
   @Test
   void applyPatchFiltersTagAuditFieldOps() {
     JsonObjectBuilder tag =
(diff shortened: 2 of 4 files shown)


ayush-shah:

PR #28084, opened 2026-05-13
fix(dq-dashboard): denormalize certification onto testCase index + filter plumbing

manerow:
Fixes open-metadata/openmetadata-collate#4080

## What was wrong
On the Data Observability dashboard, filtering by a Certification (Bronze/Silver/Gold) returned zero coverage even when certified assets existed. The same bug surfaces when navigating to a Certification tag's detail page — the Data Observability tab shows empty charts.

There are two reasons this happens, both fixed in this PR's foundation layer:
1. The test-case search index didn't carry the parent table's certification, so any filter by certification on test-case data had nothing to match.
2. The dashboard's filter-building code had no way to express a "certification" filter in the first place — only owners, tier, tags, glossary terms, and data products.

## What this PR changes
- Teaches the test-case indexer to copy each parent table's `certification` onto its testCase docs, the same way it already copies `domains`. Mapping updated for all four supported languages.
- Adds a `certification` filter slot to `DataQualityDashboardChartFilters` and a matching branch in `buildDataQualityDashboardFilters` so callers can ask the dashboard to filter by certification.

This PR is plumbing only — it doesn't change anything you'd see in the UI. The visible UI piece (Certification dropdown, Tag-list exclusion, Tag detail page wiring) lives in #28085.

## Why split from the UI fix
The dashboard component lives in a different repo on 1.12.x (Collate) than on 1.13+ (OM). Splitting the backend + filter plumbing here lets this PR cherry-pick cleanly to `1.12.9`, `1.13`, and `main`. The visible UI then lands as #28085 on main/1.13 and via a separate Collate PR on 1.12.x.

## Test plan
- [x] After deploy, a test case on a Gold-certified table has `certification.tagLabel.tagFQN: Certification.Gold` denormalized onto its testCase index doc.
- [x] `/api/v1/dataQuality/testSuites/dataQualityReport?index=testCase` with a `certification.tagLabel.tagFQN` term filter returns the expected test cases.
- [x] No regression on the existing tier / tag / owner filter paths.

## Follow-ups
- The user-facing dashboard UI: #28085.
- The Collate-side dashboard UI for 1.12.x: open-metadata/openmetadata-collate#4079.
- A separate latent routing bug surfaced for Tier (silently fails on several testCase-driven widgets that bypass the centralized filter builder) — tracked in open-metadata/openmetadata-collate#4086 and staged on branch \`fix/dq-tier-routing\`.

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/search/SearchClient.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/search/SearchClient.java
@@ -179,6 +179,18 @@ public interface SearchClient
       }
       """;
 
+  // Cascade variant: full-object replace (handles add/update) plus removal on
+  // null params, so child docs stay in sync when a parent's cert is added,
+  // changed, or removed.
+  String CASCADE_CERTIFICATION_SCRIPT =
+      """
+      if (params.certification == null) {
+        ctx._source.remove('certification');
+      } else {
+        ctx._source.certification = params.certification;
+      }
+      """;
+
   String UPDATE_GLOSSARY_TERM_TAG_FQN_BY_PREFIX_SCRIPT =
       """
       if (ctx._source.containsKey('tags')) {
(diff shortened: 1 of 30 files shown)


mohityadav766 (dismissed):

shah-harshit (dismissed):

manerow:
@gitar-bot both findings dismissed:

1. The tier-on-testCase routing bug is pre-existing and explicitly deferred to issue open-metadata/openmetadata-collate#4086 / branch `fix/dq-tier-routing` (see the "Follow-ups" section of the PR description). This PR doesn't claim to fix it — only the certification path is in scope.

2. The `entity.getEntityReference().getType()` dereference in `cascadeCertificationToChildren` is consistent with `updateEntityCertificationInSearch` immediately above, which does the same call without a guard. The entity is guaranteed populated by the upstream `handleEntityCertificationUpdate` caller; adding a guard here would create misleading asymmetry with the sibling method.

mohityadav766 (approved):

ShaileshParmar11 (approved):

shah-harshit (approved):

Issue #19901, opened 2025-02-20
Add Support for Setting a Default UI Language


thiagodeschamps:
**Is your feature request related to a problem? Please describe.**  
I’m finding it frustrating that there is currently no way to set a default UI language at the system level in OpenMetadata. Although individual users can change their language preferences, this is cumbersome for organizations operating in non-English environments. Every new or existing user must configure their language preferences independently, which creates inconsistencies and extra administrative overhead.

**Describe the solution you’d like**  
A global configuration option that allows administrators to define a default language for all users. This would mean:
- Administrators can specify the default language in a configuration file, UI, environment variable, or database setting.
- Newly created users (or those who haven’t specified a language) automatically see the UI in the default language.
- Users would still have the option to override the default language if they need to.

**Describe alternatives you’ve considered**  
1. **Manual Setup for Each User**  
   - Currently, each user must manually change their UI language, which is not scalable for larger teams or organizations.  
2. **Custom Scripts/Workarounds**  
   - Attempting to automate language settings outside the official configuration leads to potential maintenance headaches and may break with future updates.

**Additional context**  
- This feature would streamline deployments in non-English locales and ensure a consistent, localized experience for all users.  
- A screenshot of the Slack conversation discussing this request is attached for reference (see above image).

jgrocha:
Hi @thiagodeschamps,

Did you test this on recent OpenMetadata deploy? Please consider the PR #19418, landing on OM about a month ago.

1) **The default language defined by the user in the browser is respected**. No need to setup individual users. They will have their default language.

2) As admin, you can setup a redirect using a query parameter to something like: https://openmetadata.corsega.fr?lng=fr-FR All users will start with the fr-FR language, for example. 

3) As admin, you can also add/force the i18next cookie. It depends on the HTTP server you are using.

This mechanism provides a localized experience for all non-English users (if the language is already supported).

Is this ok for you?

chirag-madlani:
@thiagodeschamps closing this with consideration of the above workaround, thanks

PR #28132, opened 2026-05-15
Fix #28076: search_after pagination breaks when sort values contain ','

### LINKED ISSUE #28076: Automation fails on glossary terms with commas in names due to search_after pagination bug
## Problem
An automation workflow to update glossary term owners is failing to process all terms. Out of 1863 glossary terms, only 311 are getting updated. The automation fails with an Elasticsearch error during pagination.

## Root Cause
The issue occurs in `SearchUtils.searchAfter` method where `split(",")` is used to parse the search_after cursor. When glossary terms contain commas in their fully qualified names, the split logic breaks pagination.

## Error Details
```
metadata.ingestion.ometa.client.APIError: Search failed due to Request failed: [search_phase_execution_exception] all shards failed | Root cause: [illegal_argument_exception: search_after has 2 value(s) but sort has 1.]
```

The error shows that search_after has 2 values but sort has only 1, indicating the comma in the term name is being incorrectly parsed as a delimiter.

## Expected Behavior
Automation should process all glossary terms regardless of special characters (commas, parentheses, etc.) in their names.

## Actual Behavior
Automation stops processing when encountering glossary terms with commas in their fully qualified names, resulting in incomplete updates.

## Affected Component
- `SearchUtils.searchAfter` method
- Glossary term search pagination logic

## Steps to Reproduce
1. Create glossary terms with commas in their names
2. Run an automation workflow filtering on those glossary terms
3. Observe pagination failure during search_after operation

sonika-shah:
## Problem

Automations and any ingestion path that uses `paginate_es` truncate silently when sorted entities have a `,` in the sort value. The customer-reported case (Pylon #19901, issue #28076) processed 311 of 1,863 glossary terms because the cursor at one page boundary was an FQN containing a comma:

```
x alation archive.system glossary.sfmc (salesforce marketing cloud, exacttarget) (it system)
```

The server logs the underlying ES error:

```
illegal_argument_exception: search_after has 2 value(s) but sort has 1
```

## Root cause

Both sides of the cursor protocol assume `,` is a safe delimiter:

- Ingestion (`es_mixin.py`): `after = ",".join(last_hit.sort)`
- Server (`SearchUtils.searchAfter`): `List.of(searchAfter.split(","))`

FQNs are allowed to contain `,` per the entity-name schema (`pattern: "^((?!::).)*$"`). When a value contains a comma, the join+split round-trip produces N+1 values for an N-field sort, and ES rejects the next-page request. The wire format is fundamentally ambiguous — given `a,b,c` and 2 sort fields, the values could be `("a","b,c")`, `("a,b","c")`, or three values, with no way to disambiguate from the cursor alone.

## Fix

Switch the cursor to a **base64-encoded JSON array** — the same pattern already used in the codebase by `OpenSearchColumnAggregator.decodeCursor` / `ElasticSearchColumnAggregator.decodeCursor` and documented in the public `columnGridResponse.cursor` schema. Backward compatibility is preserved: the server tries base64-JSON first and falls back to the legacy `split(",")` when the value isn't a base64-JSON cursor, so older ingestion clients (1.6 → 1.12) keep paginating against an upgraded server.

UI is unaffected — the Explore page uses `from`+`size` offset pagination and never constructs `?search_after=`.

## Changes

| File | Change |
|---|---|
| `openmetadata-service/.../search/SearchUtils.java` | `searchAfter()` tries base64-JSON then falls through to legacy split |
| `openmetadata-service/.../resources/search/SearchResource.java` | `@Parameter` description for `search_after` documents both formats |
| `ingestion/.../ometa/mixins/es_mixin.py` | `_paginate_es_internal` emits a base64-JSON cursor |
| `openmetadata-service/.../test/.../SearchUtilsTest.java` | 5 new test methods covering the new format, legacy fallback, and the customer's exact comma-FQN |
| `ingestion/tests/integration/ometa/test_ometa_es_api.py` | `test_paginate_with_comma_in_name` — creates 5 tables with `,` in the name, paginates with `size=1`, asserts all 5 are returned |

## Verification

- `mvn -pl openmetadata-service test -Dtest=SearchUtilsTest` — 109 tests pass (5 new).
- End-to-end round-trip validated: Python encodes → Java decodes → original FQN with comma preserved verbatim.
- `OpenSearchSearchManagerIntegrationTest` / `ElasticSearchSearchManagerIntegrationTest` exercise `search_after` against a real ES/OS container and continue to pass — they validate the legacy fallback path.

Closes #28076.


ayush-shah:

sonika-shah:
Superseded by #28291 (same diff, re-opened from a branch on the main repo because the fork-authored PR was causing CI/permission issues).

PR #28146, opened 2026-05-15
feat(storage): Add compressed archive support

akashverma0786:
### Describe your changes:

Fixes #<issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### High-level design:

N/A — small change. 

#
#### Manual testing performed

#

Not applicable. 

#
### Checklist:

----
## Summary by Gitar

- **Dependency Security Updates:**
  - Patched multiple CVEs in `ingestion/setup.py` by pinning minimum versions for `protobuf`, `idna`, `lxml`, `Mako`, `mistune`, `PyJWT`, `pyasn1`, and `authlib`.
- **Airflow Stack Hardening:**
  - Added mandatory floor pins for `apache-airflow-providers` and associated web components (`tornado`, `Werkzeug`, `starlette`, `python-multipart`) to mitigate various RCE, DoS, and injection vulnerabilities.


ingestion/src/metadata/readers/archive.py
```diff
@@ -0,0 +1,603 @@
...
+
+_DF_READERS: dict[SupportedTypes, Callable[[BytesIO], object]] = {
+    SupportedTypes.CSV: partial(pd.read_csv, nrows=100),
+    SupportedTypes.CSVGZ: partial(pd.read_csv, compression="gzip", nrows=100),
+    SupportedTypes.TSV: partial(pd.read_csv, sep="\t", nrows=100),
+    SupportedTypes.PARQUET: pd.read_parquet,
+    SupportedTypes.PARQUET_PQ: pd.read_parquet,
+    SupportedTypes.PARQUET_PQT: pd.read_parquet,
+    SupportedTypes.PARQUET_PARQ: pd.read_parquet,
+    SupportedTypes.PARQUET_SNAPPY: pd.read_parquet,
```
IceS2: This reads the whole parquet file into a pandas dataframe no? for big parquet files this is complicated. Since we only need the schema, can't we just read the schema from it?

Something like:
``` 
  import pyarrow.parquet as pq

  def _read_parquet_schema_only(data: BytesIO) -> pd.DataFrame:
      """Read just the column names + arrow types — no row data."""
      data.seek(0)
      schema = pq.read_schema(data)
      return pd.DataFrame({field.name: pd.Series(dtype=field.type.to_pandas_dtype())
                           for field in schema})
```

ingestion/src/metadata/readers/archive.py
```diff
@@ -0,0 +1,603 @@
...
+                try:
+                    # py7zr may extract files with restrictive permissions — ensure we can read them.
+                    file_path.chmod(0o644)
+                    size = file_path.stat().st_size
+                    if size > _MAX_INNER_FILE_BYTES:
+                        logger.warning(f"Skipping {relative_name!r}: size {size} exceeds limit")
+                        continue
+                    yield ArchiveEntry(name=relative_name, data=BytesIO(file_path.read_bytes()), size=size)
+                except Exception as exc:
+                    logger.warning(f"Skipping archive entry {relative_name!r}: {exc}")
```
IceS2: We are storing the file and then doing two passes. Couldn't we collapse this into just one path? Something on the lines of:
```
  class SevenZipArchiveReader(ArchiveReader):
      def __init__(self, data: bytes) -> None:
          import py7zr  # noqa: PLC0415
          import tempfile  # noqa: PLC0415
          from pathlib import Path  # noqa: PLC0415

          # Validate paths from the index BEFORE writing anything to disk
          try:
              with py7zr.SevenZipFile(BytesIO(data), mode="r") as szf:
                  for info in szf.list():
                      if info.is_directory:
                          continue
                      parts = info.filename.replace("\\", "/").split("/")
                      if info.filename.startswith("/") or ".." in parts:
                          raise ValueError(f"Suspicious path in archive: {info.filename!r}")
          except py7zr.Bad7zFile as exc:
              raise ValueError("Corrupted 7Z") from exc

          # Extract once, hold tempdir until close()
          self._tmpdir_ctx = tempfile.TemporaryDirectory()
          self._tmp_path = Path(self._tmpdir_ctx.name)
          with py7zr.SevenZipFile(BytesIO(data), mode="r") as szf:
              szf.extractall(path=self._tmp_path)
          # Drop the raw bytes — no longer needed
          del data

      def entries(self) -> Iterator[ArchiveEntry]:
          for file_path in self._tmp_path.rglob("*"):
              if not file_path.is_file():
                  continue
              relative_name = str(file_path.relative_to(self._tmp_path))
              try:
                  file_path.chmod(0o644)
                  size = file_path.stat().st_size
                  if size > _MAX_INNER_FILE_BYTES:
                      logger.warning(f"Skipping {relative_name!r}: size {size} exceeds limit")
                      continue
                  yield ArchiveEntry(name=relative_name, data=BytesIO(file_path.read_bytes()), size=size)
              except Exception as exc:
                  logger.warning(f"Skipping archive entry {relative_name!r}: {exc}")

      def close(self) -> None:
          self._tmpdir_ctx.cleanup()
```

IceS2 (commented):
One small thing: We should avoid using fstrings on log, we should use %s and the likes

ingestion/src/metadata/readers/archive.py
akashverma0786: Updated the changes

ingestion/src/metadata/readers/archive.py
akashverma0786: Thanks for the suggestion, Updated the changes

ingestion/src/metadata/readers/archive.py
```diff
@@ -0,0 +1,609 @@
...
+                compressed = self._blob.read_range(data_offset, zip_entry.comp_size)
+                if zip_entry.compress_method == 0:
+                    data = compressed
+                elif zip_entry.compress_method == 8:
+                    data = zlib.decompress(compressed, -15)
+                    if len(data) > _MAX_INNER_FILE_BYTES:
+                        logger.warning(
+                            f"Skipping {zip_entry.name!r}: actual decompressed size {len(data)} exceeds limit"
+                        )
+                        continue
```
IceS2: Does this make sense? We are checking the data after we are decompressing it. Shouldn't we do that before?

ingestion/src/metadata/readers/archive.py
```diff
@@ -0,0 +1,609 @@
...
+    """
+    columns: list = []
+    for entry in reader.entries():
+        if entry.name.lower().endswith(tuple(ARCHIVE_EXTENSIONS)):
+            logger.info(f"Skipping nested archive {entry.name!r}")
+            continue
+        entry_format = detect_inner_format(entry.name)
+        if not columns and entry_format is not None:
+            columns = infer_columns_from_archive_entry(entry, entry_format)
+        yield entry, columns, entry_format
```
IceS2, line 626: From this method, it seems we are assuming everything in the zip file has the same schema, but we are still downloading all the files eagerly instead of doing it once at a time and stopping once we have the column list, right?

akashverma0786: Yes it completely makes sense, Refactoring the loading mechanism to check for schema only once

ingestion/src/metadata/readers/archive.py
akashverma0786: We already have a check before this point, this part should not be here , it should have been another check, Making the changes

ingestion/src/metadata/ingestion/source/storage/s3/metadata.py
```diff
@@ -255,6 +263,93 @@ def get_size(self, bucket_name: str, file_path: str) -> Optional[float]:  # noqa
...
+
+        archive_fqn = fqn._build(  # pylint: disable=protected-access
+            getattr(self.context.get(), "objectstore_service"),  # noqa: B009
+            bucket_name,
+            archive_path,
+        )
+        archive_entity = self.metadata.get_by_name(entity=Container, fqn=archive_fqn)
+        if archive_entity is None:
+            logger.warning(f"Archive container {archive_fqn!r} not found after creation; skipping children")
+            return
```
IceS2, line 337: Have you tested this manually? Can you do it? Since we have the bulk sink, this should probably not work at the first try due to the parent not being yet in OM.
Is there any other way we could solve this?

akashverma0786: Yes I have tested that works fine
Also for bulk sink , we have CreateContainerRequest excluded from bulk buffer

IceS2 (approved):

PR #28149, opened 2026-05-15
feat(snowflake): opt-in ACCESS_HISTORY lineage path 

ulixius9:
## Summary
- New Snowflake lineage path that reads precomputed table and column lineage from `ACCOUNT_USAGE.ACCESS_HISTORY` instead of parsing every relevant query client-side. Targets large query windows where the current sqlglot/sqlfluff/sqlparse pipeline becomes a wall-clock bottleneck.
- Opt-in via `connectionOptions.useAccessHistory: \"true\"` on the Snowflake service connection. Default off — existing pipelines see zero change. A runtime probe demotes silently to the legacy parser path when ACCESS_HISTORY isn't readable (Standard Edition or missing `IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE`).
- Single combined SQL: `MAX_BY` dedup on table edges, `ARRAY_AGG(DISTINCT OBJECT_CONSTRUCT(...))` on column pairs per `(downstream, upstream)` edge, `LEFT JOIN`'d so the cursor streams one row per directed edge with column lineage already attached. Constant client memory regardless of edge count.
- COPY_HISTORY surfaced separately for external stage → Container lineage. External stages (`s3://`, `azure://`, `gcs://`, `https://`) are resolved against ingested Container entities via `es_search_container_by_path`; internal Snowflake stages (`@~/`, `@%table/`, `@db.schema.stage/`) are skipped silently.
- Stored-procedure body lineage path (`StoredProcedureLineageMixin`) is unchanged and continues to run regardless of the flag.
- POC scope: configured via existing `connectionOptions` Map<String,String> on the Snowflake connection — no JSON schema or generated-model changes. If validated, a follow-up PR will promote the key to a first-class field.

## How to enable
Customer pipeline YAML:
```yaml
serviceConnection:
  config:
    type: Snowflake
    ...
    connectionOptions:
      useAccessHistory: \"true\"
sourceConfig:
  config:
    queryLogDuration: 180  # set as wide as needed; ACCESS_HISTORY retains ~365 days
```

## Files changed
- \`ingestion/src/metadata/ingestion/source/database/snowflake/queries.py\` — three new SQL constants: \`SNOWFLAKE_ACCESS_HISTORY_PROBE\`, \`SNOWFLAKE_ACCESS_HISTORY_LINEAGE\` (combined table+column), \`SNOWFLAKE_COPY_HISTORY_LINEAGE\`.
- \`ingestion/src/metadata/ingestion/source/database/snowflake/connection.py\` — \`probe_access_history_available(engine, schema)\` helper; failures logged at INFO (Standard Edition is a legitimate state).
- \`ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py\` — \`__init__\` reads and pops the option, runs probe, demotes on failure; \`yield_query_lineage\` dispatches to the new path or falls through to \`super()\`; new private yielders for ACCESS_HISTORY and COPY_HISTORY with LRU-cached entity resolution; \`_parse_column_pairs\` decodes the VARIANT array (handles both Python-list and JSON-string forms).
- \`ingestion/tests/unit/topology/database/test_snowflake_access_history_lineage.py\` — 25 unit tests covering SQL rendering, connectionOptions parsing and key pop, probe-failure demote, table edge yielding, column lineage attachment with single and multiple pairs, COPY edge resolution (resolved / unresolved / internal), legacy-parser-bypass regression, and \`_parse_column_pairs\` robustness.

## Test plan
- [x] \`make py_format_check\` clean (ruff lint + format)
- [x] 25 new unit tests pass
- [x] 51/51 Snowflake unit tests pass (no regression in existing \`test_snowflake.py\` or \`test_snowflake_ordinal_position.py\`)
- [ ] Manual run on an Enterprise Snowflake account with \`useAccessHistory: \"true\"\` and \`queryLogDuration: 180\`; record wall time and edge count
- [ ] Verify edge count ≥ legacy path on the same window
- [ ] Verify column lineage attaches end-to-end (e.g., \`orders.amount → revenue.total_amount\` on a known graph)
- [ ] Verify Standard Edition account: probe demotes silently and behavior is identical to today
- [ ] Verify pipeline without \`useAccessHistory\` set: behavior is bit-identical to today (default off)

## Notes
- ACCESS_HISTORY has ~45-minute freshness — daily/weekly cron pipelines unaffected; not suitable for near-real-time lineage.
- \`directSources\` is NULL for INSERT…VALUES, CALL (stored procs), and certain dynamic SQL. Proc body lineage is preserved via the existing \`StoredProcedureLineageMixin\` join.
- COPY edges depend on the upstream Container being ingested via a Storage Service ingestion; unresolved external stages are logged at INFO so the operator can decide which storage services to ingest next.
- No incremental checkpointing wired in this POC; each run scans the full configured \`queryLogDuration\` window.

----
## Summary by Gitar

- **Refined SQL lineage:**
  - Optimized `SNOWFLAKE_ACCESS_HISTORY_LINEAGE` to fetch `QUERY_TEXT` directly via `MAX_BY` instead of a secondary `LEFT JOIN` to `QUERY_HISTORY`.
- **Improved observability:**
  - Updated lineage logging to track and report the count of edges that successfully captured associated `QUERY_TEXT`.
- **Robustness improvements:**
  - Enhanced `_row_to_lower_dict` to support both SQLAlchemy `Row` objects and standard dictionaries for uniform model field resolution.
  - Relaxed type checking in `lineage.py` for improved compatibility with varied row-like structures.


ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -55,6 +93,59 @@ class SnowflakeLineageSource(SnowflakeQueryParserSource, StoredProcedureLineageM
...
+        self,
+        config: WorkflowSource,
+        metadata: OpenMetadata,
+        get_engine: bool = True,
+    ):
+        # Pop the OM-specific flag from connectionOptions BEFORE the parent
+        # creates the SQLAlchemy engine — the Snowflake URL builder copies
+        # every connectionOptions entry into the URL query string, so the
+        # driver would otherwise receive an unknown `useAccessHistory` param.
+        self._use_access_history = self._pop_access_history_flag(config)
```
IceS2, line 113: Are we sure we want to pass it through connectionOptions and pop it like this?
Feels like a recipe for future issues.
How are we sure we ain't passing it in other places we use the SnowflakeConnectior?

ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
```diff
@@ -114,6 +115,28 @@ def test_table_query(engine_wrapper: SnowflakeEngineWrapper, statement: str):
...
+            conn.execute(text(SNOWFLAKE_ACCESS_HISTORY_PROBE.format(account_usage=account_usage_schema)))
+    except Exception as exc:
+        logger.info(
+            f"ACCESS_HISTORY probe failed (will fall back to legacy lineage path): {exc}. "
+            f"Ensure the role has IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE and the account is Enterprise+."
+        )
+        return False
+    return True
+
+
```
IceS2, line 139: If we are only using it within lineage.py I would argue we could keep it there.
Eventually I think it would make sense to have it on a conditional TestConnection

ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -55,6 +93,59 @@ class SnowflakeLineageSource(SnowflakeQueryParserSource, StoredProcedureLineageM
...
+        super().__init__(config, metadata, get_engine=get_engine)
+        self._table_cache: Dict[str, Optional[Table]] = {}  # noqa: UP006, UP045
+        self._container_cache: Dict[str, Optional[Container]] = {}  # noqa: UP006, UP045
```
IceS2: I am not aware of the lifecycle here, but could we end up having OOM issues due to this maps here? Any reason for the value to be optional?

ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -55,6 +93,59 @@ class SnowflakeLineageSource(SnowflakeQueryParserSource, StoredProcedureLineageM
...
+        if self._use_access_history and self.engine is not None:
+            available = probe_access_history_available(self.engine, self.service_connection.accountUsageSchema)
+            if not available:
+                logger.info(
+                    "useAccessHistory was set in connectionOptions but the ACCESS_HISTORY probe failed; "
+                    "falling back to legacy QUERY_HISTORY parser path."
+                )
+                self._use_access_history = False
```
IceS2: Shouldn't this be a Warning?

ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -114,3 +205,305 @@ def yield_table_query(self) -> Iterator[TableQuery]:
...
+                        emitted += 1
+                        yield Either(right=edge)  # pyright: ignore[reportCallIssue]
+        except Exception as exc:
+            logger.warning(f"Failed to extract lineage from ACCESS_HISTORY: {exc}")
+            logger.debug(traceback.format_exc())
+        logger.info(
+            f"ACCESS_HISTORY lineage: emitted {emitted} edges, "
+            f"skipped {skipped} (unresolvable downstream/upstream tables)"
+        )
+
```
IceS2: nit: We should aim to move away from fstrings on logging to avoid eager resolution when uneeded

ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -114,3 +205,305 @@ def yield_table_query(self) -> Iterator[TableQuery]:
...
+        with fully qualified column names. Drops pairs where either column does
+        not exist on its parent table entity.
+        """
+        result: List[ColumnLineage] = []  # noqa: UP006
+        for d_col, u_col in column_pairs:
+            d_fqn = get_column_fqn(downstream_entity, d_col)
+            u_fqn = get_column_fqn(upstream_entity, u_col)
+            if d_fqn and u_fqn:
+                result.append(ColumnLineage(fromColumns=[u_fqn], toColumn=d_fqn))  # pyright: ignore[reportCallIssue]
+        return result
```
IceS2: Should we simplify the logic in this module by adding some Dataclasses that know about the mapping logic?

ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -114,3 +205,305 @@ def yield_table_query(self) -> Iterator[TableQuery]:
...
+        if downstream_entity is None:
+            return None
+
+        container_entity = self._resolve_container_by_path(stage_location)
+        if container_entity is None:
+            logger.info(
+                f"COPY edge unresolved: no Container ingested for stage `{stage_location}` "
+                f"(downstream table `{downstream_fqn}` skipped)"
+            )
+            return None
```
IceS2: Similar idea as above, could Dataclasses help us simplify the code here?

ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -114,3 +205,305 @@ def yield_table_query(self) -> Iterator[TableQuery]:
...
+        return stage_location.lower().startswith(EXTERNAL_STAGE_PREFIXES)
+
+    @staticmethod
+    def _row_to_dict(row: Any) -> dict:
+        """
+        SQLAlchemy result rows expose columns as attributes; normalize to a
+        lower-cased dict so downstream code can use uniform keys.
+        """
+        raw = row._asdict() if hasattr(row, "_asdict") else dict(row)
+        return {k.lower(): v for k, v in raw.items()}
```
IceS2, line 518: Could we have use the dict cursor for this?

ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -114,3 +205,305 @@ def yield_table_query(self) -> Iterator[TableQuery]:
...
+        if stage_location in self._container_cache:
+            return self._container_cache[stage_location]
+        try:
+            results = self.metadata.es_search_container_by_path(full_path=stage_location) or []
+            entity = results[0] if results else None
+        except Exception as exc:
+            logger.debug(f"Failed to resolve Container for path `{stage_location}`: {exc}")
+            entity = None
+        self._container_cache[stage_location] = entity
+        return entity
```
IceS2: Since the cache we are using is to avoid hitting the backend, we should work with cachetools caches to avoid OOM issues.

IceS2 (dismissed):

ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
ulixius9: yes this was implemented here coz we would need it in test connection... but right now don't want to implement schema changes hence it is only used it in lineage.py

harshach (approved):

PR #28160, opened 2026-05-15
fix(ingestion): stop StreamableLogHandler recursing into itself when queue is full

harshach:
## Summary

- hang in ingestion traced to infinite recursion in `StreamableLogHandler`, not to Snowflake, HTTP timeouts, or stale connection pools.
- `StreamableLogHandler` is attached to the root ingestion logger. `emit()`, `_ship_logs()`, `_worker_loop()`, and `close()` all called that **same** logger (`logger.warning(...)`, `logger.error(...)`, `logger.info("[FALLBACK] ...")`, etc.) to record their own diagnostics. Every such call was dispatched back through the handler chain and re-entered `emit()` against a still-full queue, recursing until Python's recursion limit was hit.
- Route those in-handler diagnostics through a dedicated `_internal_logger` with `propagate=False` and its own `StreamHandler(sys.stderr)`. No other behavior changes.

## What was actually happening (from a py-spy dump of a hung pod)

- **MainThread** sat ~977 frames deep, alternating between two lines forever:
  - `emit (streamable_logger.py:390)` → `logger.warning("Log queue is full…")`
  - `emit (streamable_logger.py:395)` → `logger.error(\"Error in emit: ...\")` (RecursionError caught by the outer except, which logged again, which re-entered)
- **log-shipper thread** stuck at `acquire (logging/__init__.py:917)` — blocked on the logging module's internal lock held by MainThread.
- **status-reporter thread** stuck at the same lock.
- → No logs ship anywhere. No progress. From outside the pod the connector looks dead. The previously attempted HTTP-timeout fix could not help; nothing was waiting on the network.

## Why it manifested as "Connector hang"

The bug only triggers once the bounded async log queue (`max_queue_size=10000`) fills. Snowflake metadata ingestion at scale produces hundreds of debug logs/sec (one per `yield_table` via `ExecutionTimeTracker`) and saturates the queue quickly. Smaller connectors against smaller sources never hit the condition.

## Scope (deliberately minimum)

This PR fixes only the recursion. Follow-up work to land in a separate PR:

- Per-entry size cap and per-batch byte budget
- Drop-with-count policy + periodic drop summary
- Flip log compression to default-on (server already supports it)
- Runtime diagnostics subsystem (signal handler + watchdog) — design doc in flight

## Files changed

- `ingestion/src/metadata/utils/streamable_logger.py` — add module-level `_internal_logger`; swap 6 `logger.*` callsites inside the handler hot path to use it.
- `ingestion/tests/unit/utils/test_streamable_logger.py` — new `TestStreamableLogHandlerRecursionRegression` with 3 tests that fail if anyone reintroduces a `logger.*` call inside `emit` / `_ship_logs` / `_worker_loop` / `close`.

The local-fallback path (`self.fallback_handler.emit(record)`) is unchanged — when the queue is full or shipping fails, users still see their logs in `kubectl logs` exactly as before. No log entries are dropped by this PR.

## Test plan

- [x] `pytest ingestion/tests/unit/utils/test_streamable_logger.py` — 34/34 pass (31 existing + 3 new regression)
- [x] New tests fail if `_internal_logger` is reverted to plain `logger` (manually verified by temporary revert)
- [ ] Deploy to one Snowflake-ingestion environment that previously hung; confirm the connector completes without sitting in a recursive `emit()` loop
- [ ] Confirm `kubectl logs` still shows fallback log lines on simulated queue-full conditions

### DIFF
--- a/ingestion/src/metadata/utils/streamable_logger.py
+++ b/ingestion/src/metadata/utils/streamable_logger.py
@@ -56,6 +56,7 @@ handler = setup_streamable_logging_for_workflow(
 import logging
 import os
 import queue
+import sys
 import threading
 import time
 from enum import Enum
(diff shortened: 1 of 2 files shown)


harshach:
cherry-picked manually into 1.12.8 , 1.12.9 and 1.13

PR #28161, opened 2026-05-15
ingestion: runtime diagnostics subsystem

harshach:
## Summary

Design doc for an opt-in, always-available diagnostics layer inside the ingestion framework. Goal: when a connector hangs, OOMs, or slows down, the evidence needed to identify the root cause is in `kubectl logs` — no `py-spy`, no `kubectl debug`, no ptrace, no shipping random fixes.

**This PR is the design only.** Implementation lands as six small, separately reviewable PRs described in §10.

## Why now

Three concrete production failures motivated this:

1. **The "Snowflake hang"** — actually a logging recursion bug in `StreamableLogHandler` (fixed in #28160). Took ~6 hours of investigation plus one wrong-theory fix (HTTP timeouts) to identify, because the process gave us no live state on what it was doing.
2. **Recurring OOMKills** — exit code 137, no Python traceback, no last-state log. We have no way to attribute growth to a specific object type or stage, so OOM fixes are guesses.
3. **"Is it stuck or just slow?"** questions from customers we cannot answer without `kubectl exec` and ad-hoc tooling.

The design is grounded in §1 case studies and §11 shows what these instrumentations would have told us at the time.

## Design at a glance

- **Activation:** `workflowConfig.loggerLevel == DEBUG`. No new env vars, no new config fields.
- **Cost when off:** zero. Module is dead code, `operation()` is a no-op contextmanager.
- **Cost when on:** ~250 KB RSS, <0.01% CPU.
- **Components:**
  - Operation registry — "what is each thread doing right now"
  - SIGUSR1 / SIGUSR2 handlers — on-demand dumps to stderr
  - Watchdog — auto-logs at 60s, auto-dumps at 300s
  - Heartbeat — one structured progress line every 30s
  - Memory tracker — RSS, cgroup pressure, `gc.get_objects()` top-types on dump
  - Stage backpressure — queue depths between source/processor/sink
  - HTTP introspection — OMetaClient and DB cursor `execute()`

## Implementation plan (six PRs)

| PR | Scope | LoC | Value shipped on its own |
|---|---|---|---|
| 1 | Foundation (registry + signals + minimal wiring) | ~350 | `kill -SIGUSR1` works for on-demand dumps |
| 2 | Watchdog + heartbeat | ~250 | Process logs its own hangs |
| 3 | Memory tracker | ~200 | OOMs become diagnosable |
| 4 | Stage backpressure | ~150 | We can see which stage is the bottleneck |
| 5 | DB cursor wrapping | ~200 | SQL-level visibility |
| 6 | Disk-persisted dumps | ~50 | Survives log rotation |

After PR 1 alone, the diagnostic gap that delayed the Snowflake root-cause analysis is closed.

## Anti-patterns explicitly rejected (§14)

To keep us honest, the doc enumerates what we are NOT building and why — new env vars, new config fields, periodic remote shipping of diagnostic state, a dashboard sidecar, full APM. The principle throughout: minimum new surface, maximum information per line of log output.

## Test plan

- [ ] Review for design soundness; questions in §13 ("Open questions")
- [ ] Once approved, ship PR 1 (foundation) per §10
- [ ] Validate on the next DEBUG-enabled run of a previously-hanging Snowflake pipeline

----
## Summary by Gitar

- **HTTP client enhancements:**
  - Added `post_best_effort` method for fire-and-forget requests and configurable `timeout`/`retries` for `_request`.
  - Implemented `RestTransportError` and consolidated transport-layer exceptions to improve failure observability.
- **Workflow diagnostics & metadata:**
  - Added `User-Agent` support to `OMetaClient` to improve server-side request identification.
  - Reordered workflow `stop` and `cleanup` sequences to ensure `diag.time_budget` logs are captured before shutdown.


edg956 (dismissed):
LGTM. I believe DB and HTTP introspection could be addressed by OpenTelemetry but this could work as well

PR #28199, opened 2026-05-17
fix(rdf): scope glossary graph by membership and surface glossary labels

harshach:
## Summary

- `/rdf/glossary/graph?glossaryId=<id>` silently ignored its filter and returned every term across every glossary, because the SPARQL bound `?glossary` via `OPTIONAL { ?term1 om:belongsTo ?glossary }` but the JSON-LD context maps `GlossaryTerm.glossary` to **`om:belongsToGlossary`**. The downstream `FILTER(?glossary = <…>)` was a no-op.
- The UI hierarchy view rendered each leaked glossary's terms inside a group container whose label was the raw glossary UUID (no `glossaryId`/`group` on the response, so the UI fell through `glossaryNames[id] ?? id`).
- A handful of cascading projection / read-side bugs surfaced once the predicate was fixed: term mutations dropped `rdf:type`/`rdfs:label`/`om:belongsToGlossary`; empty `skos:prefLabel = ""` won over `rdfs:label` and rendered as blank node labels; one-sided RDF deletion left a reverse triple behind; the DB-fallback was scoped via a `ListFilter` param the table doesn't recognize.

This PR fixes all of the above and adds nine integration tests that lock in the user-visible behaviors end-to-end through both the RDF path and the DB-backed term API.

## Backend (`openmetadata-service`)

- `RdfRepository.buildGlossaryTermGraphQuery` — use `om:belongsToGlossary`, make membership a required triple when scoped, COALESCE glossary label from `skos:prefLabel`/`rdfs:label`.
- `RdfRepository.parseGlossaryTermGraphResults` — emit `group` + `glossaryId` per node, backfill membership when a node is upgraded from term2→term1, blank-resilient label extraction, DB-resolved glossary-name fallback.
- `RdfRepository.addRelationship` — additive `INSERT DATA` instead of `storeEntity` (which swept identity predicates off the source entity).
- `RdfRepository.removeGlossaryTermRelation` — delete both directions so the reverse triple written by `EntityRepository.addRelationship` doesn't linger.
- `RdfRepository.getGlossaryTermGraphFromDatabase` — scope by `glossary.id` in-memory; `ListFilter` has no `glossary` predicate for the table, so the previous `addQueryParam` was a silent no-op.
- `RdfUpdater.addRelationship` / `removeRelationship` — short-circuit `glossaryTerm ⇔ glossaryTerm RELATED_TO`; the typed `addGlossaryTermRelation` path owns those edges with the precise predicate (`skos:exactMatch`, `skos:broader`, …) so type changes don't leave stale `om:relatedTo` triples.
- `RdfPropertyMapper.addTypedProperty` — skip blank `xsd:string` literals (one canonical source of truth on the write side).

## UI (`openmetadata-ui`)

- `convertRdfGraphToOntologyGraph` prefers the explicit `node.glossaryId` from the response over the FQN-prefix heuristic.
- `hierarchyGraphBuilder` combo label falls back to a node's `group` when the glossary id isn't in the caller's visible glossary list — so the container never renders as a raw UUID.
- `GraphNode` typed with the new optional `glossaryId`.

## Tests (`openmetadata-integration-tests`)

`RdfGlossaryGraphIT` (new, 9 tests):

1. `glossaryIdFilterScopesGraphToRequestedGlossary` — only requested glossary's terms returned.
2. `scopedResponseCarriesGlossaryNameAndIdPerNode` — `group` + `glossaryId` populated per term.
3. `termNodeLabelFallsBackToNameWhenDisplayNameIsAbsent` — label = term name when no displayName, never UUID.
4. `labelTracksDisplayNameLifecycle` — displayName: unset → set → renamed → cleared → still resolves correctly through both the RDF graph and the DB term API.
5. `glossaryMembershipSurvivesAddAndDeleteRelations` — `belongsToGlossary` / label preserved through add+delete cycles; cross-checked via DB.
6. `sameTermPairCanHoldMultipleRelationTypesIndependently` — two relation types between the same pair coexist; deleting one leaves the other intact.
7. `changingRelationTypeReplacesOldEdgeWithNewType` — delete-then-add to a new type leaves no orphan of the old type.
8. `crossGlossaryRelationDoesNotLeakIntoOtherGlossaryScope` — cross-glossary relation does NOT make the source term a member of the target glossary.
9. `glossaryIdFilterReturnsEmptyForGlossaryWithNoTerms` — DB-fallback path also scopes correctly.

```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0
```

## Test plan

- [x] `mvn -pl openmetadata-integration-tests test -Dtest=RdfGlossaryGraphIT` — 9/9 green
- [x] `mvn spotless:apply` — clean
- [x] `yarn organize-imports + eslint --fix + prettier` on changed UI files — clean
- [x] Manual UI verification on a local Docker stack with Fuseki enabled (Finance + Marketing glossaries with hierarchy, intra- and cross-glossary relations): scoped graphs render only the requested glossary's terms with proper group labels.

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/RdfGlossaryGraphIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/RdfGlossaryGraphIT.java
@@ -1,0 +1,780 @@
+/*
+ *  Copyright 2025 Collate
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.openmetadata.it.tests;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.time.Duration;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+import org.awaitility.Awaitility;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+import org.junit.jupiter.api.parallel.Isolated;
+import org.openmetadata.it.bootstrap.TestSuiteBootstrap;
+import org.openmetadata.it.factories.GlossaryTermTestFactory;
+import org.openmetadata.it.factories.GlossaryTestFactory;
+import org.openmetadata.it.util.SdkClients;
+import org.openmetadata.it.util.TestNamespace;
+import org.openmetadata.it.util.TestNamespaceExtension;
+import org.openmetadata.schema.api.configuration.rdf.RdfConfiguration;
+import org.openmetadata.schema.entity.data.Glossary;
+import org.openmetadata.schema.entity.data.GlossaryTerm;
+import org.openmetadata.service.rdf.RdfUpdater;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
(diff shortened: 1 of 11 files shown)


harshach:
Addressed in e5b8b512fa.

The DB fallback now drives scoping at the database level via `glossaryTermDAO.getNestedTerms(glossaryFqn)`:

```sql
SELECT json FROM glossary_term_entity WHERE fqnHash LIKE :prefix
```

which is an indexed `fqnHash`-prefix scan that only returns the requested glossary's terms. We then parse the result JSON for ids and bulk-hydrate `relatedTerms` / `parent` / `children` via `EntityRepository.get(uri, ids, fields, include)` in a single batched call — no more `listAll` + in-memory filter.

The unscoped branch (`glossaryId == null`) is unchanged because that's the documented "all glossaries" mode; the regression was strictly about silently doing a full scan when the caller DID supply a scope.

All 9 `RdfGlossaryGraphIT` tests still green:
```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0
```

harshach:
Good catch — they weren't sharing the path before. e181f948ee makes the RDF DB fallback drive through the **same** code as `GET /v1/glossaryTerms?glossary=<id>`:

```java
var listFilter = new ListFilter(null);
if (glossaryId != null) {
  var glossary = (Glossary) glossaryRepo.get(null, glossaryId, …);
  listFilter.addQueryParam("parent", glossary.getFullyQualifiedName());
}
var fetched =
    glossaryTermRepository.listAll(
        glossaryTermRepository.getFields("relatedTerms,parent,children"),
        listFilter);
```

That's exactly what `GlossaryTermResource.list(...)` does — `parent` flows through `ListFilter.getParentCondition` → `getFqnPrefixCondition` → an indexed `fqnHash LIKE '<prefix>.%'` predicate on `glossary_term_entity`. Single read path now serves both the public listing and the RDF graph DB fallback; no parallel `getNestedTerms` + manual id hydration.

Same commit also collapses the freshly-introduced FQN class names (`org.openmetadata.schema.entity.data.GlossaryTerm`, etc.) into proper imports per the convention.

All 9 `RdfGlossaryGraphIT` tests still green.

PR #4122, opened 2022-04-14
Fix Py tests

pmbrull:
### Describe your changes :

I worked on the ..... because ...

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] Documentation

#
### Frontend Preview (Screenshots) :
<p align="center">For frontend related change, please link screenshots of your changes preview! Optional for backend related changes.
</p>

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/open-source-community/developer) document.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] All new and existing tests passed.

#

PR #28214, opened 2026-05-18
Fixes open-metadata/openmetadata-collate#4122: flatten nested children to avoid ES mapping depth limit

mohityadav766:
### Describe your changes:

Fixes open-metadata/openmetadata-collate#4122

Search indexing failed at the Sink stage with `mapper_parsing_exception`
("Failed to parse") for entities whose schemas/columns nest records
deeper than ~17 levels. The recursive `children` tree pushed the
Elasticsearch object path past the default `index.mapping.depth.limit`
of 20, so the bulk item was rejected.

This PR maps the recursive `children` sub-field as `flattened` for every
affected entity, so the whole child subtree collapses into a single
field and object depth is capped at 3 regardless of nesting.

**Entities fixed (recursive `children`):**

| Entity | Field |
|--------|-------|
| Topic | `messageSchema.schemaFields.children` |
| API Endpoint | `request/responseSchema.schemaFields.children` |
| Table | `columns.children` |
| Dashboard Data Model | `columns.children` |
| Container | `dataModel.columns.children` |
| Worksheet | `columns.children` |
| Search Index | `fields.children` |

#
### Type of change:

- [x] Bug fix

#
### High-level design:

- The depth blow-up comes entirely from the **recursive** `children`
  field. Mapping it as `flattened` collapses the whole child subtree
  into one field, capping object depth at 3 regardless of nesting.
- Only `children` is flattened. Top-level `columns` / `fields` /
  `schemaFields` keep their normal analyzed mapping, so exact-match,
  sort and aggregations on the first level (and the column-grid
  aggregator) are unchanged.
- `OsUtils.transformFieldTypesForOpenSearch()` already auto-translates
  `flattened` → `flat_object`, so the change works on Elasticsearch and
  OpenSearch with no extra code.
- table / dashboardDataModel / container had **no** `children` mapping
  at all (dynamic-mapped → re-introduces the depth blow-up), so an
  explicit `flattened` `children` is added; topic(zh) / worksheet /
  searchIndex had it as a one-level object and are converted.
- `flattened` has no `.keyword` sub-field, so `.keyword` references to
  the now-flattened children fields are removed:
  - `TopicIndex` / `APIEndpointIndex` / `SearchEntityIndex` `getFields()`
  - `searchSettings.json` — Topic/APIEndpoint/Table exact-match boosts
    and the documented field list retargeted onto the flattened
    `*.children.name` virtual keys (addresses Copilot review).
- Known trade-off: highlighting and boost weighting of deeply-nested
  child column/field matches degrade (flattened highlighting is
  limited). Search still finds these values; top-level highlighting is
  unaffected. Accepted as preferable to raising the depth limit, which
  only defers the problem and bloats mappings.

Files changed: 18 `*_index_mapping.json` (topic, api_endpoint, table,
dashboard_data_model, container, worksheet, search_entity × `en`, `jp`,
`ru`, `zh`), `TopicIndex.java`, `APIEndpointIndex.java`,
`SearchEntityIndex.java`, and `searchSettings.json`.

#
#### Use cases covered
- A Topic / Table with schemas/columns nesting `children` 20 levels deep
  is reindexed without a Sink-stage "Failed to parse" failure.
- Exact-match and aggregations on top-level `*.name(.keyword)` are
  unchanged; first-level `*.children.name` remains queryable.

- [ ] Not applicable (search index mapping change; no API endpoint changes).

#### Manual testing performed
1. Reproduced the failure: indexing a deep Topic / Table document into
   the existing indices returned `mapper_parsing_exception: The depth of
   the field has exceeded the allowed limit of [20]`.
2. Built each new mapping, applied the `flattened` → `flat_object`
   transform exactly as `OsUtils` does, created the indices on OpenSearch.
3. Indexed real deep documents — a Topic (`schemaFields` depth 20,
   276 KB `schemaText`) and a Table (`columns` depth 20, 18 columns) —
   both returned `HTTP 201`.
4. Verified top-level `*.name.keyword` term queries + `terms`
   aggregations still work, and first-level `*.children.name` queries
   match.

#

Not applicable.

#
### Checklist:

- [x] My PR is linked to a GitHub issue via `Fixes` above.
- [ ] For JSON Schema changes: not applicable — this changes search index mappings, not entity JSON Schemas. Existing indices must be recreated/reindexed for the new mappings to take effect.


mohityadav766:
Addressed the Copilot review comments and extended the fix:

- `searchSettings.json` — Topic and API Endpoint exact-match boosts (and the documented field list) retargeted off `*.schemaFields.children*.keyword` onto the flattened `*.schemaFields.children.name` virtual keys, so the boosts keep working with the new mapping.
- Extended the same `children` → `flattened` fix to **Table, Dashboard Data Model, Container, Worksheet, and Search Index** — they have the identical recursive-`children` depth bug (per issue #4122, which names Topic *and* Table). `SearchEntityIndex.getFields()` `.keyword` reference dropped accordingly; `searchSettings.json` Table boost updated too.

Verified by indexing real depth-20 Topic and Table documents against OpenSearch — both index successfully with the new mappings.

harshach (approved):

pmbrull (approved):

mohityadav766:
Re: the review comment about `columns.children.name` exact-match losing case-insensitivity after switching `children` to `flattened` —

The mechanism is correct: `flattened`/`flat_object` can't carry a `normalizer`, and exact-match fields run a `term` query with `query.toLowerCase()` (`addExactMatchQueriesV2`), so a lowercased query won't match a child name stored in mixed/upper case.

However, **this is not a regression introduced by this PR.** I checked the pre-merge mappings — the `*.children.name` fields never had a `lowercase_normalizer`:

| Field | Before this PR | After |
|---|---|---|
| topic `schemaFields.children.name` | `keyword` + `.keyword` subfield — no normalizer | flattened — no normalizer |
| table `columns.children` | not mapped → dynamic `text`+`keyword` — no normalizer | flattened — no normalizer |
| apiEndpoint `…schemaFields.children.keyword` | field never existed | flattened |

Only the **top-level** `*.name.keyword` fields carry `lowercase_normalizer`, and those are untouched. So exact-match on a mixed-case **nested child** name was already case-sensitive before this change — flattening keeps the behavior identical, it doesn't worsen it.

Impact is also low: these are exact-match **boost** entries, not filters — the entity is still found via the other clauses (`columnNamesFuzzy`, standard match, top-level column fields); only a relevance bump is missed for mixed-case nested names.

Treating this as a pre-existing limitation rather than a blocker for this PR. Happy to do a follow-up if we want case-insensitive exact match on nested fields — that needs either an index-time normalized field or making the `term` query case-insensitive for flattened fields.

PR #28293, opened 2026-05-20
fix(ometa): use requests instead of httpx for SSE transport

pmbrull:
## Summary
- Switch `SSEClient` in `metadata.ingestion.ometa.sse_client` from `httpx` to `requests` for the SSE stream. The SDK's REST client already uses `requests`, so this aligns the SSE path with the proven transport.
- In some enterprise environments `httpx`'s HTTP/1.1 wire pattern (lowercase header names, request headers and body in separate TCP segments) is reset by network security middleboxes that pass `requests` / `wget` traffic through unchanged. Symptom: `httpx.ReadError: [Errno 104] Connection reset by peer` at `_receive_response_headers`, with **no server-side access-log entry** because the reset happens before any HTTP response is committed. `requests` and `wget` succeed against the same endpoint from the same pod with the same JWT and body.
- Behavior preserved: same retry loop, same SSE parser, same `Last-Event-ID` resume on retry, same INFO/ERROR log messages (`Connected to SSE stream`, `Stream terminated with event: ...`, `HTTP error: <code>`, `Connection error (retry N/M): ...`).
- **Drive-by fix:** the previous `httpx` implementation built `opts` with `allow_redirects` / `verify` / `cookies` but never passed those keys to `httpx.stream(...)` — a latent bug. `requests.Session.request` now wires them up properly (safe defaults: `verify=True`, `allow_redirects=True` when unset).
- No new dependency: `requests` is already a transitive dep used by the SDK's REST client.

## Test plan
- [ ] \`pytest ingestion/tests/unit/metadata/ingestion/ometa/test_sse_client.py\` passes (existing suite, no test changes required).
- [ ] In an environment where the documentation agent application previously failed with \`Connection reset by peer\` on every table, the workflow now logs \`Connected to SSE stream\` and \`Stream terminated with event: stream-completed\` per table.
- [ ] In a healthy environment (dev/CI), the same workflow continues to succeed end-to-end.
- [ ] Manually exercise \`metadata app -c <doc-agent-config>\` against a server with TLS enabled to confirm \`verify\` / \`cookies\` from \`ClientConfig\` now flow through the SSE path.

### DIFF
--- a/ingestion/src/metadata/ingestion/ometa/sse_client.py
+++ b/ingestion/src/metadata/ingestion/ometa/sse_client.py
@@ -101,19 +101,13 @@ class SSEClient:
                     "params": opts.get("params"),
                     "stream": True,
                     "timeout": timeout,
-                    "verify": (
-                        self.config.verify if self.config.verify is not None else True
-                    ),
+                    "verify": (self.config.verify if self.config.verify is not None else True),
                     "allow_redirects": (
-                        self.config.allow_redirects
-                        if self.config.allow_redirects is not None
-                        else True
+                        self.config.allow_redirects if self.config.allow_redirects is not None else True
                     ),
                     "cookies": self.config.cookies,
                 }
-                with requests.Session() as session, session.request(
-                    **request_kwargs
-                ) as response:
+                with requests.Session() as session, session.request(**request_kwargs) as response:
                     response.raise_for_status()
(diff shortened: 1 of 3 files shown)


IceS2 (approved):

pmbrull:
picked by hand to 1.12.9 and 1.13

PR #28300, opened 2026-05-20
perf(ingestion): collapse Databricks per-table DESCRIBE storm

ulixius9:
## Summary

The Databricks `DatabaseMetadata` workflow issued a full `DESCRIBE TABLE EXTENDED` for **every** consumer of a table's metadata — type detection, columns, comment, description and owner — plus a per-call `SHOW SCHEMAS` inside `get_view_definition`. For a ~300-table service this produced **~20k redundant warehouse round-trips** and **8h+ runtimes**, leaving the job exposed to orchestration-level kills (deadline / eviction). Diagnosis of two production run logs showed flat ~600 MB RSS throughout — this is a **query-storm / runtime** problem, not a memory OOM.

This PR collapses the storm:

- **One metadata fetch per table on DBR 16.2+** via `DESCRIBE TABLE EXTENDED ... AS JSON`. The parsed payload (columns, comment, owner, location, view text) is shared by all four consumers through a **size-1 per-dialect cache**, evicted when the next table is fetched — never accumulated across the run.
- **Bulk table-type detection**: one `information_schema.tables` query per schema replaces one `DESCRIBE` per table during enumeration.
- **Backward compatible**: runtimes **< 16.2** transparently fall back to the legacy per-statement `DESCRIBE` path after the first `AS JSON` query errors; a per-dialect flag short-circuits later tables so older runtimes pay no repeated failed-query cost.
- **Memory hygiene**: `set_inspector` now disposes the previous engine and clears the external-location map on catalog switch, bounding memory across multi-catalog services.

## Changes

| File | What |
|---|---|
| `databricks/metadata.py` | AS-JSON fetch + size-1 cache; bulk table-type query; engine disposal in `set_inspector`; `SHOW SCHEMAS` hoisted out of per-call path |
| `databricks/models.py` | Extended `DescribeJson*` Pydantic models for the full AS-JSON payload |
| `databricks/queries.py` | New `DATABRICKS_GET_TABLE_DESCRIBE_JSON` and `DATABRICKS_GET_TABLE_TYPES` constants |
| `tests/.../test_databricks_query_reduction.py` | 21 new unit tests covering AS-JSON parsing, fallback, cache bounding, and per-schema type detection |
| `.basedpyright/baseline.json` | Net reduction (88 stale entries removed, 24 refactor-position entries added) |

## Test plan

- [x] `test_databricks_query_reduction.py` — 21 passed
- [x] Existing `test_databricks.py` — 23 passed (3 pre-existing localhost-network tests deselected, fail identically on `main`)
- [x] `ruff check` + `ruff format --check` clean
- [x] `basedpyright` — 0 errors / 0 warnings on changed files
- [ ] Full ingestion run against a Databricks 16.2+ warehouse (query-count + runtime drop)
- [ ] Full ingestion run against a Databricks < 16.2 warehouse (legacy fallback)

----
## Summary by Gitar

- **Tag ingestion improvements:**
  - Refactored tag handling to support key-only tags using a new `DATABRICKS_VALUELESS_CLASSIFICATION` mapping.
  - Updated `populate_tags_cache` to store optional tag values as `None` instead of a `DEFAULT_TAG_VALUE` placeholder.
  - Added `_ometa_tag_call_args` helper to standardize classification logic and ensure graceful handling of empty tag names.


ingestion/src/metadata/ingestion/source/database/databricks/metadata.py
```diff
@@ -240,6 +255,104 @@ def _apply_nested_descriptions(
...
+        key = _json_type_to_sql_string(type_node.key_type)
+        value = _json_type_to_sql_string(type_node.value_type)
+        return f"map<{key},{value}>" if key and value else ""
+    if name == "decimal" and type_node.precision is not None:
+        if type_node.scale is not None:
+            return f"decimal({type_node.precision},{type_node.scale})"
+        return f"decimal({type_node.precision})"
+    if name in {"varchar", "char"} and type_node.length is not None:
+        return f"{name}({type_node.length})"
+    return name
```
IceS2, line 299: Wouldn't make sense to do the other way around and parse the json since it is more structured and the new API from Databricks?

ulixius9: if I go that route I would have to override this common [function](https://github.com/open-metadata/OpenMetadata/blob/a321beb7570432f829c2e5660adc818e77f48d1c/ingestion/src/metadata/ingestion/source/database/sql_column_handler.py#L247) for processing column types 

i found this approach easier and less risker compared to that 😅 let me know what you think

IceS2 (approved):

PR #28311, opened 2026-05-20
fix(reindex): batch-prefetch upstream lineage to stop Hikari connection leak

harshach:
### Describe your changes:

Reindex's doc-build executor runs 50 virtual workers that each call `SearchIndex.getLineageData -> EntityRelationshipDAO.findFrom` per entity. On JDK 21 + HikariCP 7's synchronized borrow path, those virtual threads get pinned to their carriers and can stall for >60s while holding a connection — tripping `ProxyLeakTask` and freezing the reindex (`WARN ... Apparent connection leak detected ... at SearchIndex.getLineageData ... on thread reindex-os-doc-build-N`).

This PR moves the lineage lookup off the per-entity doc-build hot path. Pre-fetch happens **once per batch in the reindex processor stage** (`PartitionWorker.writeToSink`, `SearchIndexRetryWorker`, `SearchRepository.updateEntitiesBulk`) via `ReindexingUtil.populateDocBuildContext(...)`, which stuffs a `Map<UUID, DocBuildContext>` into `contextData` under `BulkSink.DOC_BUILD_CONTEXT_KEY`. The sinks remain **transport-only** — they read the map by key and pass the per-entity `DocBuildContext` to `SearchIndex.buildSearchIndexDoc(ctx)` with zero awareness of what `ctx` carries. `LineageIndex.applyLineageFields(doc, ctx)` reads `ctx.prefetchedUpstreamLineage()` when present and falls back to the legacy per-entity DB query otherwise — so live CRUD and any other caller of `getLineageData` are unaffected.

Also codifies the project's single-return-per-method rule and the `nullOrEmpty(...)` / Java 21 `getFirst()` idioms in `CLAUDE.md`.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

- **Hot path before:** sinks submit N `addEntity` tasks to `DOC_BUILD_EXECUTOR` (50 virtual threads). Each task calls `buildSearchIndexDoc -> LineageIndex.applyLineageFields -> SearchIndex.getLineageData -> findFrom`, holding a Hikari connection per entity. Under load these threads get pinned by HikariCP's synchronized borrow and stall >60s.
- **Hot path after:** the **processor stage** (everything that builds batches before invoking `sink.write`) calls `ReindexingUtil.populateDocBuildContext(contextData, entityType, entities)` once. That helper:
  - probes `Entity.buildSearchIndex(entityType, null) instanceof LineageIndex` (result memoized per JVM in a `ConcurrentHashMap` so the probe runs at most once per type)
  - calls `SearchIndex.prefetchUpstreamLineage(entities)` — one `findFromBatch` + one `getEntityReferencesByIds` per upstream entity type
  - wraps each entry in a `DocBuildContext` record (the doc-build "carry-on" DTO, today carrying only `prefetchedUpstreamLineage`, tomorrow other batch-prefetched fields like tags/owners)
  - stuffs `Map<UUID, DocBuildContext>` into `contextData[BulkSink.DOC_BUILD_CONTEXT_KEY]`
- **Sinks**: zero lineage knowledge. `addEntity` does `docBuildContexts.getOrDefault(entity.getId(), DocBuildContext.empty())` and hands the result to `buildSearchIndexDoc(ctx)`. The sink import list lost `LineageIndex`, `EsLineageData`, and `SearchIndex.prefetchLineageIfSupported`.
- **No ThreadLocal**: an earlier iteration used a `LineagePrefetchContext` ThreadLocal; that was removed in favour of the explicit `DocBuildContext` parameter so the dependency is visible in the signature and there's no set/clear discipline at every call site.
- **Alternative considered:** switching `DOC_BUILD_EXECUTOR` to platform threads. Rejected because virtual threads are correct for I/O-bound work like reindex — the real fix is to stop doing per-entity DB I/O on those threads, not to abandon virtual threads.

#
#### Use cases covered
- Reindex of an entity type whose index implements `LineageIndex` (Table, Dashboard, Pipeline, Metric, etc.) batch-prefetches lineage once per batch — sinks see a populated `DocBuildContext` map and skip the per-entity DB call.
- Reindex of entities whose index does not implement `LineageIndex` skips prefetch entirely (the support probe is cached, so subsequent batches of the same type are O(1)).
- Prefetch failure (DB exception, missing upstream ref, invalid JSON, malformed UUID) leaves the context map unpopulated → sinks pass `DocBuildContext.empty()` → doc-build falls through to the original per-entity DB query path.
- Live CRUD / non-reindex callers of `SearchIndex.getLineageData` unaffected (they use the no-arg `buildSearchIndexDoc()` overload, which delegates with `DocBuildContext.empty()`).

#### Unit tests
- Files added/updated:
  - `openmetadata-service/src/test/java/org/openmetadata/service/search/indexes/DocBuildContextTest.java` (new, 3 tests)
  - `openmetadata-service/src/test/java/org/openmetadata/service/search/indexes/SearchIndexPrefetchTest.java` (new, 18 tests covering the prefetch helpers, malformed UUID skip, partial-failure paths, and the legacy `getLineageDataFromRefs` JSON-failure catch)
  - `openmetadata-service/src/test/java/org/openmetadata/service/search/indexes/LineageIndexTest.java` (added DocBuildContext-bound branches)
  - `openmetadata-service/src/test/java/org/openmetadata/service/workflows/searchIndex/ReindexingUtilDocBuildContextTest.java` (new, 2 tests for `populateDocBuildContext`)
  - `openmetadata-service/src/test/java/org/openmetadata/service/apps/bundles/searchIndex/OpenSearchBulkSinkBehaviorTest.java` + `ElasticSearchBulkSinkBehaviorTest.java` (sink behavior tests updated for the new transport-only contract)
- Coverage: JaCoCo (`-P static-code-analysis`) reports 100% instruction + branch coverage on every new method (`DocBuildContext.*`, `SearchIndex.prefetchLineageIfSupported(String, List)`, `supportsLineagePrefetch(String)`, `probeLineagePrefetchSupport(String)`, `prefetchUpstreamLineage` + its 5 private helpers, `parseUuidOrNull`, the legacy `getLineageDataFromRefs` catch arm, `ReindexingUtil.populateDocBuildContext`, `BulkSink.DOC_BUILD_CONTEXT_KEY`). 61 tests across the changed area pass.

- [x] Not applicable (no new API endpoint; existing reindex IT coverage exercises the path).

- [x] Not applicable.

#### Playwright (UI) tests
- [x] Not applicable.

#### Manual testing performed
1. `mvn -pl openmetadata-service test -Dtest='ReindexingUtilDocBuildContextTest,DocBuildContextTest,SearchIndexPrefetchTest,LineageIndexTest,OpenSearchBulkSink*,ElasticSearchBulkSink*'` → 61 pass, 0 fail.
2. `mvn -pl openmetadata-service verify -P static-code-analysis ...` → JaCoCo confirms 100% on every new method.
3. `mvn -pl openmetadata-service spotless:apply` → clean.

#

Not applicable.

#
### Checklist:

----
## Summary by Gitar

- **Refactored vector embedding enrichment:**
  - Replaced `existingFingerprints` map with a batch-fetched `existingEmbeddingsById` map in `OpenSearchBulkSink` to enable embedding reuse.
  - Added `canReuseCachedEmbedding` guard to safely splice pre-computed embedding fields directly into the search document JSON when fingerprints match.
- **Improved enrichment efficiency:**
  - Transitioned from `Map<String, Object>` parsing to `Jackson` `ObjectNode` manipulation for document enrichment to reduce serialization overhead.
  - Centralized embedding generation and splicing logic to support fallback to regeneration if cached data is missing or invalid.
- **Updated unit tests:**
  - Added comprehensive test coverage in `OpenSearchBulkSinkBehaviorTest` for reuse logic, cache misses, and defensive handling of malformed cached documents.


mohityadav766 (commented):
**1. ThreadLocal handoff between the sink and `LineageIndex`**

Reading `LineageIndex.applyLineageFields` in isolation, the `if (prefetched != null) ... else DB` branch has no visible link to who sets `prefetched` — it's a coupling hidden behind a static. Tests have to remember `LineagePrefetchContext.clear()` in `@AfterEach`, and the `try { setUpstream } finally { clear }` discipline at every `addEntity` callsite is bookkeeping that's easy to drop in a future refactor.

Could we thread the prefetched lineage through an explicit parameter on `buildSearchIndexDoc` / `applyLineageFields` instead? The dependency becomes visible in the signature, the clear-discipline goes away, and tests construct a context object rather than poking a static.

**2. Bulk sinks taking on lineage semantics**

After this change both sinks import `LineageIndex`, `LineagePrefetchContext`, `SearchIndex.prefetchUpstreamLineage`, and `EsLineageData` — and the `prefetchLineageIfSupported` / `supportsLineagePrefetch` block is duplicated in both. The sinks have been transport-only until now; adding lineage semantics here means two places to keep in sync, and any future cross-entity prefetch (tags, owners, etc.) would bolt on the same way.

Could the prefetch live in the reindex source/processor stage upstream of the sink, with an enriched record handed down? `LineageIndex` then reads from a field on the entity (or a sibling DTO) and the sink stays dumb.

**3. `supportsLineagePrefetch` constructs an index instance just to check a marker interface**

```java
Object index = Entity.buildSearchIndex(Entity.getEntityTypeFromObject(probe), probe);
supported = index instanceof LineageIndex;
```

A type-level check — something like `LineageIndex.class.isAssignableFrom(Entity.getSearchIndexClass(entityType))` — avoids the throwaway construction and the `try/catch` around it. Also, this only probes `entities.getFirst()`, which assumes batches are type-homogeneous; worth either asserting that or checking per-entity if mixed batches are possible.

harshach:
@mohityadav766 thanks for the review — addressed in 6171bf018d:

**#1 (ThreadLocal → explicit param):** Done. `LineagePrefetchContext` is gone. New `DocBuildContext` record is threaded through `SearchIndex.buildSearchIndexDoc(ctx)` and `LineageIndex.applyLineageFields(doc, ctx)`. No more set/clear discipline at every `addEntity` call site, no hidden static state, and the dependency is visible in the signature.

**#2 (sinks taking on lineage semantics — push prefetch upstream into the source/processor stage):** Valid architectural concern, but I'd like to take it as a follow-up rather than expanding the scope of this PR. It touches the source/reader stage and would naturally pair with similar prefetches for tags / owners / domains. I'll open a tracking issue.

**#3 (`supportsLineagePrefetch` constructs a throwaway index):** Done. The sinks now pass `entityType` (which they already have from `contextData`) to `SearchIndex.prefetchLineageIfSupported(entityType, entities)`. The probe uses `Entity.buildSearchIndex(entityType, null)` — same null-entity pattern already used by `SearchIndexFactory#getReindexFieldsFor` — and the marker check happens on that probe instance. Avoids the throwaway construction over a real entity and the `Entity.getEntityTypeFromObject` round-trip. Batches are entity-type-homogeneous (the sink's `entityType` comes from `contextData`), so the single probe is correct.

Tests: 59 pass; JaCoCo confirms 100% on every new method.

harshach:
@mohityadav766 follow-up on #2 — done in 04dc21fff8 rather than deferring.

Sinks are now transport-only:
- Dropped imports for `LineageIndex`, `EsLineageData`, and `SearchIndex.prefetchLineageIfSupported` from both sinks.
- `addEntity` does nothing more than \`docBuildContexts.getOrDefault(entity.getId(), DocBuildContext.empty())\` before \`buildSearchIndexDoc(ctx)\` — it has zero awareness of what \`ctx\` carries.

The prefetch now lives in the processor stage:
- New `BulkSink.DOC_BUILD_CONTEXT_KEY` — generic \`Map<UUID, DocBuildContext>\` channel through \`contextData\`.
- New `ReindexingUtil.populateDocBuildContext(contextData, entityType, entities)` — single place that probes for lineage support and wraps the result. Called by `PartitionWorker.writeToSink`, `SearchIndexRetryWorker`, and `SearchRepository.updateEntitiesBulk` before each `sink.write`.
- When future cross-entity prefetches show up (tags / owners / domains), they bolt into `ReindexingUtil.populateDocBuildContext` and `DocBuildContext`, not into the sinks.

Tests: 61 pass; JaCoCo 100% on the new \`populateDocBuildContext\`. All three review points (#1 explicit param, #2 transport-only sinks, #3 type-level marker check) are now addressed in this PR.

mohityadav766 (approved):

Issue #2496, opened 2022-01-30
Fix height of selected field's number


mithmatt:
The height of selected field's number is slightly smaller than the non-selected ones.

It is hard to notice. However, if you select and deselect a few times you'll see the line moving up and down.

We should fix the height such that it is same regardless of it being selected or not

![image](https://user-images.githubusercontent.com/5906984/151686275-8b03668f-5226-4223-a561-2c0fe6444019.png)

mithmatt:
@shahsank3t @Sachin-chaurasiya  Could this be a good first issue?

shahsank3t:
Yes @mithmatt it's a good first issue. Added the tag.

deweybat:
Hi, could I take on this issue? Thanks

mithmatt:
Thanks for showing interest in taking this on @deweybat 

Assigned to you

Rounak-28:
Is this issue still open?

Sachin-chaurasiya:
Hello @Rounak-28, it's fixed.

PR #28313, opened 2026-05-20
Improve Vector Embedding recalculation logic

mohityadav766:
Fixes #   

  entity's content fingerprint is unchanged, instead of regenerating the vector via
  the embedding provider (Bedrock / DJL). Embedding generation is the dominant cost
  of a full reindex when a paid provider is wired up — for entities whose semantic
  payload hasn't changed, the prior vector is just as valid as a freshly generated
  one.

  The flow:

  1. Before each batch, OpenSearchBulkSink pre-fetches _source (fingerprint +
  entity ids, using a single typed mget against OpenSearchVectorService. 
  2. In enrichWithEmbedding, for each entity:
    - Compute currentFingerprint from the entity.
    - If the cached fingerprint matches and the cached embedding array is non-empty,
  splice the cached embedding fields into the new doc. 
    - Otherwise, call the embedding provider as before.
  3. Reindex context is consulted so that during a recreateIndex, the fingerprint
  fetch reads from getOriginalIndex(entityType) (the pre-recreate active index)
  rather than the empty staged index.
 

  Type of change:

  - Improvement

  High-level design:
  
  Cache-source choice. During a full recreate, the staged index starts empty, so we
  read the cached embedding payload from ReindexContext.getOriginalIndex(entityType)
  — the pre-recreate active index. Outside a recreate, the canonical index name is
  used directly. The pre-existing getStagedIndex(...) lookup was wrong for this
  purpose: it would always return an empty hit on a recreate and force regeneration.

  Typed client + tree-model parse. The batch lookup uses the OpenSearch Java client's
   typed client.mget(...) with sourceIncludes(...) filtering to just the fields we
  need (fingerprint, embedding, textToLLMContext, textToEmbed, chunkIndex, 
  chunkCount, parentId) — no raw JSON string concatenation. The JsonData response is
  deserialized to Jackson JsonNode via a JacksonJsonpMapper so downstream code can
  use the type-tolerant tree API (.path("fingerprint").asText(null),
  .path("embedding").isArray()). This avoids the class of ClassCastException that bit
   us during an earlier iteration when the default Jakarta JSON-P provider returned
  JsonStringImpl wrappers instead of plain Strings.

  Splice rather than write a raw doc. When the fingerprint matches, the cached
  _source (an ObjectNode) is merged into the freshly built entity ObjectNode via
  setAll. The cached embedding survives the round-trip cleanly because Jackson reads
  numeric arrays as double (same precision the JSON carried) and re-serializes them
  as JSON arrays — no precision loss, same on-wire shape, OpenSearch indexes it back
  into the knn_vector field unchanged.
  
  Reject the cache defensively. canReuseCachedEmbedding requires the cached node to
  be an object, the fingerprint field to be present as text, the fingerprint to
  match, and the embedding field to be a non-empty array. A missing or malformed
  cache entry simply falls through to regeneration — never throws, never indexes a
  doc without a vector.

  Alternatives considered:

  - Per-doc copy via OpenSearch reindex API (Collate's approach in PR #2496): fits
  Collate's separate vector_search_index (one doc per text chunk, keyed by
  parent_id). In OSS we have a single entity doc with embedding fields merged in, so
  a field-level splice is the right shape — copying whole docs is over-engineered for
  - Replacing getExistingFingerprintsBatch instead of adding 
  getExistingEmbeddingsBatch: would have churned the existing caller surface.
  Additive method is safer.
  - Map<String, Object> instead of JsonNode in the cache map: tried, hit the
  JsonStringImpl cast issue. JsonNode is type-tolerant and removes a whole class of
  bugs.
  
  Backward compatibility: no schema changes, no migration, no behavior change for
  first-time indexing (cache miss → recompute, same as before). The change only
  affects the cost profile of subsequent reindexes when content is unchanged.

  

  Tests:
  
  Use cases covered

  - Recreate reindex where entity content is unchanged: embedding is reused from the
  original index, no provider call made.
  - Recreate reindex where entity content changed (fingerprint mismatch): embedding
  is regenerated via the provider, fresh values land in the new index.
  - First-time index (no cached entry): falls through to regenerate, no change in
  behavior vs. prior code.
  - Cache hit with malformed payload (no embedding field / wrong type): falls through
   to regenerate, doc is never indexed without a vector.
  
  Unit tests

  - I added unit tests for the new/changed logic.
  - Files added/updated: 
    - openmetadata-service/src/test/java/org/openmetadata/service/apps/bundles/search
  Index/OpenSearchBulkSinkBehaviorTest.java
        - enrichWithEmbeddingReusesCachedFieldsWhenFingerprintMatches — asserts
  vectorService.generateEmbeddingFields(...) is never called when fingerprint
  matches, and that the cached embedding/textToEmbed/fingerprint fields end up in the
   resulting JSON alongside the entity's own fields.
        - enrichWithEmbeddingReusesCachedFieldsWhenFingerprintMatches — asserts vectorService.generateEmbeddingFields(...) is never called when fingerprint matches, and that the
   cached embedding/textToEmbed/fingerprint fields end up in the resulting JSON alongside the entity's own fields.
      - enrichWithEmbeddingRecomputesWhenFingerprintMismatch — asserts generateEmbeddingFields(...) is called when fingerprints differ and the fresh values (not the stale cached
   ones) land in the result.
  - Run: mvn -pl openmetadata-service test -Dtest=OpenSearchBulkSinkBehaviorTest → 9/9 passing.

  Backend integration tests

  - Not applicable. No new API endpoints; the change is internal to the bulk-sink pipeline. Existing reindex integration tests continue to pass.
  
  Ingestion integration tests

  - Not applicable. No ingestion changes.

        - enrichWithEmbeddingReusesCachedFieldsWhenFingerprintMatches — asserts vectorService.generateEmbeddingFields(...) is never called when fingerprint matches, and that the
   cached embedding/textToEmbed/fingerprint fields end up in the resulting JSON alongside the entity's own fields.
      - enrichWithEmbeddingRecomputesWhenFingerprintMismatch — asserts generateEmbeddingFields(...) is called when fingerprints differ and the fresh values (not the stale cached
   ones) land in the result.
  - Run: mvn -pl openmetadata-service test -Dtest=OpenSearchBulkSinkBehaviorTest → 9/9 passing.

  Backend integration tests

  - Not applicable. No new API endpoints; the change is internal to the bulk-sink pipeline. Existing reindex integration tests continue to pass.

  Ingestion integration tests

  - Not applicable. No ingestion changes.

  Playwright (UI) tests

  - Not applicable. No UI changes.

  Manual testing performed

  1. Started a local stack with embeddings enabled.
  2. Ran a full reindex with recreateIndex=true on a populated catalog; confirmed Bedrock call count dropped to zero for entities unchanged since the last index, and that vector
   search continued to return those entities post-recreate (embeddings carried through).
  3. Modified the description of a sampling of entities, re-ran the reindex; confirmed those specific entities had generateEmbeddingFields invoked (logged) while the rest reused
   cached vectors.
  4. Verified the previously observed ClassCastException: org.glassfish.json.JsonStringImpl cannot be cast to class java.lang.String no longer reproduces under load — the
  tree-model parse handles the source map uniformly.
  

  UI screen recording / screenshots:

  Not applicable.

 
  Checklist:
  
  - I have read the CONTRIBUTING document.
  - My PR title is Fixes <issue-number>: <short explanation> — suggested title: Fixes <issue-number>: reuse cached embeddings during recreate when fingerprint unchanged
  - My PR is linked to a GitHub issue via Fixes #<issue-number> above.
  - I have commented on my code, particularly in hard-to-understand areas.
  - For JSON Schema changes: not applicable, no schema changes.
  - For UI changes: not applicable.
  - I have added tests (unit) and listed them above.

  Improvement-specific:
  - I have added tests around the new logic.
  - For connector/ingestion changes: not applicable.

----
## Summary by Gitar

- **Integration testing:**
  - Updated `VectorEmbeddingIntegrationIT` to test `getExistingEmbeddingsBatch` instead of the deprecated `getExistingFingerprintsBatch`.
  - Verified that cached embedding data (including fingerprint and vector array) is correctly retrieved during integration tests.

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/VectorEmbeddingIntegrationIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/VectorEmbeddingIntegrationIT.java
@@ -242,13 +242,29 @@ class VectorEmbeddingIntegrationIT {
     vectorService.updateEntityEmbedding(entity2, TEST_INDEX);
     Thread.sleep(1000);
 
-    Map<String, String> fingerprints =
-        vectorService.getExistingFingerprintsBatch(
-            TEST_INDEX, List.of(entity1Id.toString(), entity2Id.toString()));
+    Map<String, OpenSearchVectorService.EntityFingerprintInput> currentById =
+        Map.of(
+            entity1Id.toString(),
+            new OpenSearchVectorService.EntityFingerprintInput(
+                entity1.getUpdatedAt(),
+                () -> VectorDocBuilder.computeFingerprintForEntity(entity1)),
+            entity2Id.toString(),
+            new OpenSearchVectorService.EntityFingerprintInput(
+                entity2.getUpdatedAt(),
+                () -> VectorDocBuilder.computeFingerprintForEntity(entity2)));
 
-    assertEquals(2, fingerprints.size(), "Should retrieve fingerprints for both entities");
-    assertNotNull(fingerprints.get(entity1Id.toString()));
-    assertNotNull(fingerprints.get(entity2Id.toString()));
+    Map<String, JsonNode> cachedEmbeddings =
+        vectorService.getExistingEmbeddingsBatch(TEST_INDEX, currentById);
+
+    assertEquals(2, cachedEmbeddings.size(), "Should retrieve cached embeddings for both entities");
+    JsonNode cached1 = cachedEmbeddings.get(entity1Id.toString());
+    JsonNode cached2 = cachedEmbeddings.get(entity2Id.toString());
+    assertNotNull(cached1);
+    assertNotNull(cached2);
+    assertTrue(cached1.path("fingerprint").isTextual());
+    assertTrue(cached2.path("fingerprint").isTextual());
+    assertTrue(cached1.path("embedding").isArray() && !cached1.path("embedding").isEmpty());
+    assertTrue(cached2.path("embedding").isArray() && !cached2.path("embedding").isEmpty());
   }
 
   @Test
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/OpenSearchBulkSink.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/OpenSearchBulkSink.java
@@ -277,22 +277,23 @@ public class OpenSearchBulkSink implements BulkSink {
                 ? (ReindexContext) contextData.get(RECREATE_CONTEXT)
                 : null;
 
-        // Pre-fetch cached embeddings for entities whose fingerprint is unchanged so we can splice
-        // them into the staged doc instead of regenerating (avoids expensive embedding-provider
-        // calls). Two-step over the original index: first an mget for fingerprints only, then —
-        // for IDs whose cached fingerprint matches the entity's current fingerprint — a second
-        // mget that pulls the full embedding source. This keeps large vector payloads off the wire
-        // for entities that will be re-embedded anyway.
+        // Pre-fetch cached embeddings for entities whose state is unchanged so we can splice them
+        // into the staged doc instead of regenerating (avoids expensive embedding-provider calls).
+        // The service-layer two-step keeps large vector payloads off the wire for entities that
+        // will be re-embedded anyway, and uses the entity's `updatedAt` as a fast-path: when it
+        // matches the cached value the fingerprint supplier is never invoked.
         Map<String, JsonNode> existingEmbeddingsById = Collections.emptyMap();
         if (embeddingsEnabled) {
-          Map<String, String> currentFingerprintsById = new HashMap<>(entityInterfaces.size());
+          Map<String, OpenSearchVectorService.EntityFingerprintInput> currentById =
+              new HashMap<>(entityInterfaces.size());
           for (EntityInterface e : entityInterfaces) {
-            currentFingerprintsById.put(
-                e.getId().toString(), VectorDocBuilder.computeFingerprintForEntity(e));
+            currentById.put(
+                e.getId().toString(),
+                new OpenSearchVectorService.EntityFingerprintInput(
+                    e.getUpdatedAt(), () -> VectorDocBuilder.computeFingerprintForEntity(e)));
           }
           existingEmbeddingsById =
-              fetchExistingEmbeddings(
-                  entityInterfaces, currentFingerprintsById, indexName, reindexContext);
+              fetchExistingEmbeddings(entityInterfaces, currentById, indexName, reindexContext);
         }
 
         // Add entities to search index in parallel
@@ -833,6 +834,9 @@ public class OpenSearchBulkSink implements BulkSink {
 
       JsonNode cached = existingEmbeddingsById.get(entity.getId().toString());
       if (canReuseCachedEmbedding(cached)) {
+        // Splices chunkIndex/chunkCount/parentId along with embedding — safe because the
+        // service-layer pre-filter only admits entries whose state matches (same fingerprint or
+        // same updatedAt), and fingerprint covers the body text that determines chunk count.
         doc.setAll((ObjectNode) cached);
       } else {
         Map<String, Object> embeddingFields = vectorService.generateEmbeddingFields(entity);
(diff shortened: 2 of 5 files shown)


harshach:
Nice scoping — the recreate-path embedding reuse is a real win, and the typed `mget` with `sourceIncludes` is cleaner than the string-built `_search`. A few items I'd suggest tightening before merge:

### Code hygiene

- **Dead code**: `OpenSearchVectorService.getExistingFingerprintsBatch` (the old `_search`-based variant) no longer has any callers after this PR. Worth deleting in the same change — leaving both methods around will make the next contributor unsure which to use. Same for `getExistingFingerprint` if it's also unreferenced.
- **Misleading names**: the helper in `OpenSearchBulkSink` is still called `fetchExistingFingerprints` and the local variables are `existingFingerprints` / `finalFingerprints`, but the map values are now full embedding `JsonNode`s, not fingerprint strings. Suggest renaming to `fetchExistingEmbeddings` / `existingEmbeddings` for consistency with the renamed `getExistingEmbeddingsBatch`.

### Performance / correctness

- **`updatedAt` fast-path**: `canReuseCachedEmbedding` currently rebuilds metaLight + body and computes MD5 for every entity in every batch, even when the entity hasn't changed since last index. `entity.getUpdatedAt()` is already in the ES doc (it's serialized as part of every entity); if you add `updatedAt` to `EMBEDDING_SOURCE_FIELDS` and compare it before computing the fingerprint, the common case (unchanged entity) skips all text-building. Cheap to add, and on large reindexes where most entities haven't changed it's a meaningful saving — for tables with hundreds of columns, `buildBodyText` builds a non-trivial string per call.
- **Sort tag / owner / domain lists in the fingerprint inputs**: `buildMetaLightText` (`VectorDocBuilder.java:190-310`) joins these in insertion order, so the same set in a different order yields a different fingerprint → spurious regenerations. Not introduced by this PR, but the more aggressive reuse here makes the false-mismatch case more annoying. Easy follow-up.
- **`text.getBytes()` without explicit charset**: `TextChunkManager.computeFingerprint` at line 56 uses the platform default charset. Cross-platform fingerprint stability isn't guaranteed today — worth pinning to `StandardCharsets.UTF_8` while you're in the area.

### Tests

The two new tests cover the happy paths (match → reuse, mismatch → regenerate). A few branches in `canReuseCachedEmbedding` aren't yet exercised:

- Cache miss (`cached == null`) — first-time index of a new entity.
- Cached entry present but `embedding` array missing or empty (e.g., an entity indexed before embeddings were enabled).
- Cached entry present but `fingerprint` field absent (returns `null` from `.asText(null)`).
- Non-object cached node (`!cached.isObject()`).

Each currently falls through to regenerate, which is the correct behavior, but it'd be good to lock that in.

For end-to-end confidence on the recreate path, an integration test that:
1. Indexes a couple of entities and captures their `embedding` arrays from the active index.
2. Triggers a recreate.
3. Asserts the staged index's `embedding` arrays match the originals byte-for-byte for unchanged entities, and differ for entities whose content was modified between the two runs.

...would lock the behavior down. The manual-testing notes in the PR description are reassuring but won't catch regressions.

### Two minor things worth a one-line comment in the code

- The splice carries `chunkCount` / `chunkIndex` / `parentId` over from the cached source. That's consistent only because identical fingerprint ⇒ identical body ⇒ identical chunk count — i.e. it relies on the fingerprint covering the body text. A one-liner near the `setAll` call would make that invariant explicit for future readers.
- `parentId` is in `EMBEDDING_SOURCE_FIELDS`, so it gets spliced too. Harmless on the entity-level index where `parentId == entityId`, but worth confirming it doesn't collide with anything on the column-level path (`indexTableColumns` at `OpenSearchBulkSink.java:305-319`).

### Out of scope, but flagging

A separate gap I noticed while reviewing: the fingerprint inputs themselves still miss column-level tags (`columnsToString` at `VectorDocBuilder.java:626-639` joins name + description only) and don't cover nested sub-fields on Topic / Pipeline / MlModel / Container / SearchIndex / APIEndpoint / DashboardDataModel — `buildBodyText` only has the `instanceof Table` branch. So editing a column tag or a Topic schema field description today doesn't bust the fingerprint, and with this PR's more aggressive reuse, the stale embedding survives the recreate too. Not something to fix here, but a follow-up PR to extend `buildBodyText` to cover those entity types would close the loop.

Overall LGTM on the core change once the dead code is removed and the variable renames are in.

harshach (approved):

pmbrull (approved):

PR #28329, opened 2026-05-21
fix(ingestion): chunk Snowflake ACCESS_HISTORY lineage by day to avoid query timeouts

ulixius9:
## Problem

Running Snowflake lineage with a large `queryLogDuration` (e.g. 180 days) on the ACCESS_HISTORY path fails with:

```
000604 (57014): SQL execution was cancelled by the client due to a timeout
```

`_yield_combined_access_history` rendered the entire window as **one** `ACCESS_HISTORY × QUERY_HISTORY` query with multiple `LATERAL FLATTEN`s over VARIANT arrays. Over a long window that single statement exceeds the client `network_timeout` / server limits and Snowflake cancels it. Because the whole query sat in one `try/except`, the run then emitted **0 edges**.

The `usage` connector already avoids this by scanning the window day-by-day — the lineage ACCESS_HISTORY path did not.

### Timeout fix — chunk the window
- **Chunk into 2-day slices** (`ACCESS_HISTORY_CHUNK_DAYS = 2`) via `_iter_lineage_date_windows()`, mirroring `SnowflakeUsageSource.yield_table_queries`. Each chunk is a small, bounded query.
- **Per-window error isolation**: the `try/except` moves into `_fetch_access_history_rows()` so one slow/failing day logs a warning (with the window bounds) and the run continues, instead of zeroing out the whole job. Counters accumulate across windows for a single summary log.

### Lineage completeness — LEFT JOIN QUERY_HISTORY
ACCESS_HISTORY is the authoritative lineage source (`DIRECT_OBJECTS_ACCESSED` / `OBJECTS_MODIFIED` and column-level `directSources`); QUERY_HISTORY only contributes the query text. The previous `INNER JOIN` dropped an entire edge whenever the matching QUERY_HISTORY row was absent or filtered out.

- Switch to **`LEFT JOIN`** (ACCESS_HISTORY preserved) so an edge survives with a null query text rather than vanishing.
- The `qh.START_TIME` time-prune moves into the **`ON` clause** — it bounds the QUERY_HISTORY scan per chunk (otherwise each daily query resolves its `QUERY_ID` join against an unpruned QUERY_HISTORY) without being able to eliminate ACCESS_HISTORY rows. This also removes the boundary-skew miss risk an `INNER JOIN` time predicate would carry.
- `EXECUTION_STATUS = 'SUCCESS'` sits in the `ON` clause too, so every `qh` predicate only gates enrichment and can never drop an ACCESS_HISTORY edge. The CTE `WHERE` carries only the authoritative `ah.QUERY_START_TIME` bound.
- **Drop the dbt / OpenMetadata `NOT LIKE` noise filters**: OM runs only read-only queries (no `OBJECTS_MODIFIED`, so structurally excluded), and dbt lineage from ACCESS_HISTORY is Snowflake-computed (not parsed) and merges idempotently with dbt-connector lineage.

### `filterCondition` now scopes by table, not query history
On the ACCESS_HISTORY path, `sourceConfig.filterCondition` renders as a trailing `WHERE (...)` on the **final edge result** instead of a predicate on QUERY_HISTORY. This lets you scope lineage by the database/schema of the tables involved via the edge columns `UPSTREAM_TABLE` / `DOWNSTREAM_TABLE` (Snowflake `DB.SCHEMA.TABLE` FQNs) — e.g. `DOWNSTREAM_TABLE LIKE 'MYDB.%'` or `UPSTREAM_TABLE ILIKE 'MYDB.MYSCHEMA.%'`. Note: `qh.`-qualified predicates are no longer valid on this path.

COPY_HISTORY is left as a single query — it reads only the much smaller COPY_HISTORY view and is not the bottleneck.

### Observability
- When an edge is skipped because a table is not found in OpenMetadata, log at DEBUG which side (upstream/downstream) failed and the Snowflake table name — making the `skipped N (unresolvable ...)` summary actionable.

## Test plan

- Unit tests in `test_snowflake_access_history_lineage.py` (38 passing), including:
  - `test_combined_lineage_left_joins_query_history_to_keep_edges_without_text`
  - `test_combined_lineage_sql_prunes_query_history_by_date`
  - `test_access_history_chunks_window_by_day` (3-day span → 3 queries with distinct bounds)
  - `test_access_history_window_failure_does_not_abort_run`
  - extended `test_table_edges_skip_when_either_side_unresolvable` to assert the new debug log
- `ruff check` + `ruff format --check` clean on changed files.

----
## Summary by Gitar

- **Generated types:**
  - Updated TypeScript definitions in `openmetadata-ui` to reflect recent changes in service connection and ingestion pipeline schemas.


ingestion/src/metadata/ingestion/source/database/snowflake/lineage.py
```diff
@@ -237,48 +242,77 @@ def _yield_access_history_lineage(self) -> Iterable[Either[AddLineageRequest]]:
...
+                if row.query_text:
+                    emitted_with_sql += 1
+                yield Either(right=edge)  # pyright: ignore[reportCallIssue]
+        logger.info(
+            "ACCESS_HISTORY lineage: emitted %d edges (%d with SQL text), "
+            "skipped %d (unresolvable downstream/upstream tables)",
+            emitted,
+            emitted_with_sql,
+            skipped,
+        )
```
IceS2, line 307: should we have a try/except within the for to avoid skipping the whole history if one fails?

IceS2 (dismissed):

Rohit0301 (approved):

mohittilala (approved):

PR #28380, opened 2026-05-22
feat(ingestion): incremental metadata extraction for Unity Catalog

ulixius9:
## Description

Adds incremental metadata extraction to the Unity Catalog connector, matching the existing Snowflake/BigQuery pattern. After a first full run, only tables changed since the last successful run are processed, and dropped tables are detected explicitly.

- **Changed tables**: server-side filter on `information_schema.tables.last_altered`, then each changed table is fetched individually via the Databricks SDK (`tables.get`) — avoids enumerating the whole catalog every run.
- **Deleted tables**: detected from `system.access.audit` `deleteTable` events (`request_params.full_name_arg`). Degrades gracefully (warns and skips delete detection) when the `system.access` schema is not available.
- A table present in **both** the changed and deleted sets (dropped and recreated within the window) is kept, not deleted — `information_schema` only lists tables that currently exist.
- Reuses the shared `IncrementalConfig` framework; **no JSON schema change** (the `incremental` config already applies to every database connector via `sourceConfig.config.incremental`).

## Type of change

- [x] New feature

## How was this tested

- **Unit**: new `test_unitycatalog_incremental.py` (17 cases) covering the processor's parsing and graceful degradation, the incremental vs. full discovery path, delete handling, the dropped-and-recreated exclusion, the `mark_tables_as_deleted` incremental/full branches, and `create()` wiring. Full Unity Catalog + incremental suite is green (42 tests).
- **Live**: validated end-to-end against a real Unity Catalog workspace + OpenMetadata server — change detection, audit-log delete detection, the recreate edge case, and an incremental workflow run that processed only the changed table while leaving unchanged tables intact (none wrongly deleted).

----
## Summary by Gitar

- **Security & Reliability:**
  - Added regex validation for catalog names in `UnityCatalogIncrementalTableProcessor` to prevent SQL injection.
  - Refined audit log query to use `substring_index` for exact catalog matching instead of `LIKE` wildcards.
- **Robustness:**
  - Added `pyright` ignores for type-checking to stabilize the Unity Catalog connector build.
  - Improved `_process_table` exception handling to safely report errors when table objects lack names.

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/unitycatalog/incremental_table_processor.py
+++ b/ingestion/src/metadata/ingestion/source/database/unitycatalog/incremental_table_processor.py
@@ -1,0 +1,116 @@
+#  Copyright 2025 Collate
+#  Licensed under the Collate Community License, Version 1.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/LICENSE
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""
+Incremental Processor for Unity Catalog.
+
+Detects, for a single catalog and since a given watermark:
+  - changed tables, via `information_schema.tables.last_altered`
+  - deleted tables, via the `system.access.audit` `deleteTable` events
+
+Both queries degrade gracefully: a failure on either leaves that map empty and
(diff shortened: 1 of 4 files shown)


ingestion/src/metadata/ingestion/source/database/unitycatalog/incremental_table_processor.py
```diff
@@ -0,0 +1,132 @@
...
+
+from metadata.ingestion.source.database.unitycatalog.queries import (
+    UNITY_CATALOG_GET_CHANGED_TABLES,
+    UNITY_CATALOG_GET_DELETED_TABLES,
+)
+from metadata.utils.logger import ingestion_logger
+
+logger = ingestion_logger()
+
+SchemaToTables = Dict[str, Set[str]]  # noqa: UP006
```
IceS2, line 39: On # noqa: UP

We should aim to write new code with the new style instead of adding those comments

IceS2 (approved):

PR #28426, opened 2026-05-26
Add temporal lineage traversal support

harshach:
I worked on temporal lineage traversal because lineage views need to show edge state across historical time windows without disrupting the graph viewport.

High-level design:

Adds temporal fields to lineage APIs, cache keys, search mappings, ES/OS graph traversal, and lineage change events, then wires the UI time filter so lineage can be queried by range or point in time while preserving the rendered graph when switching between Lineage and Impact Analysis.

**Tests:**

Use cases covered

Historical lineage traversal returns edges active in the selected time window.
Lineage viewport centers/zooms on first load and does not redraw unnecessarily when switching Impact Analysis back to Lineage.
Sample ingestion data includes older and moved lineage edges for temporal traversal validation.
Unit tests

 I added unit tests for the new/changed logic.
Files added/updated: LineageProvider.test.tsx, EntityLineageUtils.test.tsx, lineageAPI.test.ts, LineageCacheKeyTest.java.
Backend integration tests

 I added integration tests for lineage temporal traversal.
Files added/updated: LineageResourceIT.java, OpenLineageLineageResolutionIT.java.
Ingestion integration tests

 Not applicable; sample ingestion lineage data was updated.
Playwright (UI) tests

 Not applicable; focused Jest coverage was added for the UI behavior.
Manual testing performed

Seeded local sample lineage edges through the REST API and verified different time windows return different active edges.
Verified last 7, 30, 90, and 180 day lineage windows for the ecommerce sample lineage graph.

https://github.com/user-attachments/assets/ee8f1d03-0269-4e3c-a739-89c734c57489

----
## Summary by Gitar

- **Backend lineage persistence:**
  - Integrated `applyTemporalFields` in `LineageRepository` to maintain `createdAt`/`createdBy` and track `updatedAt`/`updatedBy` across lineage re-emissions.
  - Added `decideEventType` to trigger `ENTITY_LINEAGE_UPDATED` events when lineage metadata or pipeline details change.
- **Search & API enhancements:**
  - Updated `ESLineageGraphBuilder` and `OSLineageGraphBuilder` to support `startTime` and `endTime` filters in graph traversal logic.
  - Expose temporal filters in `LineageResource` endpoints, including export and pagination APIs.
- **Temporal ingestion:**
  - Updated `SampleDataSource` to support ingestion of temporal fields (`createdAt`, `updatedAt`) and `TempLineageTable` metadata.
- **Audit logging:**
  - Added `AuditLogRepositoryTest` to verify that lineage lifecycle changes are correctly persisted as audit events.
- **UI components:**
  - Added `LineageTimeFilter` to UI to allow users to toggle range or point-in-time traversal filters.
  - Added `EdgeInfoDrawer` fields to display temporal provenance metadata.


chirag-madlani (dismissed):

mohityadav766 (commented):
### Temporal lineage — review notes

Overall the plumbing is clean and consistent: the ES and OS builders mirror each other, the cache key is extended with the window, `validateTemporalBounds` guards `startTime > endTime`, and there's solid IT/unit coverage. One interaction I want to flag, though.

#### Preserve-paths does not extend to the time window

A path-preservation user expects that a node which doesn't itself match the filter is still shown when something in its upstream/downstream qualifies. That holds for the **node-level query filter**, but **not** for the new `startTime`/`endTime` window — and the combination is currently a silent no-op for the temporal dimension.

**1. The time window hard-prunes during traversal.** In `fetchDownstreamNodesRecursively`:

```java
// ESLineageGraphBuilder.java:266-274
List<EsLineageData> upstreamEntities =
    getMatchingUpstreamLineageData(entityMap, hasToFqnMap.keySet(),
        lineageRequest.getStartTime(), lineageRequest.getEndTime());
if (upstreamEntities.isEmpty()) {
  continue;   // node dropped AND not enqueued for the next layer
}
```

For a path `root → A → (edge out of window) → B → (edge in window) → C`: at B the connecting edge `A→B` fails the window, so B is skipped *and never added to the next frontier*. The BFS never explores B's downstream, so edge `B→C` and node C are never discovered — even though `B→C` is inside the window. The same hard-prune pattern exists in the upstream branch and in the depth-count BFS (`getDepthWiseEntityCounts`, `allEntitiesUpToDepth`). An out-of-window edge severs discovery of everything behind it, in both directions.

**2. Preserve-path runs afterward, on an already-time-pruned graph.** Path preservation deliberately strips the *node-level query filter* from the base traversal so the full topology is available, then filters in-memory:

```java
// ESLineageGraphBuilder.java:325-333
SearchLineageRequest unfilteredRequest =
    JsonUtils.deepCopy(lineageRequest, SearchLineageRequest.class)
        .withQueryFilter(getStructuralFilterOnly(lineageRequest.getQueryFilter()));
result = searchLineageWithStrategyInternal(unfilteredRequest);
result = applyInMemoryFiltersWithPathPreservation(result, lineageRequest);
```

The deepCopy overrides only `queryFilter` — `startTime`/`endTime` survive into `unfilteredRequest`. So the "unfiltered" base graph handed to `LineagePathPreserver` is already time-pruned, and the preserver can only trace paths through edges that still exist (`buildParentAdjacencyMap` walks existing edges only). It cannot resurrect B or C.

**Net:** the query filter is *excluded* from the base traversal and applied in post-processing (preserve-able); the time window is *baked into* the base traversal as a hard edge prune (not preserve-able). A user passing `preservePaths=true` + a time window gets no path preservation for the temporal dimension, with no error or signal.

#### Recommendation

Decide the intended semantic and make it explicit:

- **If the window should honor preserve-paths** (keep an out-of-window intermediate so an in-window up/downstream node still shows): move time filtering out of the inline BFS into the same post-processing path as the node filter. Concretely — also strip `startTime`/`endTime` from `unfilteredRequest`, traverse structurally, then in the in-memory step collect endpoints of in-window edges as "matching nodes" and feed them to `LineagePathPreserver.preservePathsWithEdges(...)`, exactly as the column/node filter does.
- **If hard pruning is intended** (show the graph as of the window, severing through stale edges): that's defensible, but it should be documented on the `startTime`/`endTime` schema fields, and it should be explicit that `preservePaths` does not extend to the temporal filter — otherwise the silent interaction is confusing.

#### Smaller notes

- `SearchUtils.edgeMatchesWindow`: legacy edges with both timestamps null always match any window. Reasonable for back-compat, but a window query over mixed legacy + temporal data returns all legacy edges — worth calling out in the API docs.
- Minor directional asymmetry: downstream drops the candidate node entirely on an out-of-window connecting edge (`continue`), while upstream keeps the frontier node and only prunes onward edges. Same path-severing effect, but node retention differs between directions — worth a sanity check.

mohityadav766 (dismissed):

PR #25894, opened 2026-02-16
Task redesign

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

---

----
## Summary by Gitar

- **Test Maintenance:**
  - Marked `ActivityFeed` mention notification tests as `test.fixme` to track resolution of pending issues.
  - Disabled several `WorkflowDefinitionResourceIT` integration tests pending fixes related to the task redesign scope.

<sub>This will update automatically on new commits.</sub>

PR #28451, opened 2026-05-27
fix(workflows): GitHub Sink + event-based workflows silently reject all entities

sonika-shah:
## Fix

Restore the `!` in `FilterEntityImpl.passesExcludedFilter()` so non-JsonLogic filter input passes through correctly. Extract the JSON-filter evaluation into a small `applyJsonFilter` helper so the behavior can be unit-tested.

```diff
- passesJsonFilter =  Boolean.TRUE.equals(RuleEngine.apply(filterLogic, ...))
+ passesJsonFilter = !Boolean.TRUE.equals(RuleEngine.apply(filterLogic, ...))
```

## Why the `!` is required

`RuleEngine.apply()` returns `false` when `filterLogic` cannot be evaluated as JsonLogic — its catch branch is annotated *"falls back to triggering workflow"*. Event-based workflow triggers pass Elasticsearch query DSL through this code path, so the consumer must treat any non-`TRUE` result as "pass through". The `!` provides that semantic.

## Impact

Without the `!`, every event-based workflow with an ES-style filter (Git Sink, EventBased approval, custom user workflows) is filtered out at the trigger gate:

- Trigger workflow ends at `endEvent` in <100ms via the `${!passesFilter}` branch
- The `CallActivity` to the main workflow is skipped
- `ACT_HI_PROCINST` shows only `*Trigger` rows; the main workflow never starts
- No exception or error is surfaced

## Regression source

The `!` was originally added by [#25272](https://github.com/open-metadata/OpenMetadata/pull/25272) ("Show Relevant Exclude Fields in EventBased Workflows") and inadvertently removed by [#25894](https://github.com/open-metadata/OpenMetadata/pull/25894) ("Task redesign") — a 498-file refactor where this one-character change in `FilterEntityImpl.java` (+1/-1) was easy to miss in review.

## Why this wasn't caught by existing tests

- `FilterEntityImplTest` — 17 existing tests, all targeting `passesFieldBasedFilter`. The JSON-filter branch (`passesJsonFilter` / `RuleEngine.apply`) had no coverage.
- Integration tests (`*IT.java`) — verify workflows deploy correctly; they don't assert that a matching event drives the workflow to its terminal side effect.
- Playwright E2E — covers Workflow Builder UI flows; doesn't exercise event-driven dispatch end-to-end.

## Tests added (same PR, existing reflection pattern)

Five new tests in `FilterEntityImplTest` exercise `applyJsonFilter` via reflection — same pattern used by the existing tests for `passesFieldBasedFilter`. Uses the real `RuleEngine` (no mocks).

- `testApplyJsonFilterPassesWhenFilterLogicIsNull`
- `testApplyJsonFilterPassesWhenFilterLogicIsEmpty`
- `testApplyJsonFilterPassesWhenFilterIsNotJsonLogic` — regression test; fails if `!` is removed again
- `testApplyJsonFilterExcludesWhenJsonLogicMatches` — pins the "matching rule excludes" semantic
- `testApplyJsonFilterPassesWhenJsonLogicDoesNotMatch`

## Verification

| Before | After |
|---|---|
| Workflow ends in 50-100ms with no side effect | Workflow runs in 2-6s and produces its side effect |
| Main workflow process never starts | Main workflow runs through `CallActivity → sinkTask → endEvent` |

Test run: `mvn test -pl openmetadata-service -Dtest=FilterEntityImplTest` → **22/22 green** (17 existing + 5 new). Verified end-to-end against a Git Sink workflow with an ES filter on `databaseService` / `table` / `glossaryTerm`.

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/impl/FilterEntityImpl.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/impl/FilterEntityImpl.java
@@ -8,6 +8,7 @@ import static org.openmetadata.service.governance.workflows.elements.triggers.Ev
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


sonika-shah:
Closing — this PR diagnosed the wrong layer. The actual root cause is in the Workflow Builder UI emitting Elasticsearch query DSL for event-based trigger filters, where the schema requires JsonLogic. `FilterEntityImpl`'s behavior on `main` is correct given a properly-formatted JsonLogic filter.

Superseded by #28464, which fixes the UI to emit JsonLogic.

PR #28504, opened 2026-05-28
fix: PG cert migration cast (v1125) + v1130 heal for already-affected clusters

manerow:
## Summary

Certification migration has been silently failing on PostgreSQL since 1.12.5. Two commits — one root-cause cast in v1125, one corrective heal step in v1130 for clusters that already ran the broken v1125.

`v1125.MigrationUtil.migrateCertificationToTagUsage` binds `:metadata` as a `String` (varchar) but the PG column is `json`. PG refuses the auto-cast:
```
BatchUpdateException: column "metadata" is of type json but expression is of type character varying
```
The per-table `try/catch` swallows it as a WARN, `runDataMigration()` returns normally, framework marks `v1125 = DONE` in `SERVER_CHANGE_LOG`. Net effect on every 1.12.5+ PG cluster: `tag_usage` has zero certification rows, `entity.json` still carries the `certification` field, v1125 never re-runs. Grafana confirms the WARN has been firing nightly on PG since at least 2026-04-28.

Migration completion is keyed only on `SERVER_CHANGE_LOG.version` (no checksum on Java code), so editing v1125 alone doesn't reach clusters that already ran it. Two parts:

### Part A — root-cause cast in v1125 (commit `b2bf026aef`)
One-line SQL change adding `:metadata::json` on the PG branch. Covers fresh installs and pre-1.12.5 → 1.13 upgrades (v1125 hasn't executed there yet). MySQL branch unchanged — `INSERT IGNORE` accepts a string into a JSON column.

### Part B — corrective step in v1130 (commit `795e062224`)
Adds `healStuckCertificationOnEntityJson(handle, connType)` to `utils/v1130/MigrationUtil.java`, wired in at the end of `runDataMigration()` on both PG and MySQL `v1130/Migration.java`. Self-contained — no cross-version imports. Same 23 entity tables as v1125, same SELECT filter, same INSERT/UPDATE shape — only difference is the `::json` cast baked in from the start.

**Idempotent**: SELECT only matches rows where `entity.json` still carries `certification` with a non-null `tagFQN`. After the UPDATE strips that field, the row stops matching. `ON CONFLICT … DO NOTHING` (PG) / `INSERT IGNORE` (MySQL) covers the case where `tag_usage` already has the row.

| Cluster state | Heal effect |
|---|---|
| Fresh install of next release | SELECT returns 0 → no-op |
| MySQL 1.12.5+ (v1125 succeeded) | SELECT returns 0 → no-op |
| Pre-1.12.5 → 1.13 (any DB) | v1125 (with cast) migrates everything → SELECT returns 0 → no-op |
| **PG 1.12.5+ (stuck)** | SELECT returns stuck rows → INSERT with cast → UPDATE strips JSON → healed |
| Heal re-run for any reason | SELECT returns 0 → no-op |

## Why this stayed hidden for 5 releases

The v1125 code shipping in 1.12.5–1.12.9 is byte-identical for this method. The WARN has been firing on nightly PG runs since at least 2026-04-28 (Grafana). Four overlapping factors kept it invisible:

1. **Two stacked `try/catch` blocks** swallow the `BatchUpdateException` as a WARN; the migration framework records `v1125 = DONE`.
2. **`MigrationUtilTest` is fully mocked.** The PG-variant test passes an empty rows list, so the buggy INSERT path is never sent to a real driver.
3. **MySQL CI is green** because `INSERT IGNORE` accepts strings into JSON columns.
4. **Reads mask the broken state** — `EntityRepository.getCertification` only reads from `tag_usage`, and the entity's cert field is overwritten from `tag_usage` on every read. Affected entities look "uncertified" in the UI, so no escalation.

## Test plan

- [x] `mvn -pl openmetadata-service compile -DskipTests` — clean
- [x] `mvn -pl openmetadata-service test -Dtest=MigrationUtilTest` — 113/113 pass
- [x] `mvn -pl openmetadata-service spotless:apply` — applied

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/migration/mysql/v1130/Migration.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/migration/mysql/v1130/Migration.java
@@ -40,7 +40,7 @@ public class Migration extends MigrationProcessImpl {
     try {
       MigrationUtil.healStuckCertificationOnEntityJson(handle, ConnectionType.MYSQL);
     } catch (Exception e) {
-      LOG.error("v1130 heal of stuck certification on entity json failed; re-run to retry.", e);
+      LOG.error("v1130 heal of stuck certification on entity json failed", e);
     }
   }
 }
--- a/openmetadata-service/src/main/java/org/openmetadata/service/migration/mysql/v1130/Migration.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/migration/mysql/v1130/Migration.java
@@ -5,6 +5,7 @@ import static org.openmetadata.service.migration.utils.v1129.MigrationUtil.addTr
 
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
+import org.openmetadata.service.jdbi3.locator.ConnectionType;
 import org.openmetadata.service.migration.api.MigrationProcessImpl;
 import org.openmetadata.service.migration.utils.MigrationFile;
 import org.openmetadata.service.migration.utils.v1130.MigrationUtil;
@@ -36,5 +37,10 @@ public class Migration extends MigrationProcessImpl {
     MigrationUtil.addTableColumnSearchSettings();
     addTriggerOperationToDefaultBotPolicies(collectionDAO);
     addTriggerRuleToDataStewardPolicy(collectionDAO);
+    try {
+      MigrationUtil.healStuckCertificationOnEntityJson(handle, ConnectionType.MYSQL);
+    } catch (Exception e) {
+      LOG.error("v1130 heal of stuck certification on entity json failed; re-run to retry.", e);
+    }
   }
 }
--- a/openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v1130/MigrationUtil.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v1130/MigrationUtil.java
@@ -1,5 +1,7 @@
 package org.openmetadata.service.migration.utils.v1130;
 
+import static org.openmetadata.common.utils.CommonUtil.nullOrEmpty;
+
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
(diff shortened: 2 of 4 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/migration/mysql/v1130/Migration.java
```diff
@@ -36,5 +37,10 @@ public void runDataMigration() {
     MigrationUtil.addTableColumnSearchSettings();
     addTriggerOperationToDefaultBotPolicies(collectionDAO);
     addTriggerRuleToDataStewardPolicy(collectionDAO);
+    try {
+      MigrationUtil.healStuckCertificationOnEntityJson(handle, ConnectionType.MYSQL);
+    } catch (Exception e) {
+      LOG.error("v1130 heal of stuck certification on entity json failed; re-run to retry.", e);
```
harshach: why we need to add “re-run to retry”?

manerow: I dropped it. The phrasing was inherited from the existing v1130 catch (line 35 in this file) so I mirrored it for consistency, but you're right that it's misleading. Tracing the framework, when the catch swallows the exception, `runDataMigration()` returns normally and the workflow inserts `v1130` into `SERVER_CHANGE_LOG` via `updateMigrationStepInDB`.
Next startup, `resolveApplyMigrations` filters out anything already in `SERVER_CHANGE_LOG` (when `forceMigrations=false`), so `v1130` simply won't re-run

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v1130/MigrationUtil.java
```diff
@@ -366,4 +371,195 @@ public static void addTableColumnSearchSettings() {
...
+          .bind("tagFQN", tagFQN)
+          .bind("tagFQNHash", FullyQualifiedName.buildHash(tagFQN))
+          .bind("targetFQNHash", row.get("fqnhash").toString())
+          .bind("labelType", CERT_LABEL_TYPE)
+          .bind("state", CERT_STATE)
+          .bind("appliedAt", parseEpochMillisToTimestamp(row.get("applieddate")))
+          .bind("metadata", buildCertMetadataJson(row.get("expirydate")))
+          .add();
+    }
+    if (selectedIds.isEmpty()) {
```
harshach: can we add one return at the end and have inverse conditions

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v1130/MigrationUtil.java
```diff
@@ -366,4 +371,195 @@ public static void addTableColumnSearchSettings() {
...
+    if (rows.isEmpty()) {
+      return 0;
+    }
+
+    List<String> selectedIds = new ArrayList<>();
+    PreparedBatch batch = handle.prepareBatch(buildInsertTagUsageSql(isPostgres));
+    for (Map<String, Object> row : rows) {
+      String tagFQN = (String) row.get("tagfqn");
+      // SELECT already filters tagFQN NOT NULL, but stay defensive against concurrent edits.
+      if (tagFQN == null) {
```
harshach: same inverse condition, so that easy to follow

manerow: Done. Flipped to `if (tagFQN != null) { ... }`.

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v1130/MigrationUtil.java
manerow: Done. Single return with a `healed` accumulator + inverse guards. Same pattern applied to the other helpers in the file.

harshach (dismissed):

yan-3005 (approved):

PR #28509, opened 2026-05-28
fix(search): cap flattened children fields to prevent reindex failures

harshach:
### Describe your changes:

Fixes two interrelated OpenSearch indexing failures that surface when entities have deeply-nested column structures (e.g. Looker/DAX measure expressions embedded in struct sub-columns):

1. **Immense term** (`Document contains at least one immense term in field=...children.name`): `flattened` on Elasticsearch becomes `flat_object` on OpenSearch, and unlike ES, OpenSearch's `flat_object` does not support `ignore_above`. Any single nested leaf whose UTF-8 encoding exceeds Lucene's 32 766-byte per-term limit fails the entire bulk insert and leaves the entity permanently on the Reindex Failures tab.

2. **Mapping depth limit**: The previous fix (PR #28214) that introduced a nested `object` mapping for `children` grows the mapping depth by one level per nesting, exceeding the default `index.mapping.depth.limit: 20` for deeply-nested structs.

**Fix**: Replace `"type": "flattened"` with `"type": "object", "enabled": false` in the `children` block across all affected mappings — `table`, `dashboardDataModel`, `container`, `topic`, `apiEndpoint`, `worksheet`, and `searchEntity` — across all four language variants (`en`/`jp`/`ru`/`zh`). `enabled: false` is a terminal mapping node: the subtree is stored in `_source` for display but never parsed or indexed, so arbitrary document depth adds zero mapping depth (same property the old `flattened` mapping had) and no leaves are ever indexed (eliminating the immense-term exposure on both engines).

**Search continuity**: Since `children.name` leaves are no longer indexed, a new `columnNamesFuzzy` field (`text` with `om_analyzer`) is added to each affected index mapping. The index builders (`ContainerIndex`, `TableIndex`, `DashboardDataModelIndex`, `WorksheetIndex`) already collect the full column name hierarchy (including recursive children) into this field. Search query fields are updated to use `columnNamesFuzzy` instead of `columns.children.name` / `dataModel.columns.children.name`.

**Data migration**: Upgraded clusters retain stale `*.children.*` references in their DB-stored `SearchSettings` (highlight fields, search fields, and allowed fields) because OM's settings merge is additive. Two migrations scrub these dead entries:
- `v1130` migration extended: expands the existing single-field scrub (`dataModel.columns.children.name`) to all six stale fields across `highlightFields`, `searchFields`, and `allowedFields`.
- `v11210` migration added: runs the same scrub for clusters already past v1130.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

Two-part change:

1. **Mapping change** (28 JSON files): `"type": "flattened"` → `"type": "object", "enabled": false` on the recursive `children` block in each entity index. Adds `columnNamesFuzzy` (`text`, `om_analyzer`) as the new home for analyzed deep-column name search.

2. **Settings scrub migration** (v1130 + v11210): Removes references to the now-unindexed `*.children.*` fields from the DB-stored `SearchSettings` so highlight, search, and allowed-field lists don't point at unmapped leaves.

#
#### Use cases covered
- Table / DashboardDataModel / Container / Topic / APIEndpoint / Worksheet / SearchEntity with a struct column nested 25+ levels deep whose leaf description exceeds 32 KB now reindexes successfully instead of landing on the Reindex Failures tab.
- Deep-column names remain searchable via `columnNamesFuzzy` after the `children` field is removed from the index.
- Column Bulk Operations grid, Column Detail panel, global search, lineage, and CSV import/export continue to work — they read top-level fields (`columns.name.keyword`, `_source`), which are unaffected.
- Stale `*.children.*` entries are removed from `SearchSettings` (highlight, search, and allowed fields) on upgrade via both the v1130 and v11210 migrations.

- [x] `SearchIndexImmenseTermIT` — regression guard for the combined worst case: struct column nested 25 levels deep with a >32 KB leaf description. Asserts the entity indexes successfully and the leaf column name is searchable via `columnNamesFuzzy`.
- [x] `FlattenedChildrenHighlightSearchIT` — updated to reflect the new mapping behavior: with `object/enabled:false`, the stale highlight field is now inert (no 500) and the search succeeds before and after the scrub migration runs.

#### Unit tests
- [x] `FlattenedChildrenScrubTest` — verifies `stripFlattenedChildrenReferences` removes all six stale field variants from `highlightFields`, `searchFields`, and `allowedFields` while leaving all non-children fields intact.

- [x] Not applicable — no ingestion changes.

#### Playwright (UI) tests
- [x] `SearchIndexNestedColumns.spec.ts` — end-to-end: creates a table with a 25-level deeply-nested struct column carrying an oversized leaf description, waits for indexing, and asserts the table is discoverable via global search.

#### Manual testing performed
1. Reproduced the failure on a production OpenSearch 1.12.9 cluster: `Document contains at least one immense term in field=fields(columns.children.name).keyword whose UTF-8 encoding is longer than the maximum length 32766` on `dashboardDataModel` entities.
2. Applied `object/enabled:false` to the live index via `PUT /dashboard_data_model_search_index/_mapping`.
3. Re-ran Search Indexing scoped to Dashboard Data Model; previously failing entities indexed successfully and cleared from the Failures tab.
4. Confirmed `/v1/columns/grid` and global search continued returning the same top-level column results.

#

Not applicable.

#
### Checklist:
- [x] For index mapping changes: deployments running with `recreate-index: true` (the standard setup) pick up the new mappings on the next reindex run automatically. No DDL migration is required.
- [x] For UI changes: N/A — no UI source changes. The Playwright test exercises the existing search UI against the fixed backend mappings.

----
## Summary by Gitar

- **Elasticsearch Mappings:**
  - Replaced `flattened` mapping for `children` fields with `object` (`enabled: false`) across all entity indices to prevent `immense term` indexing failures.
  - Replaced `columns.children.name` search fields with `columnNamesFuzzy` (analyzed `text`) to preserve deep-column searchability.
- **Database Migrations:**
  - Added v1.12.10 migration to automatically scrub stale `*.children` references from `SearchSettings` (highlight, search, and allowed fields) stored in the database.
- **Testing Infrastructure:**
  - Added `SearchIndexImmenseTermIT` to verify indexing/searchability for deeply nested, oversized column descriptions.
  - Added `FlattenedChildrenScrubTest` to validate the removal of dead search settings during migration.
  - Added `SearchIndexNestedColumns.spec.ts` for end-to-end UI testing of search functionality for deeply nested column structures.


sonika-shah:
Superseded by #28530, which fixes the same reindex failure with a generic application-side value cap (trims any leaf > ~32 KB before indexing) instead of the `ignore_above` mapping change. `ignore_above` is unsupported on `flat_object`, so on OpenSearch the index is created without its mapping (dynamic fallback → `serviceType` becomes `text` → aggregations/sorting break, which is what caused the 221 Playwright failures). The #28530 approach behaves identically on Elasticsearch and OpenSearch. Suggest closing this in favor of #28530.

mohityadav766 (dismissed):

ShaileshParmar11 (approved):

PR #28510, opened 2026-05-28
fix(snyk): drop ingestion/ prefix on CWD-relative exclude paths

akashverma0786:
## Summary

The Makefile runs the ingestion Snyk scan from `ingestion/`:

```make
snyk-ingestion-report:
    cd ingestion; \
        snyk code test $(SNYK_ARGS) --json > ../security-report/ingestion-code-scan.json | true;
```

So paths in the resulting SARIF are CWD-relative (e.g. `src/metadata/sdk/examples/dq_as_code_example.py`), not repo-root-relative. The two `.snyk` patterns below were written with an `ingestion/` prefix that can never match those paths, so the excludes were silently no-ops:

- `ingestion/src/metadata/sdk/examples/**` → `src/metadata/sdk/examples/**`
- `ingestion/src/_openmetadata_testutils/**` → `src/_openmetadata_testutils/**`

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,9 +5,9 @@ patch: {}
 exclude:
   global:
     - ingestion/examples/**
-    - ingestion/src/metadata/sdk/examples/**
+    - src/metadata/sdk/examples/**
     - ingestion/tests/**
     - openmetadata-ui/src/main/resources/ui/src/pages/service/mocks/**
     - openmetadata-ui/src/main/resources/ui/src/components/common/TestConnection/TestConnection.mock.ts
     - openmetadata-service/src/test/**
-    - ingestion/src/_openmetadata_testutils/**
+    - src/_openmetadata_testutils/**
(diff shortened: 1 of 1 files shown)


.snyk
```diff
@@ -5,9 +5,9 @@ patch: {}
 exclude:
   global:
     - ingestion/examples/**
```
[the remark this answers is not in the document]
mohittilala: @akashverma0786 let's address this as well

.snyk
```diff
@@ -5,9 +5,9 @@ patch: {}
...
-    - ingestion/src/metadata/sdk/examples/**
+    - src/metadata/sdk/examples/**
```
mohittilala, line 13: can we group these and and comment that these are ingestion exlusions and the reason for why it starts with `src/` and not `ingestion/src`?

akashverma0786: Added this.

mohittilala:
Closing for #28526, which adds a dedicated ingestion/.snyk since Snyk Code only reads the policy file from the scan CWD.

PR #28514, opened 2026-05-28
feat(sdk): add getLogLocation hook on PipelineServiceClientInterface

IceS2:
## Summary

Adds a default `String getLogLocation(IngestionPipeline, String runId)` method to `PipelineServiceClientInterface` so any pipeline-service client can opt-in to exposing the storage location of a run's logs (e.g. an `s3://bucket/key` URI) without downloading them. The default returns `null`, so existing clients (Airflow, K8s, Noop) are unaffected.

## Why

Operators need the real S3 location of pipeline run logs so they can fetch the file directly. Streamable pipelines already have a deterministic S3 key derived from `LogStorageConfiguration`. Non-streamable clients (Argo in particular) write logs to their own artifact location that is only knowable per-run from the workflow status — this PR adds the generic hook so each client can expose that. First consumer is an admin-ops endpoint in `openmetadata-collate`.

## Changes

- **`PipelineServiceClientInterface`** — new `default String getLogLocation(IngestionPipeline ingestionPipeline, String runId) { return null; }`. When `runId` is `null`, the latest run is used.
- **`MeteredPipelineServiceClient`** — `@Override` forwarding to `decoratedClient.getLogLocation(...)` so the metering decorator doesn't swallow the call.
- **`IngestionPipelineRepository`** — new public `getLogLocationFromPipelineService(IngestionPipeline, UUID)` wrapper that keeps `pipelineServiceClient` encapsulated. Streamable-S3 locations remain derived from `getLogStorageConfiguration()` elsewhere; this covers the non-streamable case.

## Backward compatibility

No-op for existing clients (default returns `null`). Purely additive — no method signatures changed, no behavior changes for any path that doesn't override.

## Testing

- Suggested unit tests:
  - `MeteredPipelineServiceClientTest` — verify `getLogLocation` forwards to the delegate with the same args.
  - `IngestionPipelineRepositoryTest.getLogLocationFromPipelineService` — null client / null pipeline / forwards to client; UUID-to-String conversion.
- No existing tests are affected (additive).

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/clients/pipeline/MeteredPipelineServiceClient.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/clients/pipeline/MeteredPipelineServiceClient.java
@@ -137,6 +137,11 @@ public class MeteredPipelineServiceClient implements PipelineServiceClientInterf
   }
 
   @Override
+  public String getLogLocation(IngestionPipeline ingestionPipeline, String runId) {
+    return decoratedClient.getLogLocation(ingestionPipeline, runId);
+  }
+
+  @Override
   public Map<String, String> getIngestionLogs(
       IngestionPipeline ingestionPipeline, String after, String runId) {
     return executeWithMetering(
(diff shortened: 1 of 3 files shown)


IceS2:
Superseded — the collate consumer no longer needs a per-run log location lookup. Instead, the admin-ops endpoint will surface `runnerType` + identifiers (`argoWorkflowId` / `runnerId`) and let consumers compute the path themselves. Closing as YAGNI; can resurrect if another consumer ever needs a programmatic hook.

PR #28596, opened 2026-06-01
fix(search): scrub stale file extension aggregation on v1130 upgrade (#1A file search 500)

sonika-shah:
Cherry-pick of #28555 onto 1.13.

## Summary
Scrubs stale `extension` searchField and aggregation from the DB-stored file SearchSettings on clusters upgrading from pre-1.12.5 to 1.13. Also includes the v11210 migration for completeness (clusters on 1.12.10 upgrading to 1.13 get both scrubs in version order).

## Root cause
PR #27080 renamed `extension` → `fileExtension` in the seed but the additive settings merge never updates existing asset type configs in the DB — stale `extension` aggregation causes `illegal_argument_exception` on OpenSearch for every file search.

## Original PR
#28555


sonika-shah:
Closing — using GitHub cherry-pick flow instead

PR #28656, opened 2026-06-03
Fixes #28646: match column FQNs case-sensitively when patching column tags

### LINKED ISSUE #28646: patch_column_tags silently mis-applies tags on tables with case-distinct duplicate column names
### Affected module

Governance (Glossary / Classification / Domains)

### Describe the bug

```
  Affects
  
  openmetadata-ingestion 1.12.6 (and likely earlier — code unchanged for a while). Module: metadata.ingestion.ometa.mixins.patch_mixin.

  Summary
  
  OMetaPatchMixin.patch_column_tags calls the module-level helper update_column_tags
  (https://github.com/open-metadata/OpenMetadata/blob/1.12.6-release/ingestion/src/metadata/ingestion/ometa/mixins/patch_mixin.py#L79-L98), which matches column FQNs case-insensitively with an early break:

  def update_column_tags(columns, column_tag, operation):
      for col in columns:
          if str(col.fullyQualifiedName.root).lower() == column_tag.column_fqn.lower():
              if operation == PatchOperation.REMOVE:
                  ...
              else:
                  col.tags.append(column_tag.tag_label)
              break  # <-- only patches the first lowercased match

  When a table contains two columns whose names differ only in case (which Snowflake, S3 schema inference, and several other sources permit and produce), and the caller passes one ColumnTag per real column
  intended for both, every ColumnTag collapses onto the same first-lowercased-match column — appending the same tag multiple times to one column and never touching the other.

  This is the same class of bug as #28077 (Snowflake + Data Quality column lookup), just surfaced through a different code path. The pattern — "look up columns by FQN with case-folding, no disambiguation
  when duplicates exist" — likely affects more surfaces.

```

### To Reproduce

```
A Snowflake / S3 / arbitrary table contains two columns:

  ┌─────────────────────┬──────────────┐
  │         FQN         │ Initial tags │
  ├─────────────────────┼──────────────┤
  │ db.schema.t.country │ []           │
  ├─────────────────────┼──────────────┤
  │ db.schema.t.Country │ []           │
  └─────────────────────┴──────────────┘

  Caller code:
  
  table = om_client.get_by_name(entity=Table, fqn="db.schema.t", fields=["columns", "tags"])
  column_tags = [
      ColumnTag(column_fqn="db.schema.t.country",
                tag_label=TagLabel(tagFQN=TagFQN(root="PDG.country"), source=TagSource.Classification,
                                   labelType=LabelType.Automated, state=State.Confirmed)),
      ColumnTag(column_fqn="db.schema.t.Country",
                tag_label=TagLabel(tagFQN=TagFQN(root="PDG.country"), source=TagSource.Classification,
                                   labelType=LabelType.Automated, state=State.Confirmed)),
  ]
  om_client.patch_column_tags(table=table, column_tags=column_tags)
```

### Expected behavior

```
Expected

  Each column receives the tag independently: country.tags = [PDG.country], Country.tags = [PDG.country].

  Actual
  
  After mutating destination in memory: destination.columns[country].tags = [PDG.country, PDG.country] (duplicate), destination.columns[Country].tags = [] (never visited because of the break).

  patch_column_tags then calls jsonpatch.make_patch(source, destination). With duplicates accumulating in the array across consecutive runs, the diff library's array diff is non-deterministic — alternately
  emitting add then remove ops between runs. The server faithfully applies each patch, producing an observable add/delete flap on the table.

  Real prod symptom (PDG tag attachment job, every 2h):

  v13.9  tags PDG.country has been deleted    2026-06-02 10:07
  v13.8  tags PDG.country has been added      2026-06-02 08:15
  v13.7  tags PDG.country has been deleted    2026-06-02 07:42
  v13.6  tags PDG.country has been added      2026-06-02 06:07
  ... continues since ingestion, never converges ...

  Tables with single (non-case-colliding) column names work correctly. The bug only manifests when at least two columns in the same table share a case-insensitive name.


```

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

_No response_

_No response_

### Additional context

```
Two layers:
  
  1. Match exact-case, not lowercased — drop the .lower() in the comparison so each ColumnTag patches the column whose FQN matches exactly. This is the right behavior for sources that permit case-distinct
  columns.
  2. Disambiguate or fail loudly when the same lowercase name maps to multiple columns — at minimum, raise a clear error so callers don't silently lose work.

  Minimal patch:

  def update_column_tags(columns, column_tag, operation):
      for col in columns:
          if str(col.fullyQualifiedName.root) == column_tag.column_fqn:
              if operation == PatchOperation.REMOVE:
                  for tag in col.tags:
                      if tag.tagFQN == column_tag.tag_label.tagFQN:
                          col.tags.remove(tag)
              else:
                  col.tags.append(column_tag.tag_label)
              return
          if col.children:
              update_column_tags(col.children, column_tag, operation)

  Replacing case-insensitive matching with exact-case matching does change behavior for the dbt use case mentioned in the existing comment (# For dbt the column names in OM and dbt are not always in the 
  same case). That comment is in update_column_description, not update_column_tags, but the same case-folding pattern appears there too. If case-insensitive matching is intentional for dbt, consider:

  - Making case-sensitivity an opt-in flag at the call site, OR
  - Detecting case-collision pairs at the start of the helper and failing fast with a descriptive error.

  Workaround used downstream

  For users blocked today: bypass patch_column_tags and craft the JSON Patch directly against /v1/tables/{id} by looking up each column's exact-case FQN to its array index and emitting {"op": "add", 
  "path": "/columns/<idx>/tags/-", "value": {...}}. This avoids both the case-insensitive match and the unstable jsonpatch.make_patch array diff. (Internal MR doing this: <link to your MR if you want, or
  omit.>)
  
  Related
  
  #28077 — same root cause (case-folded column lookup) surfaced via Snowflake + Data Quality. A fix here should be coordinated with that.


```

### Pre-submission checklist

mattheworford:
### Describe your changes:

Fixes #28646 

`update_column_tags` (in `patch_mixin.py`) lowercased both sides of the column-FQN comparison, so columns that differ only in case (e.g. `country` and `Country` under the same table) collapsed onto first match. Tagging the second column applied the tag to the wrong column and left the intended one untouched.

I changed the comparison to match the FQN exactly instead of lowercasing. The auto-PII tagging path already passes the column's own `fullyQualifiedName`, so case should always agree. Only behavioral change is that the deprecated `patch_column_tag(column_fqn=...)` now requires an exact-case FQN. 

The sibling `update_column_description` helper keeps its case-insensitive match on purpose (documented inline) because dbt column names legitimately differ in case.

#
### Type of change:
- [x] Bug fix

#
### High-level design:
N/A — small change.

#
#### Use cases covered
- A table with two columns whose FQNs differ only in case (`country` / `Country`) gets each tag applied to the correct column, instead of both landing on the first one.
- Removing a tag affects only the case-exact column, leaving its case-distinct sibling untouched.
- The same case-sensitive matching holds for nested (STRUCT child) columns.

#### Unit tests
- Files added/updated:
  - `ingestion/tests/unit/test_ometa_patch_mixin.py`
- Coverage on changed classes (run `mvn jacoco:report` for backend, `yarn test:coverage` for UI,
  `make unit_ingestion` for ingestion). Target is 90% line coverage on changed classes.
- Coverage %: 100% (`update_column_tags`/patch_mixin.py:100-119)

- [ ] I added integration tests in `openmetadata-integration-tests/` for new/changed API endpoints.
- Files added/updated:

- [ ] I added/updated ingestion integration tests for connector changes.
- Files added/updated:

#### Playwright (UI) tests
- [ ] I added Playwright E2E tests under `openmetadata-ui/.../ui/playwright/` for UI changes.
- Files added/updated:

#### Manual testing performed
N/A — pure helper function with deterministic behavior covered by the unit tests above

#
Not applicable.

#
### Checklist:


mattheworford:
@harshach seems like the failures here are unrelated - are these known flakes?

harshach (approved):

harshach:
@mattheworford Yes , will merge it in. Thanks for the PR

mattheworford:
@harshach - any update here?

IceS2:
Hey @mattheworford,

I was checking the PR and to be fair I am a bit concerned about the blast radius of this change.

I understand the specific case you are mentioning, but the case the source supports and the case we receive is source dependent and this change could potentially impact other sources.

I was playing around with how we could handle this and I believe in the end it should be configurable per source,
defining if we should match for everything the fqns as uppercase, lowercase or with how we receive it. This would also guarantee we are doing the match consistently across the stack, although we would potentially need to retain the original case due to the profiler and data quality workflows.

PR #28661, opened 2026-06-03
test(activity): lower ActivityFeedAlert poll interval for CI e2e runs

manerow:
## Problem

The `ActivityAPI.spec.ts` → *Activity API - Entity Changes* tests (description / tags / owner) intermittently time out in CI with `200 []` — the activity row never appears within the 200s poll window.

The cause is **throughput, not correctness**. Activity rows are delivered by a single `ActivityFeedAlert` event-subscription consumer at the default `pollInterval=30s` / `batchSize=100` (~200 events/min, offset-ordered FIFO). The parallel e2e suite (multiple shards against one shared backend) generates change events faster than the consumer can drain them, so the backlog grows and the test's event is still queued when the poll window expires. The pipeline itself is correct — the change event is produced, passes the filter, and the row is written — just too late.

Prior fixes in this path addressed *correctness* (#28304 batch-drop on deleted users, #28539 write-path exceptions); this one addresses *throughput*.

## Change

A CI-only Playwright setup project (`activity-feed-config`) lowers the `ActivityFeedAlert` `pollInterval` to **10s** on the ephemeral test deployment, via the supported subscription API (which reschedules the consumer live without losing its offset). It's wired as a dependency of the `chromium` project that runs the spec.

- **CI-gated** (`process.env.CI`) — local dev environments are never mutated, and the suite passes locally without the boost (no parallel flood).
- **Production-safe by construction** — production never runs Playwright; no seed/default or migration change.
- **No teardown** — the deployment is ephemeral, and the backend is shared across parallel shards, so a per-shard revert would race and re-break shards still running. Tunable via `ACTIVITY_FEED_POLL_INTERVAL` if 10s proves insufficient.

## Testing

- ESLint + Prettier clean on the changed files.
- Verified on a live backend that patching `pollInterval` via the subscription API reschedules the consumer and the activity row is delivered well within the poll window.


openmetadata-ui/src/main/resources/ui/playwright/e2e/activity-config.setup.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      await apiContext.patch(
+        `/api/v1/events/subscriptions/${subscription.id}`,
+        {
+          data: [
+            { op: 'replace', path: '/pollInterval', value: POLL_INTERVAL },
+          ],
+          headers: { 'Content-Type': JSON_PATCH_CONTENT_TYPE },
+        }
+      );
```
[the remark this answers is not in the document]
manerow: Good call — addressed in 2f48499744. Kept the GET lenient (the subscription can legitimately be absent) but now capture the patch response and `expect(patchResponse.ok()).toBeTruthy()`, so a misapplied boost fails the setup project loudly instead of resurfacing as ActivityAPI flakiness.

openmetadata-ui/src/main/resources/ui/playwright/e2e/activity-config.setup.ts
```diff
@@ -0,0 +1,72 @@
...
+
+      // patch() does not throw on non-2xx — fail loudly so a misapplied boost
+      // surfaces here instead of as renewed ActivityAPI flakiness downstream.
+      expect(patchResponse.ok()).toBeTruthy();
```
chirag-madlani: @manerow if this fails, all the other tests will be skipped is that intended?

manerow: Good catch, I've fixed it.

chirag-madlani (approved):

PR #28675, opened 2026-06-03
perf: one-transaction flush + async indexing write path, read-path fixes

### LINKED ISSUE #28726: perf: reduce write/read API latency — one-transaction flush + read-path bulk-field fetch
### Problem

OpenMetadata's entity create/update/patch write path commits the same logical write 5–7 times: the per-DAO `@Transaction` annotations are decorative no-ops, so each DAO call autocommits on its own connection. This multiplies commit/fsync cost, holds a pooled DB connection open across external network round trips (Redis L2 invalidation, RDF/SPARQL, lineage and search reindex), and can leave orphan rows when a flush fails midway. On the read path, several repositories override `setFields` but not `setFieldsInBulk`, so list-typed fields are silently dropped on list endpoints and then refetched per item (N+1), inflating read latency.

### Goal

Lower read and write API latency while preserving correctness:

- Wrap each create/update/patch flush in a single real `Entity.getJdbi().inTransaction(...)` plus deadlock-retry so a write is one atomic commit that leaves no orphan rows on mid-flush failure.
- Defer every in-flush external side effect (cache L2 invalidation, RDF, lineage-ES, rename-cascade reindex) to a synchronous post-commit drain so the held DB connection makes no network call, while keeping `GET /{id}` and `/name/{fqn}` strictly read-your-write (DB plus synchronous cache write-through). Search stays consistent immediately after the write returns.
- Add the missing bulk field-fetch paths (Directory, Spreadsheet, all `*Service`, KnowledgePage, ContextMemory) and related read-path savings (batched single-entity tag inserts, a bounded Fernet key cache, and dropping a redundant CSV-export tag rescan).

### Acceptance

- Create/update/patch is atomic and deadlock-replay-safe, including glossary-term rename and data-product domain-change cascades.
- Read-your-write holds for GET by id and by name; search is consistent post-commit.
- Previously-dropped list fields populate on the relevant list endpoints.
- Full integration-test suite green on both the Elasticsearch and OpenSearch backends.

Tracked by PR #28675.

harshach:
### Describe your changes:

Fixes #28726

Optimizes the OpenMetadata read and write paths for lower API latency. **Reads:** several repositories overrode `setFields` but not `setFieldsInBulk`, silently dropping list fields and forcing per-item queries — added bulk field-fetch paths (Directory, Spreadsheet, all `*Service`, KnowledgePage, ContextMemory), dropped a redundant CSV-export tag rescan, batched single-entity tag inserts, and cached Fernet keys. **Writes — Stage 1:** the create/update/patch flush committed 5–7 times via per-DAO autocommit (the `@Transaction` annotations were decorative no-ops) and left orphan rows on mid-flush failure; it is now wrapped in one real `Entity.getJdbi().inTransaction(...)` + `DeadlockRetry`, with every external side-effect (cache-L2 invalidation, RDF/SPARQL, lineage-ES, rename-cascade reindex) deferred so the held DB connection makes no network round trip. **Writes — Stage 2:** those externals + the ES entity index now run asynchronously off the request thread on a new per-entity-ordered `OrderedLaneExecutor` (durable via the search retry outbox), while cache write-through stays synchronous post-commit so `GET /{id}` and `/name/{fqn}` remain read-your-write — only `/search` and SPARQL become eventually consistent.

#
### Type of change:
- [x] Improvement

#
### High-level design:

- **Stage 1** wraps `EntityRepository` create/update/patch flushes in `inTransaction` + deadlock-retry (mirroring the existing `cleanup()` idiom), collapsing 5–7 commits → 1 and making writes atomic. In-flush external calls are captured into thread-local collectors and drained post-commit; `EntityUpdater` state is snapshotted/restored per retry attempt (incl. the 6 cascade-bearing subclass updaters).
- **Stage 2** routes those drains plus the ES entity index (`SearchIndexHandler.isAsync()=true`) onto `OrderedLaneExecutor`, keyed by entity id so single-consumer lanes preserve per-entity ordering. A full lane sheds to `SearchIndexRetryQueue` (a fast local-DB upsert) rather than blocking the request thread; bulk create slices per-entity onto each member's own lane.
- **Consistency contract:** GET-by-id/by-name stay real-time (DB + sync cache); search/RDF/lineage are eventually consistent with durable retry. Redis stays opt-in (`CACHE_PROVIDER=none` default).
- **Honest residual:** under pathological lane overflow/hard-stop, the outbox rebuilds an entity's doc + lineage from committed DB, but RDF triples and rename-cascade *descendant* reindexes are best-effort.

#
#### Use cases covered
- Create/update/patch is atomic (mid-flush failure → zero orphan rows) and deadlock-replay-safe, incl. glossary-term rename and data-product domain-change cascades.
- Write returns before the doc is searchable; doc converges in search; same-entity create→update converges to the final value (lane ordering).
- Previously-dropped list fields now populate (Drive directories/spreadsheets, `*Service` pipelines, Knowledge Center hierarchy, Context memories).

- [x] Added/updated in `openmetadata-integration-tests/`.
- `OneTransactionFlushAtomicityIT`, `AsyncSearchIndexConsistencyIT`, `DirectoryResourceIT`, `SpreadsheetResourceIT`, `DatabaseServiceResourceIT`, `ContextMemoryIT`, `DatabaseResourceIT`, `TableResourceIT`, `KnowledgePageHierarchyIT`, `ColumnSearchIndexIT`/`TestCaseResourceIT` (Awaitility polling for async), `WritePathBenchmarkIT` (inert perf harness).

#### Unit tests
- `OrderedLaneExecutorTest`, `EntityLifecycleEventDispatcherTest`, `SearchIndexHandlerTest`, `FernetKeyCacheTest`.

#### Playwright (UI) tests

#### Manual testing performed
- Ran targeted ITs + the full IT suite (`mvn verify -Ppostgres-elasticsearch`, ~13.5k tests) on `cache=none` and `cache=redis`; all green except pre-existing async-workflow flakiness.
- Write-path benchmark on a durable (`fsync=on`) Postgres showed ~30–40% median write-latency improvement; the residual p99 tail is DB-commit/checkpoint + GC, not the OM write path.

#

Not applicable.

#
### Checklist:
- [ ] For JSON Schema changes: N/A (no schema changes).


harshach:
Thanks for the review — both findings are addressed in 18d2d923ec.

**Finding 1 (Bug — async indexer captures a mutable entity)**
`SearchIndexHandler.isAsync()` is now `true`, so the ordered lane reads the entity off the request thread *after* dispatch — but the request thread keeps mutating that same instance for the HTTP response (PII masking, `clearFields` stripping, secret masking on service connections). The lane could therefore index a masked/stripped/partial doc. Per-entity lane ordering doesn't help here because the race is on the in-memory POJO, not on lane submission order.

Fix: `executeHandler(EntityInterface, …)` now takes a `Consumer<EntityInterface>`, and the async branch deep-copies the committed state before submitting to the lane:
```java
EntityInterface snapshot =
    JsonUtils.readValue(JsonUtils.pojoToJson(entity), entity.getClass());
orderedLaneExecutor.submit(
    entity.getId(), laneTask(entity, operation, () -> handlerCall.accept(snapshot)));
```
The sync branch still binds the live entity (no copy cost when nothing is async). The `EntityReference` overload is unaffected (references aren't response-mutated). The six `onEntity*` call sites were updated from `() -> handler.onX(entity, …)` to `e -> handler.onX(e, …)`.

**Finding 2 (Quality — `invalidateCacheForTaggedEntities(String)` always returned 0)**
`deferOrRunSearchBackedInvalidation` discarded the inline count. It now takes an `IntSupplier`, so the inline (non-flush) path returns the live invalidated count:
```java
} else {
  result = invalidation.getAsInt();
}
```
The deferred path still returns `0` — the work runs post-commit and the count is only known then (it's logged inside `searchTaggedEntitiesAndInvalidate`); this is documented in the method's Javadoc.

`mvn spotless:check` and a service-module compile both pass.

harshach:
Addressed the snapshot findings in `399647568c`, and investigated the other two. Summary:

### ✅ Bug — Unguarded snapshot serialization can fail API after commit (fixed)
Correct: the snapshot ran unguarded on the request thread, and dispatch happens **post-commit** for create/update, so a serialization throw would 5xx an already-committed write *and* skip the durable outbox. Fixed with `snapshotOrEnqueueRetry`:
```java
private EntityInterface snapshotOrEnqueueRetry(EntityInterface entity, String operation) {
  EntityInterface snapshot;
  try {
    snapshot = JsonUtils.readValue(JsonUtils.pojoToJson(entity), entity.getClass());
  } catch (Throwable serializationFailure) {
    LOG.error("Snapshot serialization failed for {} '{}'; routing to durable search-index retry outbox",
        operation, entity.getId(), serializationFailure);
    enqueueLaneFailureRetry(laneTask(entity, operation, () -> {}), serializationFailure);
    snapshot = null;          // caller skips the lane submit
  }
  return snapshot;
}
```
On failure the entity locator goes to `SearchIndexRetryQueue` (the worker rebuilds the doc from committed DB state), so the committed write returns 2xx and the index still converges. I catch `Throwable` (not just `RuntimeException`) to match the lane-level retry net, so even a pathological `Error` lands in the outbox rather than escaping to the request thread.

### ✅ Performance — Entity re-serialized once per async handler (fixed)
Correct, and it bites by default: there are **two** async handlers (`SearchIndexHandler` + `DomainSyncHandler`), so a wide entity was serialized 2–3×. Fixed with `LazyEntitySnapshot`, a request-thread-only memoized `Supplier<EntityInterface>` computed at most once per entity per dispatch and shared across all async handlers. Bulk dispatch builds one per-entity supplier map (keyed by entity id) reused across handlers; the per-entity lane-ordering slicing is unchanged.

Added unit tests: async handlers share one snapshot per entity (single + bulk via `assertSame`), and a post-commit snapshot failure returns 2xx with the async handlers skipped and the entity routed to the outbox.

### ⏭️ Edge Case — deleteStale with empty seenFqns (not introduced here; intended contract)
`bulkDeleteStaleEntities` / `findStaleEntities` / `validateScopeRequest` are **byte-for-byte identical to `origin/main`** (`git diff origin/main` shows zero hunks). Empty `seenFqns` is the intended stale-reconciliation contract — "the connector saw nothing in this scope, soft-delete the stale remainder" — with safety rails already in place: soft-delete is the default (`hardDelete` is never set by the Python client), `scopeExists()` short-circuits a missing scope to a no-op, and the class javadoc documents it. The Python client deliberately sends empty `seenFqns` on an emptied scope, and `ingestion/tests/unit/topology/test_delete_stale.py` asserts that path. Rejecting empty `seenFqns` would break that legitimate flow. If hardening against operator misconfiguration is desired, that's a separate change against `origin/main`, not this latency PR.

### ⏭️ Bug — Bot PUT owner guard (not introduced here)
`updateOwners` is also **unchanged vs `origin/main`**. `git blame` attributes the `operation.isPut() && !nullOrEmpty(original.getOwners()) && updatedByBot() && !overrideMetadata` guard to commit `f49e46438d` (PR #28155, "Improve bulk endpoints for ingestion"), which is already merged to main and outside this PR's commit range (`git log --oneline origin/main..HEAD` does not contain it). The matching `updateDescription`/`updateDisplayName` guards predate this PR as well. So the owner-propagation contract is identical on both branches; any change to it belongs in a discussion on #28155.

`mvn spotless:check`, a service-module compile, and the dispatcher + ordered-lane unit suites (29 tests) all pass.

manerow (approved):
Looks good to me. The one-transaction wrap and deferring the external side effects out of the held connection reads cleanly. Approving.

PR #28676, opened 2026-06-04
fix(scim): let the SCIM bot update displayName via per-entity PUT

harshach:
### Describe your changes:

Follow-up to #28155 (fixes a regression introduced there).

PR #28155 added an in-code guard to `EntityRepository.updateDisplayName()` that reverted `displayName` on **any** bot PUT, silently breaking SCIM displayName sync (the SCIM `PATCH` returned 200 but the value never changed). A PUT — per-entity or bulk — authorizes with the coarse `EDIT_ALL` operation, which does not intersect the field-level `EditDisplayName` policy deny, so an in-code guard is needed to re-apply that deny; but it must mirror the policy rather than block every bot. This PR makes the guard **policy-aware**: `updateDisplayName()` preserves a user-curated `displayName` only when the updating bot is actually DENIED `EditDisplayName` by policy (e.g. the ingestion bot via `DefaultBotPolicy`/`IngestionBotPolicy`, where DENY is sticky over `EditAll`); bots the policy allows — notably the SCIM bot (`ScimBotPolicy` has no `DisplayName-Deny`) — update it, and `overrideMetadata=true` still force-syncs on the bulk path. `updateDescription`/`updateOwners` are unchanged.

#
### Type of change:
- [x] Bug fix

#
#### Use cases covered
- An ingestion bot (policy denies `EditDisplayName`) cannot overwrite a user-curated `displayName` via a per-entity PUT **or** the bulk endpoint.
- A bot whose policy allows `EditAll` with no `DisplayName-Deny` (modeling the SCIM bot) updates `displayName` via a per-entity PUT.
- `overrideMetadata=true` still lets a bot force-sync `displayName` on the bulk endpoint.

- [x] Added in `openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BaseEntityIT.java`:
  - `test_singleEntityPut_bot_preservesUserDisplayName` (ingestion bot → preserved)
  - `test_singleEntityPut_displayNameAllowedBot_updatesDisplayName` (policy-allowed, SCIM-like bot → updated)
  - existing `test_bulkUpdate_bot_preservesUserDisplayName` / `test_bulkUpdate_overrideMetadata_canOverrideUserDisplayName` continue to pass

#### Playwright (UI) tests

#### Manual testing performed
Ran the four displayName tests against the real testcontainers stack on `DatabaseResourceIT` — all pass; the per-entity preservation test is RED before the fix and GREEN after.

#

Not applicable.

#
### Checklist:

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BaseEntityIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BaseEntityIT.java
@@ -24,6 +24,7 @@ import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.openmetadata.it.auth.JwtAuthProvider;
 import org.openmetadata.it.bootstrap.SharedEntities;
 import org.openmetadata.it.util.BulkApi;
 import org.openmetadata.it.util.EntityValidation;
@@ -32,8 +33,19 @@ import org.openmetadata.it.util.TestNamespace;
 import org.openmetadata.it.util.TestNamespaceExtension;
 import org.openmetadata.it.util.UpdateType;
 import org.openmetadata.schema.EntityInterface;
+import org.openmetadata.schema.api.policies.CreatePolicy;
+import org.openmetadata.schema.api.teams.CreateRole;
+import org.openmetadata.schema.api.teams.CreateUser;
+import org.openmetadata.schema.auth.JWTAuthMechanism;
+import org.openmetadata.schema.auth.JWTTokenExpiry;
+import org.openmetadata.schema.entity.policies.Policy;
+import org.openmetadata.schema.entity.policies.accessControl.Rule;
+import org.openmetadata.schema.entity.teams.AuthenticationMechanism;
+import org.openmetadata.schema.entity.teams.Role;
+import org.openmetadata.schema.entity.teams.User;
 import org.openmetadata.schema.type.ApiStatus;
 import org.openmetadata.schema.type.ChangeDescription;
+import org.openmetadata.schema.type.MetadataOperation;
 import org.openmetadata.schema.type.TagLabel;
 import org.openmetadata.schema.type.api.BulkOperationResult;
 import org.openmetadata.schema.type.api.BulkResponse;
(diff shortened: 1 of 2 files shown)


harshach:
Thanks for the review — both findings were spot on. Addressed in 68326ff.

### Finding 1 (Bug) — fixed

You're right that `bulkOperation` was the wrong gate: a per-entity PUT on an existing entity authorizes with coarse `EDIT_ALL`, which doesn't intersect the `EditDisplayName` deny, so the policy layer did **not** protect the per-entity path and any bot could clobber displayName there.

The guard is now **policy-aware** — it re-applies the field-level check that `EDIT_ALL` bypasses, keying on the bot's actual `EditDisplayName` permission rather than bulk-vs-per-entity:

```java
boolean preserveUserDisplayName =
    updatedByBot()
        && !nullOrEmpty(original.getDisplayName())
        && !overrideMetadata
        && !Objects.equals(original.getDisplayName(), updated.getDisplayName())
        && updatingBotDeniedOperation(MetadataOperation.EDIT_DISPLAY_NAME);
```

`updatingBotDeniedOperation` evaluates `PolicyEvaluator.getPermission(subjectContext, entityType)` and checks for `Access.DENY` on `EDIT_DISPLAY_NAME`. Because `CompiledRule.overrideAccess` makes DENY sticky (lowest ordinal, never overridden by a later `EditAll` ALLOW), this reliably returns DENY for the ingestion bot (`DefaultBotPolicy`/`IngestionBotPolicy`) and NOT_ALLOW for the SCIM bot (`ScimBotPolicy`, which has no `DisplayName-Deny`). Net result:

- ingestion bot → preserved on **both** the per-entity and bulk paths (the gap is closed)
- SCIM bot → updates
- `overrideMetadata=true` → still force-syncs

This also resolves the asymmetry you flagged: an ingestion bot's per-entity PUT now preserves **both** description and displayName (the `bulkOperation` version overwrote displayName while preserving description). The `bulkOperation` flag is removed.

### Finding 2 (Quality) — fixed

Split into two tests that exercise the actual policy distinction the guard now keys on:

- `test_singleEntityPut_bot_preservesUserDisplayName` — the **ingestion** bot (EditDisplayName denied) does a per-entity PUT and the user value MUST survive. This is the regression guard for exactly the bug in Finding 1 — RED on the `bulkOperation` revision, GREEN now.
- `test_singleEntityPut_displayNameAllowedBot_updatesDisplayName` — a bot whose policy allows `EditAll` with **no** `DisplayName-Deny` (constructed by stripping the auto-assigned `DefaultBotRole`), modeling the SCIM bot, updates displayName via a per-entity PUT.

The real SCIM bot reaches `updateDisplayName` through the SCIM endpoint, which bypasses entity-level authorization and is enterprise-only; the second test drives the same repository code path with an equivalent (policy-allowed) bot identity. All four displayName tests pass against the testcontainers stack on `DatabaseResourceIT`.

anuj-kumary (approved):

PR #28702, opened 2026-06-04
fix(apps): sync IngestionPipeline scheduleInterval on app schedule changes

### LINKED ISSUE #28769: External app schedule not synced to IngestionPipeline when changed to on-demand
## Problem

When a customer changes an external app (e.g., Metadata Exporter) from scheduled to on-demand/manual, the backing IngestionPipeline continues to run on the old schedule.

**Root cause:** The app's schedule is stored in two places:
1. `App.appSchedule` (updated correctly by the UI)
2. `IngestionPipeline.airflowConfig.scheduleInterval` (NOT updated, remains stale)

Runner clients (K8s, Argo, Hybrid) read `scheduleInterval`, so they keep running the scheduled job even though the app is now marked as on-demand.

## Example

1. Customer creates Metadata Exporter app with cron schedule: `17 4 * * *` (daily at 4:17 AM)
   - `App.appSchedule.cronExpression` = `"17 4 * * *"`
   - `IngestionPipeline.airflowConfig.scheduleInterval` = `"17 4 * * *"` ✓ (synced)

2. Customer changes app to on-demand via UI:
   - Sets `appSchedule.scheduleTimeline` = `None`
   - Clears `appSchedule.cronExpression`
   - `App.appSchedule` updated ✓

3. But `IngestionPipeline.airflowConfig.scheduleInterval` still = `"17 4 * * *"` ❌

4. Runner clients see non-null `scheduleInterval`, so they continue running the job daily at 4:17 AM, even though the app is marked on-demand.

## Related Issues

- Affects: Metadata Exporter, Custom External Applications
- Impact: Scheduled jobs continue running after user marks app as manual/on-demand
- Severity: Medium (jobs run unexpectedly, but user can manually delete the schedule in runner)

## Solution

PR #28702 adds sync logic to ensure `IngestionPipeline.airflowConfig.scheduleInterval` is always in sync with `App.appSchedule`:
- Extract `deriveInterval(AppSchedule)` helper to derive pipeline schedule from app schedule
- Call from both pipeline creation and update paths
- Handle edge cases: null appSchedule, `ScheduleTimeline.NONE`, null AirflowConfig on old pipelines

edg956:
## Summary
Fixes #28769

- Fix app schedule sync bug where IngestionPipeline.airflowConfig.scheduleInterval was not updated when app schedule changed
- Extract `deriveInterval(AppSchedule)` helper to convert cronExpression to scheduleInterval (null if blank)
- Call from both `createAndBindIngestionPipeline` and `updateAppConfig` so pipeline schedule always matches app schedule
- Runner clients (K8s/Argo/Hybrid) read scheduleInterval, so clearing it stops scheduled jobs

## Root Cause
When customer changed Metadata Exporter app from scheduled to on-demand:
- App.appSchedule was cleared (cronExpression → null)
- But IngestionPipeline.airflowConfig.scheduleInterval retained the old cron
- Runner clients saw non-null scheduleInterval and continued running scheduled jobs

## Changes
1. **AbstractNativeApplication.java**:
   - Add `deriveInterval(AppSchedule)` helper: null schedule or blank cron → null, otherwise → cronExpression
   - Update `createAndBindIngestionPipeline` to use `deriveInterval()` instead of direct access
   - Add schedule sync in `updateAppConfig`: set `airflowConfig.scheduleInterval` to `deriveInterval()` result

2. **AppsResourceIT.java**:
   - `test_externalAppScheduleSync_clearingCronNullsOutPipelineInterval`: validates clearing cron nulls pipeline interval
   - `test_externalAppScheduleSync_changingCronUpdatedPipelineInterval`: validates changing cron updates pipeline

## Test Plan
- [ ] Run `mvn test -pl openmetadata-integration-tests -Dtest=AppsResourceIT` to verify new tests pass
- [ ] Run existing app resource tests to verify no regressions

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/AppsResourceIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/AppsResourceIT.java
@@ -1280,6 +1280,106 @@ public class AppsResourceIT {
     }
   }
 
+  @Test
+  void test_externalAppScheduleSync_removingAppScheduleNullsOutPipelineInterval(TestNamespace ns)
+      throws Exception {
+    HttpClient httpClient = SdkClients.adminClient().getHttpClient();
+    String appName = ns.prefix("extSchedRemove");
+    String initialCron = "17 4 * * *";
+
+    try {
+      AppMarketPlaceDefinition marketPlaceDef = createExternalAppMarketplace(httpClient, appName);
+
+      CreateApp createApp =
+          new CreateApp()
+              .withName(marketPlaceDef.getName())
+              .withAppConfiguration(marketPlaceDef.getAppConfiguration())
+              .withAppSchedule(
+                  new AppSchedule()
+                      .withScheduleTimeline(ScheduleTimeline.CUSTOM)
+                      .withCronExpression(initialCron));
+
+      httpClient.execute(HttpMethod.POST, "/v1/apps", createApp, App.class);
+
+      String pipelineFqn = "OpenMetadata." + appName;
+      IngestionPipeline pipeline =
+          SdkClients.adminClient().ingestionPipelines().getByName(pipelineFqn);
+      assertEquals(
+          initialCron,
+          pipeline.getAirflowConfig().getScheduleInterval(),
+          "Bound pipeline should have the initial cron schedule");
+
+      App installedApp = Apps.getByName(appName);
+      String appId = installedApp.getId().toString();
+      String removeSchedulePatch = "[{\"op\":\"remove\",\"path\":\"/appSchedule\"}]";
+
+      httpClient.executeForString(
+          HttpMethod.PATCH,
+          "/v1/apps/" + appId,
+          removeSchedulePatch,
+          RequestOptions.builder().header("Content-Type", "application/json-patch+json").build());
+
+      IngestionPipeline pipelineAfterRemove =
+          SdkClients.adminClient().ingestionPipelines().getByName(pipelineFqn);
+      assertNull(
+          pipelineAfterRemove.getAirflowConfig().getScheduleInterval(),
+          "Pipeline scheduleInterval should be null after removing appSchedule");
+
+    } finally {
+      try {
+        Apps.uninstall(appName, true);
+      } catch (Exception ignored) {
+      }
+    }
+  }
+
+  @Test
+  void test_externalAppScheduleSync_noneTimelineNullsOutPipelineInterval(TestNamespace ns)
+      throws Exception {
+    HttpClient httpClient = SdkClients.adminClient().getHttpClient();
+    String appName = ns.prefix("extSchedNone");
+    String initialCron = "17 4 * * *";
+
+    try {
+      AppMarketPlaceDefinition marketPlaceDef = createExternalAppMarketplace(httpClient, appName);
+
+      CreateApp createApp =
+          new CreateApp()
+              .withName(marketPlaceDef.getName())
+              .withAppConfiguration(marketPlaceDef.getAppConfiguration())
+              .withAppSchedule(
+                  new AppSchedule()
+                      .withScheduleTimeline(ScheduleTimeline.CUSTOM)
+                      .withCronExpression(initialCron));
+
+      httpClient.execute(HttpMethod.POST, "/v1/apps", createApp, App.class);
+
+      App installedApp = Apps.getByName(appName);
+      String appId = installedApp.getId().toString();
+      String setNoneTimelinePatch =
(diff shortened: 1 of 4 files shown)


ayush-shah:
Thanks for pushing this. This looks directionally right and it does cover the current UI-style path for clearing the Metadata Exporter cron (`scheduleTimeline=None` + removing `cronExpression`). I think there are still a couple of gaps worth tightening before we call the schedule-sync issue fully fixed.

1. `AbstractNativeApplication.install()` still returns before `scheduleExternal()` when `app.getAppSchedule() == null`:
   https://github.com/open-metadata/OpenMetadata/pull/28702/files#diff-10d32d2b6e0132740400f28ab9c4f452f09b63378bf0ecb469cc27b566e78e5cR77-R80

   Since `appSchedule` is optional in the app schema and patchable through `AppRepository.UPDATE_FIELDS`, an API client can remove `/appSchedule` entirely. In that case, the new null-safe `deriveInterval(null)` helper is never reached, and the existing backing `IngestionPipeline.airflowConfig.scheduleInterval` can remain stale.

2. `deriveInterval()` currently derives the pipeline schedule only from `cronExpression` and ignores `scheduleTimeline`:
   https://github.com/open-metadata/OpenMetadata/pull/28702/files#diff-10d32d2b6e0132740400f28ab9c4f452f09b63378bf0ecb469cc27b566e78e5cR200-R208

   That means `scheduleTimeline=None` with a stale/non-empty `cronExpression` would still keep the external pipeline scheduled. It also means non-custom timelines like `Hourly`, `Daily`, etc. become `null` unless a cron expression is present. Internal app scheduling already has explicit timeline semantics in `AppScheduler.getCronSchedule(...)`, so external app conversion is still a bit narrower than the app schedule model.

I may be missing a stronger UI/API invariant that always removes `cronExpression` and never removes `appSchedule`; if that invariant exists, please point me to it. Otherwise I think we should make the external pipeline sync resilient to both cases and add regression coverage for the API shape.

Suggested AI-agent prompt to resolve this:

```text
You are working on OpenMetadata PR #28702: "fix(apps): sync IngestionPipeline scheduleInterval on app schedule changes".

Context:
- External Native Applications create/update a backing IngestionPipeline in AbstractNativeApplication.
- The PR currently adds deriveInterval(AppSchedule) and writes the derived value into IngestionPipeline.airflowConfig.scheduleInterval.
- The customer-facing bug is that changing an app to manual/no schedule can leave the backing ingestion pipeline scheduled, so Argo/runner continues to trigger it.

Please tighten the fix with a small, focused patch:

1. In AbstractNativeApplication.install(...), do not skip the external app pipeline sync solely because app.getAppSchedule() is null.
   - For deleted apps, NoSchedule, and OnlyManual, keep the existing behavior unless there is a clear reason to change it.
   - For External apps with Scheduled/ScheduledOrManual scheduleType, scheduleExternal(installedBy) should still run so the backing IngestionPipeline can be updated to scheduleInterval = null.
   - The goal is that removing /appSchedule through the API clears the backing pipeline schedule instead of leaving the old cron behind.

2. Make schedule-to-pipeline conversion explicit.
   - null appSchedule => null scheduleInterval
   - ScheduleTimeline.NONE => null scheduleInterval, even if cronExpression is still present
   - ScheduleTimeline.CUSTOM => use non-blank cronExpression, otherwise null or a clear validation error consistent with existing app behavior
   - For Hourly/Daily/Weekly/Monthly, either convert to the cron format expected by IngestionPipeline.airflowConfig.scheduleInterval or explicitly keep current behavior and document/test it.
   - Be careful not to blindly reuse Quartz six-field cron strings if ingestion pipeline scheduleInterval expects Unix/Airflow-style five-field cron.

3. Add regression tests in AppsResourceIT:
   - Existing clear-cron UI path should still pass: replace /appSchedule/scheduleTimeline with None and remove /appSchedule/cronExpression => backing pipeline scheduleInterval is null.
   - Add a new API-shaped case: remove /appSchedule entirely from an external ScheduledOrManual app with an existing cron => backing pipeline scheduleInterval becomes null.
   - Add coverage for ScheduleTimeline.NONE with a stale cronExpression if that payload is valid through PATCH: scheduleInterval should still become null.

4. Adjust the redeploy-related test/comment.
   - If the test only asserts persisted scheduleInterval, do not claim it proves requiresRedeployment/deploy behavior.
   - If we want to prove redeploy behavior, make the pipeline deployed first and assert the deploy path/client call through the existing test utilities.

Keep the change scoped to external app schedule sync. Do not refactor unrelated app install/scheduler behavior unless required by the tests above.
```

IceS2 (dismissed):

IceS2 (dismissed):

IceS2 (dismissed):

ayush-shah (dismissed):

IceS2 (approved):

PR #28711, opened 2026-06-04
fix(search): set dynamic:false on all index mappings to stop uncontrolled dynamic mapping

mohityadav766:
## Problem

Production reindex failure :

```
failed to parse field [pipelineStatuses.config.appConfig.actions.customProperties]
of type [text] in document ... Preview of field's value: '{columndatepivot=jointure}'
```

`pipelineStatuses` (and other arbitrary config sub-trees like `config`/`appConfig`) had **no `dynamic` setting**, so ES/OS dynamically mapped unexpected fields. A **string** value auto-typed `customProperties` as `text`; a later **object** value (`{columndatepivot=jointure}`) then failed to parse → `mapper_parsing_exception`. Uncontrolled dynamic mapping is also the root cause of field-count explosion.

This is the "stop dynamic mapping completely" hardening discussed for indexing safety.

## Fix

Add **`"dynamic": false`** at the `mappings` root of **every** index mapping (all locales, 241 files). Unexpected fields are stored in `_source` but never auto-typed or indexed, so:
- no type conflicts (string-then-object),
- no field-count explosion from arbitrary keys.

All searchable/aggregatable fields stay **explicitly** mapped. Report-data/analytics indices map their `data` fields explicitly (verified), so nothing relied on dynamic indexing.

`dynamic:false` (not `strict`) is used deliberately — `strict` would *reject* documents carrying config; `false` is non-destructive (stores, doesn't reject).

## Validation

Reproduced and verified against a real OpenSearch node:
- `DynamicMappingTypeConflictTest` — reproduces the exact `text` vs object parse failure on a dynamic mapping, and proves `dynamic:false` (and the **real** `ingestion_pipeline` mapping) accept both string and object config values.
- `IndexMappingNestedFieldConsistencyTest.allIndexMappingsMustDisableDynamic` — guards that every mapping keeps `dynamic:false`.

Existing mapping/search tests pass: `IndexMappingVersionTrackerTest`, `SearchIndexReindexFieldsParityTest`, `SearchSourceBuilderFactoryTest`, `IndexMappingNestedFieldConsistencyTest`.

## Notes

- Changing mappings bumps the mapping version → triggers a **reindex on deploy** (expected).
- Complementary to the immense-term hardening PR #28671 (different root cause; both touch the mapping files but on different keys — `mappings.dynamic` here vs `mappings.properties.extension` there).


harshach:
@mohityadav766 lets file a github issue that can be reproducible. We are not going to merge this. Without clear evidence and testing we are going to add more baggage

PR #28764, opened 2026-06-05
refactor(mcp): shared trim util + global response size-budget net

Vishnuujain:
De-duplicates payload-trimming logic across MCP tools and adds a dispatch-level size cap so every tool is bounded.

### What
- **New `McpResponseTrim`** — single home for the truncation budgets, the 100k size cap, the truncate helpers, and the embedding/vector field list (previously copy-pasted across 4 tools).
- **New `McpParams`** — shared `getInt/getBoolean/getDouble` arg coercion (replaces 3 private copies).
- **Global size net** in `DefaultToolContext` — serializes the result once; if it exceeds 100k chars, returns a compact `truncated:true` envelope. This is the floor for tools with no per-tool trim (`get_entity_details`, `get_test_definitions`).
- **Error messages null-guarded** — broad catches no longer surface `"... null"` when an exception has no message.

### Notes
- Behavior-preserving: the two truncate conventions are kept distinct; per-tool smart trims (search proportional drop, lineage/RCA actionable hints) are unchanged — the global net only adds a floor.
- Backend repositories and the ES lineage paths are untouched.
- 98 unit tests pass; live-verified against a running MCP server (no payload-shape change, no vector-key leaks).


openmetadata-mcp/src/main/java/org/openmetadata/mcp/util/McpParams.java
```diff
@@ -0,0 +1,71 @@
...
+ * Lenient coercion of MCP tool arguments. MCP clients send arguments as an untyped {@code
+ * Map<String, Object>} where a numeric value may arrive as a JSON number <em>or</em> a string, so
+ * every read tool was carrying its own {@code parseIntParam}/{@code parseBooleanParam}/{@code
+ * parseDoubleParam} copy. This centralizes that logic.
+ *
+ * <p>All three readers share the same fall-through contract: a missing key, a present-but-null value
+ * and an unparseable value all yield the supplied default. Range clamping (depth/size bounds) stays
+ * with the caller because the valid range is tool-specific.
+ */
+public final class McpParams {
```
pmbrull, line 15: lets move McpParams and McpResponseTrim outside of the tools package. This way each class inside tools is a proper tool, and we can organize it better

pmbrull (approved):

PR #28793, opened 2026-06-08
Fixes #28659: typed, declarative, diagnosable Test Connection (MySQL)

### LINKED ISSUE #28659: Improve Test Connection
Problem

  The current Test Connection engine:
  - crashes (KeyError) when a definition step has no executor. Taking down the whole test, including steps that would pass;
  - returns opaque errors: a generic static message even when the driver told us the real cause (missing grant, bad host, auth failure);
  - returns only passed: bool, so the new connection-status UI (#28569) has no real per-step status, command, timing, or log to show.

  Goal

  Rework Test Connection so a connector declares its checks declaratively and the framework:
  - never crashes on definition/executor mismatch (degrades to a visible "skipped");
  - turns driver errors into actionable diagnoses (what's wrong + how to fix) while keeping the raw log;
  - emits a richer per-step result (status, timing, executed command, summary, structured log) to back the new UI.

  ▎ Design is still being iterated — details may change.

  Related

  - #28569 (connector setup UI / gate-capability taxonomy)

IceS2:
## What
An engine-agnostic Test Connection framework under `core/connections/test_connection/`, with MySQL migrated to it as the first connector.

- **Declarative checks** — a connector exposes checks as `@check(Step)` methods; `TestConnectionRunner` runs them in the order the service definition declares.
- **Gating** — a `category: ConnectionGate` step (e.g. CheckAccess) short-circuits the rest as `ConnectionNotEstablished` once it fails; a missing check is a visible `Skipped`, never a crash.
- **Diagnosis** — each connector owns an `ErrorPack` of first-match rules mapping a driver exception (matched across its `__cause__`/`__context__` chain) to an actionable `Diagnosis`; the raw `errorLog` is always retained.
- **Evidence** — checks self-report the command they ran and a short summary; a per-step timeout bounds each check.
- **Lifecycle** — `BaseConnection` owns its client: lazy build, an `ExitStack` of teardowns unwound by `close()`, context-manager support. MySQL selects a per-auth `EngineStrategy` (basic / Azure AD / CloudSQL).

## Schema
- `testConnectionStepResult` gains `status`, `diagnosis`, `durationMs`, `executedCommand`, `resultSummary`; `logEntries` removed (unused).
- `testConnectionDefinition` steps gain a `category` (`ConnectionGate` / `Capability`).
- Automation-workflow PATCH responses serialize via `model_dump(mode="json")` so the nested step enums round-trip (previous code only `.value`-converted the top-level status).

## Scope
MySQL-first strangler. Remaining connectors and the diagnosis UI are follow-ups; legacy connectors keep the existing flow until migrated.

Fixes #28659


ingestion/src/metadata/core/connections/test_connection/checks/database.py
```diff
@@ -0,0 +1,188 @@
...
+) -> Evidence:
+    """Run a statement and report the command plus a summary of its rows.
+
+    On failure, re-raise as ``CheckError`` carrying the attempted command, so the
+    failed step still reports what it ran.
+    """
+    with _captured_sql(client) as captured:
+        try:
+            with client.connect() as conn:
+                rows = conn.exec_driver_sql(statement).fetchall()
```
pmbrull: could this flow up mem potentially with fetchall?

openmetadata-service/src/main/resources/json/data/testConnections/database/mysql.json
```diff
@@ -8,6 +8,7 @@
       "description": "Validate that we can properly reach the database and authenticate with the given credentials.",
       "errorMessage": "Failed to connect to mysql, please validate the credentials",
       "shortCircuit": true,
+      "category": "ConnectionGate",
```
pmbrull, line 11: do we need this if we already have the shortCircuit?

IceS2: This is from the new jsonschema on the UI design to be honest. I think they are redundant at the moment. category feels a bit more generic, but I don't have strong feelings either way

ingestion/src/metadata/core/connections/test_connection/checks/database.py
IceS2: Yes! checking that out now

pmbrull (approved):

aniketkatkar97 (approved):

PR #28831, opened 2026-06-08
feat(mssql): use Query Store for durable lineage and usage when available

Khairajani:
## What

MSSQL lineage and usage read query history from the volatile `sys.dm_exec_*` plan cache, which is evicted on restart or under memory pressure — so query history (and the lineage built from it) is lost over time.

This adds **Query Store** (`sys.query_store_*`) as an **opt-in** durable query-history source, enabled via the `useQueryStore` connection flag (default off). When off, the connector uses the plan cache exactly as before.

## How it decides

- **Opt-in:** used only when the connection's `useQueryStore` flag is on (default off). When off, no Query Store probe runs and the plan cache is used exactly as before.
- Probe `sys.database_query_store_options` once per database (result cached for the run).
- Use Query Store only when it is **active** (`READ_ONLY`/`READ_WRITE`) **and** in **`ALL`** capture mode; otherwise fall back to the plan cache.
- `AUTO` capture mode is treated as not-usable: it skips ad-hoc / infrequent queries, so the plan cache (which still holds recently executed queries) gives more complete lineage there.
- A log line reports the chosen source, e.g. `MSSQL lineage/usage query source: Query Store (durable history)` or `plan cache (dm_exec_*)`.

## Notes

- The Query Store statement mirrors the plan-cache statement's output columns, time window, OpenMetadata/dbt comment filters and result limit. It excludes internal/parameterized queries to match the plan cache's `objtype != 'Prepared'` behaviour.
- Query Store views are per database, so the durable path is used **only for single-database services** (`ingestAllDatabases = false`); multi-database services keep the server-wide plan cache so other databases' queries are not dropped.
- Falls back safely on SQL Server < 2016 (the view is absent) and on any probe error.
- Scope: query/table lineage and usage. View and stored-procedure lineage are unchanged.

## Tests

- Unit: source detection (`READ_WRITE`/`READ_ONLY` + `ALL` → Query Store; `AUTO`/`OFF`/no-row/probe-error → plan cache), statement selection, and statement rendering for both lineage and usage filters.
- Integration (`sql_server/test_query_store.py`): asserts lineage survives `DBCC FREEPROCCACHE` via Query Store.


mohittilala:
This has been addressed as part of https://github.com/open-metadata/OpenMetadata/pull/29719

PR #28851, opened 2026-06-09
fix(audit): defer row fetch in audit logs list query to avoid full-row scan

### LINKED ISSUE #28850: Audit Logs list API (GET /api/v1/audit/logs) is extremely slow on large audit_log_event tables
## Describe the bug

`GET /api/v1/audit/logs` becomes extremely slow on large `audit_log_event` tables (millions of rows). A single page (`limit=25`) over a 24-hour window can take **90+ seconds**, with ~99% of the time spent in the database.

This is **not** caused by the usual suspects:
- The endpoint already passes a bounded `limit` (`@DefaultValue("25") @Min(1) @Max(200)`).
- The `event_ts` index (`idx_audit_log_event_ts (event_ts DESC)`) exists and is used (`type=range`).
- The per-request `COUNT(id)` is index-covering and fast (~20-40 ms) — it is not the bottleneck.

## Root cause

The list query selects **all** columns — including the large `event_json` (`LONGTEXT`, ~16 KB/row) — with `ORDER BY event_ts DESC, id DESC LIMIT n`.

To satisfy `ORDER BY + LIMIT`, the engine performs a **non-covering** index range scan: it reads the **full row** (including `event_json`) for *every* row in the time window before trimming to `limit`. For a window of ~100k+ rows that is ~100k full-row reads per page. The engine does not defer the heavy column fetch until after the limit is applied.

`EXPLAIN ANALYZE` over a ~111k-row window:

| Query | Full rows read | Time |
|---|---|---|
| Current list query | ~111,747 | ~153,000 ms |
| `COUNT` over the same rows (index-only) | 0 | ~21 ms |
| Deferred-join rewrite | 26 | ~180 ms |

Traversing the same rows index-only is ~21 ms; reading the full rows is ~153 s; deferring the row fetch to only the final page is ~0.18 s. Reading ~112k full rows is the entire cost.

## Fix

Deferred join / late row lookup — resolve the page of `id`s from the index first, then join back for the full columns:

```sql
SELECT a.<cols>
FROM audit_log_event a
JOIN (SELECT id FROM audit_log_event <condition> <orderClause> LIMIT :limit) k
  ON a.id = k.id
<orderClauseQualified>
```

The inner subquery is index-only over `idx_audit_log_event_ts` (no `event_json` read); only the final `n` rows are ever materialized. Returns an **identical result set and ordering** — `id` is the primary key, so the join is strictly 1:1 and `(event_ts, id)` is a total order, making the `LIMIT` deterministic and the post-join re-sort exact. Standard SQL — valid on both MySQL and PostgreSQL.

## Affected code

- `CollectionDAO.AuditLogDAO.list`
- `AuditLogRepository.list` (forward + backward cursor pagination)
- `exportInBatches` (delegates to `list`, fixed automatically)

## Follow-ups (not part of this fix)

- Unindexed filters on `entity_type` / `event_type` still scan rows within the window — add `(entity_type, event_ts)` / `(event_type, event_ts)` composite indexes if used at scale.
- The per-request total `COUNT` is optional for cursor pagination and could be made opt-in.
- Table size / disk reclamation is a separate retention concern (`DataRetention` app + `OPTIMIZE TABLE`).

## Severity

Performance — only manifests at scale (millions of rows); does not reproduce at small data volumes.

yan-3005:
Fixes #28850

## Problem

`GET /api/v1/audit/logs` is extremely slow on large `audit_log_event` tables (millions of rows). A single page (`limit=25`) over a 24-hour window can take **90+ seconds**, ~99% of it in the database — even though the request passes a bounded `limit` and the `event_ts` index exists and is used.

## Root cause

The list query selects **all** columns, including the large `event_json` (`LONGTEXT`, ~16 KB/row), with `ORDER BY event_ts DESC, id DESC LIMIT n`. To satisfy `ORDER BY + LIMIT`, the engine does a **non-covering** index range scan — it reads the full row (including `event_json`) for *every* row in the time window, then trims to `limit`. For a ~112k-row window that is ~112k full-row reads per page, which is the entire cost.

`EXPLAIN ANALYZE` over a ~111k-row window:

| Query | Full rows read | Time |
|---|---|---|
| Current list query | ~111,747 | ~153,000 ms |
| `COUNT` over the same rows (index-only) | 0 | ~21 ms |
| Deferred-join rewrite | 26 | ~180 ms |

## Fix

Deferred join / late row lookup — resolve the page of `id`s from the index first (index-only, no `event_json` read), then join back for the full columns of only the final page:

```sql
SELECT a.<cols>
FROM audit_log_event a
JOIN (SELECT id FROM audit_log_event <condition> <orderClause> LIMIT :limit) k
  ON a.id = k.id
<orderClauseQualified>
```

- **Inner subquery** keeps the existing `<condition>` and all `@Bind` params unchanged; it is an index-only scan over `idx_audit_log_event_ts` that picks the top-N `id`s.
- **Outer `ORDER BY`** must be qualified (`a.event_ts`, `a.id`) because `id` is ambiguous after the join — added `ORDER_DESC_QUALIFIED` / `ORDER_ASC_QUALIFIED` alongside the existing inner-scope `ORDER_DESC` / `ORDER_ASC`. Both directions are needed (backward pagination sorts ASC then reverses in Java).
- Single shared `@SqlQuery`, valid on both MySQL and PostgreSQL (standard SQL).

### Why results are identical (not just faster)

- `id` is the **primary key** (unique, non-null) → the join is strictly **1:1**: no rows dropped, none duplicated.
- The inner applies the same `WHERE` / `ORDER BY` / `LIMIT` as before, and `(event_ts, id)` is a **total order**, so the top-N is deterministic and identical to the old query.
- A `JOIN` does not preserve subquery order, so the outer re-sorts by the **same keys** — reproducing the original order exactly.

This is purely a performance change; the result set and ordering are unchanged. The only load-bearing assumption is `id` being a unique PK, which the schema guarantees.

## Changes

- `CollectionDAO.AuditLogDAO.list` — deferred-join SQL + new `@Define("orderClauseQualified")`.
- `AuditLogRepository` — added `ORDER_DESC_QUALIFIED` / `ORDER_ASC_QUALIFIED` and passed them at both `list(...)` call sites (forward + backward pagination). `exportInBatches` delegates to `list`, so it is fixed automatically.

## Tests

Added two integration tests in `AuditLogResourceIT` that exercise the deferred join with real multi-page data:

- `test_listAuditLogs_deferredJoin_forwardPaginationOrderingIsConsistent` — seeds an audit-event burst, pages forward through a fixed-`endTs` window, and asserts strict `(eventTs DESC, id DESC)` ordering and no duplicate ids across pages (covers `ORDER_DESC_QUALIFIED` + the join).
- `test_listAuditLogs_deferredJoin_backwardPaginationMatchesForward` — pages forward to obtain a `before` cursor, then backward, and asserts the backward page reproduces the forward page in identical order (covers `ORDER_ASC_QUALIFIED`).

These are correctness/regression guards. A "fails-without-the-fix" test is intentionally **not** included: the old query is correct, just slow, and the slowness only reproduces at multi-million-row scale that CI does not have — so a timing assertion would pass on the un-optimized query too and would be meaningless. The tests instead guard the real risk introduced by the rewrite (dropped/duplicated/reordered rows, ambiguous-`id` SQL), on both MySQL and PostgreSQL.

## Follow-ups (not in this PR)

- Unindexed filters (`entity_type`, `event_type`) still scan within the window — add `(entity_type, event_ts)` / `(event_type, event_ts)` composite indexes if used at scale.
- Per-request total `COUNT` could be made opt-in for cursor pagination.
- Table size / disk reclamation is a separate retention concern (`DataRetention` + `OPTIMIZE TABLE`).


harshach (approved):

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -14679,13 +14679,16 @@ interface AuditLogDAO {
     int insert(@BindBean AuditLogRecord record);
 
     @SqlQuery(
-        "SELECT id, change_event_id, event_ts, event_type, user_name, "
-            + "actor_type, impersonated_by, service_name, "
-            + "entity_type, entity_id, entity_fqn, entity_fqn_hash, event_json, search_text, created_at "
-            + "FROM audit_log_event <condition> <orderClause> LIMIT :limit")
+        "SELECT a.id, a.change_event_id, a.event_ts, a.event_type, a.user_name, "
```
harshach, line 14682: why we need this join @yan-3005

harshach (changes requested):

yan-3005:
Cherry picked to 1.12.12: https://github.com/open-metadata/OpenMetadata/commit/55ab7f30e40f1a70d5a6229d486a98f4b286dd6a

PR #28856, opened 2026-06-09
feat(search): reindex-drift + index health checks in /system/validate

### LINKED ISSUE #28849: Search - Add index drift validation & multi-lang tests
- we do not have a proper system resource validation on index status, only "ES/OS is reachable"
- we do not have proper behavioral tests for different lang indexes, and it's easy for them to drift

pmbrull:
fix https://github.com/open-metadata/OpenMetadata/issues/28849

Screenshot 2026-06-09 at 11 53 08

## Summary
Adds a **"Search Reindex Status"** step to `/system/validate` that detects when a deployed search index was not built from the current code index-mapping — so operators know a reindex is pending (stale search silently breaks RBAC / Data Quality / facets).

- `IndexMappingVersionTracker.computeDrift()` classifies each entity **CURRENT / STALE / UNTRACKED** by comparing the current code-mapping hash to the hash stored when the index was last (re)built.
- The stored hash is now stamped at **every successful reindex seam**, so the signal is reliable — `SearchIndexApp.execute()` (UI / scheduled / CLI) and `SearchRepository.createIndexes()` (fresh install / drop-create), gated on a `COMPLETED` run and only the entities actually rebuilt. The redundant CLI-only stamp in `OpenMetadataOperations` is removed.
- The check **fails** on stale or missing indexes (both reindex-fixable) and additionally **reports** orphan indexes (zero-alias rebuild leftovers) and a degraded cluster as non-failing warnings.

Replaces an earlier curated-field "canary" check that only inspected a hardcoded subset of fields on a hardcoded subset of entities.

## Test Plan
- [x] `IndexMappingVersionTrackerTest` (13), `SystemRepositoryReindexStatusTest` (8), `SearchRepositoryBehaviorTest` (110) — all green; `mvn -pl openmetadata-service spotless:check` clean.
- [ ] Manual on a live stack: run a reindex, confirm `/system/status` "Search Reindex Status" passes and `index_mapping_versions` is populated; tamper one stored hash and re-check → that entity reported as reindex-pending.


openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/SearchIndexApp.java
```diff
@@ -63,6 +68,22 @@ public void execute(JobExecutionContext ctx) {
     this.orchestrator = orch;
     orch.run(jobData);
     this.jobData = orch.getJobData();
+    stampReindexedMappings(this.jobData);
+  }
+
+  private void stampReindexedMappings(EventPublisherJob job) {
```
mohityadav766: this should be done in DefaultRecreateHandler i think , we have that to handle promotion and finalization of staged indexes

PR #28884, opened 2026-06-09
refactor(ui): make design tokens dark-mode-ready (breakpoint 1)

chirag-madlani:
## Breakpoint 1 — design-token foundation for dark mode

First of a stack of PRs adding dark theme. This one is the **token plumbing only** — it aligns the three styling layers (Tailwind/core-components, Ant Design, legacy Less) onto a single `--color-*` source of truth so dark mode can propagate without per-library color tables.

**No toggle switch / activation in this PR** — these tokens are inert and light mode is unchanged until dark mode is wired up in a later PR.

### Root cause this fixes
The app compiles core-components Tailwind with `prefix(tw)`, which **freezes** each `@theme` color into `--tw-color-*` at build time. The existing `.dark-mode` block sets the *unprefixed* `--color-*` names — which nothing reads at runtime. So dark-mode overrides were effectively dead.

### Changes
- **`globals.css`** — wrap the 374 `@theme` tokens that `.dark-mode` overrides as `var(--token, <original>)`. Tailwind preserves the `var()` into `--tw-color-*`, so the existing `.dark-mode` block finally flips them at runtime. **Light mode is byte-identical** (the original value is the fallback).
- **`variables.less`** — alias the standard `@grey-50..900` scale (a byte-exact match to `--color-gray-*`), `@background-color` and `@background-primary` to the shared `--tw-color-*` tokens. Light unchanged, now dark-aware across all files that use the scale.
- **`ExploreSearchCard`** — hardcoded card bg/border/highlight/muted-text hex → semantic tokens (reference conversion).
- 3 `fade(@grey-*)` Less calls → `color-mix()` (Less color fns can't take `var()`).

### Verification
- Compiled the real app `tailwind.css`: **619 color vars compared light-mode vs. pre-change — zero differences.**
- Confirmed each touched token resolves at `:root` in light and flips under `.dark-mode`.
- All changed `.less` files + `app.less` compile cleanly through the `less` renderer.

> ⚠️ Not yet visually validated in a running app (no dark-mode trigger exists until the activation PR). Light mode is verified unchanged.


chirag-madlani:
Handled in #29136

PR #28897, opened 2026-06-10
fix(test): fix SearchSettings spec

### LINKED ISSUE #27946: Search Preview Inconsistent Behavior
### Affected module

Search / Discovery

### Describe the bug

Search configuration parameters results do not reset consistently after reverting values. For example increasing an n-gram weight from zero to five changes the preview behavior. The results are out of order and appear to be on a different scale or inconsistent scale, but setting it back to 0 does not restore the prior previous state. The preview then shows the results in an inconsistent and random order.

### To Reproduce

1. Go to Settings -> Preferences ->Search -> Database
2. Keep track of the search results when name.ngram weight is set to 0 
3. Update the name.ngram weight from 0 to 5. Search results order changes in the preview. Save it.
4. Update the search again from 5 to 0. Search results order is not the same as the results noticed in step 2

### Expected behavior

 If a weight is changed and then reverted to its original value, the previously returned results should remain consistent. Result ordering and displayed scores should reflect the reverted configuration without residual effects from prior settings.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

_No response_

_No response_

### Additional context

Impact: This makes the preview unreliable for validating tuning changes, because reverting a configuration value does not reliably restore the original preview behavior.

### Pre-submission checklist

satender-kumar-collate:
fixes: #27946 

Describe your changes: I fixed a bug in the search settings preview where reverting a field weight (e.g. name.ngram from 5 back to 1) or clicking "Restore Defaults" would leave the preview panel frozen on the previous state instead of re-fetching with the updated config.

#
### Type of change:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### High-level design:

N/A — small change. 

#
#### Manual testing performed

#

Not applicable. 

#
### Checklist:

----
## Summary by Gitar

- **Test stabilization:**
  - Refactored `SearchSettings.spec.ts` to improve reliability of `waitForResponse` listeners during preview updates.
  - Updated test logic to revert `n-gram` weight to a static default of `1` instead of tracking initial state, ensuring predictable assertion outcomes.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SearchSettings.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SearchSettings.spec.ts
@@ -316,31 +316,12 @@ test.describe('Search Settings', () => {
       }) => {
         await settingClick(page, GlobalSettingOptions.SEARCH_SETTINGS);
         const tableCard = page.getByTestId(mockEntitySearchSettings.key);
-
-        // Register waitForResponse BEFORE the click so the on-load preview
-        // request is captured rather than missed.
-        const initialPreviewPromise = page.waitForResponse((r) =>
-          r.url().includes('/api/v1/search/preview')
-        );
-
         await tableCard.click();
 
         await expect(page).toHaveURL(
           new RegExp(mockEntitySearchSettings.url + '$')
         );
         await waitForAllLoadersToDisappear(page);
-
-        // Capture the initial preview request to read original n-gram boost.
-        const initialPreviewResponse = await initialPreviewPromise;
-        expect(initialPreviewResponse.status()).toBe(200);
-        const initialBody = initialPreviewResponse.request().postDataJSON();
-
-        const initialNgramBoost =
-          initialBody?.searchSettings?.assetTypeConfigurations
-            ?.find((c: { assetType: string }) => c.assetType === 'table')
-            ?.searchFields?.find(
-              (f: { field: string }) => f.field === 'name.ngram'
-            )?.boost ?? 0;
 
         // Expand the name.ngram field configuration panel.
         const ngramPanel = page.getByTestId(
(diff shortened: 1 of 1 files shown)


Rohit0301 (approved):

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SearchSettings.spec.ts
```diff
@@ -360,25 +341,31 @@ test.describe('Search Settings', () => {
         await saveResponse;
         await toastNotification(page, /Search Settings updated successfully/);
 
-        // Revert n-gram weight back to its original value.
-        await setSliderValue(page, 'field-weight-slider', initialNgramBoost);
+        // Re-expand the panel after save — it may have collapsed during re-render.
+        await ngramPanel.click();
```
aniketkatkar97, line 345: We should not be making assumptions here. Are we sure the panel collapses after save? also why was this not failing before? what triggered it? We need to find the root cause instead of patch fixes.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SearchSettings.spec.ts
```diff
@@ -360,25 +341,31 @@ test.describe('Search Settings', () => {
...
-        // Wait for the preview to re-fetch with the reverted config.
-        const revertedPreviewResponse = await page.waitForResponse((r) =>
+        // Register the listener BEFORE moving the slider so the response is
+        // never missed regardless of how fast the preview fires.
+        const revertedPreviewPromise = page.waitForResponse((r) =>
           r.url().includes('/api/v1/search/preview')
         );
+
+        // Revert weight to 1 (the default from mockEntitySearchConfig).
+        await setSliderValue(page, 'field-weight-slider', 1);
```
aniketkatkar97, line 354: Initially we were using `initialNgramBoost` value to set for slider, why did we change it to use hardcoded value `?

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SearchSettings.spec.ts
satender-kumar-collate: the timeout was caused by a race condition where waitForResponse was attached after setSliderValue, potentially missing a fast preview response; the panel-collapse change was based on an incorrect assumption and has been removed

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SearchSettings.spec.ts
satender-kumar-collate: Restored initialNgramBoost. It was removed when the test was simplified to avoid handling the initial API response, but now that waitForResponse is registered before tableCard.click(), we can safely capture the initial boost value from that response and reuse it for both reverting the slider and the final assertion.

PR #28944, opened 2026-06-11
fix(sse): concatenate multi-line SSE data fields instead of overwriting

pmbrull:
## Problem

`SSEClient._parse_sse_event` keeps only the **last** `data:` line of an SSE event:

```python
elif line.startswith("data:"):
    event["data"] = line.split(":", 1)[1].strip()   # overwrites every data: line
```

Per the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) a single event's `data` field may span **multiple** `data:` lines, which must be concatenated with `\n`. Overwriting drops all but the last line, so any payload split across several `data:` lines is truncated mid-string. Downstream consumers then fail with `json.JSONDecodeError: Unterminated string` — observed in the dynamic-agent ingestion app (`metadata.applications.dynamic_agent.app`) when the AI Quality agent emits a large/over-escaped `createTestCase` payload, which aborts the run with 0 tests created.

## Fix

Accumulate all `data:` line values and `"\n".join` them. Single-`data:`-line events are unchanged (a join of one).

## Tests

Two regressions added to `test_sse_client.py`:
- `test_parse_sse_event_concatenates_multiple_data_lines` — multi-`data:` join semantics
- `test_parse_sse_event_reassembles_json_split_across_data_lines` — a JSON payload chunked across `data:` lines is reassembled so `json.loads` succeeds

Verified RED→GREEN on the real `_parse_sse_event`. (The full pytest module couldn't run in my local venv due to pre-existing dependency drift, so I exercised the method via an isolated harness; CI runs the suite normally.)

Note: this bug is identical on `1.12.10`, `1.13`, and `main` — worth forward-porting.

### DIFF
--- a/ingestion/src/metadata/ingestion/ometa/sse_client.py
+++ b/ingestion/src/metadata/ingestion/ometa/sse_client.py
@@ -168,16 +168,24 @@ class SSEClient:
             dict[str, Any]: The parsed event.
         """
         event: dict[str, Any] = {}
+        data_lines: list[str] = []
         for line in event_buffer:
             if line.startswith("event:"):
                 event["event"] = line.split(":", 1)[1].strip()
                 if "complete" in event["event"] or "error" in event["event"]:
                     self.stream_completed = True
             elif line.startswith("data:"):
-                event["data"] = line.split(":", 1)[1].strip()
+                # Per the SSE spec a single event's `data` field may span multiple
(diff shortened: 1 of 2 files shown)


pmbrull:
Superseded by #28945 against `main` (the bug is in main too). Cherry-pick to 1.12.11/1.13 from there.

PR #28952, opened 2026-06-11
fix(mcp): echo OAuth state without double-encoding in redirect URI

### LINKED ISSUE #28951: MCP OAuth: redirect URI double-encodes `state`, breaking loopback clients ("State does not match")
### Description

`UriUtils.constructRedirectUri` (used by the MCP OAuth `/mcp/authorize` → `/mcp/login` → client-callback redirect) double-encodes query parameters. It first `URLEncoder.encode`s each value, then passes the already-encoded query string into the multi-argument `new URI(scheme, authority, path, query, fragment)` constructor, which re-quotes the `%` characters.

For an opaque OAuth `state` containing URL-special characters (e.g. base64 padding/`+`):

```
state = "n+eBY2DPiNEk3xEe7rqmtg=="
encoded once : n%2BeBY2DPiNEk3xEe7rqmtg%3D%3D
re-quoted    : n%252BeBY2DPiNEk3xEe7rqmtg%253D%253D   <-- sent back to client
client decodes once -> n%2BeBY2DPiNEk3xEe7rqmtg%3D%3D   != original
```

Clients that compare `state` byte-for-byte reject the callback with **"State does not match"** and OAuth login fails.

### Affected clients

- **VS Code** (MCP loopback redirect `http://127.0.0.1:PORT/`, opaque base64 state with `=`/`+`) — **fails** on first login; only succeeds via its fallback `vscode.dev/redirect` URL-handler flow which normalizes the value.
- Any MCP client using a loopback redirect with base64 (non-url-safe) state, e.g. Claude Desktop bridges.

### Not affected

- **Cursor**, **Claude (claude.ai)** — their `state` is base64url with no special characters, so single vs. double encoding produce identical output.

### Fix

Build the redirect URL by string concatenation after a single `URLEncoder.encode`, instead of passing the encoded query back through the multi-arg `URI` constructor. State then round-trips byte-for-byte. Same code path is used by the SAML/SSO callback, so the fix applies there too.

Vishnuujain:
Fixes #28951

The MCP OAuth redirect double-encoded the `state` parameter, so clients that compare it byte-for-byte (e.g. VS Code's loopback flow) failed with **"State does not match"** on login. This builds the redirect URL so `state` round-trips unchanged.

Full root-cause analysis, affected/unaffected clients, and repro are in #28951.

### DIFF
--- a/openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/util/UriUtils.java
+++ b/openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/auth/util/UriUtils.java
@@ -1,7 +1,6 @@
 package org.openmetadata.mcp.server.auth.util;
 
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


pmbrull (approved):

pmbrull:
picked to 1.13

PR #29282, opened 2026-06-22
fix(openlineage): reset Kinesis poll inactivity timer per shard (1.12.10 backport)

Khairajani:
Backport of #29276 to **1.12.10**.

## Problem
The OpenLineage Kinesis source (`_poll_kinesis`) only ingests events from the **first shard** of a multi-shard stream; every subsequent shard is skipped (reported on a multi-shard `TRIM_HORIZON` stream).

## Root cause
`empty_response_time` was initialised once **before** the `for shard in shards:` loop. For a live shard, `get_records` always returns a non-null `NextShardIterator`, so the per-shard `while shard_iterator and empty_response_time <= session_timeout` loop only exits once `empty_response_time` exceeds `sessionTimeout`. That value was never reset between shards, so after the first shard drained, shards 2..N had the loop condition already `False` and were skipped — total ingested events == one shard's worth. Deterministic for any >1-shard stream (single-shard streams appeared to work).

## Fix
Reset `empty_response_time = 0.0` at the start of each shard iteration.

## Testing
Added `TestKinesisMultiShardPolling.test_poll_kinesis_reads_events_from_every_shard` — mocks a 2-shard stream (each shard returns one record then empty polls with a live `NextShardIterator`, forcing the timeout exit path) and asserts events from **both** shards are yielded. Verified on the main fix: **2 events with the fix, 1 without**.

Cherry-picked from #29276.

### DIFF
--- a/ingestion/src/metadata/ingestion/source/pipeline/openlineage/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/pipeline/openlineage/metadata.py
@@ -1124,7 +1124,6 @@ class OpenlineageSource(PipelineServiceSource):
             iterator_type = broker.consumerOffsets.value
             pool_timeout = broker.poolTimeout
             session_timeout = broker.sessionTimeout
-            empty_response_time = 0.0
 
             for shard in shards:
                 shard_id = shard["ShardId"]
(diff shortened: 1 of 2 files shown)


Khairajani:
Closing — no 1.12.10 backport needed; the fix lands in main (#29276) and ships in the next version.

PR #29297, opened 2026-06-22
fix(it): drain async bulk worker before TestNamespace cleanup

yan-3005:
## Problem

Three flakes observed on `integration-tests-mysql-elasticsearch` / `integration-tests-postgres-elasticsearch-redis`:

### Flake 1 — `MlModelResourceIT.testAutoPaginationFluentAPI` (404 on dead service)

```
ApiException (404): Entity not found: mlmodelService 3adf0926-7d24-43b1-ac26-9aa4acf47cc6
```

Failing test doesn't own that service — it belongs to `MlModelResourceIT.test_bulkAsync_returns202`. Async bulk worker leaks an orphan model under a service that `TestNamespace` cleanup has already hard-deleted; later, any unfiltered `listEntities` call from a concurrent test trips on it.

### Flake 2 — `TestCaseResourceIT.test_listTestCasesFilteredByOwner` (404 on dead table)

```
ApiException (404): Entity not found: table pg_<shortId>.db_<shortId>.s_<shortId>.t_<shortId>
```

Test issued a global, unfiltered `testCases().list(limit=100)`. Returned an orphan TestCase whose parent table was hard-deleted by another concurrent test. TestCase stores `Relationship` rows only to TestSuite + TestDefinition; its `entityLink` to the parent table is opaque to cascade hardDelete, so the orphan persists.

### Flake 3 — `AuditLogResourceIT.test_listAuditLogs_deferredJoin_backwardPaginationMatchesForward` (forward ≠ backward)

```
Backward pagination must return the first page in identical DESC order as forward
expected: [29303, 29301, 29300, 29302, 29299]
but was:  [29301, 29300, 29308, 29302, 29299]
```

The new deferred-join pagination tests (added in #28850) bound the result set only by `[startTs, endTs]`. Other concurrent test classes keep writing audit events into the same window, so the forward read and the backward read see different snapshots. Row id `29308` was inserted between the two reads and slipped into the backward page.

### Flake 1
`BaseEntityIT.test_bulkAsync_returns202`:

1. Posts a bulk create with `async=true`, returns `HTTP 202` immediately.
2. Asserts status and `numberOfRowsProcessed` — done.
3. `TestNamespace` `@AfterEach` issues `recursive=true&hardDelete=true` on the parent service.

The async worker has not yet committed the new entity (or its service→entity relationship row). Recursive cascade scans children before the worker lands, deletes the service, and the worker subsequently persists an orphan entity whose parent no longer exists. Every `BaseEntityIT` subclass with `supportsBulkAPI=true` is exposed.

The original draft of this PR guarded the await on `supportsSearchIndex`, but `SearchIndexResourceIT` and `StoredProcedureResourceIT` both set `supportsBulkAPI=true` + `supportsSearchIndex=false`, so the race was still open for them. Final fix uses each entity's own FQN (carried in the 202 body's `successRequest`) and `getEntityByName`, removing the search-index dependency.

### Flake 2
TestCase's only `Relationship.CONTAINS` rows point at TestSuite and TestDefinition. The link to its parent table is captured in `entityLink` (a free-form string), so a recursive hardDelete of the table doesn't cascade to the test cases attached via entityLink. Orphans accumulate; any unfiltered global test-case list eventually returns one whose table is gone.

### Flake 3
`AuditLogResourceIT.test_listAuditLogs_deferredJoin_{forward,backward}*` (introduced by #28850, commit `139c6ac644`) rely on the `[startTs, endTs]` window being quiescent. Under concurrent test execution that assumption is wrong: any audit row written by another test class lands in the window and breaks both strict ordering and forward-vs-backward equality. The deferred-join SQL rewrite itself is correct — the test isolation is what's broken.

### Flake 1 — backend + integration tests
- **Server (`EntityResource.bulkCreateOrUpdateAsync`):** populate `BulkOperationResult.successRequest` with each queued entity's FQN in the 202 response. FQN is already computed synchronously in `processBulkRequest` (`setFullyQualifiedName`) before the async submit — just attach it so clients can wait on the right row. Additive change, no existing client breaks.
- **Test (`BaseEntityIT.test_bulkAsync_returns202`):** read FQNs from `successRequest`, await `getEntityByName(fqn)` for each one ignoring exceptions. Drop the search-index path entirely. Works for every entity that supports bulk regardless of search support. Existing assertions (HTTP 202, `numberOfRowsProcessed`) untouched.

### Flake 2 — integration tests only
- **Test (`TestCaseResourceIT.test_listTestCasesFilteredByOwner`):** scope the list call by `entityLink` of the test's own table via `ListParams.addFilter("entityLink", ...)`. Assertion (`size >= 2`) still holds. Other tests' orphans no longer surface in the response. Backend cascade for `testCase.entityLink` is a separate, larger change worth its own PR.

### Flake 3 — integration tests only
- **Test (`AuditLogResourceIT.test_listAuditLogs_deferredJoin_*`):** generate one unique `burstToken` (UUID-derived) per test invocation, use it as the glossary name prefix, and pass `q=<burstToken>` to every `listWindow` call. The `/v1/audit/logs?q=...` filter matches on `entity_fqn` (among other fields), so only this test's audit rows come back regardless of what concurrent classes are writing. Deferred-join SQL behavior (the actual subject under test) is unchanged.

## Out of scope
- `ColumnCustomPropertiesIT.test_tableColumn_timeCpCustomProperty` ("Unknown custom field" after a successful PUT add) — separate bug class (TypeRegistry / custom-property propagation), needs local repro before a confident fix. Tracking separately.

## Test plan

- [x] Spotless clean on `openmetadata-service` and `openmetadata-integration-tests`
- [x] `openmetadata-service` test-compile clean
- [x] `BaseEntityIT.java`, `EntityResource.java`, `TestCaseResourceIT.java`, `AuditLogResourceIT.java` compile clean (other pre-existing main-branch compile errors in `MemoryAgentIT` / `TaskResourceIT` / `TagRecognizerFeedbackIT` are unrelated)
- [ ] CI green on `integration-tests-mysql-elasticsearch` and `integration-tests-postgres-elasticsearch-redis`
- [ ] No regressions in adjacent ITs (Topic, Container, Dashboard, SearchIndex, Pipeline, StoredProcedure)

Fixes three CI flakes caused by concurrency and missing test isolation in the integration-test suite. The server change is additive: `bulkCreateOrUpdateAsync` now populates `successRequest` with each queued entity's FQN in the 202 body so clients can poll for persistence without search-index dependency.

- **Flake 1 (orphan entity after async bulk):** Server populates `BulkOperationResult.successRequest` FQNs synchronously before the async submit; `BaseEntityIT.awaitAsyncBulkEntitiesPersisted` polls each FQN via `getEntityByName` with Awaitility (60 s ceiling), ensuring the async worker has committed the entity before `TestNamespace @AfterEach` fires its recursive hard-delete cascade. Covers all entity types with `supportsBulkAPI=true`, not just those with a search index.
- **Flake 2 (orphan TestCase in global list):** Scopes the list call in `test_listTestCasesFilteredByOwner` by `entityLink` of the test-owned table; orphan test cases attached to hard-deleted tables via the free-form `entityLink` field no longer leak into the response.
- **Flake 3 (deferred-join audit-log ordering):** Generates a UUID-derived `burstToken` per test invocation, uses it as the glossary name prefix, and passes `q=burstToken` to every `listWindow` call so only that test's audit events are returned, regardless of what concurrent classes write into the same time window.
</details>

Safe to merge — the server change is purely additive (a new field on the 202 response that no existing caller reads), and all three test-isolation fixes are scoped entirely to the integration-test module.

The server change adds one new field to the async 202 response and does not alter any existing field or control flow. The FQNs written into `successRequest` are computed synchronously in `processBulkRequest` before the async submit, so the values are always available. The `awaitAsyncBulkEntitiesPersisted` polling logic correctly gates on the worker committing the entity before cleanup proceeds. The `entityLink` scoping fix and the `burstToken` audit-log isolation are minimal, targeted changes with no effect outside their respective tests.

No files require special attention.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/resources/EntityResource.java | Populates `successRequest` in the 202 async response with each entity's FQN so test clients can poll for persistence; additive change that does not break existing callers that ignore the field. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BaseEntityIT.java | Adds `awaitAsyncBulkEntitiesPersisted` that polls each accepted FQN via `getEntityByName` before returning from `test_bulkAsync_returns202`; drains the async worker for every bulk-capable entity type regardless of search-index support. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TestCaseResourceIT.java | Scopes the test-case list call by `entityLink` to prevent orphan test cases from dead tables (whose `entityLink` reference isn't cascade-deleted) from polluting unfiltered global list results in concurrent test runs. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/AuditLogResourceIT.java | Generates a per-invocation UUID-derived `burstToken` as the glossary name prefix and passes `q=burstToken` to all `listWindow` calls, isolating each test's audit event set from other concurrent writers in the same time window. |

</details>

</details>

<sub>Reviews (5): Last reviewed commit: ["fix(it): make audit-log burstToken a sta..."](https://github.com/open-metadata/openmetadata/commit/ff00a00301e21bc915e9279ac5cda3426d48fede) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39024780)</sub>


yan-3005:
Maven Collate Passed: https://github.com/open-metadata/openmetadata-collate/actions/runs/28016851959
Shows failed due to Timeout

sonika-shah (approved):

pmbrull (approved):

PR #29639, opened 2026-06-30
fix(ingestion): mssql BigInteger identity reflection on SQLAlchemy 2.0

### LINKED ISSUE #29190: MS SQL Server ingestion fails due to SQLAlchemy 2.0 long_type compat error in 1.13.0
### Connector

MS SQL Server

### Feature area

Metadata ingestion

### Describe the bug

During a Metadata Ingestion pipeline for an MS SQL Server database, the extraction process fails to get columns for multiple tables, skipping them entirely. The logs show the following warning for every affected table:

`WARNING - Unexpected exception getting columns for table [table_name] (schema: 'schema_name', db: 'db_name'): module 'sqlalchemy.util.compat' has no attribute 'long_type'`

This issue occurs regardless of the connection scheme used. I have tested it with `mssql+pytds`, `mssql+pymssql`, and `mssql+pyodbc`, and the exact same exception is raised in all of them, indicating a core issue with how OMD parses specific column types in version 1.13.0.

### To Reproduce

1. Deploy OpenMetadata 1.13.0 with the default Airflow 3.2.1 ingestion image.
2. Create a Database Service pointing to an MS SQL Server.
3. Configure a Metadata Ingestion pipeline.
4. Run the pipeline on a database containing custom data types or specific integer sizes that trigger the fallback mapping.
5. Check the Airflow task logs. The tables will be listed with 0 columns processed and the warning mentioned above.

### Expected behavior

The fallback mechanism should handle unknown or specific numeric types using a SA 2.0 compliant approach (e.g., mapping to a standard `int` or `BIGINT`) so the column extraction doesn't crash. Ultimately, it is expected that all table metadata should be successfully ingested without skipping any entities or columns.

```yaml
type: DatabaseMetadata
source:
  type: mssql
  serviceName: mssql_service
  serviceConnection:
    config:
      type: Mssql
      scheme: mssql+pytds
      username: xxxxx
      password: xxxxx
      hostPort: xxxxx
      database: xxxxx
      driver: "ODBC Driver 18 for SQL Server"
      encrypt: false
      trustServerCertificate: false
  sourceConfig:
    config:
      type: DatabaseMetadata
      ingestAllDatabases: false
      databaseFilterPattern:
        excludes:
          - "^msdb$"
          - "^model$"
          - "^tempdb$"
      schemaFilterPattern:
        excludes:
          - "^db_.*"
          - "^guest$"
          - "^information_schema$"
          - "^sys$"
```

### Logs

```shell
[2026-06-17T18:18:15.559183Z] WARNING - Unexpected exception getting columns for table [xxxxxx] (schema: 'xxxxxx', db: 'xxxxxx'): module 'sqlalchemy.util.compat' has no attribute 'long_type'
[2026-06-17T18:18:16.860691Z] WARNING - Unexpected exception getting columns for table [xxxxxx] (schema: 'xxxxxx', db: 'xxxxxx'): module 'sqlalchemy.util.compat' has no attribute 'long_type'
```

### OS

Kubernetes (OpenShift OCP4) / Linux

### Python version

3.10.20

### OpenMetadata version

1.13.0

openmetadata-ingestion==1.13.0 (Airflow 3.2.1)

### Additional context

**Root Cause Analysis:**

I've tracked this bug down to the recent migration to SQLAlchemy 2.0. In SA 2.x (which Airflow 3.2.1 uses), the `sqlalchemy.util.compat` module was cleaned up and `long_type` was entirely removed (as Python 3 only uses `int`).

In the OpenMetadata codebase, specifically in [`ingestion/src/metadata/ingestion/source/database/mssql/utils.py`](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/database/mssql/utils.py#L290) (around line 290), the code explicitly imports and uses this deprecated module when mapping `BigInteger` identity columns:

```python
from sqlalchemy.util import compat  # <--- THIS MODULE HAS CHANGED IN SA 2.0

...

        if is_identity is not None:
            # identity_start and identity_increment are Decimal or None
            if identity_start is None or identity_increment is None:
                cdict["identity"] = {}
            else:
                if isinstance(coltype, sqltypes.BigInteger):
                    start = compat.long_type(identity_start) # <--- CRASHES HERE (AttributeError)
                    increment = compat.long_type(identity_increment) # <--- CRASHES HERE (AttributeError)
```

When the MS SQL dialect encounters an Identity column of type `BigInteger`, it falls into this logic and attempts to cast the seed and increment values using `compat.long_type`, instantly crashing the column extraction for that table.

### Pre-submission checklist

Khairajani:
## Problem

fix: https://github.com/open-metadata/OpenMetadata/issues/29190

MSSQL metadata ingestion fails to reflect columns for any table that has a `BIGINT IDENTITY` column. Every such table is logged with:

```
WARNING - Unexpected exception getting columns for table [<table>] (schema: '<schema>', db: '<db>'):
module 'sqlalchemy.util.compat' has no attribute 'long_type'
```

and is ingested with **0 columns**. Test-connection passes (it only lists table names), so the failure only surfaces during the full crawl.

## Root cause

In `mssql/utils.py`, the column-reflection override cast the identity seed/increment of `BigInteger` columns with `sqlalchemy.util.compat.long_type`. That symbol was removed in **SQLAlchemy 2.0** (Python 3 has no `long`), so the call raises `AttributeError`. The per-table handler in `sql_column_handler.py` catches the exception and sets `columns = []`, silently dropping the table's columns. `INT IDENTITY` columns are unaffected because that branch already used `int()`.

## Fix

- Remove the `from sqlalchemy.util import compat` import.
- Extract the identity conversion into a small, testable `get_identity_values()` helper and use `int()` for both `Integer` and `BigInteger` (on Python 3 `long_type` *was* `int`). Non-integer numeric identity types keep their original value; a missing seed/increment still yields `{}`.

This is forward-compatible with SQLAlchemy 2.0 and preserves the previous behavior on 1.4.

## Testing

Added `MssqlIdentityColumnTest` in `tests/unit/topology/database/test_mssql.py` covering `BigInteger`, `Integer`, non-integer numeric, and missing-seed cases. Verified end-to-end against a live SQL Server: a `BIGINT IDENTITY` table that previously ingested with 0 columns now ingests all columns, with no warning.

## Reproduction

```sql
CREATE TABLE dbo.big_identity (id BIGINT IDENTITY(1,1) PRIMARY KEY, name NVARCHAR(50));
```
Run a metadata ingestion on SQLAlchemy 2.0 — before this change the table ingests with no columns and emits the `long_type` warning.

Fixes a SQLAlchemy 2.0 breakage where `BIGINT IDENTITY` tables were silently ingested with zero columns due to `sqlalchemy.util.compat.long_type` being removed in SA 2.0.

- Removes the broken `compat` import and replaces the inline `BigInteger`/`Integer` cast logic with a new `get_identity_values()` helper that uses plain `int()` for all `Integer` subtypes (correct on Python 3 where `long` no longer exists separately).
- Adds four focused unit tests covering `BigInteger`, `Integer`, non-integer numeric, and missing-seed cases, directly exercising the new helper.
</details>

Safe to merge — the change is a targeted removal of a broken SA 2.0 import with a well-tested replacement.

The fix is minimal and correct: `compat.long_type` is replaced with `int()`, which is exactly what `long_type` resolved to on Python 3. The new helper is tested with all relevant type cases, and the rest of `get_columns` is structurally unchanged.

No files require special attention.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/source/database/mssql/utils.py | Removes broken `compat.long_type` usage, extracts `get_identity_values()` helper, and cleans up the now-unnecessary `# noqa: C901` comment (complexity was reduced below the 15-branch threshold by the extraction). |
| ingestion/tests/unit/topology/database/test_mssql.py | Adds `MssqlIdentityColumnTest` with four regression cases covering the key identity column types and the missing-seed edge case. |

</details>

<a href="#gh-light-mode-only">

```mermaid
flowchart TD
    A[get_columns iterates row] --> B{is_identity not None?}
    B -->|No| G[append cdict as-is]
    B -->|Yes| C[call get_identity_values]
    C --> D{start or increment is None?}
    D -->|Yes| E[return empty dict]
    D -->|No| F{isinstance Integer?}
    F -->|Yes| H[return int start and increment]
    F -->|No| I[return raw Decimal values]
    E --> J[set cdict identity]
    H --> J
    I --> J
    J --> G
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
flowchart TD
    A[get_columns iterates row] --> B{is_identity not None?}
    B -->|No| G[append cdict as-is]
    B -->|Yes| C[call get_identity_values]
    C --> D{start or increment is None?}
    D -->|Yes| E[return empty dict]
    D -->|No| F{isinstance Integer?}
    F -->|Yes| H[return int start and increment]
    F -->|No| I[return raw Decimal values]
    E --> J[set cdict identity]
    H --> J
    I --> J
    J --> G
```

</a>
</details>

<sub>Reviews (4): Last reviewed commit: ["Merge branch &#39;main&#39; into fix-mssql-bigin..."](https://github.com/open-metadata/openmetadata/commit/fbcd01199ee46fce74ffbadf79343c6dcb7f1d13) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=40929184)</sub>


ingestion/src/metadata/ingestion/source/database/mssql/utils.py
```diff
@@ -92,9 +91,24 @@ def wrap(dialect, connection, tablename, schema=None, **kw):
...
+    """Build the reflected identity dict for an MSSQL column.
+
+    seed_value / increment_value come back from MSSQL as Decimal (or None).
+    Integer and BigInteger identities are normalised to ``int``; other numeric
+    identity types keep their original value. BigInteger previously used
+    ``sqlalchemy.util.compat.long_type``, which was removed in SQLAlchemy 2.0.
+    """
+    if identity_start is None or identity_increment is None:
+        return {}
+    if isinstance(coltype, (sqltypes.BigInteger, sqltypes.Integer)):
```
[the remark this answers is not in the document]
Khairajani: Good catch — applied in 93bb46db11. Since `BigInteger` (and `SmallInteger`) subclass `Integer`, simplified to `isinstance(coltype, sqltypes.Integer)`.

ingestion/src/metadata/ingestion/source/database/mssql/utils.py
```diff
@@ -92,9 +91,24 @@ def wrap(dialect, connection, tablename, schema=None, **kw):
     return update_wrapper(wrap, fn)
 
 
+def get_identity_values(coltype, identity_start, identity_increment):
```
TeddyCr, line 94: why do we normalize to int? Int and BigInt are different (one is 4 bytes and the other 8 bytes I believe)

TeddyCr: Was looking into this deeper, this mapping is on the python type right. not a DB override?Was looking into this deeper, this mapping is on the python type right. not a DB override?

IceS2: I'm also trying to understand this. Feels this is internal only, not type override (based on the tests) 🤔

IceS2 (approved):

ingestion/src/metadata/ingestion/source/database/mssql/utils.py
IceS2: From some research indeed what we are doing is something similar to what was being done before, just getting the Seed and Increment numbers. Seems good

Khairajani: Yeah, it's not touching the column type. The int() is only on the identity seed/increment values (MSSQL hands those back as Decimal). The INT vs BIGINT type is untouched. 

It's also what SA 2.0's own mssql dialect does now - it swapped the old compat.long_type for int() on both Integer and BigInteger. Collapsed it to just Integer here since BigInteger subclasses it.

PR #29833, opened 2026-07-08
Make search embeddings, reindex auto-tune, and audit export resilient on any pod size

mohityadav766:
Three search/export-side reliability fixes that all surfaced as collate server OOMKills / restarts — under an unreachable embedding provider, a memory-constrained pod, and concurrent audit-log exports.

## 1. Embedding provider circuit breaker

When an embedding provider is persistently unavailable — e.g. AWS Bedrock `AccessDeniedException` when the deployment's role has no Bedrock permission, throttling, or an outage — the vector pipeline called the provider for **every** entity lifecycle event with no back-off (observed ~12 failures/sec, each a full stack trace, continuously), burning threads/CPU and flooding logs.

- Provider-agnostic circuit breaker in `EmbeddingClient`: after 5 consecutive failures (or immediately on a permanent auth/config failure) the circuit opens and `embed()` fast-fails with `EmbeddingUnavailableException` **without a provider call** — one WARN + a `search.embedding.circuit_opened` metric instead of per-entity stack traces. Half-opens after a cooldown (60s transient / 300s permanent) and closes on recovery.
- `BedrockEmbeddingClient` classifies 401/403/404 as permanent and drops its per-call error stack traces to debug.
- `OpenSearchVectorService` skips embedding work entirely while the circuit is open, so entities are still indexed without vectors and enrichment resumes automatically.

## 2. Container-aware reindex auto-tune memory bound

`SearchClusterMetrics` auto-tune sized reindex memory against the JVM heap (`Runtime.getRuntime().maxMemory()` = `-Xmx`) and capped concurrency and payload size **independently**, but never bounded their product. Bulk-request payloads are held largely in **off-heap direct buffers**, so `concurrentRequests × maxPayloadSize` (up to ~cores×10 × ~90 MB ≈ 3.6 GB) could exceed the container cgroup limit while the heap looked fine → **OOMKilled (137)**. Auto-tune never read the pod's own memory ceiling, so it did not "just work" on memory-constrained pods.

- Cap concurrent bulk requests so `concurrentRequests × maxPayloadSize` fits **half the off-heap headroom** (`containerLimit − maxHeap`), reading the container limit from `com.sun.management.OperatingSystemMXBean.getTotalMemorySize()` (container-aware with `UseContainerSupport`). Applied to both the metrics-based and conservative-default paths.
- Added an off-heap in-flight bulk-memory estimate to the recommendation log (it previously logged only queue heap, blind to its biggest consumer).

## 3. Audit-log export streaming + spool download

Concurrent audit-log exports (`GET /v1/audit/logs/export`) were the actual driver of the nightly restarts. The handler loaded up to 100,000 `AuditLogEntry` records into a `List`, serialized the whole thing to one JSON `String`, and sent it **inline over the `csvExportChannel` WebSocket** — on the shared async executor. Under parallel exports (e.g. 4 Playwright shards) these multi-hundred-MB payloads piled up as G1 humongous `byte[]`: first exhausting the heap (Java OOM + dump), and — once concurrency alone was bounded — the off-heap netty send buffer (OOMKilled 137). Reproduced locally end-to-end: ~45 concurrent exports produced a 3.8 GB heap dump whose `byte[]` are exactly the `csvExportChannel` audit payloads. This path was never covered by the entity/search CSV-export spool fix.

- `AuditLogRepository.streamExportAsJson` writes each 1000-row batch straight to a `CsvExportSpool` file (one batch in heap at a time), never materializing the full list or string.
- The `COMPLETED` WebSocket event now carries only a **download reference** (`data: null`) instead of the inline payload; a new `GET /v1/audit/logs/export/{jobId}/result` streams the spooled file via `StreamingOutput`.
- A dedicated 2-thread `AUDIT_EXPORT_EXECUTOR` also caps how many exports run at once (DB load / spool writers), mirroring the entity/search CSV export's already-bounded worker pool.

> Frontend follow-up (separate UI PR): `useCSVExportWebSocket` should download the audit export from the result endpoint on the `COMPLETED` event, the same way the entity/search CSV export already does.

## Tests

- `EmbeddingClientTest`: circuit opens after N failures / opens immediately on a permanent failure / recovers when the provider returns.
- `SearchClusterMetricsTest`: `boundConcurrentRequestsToMemory` shrinks concurrency to the off-heap budget, leaves it untouched when memory is ample, and degrades to serial when there's no headroom.
- Audit export: verified by the concurrent-export reproduction — the 150-concurrent hammer that previously OOM'd (3.8 GB heap dump, then OOMKilled) now leaves the server running (peak G1 humongous 322, GC'd to ~3) and the 180 MB result downloads correctly.

## Verifying in AUT

Both surfaced in the AUT nightly (Bedrock unauthorized). With the circuit breaker deployed, the `BedrockEmbeddingClient` log flood dropped from thousands/5min to ~0 and `search.embedding.circuit_opened` engaged; the auto-tune bound addresses the reindex-time off-heap OOM on memory-constrained pods; and the audit-export streaming fix targets the **confirmed root cause of the nightly restarts** — the captured heap dump's `byte[]` are `csvExportChannel` audit payloads.

----
## Summary by Gitar

- **Embedding circuit breaker:**
  - Added `EmbeddingClient` circuit breaker to handle transient/permanent provider failures.
  - Implemented `isAvailable()` check in `OpenSearchVectorService` and `VectorDocBuilder` to skip embedding work during outages.
  - Added `EmbeddingUnavailableException` for fast-failure and ensured `BedrockEmbeddingClient` identifies non-retryable status codes.
- **Bulk reindexing safety:**
  - Added `boundConcurrentRequestsToMemory` and `boundQueueSizeToHeap` to `SearchClusterMetrics` to prevent OOMs by sizing buffers against container limits and JVM heap.
- **Audit log reliability:**
  - Replaced in-memory `List<AuditLogEntry>` exports with `streamExportAsJson` using `CsvExportSpool` to prevent heap exhaustion.
  - Added `MAX_CONCURRENT_AUDIT_EXPORTS = 2` and a dedicated `AUDIT_EXPORT_EXECUTOR` to bound resource contention.
  - Added `downloadAuditExportResult` endpoint to retrieve spooled files via `StreamingOutput` after receiving completion notifications.

This PR makes search embeddings, reindex sizing, and audit export more resilient. The main changes are:

- Added an embedding provider circuit breaker and outage-aware vector indexing paths.
- Bounded reindex bulk concurrency and queue size using container and heap memory estimates.
- Streamed audit-log exports to spool files with a result download endpoint.
- Added tests for embedding circuit behavior, reindex memory bounds, and vector outage handling.
</details>

This is close, but the search outage path should be fixed before merging.

- The indexing and chunk-update paths now avoid destructive chunk replacement during embedding outages.
- The half-open circuit path now limits recovery probing.
- Semantic search can still fail the request when query embedding fast-fails through the new circuit path.

openmetadata-service/src/main/java/org/openmetadata/service/search/vector/client/EmbeddingClient.java and the semantic search caller in OpenSearchVectorService
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/search/vector/client/EmbeddingClient.java | Adds circuit-breaker state handling for embedding calls, but query embeddings can now fast-fail a search request during an outage. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/vector/OpenSearchVectorService.java | Skips unavailable embedding work for indexing paths and catches chunk-build outages, while the search path still needs outage handling. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/vector/VectorDocBuilder.java | Uses an explicit unavailable exception instead of returning an empty chunk list during provider outages. |
| openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/searchIndex/OpenSearchBulkSink.java | Avoids generating new embeddings during reindex when the provider circuit is unavailable. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/SearchClusterMetrics.java | Adds memory-aware bounds for reindex concurrency and queue sizing. |
| openmetadata-service/src/main/java/org/openmetadata/service/audit/AuditLogRepository.java | Streams audit export JSON directly to an output stream. |
| openmetadata-service/src/main/java/org/openmetadata/service/resources/audit/AuditLogResource.java | Runs audit exports on a bounded executor and exposes a spooled result download endpoint. |

</details>

<details><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `openmetadata-service/src/main/java/org/openmetadata/service/search/vector/client/EmbeddingClient.java`, line 86-88 ([link](https://github.com/open-metadata/openmetadata/blob/574785afd9bd97ffeb3eb80c8dc130a38fbe18ee/openmetadata-service/src/main/java/org/openmetadata/service/search/vector/client/EmbeddingClient.java#L86-L88)) 

   <a href="#">P1</a> **`embedQuery` bypasses the circuit breaker**

   `embedQuery` calls `embedWithLimit(() -> doEmbedQuery(text))` directly, skipping `guardCircuit()`. The circuit can be open while `embedQuery` still issues provider calls — every semantic search query routed through `OpenSearchVectorService.search()` (which calls `embeddingClient.embedQuery(query)`) will keep hitting a failed provider and propagate as an uncaught `RuntimeException`, rather than receiving the fast-fail `EmbeddingUnavailableException` the entity-write path now correctly gets. `embedQuery` needs to call `guardCircuit()` and route through `recordSuccess`/`recordFailure` the same way the new `embed` method does.

2. `openmetadata-service/src/main/java/org/openmetadata/service/search/vector/client/EmbeddingClient.java`, line 78 ([link](https://github.com/open-metadata/openmetadata/blob/c1203c88b451e41770df5d89a5714cb70760c2a1/openmetadata-service/src/main/java/org/openmetadata/service/search/vector/client/EmbeddingClient.java#L78)) 

   <a href="#">P1</a> **Search fast-fails**

   `embedQuery()` now goes through the circuit breaker, but the semantic search caller does not handle the new fast-fail exception. When the provider circuit is open, `OpenSearchVectorService.search()` calls `embeddingClient.embedQuery(query)` before its OpenSearch error handling and can let `EmbeddingUnavailableException` escape as a server error. Search should degrade cleanly while embeddings are unavailable instead of failing the request.

</details>

<sub>Reviews (21): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/embedding-p..."](https://github.com/open-metadata/openmetadata/commit/c1203c88b451e41770df5d89a5714cb70760c2a1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42659342)</sub>

</details>


harshach (dismissed):

harshach:
@mohityadav766 lets address bot reviews

mohityadav766:
Addressed the open review threads in `51ca7466`.

**Fixed**
- **Half-open probe fan-out** (EmbeddingClient L145/L147, `isAvailable` L113, permit-race L84): `HALF_OPEN` is now a deterministic single-probe state — concurrent callers are rejected by *state* rather than a time deadline, so a probe that outlives the cooldown can't leak extra probes. `isAvailable()` returns `false` while a probe is in flight, and a permit-acquisition failure is now recorded so a promoted probe always resolves (closed/reopened) and never wedges the circuit half-open. New latch-based `testHalfOpenAllowsOnlyOneConcurrentProbe`.
- **Empty-list deletes existing vectors** (VectorDocBuilder L156; OpenSearchVectorService L225/L752): `fromEntity` now throws `EmbeddingUnavailableException` on outage instead of returning `List.of()`, which callers couldn't distinguish from "no chunks". The racy `chunksEmptyFromOutage` re-check is removed and both write paths catch the exception, so `replaceChunks` never runs during an outage.
- **Reindex sink** (OpenSearchBulkSink): the embedding call is skipped when the circuit is open, so a transient outage indexes entities without embeddings (self-heals on the next reindex) rather than marking every entity failed.
- **Int overflow** (SearchClusterMetrics L445): the memory cap is computed in `long` and cast after `Math.min`, so a tiny payload can't overflow to negative concurrency. Regression test added.
- **Audit export** (AuditLogResource L262/L323): jobId is validated as a server-generated UUID before touching the spool path (path-traversal guard), and a partial spool is deleted on export failure.

**Already resolved / stale** (from earlier commits): `openCause` null-guard (L182-183), the "syntactically broken"/"breaker bypassed" comments, the `ponytail:` marker, and the conservative-queue ordering (queue is sized from the *bounded* concurrency).

**Deliberately not changed**
- **Permit-race fail-fast** (L67): inherent to any breaker with in-flight requests — bounded by the concurrency limiter and self-correcting via `recordFailure`. A naive re-guard after the permit would reject the legitimate probe.
- **Off-heap throttle to ~1 on high-Xmx** (L454) and **queue floor of 1000** (L472): intended safety tradeoffs (both already `LOG.info`-warned). Throttling to fit off-heap headroom is exactly the OOMKill-137 fix this PR is about; loosening it reintroduces the bug.

mohityadav766:
> @mohityadav766 lets address bot reviews

done @harshach

chirag-madlani (approved):

PR #29871, opened 2026-07-09
feat(contextCenter): sortBy/sortOrder on page hierarchy search

sonika-shah:
Issue: open-metadata/openmetadata-collate#4781

## Companion / related PRs

- **open-metadata/openmetadata-collate#4932** — applies the same sort in the Collate ES/OpenSearch client extensions (depends on this PR).
- **#29875** — Context Center archive page current-user filter for drive files (independent OSS change, split out of this PR).

## What

`GET /v1/contextCenter/pages/search/hierarchy` now accepts:

- `sortBy` — one of `name`, `createdAt`, `updatedAt`
- `sortOrder` — `asc` | `desc`

**Default is `updatedAt` `desc`**, so a newly created article or quick link surfaces at the top of the left-tree hierarchy. `fullyQualifiedName` is always appended as a secondary sort key, keeping `from`/`size` pagination stable when the primary field (e.g. `updatedAt`) is non-unique. Invalid `sortBy`/`sortOrder` values return `400`.

The sort is threaded through `KnowledgePageRepository` → `SearchClient` → the Elastic/OpenSearch hierarchy queries (both the parent-scoped and active-page paths).

## Tests

`KnowledgePageHierarchyIT`: ascending/descending name ordering via `/search/hierarchy`; `400` on invalid `sortBy` and `sortOrder`.

## Notes

Not addressed here (flagged as "to be discussed" in the design review): how a **child** page update should reflect in the main hierarchy ordering.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR wires `sortBy`/`sortOrder` query parameters into the `/v1/contextCenter/pages/search/hierarchy` endpoint, defaulting to `updatedAt desc` so newly created articles surface at the top of the left-tree. It also threads a stable `fullyQualifiedName` tiebreaker through both ES and OS clients to keep `from`/`size` pagination deterministic.

- **Sort parameter plumbing** — `KnowledgePageResource` validates the sort params (`IllegalArgumentException` → 400 for unknown values), builds a `SearchSortFilter`, and forwards it through `KnowledgePageRepository` → `SearchClient` → `ElasticSearchClient`/`OpenSearchClient`, where `buildPageHierarchySortOptions` produces the final sort list.
- **`fullyQualifiedName` tiebreaker preserved** — both hierarchy query paths always append `fullyQualifiedName asc` as a secondary sort after the user-requested primary, keeping pagination stable when the primary field (e.g. `updatedAt`) is non-unique.
[...]

PR #29875, opened 2026-07-09
feat(contextCenter): filter drive files by updatedBy (archiver)

sonika-shah:
Issue: open-metadata/openmetadata-collate#4781

## What

`GET /v1/contextCenter/drive/files` now accepts an `updatedBy` query param (a username), backed by a new `updatedBy` condition in `ListFilter`. Combined with `include=deleted`, it scopes the archive listing to files **archived by** that user.

**Why `updatedBy`, not `createdBy`:** archiving a document is a soft-delete, so the person who archived it is recorded in `updatedBy`. An archived file can't be edited, so `updatedBy` stays pinned to whoever archived it. `createdBy` would wrongly match the original author even when someone else archived the file.

`updatedBy` is a plain username string (`json ->> 'updatedBy'`, a generated column on the entity table), so the filter matches it directly — no relationship join or hashing — and applies on both the cursor list path (`listInternal`) and the `orderBy` path (`listByUpdatedAt`).

## Tests

`ContextFileIT`: after archiving a file, the `updatedBy` filter (using the archiver captured post-delete) includes it, a different user excludes it, and the unfiltered `include=deleted` list still returns it.

## Context

Split out from #29871 (Context Center hierarchy `sortBy`/`sortOrder`) — independent OSS work, no downstream Collate coupling.

This PR adds an `updatedBy` query parameter to `GET /v1/contextCenter/drive/files`, allowing the archive page to scope results to files archived by a specific user. The filter is implemented as a direct column match on the `updatedBy` field (avoiding the relationship-table lookup used by `createdBy`), and applies on both the cursor list path and the `orderBy` path.

- **`ListFilter.getUpdatedByCondition()`** — new method emitting a parameterised `updatedBy IN (…)` predicate, registered in the global `getCondition()` list; it is a no-op for all entity types that don't explicitly populate the `updatedBy` query param.
- **`ContextFileResource.list()`** — new `updatedBy` `@QueryParam` forwarded to `ListFilter`; guard (`!= null && !isBlank`) is consistent with the existing `folderId` guard.
- **`ContextFileIT.testListArchivedFilesFilteredByUpdatedBy`** — integration test covering include/exclude and an unfiltered control check to confirm the filter (not the soft-delete flag) drives the scoping.
</details>

Safe to merge; the change is a straightforward additive filter on a direct entity column with parameterised binding and targeted integration test coverage.

The filter uses JDBI named bind parameters throughout, so there is no SQL injection risk. The updatedBy column is a standard field present on all entity tables, and the condition is gated on an explicit addQueryParam call in the resource — making it a no-op for every other entity type. The only finding is a mismatch between the PR title/description (which say createdBy) and the actual parameter (updatedBy), which has no impact on runtime behaviour.

No files require special attention; all three changed files are straightforward.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/resources/drive/ContextFileResource.java | Adds updatedBy query parameter to the list endpoint and passes it to ListFilter; filter is applied on both the cursor (listInternal) and orderBy (listByUpdatedAt) code paths. |
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/ListFilter.java | New getUpdatedByCondition() generates a parameterised updatedBy IN predicate directly against the entity table column, registered in the global getCondition() list; safe because it only fires when the queryParam key is present. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/drive/ContextFileIT.java | Adds testListArchivedFilesFilteredByUpdatedBy covering include/exclude behaviour and a control check that unfiltered archive still returns the file. |

</details>

<sub>Reviews (3): Last reviewed commit: ["feat(contextCenter): filter drive files ..."](https://github.com/open-metadata/openmetadata/commit/b902b856f825f23288f107e90f65ca1e9b831171) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42974001)</sub>

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/drive/ContextFileIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/drive/ContextFileIT.java
@@ -2,6 +2,7 @@ package org.openmetadata.it.drive;
 
 import static org.awaitility.Awaitility.await;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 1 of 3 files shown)


sonika-shah:
Superseded by #29902, which carries these exact backend changes (`ListFilter`, `ContextFileResource`, `ContextFileIT` are byte-identical) plus the archive-page UI + pagination. Closing in favor of that PR. Tracked by open-metadata/openmetadata-collate#4781.

PR #30039, opened 2026-07-14
Fixes #30037: refactor(ingestion): migrate Looker test-connection to the @check framework

### LINKED ISSUE #30037: Migrate Looker test-connection to the @check framework
Looker's test connection still runs through the imperative `test_connection_steps` helper: the four steps are a dict of lambdas, failures surface as raw `SDKError` tracebacks with no remediation, and the `CheckAccess` step is not marked as a `ConnectionGate`, so an unreachable host makes every subsequent step re-dial it.

Migrate it to the declarative `@check` / `ChecksProvider` / `ErrorPack` framework, as already done for PowerBI and Tableau:

- `LookerChecks` provider with one `@check` per step (`CheckAccess`, `ValidateVersion`, `ListDashboards`, `ListLookMLModels`), reusing the shared dashboard helpers.
- `LOOKER_ERRORS` pack mapping the failures the Looker SDK actually raises (rejected API3 credentials, 401/403/404, unsupported API version, and the network conditions the SDK transport flattens into an `SDKError` message) to actionable diagnoses.
- Mark `CheckAccess` as the `ConnectionGate` in `testConnections/dashboard/looker.json` so a failed gate short-circuits the remaining steps.
- Drop the `test_connection` override so `BaseConnection.test_connection` drives the checks through the runner.

IceS2:
Migrates Looker's test connection off the imperative `test_connection_steps` helper onto the declarative `@check` / `ChecksProvider` / `ErrorPack` framework, following the merged PowerBI and Tableau connectors.

Closes #30037

## What changed

- **`LookerChecks` provider** — one `@check` per step (`CheckAccess`, `ValidateVersion`, `ListDashboards`, `ListLookMLModels`), reusing the shared dashboard helpers (`verify_access`, `call_endpoint`, `fetch_list`). The `test_connection` override is gone; `BaseConnection.test_connection` now drives the checks through the runner.
- **`BaseConnection` owns the client.** The provider builds nothing itself: it takes a `connect` thunk over `BaseConnection.client` (the same wiring as Athena, BigQuery, Glue, and S3), so every step shares the one client the connection builds, caches, and closes — one owner, one client, one lifetime. Passing a thunk rather than the client keeps the build inside the first check, so nothing dials the instance before the runner's gate.
- **`CheckAccess` is the `ConnectionGate`** in `testConnections/dashboard/looker.json`, so a failed gate short-circuits the remaining steps instead of each one re-dialling an unreachable host.
- **`LOOKER_ERRORS` pack** — maps the failures the SDK actually raises to actionable diagnoses. Two Looker-specific signals drive it:
  - `SDKError` carries **no status code**; the response body is deserialized into `message` / `documentation_url`, and it is the *documentation URL* that encodes the status and the endpoint (`.../r/err/4.0/404/post/api/4.0/login`). The pack parses the status from there. Because Looker answers rejected API3 credentials with a **404 on `/login`** rather than a 401, the endpoint — not the status — is what separates bad credentials from a wrong host, so that rule is ordered ahead of the generic 404.
  - The SDK's transport **catches every `IOError`** (DNS, refused, TLS, timeout) and re-raises it as an `SDKError` whose message is the stringified original, so the exception *type* is lost. Those conditions are therefore matched on text, ahead of the folded `NETWORK_ERRORS` type-based rules, which remain the fallback for anything raised outside the transport.
- **Extra `ListDashboards` / `ListLookMLModels` caveats** — an empty result now passes with a non-blocking Warning explaining what a missing permission would cost (dashboards not visible; no lineage without LookML models) instead of a bare `assert`.
- Adds `ValidateVersion`, `ListDashboards`, and `ListLookMLModels` to the shared `DashboardStep` enum (first occurrence of each; existing members untouched).

Step order, names, and `mandatory` flags are preserved exactly; only the gate `category` is added to the definition.

## Note for review

The error pack was derived by reading the `looker_sdk` transport and error-handling source, not against a live instance — the status-from-documentation-URL and the flattened-`IOError` text rules are the two behaviours worth confirming during live validation.

This PR moves Looker test-connection checks onto the shared check framework. The main changes are:

- New declarative Looker checks for access, API version, dashboards, and LookML models.
- Looker-specific error classification for SDK and transport failures.
- A shared SDK client owned by `BaseConnection`.
- A three-minute per-step timeout for Looker checks.
- `CheckAccess` marked as the connection gate in the test-connection definition.
</details>

This looks safe to merge.

- The timeout behavior now preserves the Looker three-minute budget through the base runner.
- The documented Looker error guards now keep structured Looker errors out of the generic network and host fallback rules.

</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/source/dashboard/looker/connection.py | Adds the Looker checks provider, SDK settings wrapper, error pack, and three-minute per-step timeout. |
| ingestion/src/metadata/core/connections/test_connection/checks/dashboard.py | Adds the Looker dashboard step names used by the new check provider. |
| openmetadata-service/src/main/resources/json/data/testConnections/dashboard/looker.json | Marks `CheckAccess` as the connection gate for Looker test connections. |
| ingestion/tests/unit/source/dashboard/looker/test_connection.py | Adds tests for the new Looker checks, SDK configuration, error classification, and timeout behavior. |

</details>

<sub>Reviews (4): Last reviewed commit: ["fix(ingestion): don&#39;t read a structured ..."](https://github.com/open-metadata/openmetadata/commit/4659e90d52d75c43ad97eb216b7ccb360a469cee) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44187720)</sub>

</details>

### DIFF
--- a/ingestion/src/metadata/core/connections/test_connection/checks/dashboard.py
+++ b/ingestion/src/metadata/core/connections/test_connection/checks/dashboard.py
@@ -47,11 +47,14 @@ class DashboardStep(StepName):
     GetDashboards = "GetDashboards"
     ServerInfo = "ServerInfo"
     ValidateApiVersion = "ValidateApiVersion"
+    ValidateVersion = "ValidateVersion"
     ValidateSiteUrl = "ValidateSiteUrl"
     GetWorkbooks = "GetWorkbooks"
     GetViews = "GetViews"
     GetOwners = "GetOwners"
     GetDataModels = "GetDataModels"
+    ListDashboards = "ListDashboards"
+    ListLookMLModels = "ListLookMLModels"
 
 
 def _count(number: int, noun: str, cap: int | None = None) -> str:
--- a/ingestion/src/metadata/ingestion/source/dashboard/looker/connection.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/looker/connection.py
@@ -115,6 +115,10 @@ def _contains_any(*tokens: str) -> Matcher:
     return match
 
 
+# The statuses handled here are the ones the API spec documents for the endpoints
+# these checks call (/login, /user, /versions, /dashboards, /lookml_models): 400,
+# 404 and 429. It documents no 401 anywhere, and no 403 on any of them, so an
+# unmatched status keeps its raw errorLog rather than a made-up diagnosis.
 LOOKER_ERRORS = ErrorPack(
     # Ordered before the generic 404: Looker rejects credentials with a 404 on
     # /login, so the endpoint is what separates them from a wrong host.
@@ -128,19 +132,13 @@ LOOKER_ERRORS = ErrorPack(
         fix="Provide both the Client ID and the Client Secret.",
         doc=API_SDK_DOC,
     ),
-    when(_http_status(401)).diagnose(
-        "Authentication failed",
-        fix="Looker did not accept the credentials. Check the Client ID and Client Secret.",
-        doc=API_SDK_DOC,
-    ),
-    when(_http_status(403)).diagnose(
-        "Insufficient permissions",
-        fix="The credentials are valid but not authorized for this call. Grant the user access to "
-        "the dashboards and models to ingest.",
-    ),
     when(_http_status(404)).diagnose(
         "Resource not found",
         fix="Looker could not find the requested resource. Check that Host Port is the instance URL.",
+    ),
+    when(_http_status(429)).diagnose(
+        "Rate limited by Looker",
+        fix="Looker is throttling the requests. Retry once the instance is under less load.",
     ),
     when(Matchers.exception(UnsupportedApiVersionError)).diagnose(
         f"API {SDK_API_VERSION} is not supported by this instance",
--- a/ingestion/src/metadata/ingestion/source/dashboard/looker/connection.py
+++ b/ingestion/src/metadata/ingestion/source/dashboard/looker/connection.py
@@ -59,6 +59,10 @@ API_SDK_DOC = "https://cloud.google.com/looker/docs/api-sdk"
 # https://cloud.google.com/looker/docs/r/err/4.0/404/post/api/4.0/login
 _ERROR_DOC_URL = re.compile(r"/r/err/[^/]+/(?P<status>\d{3})/(?P<path>\S*)", re.IGNORECASE)
 
+# A 404 from a server that is not Looker. Word-bounded so a port or byte count
+# containing 404 does not match.
+_FOREIGN_404 = re.compile(r"\b404\b|not found", re.IGNORECASE)
+
 
 class UnsupportedApiVersionError(Exception):
     """The Looker instance does not list the API version the SDK speaks."""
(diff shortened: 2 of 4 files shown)


IceS2:
Thanks — both addressed in 17a9d08.

**1. Doc-URL parsing (bug): confirmed and fixed.** I verified this against the floor of the pin rather than assuming. On `looker-sdk 22.20.0`, `SDKError` has no custom `__str__` (it arrives in 24.x), so:

```
>>> str(SDKError("Not found", documentation_url=".../r/err/4.0/404/post/api/4.0/login"))
'Not found'                     # doc URL absent from the message
>>> err.documentation_url
'https://.../r/err/4.0/404/post/api/4.0/login'   # present on the attribute
```

Matching `str(error)` alone would therefore have silently stopped classifying **every** status rule for users on that version. `_error_doc` now searches the message *and* the attribute, which also keeps the login case working (a failed login raises the raw body as the message and leaves the attribute empty — the reason it was reading `str()` in the first place). Added a regression test built on an error whose `__str__` omits the URL; it fails without the fix.

**2. Foreign-404 fallback: tightened, though the ordering premise doesn't hold.** The fallback is matched on a word boundary now (`\b404\b|not found`), so a port or byte count containing `404` no longer reads as an HTTP 404.

One correction on the reasoning: the rule does not precede the network rules. The DNS / refused / timeout / TLS / `max retries exceeded` rules are all ordered **above** it, and the SDK's transport flattens every `IOError` into an `SDKError` message, so a genuine network failure is classified there and never reaches the fallback — the `.including(NETWORK_ERRORS)` type-based pack is the last-resort layer for errors raised *outside* the transport. Added a test pinning that: a refused connection on port `8404` still diagnoses as "Connection refused".

Both re-verified live against a real Looker instance; unit tests now 33, ruff and basedpyright clean.

IceS2:
@greptile-apps both findings were correct — fixed in 1384634.

**P1 — timeout regression.** Confirmed: the imperative handler defaulted to `THREE_MIN`, while `BaseConnection.step_timeout_seconds` defaults to `STEP_TIMEOUT_SECONDS = 60`, so the migration silently cut the budget to a third and a slow `all_dashboards()` / `all_lookml_models()` on a large instance could fail an otherwise working connection. Pinned `step_timeout_seconds = THREE_MIN`, the same way the merged Tableau connector preserves its legacy budget.

**P2 — Looker error read as a foreign host.** Also correct, and the failure mode is exactly as described: the fallback matched the bare text, so a Looker error carrying a status this pack does not diagnose (a `400` reading "parameter not found" — `400` is documented for every endpoint these checks call) was reported as "The host is not serving the Looker API", pointing the user at Host Port instead of the real error. The fallback now requires the *absence* of a Looker error document: any Looker answer carries one, whether or not its status is diagnosed, so it keeps its raw `errorLog` instead. Regression test added for the `400 parameter not found` case.

Unit tests now 35; ruff and basedpyright clean.

IceS2:
@greptile-apps "Guard documented errors" (P1) — confirmed and fixed in 4659e90.

Reproduced it: a structured Looker `400` carrying an error document, whose message reads `"Query exceeded the timeout"`, was classified as *"Connection timed out"* — the text-based transport rules fired even though Looker returned a structured response, sending the user toward a network fix.

It's the same guard `_foreign_404` already had, generalised into `_transport_text(...)`: the DNS / refused / timeout / TLS / cannot-reach rules now skip when any error in the chain carries a Looker error document. This is safe because a real transport failure is flattened by the SDK into a message with **no** `/r/err/...` document, so it still matches — verified both directions:

- Looker `400` + `"timeout"` / `"connection error"` in message → `None` (raw errorLog kept)
- real `Connection refused`, `Read timed out`, DNS failure (no document) → still diagnosed correctly

Regression test added. Unit tests now 36; ruff and basedpyright clean.

ulixius9 (approved):

PR #30040, opened 2026-07-14
Fixes #30038: refactor(ingestion): migrate dbt Cloud test-connection to the @check framework

### LINKED ISSUE #30038: Migrate dbt Cloud test-connection to the @check framework
Migrate the dbt Cloud test connection off the imperative `test_connection_steps` helper onto the declarative `@check` / `ChecksProvider` / `ErrorPack` framework, as already done for PowerBI, Tableau, MySQL, Postgres, MSSQL, Redshift, Snowflake, Unity Catalog and S3.

dbt Cloud is the first Pipeline connector on the framework, so it also seeds the Pipeline vertical.

Scope:
- New `checks/pipeline.py` scaffolding: a `PipelineStep` enum (`CheckAccess`, `GetJobs`, `GetRuns`) and the REST helpers (`verify_access`, `fetch_list`) the next pipeline connectors (Airflow, Fivetran) reuse.
- `DBTCloudChecks` provider with one `@check` per step, each returning `Evidence`.
- `DBTCLOUD_ERRORS` `ErrorPack` mapping HTTP 401/403/404/429 and the transport failures to actionable diagnoses, folding in the shared network pack.
- A new `CheckAccess` `ConnectionGate` step so a bad token or a wrong account id short-circuits the remaining steps instead of re-dialling the API for each one.
- Surface the HTTP status of a failed dbt Cloud call: its error body nests `code` under `status`, so the shared REST client neither raises nor returns and a rejected token currently reads as an opaque validation error rather than "Authentication failed".

IceS2:
Fixes #30038

Moves dbt Cloud off the imperative `test_connection_steps` helper onto the declarative `@check` / `ChecksProvider` / `ErrorPack` framework. dbt Cloud is the first Pipeline connector on the framework, so this also seeds the Pipeline vertical.

**Pipeline scaffolding (new)** — `core/connections/test_connection/checks/pipeline.py`: a `PipelineStep` enum (`CheckAccess`, `GetJobs`, `GetRuns`) plus the REST helpers `verify_access` and `fetch_list`, modelled on `checks/dashboard.py`. Airflow and Fivetran reuse these next. `verify_access` / `fetch_list` / `_count` are near-duplicates of the dashboard ones — left duplicated on purpose, per the scaffolding policy; a later pass can consolidate the REST helpers across verticals.

**Connector** — `DBTCloudChecks` with one `@check` per step, and `DBTCLOUD_ERRORS` mapping HTTP 401/403/404/429 and the transport failures (TLS, timeout, unreachable host) to actionable diagnoses, folding in the shared network pack. The provider checks the client `BaseConnection` owns rather than opening a second one; constructing it opens no connection (it only builds the REST client config and a session), so the gate stays the first call that reaches dbt Cloud. `GetJobs` on an account with no jobs passes with a non-blocking caveat rather than silently looking healthy.

**Gate** — a new `CheckAccess` step, `"category": "ConnectionGate"`, reads the first page of one job (`limit=1`). That is the smallest authenticated call proving Host, Token and Account Id together, and it exercises the same account-scoped path ingestion reads, so a token that passes it cannot fail the later steps for auth. dbt Cloud exposes no cheaper whoami endpoint the client already wraps, and probing a broader account-level endpoint would risk a false 403 for a narrowly-scoped service token. A bad token or wrong account now short-circuits instead of re-dialling the API for each remaining step.

**HTTP status surfacing** — a dbt Cloud error body nests its error `code` under `status`, so the shared REST client's error branch neither raises nor returns: the caller got a bare `None` and a rejected token surfaced as an opaque pydantic validation error. The three test-connection calls now issue the request directly and raise `DBTCloudApiError` carrying the status, so 401 (bad token) is told apart from 404 (wrong account) and reaches the error pack. Ingestion keeps using the shared client untouched.

This PR migrates the dbt Cloud test connection to the declarative check framework. The main changes are:

- New Pipeline check steps and shared REST-list helpers.
- dbt Cloud `ChecksProvider` wiring with a `CheckAccess` gate.
- Raw-response REST support so test calls keep status and retry behavior.
- dbt Cloud-specific error diagnoses for auth, account, rate limit, network, and non-JSON responses.
- Unit coverage for the new check flow and REST raw-response behavior.
</details>

This looks safe to merge.

</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/core/connections/test_connection/checks/pipeline.py | Adds Pipeline step names plus helpers for authenticated checks and bounded list evidence. |
| ingestion/src/metadata/ingestion/ometa/client.py | Adds raw GET support while preserving the existing retry and limit handling path. |
| ingestion/src/metadata/ingestion/source/pipeline/dbtcloud/client.py | Routes dbt Cloud test calls through raw REST responses and raises status-carrying API errors. |
| ingestion/src/metadata/ingestion/source/pipeline/dbtcloud/connection.py | Defines the dbt Cloud check provider, gate step, and connector-specific error pack. |
| openmetadata-service/src/main/resources/json/data/testConnections/pipeline/dbtcloud.json | Adds the `CheckAccess` connection gate to the dbt Cloud test-connection definition. |

</details>

<sub>Reviews (6): Last reviewed commit: ["refactor(ingestion): REST.get\_raw keeps ..."](https://github.com/open-metadata/openmetadata/commit/16ee5662f50126ac8babb9e304dd4754a9e3280a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44194887)</sub>

</details>


IceS2:
> Unless config validation already rejects scheme-less hosts, restore a matcher for MissingSchema/InvalidURL.

It does, so the matcher would be unreachable and I'm leaving it out.

`host` is `AnyUrl` in the generated config (`dbtCloudConnection.json` declares `"format": "uri"`), so a scheme-less Host fails pydantic validation before any check runs:

```
'cloud.getdbt.com'          REJECTED -> Input should be a valid URL, relative URL without a base
'https://cloud.getdbt.com'  ACCEPTED
```

Manual UI testing confirms it: a scheme-less Host never reaches the connector. The automation payload fails validation and the request dies in `parse_automation_workflow_gracefully`'s error path, so `requests.get` is never called and `MissingSchema`/`InvalidURL` cannot be raised. Every rule in `DBTCLOUD_ERRORS` is backed by a response observed against the live dbt Cloud API (401 `Invalid token.`, 403 `Token is not scoped to account.`) or by the [rate-limit docs](https://docs.getdbt.com/docs/dbt-apis/rate-limits) (429), and I removed this one for the same reason I removed the 404 rule: it can't fire.

Good catch on the PR description, though — it still claimed non-absolute Host was mapped. Fixed.

Separately, that validation path has a pre-existing bug unrelated to this PR: `get_connection_class` (`ingestion/src/metadata/ingestion/api/parser.py:294`) derives the module as `source_type[0].lower() + source_type[1:] + "Connection"` → `dBTCloudConnection`, falling back to all-lowercase `dbtcloudConnection`, while the real module is `dbtCloudConnection`. So any dbt Cloud payload that fails validation 500s in the error handler instead of reporting the validation error. Reproduces on `main`; will be filed separately.

IceS2:
Thanks @greptile-apps — all three are real. Fixed in 68ec86a.

**Gate skips REST retries.** Correct: the test calls bypass `TrackedREST`, so they lost its transport retries and a transient connect failure would have reported an unreachable host. They now go through a session with the same `mount_resilient_adapter` the REST client mounts (total=3, connect=2, read=1), so a blip retries instead of failing the gate. Covered by a test asserting the adapter is mounted.

**Successful HTML becomes generic failure.** Correct, and reachable in practice: a Host that is a valid URL but not the API — e.g. `https://www.getdbt.com` — answers `308` and redirects to an HTML page, which returns 200 and fails to decode. That now maps to **"Host is not the dbt Cloud API"** with the fix to set Host to the dbt Cloud access URL, instead of a raw `JSONDecodeError`.

**404 status loses diagnosis.** Fair. I had removed the 404 rule because a wrong Account Id answers `403 "Token is not scoped to account."` (confirmed by manual testing), never 404 — so the rule as written was wrong. But an unmatched 404 should still be actionable, so it is back as **"Endpoint not found"**, worded around what is verifiable: Host and Account Id are what build the path the API answered 404 for.

61 unit tests, basedpyright clean, and the two new diagnoses verified against the running stack.

IceS2:
Fixed in 874ca06. Right that there was a gap — the mounted adapter retries connect/read but not HTTP status (`status=0`), so a transient 502/503/504 failed the gate immediately while the REST client retries it.

One nuance shaped the fix: the REST client retries 504 on `retry_wait=30` → 30/60/90s sleeps, and a single such sleep would blow the 60s step budget. So instead of importing that behavior, the test session now retries gateway 5xx at the adapter layer with the fast urllib3 backoff (0/2/4s), bounded by `total=3`, derived from `build_transport_retry()` so it stays in step with the shared transport config. A test asserts `status_forcelist == {502, 503, 504}`.

Latest state: 70 unit tests, basedpyright clean, verified against the running stack.

ulixius9 (approved):

PR #30094, opened 2026-07-15
Fixes #30093: database sources own one BaseConnection to avoid a second sign-in

### LINKED ISSUE #30093: database sources own one BaseConnection to avoid a second sign-in
Database sources build their working client in `__init__` and then a second, throwaway connection for the startup test-connection step. On single-session auth (Tableau Personal Access Tokens, some OAuth flows) the second sign-in invalidates the first, so ingestion fails after the test reports success.

This is the database slice of the source-owns-BaseConnection work (dashboard shipped in #29870). Each database source now owns one `BaseConnection`, reuses it for the test step, and disposes it on close:

- `CommonDbSourceService` / `CommonNoSQLSource` bases (all engine + NoSQL sources); `set_inspector` swaps the per-database owner, disposing the previous one.
- The leaf sources that extend `DatabaseServiceSource` directly and built a raw client (datalake, deltalake, domodatabase, glue, salesforce, sas).
- Unity Catalog (metadata + lineage) and burstiq.

`connection_obj` and `test_connection_common` stay as compat surfaces for non-migrated / custom connectors.

Tracked PR: #30094

IceS2:
Fixes #30093

Database sources built their working client in `__init__` and then a second, throwaway connection for the startup test-connection step. On single-session auth (Tableau PATs, some OAuth) the second sign-in invalidated the first, so ingestion 401'd after the test reported success. This is the database slice of the source-owns-BaseConnection work (dashboard: #29870).

## What
Every database source now owns one `self._connection` `BaseConnection` and reuses it for `test_connection()`:

- `CommonDbSourceService` / `CommonNoSQLSource` bases (all engine + NoSQL sources); `set_inspector` swaps the per-database owner, disposing the previous one.
- The six leaf sources that extend `DatabaseServiceSource` directly and built a raw client — datalake, deltalake, domodatabase, glue, salesforce, sas. salesforce/sas drop their `test_connection` overrides so they inherit the base (which routes through the owned connection); this also fixes salesforce's double *login*.
- Unity Catalog (metadata + lineage) and burstiq.

`connection_obj` and `test_connection_common` stay as compat surfaces for non-migrated / custom connectors.

## How
- Unity Catalog's source engine now comes from the owner's SQL sub-owner (`_connection.sql.client`) instead of a second `get_sqlalchemy_connection` build, so the warehouse connection is created once and reused by the test-connection checks.
- The raw `DatabricksClient` (REST API) becomes an owned `DatabricksApiConnection` sub-owner on `DatabricksConnection` and `UnityCatalogConnection`.
- burstiq's module-level `get_connection` is inlined into `_get_client` and its client cache bounded; other connectors keep their `get_connection` (used by their own `_get_client`, and databricks downstream).

This PR gives database sources one lifecycle-owned connection and reuses it for startup checks. The main changes are:

- Shared ownership and cleanup in the database source bases.
- Owned connections for the migrated leaf sources.
- SQL and REST sub-owners for Databricks and Unity Catalog.
- Bounded BurstIQ client caching.
- Tests for connection reuse, cleanup, and inherited behavior.
</details>

This looks safe to merge.

- The updated leaf sources now close their owners during normal shutdown.
- Failed startup checks close the retained owner before propagating the error.
- Datalake cleanup is registered with the owner instead of relying on the removed source override.
- No blocking issue remains in the reviewed fixes.

</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/source/database/database_service.py | Adds shared testing and cleanup for lifecycle-owned connections. |
| ingestion/src/metadata/ingestion/source/database/common_db_source.py | Reuses an owned connection and closes it when switching databases or shutting down. |
| ingestion/src/metadata/ingestion/source/database/datalake/connection.py | Registers datalake client cleanup with the connection owner. |
| ingestion/src/metadata/ingestion/source/database/salesforce/metadata.py | Reuses one authenticated owner for startup testing and ingestion. |
| ingestion/src/metadata/ingestion/source/database/unitycatalog/connection.py | Adds owned SQL and REST sub-connections with coordinated cleanup. |

</details>

<sub>Reviews (7): Last reviewed commit: ["address review: close-on-failure guard i..."](https://github.com/open-metadata/openmetadata/commit/ce7df5a5567a8b940057e04eb81e4a71660190be) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44463093)</sub>

**Context used:**

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/burstiq/connection.py
+++ b/ingestion/src/metadata/ingestion/source/database/burstiq/connection.py
@@ -13,8 +13,9 @@ Source connection handler for BurstIQ
 """
 
 import hashlib
-from collections import OrderedDict
 from typing import Optional
+
+from cachetools import LRUCache
 
 from metadata.generated.schema.entity.automations.workflow import (
     Workflow as AutomationWorkflow,
@@ -36,7 +37,7 @@ logger = ingestion_logger()
 
 CLIENT_CACHE_SIZE = 100
 
-_CLIENT_CACHE: "OrderedDict[str, BurstIQClient]" = OrderedDict()
+_CLIENT_CACHE: "LRUCache[str, BurstIQClient]" = LRUCache(maxsize=CLIENT_CACHE_SIZE)
 
 
 class BurstIQConnection(BaseConnection[BurstIQConnectionConfig, BurstIQClient]):
@@ -44,11 +45,11 @@ class BurstIQConnection(BaseConnection[BurstIQConnectionConfig, BurstIQClient]):
         """Return a BurstIQ client, cached by a digest of the serialised config."""
         connection = self.service_connection
         key = hashlib.sha256(connection.model_dump_json().encode()).hexdigest()
-        if key not in _CLIENT_CACHE:
-            _CLIENT_CACHE[key] = BurstIQClient(config=connection)
-            if len(_CLIENT_CACHE) > CLIENT_CACHE_SIZE:
-                _CLIENT_CACHE.popitem(last=False)
-        return _CLIENT_CACHE[key]
+        client = _CLIENT_CACHE.get(key)
+        if client is None:
+            client = BurstIQClient(config=connection)
+            _CLIENT_CACHE[key] = client
+        return client
 
     def test_connection(
         self,
--- a/ingestion/src/metadata/ingestion/source/database/burstiq/connection.py
+++ b/ingestion/src/metadata/ingestion/source/database/burstiq/connection.py
@@ -13,7 +13,8 @@ Source connection handler for BurstIQ
 """
 
 import hashlib
-from typing import Dict, Optional  # noqa: UP035
+from collections import OrderedDict
+from typing import Optional
 
 from metadata.generated.schema.entity.automations.workflow import (
     Workflow as AutomationWorkflow,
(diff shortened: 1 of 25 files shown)


IceS2:
Thanks — addressed both:

**1. Leaf sources never released the owned `_connection` (P1 ×6 / Quality)** — fixed in d24283fd11. Rather than repeat the release per source, `DatabaseServiceSource` now has a default `close()` that releases the owner:
```python
def close(self) -> None:
    if self._connection is not None:
        self._connection.close()
        self._connection = None
```
The five leaf sources whose only teardown is the owner drop their `close()` and inherit it. Datalake registered its client teardown on the owner (`_on_close(lambda: client.close(...))` in `DatalakeConnection._get_client`, which does the real GCS temp-file / Azure-client cleanup), so it inherits the base `close()` too — the owner's `close()` now tears down everything it built, including on the `__init__` failure path, with no per-source special case.

**2. Leaf sources assume `create_connection` is non-None (Edge Case)** — intentional. These are migrated leaf connectors that always register a `connection_class`, so `create_connection` never returns `None` and the base-class fallback would be dead code. The base classes keep the fallback only because they also serve non-migrated / custom connectors. The `cast` documents the invariant; if a spec ever regressed, failing loud at `__init__` is preferable to silently falling back to an unowned client, which would reintroduce the double-connection this PR fixes.

IceS2:
Fixed the Salesforce close-on-failure gap in ce7df5a556 — Salesforce now wraps its startup `test_connection()` in the same `try/except: self.close(); raise` the other five leaf sources use, so a failed startup releases the owned connection.

The other open finding (leaf sources assume `create_connection` is non-None) is intentional, as noted earlier: these are migrated leaf connectors that always register a `connection_class`, so the fallback would be dead code; the base classes keep it only because they also serve non-migrated / custom connectors.

pmbrull (approved):

Issue #27107, opened 2026-04-06
Soft deleted users still appear in Experts/Reviewers field in Data Product


Prajwal214:
**Affected module**
Does it impact the UI, backend or Ingestion Framework? - Backend

**Describe the bug**
- Soft deleted users are still visible in the Experts and Reviewers fields of a Data Product. These users should not appear once they are soft deleted.

**To Reproduce**

Screenshots or steps to reproduce
-Create or assign a user as Expert or Reviewer in a Data Product.
- Soft delete the user.
- Navigate to the Data Product.
- Observe that the soft deleted user is still present in the Experts/Reviewers field.

**Expected behavior**
- Soft deleted users should not appear in the Experts or Reviewers field.

**Version:**
 - Python version:
 - OpenMetadata version: [e.g. 0.8]
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

PR #30118, opened 2026-07-16
Fixes #30117: never surface soft-deleted users as owners/followers/experts/reviewers under include=all

### LINKED ISSUE #30117: Editing owners fails with "array item index out of range" when an owner user is soft-deleted
### Summary

Editing the **owners** of an entity fails when one of the current owners is a **soft-deleted user**:

```
PATCH /api/v1/dataQuality/testSuites/{id}
[JsonException] An array item index is out of range. Index: 6, Size: 6
```

The soft-deleted user also still appears in the owner list, and the user cannot remove them. Reported by a customer on **1.13.1** for a Test Suite, but it is generic — it affects **any entity** edited from a detail page that fetches with `include=all` (Database, Database Schema, Service, Team, Persona, Test Suite, …).

### Root cause

Two layers.

**1. The crash — a client ↔ server owner-array mismatch on a positional JSON Patch.**
- The UI loads the entity with `include=all` (e.g. `TestSuiteDetailsPage` → `getTestSuiteByName(fqn, { fields:[owners], include: Include.All })`). Under `include=all` the single-entity GET still resolves owners **including the soft-deleted user**, so the client's JSON-Patch base (`compare(entity, updated)`) has *N* owners.
- The server's `EntityRepository.patch()` loads the patch `original` with `Include.NON_DELETED`, which **drops** the soft-deleted owner → *N-1* owners.
- Applying the client's patch (which references `/owners/{N-1}`) to the *N-1*-sized server array throws `array item index out of range`. (The relationship add/remove logic, `updateOwners`, is already ID-based and robust — only the positional `applyPatch` step is fragile.)

**2. The deeper problem — soft-deleted users are hidden from owner lists but are still assigned.**
- Soft-delete only sets `deleted = true`; it **keeps all relationship rows** (`OWNS`, `EXPERT`, `REVIEWS`, `FOLLOWS`). Only *hard* delete runs `cleanup()` → `relationshipDAO.deleteAll(...)`. So the soft-deleted user is genuinely still an owner in the DB.
- Recent changes made most read paths **hide** these soft-deleted users instead of surfacing them. The result is an entity that is still owned by a user you can neither see (on filtered paths) nor remove (the edit crashes) — an invisible, un-cleanable dangling ownership.

### PRs that led to this (all confirmed in the `1.13.1-release` tag)

| PR | What it did | Merged |
|---|---|---|
| #25284 — *Filter deleted users from ownership relationships in GET operations* | Introduced `NON_DELETED` filtering of **owners** on GET. Authored by the `Copilot` agent; merged 2026-01-19. | ✅ in 1.13.1 |
| #27120 — *soft-deleted users still appear in Experts/Reviewers across all entities* (fixes #27107) | Extended the same `NON_DELETED` filtering to **owners / followers / experts / reviewers** (bulk + default-include paths). Merged 2026-04-24. | ✅ in 1.13.1 |

Both verified as ancestors of the `1.13.1-release` tag (cut 2026-06-27). Neither PR closed the **single-entity `include=all`** path, so today the behavior is inconsistent: soft-deleted owners are hidden on list/default-GET paths but returned on single-entity `include=all` — which is exactly the divergence that triggers the crash.

### Why "hide the soft-deleted user" is the wrong direction

Filtering soft-deleted users out of owner/expert/reviewer lists diverges from the industry norm and from OpenMetadata's own semantics:
- Mature systems (Jira, GitHub, ServiceNow, Google Workspace) **keep** a deactivated user's existing assignments, render them as *inactive* (badge/tombstone), and let you **reassign or remove** them. They exclude deactivated users only from *new-assignment pickers*. They do not silently drop existing assignments — that destroys audit/accountability and creates phantom "ownerless" records.
- OpenMetadata soft-delete is **reversible** (a soft-deleted user can be restored), i.e. it is *deactivation*, not deletion. Hiding a reference that reappears on restore is self-inconsistent.

### Recommended direction (for the owning team to decide)

- **Show** soft-deleted users in owner/expert/reviewer lists, flagged as inactive; **allow removal/reassignment** through the UI.
- Fix the crash where it actually lives: the PATCH must load `original` with the soft-deleted owners **included**, so (a) the positional patch stays in range and (b) the ID-based `updateOwners` diff detects and removes the dangling `OWNS` row. (Optionally harden by having the owner widget send a whole-array `replace /owners` instead of a positional diff.)
- Exclude soft-deleted users only from the **add-owner picker**.
- Revisit #27120's approach for **experts / reviewers / followers** — they carry the identical "hidden but still assigned" flaw and the same latent crash.

### Steps to reproduce

1. Create an entity (e.g. a logical Test Suite) with ≥2 owners; make one owner's name sort **last**.
2. Soft-delete that last owner (user).
3. Open the entity page (fetched with `include=all`) → the deleted user still shows as an owner.
4. Try to remove/change owners → `400 An array item index is out of range. Index: N, Size: N`.

### Customer context

- Customer: Remitly · Version: 1.13.1 · SaaS · Priority: high
- Deleted owner: `nickbu` on the *TML Fraud Mart Test Suite*.
- **Workaround:** permanently (hard) delete the removed user to purge the orphaned `owns` relationship, which unblocks owner edits.

cc @TeddyCr @yan-3005 @mohityadav766 @sonika-shah — flagging for the design call on hide-vs-show for soft-deleted user references.

manerow:
Fixes #30117

## Problem

Editing the **owners** of an entity fails when one of the current owners is a **soft-deleted user**:

```
PATCH /api/v1/dataQuality/testSuites/{id}
[JsonException] An array item index is out of range. Index: 6, Size: 6
```

The soft-deleted user also keeps showing in the owner list, so the user can't remove them. Reported by a customer on 1.13.1 for a Test Suite, but it is generic — it hits **any entity** edited from a detail page that fetches with `include=all` (Database, Database Schema, Service, Team, Persona, Test Suite, …).

## Root cause

A client ↔ server **owner-array size mismatch**:

- The UI loads the entity with `include=all`; under `include=all` the server resolves owners **including soft-deleted users**, so the JSON-Patch base (`compare(entity, updated)`) has *N* owners.
- `EntityRepository.patch()` loads the patch `original` with `NON_DELETED`, which **drops** the soft-deleted owner → *N-1* owners.
- Applying the client patch (which references `/owners/{N-1}`) against the *N-1*-sized array throws `array item index out of range`.

This is the same class of bug as #27107 / #27120 (soft-deleted users leaking into experts/reviewers/owners). #27120 fixed the bulk-list and default-include paths, but the **single-entity GET path still honored `include=all`** for owner/expert/reviewer/follower resolution, so soft-deleted users still leaked there.

## Fix

Soft-deleted users are never valid owners/followers/experts/reviewers, so these **user-identity relationship fields always resolve with `NON_DELETED`, regardless of the entity-level `include`** — completing #27120 for the single-entity read path. This is done at the single choke point both the read planner and `setFieldsInternal` route through: `RelationIncludes.getIncludeFor`.

- An explicit per-field override (`include-relations=owners:all`) remains authoritative.
- Non-user relationship fields (`domains`, `dataProducts`, `children`, …) are unchanged.

With this, the UI's `include=all` view and the server's `NON_DELETED` patch load agree, so owner edits succeed and soft-deleted users no longer appear as owners.

## Test plan

- Unit: `EntityUtilTest.testUserIdentityRelationFieldsForceNonDeletedUnderIncludeAll` — owners/followers/experts/reviewers force `NON_DELETED` under `include=all`; non-user fields keep the requested include; explicit override wins.
- Existing `RelationIncludes` / `ReadPlanner` tests remain green (their differing values come from explicit per-field overrides).
- Manual repro (before → after): create a logical Test Suite with ≥2 owners, soft-delete the last-sorting owner, then remove owners via `PATCH`. Before: `400 array item index out of range`. After: succeeds and the deleted user no longer shows as owner.

This PR prevents soft-deleted user identities from appearing through the default `include=all` relation path. The main changes are:

- Forces owners, followers, experts, and reviewers to use `NON_DELETED` by default.
- Keeps explicit per-field relation overrides authoritative.
- Adds unit coverage for identity fields, other relations, and override precedence.
</details>

The explicit owner include override can still trigger the PATCH array mismatch.

- The default `include=all` path now filters soft-deleted user identities.
- `owners:all` still exposes deleted owners while PATCH reconstructs the entity with `NON_DELETED`.
- The resulting arrays can have different sizes and reject a valid client patch.

openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java | Adds a default non-deleted include policy for user-identity relations, but explicit overrides can restore the GET/PATCH mismatch. |
| openmetadata-service/src/test/java/org/openmetadata/service/util/EntityUtilTest.java | Adds focused tests for identity relation filtering, unaffected relation fields, and explicit override precedence. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(entity): never surface soft-deleted ..."](https://github.com/open-metadata/openmetadata/commit/2c3222f32b760fb78b15292d40294178aff093b6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44718282)</sub>

</details>

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/util/EntityUtil.java
@@ -17,6 +17,10 @@ import static org.openmetadata.common.utils.CommonUtil.listOrEmpty;
 import static org.openmetadata.common.utils.CommonUtil.nullOrEmpty;
 import static org.openmetadata.schema.type.Include.ALL;
 import static org.openmetadata.schema.type.Include.NON_DELETED;
+import static org.openmetadata.service.Entity.FIELD_EXPERTS;
+import static org.openmetadata.service.Entity.FIELD_FOLLOWERS;
+import static org.openmetadata.service.Entity.FIELD_OWNERS;
+import static org.openmetadata.service.Entity.FIELD_REVIEWERS;
 import static org.openmetadata.service.jdbi3.ListFilter.NULL_PARAM;
 import static org.openmetadata.service.jdbi3.RoleRepository.DOMAIN_ONLY_ACCESS_ROLE;
 import static org.openmetadata.service.security.DefaultAuthorizer.getSubjectContext;
@@ -630,6 +634,12 @@ public final class EntityUtil {
    */
   @Getter
   public static class RelationIncludes {
+    // Soft-deleted users must never surface as owners/followers/experts/reviewers, even under
+    // include=all: returning them desyncs the client's PATCH base (fetched with include=all) from
+    // the server's NON_DELETED patch load, causing "array item index out of range" on owner edits.
+    private static final Set<String> USER_IDENTITY_RELATION_FIELDS =
+        Set.of(FIELD_OWNERS, FIELD_FOLLOWERS, FIELD_EXPERTS, FIELD_REVIEWERS);
+
     private final Include defaultInclude;
     private final Map<String, Include> fieldIncludes;
 
@@ -649,7 +659,14 @@ public final class EntityUtil {
     }
 
     public Include getIncludeFor(String fieldName) {
-      return fieldIncludes.getOrDefault(fieldName, defaultInclude);
+      Include explicitInclude = fieldIncludes.get(fieldName);
+      if (explicitInclude != null) {
+        return explicitInclude;
+      }
+      if (USER_IDENTITY_RELATION_FIELDS.contains(fieldName)) {
+        return NON_DELETED;
+      }
+      return defaultInclude;
     }
 
     public boolean hasFieldSpecificInclude(String fieldName) {
--- a/openmetadata-service/src/test/java/org/openmetadata/service/util/EntityUtilTest.java
+++ b/openmetadata-service/src/test/java/org/openmetadata/service/util/EntityUtilTest.java
@@ -283,6 +283,26 @@ class EntityUtilTest {
   }
 
   @Test
+  void testUserIdentityRelationFieldsForceNonDeletedUnderIncludeAll() {
+    // Under include=all, soft-deleted users must not leak into owners/followers/experts/reviewers,
+    // otherwise the client's include=all PATCH base desyncs from the server's NON_DELETED patch
+    // load and owner edits fail with "array item index out of range".
+    EntityUtil.RelationIncludes includeAll = EntityUtil.RelationIncludes.fromInclude(Include.ALL);
+    assertEquals(Include.NON_DELETED, includeAll.getIncludeFor("owners"));
+    assertEquals(Include.NON_DELETED, includeAll.getIncludeFor("followers"));
+    assertEquals(Include.NON_DELETED, includeAll.getIncludeFor("experts"));
+    assertEquals(Include.NON_DELETED, includeAll.getIncludeFor("reviewers"));
+    // Non-user relationship fields still honor the requested include.
+    assertEquals(Include.ALL, includeAll.getIncludeFor("domains"));
+    assertEquals(Include.ALL, includeAll.getIncludeFor("dataProducts"));
+    assertEquals(Include.ALL, includeAll.getIncludeFor("unknown"));
+    // An explicit per-field override remains authoritative.
+    EntityUtil.RelationIncludes explicitOwners =
+        new EntityUtil.RelationIncludes(Include.ALL, "owners:all");
+    assertEquals(Include.ALL, explicitOwners.getIncludeFor("owners"));
+  }
+
+  @Test
   void testVersionAndFieldNameHelpers() {
     Column column = new Column().withName("amount");
     Topic topic = new Topic().withFullyQualifiedName("service.topic");


manerow:
Closing this in favor of a different direction.

This PR took the **hide** approach — filtering soft-deleted users out of owner/follower/expert/reviewer resolution so the client and server agree. Discussion on #30117 concluded that hiding a **still-assigned** soft-deleted user (the OWNS relationship persists through soft-delete) diverges from the industry norm and from OpenMetadata's own reversible-soft-delete (= deactivation) semantics.

The better fix is to **show** soft-deleted users (flagged inactive) and let them be **removed/reassigned**, fixing the crash at the PATCH layer (load the patch `original` with soft-deleted owners included) rather than hiding them on read. Because that flips the direction established by #25284 and #27120 — and applies equally to experts/reviewers/followers — it's a design call for the owning team.

Delegating to the owners of #30117. See the updated issue for the full root cause, PR lineage (both confirmed in 1.13.1), and recommended direction.

PR #30174, opened 2026-07-17
Fixes #30173: fix(ingestion): security/metadata/mcp sources own their BaseConnection; dedupe close-on-failure

### LINKED ISSUE #30173: Security, metadata and mcp sources open two connections; close-on-failure has drifted across the service bases
The `source-owns-BaseConnection` work is complete for the 9 standard service types (#29870 dashboard, #30094 database, #30148 pipeline, #30133 remaining). Two loose ends are left on the OSS side.

## 1. Sources that still open two connections

`SecurityServiceSource.__init__` builds **two** clients: one through the owner-discarding `get_connection` seam and a second one through its `get_client` hook. It then calls `self.client.test_connection()` and **discards the result**, so a failing connection never raises and ingestion proceeds against a broken client.

The `metadata` (atlas, amundsen, alationsink) and `mcp` sources have no service base, but each carries the same pattern in its own `__init__`: a client is built through the seam, then `test_connection_common` builds a second one to test with.

All of these should own a single `BaseConnection` and reuse it for the test step, like the migrated bases do.

## 2. `close_on_failure`

Every service base re-derives `try: self.test_connection() except: <dispose>; raise` in its constructor, and the shape has drifted: most call `self.close()`, pipeline disposes `self._connection` directly.

`self.close()` is the wrong teardown for a constructor failure. It is successful-run teardown, and it is polymorphic — it dispatches to subclass overrides that touch attributes set *after* `super().__init__()` returns (airflow's `self.session`, databrickspipeline's `self._table_lookup_cache`), so an `AttributeError` masks the connection error the user needs to see. `saperp`'s override goes further and closes the OpenMetadata client.

A single `close_on_failure` context manager in `source/connections.py` releases only the owner — the only resource in flight when the test step runs.

## Scope

- Migrate `security_service.py` plus the four `metadata`/`mcp` leaves; audit `drives/` (dead package, N/A).
- Add `close_on_failure` and apply it to every base and leaf that owns a connection.
- Fix `airflow` and `openlineage` `close()`, which never call `super().close()` and so never dispose the owner.

Additive and backward-compatible: `connection_obj`, `test_connection_common`'s signature and the `get_connection`/`import_connection_fn` seam are all preserved for Collate and custom connectors.

Tracked PR: #30174

IceS2:
Fixes #30173

Closes out the OSS side of `source-owns-BaseConnection`. Two commits, splittable.

## 1. Security, metadata and mcp sources own their BaseConnection (`62577c2381`)

`SecurityServiceSource` built **two** clients — one through the owner-discarding `get_connection` seam, a second through its `get_client` hook — and then discarded the test-connection result, so a failing connection never raised. The base now owns one `BaseConnection`, takes its client from the owner, tests through `run_test_connection` (which raises), and disposes it on `close()`. `get_client` is demoted to the legacy path for connectors without a `connection_class`, and is now declared on the base instead of being an implicit contract.

**Key finding:** for the only consumer of this base, the client's constructor performs a real login. Routing just the unused `self.connection` attribute through the owner would have left the second login in place — the fix only works because `self.client` comes from the owner too.

The `metadata` (atlas, amundsen, alationsink) and `mcp` sources have no service base but carry the same pattern in their own `__init__`. Each now owns its connection and tests through it. `drives/` (plural) is a dead package containing only `__init__.py` — nothing to migrate.

## 2. Dedupe close-on-failure behind a context manager (`c4fcc7319b`)

`close_on_failure` in `source/connections.py` replaces the drifted `try: self.test_connection() except: <dispose>; raise` in 24 constructors.

`self.close()` was the wrong teardown for a constructor failure: it is successful-run teardown, and it is polymorphic, so it dispatched to subclass overrides touching attributes set *after* `super().__init__()` returned — airflow's `self.session`, databrickspipeline's `self._table_lookup_cache` — masking the real connection error behind an `AttributeError`. `saperp`'s override closes the OpenMetadata client outright. The context manager releases only the owner, which is the only resource in flight when the test step runs.

`CommonDbSourceService` keeps an explicit failure branch: its SSL temp files are owned by the source, not the connection, until they move into a `BaseConnection` hook.

## Pre-existing bugs fixed while here

| | |
|---|---|
| `airflow`, `openlineage` | `close()` never called `super().close()`, so the owner was never disposed. airflow closes its session before the engine dispose `super().close()` triggers. |
| `OpenLineageConnection` | never registered the Kafka consumer teardown |
| `SftpConnection` | never registered the SFTP transport teardown |
| `AmundsenConnection` | never registered the Neo4j driver teardown |

The last three leaked one client per *temporary* owner — i.e. on every Automation test-connection.

## Compatibility

Additive. `connection_obj` is still set, `test_connection_common` keeps its 3-arg signature, and the `get_connection` / `import_connection_fn` seam is untouched — Collate and custom connectors are unaffected.

This PR moves ingestion sources to owned connections and centralizes failure cleanup. The main changes are:

- Reuses one owned client for security, metadata, and MCP sources.
- Adds `close_on_failure` for constructor verification failures.
- Registers missing SFTP, Amundsen, Kafka, and Kinesis cleanup callbacks.
- Routes Airflow and OpenLineage cleanup through their base source.
</details>

This looks safe to merge.

- Verification failures remain visible when owned connection cleanup raises.
- The Kinesis client is now registered with the connection owner.
- No blocking incomplete or unsafe fix remains in the reviewed lifecycle paths.

</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/source/connections.py | Adds shared failure cleanup that preserves the connection verification error when normal teardown fails. |
| ingestion/src/metadata/ingestion/source/pipeline/openlineage/connection.py | Registers both Kafka and Kinesis clients with the owning connection for cleanup. |
| ingestion/src/metadata/ingestion/source/security/security_service.py | Uses one owned connection and client for login, verification, ingestion, and cleanup. |
| ingestion/src/metadata/ingestion/source/database/common_db_source.py | Keeps explicit legacy engine and SSL cleanup while using owner cleanup for migrated connectors. |

</details>

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/security-an..."](https://github.com/open-metadata/openmetadata/commit/eab63ec4605a04b6e3b36df8eafce590eaf93915) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45043686)</sub>

</details>


IceS2:
Review addressed in `69330991b0`. All three findings were valid.

**gitar: legacy-path engine leaked on test-connection failure** — correct, and it was a regression I introduced. `close_on_failure` releases only the owner, but a connector without a `connection_class` gets its engine from the legacy seam, where nobody owns it; the previous `self.close()` disposed it via `_release_engine()`. This is live for the connectors that resolve through the seam rather than a `connection_class`. The failure branch now calls `_release_engine()` when `self._connection is None`. Locked by `test_legacy_engine_disposed_when_test_connection_fails`, verified failing without the fix.

**greptile: cleanup masks connection failure** — valid. Not a regression (the old `self.close()` had the same exposure), but this PR exists to stop teardown from masking the connection error, so leaving it would have been inconsistent. A failing teardown is now logged and the verification error propagates.

**greptile: Kinesis client remains unowned** — valid. Registering the teardown for the Kafka branch and not its sibling was an inconsistency on my side. Both branches now register it.

Two more changes from self-review:

- `SecurityServiceSource.get_client` is gone. It duplicated the seam: for the only implementation, `get_client()` and the seam's `get_connection()` build the same client the same way, and the seam already covers connectors without a `connection_class`.
- atlas, amundsen, alationsink and mcp all ship a `connection_class`, so `create_connection` never returns `None` for them. The `else get_connection(...)` fallback was unreachable; they now take the client from the owner with a `cast`, matching the other migrated leaf sources.

Copilot could not review (quota).

pmbrull (approved):

PR #30213, opened 2026-07-19
fix(migration): preserve task creator/comments/reactions/resolution and replay incident state on 2.0 upgrade

sonika-shah:
## Problem

On a `1.12.1 / 1.13 → 2.0` upgrade, the thread→task migration preserves the task rows and their type/status mapping, but drops several fields and resets incident state. On migrated (pre-existing) data:

- **Created By is blank** — the migration never carries the original requester onto the task.
- **Completed tasks lose their resolver** — `resolution.resolvedBy` is empty and the comment is a generic `"Migrated from thread-based task system"`.
- **Task comments are dropped** — the thread's posts (replies + resolve/close records) don't reach `task_entity.comments`.
- **Task reactions are dropped**.
- **Migrated incidents reset to New / No Owners** — the incident workflow restarts at `NewStage` and never replays the recorded `test_case_resolution_status` lifecycle, so a pre-upgrade Ack/Assigned incident loses its stage and assignee (and the stale state then makes the "Update Status" resolve fail).

New tasks/incidents created *in* 2.0 are unaffected — this is migration-only.

## Fix

`v200/MigrationUtil.java` — in the thread→task migration:
- set `createdBy` (entityReference) in addition to `createdById`;
- set `resolution.resolvedBy` from the legacy `closedBy` (the original close note now survives as a migrated comment, so the placeholder is dropped);
- migrate the thread's `posts[]` into task `comments[]` (+ `commentCount`);
- carry thread-level `reactions[]` onto the task.

For incidents, after `backfillOpenTasksToWorkflowInstances` starts the workflow, replay the test case's latest resolution status onto the instance, reusing the existing runtime bridge (`resolveLegacyTransitionId` / `buildLegacyResolvedPayload`) so migration and runtime stay consistent. New public entry point: `TestCaseResolutionStatusRepository.backfillMigratedIncidentTaskStage`. Idempotent — no-ops when the task is already at the recorded stage.

Complementary to #29978 (which restores the `ASSIGNED_TO` / `MENTIONED_IN` relationships on Postgres); this PR restores the field- and incident-state fidelity that #29978 does not cover.

## Testing

- `openmetadata-service` compiles; spotless applied.
- **Reviewers:** please verify the incident replay end-to-end on a migrated instance — a test case that was Ack/Assigned before upgrade should come back at the same stage with its assignee, and "Update Status" should transition without error. (This path touches Flowable workflow state and warrants a live migration check.)

---

## Additional fix: task resolve/close change events (2.0 regression)

[...]

PR #30248, opened 2026-07-20
fix(migration): preserve task creator/comments/resolution and replay incident state on 2.0 upgrade

sonika-shah:
## Problem

On a `1.12.1 / 1.13 → 2.0` upgrade, the thread→task migration preserves the task rows and their type/status mapping, but drops several fields and resets incident state. On migrated (pre-existing) data:

- **Created By is blank** — the migration never carries the original requester onto the task.
- **Completed tasks lose their resolver** — `resolution.resolvedBy` is empty and the comment is a generic `"Migrated from thread-based task system"`.
- **Task comments are dropped** — the thread's posts (replies + resolve/close records) don't reach `task_entity.comments`.
- **Migrated incidents reset to New / No Owners** — the incident workflow restarts at `NewStage` and never replays the recorded `test_case_resolution_status` lifecycle, so a pre-upgrade Ack/Assigned incident loses its stage and assignee (and the stale state then makes the "Update Status" resolve fail).

New tasks/incidents created *in* 2.0 are unaffected — this is migration-only.

## Fix

`v200/MigrationUtil.java` — in the thread→task migration:
- set `createdBy` (entityReference) in addition to `createdById`;
- set `resolution.resolvedBy` from the legacy `closedBy` (the original close note now survives as a migrated comment, so the placeholder is dropped);
- migrate the thread's `posts[]` into task `comments[]` (+ `commentCount`), with per-post reactions riding along in `taskComment.reactions`.

> Note: `Task` has no top-level `reactions` field, so thread-level task reactions are intentionally not carried over (only per-comment reactions survive). Writing a top-level `reactions` field corrupts the row on read — see the dedicated fix below.

For incidents, after `backfillOpenTasksToWorkflowInstances` starts the workflow, replay the test case's latest resolution status onto the instance, reusing the existing runtime bridge (`resolveLegacyTransitionId` / `buildLegacyResolvedPayload`) so migration and runtime stay consistent. New public entry point: `TestCaseResolutionStatusRepository.backfillMigratedIncidentTaskStage`. Idempotent — no-ops when the task is already at the recorded stage.

Complementary to #29978 (which restores the `ASSIGNED_TO` / `MENTIONED_IN` relationships on Postgres); this PR restores the field- and incident-state fidelity that #29978 does not cover. This branch is merged up to current `main`, so it already includes #29978.

## Additional fix: task resolve/close change events (2.0 regression)

**Symptom (verified on 1.12.1→2.0, Postgres, in the `change_event` table):** in 1.12.1, resolving/closing a task emitted change events (`taskResolved`, `taskClosed` under `entityType=THREAD`), which drove webhook/email/Slack alerts. After the 2.0 task refactor, resolving/closing a migrated task produces **zero** change events — so task-resolution alerts silently stop working after upgrade.

**Origin — PR #25894 (Task redesign):** the pre-redesign feed endpoints returned via RestUtil's `.toResponse()` wrapper, which sets the `X-OpenMetadata-Change` header automatically — so `ChangeEventHandler` recorded a `taskResolved`/`taskClosed` event. PR #25894's new `TaskResource` `resolveTask`/`closeTask`/`suggestion-apply` endpoints return a bare `Response.ok(task).build()`, so no change event is recorded. The `add-comment` endpoint (header added in #29023) proves the mechanism is intended for tasks; resolve/close were simply missed.

**Fix:** set `RestUtil.CHANGE_CUSTOM_HEADER = ENTITY_UPDATED` on the resolve/close/suggestion-apply responses, restoring task lifecycle events (and thus alerting). `create` already fires `entityCreated` via the 201 path.

## Testing

- `openmetadata-service` compiles (full `mvn clean install -pl openmetadata-service,openmetadata-integration-tests -am -DskipTests` → BUILD SUCCESS); spotless applied.
- **Reviewers:** please verify the incident replay end-to-end on a migrated instance — a test case that was Ack/Assigned before upgrade should come back at the same stage with its assignee, and "Update Status" should transition without error.

---

Re-raised from the same-repo branch because the fork-based PR is blocked by GitHub's new `pull_request_target` fork-checkout guard. Supersedes #30213 (identical diff).

This PR preserves legacy task and incident state during the 2.0 migration. The main changes are:

- Migrates task creators, resolvers, comments, and per-comment reactions.
- Replays recorded incident stages after workflow instances are created.
- Retries incident replay on later migration runs after transient failures.
- Restores change events for task resolution, closure, and suggestion application.
</details>

This looks safe to merge.

- Incident replay now runs independently of one-time workflow creation.
- Exceptions during replay remain visible and can be retried on a later migration run.
- No blocking issue was found in the updated code.

</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseResolutionStatusRepository.java | Exposes the existing legacy-status bridge so migration code can replay recorded incident stages. |
| openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java | Preserves legacy task metadata and retries incident-state replay independently of workflow creation. |
| openmetadata-service/src/main/java/org/openmetadata/service/resources/tasks/TaskResource.java | Adds entity-update headers to task lifecycle responses so change events are emitted. |

</details>

<sub>Reviews (8): Last reviewed commit: ["refactor: drop heavy JSON conversions in..."](https://github.com/open-metadata/openmetadata/commit/e590b952caa6ef2f94053f432df19af9c6aca829) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45675772)</sub>

**Context used:**

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
@@ -1628,15 +1628,16 @@ public class MigrationUtil {
    * falls back to the admin user, never a fabricated id. Throws only if admin is also unresolvable.
    */
   static EntityReference resolveUserReference(String userName) {
-    if (nullOrEmpty(userName)) {
-      return resolveAdminReference();
+    EntityReference reference = null;
+    if (!nullOrEmpty(userName)) {
+      try {
+        reference = Entity.getEntityReferenceByName(Entity.USER, userName, Include.ALL);
+      } catch (Exception e) {
+        LOG.debug(
+            "Unable to resolve user '{}', falling back to admin: {}", userName, e.getMessage());
+      }
     }
-    try {
-      return Entity.getEntityReferenceByName(Entity.USER, userName, Include.ALL);
-    } catch (Exception e) {
-      LOG.debug("Unable to resolve user '{}', falling back to admin: {}", userName, e.getMessage());
-      return resolveAdminReference();
-    }
+    return reference != null ? reference : resolveAdminReference();
   }
 
   private static EntityReference resolveAdminReference() {
@@ -1648,12 +1649,13 @@ public class MigrationUtil {
    * POJOs). Returns null only if even the admin fallback cannot be resolved.
    */
   private static ObjectNode buildUserRef(String userName) {
+    ObjectNode ref = null;
     try {
-      return userRefToJson(resolveUserReference(userName));
+      ref = userRefToJson(resolveUserReference(userName));
     } catch (Exception e) {
       LOG.warn("Could not resolve a user reference for '{}': {}", userName, e.getMessage());
-      return null;
     }
+    return ref;
   }
 
   private static ObjectNode userRefToJson(EntityReference ref) {
@@ -1672,26 +1674,34 @@ public class MigrationUtil {
     ArrayNode comments = JsonUtils.getObjectNode().arrayNode();
     if (posts != null && posts.isArray()) {
       for (JsonNode post : posts) {
-        if (!post.has("message")) {
-          continue;
+        ObjectNode comment = buildTaskComment(post);
+        if (comment != null) {
+          comments.add(comment);
         }
-        ObjectNode author = buildUserRef(post.path("from").asText(null));
-        if (author == null) {
-          continue;
-        }
-        ObjectNode comment = JsonUtils.getObjectNode();
-        comment.put("id", post.has("id") ? post.get("id").asText() : UUID.randomUUID().toString());
-        comment.put("message", post.get("message").asText());
-        comment.set("author", author);
-        comment.put("createdAt", post.has("postTs") ? post.get("postTs").asLong() : 0);
-        if (post.has("reactions") && post.get("reactions").isArray()) {
-          comment.set("reactions", post.get("reactions"));
-        }
-        comments.add(comment);
       }
     }
     taskJson.set("comments", comments);
     taskJson.put("commentCount", comments.size());
+  }
+
+  /**
+   * Build a {@code taskComment} from a legacy post, or {@code null} to skip it (no message, or an
+   * author that cannot be resolved even to the admin fallback). Per-post reactions are carried over.
+   */
+  private static ObjectNode buildTaskComment(JsonNode post) {
+    ObjectNode comment = null;
+    ObjectNode author = post.has("message") ? buildUserRef(post.path("from").asText(null)) : null;
+    if (author != null) {
+      comment = JsonUtils.getObjectNode();
+      comment.put("id", post.has("id") ? post.get("id").asText() : UUID.randomUUID().toString());
+      comment.put("message", post.get("message").asText());
+      comment.set("author", author);
+      comment.put("createdAt", post.has("postTs") ? post.get("postTs").asLong() : 0);
+      if (post.has("reactions") && post.get("reactions").isArray()) {
+        comment.set("reactions", post.get("reactions"));
+      }
+    }
+    return comment;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
```diff
@@ -2169,13 +2249,36 @@ private int backfillOpenTasksToWorkflowInstances() {
...
+                Entity.getEntityTimeSeriesRepository(Entity.TEST_CASE_RESOLUTION_STATUS);
+        if (incidentRepository.backfillMigratedIncidentTaskStage(task)) {
+          LOG.info("Replayed pre-migration incident state onto task {}", task.getId());
+        }
+      } catch (Exception e) {
+        LOG.warn(
+            "Failed to replay pre-migration incident state for task {}: {}",
+            task.getId(),
+            e.getMessage());
+      }
```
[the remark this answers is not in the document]
sonika-shah: Fixed in b2ff068365.

The replay is no longer coupled to the instance-creation guard. The loop now keeps the one-shot trigger behind `workflowInstanceId == null`, but attempts the incident replay for **every** open incident task on **every** run:

```java
if (task.getWorkflowInstanceId() == null && startWorkflowInstanceForTask(task)) {
  backfilled++;
}
if (task.getCategory() == TaskCategory.Incident && !replayMigratedIncidentState(task)) {
  failedIncidentReplays++;
}
```

This is safe because `backfillMigratedIncidentTaskStage` is idempotent — `resolveLegacyTransitionId` returns null (no-op) once the task is already at its recorded stage — so a re-run only re-applies the replays that actually failed and leaves successful ones untouched. That keys re-run eligibility off the workflow stage (via the callee's existing check) rather than `workflowInstanceId`, so a transient first-run failure is recoverable by re-running the migration, consistent with the rest of this migration's "re-run to retry" contract.

`replayMigratedIncidentState` now returns a boolean, and failures are tallied and surfaced so operators know remediation is needed:

```
LOG.warn("{} migrated incident task(s) could not replay their pre-upgrade state; re-run the migration or remediate manually", failedIncidentReplays);
```

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
```diff
@@ -2169,13 +2249,36 @@ private int backfillOpenTasksToWorkflowInstances() {
...
+    }
```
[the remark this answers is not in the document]
sonika-shah: Fixed in b2ff068365 (same as the parallel review thread).

The replay is decoupled from the instance-creation guard. The one-shot trigger stays behind `workflowInstanceId == null`, but the incident replay now runs for every open incident task on every run:

```java
if (task.getWorkflowInstanceId() == null && startWorkflowInstanceForTask(task)) {
  backfilled++;
}
if (task.getCategory() == TaskCategory.Incident && !replayMigratedIncidentState(task)) {
  failedIncidentReplays++;
}
```

Safe because `backfillMigratedIncidentTaskStage` is idempotent — it no-ops once the task is at its recorded stage — so a re-run re-applies only the replays that actually failed. This keys re-run eligibility off the workflow stage (via the callee's existing check) rather than `workflowInstanceId`, so a transient failure is recoverable by re-running the migration. `replayMigratedIncidentState` now returns a boolean, and failed replays are tallied and surfaced via a WARN so operators know remediation is needed — covering both suggestions in the comment.

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
```diff
@@ -1613,6 +1621,78 @@ private static String lookupUserId(Handle handle, String userName) {
...
 
+  /**
+   * Resolve a legacy username to a user {@link EntityReference}; shared by both migration paths.
+   * A soft-deleted user keeps its real id; anything unresolvable (hard-deleted / "system" / null)
+   * falls back to the admin user, never a fabricated id. Throws only if admin is also unresolvable.
+   */
+  static EntityReference resolveUserReference(String userName) {
+    if (nullOrEmpty(userName)) {
+      return resolveAdminReference();
+    }
```
harshach, line 1640: should there be else instead of depending on the return?

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
```diff
@@ -1613,6 +1621,78 @@ private static String lookupUserId(Handle handle, String userName) {
...
+    try {
+      return Entity.getEntityReferenceByName(Entity.USER, userName, Include.ALL);
+    } catch (Exception e) {
+      LOG.debug("Unable to resolve user '{}', falling back to admin: {}", userName, e.getMessage());
+      return resolveAdminReference();
```
harshach: can this be written better we are calling resolveAdminReference() twice in the same method , can the conditions be collapsed

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
sonika-shah: Done in 894897f3 — single trailing return now (result var, no scattered returns).

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
sonika-shah: Done in 894897f3 — collapsed to one admin lookup at the end; resolveAdminReference removed.

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
```diff
@@ -2124,58 +2213,118 @@ private MigrationStats migrateLegacyThreadTasks() {
 
     private int backfillOpenTasksToWorkflowInstances() {
```
TeddyCr, line 2214: @sonika-shah what does there replay part of this file do exactly?

sonika-shah: @TeddyCr  During migration, an incident's task comes back at the workflow's starting new stage — so it loses whatever state it was in (Acknowledged, or Assigned to someone). 

`replayMigratedIncidentState` reads the incident id stored on the migrated task, looks up that incident's last known status, and moves the workflow to that status — reusing the same code that already runs when someone changes an incident's status in the app (so migration and normal usage stay consistent).

 Without it, every migrated incident comes back as New / No Owners, losing its assignee and progress.

 The comment on the line just notes that this replay isn't behind the workflowInstanceId guard above it: it's safe to repeat (does nothing if the incident is  already at the right status), so a one-off failure is fixed by simply re-running the migration.

yan-3005 (approved):

PR #30120, opened 2026-07-16
Fixes 30119: Stabilize integration-test CI

harshach:
### Describe your changes:

Fixes #30119

This change makes Java, Playwright, and Python integration-test outcomes deterministic without reducing parallel throughput.
It builds each Java runtime once before four concurrent isolation lanes, classifies assertion, setup, infrastructure, missing-report, cancellation, and hung-test outcomes, and retains fail-closed diagnostics.
It also fixes hard-delete cache consistency, stale context/search updates, bulk-update fallback ordering, restore/test-suite update races, and retry-dependent Oracle, Trino, and Playwright readiness checks.
The Playwright zero-retry policy remains shadow-only while stable retry signatures are collected.

#
### Type of change:

- [x] Bug fix

#
### High-level design:

- Immutable, checksummed, source-SHA-bound Maven reactor artifact shared by profile lanes
- Concurrent `parallel`, `global-state`, `multi-node`, and `retry-queue` Failsafe lanes with unchanged required check names
- Shared JUnit and Playwright classifiers plus always-uploaded reports, environment data, Maven logs, dumps, and bounded container logs
- Authoritative hard-delete markers and compare-and-delete invalidation across local/Redis cache layers
- Conditional context processing, non-overlapping search fallback, durable per-entity relationship revisions for logical test suites, and exact-state polling
- Transactional per-TestCase and per-logical-TestSuite revisions reject late search publications without database advisory locks
- No blanket test reruns; shadow telemetry before any zero-retry enforcement

#
#### Use cases covered

- Concurrent reads cannot repopulate entities after hard delete
- Deleted context files reject late background updates
- Timed-out bulk indexing closes before safe individual fallback
- Concurrent logical-suite membership updates converge in both TestCase and TestSuite search documents
- Edited test cases publish fenced logical-suite relationship state instead of stale incremental changes
- Java lanes fail closed for assertions, setup failures, missing reports, cancellation, and hangs
- Playwright shards and Python suites preserve their real test outcome and diagnostics
- Glossary term navigation waits for the populated tree state before deciding whether to expand
- Oracle lineage, Trino rows, and Lineage UI entities wait for the exact published state

#### Unit tests

- [x] Added and updated tests for cache markers, Elasticsearch/OpenSearch bulk sinks, context extraction, entity restore, search updates, and both outcome classifiers
- Classifier tests: 21 passed
- Focused Java tests: 258 passed
- Latest `TestCaseRepositoryTest` and `SearchRepositoryTest`: 22 passed
[...]

PR #30267, opened 2026-07-21
fix(ci): install the ANTLR CLI from the distro archive when available

ShaileshParmar11:
## Problem

`make install_antlr_cli` downloads the ANTLR 4.9.2 CLI from a public host at build time, and that download keeps getting rate-limited.

#30120 moved it from `www.antlr.org` (GitHub Pages) to Maven Central and added SHA256 verification. Central then rate-limited us from the same shared CI egress IP. Nightly run [29776192824](https://github.com/open-metadata/openmetadata-nightly/actions/runs/29776192824) started ~7h after that merge and failed with 12 × HTTP 429 from `repo1.maven.org`. Runs are still failing today, and it also breaks collate PR checks.

Relocating the download between public hosts doesn't help — the throttling follows it.

## Why fix it here

Every consumer already calls this one target:

| Repo | Call sites | Covered by this PR? |
|---|---|---|
| OpenMetadata | 20 — 16 workflows + the 3 shared composite actions | ✅ all |
| openmetadata-collate | 3 Dockerfiles that run OpenMetadata's Makefile (`WORKDIR openmetadata/` or `cd ../OpenMetadata`) | ✅ via `git submodule update --init --remote`, which tracks OpenMetadata `main` |
| openmetadata-collate | root `Makefile` target + its caller `ui-checkstyle.yml` | ❌ **not covered — separate collate PR needed** |
| openmetadata-nightly | 7 | n/a — fixed directly in openmetadata-nightly#260 (merged) |

Fixing the target here covers the OpenMetadata call sites and the collate Docker builds, instead of editing ~25 call sites across two repositories.

### Scope gap (collate)

Collate has its **own** root `Makefile` with a duplicate `install_antlr_cli` that is still the original pre-#30120 version — it downloads from `antlr.org` and has no `curl --fail`, so a 429 response body gets appended into the jar and fails later with an unrelated error. Its only caller is `.github/workflows/ui-checkstyle.yml`.

Nothing in this PR touches that. It needs a separate collate PR, ideally deleting the duplicate so there is one implementation rather than two that drift. Thanks @ayush for catching that this PR's description previously overstated its collate coverage.

## Change

`install_antlr_cli` now resolves in three steps, cheapest first:

1. **Already on PATH** at the pinned version → nothing to do.
2. **Distro package**, when apt offers exactly `4.9.2`. Ubuntu noble and later ship 4.9.2 — matching every pinned runtime in this repo — so no public host is involved.
3. **Pinned, checksum-verified download**, unchanged.

The version pins step 2 has to satisfy:

| Pin | Location |
|---|---|
| `antlr4-python3-runtime==4.9.2` | `ingestion/setup.py:147` |
| `"antlr4": "4.9.2"` | `openmetadata-ui/.../package.json:100` |

The match is exact and checked with `apt-cache policy` *before* installing, so a distro carrying a different ANTLR (jammy ships 4.7.2) falls through to the pinned download rather than silently generating parsers those runtimes reject. apt failures — no root, unreachable archive — fall through too.

Step 3 keeps `ANTLR_MAVEN_BASE` so environments without a distro package (macOS, non-Debian images) can point at an internal Maven mirror instead of Central.

## Testing

| Scenario | Result |
|---|---|
| 4.9.2 already on PATH | skipped, nothing written |
| apt offers 4.9.2 (noble) | installed via apt, **no download** |
| apt offers 4.7.2 (jammy) | rejected, fell through to download, checksum `OK` |
| apt install fails (no root) | fell through to download, checksum `OK` |
| no apt at all (macOS) | download, checksum `OK` |
| 4.7.2 binary on PATH | did not skip, downloaded 4.9.2 |
| stale 4.7.2 shadowing apt's 4.9.2 on PATH | detected, reported the shadowing path, fell back to download |
| mirror unreachable | fell back to Central, checksum `OK` |
| all sources 404 | exit 2, no partial install |

The apt branches were exercised with stubbed `apt-get`/`apt-cache` (this was developed on macOS), so the branch logic is verified but **real apt behaviour on a noble runner is not**. Worth confirming on CI before merge.

Pinned hash verified against the published artifact: `bb117b1476691dc2915a318efd36f8957c0ad93447fb1dac01107eb15fe137cd`.

## Notes

- Version stays 4.9.2. Aligning with the Java side's `antlr.version` of 4.13.2 would need the Python and JS runtimes bumped and parsers regenerated — separate change.
- open-metadata/openmetadata-nightly#260 fixes the nightly workflows directly so AUT is unblocked without waiting for this.
- Credit to @matias for suggesting the distro package.

This PR makes ANTLR CLI installation prefer local and distro-provided copies before downloading. The main changes are:

- Reuses an existing ANTLR 4.9.2 executable on `PATH`.
- Installs the exact pinned version from apt when available.
- Verifies the resolved executable after apt installation.
- Retains checksum-verified downloads with configurable primary and fallback Maven sources.
</details>

This looks safe to merge.

- The apt path now verifies the executable selected by `PATH` before returning success.
- An incompatible shadowing executable causes the recipe to continue to the pinned download.
- No blocking issue remains in the updated code.

</details>

| Filename | Overview |
|----------|----------|
| Makefile | Adds exact-version PATH and apt resolution, post-install verification, and a configurable verified download fallback. |

</details>

<sub>Reviews (3): Last reviewed commit: ["fix(ci): portable checksum tool and anch..."](https://github.com/open-metadata/openmetadata/commit/259ad45269bea45909e2df51ffbcd59992a6e91c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45866084)</sub>

</details>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,10 +102,15 @@ install_antlr_cli:  ## Install antlr CLI locally
 		case "$$candidate" in \
 			$(ANTLR_VERSION)|$(ANTLR_VERSION)[-+~]*) \
 				if apt-get install -y -qq antlr4 > /dev/null 2>&1; then \
-					echo "Installed ANTLR $$candidate from the distro archive."; \
-					exit 0; \
+					if command -v antlr4 > /dev/null 2>&1 \
+						&& antlr4 2>&1 | grep -q "Version $(ANTLR_VERSION)"; then \
+						echo "Installed ANTLR $$candidate from the distro archive."; \
+						exit 0; \
+					fi; \
+					echo "Distro package installed, but antlr4 on PATH resolves to $$(command -v antlr4 2>/dev/null || echo none) which is not $(ANTLR_VERSION); falling back to pinned download." >&2; \
+				else \
+					echo "apt-get install antlr4 failed; falling back to pinned download." >&2; \
 				fi; \
-				echo "apt-get install antlr4 failed; falling back to pinned download." >&2; \
 				;; \
 			*) \
 				echo "Distro ANTLR candidate '$$candidate' is not $(ANTLR_VERSION); using pinned download." >&2; \
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ install_antlr_cli:  ## Install antlr CLI locally
 			candidate=$$(apt-cache policy antlr4 2>/dev/null | awk '/Candidate:/ {print $$2}'); \
 		fi; \
 		case "$$candidate" in \
-			$(ANTLR_VERSION)*) \
+			$(ANTLR_VERSION)|$(ANTLR_VERSION)[-+~]*) \
 				if apt-get install -y -qq antlr4 > /dev/null 2>&1; then \
 					echo "Installed ANTLR $$candidate from the distro archive."; \
 					exit 0; \
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,15 @@ generate:  ## Generate the pydantic models from the JSON Schemas to the ingestio
 
 .PHONY: install_antlr_cli
 ANTLR_VERSION := 4.9.2
-ANTLR_COMPLETE_JAR_URL := https://repo1.maven.org/maven2/org/antlr/antlr4/$(ANTLR_VERSION)/antlr4-$(ANTLR_VERSION)-complete.jar
+# Override ANTLR_MAVEN_BASE to pull from an internal Maven mirror (e.g. the
+# in-cluster Pulp proxy) instead of Maven Central, which rate-limits shared CI
+# egress IPs. The fallback base is tried if the primary is unreachable, so the
+# default (both = Central) stays correct on GitHub-hosted runners and locally.
+ANTLR_MAVEN_BASE ?= https://repo1.maven.org/maven2
+ANTLR_MAVEN_FALLBACK_BASE ?= https://repo1.maven.org/maven2
+ANTLR_COMPLETE_JAR_PATH := org/antlr/antlr4/$(ANTLR_VERSION)/antlr4-$(ANTLR_VERSION)-complete.jar
+ANTLR_COMPLETE_JAR_URL := $(ANTLR_MAVEN_BASE)/$(ANTLR_COMPLETE_JAR_PATH)
+ANTLR_COMPLETE_JAR_FALLBACK_URL := $(ANTLR_MAVEN_FALLBACK_BASE)/$(ANTLR_COMPLETE_JAR_PATH)
 ANTLR_COMPLETE_JAR_SHA256 := bb117b1476691dc2915a318efd36f8957c0ad93447fb1dac01107eb15fe137cd
 ANTLR_INSTALL_DIR ?= /usr/local/bin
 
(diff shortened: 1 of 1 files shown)


ShaileshParmar11:
> **Edge Case:** apt candidate glob `4.9.2*` could match unintended versions

Good catch — fixed in b7954e2. The glob is now anchored on a version delimiter:

```make
case "$candidate" in \
    $(ANTLR_VERSION)|$(ANTLR_VERSION)[-+~]*) \
```

Verified against the version formats apt actually produces:

| Candidate | Result |
|---|---|
| `4.9.2` | accepted → apt |
| `4.9.2-2` | accepted → apt |
| `4.9.2-3build1` (what noble ships) | accepted → apt |
| `4.9.2~rc1` | accepted → apt |
| `4.9.2+dfsg-1` | accepted → apt |
| `4.9.20-1` (the reported case) | **rejected** → pinned download |
| `4.9.2preview` | **rejected** → pinned download |
| `4.7.2-5` (jammy) | **rejected** → pinned download |
| `4.10-1` | **rejected** → pinned download |

`~` is included in the delimiter set since a `4.9.2~rc` is still upstream 4.9.2. An epoch-prefixed candidate (`1:4.9.2-2`) would no longer match and would fall through to the pinned download — the safe direction.

Also worth noting for reviewers: the apt path is verified end-to-end on `ubuntu:24.04` arm64, matching our runners.

```
apt-cache policy antlr4  ->  Candidate: 4.9.2-2
apt-get install -y antlr4 ->  OK
command -v antlr4         ->  /usr/bin/antlr4
antlr4                    ->  ANTLR Parser Generator  Version 4.9.2
```

tutte (dismissed):

ShaileshParmar11:
Thanks @ayush — both code points were valid. Addressed:

**1. Collate scope — you're right, the description overstated it.** Corrected above. The accurate split:

| collate call site | Covered here? |
|---|---|
| `collate-local-webserver/docker/Dockerfile-dev:90` (`WORKDIR openmetadata/`) | ✅ runs OpenMetadata's Makefile |
| `ingestion-extension/operators/docker/Dockerfile-dev:109` (`WORKDIR openmetadata/`) | ✅ same |
| `collate-local-webserver/docker/Dockerfile:45` (`cd ../OpenMetadata &&`) | ✅ same |
| collate root `Makefile` + `ui-checkstyle.yml` | ❌ **needs a separate collate PR** |

The Docker ones reach this fix because collate CI runs `git submodule update --init --remote` and `.gitmodules` on collate main tracks `branch = main` — so they pick up OpenMetadata main's tip without a submodule bump. The `ui-checkstyle` path calls collate's own duplicate target, which this PR cannot touch.

**2. Verify the resolved binary after `apt-get install` — real bug, fixed in edb6d75.**

You were right that it was unprotected. apt installs to `/usr/bin/antlr4`, but `/usr/local/bin` comes earlier on PATH and is exactly where the *previous* install method wrote. A machine carrying a stale binary there would have had `apt-get install` succeed, the target exit 0, and `antlr4` still resolve to the old version — silently generating parsers the pinned runtimes reject.

It now re-checks the resolved binary and falls through to the pinned download (which writes to `ANTLR_INSTALL_DIR` and takes precedence):

```
Distro package installed, but antlr4 on PATH resolves to /.../shadow-early/antlr4
which is not 4.9.2; falling back to pinned download.
```

Verified by putting a stale 4.7.2 ahead of apt's 4.9.2 on PATH — it detects the shadowing, names the offending path, and downloads. Happy paths re-checked for regressions.

**3. Metadata check** — noted, it wants a linked issue (`REQUIRE_LINKED_ISSUE`) or the `skip-pr-checks` bypass label. Deferring to the team on which is appropriate here.

**4. Confirming a fresh AUT run reaches sample-data ingestion** — agreed, that's the real signal. #260 is merged so the next nightly is the first end-to-end proof; nothing so far is more than local plus one `ubuntu:24.04` arm64 container.

ayush-shah (dismissed):

ShaileshParmar11:
Both findings were valid — thanks. Fixed, though the first one needed a different fix than suggested.

## 1. `shasum` vs `sha256sum` — real bug, but `sha256sum` alone isn't portable either

The core claim is correct and I verified it in a container: **`shasum` is absent on `ubuntu:24.04`**. So the download fallback would fail its checksum step on exactly the runners we target, and report `download failed after 3 attempts` instead of naming the missing tool.

Worth noting the implication: **the download path introduced in OpenMetadata#30120 has never actually succeeded on Ubuntu CI.** The 429 always hit first and masked it.

But switching to `sha256sum` would break macOS, which is the documented download path. macOS ships `/sbin/sha256sum` from BSD, which has **no `--check`**:

```
$ /sbin/sha256sum -c
usage: sha256sum [-bctwz] [files ...]
```

Measured on both platforms:

| | `shasum -a 256 --check` | `sha256sum --check` |
|---|---|---|
| macOS | ✅ works | ❌ fails (BSD, no `--check`) |
| ubuntu:24.04 | ❌ missing | ✅ works |

Neither name is safe to select by existence alone. So: **prefer `shasum`, fall back to `sha256sum`, and fail with a clear message if neither exists.**

Verified on ubuntu:24.04 arm64:
```
selected: sha256sum --check
/chk.txt: OK
verify: PASS
negative: correctly rejected
```

## 2. Unanchored version match — fixed

Correct on both counts: unanchored, and the unescaped dots were regex wildcards. Now:

```make
ANTLR_VERSION_RE := $(subst .,\.,$(ANTLR_VERSION))
... | grep -Eq 'Version $(ANTLR_VERSION_RE)([^0-9]|$$)'
```

Applied to both the early no-op check and the post-apt re-verification. Verified a stub reporting `Version 4.9.20` is now rejected and falls through to the pinned download, while exact `4.9.2` still short-circuits.

---

Ported here from the collate counterpart (open-metadata/openmetadata-collate#5176) since this file carries the identical logic.

PR #30349, opened 2026-07-22
ci(playwright): skip unused runner-side ingestion Python installs

chirag-madlani:
## Summary

- The composite action `.github/actions/setup-openmetadata-test-environment` installs `ingestion[all]`, `ingestion[test]`, `cx_Oracle`, `sqlalchemy-redshift`, `sqlalchemy-ibmi`, `pydoris-custom`, and `nox` onto the runner in its \"Install Python Dependencies\" step.
- Those packages are consumed by `py-tests-*` / `py-cli-e2e-tests` / `py-operator-*` workflows that invoke Python on the runner.
- **Playwright E2E workflows don't touch runner-side Python at all** — the ingestion Docker container runs its own installs. So the step is dead weight for those runs, costing ~3–4 minutes per shard.

## Change

- Add an `install-test-dependencies` input (default `\"true\"`, preserves current behavior for py-tests) and gate the \"Install Python Dependencies\" step on it.
- Flip all nine Playwright workflows using this composite to `install-test-dependencies: \"false\"`.

## Files changed

- `.github/actions/setup-openmetadata-test-environment/action.yml` (new input + `if:` gate)
- 9 playwright-*.yml workflows (one-line addition each)

Note: py-tests / py-cli-e2e / py-operator / postgresql-nightly-e2e / mysql-nightly-e2e workflows are **unchanged** and continue to install everything as before.

## Impact estimate

On the [postgresql-e2e run 29907894464](https://github.com/open-metadata/OpenMetadata/actions/runs/29907894464) each shard's `Setup Openmetadata Test Environment` step ran 12–14 min. This should shave ~3–4 min off that on every Playwright shard.

## Test plan

- [ ] `Postgresql PR Playwright E2E Tests` on this PR runs green; \"Setup Openmetadata Test Environment\" step is measurably shorter.
- [ ] `py-tests-postgres` / `py-cli-e2e-tests` on this PR still install everything (default `\"true\"` unchanged) and pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<details open><summary><h3>Greptile Summary</h3></summary>

This PR avoids installing unused runner-side Python test packages in Playwright workflows. The main changes are:

- Adds a default-enabled `install-test-dependencies` input to the shared setup action.
- Gates the broad ingestion test dependency installation with that input.
- Disables the installation in nine Playwright workflows.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

This looks safe to merge.

[...]

PR #30350, opened 2026-07-22
ci: cache ingestion Python venv in setup-openmetadata-test-environment

chirag-madlani:
## Summary

- Every job using the shared composite action `.github/actions/setup-openmetadata-test-environment` builds a fresh Python venv and runs `uv pip install ingestion[dev]` (plus, for py-tests workflows, `[all]`/`[test]` and extras). That's ~2–4 min per shard, every run.
- Cache the `env/` venv with `actions/cache@v4`, keyed on runner OS + python-version + `hashFiles` of `ingestion/setup.py`, `ingestion/pyproject.toml`, and the action file itself.

## How it works

- Cache is scoped to a manifest hash — any change to ingestion package deps or to the install-script code in the action invalidates it cleanly.
- `Generate Models` now only runs `python3 -m venv env` if `env/` isn't already restored.
- **The install steps are intentionally left in place.** `uv pip install` skips already-satisfied packages on cache hit, so they become fast no-ops. This keeps the change self-contained and safe for every consumer of this action regardless of which install steps they hit.

## Files changed

- `.github/actions/setup-openmetadata-test-environment/action.yml`
- No workflow file changes required.

## Impact estimate

On [postgresql-e2e run 29907894464](https://github.com/open-metadata/OpenMetadata/actions/runs/29907894464), the `Generate Models` + `Install Python Dependencies` steps combined for ~5–8 min per shard. Cache hit reduces both to ~10–30 s. That's ~2–4 min saved per shard × 7 shards = **~15–25 runner-minutes per run**. All Playwright / py-tests / py-cli-e2e workflows using this action benefit.

## Test plan

- [ ] First run of any workflow on this branch: cache miss (normal install path runs), \"Cache ingestion Python venv\" step reports `cache-hit=false` and saves at end of job.
- [ ] Subsequent run without dep-manifest changes: cache hit, `Generate Models`'s `uv pip install` becomes near-instant.
- [ ] py-tests-postgres on this PR still passes (venv fully populated on cache hit; installs remain functional on cache miss).

## Follow-ups (out of scope for this PR)

- Once #30349 lands (skip runner-side `Install Python Dependencies` for Playwright workflows), the cache benefit is realized on the biggest step. This PR helps both before and after that lands.

This PR caches the ingestion Python environment used by the shared test setup action. The main changes are:

- Cache `env/` using the runner OS, resolved Python version, and dependency manifests.
- Avoid fallback restoration of obsolete environments.
- Rebuild a restored environment when its Python interpreter cannot run.
- Keep dependency installation active after cache restoration.
</details>

The cache should be partitioned by dependency mode before merging.
- The resolved Python version prevents restoration across interpreter patch changes.
- Removing fallback keys prevents restoration after dependency manifests change.
- Jobs that install different dependency sets can still restore the same immutable cache, making results depend on which job populated it first.

**Files Needing Attention:** .github/actions/setup-openmetadata-test-environment/action.yml
</details>

| Filename | Overview |
|----------|----------|
| .github/actions/setup-openmetadata-test-environment/action.yml | Adds resolved-version venv caching and interpreter recovery, but different dependency-install modes still share one cache key. |

</details>

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into ci/cache-python..."](https://github.com/open-metadata/openmetadata/commit/f018a4889ad58984ff51212ce278446bfba3b7b3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46287559)</sub>


chirag-madlani:
@greptile-apps thanks — both P1s addressed in 629495f:

1. **Resolved interpreter** — added `id: setup-python` and switched the cache key to `steps.setup-python.outputs.python-version` (resolved patch). Also added a validation guard in `Generate Models` that rebuilds the venv if `env/bin/python --version` can't run, so a broken restore fails loud instead of dying on an opaque shebang error later.
2. **restore-keys fallback** — dropped entirely. Cache miss now guarantees a fresh install; no risk of importing a package the current manifest doesn't declare.

ulixius9 (dismissed):

chirag-madlani:
The shard-4 failures on [run 29997487494](https://github.com/open-metadata/OpenMetadata/actions/runs/29997487494/job/89234406200?pr=30350) are **not caused by the venv-cache changes** — they're pre-existing broken tests that were landed as skips on main after that CI run:

| File | Failures | Status on main |
|---|---|---|
| `Features/ServiceAgentsPauseResume.spec.ts` | 3 | **Skipped in [#30408](https://github.com/open-metadata/OpenMetadata/pull/30408)** — `0a4a79f7b6 Skip the ServiceAgentsPauseResume` |
| `Features/OntologyExplorerRdf.spec.ts` | 6 | Same failure class — RDF endpoint / graph-render (not covered by #30408 but likely same intermittent) |

Timeline:

```
Failing CI run:              2026-07-23 09:58 UTC
Skip commit 0a4a79f7b6:       2026-07-23 16:15 UTC   ← landed AFTER the run
```

Verified: PR head `dfdca4fd95…` did NOT contain the skip; `origin/main` currently does. Rebased on latest main → new head `3874d6cfa7` now includes the skip. Force-pushed just now.

The venv-cache logic itself only affects Python env setup on the runner; it can't influence UI/backend test behavior. Fresh CI on the rebased branch should be clean.

chirag-madlani:
Rebased on latest main (which now includes #30310) → head `79ec02aa14`.

**Verified: #30310 did NOT add venv caching.** Its `fast-mode` gates *skip* the venv steps entirely on fixture-restored shards, but every consumer that runs in normal mode (py-tests, py-cli-e2e, py-operator, and any Playwright path that hasn't opted into fast-mode) still pays the ~2–4 min of `uv pip install` on every run. Our cache still applies to all of those.

Conflict resolution — `action.yml` line 96-ish:
- Kept #30310's `if: ${{ inputs.fast-mode != 'true' && inputs.lightweight-ingestion != 'true' }}` gate
- Kept #30310's `actions/setup-python@v6` bump
- Kept our `id: setup-python` (needed for the cache key's resolved-patch reference)
- Added the same `if:` gate to the `Cache ingestion Python venv` step so fast-mode consumers skip both the setup and the cache — no wasted cache lookups on runs that will never use the venv.

mohityadav766 (approved):

PR #30348, opened 2026-07-22
fix(playwright): stop CustomProperties describe.serial retry cascade

chirag-madlani:
## Summary

- Overrides `retries: 0` on the `describe.serial` block that iterates `ALL_ENTITIES` in `Pages/CustomProperties.spec.ts` (~20 tests per entity type).
- Playwright reruns an entire `describe.serial` on any test failure; with the global `retries: 2`, a single flake cascades into ~3× block-level wall time. This was the single biggest source of retry waste on `postgresql-e2e`.

## Evidence

Numbers from two recent `Postgresql PR Playwright E2E Tests` runs:

| Run | CustomProperties tests with retries | Runs per test | Retry-time waste |
|---|---|---|---|
| [29881625038](https://github.com/open-metadata/OpenMetadata/actions/runs/29881625038) | ~20 | up to 4 (2 retries) | ~15+ minutes on shard 5 |
| [29907894464](https://github.com/open-metadata/OpenMetadata/actions/runs/29907894464) | 22 | 2 (1 retry each) | 478s on shard 5 |

Shard 5 is consistently the slowest (154m / 133m wall in the two runs above) because it draws the heaviest chromium spec files by file-hash sharding — the CustomProperties cascade dominates its retry cost.

## Why not remove `.serial()` instead?

Cross-test state shared via describe scope (`mainEntity`, `dashboardTopic1/2`, `cpasTestData`, `propertyNames`, `users`) means removing `.serial()` would need per-test dependency proofs across ~3900 lines. Out of scope for this fix; can be revisited separately.

## Consequence

Flakes in this block will now surface as PR failures instead of 20+ minutes of silent retry work. That's the intent — force real flake fixes rather than paying the retry tax.

## Test plan

- [ ] `Postgresql PR Playwright E2E Tests` on this PR runs the `CustomProperties.spec.ts` block once (no retries) and either passes or surfaces a real flake to fix.
- [ ] Wall clock on shard 5 drops by the equivalent of the previous retry cost (~5-8 min minimum, larger when historical flakes were present).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<details open><summary><h3>Greptile Summary</h3></summary>

This PR reduces retry waste in the CustomProperties Playwright suite. The main changes are:

- Disables retries for each entity-specific serial test block.
- Uses plain property names for Dashboard and Pipeline search-settings persistence checks.
- Retains special-character coverage in the general custom-property CRUD tests.
</details>

[...]

PR #30390, opened 2026-07-23
fix(playwright): warm search index + cap slow-test timeouts on Roles/Entity flakes

chirag-madlani:
## Summary

Two ~180 s timeout hangs consistently surface on postgresql-e2e:

- \`Pages/Roles.spec.ts › Roles page should work properly\` (shard 1)
- \`Pages/Entity.spec.ts › User as Owner with unsorted list\` (heaviest chromium shards)

Both declare \`test.slow()\` (180 s ceiling) and both **pass in ~30–40 s on retry**. All the waste is in how long the first attempt hangs before Playwright's per-test retry kicks in — 180 s / attempt in the worst case.

## Root cause per test

**Entity › User as Owner with unsorted list**
The test creates two fresh users via API, then immediately drives the UI owner picker to search for them. Under CI load the async \`user_search_index\` lags creation by several seconds. The UI search returns empty and each \`ownerItem.waitFor({state:'visible'})\` inside \`addMultiOwner\` hangs 30 s. Two adds + a remove chain waste up to 90 s from one cold index — enough to push the whole test over the 180 s ceiling.

**Roles › Roles page should work properly**
The test runs 8 sequential \`test.step\` blocks, each renavigating to the roles list and calling \`getElementWithPagination\` (a 50-page loop with a 30 s loader wait per page). Under load the cumulative delays exceed 180 s.

## Change

**Entity.spec.ts** — root-cause fix + insurance cap
- Poll \`/api/v1/search/query?q=<username>&index=user_search_index\` after both \`OWNER1.create()\` / \`OWNER2.create()\` return, waiting up to 15 s (500 ms → 2 s intervals) until both users appear. Subsequent UI searches now find them on the first click.
- Cap \`test.setTimeout(120_000)\` instead of \`test.slow()\` so if any other hang mode remains, we fail at 2 min and let the retry recover with a fresh 120 s budget.

**Roles.spec.ts** — fail-fast cap only
- Refactoring the 8 \`test.step\` blocks is out of scope for a flakiness PR. Cap \`test.setTimeout(120_000)\` (2× happy-path runtime) so a hung attempt fails at 2 min instead of 3.

## Expected impact

Per flake occurrence:

|                              | Attempt 1 hang | Attempt 2 (typical) | Total wall time | Savings |
|-----------------------------|----------------|---------------------|-----------------|---------|
| Before (with \`test.slow()\`)  | 180 s          | 30–40 s             | 210–220 s       | —       |
| After (Entity, warmup works) | 0 s (passes on 1st try) | —          | 30–40 s         | ~180 s  |
| After (fallback cap)         | 120 s          | 30–40 s             | 150–160 s       | ~60 s   |

## Test plan

- [ ] \`Postgresql PR Playwright E2E Tests\` on this PR: both tests pass, ideally on first attempt (Entity via warmup).
- [ ] If Entity's warmup ever fails (search index never lands), it fails at 15 s inside \`expect.poll\` with a clear message rather than the opaque \`Test timeout exceeded\`.

## Related

- [#30348](https://github.com/open-metadata/OpenMetadata/pull/30348) CustomProperties retry cascade (same effort, different flake pattern).

The PR reduces Playwright flake duration by warming the user search index before owner selection, imposing two-minute test ceilings, and replacing repeated Roles-list pagination with deterministic detail-page navigation.
</details>

The PR appears safe to merge.

No blocking failures remain.

</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Entity.spec.ts | Adds search-index readiness polling for newly created owners and replaces the slow-test multiplier with a 120-second timeout. |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Roles.spec.ts | Caps test duration and navigates directly to the generated role detail URL for repeated role operations. |

</details>

<sub>Reviews (4): Last reviewed commit: ["fix(playwright): poll user search index ..."](https://github.com/open-metadata/openmetadata/commit/c24ffcdce0a8d60c5708bdc40fba3b2732c7dc70) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46633218)</sub>

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Entity.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Entity.spec.ts
@@ -244,7 +244,12 @@ Object.entries(entities).forEach(([key, EntityClass]) => {
      * and verifying the owner list maintains proper state
      */
     test('User as Owner with unsorted list', async ({ page }) => {
-      test.slow(true);
+      // Cap at 120s instead of test.slow()'s 180s — see rationale on the
+      // Roles spec: hitting the slow ceiling on a hung wait burns 3
+      // minutes before retry kicks in. The warmup below eliminates the
+      // main hang source (search-index freshness), but keep a tighter
+      // ceiling as insurance.
+      test.setTimeout(120_000);
 
       const { afterAction, apiContext } = await getApiContext(page);
       const owner1Data = generateRandomUsername('PW_A_');
@@ -253,6 +258,34 @@ Object.entries(entities).forEach(([key, EntityClass]) => {
       const OWNER2 = new UserClass(owner2Data);
       await OWNER1.create(apiContext);
       await OWNER2.create(apiContext);
+
+      // Wait for the freshly created users to land in user_search_index
+      // before we open the owner picker. Under CI load the async indexer
+      // can lag creation by several seconds; when addMultiOwner's search
+      // returns empty each ownerItem.waitFor({visible}) hangs the default
+      // 30s. Two adds + a remove = up to 90s of waste from one cold index.
+      // Poll the search API up-front so the UI dropdown finds them on
+      // the first search.
+      await expect
+        .poll(
+          async () => {
+            const [r1, r2] = await Promise.all([
+              apiContext.get(
+                `/api/v1/search/query?q=${OWNER1.getUserName()}&index=user_search_index`
+              ),
+              apiContext.get(
+                `/api/v1/search/query?q=${OWNER2.getUserName()}&index=user_search_index`
+              ),
+            ]);
+            const [d1, d2] = await Promise.all([r1.json(), r2.json()]);
+            return (
+              (d1.hits?.total?.value ?? 0) > 0 &&
+              (d2.hits?.total?.value ?? 0) > 0
+            );
+          },
+          { timeout: 15_000, intervals: [500, 1000, 2000] }
+        )
+        .toBeTruthy();
 
       await addMultiOwner({
         page,
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Entity.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Entity.spec.ts
@@ -266,15 +266,19 @@ Object.entries(entities).forEach(([key, EntityClass]) => {
       // 30s. Two adds + a remove = up to 90s of waste from one cold index.
       // Poll the search API up-front so the UI dropdown finds them on
       // the first search.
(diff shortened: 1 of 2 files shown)


chirag-madlani:
Amended with [2343d80527](https://github.com/open-metadata/OpenMetadata/pull/30390/commits/2343d80527) per @madlanichirag's suggestion — direct-nav to the role detail page instead of paginating.

Four test.step blocks in `Roles page should work properly` were doing the same thing: navigate to roles list → paginate to find the row → click. The role URL is deterministic from `roleName`, so each hop becomes a single `page.goto(`/settings/access/roles/${roleName}`)`.

- **Steps affected**: 3 (edit description), 5 (add policy), 6 (remove policy), 7 (check last policy).
- **Steps left with pagination**: 1 (verify role appears in list — the point of that assertion is the list view) and 8 (delete via list row's `delete-action-<name>` button).
- **Net**: −37 lines, no new helpers, root cause of the pagination-loop-in-hang mode removed for the four affected steps.

Combined with the earlier `test.setTimeout(120_000)` cap this test should now consistently pass well under the ceiling on cold shards.

ShaileshParmar11 (dismissed):

chirag-madlani:
Rebased on latest main (which now includes #30310) → head `c24c647942`. Clean rebase, no conflicts.

- `Roles.spec.ts` + `utils/roles.ts` were **untouched** by #30310 — our direct-nav fix (steps 3/5/6/7) and the 120s timeout cap on `Roles page should work properly` apply as-is.
- `Entity.spec.ts` got small structural changes from #30310 (`performAdminLogin` → `createAdminApiContext` in beforeAll/afterAll, `{ storageState: undefined }` on `browser.newPage()` calls) but those are around lines 94–140. Our edits at line ~244 (`User as Owner with unsorted list` warmup + timeout cap) are unaffected — `test.slow(true)` is still on main so the flake mode this PR fixes is still there.

chirag-madlani:
Addressed @gitar-bot's review in [c24ffcdce0](https://github.com/open-metadata/OpenMetadata/pull/30390/commits/c24ffcdce0).

Confirmed the mapping (bot had the names swapped but the concern was real):
- `UserClass.getUserName()` returns `responseData.name`
- `UserClass.getUserDisplayName()` returns `responseData.displayName ?? responseData.name`

At line 292 and downstream (301, 311, 319) `addMultiOwner` / `removeOwnersFromList` / `removeOwner` all drive the picker with `getUserDisplayName()`. My warmup was polling `getUserName()` — coincidentally identical today for users generated via `generateRandomUsername` (no displayName set → falls back to name), but future divergence would let the warmup pass while the UI search still misses. Switched both poll calls to `getUserDisplayName()` so they cover the exact search term the picker uses.

PR #30148, opened 2026-07-16
Fixes #30147: pipeline sources own one BaseConnection to avoid a second sign-in

IceS2:
Fixes #30147

`PipelineServiceSource` built its client through the `get_connection` seam, which discards the `BaseConnection` owner, and then ran its test step through `test_connection_common` — which built a **second** connection, i.e. a second sign-in. Same root cause already fixed for dashboard (#29870) and database (#30094).

The base now owns one `BaseConnection` and reuses it:

- `__init__` builds the owner with `create_connection` and takes its client; `test_connection()` goes through `run_test_connection`, reusing that owner. Unmigrated/custom connectors keep the legacy `test_connection_common` path.
- `close()` disposes the owner. If the in-`__init__` test fails, the owned connection is released directly rather than via `self.close()` — `close()` is successful-run teardown (it runs `compute_percentile`) and is polymorphic, so subclass overrides would touch attributes their `__init__` had not set yet and mask the original error.
- `connection_obj`, `self.connection` and `self.client` still point at the owner's live client: pipeline subclasses read `self.connection`, and Collate reads `connection_obj`.

Each connector's redundant module-level `get_connection` is inlined into its `_get_client` and deleted (airbyte, airflow, databrickspipeline, fivetran, flink, openlineage). `databrickspipeline.get_connection_url` is kept — it has external importers. Airflow keeps its `@check` provider, `Engine` dispose registration and singledispatch builder from #30089.

Fixes a latent bug surfaced while inlining: the databricks pipeline connector subscripted `connectionArguments.root` without guarding against `None`.

`self.test_connection()` stays a visible polymorphic call in `__init__` — Collate subclasses (`SnowplowSource`) override it to run a different check.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR makes pipeline sources own and reuse one connection throughout a run. The main changes are:

- Creates pipeline clients through an owned `BaseConnection`.
- Reuses the owned connection for connection tests.
- Closes partially built clients when construction fails.
- Registers connector resources for teardown.
- Keeps the legacy path for custom and unmigrated connectors.
- Inlines connector-specific client factories.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

This looks safe to merge.

- Client construction failures now release registered resources.
- The owned connection is reused by the connection test.
- No additional blocking issue was found in the updated code.

</details>

<details open><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
[...]

PR #30402, opened 2026-07-23
Fixes #30399: Fix connector teardown regressions from the source-owns-BaseConnection refactor

### LINKED ISSUE #30399: Fix connector teardown regressions from the source-owns-BaseConnection refactor
The source-owns-BaseConnection refactor (#29870 / #30094 / #30148 / #30133 / #30174) moved connection lifecycle onto the `BaseConnection` owner and replaced the in-`__init__` `self.close()` on a failed test-connection with `close_on_failure(self._connection)`. `close_on_failure` disposes only the owner (it unwinds the owner's `_on_close` teardowns); it never runs a source `close()` override. That surfaced four follow-up regressions:

1. **Init-failure cleanup skipped (MicroStrategy, Elasticsearch/OpenSearch).** When `test_connection` raised in `__init__`, connector teardown that lived in a source `close()` override was lost: MicroStrategy leaked its server-side API session; Elasticsearch/OpenSearch left the client private key on disk in the SSL staging dir. Fix: move the teardown onto the owner by registering it with `_on_close` inside `_get_client`, so it runs on both the normal close and the `close_on_failure` path; drop the now-redundant source overrides.
2. **Dashboard `close()` overrides skip `super().close()`** (tableau, looker, powerbi) → the owned `BaseConnection` was never disposed on the normal path.
3. **Airflow 2.x disposes its process-global engine.** The `BackendConnection` engine is `airflow.settings.engine` (borrowed, not OpenMetadata-built); the owner registered `engine.dispose` unconditionally for any `Engine`, tearing down the host Airflow connection pool on close and on every Automation test-connection. Fix: only dispose engines this connection built.
4. **Dashboard base dropped `self.connection_obj`** — the only service base that stopped setting it, breaking the kept, deprecated seam for out-of-tree custom dashboard connectors.

Tracked PR: #30402

IceS2:
Fixes #30399

Four follow-up regressions from the source-owns-BaseConnection refactor (#29870 / #30094 / #30148 / #30133 / #30174).

**Key finding:** `close_on_failure(self._connection)` in each base `__init__` disposes only the owner (it unwinds the owner's `_on_close` teardowns); it never runs a source `close()` override. So any connector teardown that lived in a `close()` override was skipped when `test_connection` raised in `__init__`. The fix pattern is to move connector teardown onto the owner — register it with `_on_close` inside `_get_client` — so it fires on both the normal `super().close()` and the `close_on_failure` path, then drop the redundant override.

| # | Regression | Fix |
|---|---|---|
| 1 | Init-failure cleanup skipped: MicroStrategy leaked its server-side API session; Elasticsearch/OpenSearch left the client private key on disk in the SSL staging dir | register `close_api_session` / staging-dir removal on the owner in `_get_client`; delete the source `close()` overrides |
| 2 | tableau/looker/powerbi `close()` overrides never chained `super().close()` → the owned `BaseConnection` was never disposed on the normal path | chain `super().close()`; tableau drops its direct `sign_out()` (the owner already registers it via `_on_close`, so keeping both signs out twice); powerbi moves `file_client.delete_tmp_files` onto the owner |
| 3 | Airflow 2.x disposed `airflow.settings.engine` — the borrowed process-global engine returned by `BackendConnection` — on close and on every Automation test-connection, tearing down the host Airflow pool | only register `dispose` for engines this connection built (`isinstance(client, Engine) and client is not settings.engine`) |
| 4 | The dashboard base was the only service base that stopped setting `self.connection_obj`, breaking the kept, deprecated seam for out-of-tree custom dashboard connectors | set `connection_obj` and read it in the legacy `test_connection` fallback (same object, behavior unchanged) |

One commit per fix (splittable).

Fixes connector teardown ownership regressions across dashboard, search, and Airflow integrations.
- Moves connector-specific cleanup callbacks onto their owning `BaseConnection`.
- Restores base dashboard connection disposal and the legacy `connection_obj` seam.
- Avoids disposing Airflow’s borrowed process-global engine.
- Adds regression coverage for normal close and initialization-failure cleanup paths.
</details>

The shared TLS staging-directory cleanup must be isolated per connection before this PR is safe to merge.

Closing one search connection recursively deletes the default staging directory used by other live connections, removing their TLS files and breaking later TLS setup or reconnects.

**Files Needing Attention:** ingestion/src/metadata/ingestion/source/search/elasticsearch/connection.py and ingestion/src/metadata/ingestion/source/search/opensearch/connection.py
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/source/search/elasticsearch/connection.py | Registers TLS staging cleanup before SSL-context construction, but the callback still recursively deletes a staging directory shared by other live connections. |
| ingestion/src/metadata/ingestion/source/search/opensearch/connection.py | Transfers by-value TLS staging cleanup to the connection owner while retaining the same shared-directory ownership problem. |
| ingestion/src/metadata/ingestion/source/dashboard/dashboard_service.py | Restores the deprecated connection-object compatibility seam while retaining owned connection lifecycle handling. |
| ingestion/src/metadata/ingestion/source/pipeline/airflow/connection.py | Limits engine disposal callbacks to connection-owned engines rather than Airflow’s process-global engine. |
| ingestion/tests/unit/topology/search/test_search_owned_connection.py | Covers normal and failed-initialization cleanup but does not cover two live connections sharing the default TLS staging directory. |

</details>

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/source-owns..."](https://github.com/open-metadata/openmetadata/commit/a9d874fd296e7e015bbfa55c7a014141e535317c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46646325)</sub>

**Context used:**


IceS2:
Thanks — addressing both findings:

**greptile P1 — SSL failure leaves staged secrets:** Valid, fixed in 15e7046. The `_on_close` cleanup was registered *after* `get_ssl_context` (Elasticsearch) / `_handle_ssl_context_by_value` (OpenSearch), which write the cert/key files and can raise while parsing them — so a failed SSL build left the staged private key on disk, since `_build_client` only unwinds teardowns registered *before* the failure point. Now registered as the staging material is acquired, in both connectors. Added a regression test (`test_elasticsearch_removes_staging_dir_when_ssl_context_build_fails`, red before the reorder).

**gitar — cleanup may delete a shared cert dir:** Real observation, but pre-existing and out of scope for this regression PR. The prior source `close()` already did `shutil.rmtree(...stagingDir)` on the whole shared default dir (`/tmp/openmetadata-certs`); this PR relocates that teardown onto the owner, it doesn't introduce the shared-dir behavior. Two by-value connections on the default dir also already collide on *write* (identical `root.pem`/`client.pem`/`client_key.pem` filenames), so the shared-dir design is broken independently of deletion. The robust fix — per-connection unique subdirs in the shared SSL-staging helpers — is a separate change; I'll open a follow-up rather than widen this PR.

pmbrull (approved):

PR #30468, opened 2026-07-24
fix(playwright): green playwright-summary when gate skips the pipeline

chirag-madlani:
## Summary

Regression from #30453 that surfaced on PR #30454 ([run 30090391086](https://github.com/open-metadata/OpenMetadata/actions/runs/30090391086/job/89472165978)):

- \`pull_request_target\` fires on a same-repo PR
- \`gate\` correctly returns \`should_run=false\` (redundant — \`pull_request\` already handles it)
- Every upstream job is legitimately \`skipped\`
- **But** \`playwright-summary\` runs the render step unconditionally (\`if: always()\`)
- \`render_playwright_summary.cjs\` counts each skipped upstream as a \"CI/reporting failure\" and calls \`core.setFailed()\`
- The required \`playwright-summary\` check goes 🔴 with \`0 Playwright test failure(s); 9 CI/reporting failure(s)\`
- Branch protection blocks the PR, even though the sibling \`pull_request\` run passed

## Fix

Three-line change on the \`playwright-summary\` job:

1. Add \`gate\` to its \`needs:\` list (so it can read \`needs.gate.outputs.should_run\`)
2. Gate the render step on \`should_run == 'true'\` — skips when gate said skip
3. Add a tiny \"Report gate-skipped run as green\" step with the opposite condition that echoes success — keeps the job's conclusion at ✅ so branch protection sees green regardless of which event fired

## Behaviour matrix after this fix

| Event | gate | render | job conclusion |
|---|---|---|---|
| merge_group / workflow_dispatch / schedule | true | run | real result |
| same-repo PR + pull_request | true | run | real result |
| same-repo PR + pull_request_target | false | skipped | ✅ green (this fix) |
| fork PR + pull_request_target + label | true | run | real result |
| fork PR + pull_request_target, no label | false | skipped | ✅ green (this fix) |
| fork PR + pull_request | false | skipped | ✅ green (this fix) |
| any draft PR | false | skipped | ✅ green (this fix) |

## Diff scope

\`\`\`
 .github/workflows/playwright-postgresql-e2e.yml | 19 ++++++++++++++++++-
 1 file changed, 18 insertions(+), 1 deletion(-)
\`\`\`

Pure workflow change. No script edits, no test edits, no logic changes to the render pathway.

## Test plan

[...]

PR #30470, opened 2026-07-24
fix(playwright): pin merge-multiple=false on results download

chirag-madlani:
## Summary

Follow-up to #30468. Single-shard PR runs (spec-only mode, one chromium lane) fail \`playwright-summary\` with:

\`\`\`
Shard chromium-01 did not upload a usable Playwright results artifact.
Shard chromium-01 did not upload its execution status.
\`\`\`

Seen on [PR #30437 run 30088248354](https://github.com/open-metadata/OpenMetadata/actions/runs/30088248354/job/89474166397?pr=30437).

## Root cause

Verified against the run's artifacts API — \`playwright-results-json-chromium-01\` **was** uploaded correctly, archive contents:

\`\`\`
results.json      (441110 bytes)
ci-status.json    (470 bytes)
\`\`\`

The download step reports \`success\`, but \`render_playwright_summary.cjs\` expects the on-disk layout:

\`\`\`
results/playwright-results-json-<shardId>/results.json
\`\`\`

That layout is what \`pattern: playwright-results-json-*\` produces **when multiple artifacts match**. When only one artifact matches (single-shard runs), some \`download-artifact@v7\` configurations flatten the contents directly into \`path:\` (\`results/results.json\`) — breaking the \`readdirSync\`-based shard discovery in the renderer.

Multi-shard runs never hit this because there's always >1 artifact matched — which is why we only see this on impact-selected / spec-only PRs like #30437.

## Change

One-line addition to the \`Download all results JSON\` step:

\`\`\`yaml
merge-multiple: false
\`\`\`

Locks the per-artifact subdirectory layout regardless of match count, so single-shard runs render exactly the same way multi-shard runs do. Also added an inline comment explaining the failure mode + linking the reference run.

## Diff scope

\`\`\`
 .github/workflows/playwright-postgresql-e2e.yml | 11 +++++++++++
 1 file changed, 11 insertions(+)
\`\`\`

Pure workflow config. No script changes.

## Not in scope (deferred)

A more resilient fix would harden \`render_playwright_summary.cjs\` to handle either layout (loose files at \`results/\` root vs. per-shard subdirs). That's Option B from the earlier analysis — can be a follow-up if this pin turns out not to help on single-shard runs.

## Test plan

- [ ] This PR's own run: single-shard scenario should pass. If it doesn't, we know the pin isn't enough and need Option B.
- [ ] Multi-shard runs (merge queue, full-suite dispatch): unaffected — behaviour was already correct there.

Adds an explicit `merge-multiple: false` setting to preserve per-artifact Playwright result directories and logs the downloaded layout for diagnosing single-shard summary failures.
</details>

The PR appears safe to merge because no blocking failures eligible for this follow-up review remain.

No blocking failure remains.

</details>

| Filename | Overview |
|----------|----------|
| .github/workflows/playwright-postgresql-e2e.yml | Pins the Playwright artifact download layout and adds temporary diagnostics for downloaded files and expected shard IDs. |

</details>

<sub>Reviews (2): Last reviewed commit: ["add diagnostic ls -R for downloaded resu..."](https://github.com/open-metadata/openmetadata/commit/9c79d8739183f991d43e1e4f0ed35c603c6629ab) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46991618)</sub>

### DIFF
--- a/.github/workflows/playwright-postgresql-e2e.yml
+++ b/.github/workflows/playwright-postgresql-e2e.yml
@@ -1393,6 +1393,25 @@ jobs:
           # See run 30088248354 for the failure this addresses.
           merge-multiple: false
 
+      # Diagnostic: log the actual on-disk layout the renderer sees. The
+      # explicit merge-multiple pin above matches the documented default,
+      # so if single-shard runs still fail the same way we need this
+      # trace to see whether download-artifact placed files at
+      # `results/playwright-results-json-<shardId>/results.json` (expected)
+      # or somewhere else. Remove once single-shard runs are consistently
+      # green.
+      - name: Diagnose downloaded results layout
+        if: ${{ always() && needs.playwright-ci-postgresql.result != 'skipped' }}
+        continue-on-error: true
+        run: |
+          echo "=== results/ tree ==="
+          ls -laR results 2>&1 || echo "(results/ does not exist)"
+          echo
+          echo "=== expected shardIds (from plan-playwright.matrix) ==="
+          echo '${{ needs.plan-playwright.outputs.matrix }}' \
+            | jq -r '.include[].shardId' \
+            || echo "(jq parse failed)"
+
       - name: Setup Node.js
         id: setup-node
         if: ${{ always() && needs.playwright-ci-postgresql.result != 'skipped' }}
--- a/.github/workflows/playwright-postgresql-e2e.yml
+++ b/.github/workflows/playwright-postgresql-e2e.yml
@@ -1381,6 +1381,17 @@ jobs:
         with:
           pattern: playwright-results-json-*
           path: results
+          # Pin the per-artifact subdirectory layout the renderer expects
+          # (results/playwright-results-json-<shardId>/results.json). When
+          # only one artifact matches the pattern — typical for spec-only
+          # single-shard PR runs — some download-artifact configurations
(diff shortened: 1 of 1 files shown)


chirag-madlani:
Pushed [9c79d87391](https://github.com/open-metadata/OpenMetadata/pull/30470/commits/9c79d87391) — added a diagnostic `ls -laR results/` step right after the download.

Reason: after re-reading the actions/download-artifact release notes, the `merge-multiple: false` pin in the previous commit matches what should already be the default for pattern-based downloads. If single-shard runs still fail the same way after this PR merges, my earlier flatten hypothesis was wrong. Rather than speculate more, the diagnostic step will print the actual on-disk layout the renderer sees so we can pick the right follow-up with data instead of guesses.

Comment on the step notes it's removable once single-shard runs are consistently green.

**Also — to answer @Chirag Madlani's question about mixing upload-artifact@v6 with download-artifact@v7**: no compatibility issue there. Both are on the same post-v4 storage backend (v4 introduced the immutable-artifacts model); download@v7 handles v6-uploaded zips identically. The v6→v7 upload change was purely additive (an `archive: false` opt-in for skipping the zip on single-file uploads) — v6-uploaded zips have identical structure to what v7 would produce.

pmbrull (approved):

PR #30354, opened 2026-07-22
chore(ci): remove workflows which are not in use

chirag-madlani:
### Describe your changes:

Fixes #<issue-number>

Clean up unused workflows 

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### High-level design:

N/A — small change. 

#
#### Manual testing performed

#
### UI screen recording / screenshots:

Not applicable. 

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] My PR is linked to a GitHub issue via `Fixes #<issue-number>` above.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.
- [ ] For UI changes: I attached a screen recording and/or screenshots above.
- [ ] I have added tests (unit / integration / Playwright as applicable) and listed them above.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR removes unused GitHub Actions workflows. The main changes are:

- Remove MySQL and PostgreSQL ingestion E2E workflows.
- Remove the MySQL UI Playwright workflow.
- Remove the SSO Playwright workflow.
- Remove the Playwright documentation check workflow.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

No additional blocking issue was found in this follow-up.
- The remaining candidates repeat the previously reported loss of test coverage.
- No separate workflow caller or required-check failure was established.
</details>

<details open><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| .github/workflows/playwright-integration-tests-mysql.yml | Removes PR and merge-queue execution of ingestion E2E tests against MySQL. |
| .github/workflows/playwright-integration-tests-postgres.yml | Removes PR and merge-queue execution of ingestion E2E tests against PostgreSQL. |
| .github/workflows/playwright-mysql-e2e.yml | Removes the MySQL-backed UI Playwright workflow. |
| .github/workflows/playwright-sso-tests.yml | Removes PR-time mock-OIDC tests and manually dispatched provider tests. |

</details>

[...]

PR #30523, opened 2026-07-27
fix(playwright): renderer fallback for flat single-shard download layout

chirag-madlani:
## Summary

Follow-up to #30470. The \`merge-multiple: false\` pin turned out to be a placebo — \`actions/download-artifact@v7\` STILL flattens a single-artifact \`pattern:\` match directly into \`path:\` even with the flag explicit.

The diagnostic step I added in #30470 finally proved it on [PR #30354 run 30261950909](https://github.com/open-metadata/OpenMetadata/actions/runs/30261950909/job/89975437662?pr=30354):

\`\`\`
=== results/ tree ===
results:
-rw-r--r--  ... 470   ci-status.json
-rw-r--r--  ... 1.3M  results.json

=== expected shardIds (from plan-playwright.matrix) ===
chromium-01
\`\`\`

Renderer expects \`results/playwright-results-json-<shardId>/results.json\` and its \`readdirSync\` loop walks \`results.json\`/\`ci-status.json\` as if they were shard subdirectories, finds no inner file, and reports the shard as \"did not upload a usable Playwright results artifact\". Impact-selected / spec-only PRs (that plan a single shard) are the only ones affected; multi-shard runs escape it because ≥2 matched artifacts keep download-artifact in per-subdirectory mode.

## Change

Add a **workflow-level normalize step** right after the diagnostic step:

\`\`\`yaml
- name: Normalize single-shard download layout
  if: ${{ always() && needs.playwright-ci-postgresql.result != 'skipped' }}
  env:
    MATRIX: ${{ needs.plan-playwright.outputs.matrix }}
  run: |
    set -euo pipefail
    if [[ ! -f results/results.json ]]; then
      echo \"Multi-shard layout detected; nothing to normalize.\"
      exit 0
    fi

    shard_ids=$(printf '%s' \"$MATRIX\" | jq -r '.include[].shardId')
    shard_count=$(printf '%s\\n' \"$shard_ids\" | grep -c '^' || true)
    if [[ \"$shard_count\" -ne 1 ]]; then
      echo \"::warning::Detected flat results/ layout but plan matrix has ${shard_count} shards — refusing to guess an artifact name.\"
      exit 0
    fi

    shard_id=$(printf '%s' \"$shard_ids\" | head -n 1)
    target=\"results/playwright-results-json-${shard_id}\"
    mkdir -p \"$target\"
    mv results/results.json \"$target/\"
    mv results/ci-status.json \"$target/\"
    echo \"Moved flat results/ files to ${target}/\"
    ls -laR results
\`\`\`

## Behaviour

| Scenario | Detection | Action |
|---|---|---|
| Multi-shard run (2+ shards planned) | `results/results.json` absent (files are already in per-shard subdirs) | no-op |
| Single-shard run, files flattened by v7 | `results/results.json` present + plan matrix has 1 shard | move into `results/playwright-results-json-<shardId>/` |
| Weird case: flat layout but plan has >1 shards | Refuses to guess | log warning, no-op (defensive — shouldn't happen) |

## Why workflow-level and not renderer-side

Two reasons:
1. Renderer stays untouched — no JS special cases to test.
2. The plan matrix is only easily accessible to workflow expressions; passing it into the JS via yet another env var is ugly.

## Test plan

- [ ] Rerun any spec-only PR that previously failed \`playwright-summary\` with \"did not upload a usable Playwright results artifact\" — the render step should now find the files.
- [ ] Multi-shard runs (merge queue, full-suite dispatch): step no-ops, no regression.
- [ ] The diagnostic \`ls -R results/\` step also runs after the normalize step to prove the layout is correct on both paths.

Updates the Playwright summary renderer to normalize flattened single-shard artifact downloads.
- Detects a top-level `results/results.json` when exactly one shard is expected.
- Moves available result and CI-status files into the renderer’s existing per-shard directory structure.
- Leaves normal multi-shard artifact layouts unchanged.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| .github/scripts/render_playwright_summary.cjs | Normalizes flattened single-shard downloads into the directory layout already consumed by the summary renderer. |

</details>

```mermaid
flowchart TD
  A[Downloaded Playwright results] --> B{Top-level results.json exists?}
  B -- No --> E[Process existing per-shard directories]
  B -- Yes --> C{Exactly one expected shard?}
  C -- No --> E
  C -- Yes --> D[Move files into expected shard directory]
  D --> E
  E --> F[Render Playwright summary]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(playwright): renderer-side fallback ..."](https://github.com/open-metadata/openmetadata/commit/75475198817ff2b4fddc2775413fa5f759a6c334) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47565723)</sub>

### DIFF
--- a/.github/scripts/render_playwright_summary.cjs
+++ b/.github/scripts/render_playwright_summary.cjs
@@ -210,6 +210,32 @@ async function renderPlaywrightSummary({ github, context, core }) {
 
   const shardResults = [];
   const resultsDir = 'results';
+
+  // actions/download-artifact@v5+ flattens the archive contents directly
+  // into `resultsDir` when the `pattern:` matches only ONE artifact —
+  // typical for impact-selected / spec-only PRs that plan a single
+  // chromium shard. The loops below assume the per-artifact subdirectory
+  // layout (`resultsDir/playwright-results-json-<shardId>/results.json`),
+  // so migrate the flat files into the expected subdirectory when we
+  // detect it. Multi-shard runs are untouched because they always land
+  // in the per-subdirectory layout.
+  if (
+    fs.existsSync(path.join(resultsDir, 'results.json')) &&
+    expectedShards.length === 1
+  ) {
+    const subDir = path.join(
+      resultsDir,
+      `playwright-results-json-${expectedShards[0]}`
+    );
+    fs.mkdirSync(subDir, { recursive: true });
+    for (const file of ['results.json', 'ci-status.json']) {
+      const from = path.join(resultsDir, file);
+      if (fs.existsSync(from)) {
+        fs.renameSync(from, path.join(subDir, file));
+      }
+    }
+  }
+
   if (fs.existsSync(resultsDir)) {
     for (const dir of fs.readdirSync(resultsDir).sort()) {
       const jsonPath = path.join(resultsDir, dir, 'results.json');


chirag-madlani:
Rewrote per @Chirag Madlani's feedback — same fix, no custom code.

Root cause per actions/download-artifact release notes:

> **v5.0.0** — BREAKING fix: inconsistent path behavior for single artifact downloads
> ...single artifact downloads behaved differently depending on how you specified the artifact...

v5+ flattens the archive contents directly into `path:` when `pattern:` matches only one artifact — even with `merge-multiple: false` explicit. v4 keeps per-artifact subdirectories regardless of match count.

Fix is just a **version pin on two steps**:

- `Download timing metrics` → `@v4`
- `Download all results JSON` → `@v4`

**Left alone:**
- `Download blob reports` — uses `merge-multiple: true` intentionally (`playwright merge-reports` wants a flat directory); its behaviour is unchanged across v4/v5/v6/v7.
- `Download shard plans` — uses `name:` (not `pattern:`), never affected.

Also removed:
- The placebo `merge-multiple: false` pin from #30470 (redundant with the v4 default; misleading with v7).
- The `Diagnose downloaded results layout` step from #30470 — its whole purpose was to confirm the layout, which it did on run 30261950909 (thanks!). No need to keep it as ongoing overhead now that we know the answer.

**Diff:** −13 lines net. Just the pin and a comment explaining why v4 is chosen.

chirag-madlani:
Rewrote once more per feedback — no version pin, stay on latest actions/download-artifact.

Fix moves to the renderer: 24 lines of JS at the top of the results-parsing section that detects the flat layout and renames the files into the expected subdirectory using `expectedShards[0]`. Both downstream loops (results.json and ci-status.json parse) then run uniformly regardless of the download tool's per-version behaviour.

**Diff:**
```
.github/scripts/render_playwright_summary.cjs | 26 ++++++++++++++++++++++++++
1 file changed, 26 insertions(+)
```

- **No workflow changes** (rolled back the v4 pin from the previous force-push).
- **Multi-shard runs:** `results/results.json` at top level doesn't exist → migration is a no-op.
- **Single-shard flat runs:** files land under `results/playwright-results-json-<shardId>/` before any parsing loop runs.
- **Defensive:** guarded on `expectedShards.length === 1` — never guesses a name if the plan reports a different shard count.

Why not the v4 pin: staying on latest keeps hash validation, Node runtime updates, and any future security patches without a manual version lock we'd have to remember to review if v4 gets EOL'd.

pmbrull (approved):

PR #20740, opened 2025-04-09
feat(ui): allow add and edit data products from entity page

pranita09:
Fix #20568

Worked on Add/Edit Data Product for each entity page.
- If there is domain user can add/edit data products with-in that domain
- If the table doesn't have domain, user will be asked to select a domain to add data products.
- Data products data should be visible inside explore page summary panel.
- Domain and Data Products are visible for table's lineage summary panel.

https://github.com/user-attachments/assets/11ee6e52-2aa7-402a-8cc7-5d9d6d52462f

Screenshot 2025-04-10 at 10 07 30 PM

Screenshot 2025-04-10 at 10 08 40 PM

#
### Type of change:

- [ ] Bug fix
- [x] Improvement
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

PR #30531, opened 2026-07-27
fix(dataContract): persist testSuite -> dataContract reverse relationship

### LINKED ISSUE #30297: Data Contract stays running forever
### Affected module

Data Quality / Profiler

### Describe the bug

Hi - we hit what looks like a regression in 1.13.0. After upgrading, every data contract that has qualityExpectations never completes validation. It sits at contractExecutionStatus: Running with qualityValidation: null indefinitely. Contracts stranded at the first scheduled run after the upgrade; the same contracts had been closing green daily on 1.12.3.

What works: the test suite runs, the test case result is written (we see genuine Failed results), the pipeline posts its status back, and the server even computes the right answer - the log shows TestSuiteRepository - Created ChangeEvent for TestSuite <contract-id> (passed: 0/1).

What breaks: that result is never written back to the contract. TestSuiteRepository.onTestSuiteExecutionComplete() does:

if (testSuite.getDataContract() != null) {
    dataContractRepository.updateContractDQResults(testSuite.getDataContract(), testSuite);
}

and testSuite.getDataContract() is always null, so the contract is silently skipped - no exception, no log.

Root cause (we think): in 1.13.0, store() was changed from nullifyEntityFields() to serializeForStorage() / storageJsonNode(), and the new FIELDS_STORED_AS_RELATIONSHIPS constant in EntityRepository.java (~line 4522) added "dataContract" to the strip list:

List.of("href", "owners", "children", "tags", "domains", "dataProducts",
        "dataContract",   // new in 1.13.0, was not nulled in 1.12.3
        "followers", "experts", "reviewers", "certification");

So dataContract is now stripped from every entity's stored JSON but TestSuiteRepository has no setFields/relationship handling to rebuild it on read, and nothing writes the relationship row. DataContractRepository.getOrCreateTestSuite() sets it via .withDataContract(...) at creation, it's dropped at serialization, and it's never reconstructed. The field is simply gone.

### To Reproduce

Add a test suite to a data contract of a table and then trigger the data contract.

### Expected behavior

After the test pipeline result we should have the data contract done as well.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

_No response_

_No response_

### Additional context

_No response_

### Pre-submission checklist

yan-3005:
## Problem

Regression from **1.13.0** (also OSS issue #30297 and Crocs ticket #20740).

`dataContract` was added to `FIELDS_STORED_AS_RELATIONSHIPS` on `TestSuite`, so the field is stripped from `TestSuite` JSON before storage. But logical test-suite creation never persisted the reverse edge (`testSuite CONTAINS dataContract`).

When a DQ pipeline completes and posts `pipelineStatus`:

1. `TestSuitePipelineStatusHandler` fires and calls `onTestSuiteExecutionComplete`.
2. It reloads the `TestSuite` — `testSuite.getDataContract()` returns `null` (never stored, no reverse fetch).
3. The guard `if (testSuite.getDataContract() != null) { updateContractDQResults(...); }` **silently skips**.
4. Contract sits at `Running` forever, even though the test-case result and pipeline status are written correctly.

Reproduced locally against MySQL + Argo: two contracts on real ingested tables, both pipelines completed with `pipelineState=success`, both contracts stuck at `Running`.

## Fix

**`DataContractRepository.postCreateOrUpdate`** now:

- Writes the reverse relationship (`testSuite -CONTAINS-> dataContract`) **immediately** after the TestSuite is available, before any pipeline deploy work. Idempotent: `findTo` guard before insert.
- Wraps `prepareAndDeployIngestionPipeline` in try/catch — pipeline-service outage no longer aborts contract creation nor loses the reverse link. Mirrors the same catch-all already used in `validateContract` for the runtime path.

**Backfill (v1133 Java migration):** scans `data_contract_entity` in batches of 500, extracts `testSuite.id` from JSON, inserts missing relationship rows. Uses `INSERT IGNORE` (MySQL) / `ON CONFLICT DO NOTHING` (Postgres) — safe to re-run.

## Test plan

- [x] Unit: `MigrationUtilTest` — 8 mocked cases covering empty result, contract without testSuite, single/multiple contracts, batch pagination, idempotency, dialect-specific SQL shape. **8/8 pass.**
- [x] IT: `DataContractResourceIT.testDataContractCreatesReverseTestSuiteRelationship` — end-to-end guard asserting `testSuite.dataContract` is populated after contract create with `qualityExpectations`. **1/1 pass.**
- [x] Live reproduction: created 2 contracts on real MySQL tables via Argo, both stuck at Running with `pipelineState=success`. Applied fix + ran v1133 migration → `entity_relationship` rows inserted → `Run` in UI flips to `Success` on next callback.
- [ ] Backport to 1.13.3 (this PR targets `main`; migration filed under `bootstrap/sql/migrations/native/1.13.3` and `migration/*/v1133` for direct cherry-pick).

## Notes for reviewers

- Backfill also protects existing 1.13.0/1.13.1/1.13.2 → 2.0.0 upgrade paths — the v1133 migration runs on the way up.
- Migration is Java-only (empty `schemaChanges.sql`); `runDataMigration` in `mysql/v1133/Migration.java` and `postgres/v1133/Migration.java` both delegate to the shared `utils/v1133/MigrationUtil.backfillDataContractTestSuiteRelationships(handle, connectionType)`.
- No functional change to the happy path when the pipeline service is healthy at contract create time.

Fixes #30297

This PR restores and backfills the TestSuite-to-DataContract reverse relationship.
- Persists the reverse relationship during data-contract creation before pipeline deployment.
- Makes initial ingestion-pipeline deployment best-effort.
- Adds MySQL and PostgreSQL v1133 backfills plus unit and integration coverage.
</details>

This PR is not yet safe to merge because a failed or partial backfill can still be permanently recorded as successful.

The migration utility catches insert and query failures, and both vendor migration entry points also return normally after exceptions; the migration runner therefore cannot detect the incomplete backfill and retry it, leaving some reverse relationships absent.

**Files Needing Attention:** openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v1133/MigrationUtil.java; openmetadata-service/src/main/java/org/openmetadata/service/migration/mysql/v1133/Migration.java; openmetadata-service/src/main/java/org/openmetadata/service/migration/postgres/v1133/Migration.java
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataContractRepository.java | Persists the reverse relationship before best-effort pipeline deployment. |
| openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v1133/MigrationUtil.java | Implements batched, dialect-specific relationship backfill; the previously reported failure-suppression issue remains. |
| openmetadata-service/src/main/java/org/openmetadata/service/migration/mysql/v1133/Migration.java | Invokes the MySQL backfill but still converts backfill exceptions into successful completion. |
| openmetadata-service/src/main/java/org/openmetadata/service/migration/postgres/v1133/Migration.java | Invokes the PostgreSQL backfill but still converts backfill exceptions into successful completion. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DataContractResourceIT.java | Adds an end-to-end regression assertion for reverse-relationship resolution after contract creation. |
| openmetadata-service/src/test/java/org/openmetadata/service/migration/utils/v1133/MigrationUtilTest.java | Covers batching inputs and dialect SQL shape but intentionally expects swallowed failures rather than migration failure propagation. |

</details>

```mermaid
sequenceDiagram
  participant DC as DataContractRepository
  participant TS as TestSuite
  participant DB as Relationship Store
  participant PS as Pipeline Service
  participant MIG as v1133 Migration
  DC->>TS: Get or create logical test suite
  DC->>DB: Persist TestSuite CONTAINS DataContract
  DC->>PS: Best-effort pipeline deployment
  MIG->>DB: Backfill relationships from stored contracts
  TS->>DB: Resolve DataContract on execution callback
```
</details>

<sub>Reviews (3): Last reviewed commit: ["address review: narrow catch clauses to ..."](https://github.com/open-metadata/openmetadata/commit/0353bf7c6b08e11f2bfda89c5a4c94faf24ae6b7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47580705)</sub>

**Context used:**

- Knowledge Base — [Bootstrap SQL and the Database Migration Runner](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/bootstrap-migrations.md)


yan-3005:
@gitar-bot on both remaining findings:

**1. Migration.java swallow of backfill failure — intentional, not fixing.**

Throwing here would abort `MigrationWorkflow.runDataMigration()` and halt server startup for the whole cluster. That is a worse operational failure than a silent backfill: contracts already stuck at `Running` are recoverable via `POST /dataContracts/{id}/validate` (or the daily `DataContractValidationApp` cron), but a cluster that will not boot blocks every user of the platform.

Trade-off accepted:
- Backfill failure logs `LOG.error("v1133: batch at offset={} failed, aborting backfill loop", offset, e)` (visible in ops logs).
- `SERVER_CHANGE_LOG` marks `1.13.3` applied → migration is not retried automatically. Operators re-run via `openmetadata-ops.sh migrate --force` if they need to.
- Same pattern is used in v200 for `taskWorkflow` migration ([Migration.java:52-63](https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-service/src/main/java/org/openmetadata/service/migration/mysql/v200/Migration.java#L52-L63)) — precedent in this codebase.

If we want operator paging on backfill failure, that belongs in metrics/alerting, not in a startup-abort throw.

**2. MigrationUtilTest brittle SQL-string asserts — acknowledged, not fixing in this PR.**

Real critique: yes, `insertSql.contains(\"INSERT IGNORE\")` will break if we swap to `INSERT ... ON DUPLICATE KEY UPDATE`. Trade-off:
- What the mock test IS validating: dialect branching (MySQL vs Postgres path), per-row skip when `testSuite` absent from JSON, per-row skip on JDBI exception, batch pagination guard.
- What it is NOT validating: the SQL actually runs against a real DB.
- End-to-end coverage of the code path exists in `DataContractResourceIT.testDataContractCreatesReverseTestSuiteRelationship` — that hits the runtime write (`postCreateOrUpdate → ensureTestSuiteToDataContractRelationship`) against a real MySQL container. The backfill path uses the same target row, same relationship shape, same PK constraint — a rewrite that broke real backfill would also break that IT.
- A dedicated Testcontainers migration IT would be additional insurance but adds Docker boot cost per test and duplicates ~90% of the IT's coverage.

Filing a follow-up if we get bitten by a SQL-rewrite that both mocks miss and the create-path IT misses.

sonika-shah (approved):

PR #30538, opened 2026-07-27
fix(profiler): sanitize BigQuery uniqueCount SQL for TIMESTAMP/DATE and STRUCT-subfield columns (#30152)

### LINKED ISSUE #30152: BigQuery profiler: uniqueCount emits invalid SQL for TIMESTAMP/DATE and nested STRUCT-subfield columns
> **Type:** Bug · **Suggested labels:** `bug`, `Ingestion`, `profiler`

## Affected module

BigQuery Profiler (data profiling), `uniqueCount` metric.

## Describe the bug

On BigQuery, `uniqueCount` generates invalid SQL and fails for two kinds of columns:

1. Non-integer flat columns such as TIMESTAMP and DATE.
2. Nested STRUCT-subfield columns such as `customer.email`.

The profiler run then finishes as a partial success: a per-column WARNING is logged and `uniqueCount` is missing in the UI for the affected columns. `distinctCount` is unaffected, which is why the run does not fail outright.

Both failures share one root cause: the outer `countif(...)` is built from a *reconstructed* column that carries the source column's name and type, instead of referencing the integer occurrence-count that the wrapping subquery produces.

The outer expression, in `ingestion/src/metadata/profiler/metrics/static/unique_count.py`, `UniqueCount.query()` (BigQuery branch):

```python
col = column(self.col.name, self.col.type)
if session.get_bind().dialect.name == Dialects.BigQuery:
    return func.countif(col == 1).label(self.name())
```

The wrapping subquery, in `profiler_interface.py`, `_compute_query_metrics` (the `isinstance(metric_query, Label)` block, commented "hotfix to handle transition of unique count implementation"):

```python
subquery = (
    self.session.query(Count(sample_column).fn().label(column.name))  # count(col) AS <column.name>
    .select_from(sample).group_by(sample_column).subquery()
)
```

### Failure mode 1: non-integer flat column (for example TIMESTAMP)

Table `analytics.orders` with a TIMESTAMP column `created_at`.

**Error:**

```
400 ... Unparseable query parameter `created_at_1` in type `TYPE_TIMESTAMP`, Invalid timestamp: '1' value: '1'
```

**Generated SQL:**

```sql
SELECT countif(`created_at` = %(created_at_1:TIMESTAMP)s) AS `uniqueCount`
FROM (SELECT count(`created_at`) AS `created_at`
      FROM `analytics`.`orders` GROUP BY `orders`.`created_at`) AS `anon_1`
```

`col == 1` binds `1` as a parameter typed with the column's own type, so it is declared `%(created_at_1:TIMESTAMP)s`. BigQuery rejects it while parsing the value `1` as a TIMESTAMP, before any comparison runs. Any non-integer type is affected (TIMESTAMP, DATE, and so on). INT and STRING columns happen to work only because `1` is a valid value for those parameter types.

### Failure mode 2: nested STRUCT-subfield column

Table `analytics.orders` with `customer STRUCT<email STRING>`, profiled as `customer.email`.

**Error:**

```
400 ... Unrecognized name: customer; Did you mean customer_email? at [2:16]
```

**Generated SQL:**

```sql
SELECT countif(`customer`.`email` = %(customer.email_1:INT64)s) AS `uniqueCount`
FROM (SELECT count(`customer`.`email`) AS `customer_email`
      FROM `analytics`.`orders` GROUP BY `orders`.`customer`.`email`) AS `anon_1`
```

`.label(column.name)` sanitizes the dotted name to `customer_email`, but the outer `countif` re-emits the raw path `` `customer`.`email` ``, which does not exist in the subquery. A third, latent problem hides behind this: the auto-derived bind-parameter name also contains a dot (`customer.email_1`), which is an illegal BigQuery parameter name. It surfaces once the identifier mismatch is fixed.

## Root cause

In SQLAlchemy, `col == 1` (where `col = column(self.col.name, self.col.type)`) derives three things from `col`, and each is wrong here:

1. The reference identifier becomes the dotted `` `customer`.`email` ``, which does not match the subquery's sanitized `customer_email` label (failure mode 2).
2. The bind-parameter type is taken from the left operand via `col.type.coerce_compared_value(op, 1)`; for the BigQuery temporal column type this returns the column's type, so `1` is declared as `:TIMESTAMP` (failure mode 1).
3. The bind-parameter name is derived from the column, so it contains a dot, for example `customer.email_1` (the latent problem above).

The outer aggregate should instead count the subquery's integer occurrence-count column against an integer `1`, independent of the source column's SQL type or name.

## To reproduce

1. Create a BigQuery table with (a) a TIMESTAMP column and (b) a STRUCT column with a scalar subfield.
2. Run the profiler with `uniqueCount` enabled (the default).
3. The run completes as a partial success; the logs show the two 400 errors above, and those columns have no `uniqueCount`.

## Expected behavior

`uniqueCount` computes correctly for TIMESTAMP and DATE columns and for STRUCT-subfield columns.

## Suggested fix

Keep the single-pass COUNTIF optimization from #24433 (COUNT and UNIQUE_COUNT computed over one grouped CTE). `countif` is BigQuery-only and must stay dialect-gated — it already is, because `UniqueCount.query()` returns a `countif` `Label` only under `if dialect == BigQuery` and every other dialect returns a full query. Rather than introduce a bare `func.countif(...)` in dialect-generic code, keep the countif in that branch but make it reference a shared, dot-free, integer-typed alias, and have both CTE builders label their grouped column with that same alias.

Three coordinated edits, one shared constant `ALIAS = "value_count"`:

1. **`unique_count.py`, `UniqueCount.query()` BigQuery branch** — reference the alias instead of the reconstructed `col`:

   ```python
   if session.get_bind().dialect.name == Dialects.BigQuery:
       return func.countif(column(ALIAS, Integer()) == 1).label(self.name())
   ```

   `Integer()` makes `1` bind as `INT64`; the plain `value_count` name yields a clean identifier and parameter name.

2. **`profiler_interface.py`, `_compute_query_metrics`** (the `isinstance(metric_query, Label)` block) — label the grouped count column with the alias: `Count(sample_column).fn().label(ALIAS)` (was `column.name`).

3. **`columnValuesToBeUnique.py`, `_run_results`** — same relabel: `select(count.label(ALIAS))...` and read it back as `func.sum(grouped_cte.c[ALIAS])` (was `column.name`). Its COUNT has the same dotted-alias flaw for nested columns, so the alias fixes COUNT there too.

Because the countif comes from `UniqueCount.query()`, edit 1 fixes the expression for both call sites; edits 2 and 3 only relabel the grouped column so the reference resolves. This compiles cleanly for flat and nested columns:

```sql
SELECT countif(`value_count` = @value_count_1)
FROM (SELECT count(<col>) AS `value_count` ... GROUP BY <col>) AS `anon_1`
```

**Alternative** (simpler, but drops the optimization): remove the BigQuery COUNTIF special-case in `UniqueCount.query()` so BigQuery falls through to the dialect-agnostic path, `SELECT count(*) FROM (... GROUP BY col HAVING count(col) = 1)`. The outer `count(*)` has no column name or type reference, so it is immune to both failure modes. The trade-off is reverting #24433's single-pass optimization.

## Test coverage gap

`test_unique_count` in `ingestion/tests/unit/observability/profiler/sqlalchemy/test_metrics.py` runs against in-memory SQLite, so it never exercises the BigQuery COUNTIF branch, a nested STRUCT column, or a TIMESTAMP/DATE column. A fix should add BigQuery regression coverage for a TIMESTAMP column and a nested STRUCT-subfield column.

## Version

- OpenMetadata version: 1.12.13
- OpenMetadata Ingestion package version: 1.12.13
- Source: BigQuery
- Python: 3.10

## Additional context

Self-hosted open-source OpenMetadata. Introduced by #24433 ("move to use COUNTIF to not re query the sampled CTE"), still present on `main` as of 2026-07-16. Related but different (closed) issues: #6795 (BigQuery grouping by a whole STRUCT, distinct from our scalar-subfield case), #7159 (STRUCT subfields materialized as dotted columns), #21447 (nested case-sensitivity). Workaround: disable `uniqueCount` via Profiler Configuration (`profilerConfiguration.metricConfiguration[].disabled`), or exclude the affected columns from profiling.

kshitij1439:
### Description
Fixes #30152.

On BigQuery, the `uniqueCount` metric generated invalid SQL and failed for:
1. Non-integer columns (`TIMESTAMP`, `DATE`).
2. Nested `STRUCT` subfield columns (e.g., `customer.email`).

This is distinct from #27256, which fixed a separate BYTES/binary-column type
mismatch in the same function. That fix did not resolve the TIMESTAMP/DATE or
STRUCT-subfield cases described here — confirmed still present on `main` as of
the issue being filed.

### Root Cause
- **Type Mismatch (`TIMESTAMP`/`DATE`)**: `UniqueCount.query` built the outer `COUNTIF` using `column(col.name, col.type)`. This caused SQLAlchemy to bind parameter `1` matching the column's native `TIMESTAMP`/`DATE` type, triggering BigQuery type conversion errors.
- **Dotted Path Resolution (`STRUCT` subfields)**: The outer `COUNTIF` used raw dotted identifiers (e.g. `customer.email`) while the subquery grouped and labeled using sanitized identifiers, resulting in unresolved column errors (`400 Unrecognized name: customer`) and invalid parameter names (`customer.email_1`).

### Solution
- Introduced a shared, untyped integer column alias `VALUE_COUNT_ALIAS = "value_count"`.
- Updated `UniqueCount.query()` to construct `func.countif(column(VALUE_COUNT_ALIAS, Integer()) == 1)`.
- Synchronized `profiler_interface.py` (`_compute_query_metrics`) and `columnValuesToBeUnique.py` (`_run_results`) to label grouped CTE subqueries with `VALUE_COUNT_ALIAS`.
- Added unit tests compiling against `BigQueryDialect` for `TIMESTAMP`, `DATE`, and `STRUCT` columns.

### Affected Modules
- `ingestion/src/metadata/profiler/metrics/static/unique_count.py`
- `ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py`
- `ingestion/src/metadata/data_quality/validations/column/sqlalchemy/columnValuesToBeUnique.py`
- `ingestion/tests/unit/observability/profiler/sqlalchemy/bigquery/test_unique_count_bigquery.py`

### Testing
Ran unit tests with `pytest`:
```
pytest ingestion/tests/unit/observability/profiler/sqlalchemy/bigquery/test_unique_count_bigquery.py
pytest ingestion/tests/unit/observability/profiler/ -q
```

Updates BigQuery unique-count query generation to use a shared integer count alias.
- Labels grouped profiler and data-quality count expressions as `value_count`.
- Makes the outer BigQuery `COUNTIF` reference that alias instead of the original typed or dotted column identifier.
- Adds SQL-compilation coverage for TIMESTAMP, DATE, and nested STRUCT-subfield columns.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/profiler/metrics/static/unique_count.py | Introduces the shared count alias and uses an integer-typed alias reference in BigQuery COUNTIF generation. |
| ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py | Labels the grouped profiler count with the alias expected by the BigQuery outer expression. |
| ingestion/src/metadata/data_quality/validations/column/sqlalchemy/columnValuesToBeUnique.py | Keeps the validator’s grouped CTE producer and aggregate consumer synchronized on the shared alias. |
| ingestion/tests/unit/observability/profiler/sqlalchemy/bigquery/test_unique_count_bigquery.py | Adds compilation tests covering profiler and validator SQL for temporal and nested STRUCT-subfield columns. |

</details>

```mermaid
sequenceDiagram
    participant Caller as Profiler / Validator
    participant CTE as Grouped subquery
    participant BQ as BigQuery
    Caller->>CTE: COUNT(source_column) AS value_count
    CTE->>BQ: Group by source_column
    Caller->>BQ: "COUNTIF(value_count = 1)"
    BQ-->>Caller: uniqueCount
```
</details>

<sub>Reviews (5): Last reviewed commit: ["fix(profiler): gracefully skip BigQuery ..."](https://github.com/open-metadata/openmetadata/commit/ddb9819f5b76c632e3935ee762a7866ad4ef1085) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47729823)</sub>


kshitij1439:
Hi team 👋 this PR seems blocked because #30152 is on Roadmap, not Shipping.
Could someone move it or add skip-pr-checks? Tests pass locally, thanks! 🙏

PR #30606, opened 2026-07-28
Fixes #28911: attach dbt relationships tests to the model holding the foreign key

### LINKED ISSUE #28911: dbt relationship tests incorrectly assigned to referenced table instead of source table
### Affected module

Ingestion Framework

### Describe the bug

Describe the bug
dbt relationship tests are being incorrectly assigned to the referenced table instead of the source table where the foreign key column exists. This causes "Invalid column name" errors during dbt ingestion because OpenMetadata tries to create test cases on columns that don't exist on the target dimension tables.

Example:

dbt test defined on FACT_PRODUCT_STOCK_DETAIL.movement_line_location_key (foreign key)
Test references DIM_LOCATION.location_key (primary key)
OpenMetadata incorrectly tries to create test on DIM_LOCATION.movement_line_location_key
Error: "Invalid column name movement_line_location_key" (column doesn't exist on DIM_LOCATION)
This affects 265 relationship tests in our project (66% of all relationship tests), while other test types (not_null, unique, accepted_values) ingest correctly.

Versions affected: 1.12.10, 1.13.0

### To Reproduce

1. Create a dbt model with a relationship test:

```
# models/fact_product_stock_detail.yml
models:
  - name: fact_product_stock_detail
    columns:
      - name: movement_line_location_key
        description: Foreign key to dim_location
        tests:
          - relationships:
              to: ref('dim_location')
              field: location_key
```

2. Ensure both tables exist in Snowflake:

FACT_PRODUCT_STOCK_DETAIL with column movement_line_location_key
DIM_LOCATION with column location_key

3. Run dbt test successfully:

```
dbt test --select fact_product_stock_detail
# Test passes in dbt
```

4. Configure OpenMetadata dbt ingestion pipeline with:

- manifest.json, catalog.json, run_results.json from dbt artifacts

5. Run dbt ingestion in OpenMetadata

Actual Result:

```

Failed to ingest CreateTestCaseRequest [relationships_fact_product_stock_detail_movement_line_location_key__location_key__ref_dim_location_]
APIError: Invalid column name movement_line_location_key
400 Client Error: Bad Request for url: /api/v1/dataQuality/testCases
```

### Expected behavior

The test case should be created on the source table where the foreign key column is defined:

Expected FQN:

`dev-snowflake-warehouse.DEV_WAREHOUSE.MART.FACT_PRODUCT_STOCK_DETAIL.movement_line_location_key.relationships_...
`
The test validates that values in FACT_PRODUCT_STOCK_DETAIL.movement_line_location_key exist in DIM_LOCATION.location_key.

Expected result: Test case created successfully, appears in OpenMetadata UI under the source table's data quality tab.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

_No response_

_No response_

### Additional context

Additional context
Scale of impact:

265 relationship tests failed in our production deployment
Affects fact tables (181 tests), dimension tables (66 tests), link tables (17 tests)
Most common failure patterns: date keys, currency keys, employee/customer keys
Root cause:
The dbt parser assigns tests to to: ref('target_table') instead of the source model where the test is defined. In dbt, relationship tests belong to the source table with the foreign key, not the referenced table.

Root cause:
The dbt parser assigns tests to to: ref('target_table') instead of the source model where the test is defined. In dbt, relationship tests belong to the source table with the foreign key, not the referenced table.

Environment:

Data warehouse: Snowflake
Deployment: Kubernetes
dbt project: 900+ models
Other test types work correctly (not_null, unique, accepted_values)
Suggested fix:

```
# Current (incorrect)
test_table = relationship_test.target_table  # dimension table

# Should be
test_table = relationship_test.source_table  # fact table with FK
```

### Pre-submission checklist

ulixius9:
## Describe your changes

dbt `relationships` tests were created on the referenced (parent) table instead of the model holding the foreign key, so the server rejected them with `Invalid column name`. This resolves the tested model by dbt node name, so the test lands on the correct table.

Fixes #28911

## Root cause

The reporter's diagnosis ("the parser assigns tests to `to: ref('target_table')`") is not quite right — logic to read `test_metadata.kwargs['model']` already exists (#27366, present in 1.12.10 and 1.13). The defect is in how the extracted name is matched, and what happens when the match fails:

1. `re.search(r"ref\(...")` on `kwargs['model']` yields the dbt model **name** (`orders`).
2. It is matched with `fqn.endswith(f".{name}")` against `upstream_list`, whose FQNs are built in `parse_upstream_nodes` from `get_dbt_model_name(parent_node)` = **`alias` if set, else name**.
3. Name vs alias (or case) diverge → no match → silent fallback to `upstream_list[0]`.
4. `upstream_list` preserves `depends_on.nodes` order, and dbt lists the **parent `to` model first**.

So the fallback deterministically picks the referenced table and attaches the FK column to it. Verified against this repo's own fixture `manifest_versionless.json`, where a test on `orders.customer_id` has `depends_on = ['...customers', '...orders']`.

Three confirmed triggers for the match failing:
- model has an `alias` differing from its name (common on Snowflake with `FACT_`/`DIM_` uppercase naming)
- the test targets a `source()` — the old regex only handled `ref(`
- the child table is absent from `upstream_list` (filtered by `is_filtered`, or an `_get_table_entity` ES miss)

This explains the reporter's partial 66% failure rate: it is per-model, not universal.

## Changes

**`dbt_utils.py`** — new `get_dbt_test_primary_table_fqn()` resolves the tested table in three steps:
1. exact lookup of the `ref()`/`source()` name against a new dbt node name → FQN map
2. case-insensitive suffix match (handles warehouses that upper-case identifiers)
3. for relationships tests, exclude the `to:` upstream — if exactly one candidate remains it is the tested table

`_DBT_REF_PATTERN` now also handles `source('schema', 'table')` and multi-arg `ref('pkg', 'model')`.

**`metadata.py`** — `parse_upstream_nodes_with_names()` returns `(dbt node name, FQN)` pairs; `parse_upstream_nodes()` keeps its existing FQN-list contract, so lineage/exposure/data-model callers are unaffected.

**Second defect fixed:** `create_dbt_test_case_results` looped over *every* upstream and posted results to a test case FQN on each, so multi-upstream tests wrote results to a table that never had the test case (the failure was swallowed at debug level). It now resolves the same single table.

## Type of change

## Tests

Five regression tests added to `ingestion/tests/unit/test_dbt.py`, driven by the real `relationships` node in `manifest_versionless.json` parsed through `collate_dbt_artifacts_parser` — covering the aliased-upstream case, resolution without a name map, name-based resolution, and `get_dbt_test_primary_table_fqn` for both single- and multi-upstream tests.

Confirmed non-vacuous by running the same scenarios against the pre-fix code, which returns `CUSTOMERS` where the fixed code returns `ORDERS`.

```
157 passed in 4.06s
ruff check: All checks passed!
ruff format --check: 9 files already formatted
```

This PR corrects dbt test ownership and result association.
- Preserves dbt node names and namespaces alongside upstream table FQNs.
- Resolves `ref()` and `source()` expressions to the tested table, including aliases, case differences, and multi-upstream relationship tests.
- Posts test results only to the table on which the test case was created.
- Adds regression coverage for aliases, ambiguous names, namespaces, and relationship-test resolution.
</details>

The functional changes appear safe to merge, with only the previously reported test-style cleanup still outstanding.

The package-qualified identity fix now retains namespace-qualified keys and marks ambiguous bare names rather than silently selecting one table; no blocking failure remains, while the added tests still use the previously flagged unittest assertion style.

**Files Needing Attention:** ingestion/tests/unit/test_dbt.py
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/source/database/dbt/dbt_utils.py | Adds namespace-aware upstream indexing and centralized primary-table resolution for dbt tests. |
| ingestion/src/metadata/ingestion/source/database/dbt/metadata.py | Retains upstream node identity during parsing and routes each test result to one resolved table. |
| ingestion/src/metadata/ingestion/source/database/dbt/models.py | Introduces the typed upstream-node representation used to retain dbt identity alongside table FQNs. |
| ingestion/tests/unit/test_dbt.py | Adds behavioral regression coverage for relationship-test ownership, aliases, ambiguous names, and source namespaces. |

</details>

```mermaid
flowchart LR
  M[dbt test manifest node] --> D[Resolve upstream dependencies]
  D --> N[Build bare and namespace-qualified name map]
  M --> R[Extract model ref or source expression]
  N --> P[Resolve primary tested table FQN]
  R --> P
  P --> T[Create test case entity link]
  P --> O[Post test result to the same table]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(ingestion): skip upstream nodes whos..."](https://github.com/open-metadata/openmetadata/commit/c58163bc3e7e0f24399b67454ec33a989a91c09b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48061236)</sub>

**Context used:**

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/dbt/constants.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt/constants.py
@@ -159,6 +159,7 @@ class DbtCommonEnum(Enum):
     RESOURCETYPE = "resource_type"
     MANIFEST_NODE = "manifest_node"
     UPSTREAM = "upstream"
+    UPSTREAM_BY_NAME = "upstream_by_name"
     RESULTS = "results"
     TEST_SUITE_NAME = "test_suite_name"
     DBT_TEST_SUITE = "DBT_TEST_SUITE"
--- a/ingestion/src/metadata/ingestion/source/database/dbt/dbt_utils.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt/dbt_utils.py
@@ -31,7 +31,7 @@ from metadata.ingestion.source.database.dbt.constants import (
     DbtCommonEnum,
     RawQueriesEnum,
 )
-from metadata.ingestion.source.database.dbt.models import SnapshotNodeLocation
+from metadata.ingestion.source.database.dbt.models import SnapshotNodeLocation, UpstreamNode
 from metadata.utils import entity_link
 from metadata.utils.logger import ingestion_logger
 
@@ -718,10 +718,15 @@ def get_dbt_test_description(manifest_node) -> Optional[str]:  # noqa: UP045
     return description
 
 
-# Matches a dbt ``ref()``/``source()`` expression, capturing the last quoted argument,
-# which is the model or table name. Handles ref('t'), ref("pkg", "t"), source('s', 't')
-# and the get_where_subquery(...) wrapper dbt adds to test_metadata.kwargs['model'].
-_DBT_REF_PATTERN = re.compile(r"\b(?:ref|source)\(\s*['\"](?:[^'\"]+['\"]\s*,\s*['\"])*([^'\"]+)['\"]\s*\)")
+# Matches a dbt ``ref()``/``source()`` expression, capturing the optional namespace
+# (package for ref, source name for source) and the model/table name. Handles ref('t'),
+# ref("pkg", "t"), source('s', 't') and the get_where_subquery(...) wrapper dbt adds to
+# test_metadata.kwargs['model'].
+_DBT_REF_PATTERN = re.compile(r"\b(?:ref|source)\(\s*['\"]([^'\"]+)['\"]\s*(?:,\s*['\"]([^'\"]+)['\"]\s*)?\)")
+
+# Marks a bare dbt name claimed by more than one upstream table. Looking such a name up
+# must fall through to the other resolution strategies rather than pick an arbitrary one.
+_AMBIGUOUS_NAME = None
 
 
 def _get_test_kwargs(manifest_node) -> Dict[str, Any]:  # noqa: UP006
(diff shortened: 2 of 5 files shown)


harshsoni2024 (dismissed):
lgtm, can we look into this edge case marked - https://github.com/open-metadata/OpenMetadata/pull/30606#discussion_r3668515372

ulixius9:
Thanks @gitar-bot — the finding is valid and now fixed in d3b7f1a.

I checked whether the collision is actually reachable rather than assuming it isn't. It is: `manifest_entities` merges `nodes` and `sources` ([metadata.py:303-307](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/database/dbt/metadata.py#L303-L307)), and a source node's `name` is its *table* name. So a model `orders` and a source table `orders` can both be upstreams of the same test — dbt namespaces them separately (`model.pkg.orders` vs `source.pkg.raw.orders`), so nothing prevents it. The two-argument `ref('pkg', 'model')` form is reachable the same way.

I went with disambiguation rather than "prefer the first entry", since first-vs-last is equally arbitrary:

- Upstreams are now `UpstreamNode` records carrying the dbt name plus its namespace (`package_name` for models, `source_name` for sources).
- `build_upstream_name_map` indexes **both** the qualified (`raw.orders`) and bare (`orders`) names, and marks a bare name claimed by two different tables as ambiguous so lookups fall through to the remaining strategies instead of guessing.
- `_DBT_REF_PATTERN` now captures the namespace, so `ref('pkg', 'model')` and `source('src', 'table')` resolve against the qualified key.
- `_match_fqn_by_table_name` likewise returns nothing when several upstreams match, rather than taking the first.

On the "the suffix fallback masks most of this" point — it didn't, it had the same defect, so that's fixed too.

Verified the new tests aren't vacuous by running both orderings against the previous `dict()` behaviour:

```
                 previous dict()       new map
[source, model]  svc.db.sch.orders     svc.db.sch.orders
[model, source]  svc.db.raw.orders     svc.db.sch.orders   <-- previous picks the wrong table
```

The `[model, source]` ordering resolves to the wrong table under the old last-wins map and to the correct one now; the test asserts both orderings so the result cannot depend on index order.

161 passed, ruff clean.

harshsoni2024 (approved):

PR #30660, opened 2026-07-29
ci(playwright): instrument prepare-playwright-fixture phase timings

chirag-madlani:
## Summary

`prepare-playwright-fixture` costs ~17 min on the cache-miss path; ~614s (60%) sits in the monolithic `Prepare seeded environment` step. This PR adds phase-level instrumentation so we know where the 10 min actually goes before proposing fixes.

- Wrap 9 phases of `run_local_docker_main` (`compose_down_prev`, `docker_build_up`, `wait_search_index`, `wait_airflow_api`, `wait_sample_data_dag`, `wait_om_server`, `dag_ingest_sample_data`, `dag_ingest_secondary`, `reindex_search`) with opt-in `om_phase_start`/`om_phase_end` helpers.
- When `OM_TIMING_LOG` is set, each phase writes `<name>\t<seconds>` to that file; when unset, helpers just print human-readable markers (local runs unaffected).
- Workflow initializes the log before `Prepare seeded environment`, renders it into `$GITHUB_STEP_SUMMARY` as a markdown table, and uploads the raw TSV (14-day retention) so we can diff across runs.

`docker/run_local_docker_common.sh` is on the fixture fingerprint prefix list, so this PR intentionally busts the fixture cache — the cache-miss cold path runs on this branch's own CI, giving us real data.

## Test plan

- [x] `bash -n docker/run_local_docker_common.sh` — clean
- [x] YAML parse — clean
- [x] Local dry-run of the helpers writes the TSV correctly and awk renders the table
- [ ] CI run on this branch produces a populated **run_local_docker.sh — phase timings** table in the `prepare-playwright-fixture` step summary
- [ ] `prepare-fixture-phase-timings-<run>-<attempt>` artifact contains the TSV

## Follow-ups (separate PRs, once data lands)

- If `docker_build_up` dominates → cache image layers.
- If `dag_ingest_sample_data` dominates → cut sample-data volume for fixture-mode.
- If `reindex_search` dominates → skip nightly-only indices for the fixture.
- Parallelize `Free Disk Space` with `Checkout` + `Download distribution`.

This PR instruments fixture preparation and adds reuse of cached ingestion images.
- Records phase durations from `run_local_docker_main` and publishes them as a workflow summary and retained TSV artifact.
- Restores, validates, and loads a cached ingestion image to bypass rebuilding it when available.
- Expands ingestion-image snapshot validation and cache saving to fixture rebuilds that do not currently require Airflow.
- Adds a Compose override for the preloaded ingestion image and documents the fast Playwright overlay.
</details>

The PR appears safe to merge, although its timing report remains inaccurate for retried or failed fixture setup attempts.

Retried startup commands reuse an append-only timing file, while three active phases can terminate before recording their duration; these limitations affect the new diagnostics rather than fixture execution itself.

**Files Needing Attention:** docker/run_local_docker_common.sh and .github/workflows/playwright-postgresql-e2e.yml
</details>

| Filename | Overview |
|----------|----------|
| docker/run_local_docker_common.sh | Adds phase instrumentation and a cached-ingestion-image startup path; the previously reported retry accumulation and failed-phase omission remain. |
| .github/workflows/playwright-postgresql-e2e.yml | Initializes, reports, and uploads timing data while adding ingestion-image restore and broader cache population behavior. |
| docker/development/docker-compose-fixture-cache.yml | Adds a CI-only overlay that selects the validated preloaded ingestion image and disables pulling. |
| docker/development/docker-compose-playwright-fast.yml | Adds explanatory comments without changing Compose behavior. |

</details>

<sub>Reviews (4): Last reviewed commit: ["ci(playwright): bust fixture cache to va..."](https://github.com/open-metadata/openmetadata/commit/c12d0005c8164ec67d5a6b07655a0d792b4cacae) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48369845)</sub>

### DIFF
--- a/.github/workflows/playwright-postgresql-e2e.yml
+++ b/.github/workflows/playwright-postgresql-e2e.yml
@@ -754,6 +754,13 @@ jobs:
           name: openmetadata-distribution
           path: openmetadata-dist/target
 
+      - name: Initialize fixture phase-timing log
+        if: ${{ needs.restore-playwright-fixture.outputs.fixture_cache_hit != 'true' || (needs.plan-playwright.outputs.requires_airflow == 'true' && needs.restore-playwright-fixture.outputs.ingestion_cache_hit != 'true') }}
+        run: |
+          mkdir -p "$RUNNER_TEMP/fixture-timings"
+          : > "$RUNNER_TEMP/fixture-timings/run_local_docker_phases.tsv"
+          echo "OM_TIMING_LOG=$RUNNER_TEMP/fixture-timings/run_local_docker_phases.tsv" >> "$GITHUB_ENV"
+
       - name: Prepare seeded environment
         if: ${{ needs.restore-playwright-fixture.outputs.fixture_cache_hit != 'true' || (needs.plan-playwright.outputs.requires_airflow == 'true' && needs.restore-playwright-fixture.outputs.ingestion_cache_hit != 'true') }}
         uses: ./.github/actions/setup-openmetadata-test-environment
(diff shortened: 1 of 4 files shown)


chirag-madlani:
Closing — the profiling data was useful (docker_build owns 5:13 = 62% of run_local_docker time on cache miss) but the fix approach (cache the built ingestion image + docker load on next run) doesn't deliver in practice: GHA repo cache is at 9.4GB/10GB, PR-scoped 1.2GB entries get LRU-evicted within ~30 min, and the workflow has no push trigger to populate main-scope caches that PR runs could fall back to. Fast path would only fire on back-to-back PR runs within the same batch. Not worth the complexity for that edge case. If we want the win, better path is (a) add a main-push cache-warmer job or (b) reduce ingestion image size — separate work.

PR #30667, opened 2026-07-29
fix(ingestion): correct BigQuery uniqueCount SQL for TIMESTAMP/DATE and nested STRUCT columns

### LINKED ISSUE #30152: BigQuery profiler: uniqueCount emits invalid SQL for TIMESTAMP/DATE and nested STRUCT-subfield columns

rajivbasnet:
### Describe your changes:

Fixes #30152

I fixed the BigQuery `uniqueCount` metric because it emitted invalid SQL and silently dropped the metric (partial-success WARNING, missing value in the UI) for two kinds of columns:

1. **Non-integer flat columns** (TIMESTAMP, DATE) — `400 Unparseable query parameter ... in type TYPE_TIMESTAMP`.
2. **Nested STRUCT-subfield columns** (`customer.email`) — `400 Unrecognized name: customer`.

Both share one root cause. The single-pass COUNTIF optimization from #24433 builds the outer `countif(...)` from a *reconstructed* column (`col = column(self.col.name, self.col.type)`), so `col == 1`:
- binds `1` with the source column's own SQL type (`1` declared as `:TIMESTAMP`), and
- re-emits the source column's identifier (the dotted `` `customer`.`email` ``), which does not exist in the wrapping grouped subquery (whose label is sanitized to `customer_email`).

The outer aggregate should instead count the subquery's **integer occurrence-count** column against an integer `1`, independent of the source column's SQL name or type.

**The fix** keeps the single-pass COUNTIF optimization and introduces one shared, dot-free, integer-typed alias `value_count` (`UNIQUE_COUNT_VALUE_ALIAS`). The BigQuery branch references that alias, and every producer of the grouped occurrence-count column labels it with the same alias so the reference resolves:

1. `metrics/static/unique_count.py` — `func.countif(column(UNIQUE_COUNT_VALUE_ALIAS, Integer()) == 1)` (`Integer()` binds `1` as INT64; the plain name yields a clean identifier and parameter name).
2. `interface/sqlalchemy/profiler_interface.py` — the transition hotfix subquery labels its `Count(...)` with the alias.
3. `data_quality/.../columnValuesToBeUnique.py` — the grouped CTE labels its count with the alias and reads it back via `func.sum(grouped_cte.c[UNIQUE_COUNT_VALUE_ALIAS])` (this had the same dotted-alias flaw for nested columns).

Resulting SQL compiles cleanly for flat and nested columns:

```sql
SELECT countif(`value_count` = @value_count_1)
FROM (SELECT count(<col>) AS `value_count` ... GROUP BY <col>) AS `anon_1`
```

#
### Type of change:

- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered
- BigQuery profiler computes `uniqueCount` for a TIMESTAMP column without emitting an invalid `:TIMESTAMP`-typed bind for the integer `1`.
- BigQuery profiler computes `uniqueCount` for a nested STRUCT-subfield column (`customer.email`) without re-emitting the dotted path or a dotted bind-parameter name.

#### Unit tests
- Files added: `ingestion/tests/unit/observability/profiler/sqlalchemy/bigquery/test_unique_count.py`
- The existing `test_unique_count` runs against in-memory SQLite and never exercised the BigQuery COUNTIF branch (noted in the issue's "Test coverage gap"). The new tests compile the generated SQL against the BigQuery dialect (no live BigQuery needed) and assert the identifier and the INT64 bind are well-formed for TIMESTAMP and nested STRUCT-subfield columns.

- [x] Not applicable (unit-level regression coverage is sufficient; the SQL is compiled against the BigQuery dialect).

#### Playwright (UI) tests

#### Manual testing performed
1. `make py_format_check` equivalent (`ruff check` + `ruff format --check`) on the changed files — clean.
2. Ran the new BigQuery regression suite and the surrounding profiler / data-quality unit suites (metrics, impact scoring, dataframe validator) — all pass.

#

Not applicable.

#
### Checklist:

- [x] My PR is linked to a GitHub issue via `Fixes #30152` above.
- [x] I have added a test that covers the exact scenario we are fixing (TIMESTAMP and nested STRUCT-subfield columns on the BigQuery dialect). The test references issue #30152.

The PR corrects BigQuery unique-count SQL generation.
- Introduces a shared, integer-typed `value_count` alias for grouped occurrence counts.
- Updates profiler and data-quality query producers to use the shared alias.
- Adds BigQuery dialect regression coverage for timestamp and nested STRUCT columns.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/profiler/metrics/static/unique_count.py | The BigQuery COUNTIF expression now references the integer occurrence-count alias instead of reconstructing the source column. |
| ingestion/src/metadata/profiler/interface/sqlalchemy/profiler_interface.py | The profiler's grouped count subquery now labels its output consistently with the BigQuery COUNTIF consumer. |
| ingestion/src/metadata/data_quality/validations/column/sqlalchemy/columnValuesToBeUnique.py | The non-dimensional uniqueness validator consistently labels and reads the grouped occurrence-count column. |
| ingestion/src/metadata/profiler/orm/functions/unique_count.py | Defines the shared occurrence-count alias used by profiler and data-quality query paths. |
| ingestion/tests/unit/observability/profiler/sqlalchemy/bigquery/test_unique_count.py | Adds SQL-compilation regressions for timestamp typing, nested identifiers, and integer bind parameters. |
| ingestion/tests/unit/profiler/metrics/test_unique_count.py | Updates metric-level assertions for the new alias and integer SQLAlchemy type. |

</details>

<sub>Reviews (2): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/mai..."](https://github.com/open-metadata/openmetadata/commit/f8a869a92da249c178c7e7d8c5140baa1388f1eb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48444087)</sub>


ingestion/src/metadata/profiler/orm/functions/unique_count.py
```diff
@@ -25,6 +25,8 @@
 from metadata.profiler.orm.registry import Dialects
 from metadata.profiler.orm.types.custom_image import CustomImage
 
+UNIQUE_COUNT_VALUE_ALIAS = "value_count"
```
TeddyCr, line 28: I am not sure that would be correct, as we do have another metric that is actual value

PR #30686, opened 2026-07-30
fix(dar): validate resolve intent + transitions, block workflow-owned PATCH, freeze payload after Open, bound expirationDate, and reject phantom-target entity types

yan-3005:
## Summary

Plugs 8 HIGH-severity + 5 MEDIUM-severity backend defects flagged in the DAR 2.0.0 QA report (pre-release). Each fix is a design-level guard, not a workaround.

DAR (Data Access Request) is a 2.0.0 feature that has not shipped yet, so nothing here needs a migration for existing data.

## HIGHs (security / integrity — fix first)

| # | Bug | Fix |
|---|---|---|
| **H1** | `POST /resolve` with empty body / unknown `transitionId` silently drove the task to Approved via the workflow handler's positive-default path | `requireExplicitResolveIntent` rejects empty body (no transitionId AND no resolutionType); `validateTransition` rejects unknown transitionIds against the task's `availableTransitions`. |
| **H2** | `transitionId=reject + resolutionType=Approved` was silently accepted (short-circuit on the caller-supplied resolutionType) | `validateTransition` cross-checks the caller's resolutionType against the transition's declared resolutionType and 400s on mismatch. |
| **H3** | `POST /close` on a Granted DAR marked it Cancelled without running the revoke enforcement path (PolicyAgent, revoke boundary timer) | `validateTaskCanBeResolved` blocks close/resolve when the task is in an authorization-outcome state unless the workflow still exposes forward transitions. |
| **H4** | Any user could `PATCH /status = Granted` (or Approved/Rejected/…) and skip the workflow entirely, forging an authorization decision | `checkRestrictedTarget`: (a) always-block PATCH on forge-attribution fields — `/resolution`, `/approvedBy*`, `/aboutFqnHash`; (b) `/status` PATCH allowed only for progress markers (Open/InProgress/Pending), never for workflow-decision statuses (Granted/Approved/Rejected/ManualRevoke/Cancelled). Empty JSON-Pointer (RFC 6901 whole-doc target) also rejected. |
| **H5** | After a DAR was Approved, the requester could `PATCH /payload` and widen their own access (Masked → FullAccess) retroactively | `/payload` frozen once the task has left Open. Same guard in the resource layer covers PUT and PATCH. |
| **H6** | `PATCH /about` retargeted a DAR to a new entity, bypassing the "one active DAR per (creator, target)" dedup check | `/about` frozen after Open; while Open, `TaskUpdater` re-runs `validateNoDuplicateActiveDataAccessRequest` with self-exclusion so the dup check catches the retarget. |
| **H7** | `expirationDate: 1e400` parsed as `Double.POSITIVE_INFINITY`, Long coercion in `readDataAccessPayload` silently swallowed the failure, task stored "never-expires" access | Schema clamps `expirationDate` to `1..7258118400000` (~year 2200); `TaskFieldValidator.validateDataAccessRequestExpiry` also caps at 10 years from now; `readDataAccessPayload` no longer swallows `IllegalArgumentException`. |
| **H8** | Concurrent DAR create by the same user won the SELECT-then-INSERT TOCTOU and produced two active DARs on the same entity | Java-only enforcement: `validateNoDuplicateActiveDataAccessRequest` runs on both create and PATCH (with self-id exclusion). The narrow parallel-create window is accepted as a design tradeoff — a DB constraint would need a follow-up migration that is not warranted for a not-yet-released feature. |

## MEDIUMs

| # | Bug | Fix |
|---|---|---|
| **M4** | `POST /tasks/{id}/comments` accepted from anyone; response leaked the full task payload (private DAR reason, access type, column list) to unrelated callers | Anyone-can-comment stays (collaboration surface). Payload-leak concern is moved to a view-side redaction TODO on GET paths — noted in the code, deliberately out of scope here. |
| **M7** | While a DAR was Open, the requester could `PATCH /payload/expirationDate` to a past timestamp | `TaskUpdater.updatePayload` rejects the update when `expirationDate` changed and is `<= now`, with the same 400 message as the create-time check. |
| **M9** | DAR payload accepted unknown fields; `columns` array unbounded on `ColumnLevel` requests | `additionalProperties: false`, `columns` items `minLength: 1` + `uniqueItems: true`, `reason` `minLength: 10, maxLength: 2000`, `pattern: "\\S"`, `ticketId` `maxLength: 128`. Kept `manualGrantReason` explicitly (workflow write path). Columns intentionally left un-`maxItems`-bounded — wide tables (500+ columns) are legitimate. |
| **M10** | DAR created against a dashboard / topic / pipeline was accepted, but no enforcement path exists for those entity types → phantom "granted" record | `validateDataAccessCapabilities` rejects DAR unless `about.type` is in the allow-list (table, databaseSchema, database, storedProcedure, dataProduct). |
| **M11** | `POST /tasks` with empty / null body threw NPE inside `getTask`, surfaced as 500 with class/method leaked in the log line | Null-body guard on `create()` + `createOrUpdate()` returns a clean 400 up front. |
| ~~M1~~ | Rolled back — mapping `IllegalStateException → 409` globally was too broad; would mask real 500s. Re-check at the resolve race site with a typed exception if needed. |
| ~~M3~~ | Superseded by H1's `requiresComment` enforcement — was reject/revoke without a comment; the workflow's `requiresComment` flag is now honored at the API boundary. |

## Design decisions worth flagging

- **H4 admin does NOT bypass.** Admin PATCH still cannot land the task in Granted / Approved / … — Flowable doesn't advance for a direct-PATCH, so admin bypass would create the exact runtime/row desync the guard exists to prevent. The OSS test that seeded a Granted DAR via `SDK.update()` moved to collate-integration-tests where the fixture walks the real `approve → markAsGranted` workflow.
- **No DB migration.** DAR ships in 2.0.0 which has not been released, so there is no live data to protect. The earlier attempt at a partial unique index on `task_entity` for H8 has been fully rolled back — `bootstrap/sql/migrations/native/2.0.0/{mysql,postgres}/schemaChanges.sql` no longer touch `task_entity`.
- **DAR IT relocated.** `DataAccessRequestValidationIT` and the DAR-Granted status-group tests (previously in `openmetadata-integration-tests`) required the `policyAgentTask` node handler and the DAR workflow, both of which are Collate-only. `OpenMetadataApplication` cannot deploy the workflow (`policyAgentTask is a Collate-only feature — handler not registered`), so those tests never made sense in this module — they moved to `collate-integration-tests` in the paired collate PR.

## Test plan

- [x] Unit — `SetApprovalAssigneesImplTest` (14) + `TaskFieldValidatorTest` (5) + `SetEntityAttributeImplTest` + `RunIngestionPipelineImplTest` + `TaskFormSchemaValidatorTest`: 25/25 pass locally.
- [x] IT — `TaskResourceIT` + `TaskCommentsIT` + `FeedTaskAuthzIT` + `TaskFormSchemaResourceIT`: `mvn test` green locally on the OSS module (no residual failures after the DAR-Granted tests moved to collate).
- [x] IT — `DataAccessRequestValidationIT` (20 tests) + `DarStatusGroupIT` (4 tests): 24/24 pass locally against `CollateApplication` (paired collate PR).
- [x] Spotless — `mvn spotless:apply` clean.
- [ ] CI to run full backend + collate lanes.


yan-3005:
Gitar-bot's remaining finding — "Removing DAR constraint translation reintroduces 500 on race" — is stale as of the migration-revert commits.

The whole DB-layer H8 defense (partial unique index on Postgres, virtual column + unique key on MySQL, and the `translateActiveDarConstraintViolation` translator in `TaskRepository.storeEntity`) was rolled back per the design call to keep DAR out of the migration surface entirely (2.0.0 is unreleased, no live data to protect). With no constraint on the row anymore, there is no `UnableToExecuteStatementException` to translate — the losing thread in a concurrent DAR create no longer surfaces a 500, it just wins the second `INSERT` after the Java-level pre-check races, producing duplicate active DARs. That narrow window is a known TOCTOU gap; documented in the PR body, accepted as the design trade-off for keeping the fix Java-only.

If we later decide to reintroduce the DB constraint in a proper 2.1.0 migration (with a dedup pre-step for real production data), the translator comes back with it. Not in scope here.

yan-3005:
@gitar-bot the "Removing DAR constraint translation reintroduces 500 on race" finding is stale — the DB unique index and the `translateActiveDarConstraintViolation` translator were both removed together in the migration-revert commits (see PR body under "Design decisions"). No constraint on the row → no `UnableToExecuteStatementException` to translate → losing thread does not 500, it just produces a duplicate (accepted TOCTOU trade-off for keeping DAR out of the 2.0.0 migration surface). If we later add the constraint back in a proper 2.1.0 migration with a dedup pre-step, the translator comes back with it. Please re-scan and mark resolved.

yan-3005:
Superseded by #30969 — same branch, opened from the upstream open-metadata/OpenMetadata head instead of a fork so the collate submodule CI can fetch the PR head and pin correctly (fork PR failed the data-access-request-e2e job at the submodule checkout step).

PR #30693, opened 2026-07-30
feat(ui): component-reuse guard, SonarJS rules, and a required Sonar quality gate

karanh37:
### Describe your changes:

Fixes #<issue-number>

I worked on a **UI code-quality gate** because AI-generated frontend code has been landing with
avoidable slop: components hand-rolled from raw `<div>`s when `@openmetadata/ui-core-components`
already exports them, needless complexity and duplication, and no consistent enforcement point.

The governing rule throughout: **new code must be clean; existing debt is fixed gradually.** Every
gate here is scoped to what a change *adds*, so the backlog can never block a PR.

#
### Type of change:
- [x] Improvement

#
### High-level design:

**Why "added lines only" instead of a warn tier.** `ui-checkstyle` runs `eslint --fix` and fails on
the resulting git diff, and `lint:base` has no `--max-warnings`. So a `warn` rule is useless in this
repo in exactly two ways: an auto-fixable rule at `warn` still rewrites files and hard-fails the
required check on legacy code, while a non-fixable rule at `warn` enforces nothing at all. Every new
check therefore either inspects only added lines (following the existing `tw-deprecation-guard`
precedent) or was measured to have a zero backlog first.

**1. `scripts/reuse-audit.js` — component reuse.** No off-the-shelf linter knows this design system,
so this is the one bespoke check. Four detectors: ARIA-role widgets, raw native controls inside
`src/components/**`, hand-rolled arrow-key navigation, and `createPortal` overlays. Three modes
(pre-commit / CI base-ref / `--file`) sharing one implementation.

**2. SonarJS in ESLint.** `eslint-plugin-sonarjs` is the same engine and the same `Sxxxx` rule ids as
the SonarCloud analysis already running on every UI PR. Enabled **16 rules measured to have zero
existing violations**; nine high-backlog rules are deliberately excluded (`no-duplicate-string` 640,
`cognitive-complexity` 85, `no-collapsible-if` 21, …) because ESLint reports per *file* and those
belong to Sonar's new-*line* model. Pinned exactly at `4.2.0` — analyzer drift is silent.

**3. Sonar quality gate as a required check.** New `ui-sonar-gate` job. The result comes from the
scanner via `-Dsonar.qualitygate.wait=true`; an earlier polling approach was rejected because it
races SonarCloud's asynchronous report processing and silently *passes* on timeout, which would make
a required check meaningless.

**4. Local/CI parity.** `make ui-checkstyle-changed` previously ran only the fixing steps — `tw-audit`,
`tw-guard` and `reuse-audit` were missing, so local could pass while CI failed. All three now run.

**Alternatives rejected:** a new dedicated workflow (a new required check needs fresh branch-protection
wiring and reintroduces the paths-filter deadlock that the always-runs `ui-checkstyle` job already
solves); making SonarCloud's own check required (it is gated behind paths-filter and the `safe to
test` label, so PRs with no UI changes would block forever on a check that never posts).

**Requires admin action before the gate is live** — see `docs/ui-code-quality-gate.md`:
- Create the SonarCloud quality gate: **Coverage on New Code ≥ 90%**, Issues 0, Hotspots Reviewed
  100%, Duplicated Lines ≤ 3%, and **no condition on Overall Code**.
- Mark `ui-sonar-gate` required on `main`.

> ⚠️ **90% coverage on new code is stricter than Sonar's 80% default**, and the UI has no coverage
> floor today (`jest.config.js` sets `collectCoverageFrom` but no `coverageThreshold`). Expect this to
> be the condition that fails most PRs at first — new components, hooks and utils will need tests in
> the same PR. That is the intent, but the team should hear it before they hit a red check.

#
#### Use cases covered
- A PR adding `<div role="listbox">` instead of `Select` fails `ui-checkstyle` with the component to
  use named in the PR comment.
- A PR touching a legacy file that already contains hand-rolled UI passes — only added lines count.
- A PR with no UI changes passes `ui-sonar-gate` rather than hanging on a check that never posts.
- `make ui-checkstyle-changed` reproduces the CI result locally, listing every failing gate.

#### Unit tests
Not applicable — no product code changed. This PR is CI configuration, node scripts and docs; it
contains **zero `.ts`/`.tsx` changes** by design.

Verified by execution instead:
- All four `reuse-audit` detectors fire with correct line numbers and library mappings.
- Untracked new files are detected — `git diff` alone reports nothing for them, which silently passed
  the exact case the guard exists for until fixed.
- **0 findings across the last 120 commits of `main`** while still matching 14 real instances in
  existing source: live detectors, no legacy blocking.
- First pass produced 6 findings, all false positives (Playwright `[role=]` selectors, test-double
  `<button>`s) — tests/mocks/stories/playwright excluded and a lookbehind added for bracket selectors.
- ESLint: **0 errors and 0 sonarjs violations repo-wide** across `src` and `playwright`; rules
  confirmed live against synthetic violations.
- All five `ui-sonar-gate` decision branches simulated (no UI changes → pass, gate green → pass, gate
  red → fail, jest failed → fail, push event → pass).
- Both workflows parse; `make harness-check` adds no new warnings.

#### Playwright (UI) tests
- [x] Not applicable — no runtime UI behaviour changes; this is CI/tooling only.

#### Manual testing performed
1. `make ui-checkstyle-changed` on a clean branch → exit 0.
2. Added `src/components/GateSim/X.tsx` containing `<div role="menu" />` → exit 1, `✖ ui-checkstyle
   failed: reuse-audit`, pointing at `Dropdown`.
3. Simulated the CI step with the real `BASE_SHA` → exit 1 with the exact payload posted to the PR.
4. `yarn lint` repo-wide with the new SonarJS rules → 0 errors.

#
### Follow-ups (deliberately not in this PR)
- **`.claude/**` harness changes** — the component-library reinvention table, `frontend-performance.md`,
  `frontend-a11y.md`, and the `PostToolUse` hook that runs `ui-guard.js`. Parked on branch
  `claude-harness-rules`. `scripts/ui-guard.js` ships here since it is useful standalone; the
  follow-up only adds the hook that invokes it.
- **Prettier 2.8.8 → 3.9.6** — parked on branch `prettier-3-upgrade` (463 mechanical files, kept out
  of this PR so it contains no `.ts`/`.tsx`). Prerequisite for `prettier-plugin-tailwindcss`.
- Semantic i18n audit — blocked on a design question: `sync-i18n --fn` fills new keys with English by
  design, so naive "identical to en-us" detection is 100% false positives.

<sub>Supersedes #30692, which GitHub auto-closed when this branch was renamed from `prettier-upgrade-scoped-reformat`. Same commits, same content.</sub>

This PR introduces frontend code-quality enforcement and supporting documentation.
- Adds a changed-line component-reuse audit with CI, local, and single-file modes.
- Enables sixteen SonarJS ESLint rules and adds the pinned plugin dependency.
- Adds an always-running required-job shim for the SonarCloud quality-gate result.
- Extends local changed-file checkstyle to run Tailwind and component-reuse audits.
- Updates generated API-reference counts and documents quality-gate administration.
</details>

The shell-command injection in the reuse audit's single-file mode should be fixed before merging.

A filename supplied to the new generation-time audit is incorporated into shell commands without safe argument passing, allowing a single quote in the filename to escape the intended pathspec.

**Files Needing Attention:** openmetadata-ui/src/main/resources/ui/scripts/reuse-audit.js and openmetadata-ui/src/main/resources/ui/scripts/ui-guard.js
</details>

The new single-file reuse audit passes a filename through shell command interpolation, allowing a crafted filename to execute commands when that mode is invoked. **How this was verified:** The `--file` argument is placed inside unescaped single quotes and concatenated into commands executed by `execSync`.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/scripts/reuse-audit.js | Adds the changed-line component-reuse detector, but its single-file mode shell-interpolates an untrusted filename. |
| openmetadata-ui/src/main/resources/ui/scripts/ui-guard.js | Adds generation-time orchestration for reuse and Tailwind audits and exposes filenames to the affected single-file audit path. |
| .github/workflows/yarn-coverage.yml | Propagates the scanner step outcome into an always-running Sonar quality-gate check. |
| .github/workflows/ui-checkstyle.yml | Integrates component-reuse auditing and reports its failures through the aggregate required check. |
| openmetadata-ui/src/main/resources/ui/eslint.config.mjs | Registers SonarJS and enables sixteen correctness rules. |
| openmetadata-ui/src/main/resources/ui/scripts/ui-checkstyle-changed.sh | Extends local changed-file checks to collect and report all three non-fixing audit failures. |
| docs/ui-code-quality-gate.md | Documents local enforcement, component reuse, Sonar configuration, and branch-protection requirements. |

</details>

```mermaid
flowchart LR
  Change[UI source change] --> Local[ui-checkstyle-changed]
  Change --> CI[ui-checkstyle workflow]
  Change --> Agent[ui-guard single-file mode]
  Local --> Audits[Tailwind and reuse audits]
  CI --> Audits
  Agent --> Audits
  Change --> Coverage[Jest coverage job]
  Coverage --> Sonar[SonarCloud PR scan]
  Sonar --> Gate[ui-sonar-gate required check]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["chore: keep the reuse guard out of pre-c..."](https://github.com/open-metadata/openmetadata/commit/24e55aca6446a7cab67d1333e40f96d7f275172d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48639791)</sub>


karanh37:
### Review findings addressed — 18533e36

All four automated findings were real and are fixed in `scripts/reuse-audit.js`.

| # | Finding | Fix | Verified by |
|---|---|---|---|
| 1 | `sh()` swallowed git failures → gate printed *"No new hand-rolled components"* and exited 0 without inspecting the diff | git failures throw; process exits 1 with the failing command | bad base ref → exit 1 + diagnostic |
| 2 | CI ran from repo root, local from the UI app → bare `'*.tsx'` made CI scan the whole repository | pathspec pinned to `:(top)openmetadata-ui/.../src`, identical from any cwd | repo-root and UI-dir runs return identical detections; non-UI `.ts` adding `ArrowDown` now out of scope |
| 3 | `data-role="menu"` matched (`\b` also matches after `-`) | requires whitespace or `<` before `role=` | `data-role`/`user-role` → exit 0; real `role=` → exit 1 |
| 4 | **P1** `--file` interpolated into a shell string → command injection, and `--file` is fed by the PostToolUse hook | all git calls use `execFileSync` with an argument array, no shell | filename containing `$(touch /tmp/PWNED)` executes nothing |

Finding 1 deserves calling out: it was the same silent-pass failure mode this PR removes from the Sonar check, sitting in the new guard itself.

Fixing 4 surfaced a further bug of my own — `git ls-files` returns **cwd-relative** paths, which broke the untracked-file read once cwd could be the UI app. `--full-name` pins them repo-root-relative, matching `git diff`'s `+++ b/` form.

Thanks — these were good catches.

chirag-madlani (approved):

PR #30710, opened 2026-07-30
Fixes #30709: fix(ingestion): build a deferred model's schema on first instantiation

### LINKED ISSUE #30709: Ingestion: deferred models break serialization with "'MockValSer' object cannot be converted to 'SchemaSerializer'"
### Affected module

Ingestion Framework

### Describe the bug

Since #30377 (`defer_build=True` on the generated-model base class), any `model_dump()` that reaches a nested model whose own schema was never built dies with:

```
TypeError: 'MockValSer' object cannot be converted to 'SchemaSerializer'
```

Every Snowflake CLI ingestion hits it. `MetadataWorkflow._get_source` calls `self.config.source.model_dump()` (`ingestion/src/metadata/workflow/metadata.py:50`), and the workflow aborts before it does any work:

```
[2026-07-30 14:51:16] INFO     {metadata.OMetaAPI:server_mixin:98} - OpenMetadata client running with Server version [2.0.0] and Client version [2.0.0.0]
[2026-07-30 14:51:16] ERROR    {metadata.Utils:logger:222} -
Error initializing metadata: 'MockValSer' object cannot be converted to 'SchemaSerializer'
```

Root cause: `defer_build` means a nested model's schema only ever gets **inlined into its parent**. Validating `Source` therefore never touches `SnowflakeConnection`'s own validator or serializer, so that class keeps the `MockValSer` that pydantic puts on it in place of a real serializer (`_mock_val_ser.set_model_mocks`). When serialization falls back to type inference for a union member, pydantic-core reads `type(value).__pydantic_serializer__` **directly off the class** — a plain attribute lookup that never goes through `MockValSer.__getattr__`, so the lazy rebuild that would normally repair this never fires. pydantic-core then fails to convert the mock.

Four instantiated-but-unbuilt classes are involved in the Snowflake case: `SnowflakeConnection`, `DatabaseConnection`, `SourceConfig`, `DatabaseServiceMetadataPipeline`. Force-building any one of them individually still fails; building all four makes the dump succeed. Not connector-specific plumbing — it is any instantiated model whose class was never built. Postgres happens not to trip it.

Not a pydantic version issue: reproduced on 2.11.10, 2.11.9, 2.11.7 and 2.10.6.

### To Reproduce

No server or credentials needed:

```python
from metadata.generated.schema.metadataIngestion.workflow import Source

cfg = {
    "type": "snowflake",
    "serviceName": "test",
    "serviceConnection": {
        "config": {
            "type": "Snowflake",
            "account": "account",
            "username": "username",
            "password": "password",
            "warehouse": "warehouse",
        }
    },
    "sourceConfig": {"config": {"type": "DatabaseMetadata"}},
}
Source.model_validate(cfg).model_dump()  # TypeError
```

`OM_PYDANTIC_DEFER_BUILD=false` makes it pass, which confirms the source.

### Expected behavior

`model_dump()` works on a validated workflow config, and `metadata ingest -c <snowflake config>` runs — with `defer_build` left on by default.

### OS

macOS 15.5 (Darwin 25.5.0)

### Python version

3.11.13

### OpenMetadata version

2.0.0

openmetadata-ingestion==2.0.0.0.dev0 (built from `main`)

### Additional context

`ingestion/tests/unit/test_generated_models_defer_build.py` checks buildability (`model_rebuild(force=True)`) and the env toggle, but never dumps a parent whose nested child class was never built — which is the entire failure mode, hence the gap in CI.

Workaround until fixed: `export OM_PYDANTIC_DEFER_BUILD=false` (the kill-switch #30377 shipped), at the cost of the import-time RSS saving.

pmbrull:
### Describe your changes:

Fixes #30709

`defer_build=True` (#30377) means a nested model's schema only ever gets **inlined into its parent**, so the nested class keeps the `MockValSer` that pydantic puts on it in place of a real serializer (`_mock_val_ser.set_model_mocks`). Pydantic's serialization fallbacks read `type(value).__pydantic_serializer__` straight off the class — a plain attribute lookup that never goes through `MockValSer.__getattr__`, so the lazy rebuild that would normally repair this never fires. pydantic-core then fails to convert the mock:

```
TypeError: 'MockValSer' object cannot be converted to 'SchemaSerializer'
```

That takes out **every Snowflake CLI ingestion**. `MetadataWorkflow._get_source` calls `self.config.source.model_dump()`, and `SnowflakeConnection`, `DatabaseConnection`, `SourceConfig` and `DatabaseServiceMetadataPipeline` are all instantiated there without ever being built, so the workflow dies at init before doing any work.

I build the schema in `model_post_init` when the class is still incomplete. An instantiated model is a used model, and importing a module instantiates nothing, so #30377's import-time saving survives — the deferral just now ends at first use rather than first *direct* use.

`_parent_namespace_depth=0` is deliberate: the default (2) would resolve forward refs against whichever frame happened to instantiate the model instead of against the model's own module.

#
### Type of change:

- [x] Bug fix

#
### Checklist:

- [x] My PR title is following the [convention](https://github.com/open-metadata/OpenMetadata/blob/main/CONTRIBUTING.md).
- [x] I have added tests.

#
### Verification

Peak RSS over a Snowflake workflow source plus 200 `Table` validate/dump cycles:

| config | `Source.model_dump()` | maxrss |
|---|---|---|
| `main` | **TypeError** | 202 MB |
| this PR | OK | **203 MB** |
| `OM_PYDANTIC_DEFER_BUILD=false` | OK | 305 MB |

So the fix costs ~1 MB and keeps ~100 MB of #30377's win. `Table.__pydantic_complete__` is still `False` after an import-only run, i.e. deferral is intact.

Also checked by hand: `metadata ingest -c <snowflake config>` now gets past init to authentication instead of dying on the dump, and a recursive `Table` → `Column` → `children` validate + `model_dump` + `model_dump_json` round-trips, so rebuilding inside `model_post_init` while the parent's validator is running is not re-entrant-hostile.

`tests/unit/test_generated_models_defer_build.py`, `test_pydantic_v2.py`, `test_workflow_parse.py`, `test_workflow_parse_example_config.py` and `test_filter_pattern.py` pass.

#
### Test gap this closes

The existing guards check buildability (`model_rebuild(force=True)`) and the env toggle, but never dump a parent whose nested child class was never built — which is the entire failure mode, so CI was green while every Snowflake ingestion was broken. The new `test_deferred_nested_models_are_serializable` does exactly that, in a fresh interpreter with the toggle unset. Verified that it fails on `main` with the `TypeError` above.

#
### Note for reviewers

- **Thread safety.** ~~First instantiation from the multithreaded topology/profiler paths can race on `model_rebuild`.~~ Addressed in 7cb62713e6 — the race turned out to be real and pre-existing, see the comment below.
- **Error handling.** I deliberately left the rebuild *outside* the existing `try`, so a genuine build failure raises loudly instead of degrading into the opaque `MockValSer` error. `test_all_generated_models_are_buildable` already guarantees every generated model builds.

Fixes deferred Pydantic model serialization while retaining import-time schema deferral.
- Rebuilds incomplete model schemas on first instantiation.
- Serializes schema rebuilds with a reentrant lock to protect concurrent and nested builds.
- Adds regression coverage for nested serialization, configuration toggling, buildability, and concurrent first use.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/models/custom_pydantic.py | Adds locked schema rebuilding on first model instantiation to make deferred nested models serializable. |
| ingestion/tests/unit/test_generated_models_defer_build.py | Adds behavioral regression tests for nested serialization and concurrent deferred-model initialization. |

</details>

```mermaid
sequenceDiagram
  participant Caller
  participant Model as Deferred Model
  participant Lock as Rebuild RLock
  participant Pydantic
  Caller->>Model: First validation/instantiation
  Pydantic->>Model: model_post_init()
  Model->>Model: Check __pydantic_complete__
  alt Schema incomplete
    Model->>Lock: Acquire
    Model->>Pydantic: "model_rebuild(depth=0)"
    Pydantic-->>Model: Validator and serializer ready
    Model->>Lock: Release
  end
  Model-->>Caller: Serializable model instance
```
</details>

<sub>Reviews (3): Last reviewed commit: ["docs(ingestion): correct where a deferre..."](https://github.com/open-metadata/openmetadata/commit/eb86e90318665f46f98d1f85cda079bc6bbeee1e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48686702)</sub>

**Context used:**

### DIFF
--- a/ingestion/src/metadata/ingestion/models/custom_pydantic.py
+++ b/ingestion/src/metadata/ingestion/models/custom_pydantic.py
@@ -37,8 +37,12 @@ SECRET = "secret:"
 JSON_ENCODERS = "json_encoders"
 
 # model_rebuild() deletes __pydantic_core_schema__/__pydantic_validator__/__pydantic_serializer__
-# before regenerating them, so concurrent first uses of the same class can observe the inherited
-# MockValSer or lose the delete race outright ("AttributeError: __pydantic_core_schema__").
+# before regenerating them, and says so itself: "as model_rebuild() isn't thread-safe, concurrent
+# model instantiations can lead to the parent validator being used". Two ways that bites us. A
+# first build only deletes the core schema, since the validator and serializer mocks are skipped
+# by pydantic's isinstance(..., MockValSer) guard, so a racing thread loses the delete outright
+# ("AttributeError: __pydantic_core_schema__"). A force rebuild of an already built class deletes
+# the real objects, so a racing thread falls through the MRO onto PydanticBaseModel's own mocks.
 # Reentrant because building one schema can rebuild the nested models it references.
 _MODEL_REBUILD_LOCK = threading.RLock()
 
@@ -78,11 +82,12 @@ class BaseModel(PydanticBaseModel):
         but a FilterPattern object is required in the generated code.
         """
         # Build the schema of a model the first time one is instantiated. With
-        # defer_build a nested model's schema only ever gets inlined into its parent,
-        # so the class itself keeps the MockValSer it inherits from pydantic.BaseModel.
-        # Serialization fallbacks read type(value).__pydantic_serializer__ straight off
-        # the class, which does not go through MockValSer.__getattr__ and so never
-        # triggers its lazy rebuild; pydantic-core then fails to convert the mock with
+        # defer_build a nested model's schema only ever gets inlined into its parent, so
+        # the class keeps the MockValSer that pydantic put on it in place of a real
+        # serializer (see _mock_val_ser.set_model_mocks). Serialization fallbacks read
+        # type(value).__pydantic_serializer__ straight off the class, which does not go
+        # through MockValSer.__getattr__ and so never triggers its lazy rebuild;
+        # pydantic-core then fails to convert the mock with
         # "'MockValSer' object cannot be converted to 'SchemaSerializer'". Building here
         # keeps the import-time saving, since importing a module instantiates nothing.
         # _parent_namespace_depth=0: forward refs must resolve against the model's own
--- a/ingestion/src/metadata/ingestion/models/custom_pydantic.py
+++ b/ingestion/src/metadata/ingestion/models/custom_pydantic.py
@@ -58,6 +58,20 @@ class BaseModel(PydanticBaseModel):
         This is needed because dict is defined in the JSON schema for the FilterPattern field,
         but a FilterPattern object is required in the generated code.
         """
+        # Build the schema of a model the first time one is instantiated. With
+        # defer_build a nested model's schema only ever gets inlined into its parent,
+        # so the class itself keeps the MockValSer it inherits from pydantic.BaseModel.
+        # Serialization fallbacks read type(value).__pydantic_serializer__ straight off
+        # the class, which does not go through MockValSer.__getattr__ and so never
+        # triggers its lazy rebuild; pydantic-core then fails to convert the mock with
+        # "'MockValSer' object cannot be converted to 'SchemaSerializer'". Building here
+        # keeps the import-time saving, since importing a module instantiates nothing.
+        # _parent_namespace_depth=0: forward refs must resolve against the model's own
+        # module, not against whichever frame happened to instantiate it.
+        cls = type(self)
+        if not cls.__pydantic_complete__:
+            cls.model_rebuild(_parent_namespace_depth=0)
+
         # pylint: disable=import-outside-toplevel
         try:
             if not self.__class__.__name__.endswith("Connection"):
--- a/ingestion/src/metadata/ingestion/models/custom_pydantic.py
+++ b/ingestion/src/metadata/ingestion/models/custom_pydantic.py
@@ -19,6 +19,7 @@ be self-sufficient with only pydantic at import time.
 import json
 import logging
 import os
+import threading
 from typing import Any, Callable, Dict, Literal, Optional, Union  # noqa: UP035
 
 from pydantic import BaseModel as PydanticBaseModel
@@ -34,6 +35,12 @@ logger = logging.getLogger("metadata")
 
 SECRET = "secret:"
 JSON_ENCODERS = "json_encoders"
+
+# model_rebuild() deletes __pydantic_core_schema__/__pydantic_validator__/__pydantic_serializer__
+# before regenerating them, so concurrent first uses of the same class can observe the inherited
+# MockValSer or lose the delete race outright ("AttributeError: __pydantic_core_schema__").
+# Reentrant because building one schema can rebuild the nested models it references.
+_MODEL_REBUILD_LOCK = threading.RLock()
 
 
 class BaseModel(PydanticBaseModel):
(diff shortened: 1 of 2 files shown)


pmbrull:
Reviewed 7cb62713e6 — the lock is right, and I verified the reasoning against pydantic 2.11.10 rather than taking it on faith:

- `model_rebuild()` does `delattr` the three attributes before regenerating, and pydantic's own inline comment there says *"as `model_rebuild()` isn't thread-safe, concurrent model instantiations can lead to the parent validator being used"*. The premise is pydantic-endorsed, not speculative.
- The claim that pydantic's lazy repair funnels through the override holds: `_mock_val_ser.set_model_mocks` builds `attempt_rebuild_fn` around `cls.model_rebuild(raise_errors=False, _parent_namespace_depth=5)`, a bound lookup, so the MRO hits our override.
- Depth accounting is right on both routes. Default `2` resolves to the caller, the mock's `5` to its caller, and the override inserts exactly one frame into each, so `+1` is universally correct. The `if _parent_namespace_depth > 0` guard mirrors pydantic's own `elif _parent_namespace_depth > 0`.

I pushed eb86e90318 on top, comments only. Two corrections:

**Where the mock comes from.** I had written that a deferred class keeps the `MockValSer` it *inherits* from `pydantic.BaseModel`. It does not — `set_model_mocks(cls)` puts one on the class itself:

```
"__pydantic_serializer__" in SnowflakeConnection.__dict__                          -> True
SnowflakeConnection.__pydantic_serializer__ is BaseModel.__pydantic_serializer__   -> False
```

The mechanism is unchanged, only the provenance. My wording, my mistake — it had also leaked into the issue and the PR description, both now fixed.

**The lock covers two different races, not one.** A *first* build deletes only the core schema, because pydantic skips the validator and serializer under its `isinstance(..., MockValSer)` guard. So a racing thread loses that one delete, which is why `test_concurrent_first_instantiation_is_safe` reports `AttributeError: type object 'Parent' has no attribute '__pydantic_core_schema__'` rather than anything mock-shaped. Only a *force* rebuild of an already built class deletes the real objects and leaves a racing thread falling through the MRO onto `PydanticBaseModel`'s mocks. Both are real; splitting them in the comment makes the test's failure mode legible.

Two things worth recording for whoever reviews this:

**The race predates this PR.** `test_concurrent_first_instantiation_is_safe` fails on `Parent`, which is validated directly and therefore repaired through pydantic's own `MockValSer.attempt_rebuild` — nothing to do with the `model_post_init` build added in da2387e2ee. Every deferred model has been exposed to this since #30377; the lock closes a pre-existing hole rather than paying for this PR.

**Lock ordering against the import lock.** A thread holding `_MODEL_REBUILD_LOCK` that triggers an import, while another thread is mid-import of a generated module whose module scope calls `Column.model_rebuild()`, is the classic deadlock shape. I think we are fine — schema generation resolves forward refs out of already-populated namespaces instead of importing, and 3.11 uses per-module import locks — but it is the one part I would not want to sign off alone.

### Verification

- `tests/unit/test_generated_models_defer_build.py`: 5 passed.
- Removed the `model_rebuild` override and reran the concurrency test: **17 failures out of 1600 runs**, all `AttributeError: type object 'Parent' has no attribute '__pydantic_core_schema__'`. The test fails without the lock, so it is guarding something real.
- `ruff format --check` and `ruff check` clean. `str | None` in the new helper is fine under `requires-python = ">=3.10"` / `target-version = "py310"`.
- Peak RSS re-measured against the repo source (Snowflake workflow source plus 200 `Table` validate/dump cycles), so these supersede the numbers in the PR description, which came from an installed build:

| config | `Source.model_dump()` | maxrss |
|---|---|---|
| `main` | **TypeError** | 286 MB |
| this branch | OK | **285 MB** |
| `OM_PYDANTIC_DEFER_BUILD=false` | OK | 449 MB |

The lock costs nothing measurable, and #30377's saving is intact.

One detail in the new test worth keeping if anyone refactors it: it exercises both rebuild routes at once — `Parent` is validated directly and repaired via `attempt_rebuild`, while `Nested` is only ever reached through `model_post_init` — and `sys.setswitchinterval(1e-6)` is what makes a few-bytecode window reproduce at all.

IceS2 (approved):

pmbrull:
picked to 2.0

PR #30716, opened 2026-07-30
Fixes 30715: reject bodiless clauses in persona AI context rule filters

### LINKED ISSUE #30715: Persona AI context rule preview fails with "Unexpected JSON event 'END_OBJECT' instead of 'KEY_NAME'"
### Describe the bug

Editing a Persona AI-context rule in **Settings → Persona → AI Context** fails as soon as a query-builder row has a field and an operator selected but no value entered yet.

The rule editor auto-previews on a 300 ms debounce, so just pausing mid-edit fires `POST /api/v1/personas/{id}/aiContext/rules/preview` and it returns a 500.

Two defects compound here.

**1. The query builder emits a bodiless clause.**

`buildEsRule` builds `{[queryType]: {...parameters}}` from parameters that are still `undefined`. `JSON.stringify` drops `undefined` values, so the stored `queryFilter` contains `{"term":{}}` (likewise `match`, `range`, `wildcard`). The top-level `isEmpty(data)` guard in `QueryBuilderWidgetV1` only inspects the outermost object, so the poisoned clause is sent and persisted.

Both engines reject a bodiless field-keyed clause. Reproduced directly against the clients:

```
opensearch-java 2.26.0   {"term":{}} -> UnexpectedJsonEventException: Unexpected JSON event 'END_OBJECT' instead of 'KEY_NAME'
elasticsearch-java 9.2.4 {"term":{}} -> JsonpMappingException: ... Unexpected JSON event 'END_OBJECT' instead of 'KEY_NAME' (JSON path: term)
```

`{"bool":{}}`, `{"bool":{"must":[]}}`, `{"match_all":{}}` and `{"match_none":{}}` all parse fine — only field-keyed clauses break.

**2. The resulting error is masked.**

`PersonaContextBuilder.search` wraps the failure and then calls `initCause()`:

```java
ServiceUnavailableException unavailable = new ServiceUnavailableException("Failed to evaluate persona context rule '" + rule.getName() + "': " + exception.getMessage());
unavailable.initCause(exception);
```

`ServiceUnavailableException(String)` routes through `WebApplicationException(message, null, response)` → `super(message, cause)` with `cause == null`. That *sets* the cause to null instead of leaving it unset, so `initCause()` always throws:

```
java.lang.IllegalStateException: Can't overwrite cause with java.io.IOException: Failed to parse query filter: Unexpected JSON event 'END_OBJECT' instead of 'KEY_NAME'
	at java.base/java.lang.Throwable.initCause(Throwable.java:464)
	at org.openmetadata.service.aicontext.PersonaContextBuilder.search(PersonaContextBuilder.java:272)
	at org.openmetadata.service.aicontext.PersonaContextBuilder.preview(PersonaContextBuilder.java:300)
	at org.openmetadata.service.resources.teams.PersonaResource.previewAiContextRule(PersonaResource.java:369)
```

This is unconditional, not a race — every search failure in persona context materialization is converted to a 500 with the original stack trace dropped from the cause chain.

### To Reproduce

1. Settings → Persona → pick a persona → AI Context → add a context rule.
2. In the rule's query builder, pick a field and an operator. Do not enter a value.
3. Wait ~300 ms. The preview call returns 500.
4. Save the rule. `GET /api/v1/personas/{id}/aiContext/document` and the `get_persona_context` MCP tool now fail too, because `selectRules` evaluates every rule and one bad rule fails the whole build.

### Expected behavior

An incomplete query-builder row should contribute no constraint, exactly as the query builder itself renders it — not fail the rule, the persona context document, and the MCP tool. And when a search genuinely does fail, the API should return its real cause rather than an `IllegalStateException` from the error-handling path.

### Version

- OpenMetadata version: 2.0.0
- Search: OpenSearch (also reproducible on Elasticsearch)

pmbrull:
### Describe your changes:

Fixes #30715

A persona AI-context rule whose Explore query-builder row has a field and an operator but no value serialized to `{"term":{}}`. Both OpenSearch and Elasticsearch reject a bodiless field-keyed clause outright, so evaluating the rule failed with `Failed to parse query filter: Unexpected JSON event 'END_OBJECT' instead of 'KEY_NAME'`. The editor auto-previews on a 300 ms debounce, so pausing mid-edit was enough to trigger it, and once the rule was saved the same filter broke the whole persona context document and the `get_persona_context` MCP tool — `selectRules` evaluates every rule, so one bad rule fails the entire build.

**Approach: fix it on the UI, don't sanitize on the backend.** The empty filter must never be emitted or saved in the first place — that is a UI concern. The backend does *not* try to clean up query filters: ES/OS already validate queries, so re-implementing that as recursive JSON pruning in the service is duplicated work. If a malformed filter ever does reach the backend, ES/OS reject it and the failure now surfaces with its real cause instead of being masked.

**UI — don't let an empty filter through**

- `buildEsRule` stops emitting a rule whose values are all `undefined` (this also covers the nested custom-property path, which builds its clause separately), and the multiselect branch filters out the resulting holes instead of serializing them as `null`.
- The rule editor refuses the save while a condition is unfinished and says so inline. A rule built as "tier is ___" would otherwise persist as *no filter at all* — matching everything and pulling the entity cap instead of the handful of assets the user picked.
- `hasUnfinishedRule` answers that by asking `elasticSearchFormat` what a row actually produces, so the guard and `buildEsRule`'s drop decision cannot drift apart. A group with no conditions stays valid — "empty selects every entity of the configured type" is documented behaviour.
- `QueryBuilderWidgetV1` exposes this through a new optional `onValidityChange` prop. Only the persona rule editor opts in, so Explore and the other query-builder consumers are untouched. The message appears on the save attempt rather than the moment a row is added.

**Backend — stop masking a real search failure**

- `PersonaContextBuilder.search` passes the cause through the `(String, Response, Throwable)` constructor instead of calling `initCause()`, so the 503 keeps its real cause. `ServiceUnavailableException(String)` routes through `WebApplicationException(message, null, response)` → `super(message, cause)` with `cause == null`, which *sets* the cause instead of leaving it unset, so the following `initCause()` always threw `IllegalStateException: Can't overwrite cause with ...` — every search failure surfaced as a 500 with the original stack trace dropped.

> The earlier revision of this PR also pruned bodiless clauses out of the requested filter in the service. That was removed after review — the backend should not re-do validation ES/OS already perform, and sanitization belongs on the UI.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

The UI guard is deliberately a pure function over the same tree and config the emitted filter is built from, rather than a call into react-awesome-query-builder's own validation (`QbUtils.isValidTree`).

Two reasons. It cannot disagree with `buildEsRule` about what counts as unfinished, since it asks `elasticSearchFormat` rather than reimplementing the rule. And it is unit-testable: every jest suite in this repo mocks RAQB wholesale (`jest.mock('@react-awesome-query-builder/antd')` in `QueryBuilderWidgetV1.test.tsx`), so anything reaching into RAQB's config or validation can only be verified in a browser. Building RAQB's extended config standalone does not work either — OM's `getTreeConfig` output is not complete until `<Query>` extends it, so `ConfigUtils.extendConfig` throws on it.

#
#### Use cases covered

- Admin picks a field for a condition, leaves the value empty, hits Save → save is blocked with an inline message; no `POST /aiContext/rules` is fired.
- Admin saves a rule with **no filter conditions at all** → still saves.
- Admin removes the unfinished condition → the error clears and the rule saves.
- A search failure that is not the caller's fault (engine unavailable) surfaces as a 503 carrying the underlying exception as its cause.

#### Unit tests

Backend — `openmetadata-service/.../aicontext/PersonaContextBuilderTest.java`. `searchSurfacesTheUnderlyingSearchFailureWithItsCause` fails on `main` with the exact production `IllegalStateException`.

```
PersonaContextBuilderTest    tests=12 errors=0 skipped=0 failures=0
```

UI — `src/utils/QueryBuilderElasticsearchFormatUtils.test.js`, 12 tests: 4 covering the emitted filter (unentered rule, bodiless clause, multiselect null hole, entered-value regression guard) and 8 covering `hasUnfinishedRule` (unentered value, no field picked, multiselect with no option, fully entered, **group with no conditions**, undefined tree, unfinished nested in a group, all-entered group).

- [x] Not applicable (no API contract change — same endpoints, same shapes).

#### Playwright (UI) tests

- [x] Added to `playwright/e2e/Features/PersonaAIContext.spec.ts`:
  - `blocks saving a rule whose condition has no value entered`
  - `saves a rule that has no filter conditions at all`
  - `clears the filter error once the unfinished condition is removed`

These drive the real query builder end-to-end, which jest cannot (see High-level design). The middle one is the important guard: if the check ever starts rejecting an empty filter, saving a "match everything" rule breaks, and that is a worse regression than the bug being fixed.

#### Manual testing performed

The UI behaviour is covered by the Playwright cases above.

#
### i18n

`message.persona-context-rule-filter-incomplete` added and synced with `yarn i18n`. All 19 non-English locales carry a real translation rather than the English placeholder `yarn i18n` inserts. Note `pr-pr` is Persian, not Portuguese.

#

Not attached — no local stack available to record against. The three Playwright cases document the behaviour end-to-end instead.

#
### Follow-up (not in this PR)

`PersonaContextBuilder.selectRules` still evaluates rules unguarded, so any future rule-level failure takes down the entire persona context document and the MCP tool rather than degrading that one rule. Isolating per rule is a behaviour change worth its own discussion — happy to open a separate issue.

#
### Checklist:


Rohit0301 (dismissed):

mohityadav766 (approved):

Rohit0301 (approved):

pmbrull:
picked to 2.0

PR #30719, opened 2026-07-30
Fixes 30715: block saving a persona AI context rule with an unfinished condition

pmbrull:
### Describe your changes:

Relates to #30715. **Stacked on #30716** — base is that PR's branch, so this diff shows only the follow-up. Retarget to `main` once #30716 merges.

#30716 stopped a half-built condition from crashing the search: `{"term":{}}` is rejected outright by both OpenSearch and Elasticsearch, so the emitted `queryFilter` now drops it. That fixes the 500, but leaves a worse-shaped UX bug behind it — dropping the condition *silently on save*. A rule the user built as "tier is ___" persists as **no filter at all**, so it matches everything and pulls the entity cap instead of the handful of assets they picked. Nothing tells them.

So the editor now refuses the save while a condition is unfinished, and says so inline, rather than quietly widening the rule. Either the condition is in, or it is out.

`QueryBuilderWidgetV1` reports tree validity through a new optional `onValidityChange` prop. It uses `QbUtils.isValidTree` with the config react-awesome-query-builder has already extended in its own `onChange` — the same validity RAQB uses internally, so operators with no value (`is_null` and friends) are not false-flagged. Only the persona rule editor opts in; Explore and the other query-builder consumers are untouched.

The backend prune from #30716 stays regardless. It is the safety net, not the UX: rules with `{"term":{}}` are already persisted in `persona.contextDefinition` today and a UI fix does not heal them, and `POST /aiContext/rules` is a public API that must not 500 on bad input.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered

- Admin picks a field for a condition, leaves the value empty, hits Save → save is blocked, an inline message explains why, and no `POST /aiContext/rules` is fired.
- Admin saves a rule with **no filter conditions at all** → still saves. This is the regression guard, see the note below.
- Admin removes the unfinished condition → the error clears and the rule saves.

#### Unit tests

- [x] Not applicable to the validity check itself — see the honest note below.
- Existing jest suites re-run for regressions: `QueryBuilderElasticsearchFormatUtils`, `AdvancedSearchClassBase`, `RuleQueryBuilderField`, `AdvancedAssetsFilterField`, `QueryBuilderWidget`, `QueryBuilderWidgetV1`, `ContractSemanticFormTab`, and every `PersonaAIContext` suite.

```
Test Suites: 13 passed, 13 total
Tests:       147 passed, 147 total
```

#### Playwright (UI) tests

- [x] Added to `openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/PersonaAIContext.spec.ts`:
  - `blocks saving a rule whose condition has no value entered`
  - `saves a rule that has no filter conditions at all`
  - `clears the filter error once the unfinished condition is removed`

**Why Playwright and not jest — please read before reviewing.** Every jest suite in this repo mocks react-awesome-query-builder wholesale (`jest.mock('@react-awesome-query-builder/antd')` in `QueryBuilderWidgetV1.test.tsx`), so `QbUtils` is a stub there and a unit test would assert nothing about real `isValidTree` behaviour. Building the real config standalone does not work either — OM's `getTreeConfig` output is not complete until RAQB's `<Query>` extends it, so `ConfigUtils.extendConfig` throws on it. Playwright is the only place the real query builder actually runs.

**The second test is the load-bearing one.** `isValidTree` is RAQB's own validity check and I could not verify locally how it treats a tree with no rules. If it reports an empty tree as invalid, this change would block saving every "match everything" rule — which the schema documents as legal ("Empty selects every entity of the configured type") and which would be a worse regression than the bug being fixed. That test exists to fail loudly in CI if that is the case. I have deliberately not run it locally; **please watch that job.** If it goes red, the fix is to gate the check on the tree being non-empty, not to loosen the test.

- [x] Not applicable (no backend change in this PR).

- [x] Not applicable.

#### Manual testing performed

None — flagged deliberately. The local stack was unavailable, which is exactly why the verification is pushed into the Playwright job rather than asserted here.

Lint and type-check were run:

- `eslint` on the changed spec — 0 errors, 0 warnings.
- `yarn lint:playwright` — 0 errors (280 pre-existing warnings elsewhere in the suite, none in the changed file).
- The changed src files pass the CI organize-imports → eslint → prettier sequence with no diff.
- `tsc --noEmit` reports one error in `QueryBuilderWidgetV1.test.tsx:297`, confirmed pre-existing on the base commit and untouched by this change.

#

Not attached — no local stack available to record against. The three Playwright cases document the behaviour end-to-end instead.

#
### i18n

`message.persona-context-rule-filter-incomplete` added and synced with `yarn i18n`. All 19 non-English locales carry a real translation rather than the English placeholder `yarn i18n` inserts. Note `pr-pr` is Persian, not Portuguese.

#
### Checklist:
- [x] My PR is linked to a GitHub issue.
- [x] For UI changes: recording explained above.

The PR prevents saving unfinished persona AI-context filters.
- Adds query-tree validity reporting to QueryBuilderWidgetV1 and forwards it through RuleQueryBuilderField.
- Blocks persona context-rule submission and displays an inline localized error while the filter is invalid.
- Adds Playwright coverage for unfinished, empty, and subsequently removed conditions.
- Adds the new validation message to all supported locale files.
</details>

The validity lifecycle needs to be fixed before merging because valid rules can remain blocked after resets while invalid persisted rules can bypass the guard without an edit.

Validity is retained after the displayed filter is reset, but it is not initialized when an existing filter tree loads, causing both false save rejection and missed validation.

**Files Needing Attention:** openmetadata-ui/src/main/resources/ui/src/components/Settings/Persona/PersonaAIContext/ContextRuleEditor/ContextRuleEditor.component.tsx; openmetadata-ui/src/main/resources/ui/src/components/common/QueryBuilderWidgetV1/QueryBuilderWidgetV1.tsx
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/src/components/Settings/Persona/PersonaAIContext/ContextRuleEditor/ContextRuleEditor.component.tsx | Adds save blocking and inline feedback, but retains stale invalidity across form resets and entity-type changes. |
| openmetadata-ui/src/main/resources/ui/src/components/Settings/Persona/PersonaAIContext/ContextRuleEditor/RuleQueryBuilderField.component.tsx | Forwards the optional validity callback to the shared query-builder widget. |
| openmetadata-ui/src/main/resources/ui/src/components/common/QueryBuilderWidgetV1/QueryBuilderWidgetV1.tsx | Reports validity only during query-builder changes, leaving initially loaded invalid trees unreported. |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/PersonaAIContext.spec.ts | Adds end-to-end coverage for blocking unfinished conditions while preserving empty-rule saves and recovery after deletion. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(ui): block saving a persona rule wit..."](https://github.com/open-metadata/openmetadata/commit/4b87fe819b11ed2ebe5cb6b50ad6385f1496cfb9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48729019)</sub>

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/PersonaAIContext.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/PersonaAIContext.spec.ts
@@ -1102,4 +1102,125 @@ test.describe.serial('Persona AI Context', () => {
     await expect(adminPage.getByRole('dialog')).toBeVisible();
     await expect(adminPage.getByText('7 truncated')).toBeVisible();
   });
+
+  // A condition with a field but no value serializes to `{"term":{}}`, which the search engines
+  // reject outright. The emitted filter now drops it, so saving such a rule would silently widen it
+  // to match everything — the editor has to stop that at the source. See issue #30715.
+  test('blocks saving a rule whose condition has no value entered', async ({
+    adminPage,
+  }) => {
+    await mockPersonaContextApi(
+      adminPage,
+      persona.responseData.id as string,
+      []
+    );
+    await openPersonaContext(adminPage);
+
+    await adminPage.getByTestId('empty-add-context-rule').click();
+    await adminPage.getByTestId('context-rule-name').fill('Unfinished filter');
+    await adminPage.getByTestId('add-context-condition').click();
+
+    const condition = adminPage.getByTestId('delete-condition-button');
+    await expect(condition).toHaveCount(1);
+
+    const fieldSelect = adminPage
+      .getByRole('dialog')
+      .locator('.rule--field .ant-select')
+      .first();
+    await fieldSelect.click();
+    await adminPage
+      .locator('.ant-select-dropdown:visible .ant-select-item-option')
+      .filter({ hasText: 'Display Name' })
+      .first()
+      .click();
+
+    let saveRequested = false;
+    await adminPage.route('**/aiContext/rules', async (route) => {
+      saveRequested = true;
+      await route.abort();
+    });
+
+    await adminPage.getByRole('button', { name: 'Save Rule' }).click();
+
+    await expect(
+      adminPage.getByTestId('context-rule-filter-error')
+    ).toBeVisible();
+    await expect(adminPage.getByTestId('form-heading')).toBeVisible();
+    expect(saveRequested).toBe(false);
+  });
+
+  // Guards the other side of the same check: "Empty selects every entity of the configured type" is
+  // documented behaviour, so a rule with no filter at all must stay saveable.
+  test('saves a rule that has no filter conditions at all', async ({
+    adminPage,
+  }) => {
+    await mockPersonaContextApi(
+      adminPage,
+      persona.responseData.id as string,
+      []
+    );
+    await openPersonaContext(adminPage);
+
+    await adminPage.getByTestId('empty-add-context-rule').click();
+    await adminPage.getByTestId('context-rule-name').fill('Every table');
+
+    const createRuleRequest = adminPage.waitForRequest(
+      (request) =>
+        request.url().endsWith('/aiContext/rules') &&
+        request.method() === 'POST'
+    );
+    await adminPage.getByRole('button', { name: 'Save Rule' }).click();
+
+    const createdRule = (await createRuleRequest).postDataJSON() as ContextRule;
+    expect(createdRule).toMatchObject({ name: 'Every table', queryFilter: '' });
+    await expect(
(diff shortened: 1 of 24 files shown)


pmbrull:
Folded into #30716 — the UI guard belongs in the same PR as the backend fix, and the guard itself was reworked to be deterministic and unit-testable (see below). Closing this stacked PR.

PR #30763, opened 2026-07-31
fix(ui): a 401 should only sign the user out when the session is dead

aniketkatkar97:
Fixes open-metadata/openmetadata-collate#4647

Companion to that issue's backend fix, open-metadata/openmetadata-collate#5422. This is the UI half and stands on its own.

## Problem

The shared axios response interceptor treats **every** 401 as a dead session. It calls `TokenService.refreshToken()` and reads a falsy return as "refresh failed" → `resetUserDetails(true)` → forced logout to `/signin`.

The catch is one layer down. `refreshToken()` only renews when the stored token is actually expiring:

```ts
// TokenServiceUtil.ts
if (isExpired || timeoutExpiry <= 0) { ...renew... }
else { this.clearRefreshInProgress(); return null; }   // <- falsy
```

So **a 401 arriving while our own token is still valid never triggers a refresh at all — it goes straight to a forced logout.** Any 401 an endpoint raises for its own reasons (most often a service the API depends on rejecting *its* credentials) destroys a healthy session and replaces the real error with "Your session has timed out!".

`showErrorToast` then swallows the original message too, because it suppresses all 401s — so nothing anywhere tells the user what actually went wrong. The reported case was a misconfigured Argo token: every Test Connection threw the developer back to the login page with no diagnosis.

## Fix

A 401 forces a logout only when the stored token is expired or expiring — the same predicate `refreshToken()` already uses.

- `TokenService.isSessionExpired()` — new, reusing the extracted `isTokenExpiredOrExpiring` so the gate and `refreshToken` cannot drift apart. The compound form matters: a token with no `exp` claim reports `isExpired: false` but `timeoutExpiry: 0`, and the two predicates disagree on exactly that input.
- `AuthProvider` — gate the refresh/logout path on it; a 401 on a still-valid token is rethrown to the caller.
- `NON_SESSION_AUTH_ERROR` — the interceptor flags those errors so `showErrorToast` shows them, while staying quiet for genuine session failures (whose forced logout already displays its own message). The GET-403 suppression is untouched: pages deliberately render their own permission placeholder.

## Why no SSO provider's login or refresh changes

The gate reuses the predicate that already guards `fetchNewToken`, so **no renewal that happens today is removed** — it only changes a branch that today can *only* produce a logout.

- **Basic/LDAP**, **SAML**, and every **`ClientType.Confidential`** client (all `GenericAuthenticator`, refreshing via `GET /api/v1/auth/refresh`) — refresh is entered only on expiry, unchanged.
- **Auth0, Okta, Azure/MSAL, Google, CustomOidc, AwsCognito** — same; `renewIdToken` is only reachable through `fetchNewToken()`, itself only reachable when the token is expired/expiring.
- **Unreadable token** — Okta and Auth0 can store `''`, and `getOidcToken()` transiently returns `''` during the service-worker handshake. `extractDetailsFromToken('')` reports `isExpired: true`, so the gate falls through to today's path. It fails *safe*.

**The one real behaviour change:** a token revoked server-side *before* its `exp` (JWKS rotation, `Invalid session.`, clock skew) no longer logs out immediately — requests fail with a visible error until the token expires, and the pre-expiry timer (`EXPIRY_THRESHOLD_MILLES`, 60 s) or the next reload then handles it. `/auth/refresh` 401s for those cases anyway, so they always ended in a logout; only the route there changes. That is the deliberate trade against today's rule, where a 401 from *any* cause silently destroys a valid session.

## Tests

All new tests fail without the fix (verified by stashing the source changes: 8 failures across the three suites).

**Jest** — `Tests: 40 passed, 40 total`
- `AuthProvider.test.tsx`: a 401 on a valid session rejects with the original error, attempts no refresh and retries nothing; the error carries the flag; an expired token still forces the logout.
- `TokenServiceUtil.test.ts`: `isSessionExpired` across expired / within-threshold / absent / nil-`exp` / valid tokens, plus that it agrees with `refreshToken`.
- `ToastUtils.test.ts` (new): flagged 401 and non-GET 403 are shown; session 401 and GET 403 stay suppressed.

**Playwright** — `playwright/e2e/Pages/SessionAuthErrors.spec.ts` (new, default chromium project — deliberately outside `e2e/Auth/`, which the project ignores):
1. expired token + failing refresh → `/signin`
2. dependency 401 on a valid session → toast, no redirect, still signed in *(fails today)*
3. 403 on a write → toast, no redirect
4. 403 on a read → silent and still signed in

**SSO suites.** The three `SSOAuthentication` "401 Interceptor" tests mocked a 401 while the OIDC token was still valid and expected a logout — that is precisely the bug (their own comment said so: *"since the token isn't truly expired at the OIDC level, refreshToken() returns null → forced logout"*). They now plant a genuinely expired token, so they still assert the logout for a *real* failure. A dependency-401 case was added there and in `SSORenewal.spec.ts` (real Keycloak SAML), whose three existing renewal tests are unchanged and are the regression proof that refresh still works.

## Not verified locally

The Playwright specs need a running OM server (and the SSO suites need the mock-OIDC / Keycloak docker stacks), which I could not start here — they are unrun and rely on CI. Jest, `tsc --noEmit` (no new errors; one pre-existing error in `AuthProvider.tsx` is only line-shifted) and the UI checkstyle sequence all pass locally.

## CI config change, flagged for review

`ci(playwright)` is a second commit and is deliberately separable. Any PR touching
`src/components/Auth/**` currently fails `plan-playwright` before a single test runs:

```
Playwright selectors matched no runnable tests:
  playwright/e2e/Flow/ServiceCreationPermissions.spec.ts
```

That mapping selects `playwright/e2e/**/*Permission*.spec.ts` but lists only the `chromium`,
`Basic`, `SearchRBAC` and `ImportExport` projects. The glob expands to
`ServiceCreationPermissions.spec.ts`, which carries `PLAYWRIGHT_INGESTION_TAG_OBJ` and — with
`PW_DEDICATED_INGESTION` on, as CI runs it — is runnable **only** in the `Ingestion` project.
Verified by listing the spec per project: 8 tests under `Ingestion`, none under any of the four.

So the commit adds `Ingestion` to that mapping's projects, and registers the new
`SessionAuthErrors.spec.ts` (4 tests, `chromium`) so future auth changes pick up this coverage.
Happy to drop it if maintainers would rather fix the mapping separately.

## Two adjustments to existing tests, called out

- `AuthProvider.test.tsx` "should set up response interceptor with correct signature" asserted `mockRefreshToken` **synchronously** after invoking the handler. With the `await` in front of the refresh that call is now a microtask later, so the assertion moved after an `await` of the handler. Intent preserved.
- The same file's `react-router-dom` mock returned `undefined` from `useNavigate`, so any test reaching the forced-logout path crashed the worker with an unhandled `navigate is not a function` rather than failing an assertion. It now returns a `jest.fn()`.

The PR distinguishes endpoint-originated 401 responses from expired-session failures.
- Adds a shared token-expiry predicate and exposes it through TokenService.
- Gates refresh and forced logout on the stored token's expiry state.
- Flags non-session 401 errors so callers can display their original messages.
- Adds Jest and Playwright coverage for session failures, dependency 401s, and 403 behavior.
</details>

The stale-request refresh race should be fixed before merging because a successfully renewed session can still surface a recoverable request as a 401 failure.

The interceptor compares a delayed 401 against mutable token storage, so a concurrent refresh can make it bypass the existing request queue and retry path.

**Files Needing Attention:** openmetadata-ui/src/main/resources/ui/src/components/Auth/AuthProviders/AuthProvider.tsx
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/src/components/Auth/AuthProviders/AuthProvider.tsx | Adds the session-expiry gate and non-session error flag, but checks mutable current token state rather than the failed request's token. |
| openmetadata-ui/src/main/resources/ui/src/utils/Auth/TokenService/TokenServiceUtil.ts | Centralizes the existing expired-or-expiring predicate and exposes it for interceptor classification. |
| openmetadata-ui/src/main/resources/ui/src/utils/ToastUtils.ts | Displays flagged endpoint-originated 401 errors while retaining suppression for session 401s and read-only 403s. |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/SessionAuthErrors.spec.ts | Adds end-to-end coverage for expired sessions, dependency 401 responses, and read/write 403 handling. |

</details>

```mermaid
sequenceDiagram
  participant Req as API request
  participant Int as Axios interceptor
  participant Store as Token storage
  participant Refresh as Token refresh
  Req->>Int: Send with token T1
  Refresh->>Store: Persist refreshed token T2
  Req-->>Int: 401 for T1
  Int->>Store: Read current token
  Store-->>Int: T2 is valid
  Int-->>Req: Reject as non-session 401
  Note over Int,Req: Request is not retried with T2
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(ui): a 401 should only sign the user..."](https://github.com/open-metadata/openmetadata/commit/a8ea49abf11fe78719bd481041dea9406ffa734e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49025491)</sub>

**Context used:**

### DIFF
--- a/.github/playwright/impact-map.json
+++ b/.github/playwright/impact-map.json
@@ -258,10 +258,11 @@
         "openmetadata-ui/src/main/resources/ui/src/components/Auth/**",
         "openmetadata-ui/src/main/resources/ui/src/pages/Login/**"
       ],
-      "projects": ["chromium", "Basic", "SearchRBAC", "ImportExport"],
+      "projects": ["chromium", "Basic", "SearchRBAC", "ImportExport", "Ingestion"],
       "specs": [
         "playwright/e2e/**/*Permission*.spec.ts",
         "playwright/e2e/Pages/Login*.spec.ts",
+        "playwright/e2e/Pages/SessionAuthErrors.spec.ts",
         "playwright/e2e/Flow/SearchRBAC.spec.ts"
       ]
     },
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Auth/SSOAuthentication.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Auth/SSOAuthentication.spec.ts
@@ -33,8 +33,19 @@ import {
   setTokenExpiry,
   waitForMockOidcReady,
 } from '../../utils/mockOidc';
+import { setToken } from '../../utils/tokenStorage';
 
 const MOCK_OIDC_URL = process.env.MOCK_OIDC_URL || 'http://localhost:9090';
+
+// A syntactically valid JWT whose exp is far in the past, used to make a session genuinely dead.
+const EXPIRED_JWT = [
+  'eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9',
+  'eyJzdWIiOiJhZG1pbiIsImlzQm90IjpmYWxzZSwiZXhwIjoxMDAwMDAwMDAwfQ',
+  'not-a-real-signature',
+].join('.');
+
+const DEPENDENCY_401_MESSAGE =
+  'The configured runner token is not authorized for this namespace';
 
 const performOidcLogin = async (page: Page): Promise<void> => {
   await page.goto('/');
@@ -342,12 +353,20 @@ test.describe('SSO Authentication with Mock OIDC Provider', () => {
   // These tests use page.route() to simulate 401 responses from the
   // OM API, triggering the frontend's axios interceptor to refresh
   // the token and retry failed requests.
+  //
+  // A 401 only ends in a logout when the stored token has actually expired —
+  // refreshToken() refuses to renew a still-valid token, so a 401 in that
+  // state cannot be recovered by refreshing and must surface as an error
+  // instead. Tests that want the logout path therefore plant an expired token
+  // rather than relying on a mocked 401 alone.
+  // See https://github.com/open-metadata/openmetadata-collate/issues/4647
   // ---------------------------------------------------------------
 
   test.describe('401 Interceptor and Request Retry', () => {
     test('should trigger session expired redirect on 401 with expired token message', async ({
       page,
       browserName,
+      request,
     }) => {
       // WebKit processes page.route() 401 interceptions with different event
       // loop timing — the async logout chain doesn't complete before the page
@@ -360,10 +379,11 @@ test.describe('SSO Authentication with Mock OIDC Provider', () => {
       await performOidcLogin(page);
       await verifyAuthenticated(page);
 
-      // Intercept the loggedInUser API call and return 401 with 'Expired token!'
-      // message. The interceptor will attempt to refresh, but since the token
-      // isn't truly expired at the OIDC level, refreshToken() returns null →
-      // forced logout with "session expired" toast.
+      // Make the session genuinely dead: the stored token is expired and the
+      // provider will refuse to renew it silently.
+      await setToken(page, EXPIRED_JWT);
+      await forceInteractionRequired(request);
+
       let intercepted = false;
       await page.route('**/api/v1/users/loggedInUser*', async (route) => {
         if (!intercepted) {
@@ -399,6 +419,10 @@ test.describe('SSO Authentication with Mock OIDC Provider', () => {
       await verifyAuthenticated(page);
 
       await resetMetrics(request);
+
+      // Expired token, so the 401s below are real session failures and the
+      // interceptor genuinely attempts a renewal to coalesce.
+      await setToken(page, EXPIRED_JWT);
 
       // Track how many 401s we injected
       const interceptedUrls = new Set<string>();
@@ -447,7 +471,9 @@ test.describe('SSO Authentication with Mock OIDC Provider', () => {
       await performOidcLogin(page);
       await verifyAuthenticated(page);
 
-      // Force the OIDC provider to reject the next silent renewal
+      // Expired token + a provider that refuses silent renewal: the session is
+      // unrecoverable, which is exactly when a logout is the right answer.
+      await setToken(page, EXPIRED_JWT);
       await forceInteractionRequired(request);
 
       // Intercept ALL API calls to return 401, simulating expired session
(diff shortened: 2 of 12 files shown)


pmbrull (dismissed):

aniketkatkar97:
Addressed the three bot findings in `fix(ui): replay a 401 whose token was refreshed mid-flight`.

**1. Stale-token race (greptile P1 / Copilot ×2)** — real, and now fixed by replaying the request.

One correction to the framing, because it changes what the fix has to be: both reviews say this "bypasses the existing queue/retry path". That path isn't reachable in this scenario. With a *valid* stored token `refreshToken()` short-circuits at `TokenServiceUtil.ts:117` and returns `null`, and the pre-existing interceptor read `null` as "refresh failed" → `resetUserDetails(true)` → **forced logout**. So before this PR the stale-token race logged the user out; the gate had already turned that into a visible error. Neither version retried.

The gap is real all the same, so the interceptor now compares the request's own `Authorization` header against the current stored token and, when they differ, replays the request once:

```ts
const currentToken = await getOidcToken();
const requestToken = getRequestBearerToken(error.config);
const isStaleRequest = Boolean(requestToken) && Boolean(currentToken) && requestToken !== currentToken;

if (isStaleRequest && !error.config[STALE_TOKEN_RETRIED]) {
  error.config[STALE_TOKEN_RETRIED] = true;

  return axiosClient.request(error.config);
}
```

Bounded by `STALE_TOKEN_RETRIED` so an endpoint answering 401 for its own reasons can't loop; the request interceptor re-reads storage, so the replay carries the new token.

**2. `setToken` fallback unreachable** — fixed. The service-worker exchange is wrapped in `try/catch` returning `false`, so a MessageChannel timeout or malformed payload falls through to the documented IndexedDB write instead of rejecting `page.evaluate`.

**3. `ToastUtils.test.ts` unmock under ts-jest ESM** — fixed. Now `jest.requireActual<typeof import('./ToastUtils')>('./ToastUtils')` rather than a static import plus `jest.unmock`.

**Tests.** The replay case fails without the change (verified by stashing it). The "at most once" case is a bound rather than a regression guard — it passes either way by construction, and exists so a future edit can't turn the replay into a loop. `Tests: 42 passed` across `AuthProvider`, `TokenServiceUtil` and `ToastUtils`; checkstyle clean; `tsc` adds no new errors. Also rebased onto latest `main`.

Issue #30779, opened 2026-07-31
Audit: ellipsis-tooltip nested inside interactive elements (~32 sites)


chirag-madlani:
## Context

While hand-finishing the Typography sweep's remaining conversions (part of #30565), two real bugs were found and fixed where core `Typography`'s `ellipsis={{ tooltip: true }}` renders a real `<button>` trigger (via `TooltipTrigger`) nested inside another interactive element:

- `UserProfileIcon.component.tsx` — nested inside a dropdown-trigger `Button`.
- `PersonaDetailsCard.tsx` — nested inside a whole-card `onClick` handler.

Both were fixed by switching to a plain `ellipsis` boolean plus a native `title` attribute, which preserves the truncation/hover-text behavior without the invalid button-in-button nesting (which also silently swallowed the parent's click).

## Scope of the follow-up

A repo-wide grep for `ellipsis={{ tooltip` currently turns up 44 files using the tooltip-ellipsis shape. A prior pass estimated roughly 32 of these sit inside another interactive element (button, link, clickable row/card) where the same invalid-nesting / swallowed-click bug class can occur. This issue tracks auditing those sites individually and applying the same fix where warranted.

## Tasks

- [ ] Enumerate every `ellipsis={{ tooltip: ... }}` call site nested inside a `<Button>`, `<Link>`, an `onClick`-bearing element, or another `TooltipTrigger`/interactive ancestor.
- [ ] For each confirmed nesting violation, replace with `ellipsis` (boolean) + native `title` (or restructure so the tooltip trigger isn't nested), following the pattern already used in `UserProfileIcon.component.tsx` and `PersonaDetailsCard.tsx`.
- [ ] Consider a lint rule or codemod check to prevent regressions once the audit is complete.

Sub-issue of #30565.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

chirag-madlani:
Follow-up evidence for the `div.prose` wrapper's blast radius, found while fixing CI on PR #30780 (the final Typography sweep PR). This is the more general case than the ellipsis-tooltip nesting this issue was originally scoped to — core `Typography` wraps its children in `<div className="prose">` on **every** render, regardless of `as`, `ellipsis`, or any other prop. Three concrete, confirmed breakages from that PR's CI run:

**1. `components/Explore/ExploreTree/ExploreTree.tsx` — structural e2e selector broke**

[...]

Issue #30565, opened 2026-07-28
Wave 1: Mechanical primitive sweeps (Typography, Button, Space/Divider, Tag, Tooltip, Badge, Skeleton, Alert, Avatar)


chirag-madlani:
Part of the AntD → `@openmetadata/ui-core-components` migration program.

Codemod-driven sweeps of the highest-count, lowest-risk antd primitives. Each sweep follows the playbook in `docs/antd-migration/README.md`: gap check → mapping guide → codemod → chunked PRs (~20–40 files) → ledger update. Each PR deletes the `.less`/dead code it orphans and regenerates visual baselines for intentionally changed pages.

Sweeps (ledger counts at program start, openmetadata-ui):
- [ ] Typography → core `Typography` (~422 imports)
- [ ] Button → `buttons` (~310)
- [ ] Space / Divider → flex `gap` utilities / `divider`
- [ ] Tooltip → `tooltip`
- [ ] Tag → `tags`
- [ ] Badge → `badges`
- [ ] Skeleton → `skeleton`
- [ ] Alert → `alert`
- [ ] Avatar → `avatar`

Per component: upstream sweep first, Collate repos follow immediately after merge. Done = ledger row reads zero across all repos.

chirag-madlani:
Typography (#30796) and Button (#30809) are done and merged into `antd-migration/wave-1`, both at **0** upstream.

`Space`/`Divider` remain, and they move to the layout work rather than staying here — `Space` is 241 sites and shares a codemod with `Row`/`Col`. See #30566.

Leaving this open only until #30941 (bulk edit tag loss, a regression from the Button sweep) is closed, since that is the outstanding item from wave 1.

PR #30780, opened 2026-07-31
refactor(ui): complete antd Typography migration to ui-core-components

### LINKED ISSUE #30796: Wave 1: complete antd Typography migration (chunks 2-13 + holdouts)
Sub-task of #30565 (epic #30570). Completes the antd Typography → `@openmetadata/ui-core-components` migration for openmetadata-ui: chunks 2–13 (~378 files) plus the final hand-finished holdouts, taking the ledger from 422 → **0**.

Includes the DOM-shape fixes for core Typography's `div.prose` wrapper (ExploreTree testid parent, nested-column links) — see #30779 for the broader audit — and the visual-baseline maintenance for pages whose seeded data is non-deterministic (see #30783 for the harness repair).

Delivered by PR #30780 into `antd-migration/wave-1`.

chirag-madlani:
## Summary

Completes the Typography sweep for `openmetadata-ui` (part of #30565, AntD → `@openmetadata/ui-core-components` migration program).

- **Scope:** ~442 files changed across the sweep, split into chunks 2–13 as separate commits (plus this final hand-finish commit) so the change is bisectable if a visual regression surfaces later. Chunk 1 (`utils/*`) landed in an earlier PR.
- **Ledger:** `openmetadata-ui`'s `Typography` row in `docs/antd-migration/LEDGER.md` goes **422 → 0** (baseline count per `docs/antd-migration/typography.md`). Verified by re-running `tooling/antd-migration/ledger.mjs` against this branch — the `Typography` row is now absent entirely.
- **This PR's own contribution:** hand-finishes the four files the codemod could only partially convert (`AlertDetails.component.tsx`, `ReindexFailures.component.tsx`, `AdminPermissionDebugger.component.tsx`, `UserPermissions.component.tsx`), which had no core equivalent for antd's `code`, `copyable`, and `ellipsis.expandable`. Resolved per the mapping guide's "no direct equivalent" section:
  - `Typography.Text code` → `Typography as="code"` with the rounded/mono/bg-secondary className idiom already established in `ContextRuleCard.component.tsx`.
  - `Typography.Text/Paragraph copyable` → `Typography` composed with the existing `CopyToClipboardButton` in a flex row; `ellipsis.expandable` dropped in favor of `rows: 2` (the cell already sits inside a `Tooltip` showing the full text, so the expand affordance was redundant).
  - Also fixed a regression this hand-edit introduced: core `Typography` always wraps content in an outer `div.prose` regardless of `as`, so nesting the new `as="code"` Typography inside the file's pre-existing `as='p'` Typography produced an invalid `<div>`-in-`<p>` (a `validateDOMNesting` warning not caught by any existing test assertion). Fixed by switching the outer wrapper to `as="div"`.

## Approved mapping decisions (docs/antd-migration/typography.md)

- `Typography.Title level={N}` → `as="hN"` + `size` via the approved `LEVEL_SIZE_MAP` (1→display-sm, 2→display-xs, 3→text-xl, 4→text-lg, 5→text-md). Level 5 dominates real usage (>70%).
- `type="secondary"/"success"/"warning"/"danger"` → `color="secondary"/"success"/"warning"/"danger"` (core `color` prop landed in main ahead of the bulk sweep).
- `strong` → `weight="bold"` (literal-only; dynamic `strong={expr}` was a codemod skip, hand-resolved where it occurred).
- `underline` → `className` gets `tw:underline` appended.
- `Typography.Text` → `Typography` (`as="span"` default); `Typography.Paragraph` → `Typography as="p"`; `Typography.Link` → `Typography as="a"`.

## Codemod vs. hand-finish

The codemod (`tooling/antd-codemods/transforms/antd-typography-to-core.js`) handled the mechanical prop/import rewrites above across chunks 2–13. Hand-finished on top of the codemod:
- Bare `<Typography>` wrappers (codemod deliberately skips these — antd's bare form renders an `<article>`, core's default is `span`; each needed the intended element verified by hand): `BlockEditor/*`, `CustomStatistic.tsx`, `TestCaseIncidentManagerStatus.component.tsx`, `SuccessScreen.tsx`, `SummaryTagsDescription.component.tsx`, `DataInsightHeader.component.tsx`, `ProfilerObjectFieldTemplate.tsx`, `MetricExpression.tsx`, `TourEndModal.tsx`, `Sso*FieldTemplate.tsx`, `ErrorPlaceHolderIngestion.tsx`, `WorkflowArrayFieldTemplate.tsx`.
- Dynamic `Title level={expr}` and `type={expr}`/unsupported `type` values (codemod skips, no fixed value to look up at codemod time).
- The `code`/`copyable`/`ellipsis.expandable` hard gaps in this PR's four files (above).
- **Two real bugs found and fixed** in `UserProfileIcon.component.tsx` and `PersonaDetailsCard.tsx`: core `Typography`'s `ellipsis={{ tooltip: true }}` renders a real `<button>` trigger (via `TooltipTrigger`), which was nested inside another interactive element — a dropdown-trigger `Button` in one case, a whole-card `onClick` in the other. This is invalid HTML (button-in-button) and silently swallowed the parent's click. Fixed by switching to a plain `ellipsis` boolean + native `title` attribute, which preserves the hover-tooltip truncation without the interactive nesting.
  - **Residual risk:** a repo-wide grep for `ellipsis={{ tooltip` turns up 44 files using this shape; a prior pass estimated ~32 of these sit inside another interactive element and share the same bug class. Filed **#30779** (sub-issue of #30565, assigned to @chirag-madlani) to audit and fix the remainder.
- **`PageHeader.interface.ts` fix:** `titleProps`/`subHeaderProps` were still typed against antd's `Typography`/`Paragraph` props after `PageHeader.component.tsx` itself was converted to core Typography (sweep chunk 4/13). antd's `ParagraphProps` inherits a generic HTML `color?: string` from `HTMLAttributes`, which isn't assignable to core Typography's narrower `color?: TypographyColor` union, so `{...subHeaderProps}` stopped type-checking once the component switched. Retyped both props against core's `TypographyProps` — the one genuine new `tsc` error the sweep introduced, isolated by diffing against the pre-sweep baseline.

## Gate results

- **tsc --noEmit:** 0 new errors (523 pre-existing baseline errors, unrelated to this sweep, unchanged before/after — isolated by diffing against `origin/antd-migration/wave-1`).
- **jest, full suite:** `Test Suites: 3 skipped, 1155 passed, 1155 of 1158 total` / `Tests: 52 skipped, 13637 passed, 13689 total`. 0 failures.
- **eslint / prettier / organize-imports:** clean on this PR's own hand-finish commit (one import-order violation and one formatting nit introduced by the hand-edit were caught and fixed before commit; all remaining eslint findings across the four touched files are pre-existing and unrelated, confirmed by diffing lint output against the pre-edit tree).

## Notes

- The Collate-side sweep (`collate-ui`, `collate-local-webserver`) is tracked as a separate PR in the `openmetadata-collate` repo.
- Every converted call site carries DOM-wrapper layout-regression risk per the mapping guide (core `Typography` always renders inside `div.prose`) — the visual-regression baselines under the collate visual project should be checked against the affected pages before merging past this PR into `main`.

## Test plan

- [x] `tsc --noEmit` — 0 new errors vs. pre-sweep baseline
- [x] `jest` full suite — 1155/1158 suites, 13637/13689 tests, 0 failures
- [x] `eslint` / `prettier` / `organize-imports` clean on the hand-finish commit
- [x] Ledger regenerated — `Typography` row absent for `openmetadata-ui`
- [ ] Visual-regression baselines reviewed for pages touched by this sweep (tracked separately, see Notes)

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

Fixes #30796


chirag-madlani:
Addressed the review in two commits — `eec96c9` (core) and `d024312` (call sites).

**Findings 1 + 2 share one root cause**, so they're fixed once in core rather than at 442 call sites.

`typography.css` works in two layers: `.prose:not(...)` on the element itself sets only the `--tw-prose-*` vars plus `color`/`font-size`/`line-height`, while `.prose :not(...)` — a *descendant* selector — carries every real typographic rule, each gated on an element type (`p`, `h1`-`h6`, `ol`, `ul`, `li`, `blockquote`, `a`, `code`, `pre`, `img`, `figure`, table).

So the wrapper is genuinely load-bearing for those types — putting `prose` on a `<p>` directly would stop `.prose p` matching and silently drop its margins. But **`span` and `div` are targeted by no rule in that block**, so for them the wrapper contributes only inherited properties, and setting `prose` on the element gives an identical computed style. Those now render without a wrapper; ellipsis and non-default quote variants keep it. Deliberately a small allowlist.

A scan of all 591 files containing `<Typography` reports **no remaining invalid nesting**. The two that survived the core fix (`DataAssetsHeaderExtraInfo`, `MarketPlaceAppDetails`) were inline Typography used purely as layout containers and are now `as="div"`.

**Finding 5 — I did not apply the suggested fix.** The diagnosis is right that the handler never fires today, but the proposed wrapper `<div onClick>` would introduce a different bug. `handleCardClick` is a *toggle*, and the row-level `onClick` on the `<Space>` is still attached. Once #30804 stops the TooltipTrigger swallowing propagation, both handlers fire on one click and cancel each other out — selection would appear frozen. Removed the re-wires instead and left selection to the single row-level handler. This makes the PR depend on #30804, which was already in the merge chain.

Findings 3 and 4 applied as suggested.

**Note for baselines:** this restores inline flow, so it legitimately changes rendering. The current visual-regression baselines were captured *with* the block wrapper, i.e. they encode the regression as expected. They will need regenerating from CI after #30804 and this land — the diffs are the fix, not a new break.

Verified: full consumer suite green (1155 suites / 13,637 tests), core suite 21/21, eslint clean.

Issue #30803, opened 2026-08-01
Typography ellipsis tooltip swallows clicks intended for ancestor elements


chirag-madlani:
`Typography` with `ellipsis={{ tooltip: … }}` wraps its content in `TooltipTrigger`, which renders a react-aria `Button`. react-aria's `usePress` sets `shouldStopPropagation: true` by default, so **a click on the truncated text never reaches an ancestor's `onClick`**.

The wrapper exists only to host a hover/focus tooltip over non-interactive text — it should not consume clicks.

## Confirmed breakage

Found while migrating antd Typography → ui-core-components (#30565):

- **Persona switcher** — clicking a persona in the profile dropdown silently did nothing; the `onClick` lives on an ancestor row, so persona/navigation settings never applied.
- **Test-case selection cards** (`AddTestCaseList`) — cards could not be selected at all; three Playwright specs hung waiting on API calls that never fired.

## Why this needs a core fix

There are ~57 `ellipsis={{ tooltip: true }}` call sites in `openmetadata-ui`. Only a handful sit on e2e-covered paths, so an unknown number of click targets are silently broken today, and every future migration sweep would reintroduce the pattern.

Fixed by PR (linked below): calls `continuePropagation()` on the ellipsis tooltip's press event only. Deliberately **not** changed on the shared `TooltipTrigger`, because some consumers rely on the swallowing default (e.g. a help-icon tooltip inside a sortable table header must not also trigger sort).

Issue #30570, opened 2026-07-28
Epic: Migrate UI from Ant Design to @openmetadata/ui-core-components (Untitled + Tailwind)


chirag-madlani:
## Goal

Remove `antd` / `@ant-design/*` entirely from the OpenMetadata UI (and downstream consumers), adopting `@openmetadata/ui-core-components` (Untitled UI + Tailwind v4 + react-aria-components) as the single component system. Visual change to the new design language is expected and desired.

## Approach

Codemod-assisted **per-component sweeps**, ordered by risk: mechanical primitives first (codemod-driven), then structural primitives, then the layout shell, then complex controls (hand-migrated), then teardown. Every sweep follows the playbook in `docs/antd-migration/` (lands with Wave 0): gap check in `ui-core-components` → mapping guide (the review contract) → codemod where mechanical → chunked ~20–40-file PRs → sweep-ledger update. Each PR also deletes the `.less`/`.ant-*` overrides and dead code it orphans.

Guardrails:
- **Visual regression**: Playwright screenshot baselines (docker-generated for Linux-rendering parity) gate every sweep PR; intentional changes regenerate baselines in the same PR so reviewers see before/after.
- **Sweep ledger** (`tooling/antd-migration`): antd imports per component per repo + `.ant-*` selector counts. A component is done when its row reads zero everywhere.
- antd's global reset stays until the final teardown so unmigrated screens don't shift.
- QueryBuilder is excluded — owned by #29849.

## Waves

- [ ] #30564 — Wave 0: Foundations (harness, codemods, ledger) — PR #30560
- [ ] #30565 — Wave 1: Mechanical primitives
- [ ] #30566 — Wave 2: Structural primitives
- [ ] #30567 — Wave 3: Layout shell
- [ ] #30568 — Wave 4: Complex controls + RJSF
- [ ] #30569 — Wave 5: Teardown

chirag-madlani:
## Re-scoped: AI-mode-first, and Collate runs independently

The wave model below is superseded. Three things changed it.

**1. Scope is now AI-mode-first.** The revamped pages under `plugins/ai-chat` are moving to OSS with Classic and AI both shipping. Scoping to what AI mode can reach gives a boundary that is shippable and demonstrable rather than a ~1,600-site sweep with no natural stopping point.

[...]

PR #30810, opened 2026-08-01
refactor(ui): complete antd Button migration to ui-core-components

### LINKED ISSUE #30809: Wave 1: complete antd Button migration to ui-core-components
Sub-task of #30565 (epic #30570). Migrates all antd `Button` usage in openmetadata-ui to `@openmetadata/ui-core-components` per docs/antd-migration/button.md, taking the ledger from **310 → 0**.

Includes the hand-finish punch list (shape, Button.Group redesign, dynamic sizes/block/danger) and fixes for three real interaction regressions surfaced during the sweep — see PR for detail.

chirag-madlani:
## Summary

Second Wave 1 primitive (#30565, epic #30570). Migrates all antd `Button` usage in `openmetadata-ui` per `docs/antd-migration/button.md`. **Ledger: `Button` 310 → 0.** 336 files, 14 commits kept per-area for bisectability.

Stacked on the Typography sweep (#30780) so the codemod ran over already-migrated files — base retargets to `antd-migration/wave-1` once that lands.

## Applied mapping (approved 2026-07-30)

`type→color` incl. the default inversion (bare `<Button>` gets explicit `color="secondary"`, since antd's default is *default* but core's is *primary*); `danger` folds into `-destructive`; `ghost`→`tertiary`; sizes `small/middle/large → xs/sm/md`; `disabled→isDisabled`; `loading→isLoading`; `icon→iconLeading`; `htmlType→type`; `block→tw:w-full`; `ref` passes through (forwardRef from #30664).

## Hand-finished (27 sites the codemod deliberately skipped)

`shape="circle"/"round"` (4, via `tw:rounded-full`), `Button.Group` + subpath imports (5 — core's `ButtonGroup` is a single-select toggle, so these were redesigned as flex-row compositions rather than mechanically swapped), dynamic `size` (8), dynamic `block`+className (4), dynamic `danger` (1), plus 2 pre-existing `Button as CoreButton` alias collisions.

## Real regressions found and fixed

Beyond test-assertion drift, the sweep surfaced three genuine interaction bugs:

1. **Click swallowing in `ApplicationCard`** — react-aria's `usePress` stops propagation that antd's plain button allowed to bubble to an ancestor `onClick`. (Same class as #30803/#30804.)
2. **Core Button silently drops a bare `onFocus` prop** (confirmed empirically) — broke MetaPilot focus tracking in 4 files; worked around with a wrapping `<span onFocus>`. Worth a core follow-up.
3. **`PropertyValue.tsx` passed antd-shaped `saveButtonProps={{ htmlType: 'submit', … }}` into the now-core-based `InlineEdit`**, silently breaking submit-on-click across 10 property-type editors. `tsc` could not catch this because `InlineEdit`'s prop type was still declared against antd's `ButtonProps` — a reminder that intermediate prop types need migrating alongside call sites.

Also: `.ant-btn-loading` class assertions → `data-loading` attribute (7 files), and core-components jest mock factories widened to forward `data-testid`/`isDisabled` (12 files, 4 of them pre-existing gaps this sweep exposed).

## Verification

| Gate | Result |
|---|---|
| `tsc --noEmit` (isolated vs stashed baseline) | **0 new errors** — 521 head vs 674 baseline, i.e. the sweep *fixed* 153 pre-existing |
| Full jest | **1155/1155 suites, 13637/13637 tests** |
| eslint + prettier (full diff) | clean |

Visual baselines are intentionally **not** regenerated locally — per `docs/antd-migration/README.md`, baselines come from CI artifacts only. Button changes appearance by design, so expect intentional baseline updates from the CI run.

## For reviewers

- Orphaned `.ant-btn-group.spaced` CSS left in place (not deleted) pending confirmation nothing else uses it
- `GithubStarCard`'s joined-pill visual not verified locally (no-local-baseline rule) — worth an eye on the CI baseline
- Suggest a repo-wide grep for other stale `ButtonProps` type imports from 'antd' outside this sweep's touched files

Fixes #30809


chirag-madlani:
Merging into `antd-migration/wave-1` with three e2e shards knowingly red — `data-asset-rules-01`, `ingestion-01`, `reindex-01` — tracked in #30941.

Those three are a confirmed regression from this sweep, not flake: they pass on #30780 (one commit earlier, same pipeline and seeded data) and on unrelated PRs, and they fail consistently across two reruns of the same commit here. The trace shows bulk edit runs both its preview and apply calls, but the selected tag never reaches the import payload — so a bulk edit silently drops tags.

Merging anyway because wave-1 is an integration branch for review rather than a release, and this should not hold up the rest of the wave. **#30941 must be closed before wave-1 goes to `main`.**

Two real bugs were found and fixed on this branch along the way:
- `data-icon` — icons passed as elements rendered at 0×0, collapsing every icon-only button (measured in Chrome: 36×36 with the fix, 16×16 with a 0×0 icon without). This took out all 11 chromium shards and would have shipped invisible edit buttons across the product.
- `InlineEdit` — converted props were emitted after the `{...saveButtonProps}` spread, inverting precedence so caller overrides were silently clobbered.

Nine visual baselines were refreshed, each compared against its predecessor rather than adopted wholesale.

PR #30851, opened 2026-08-03
fix(BE): evaluate task conditions for per-entity task permissions

### LINKED ISSUE #30850: Permissions API: task permission does not evaluate task conditions, so ResolveTask is denied for all users
## Summary
`GET /v1/permissions/{resource}/{id}` cannot report correct permissions for the `task` resource because it builds a generic `ResourceContext` that does not load a task's `assignees` or `createdBy`. As a result the task-scoped SpEL policy conditions — `isTaskAssignee()`, `isTaskFiler()`, `isTaskReviewer()` — always evaluate to `false`, and task operations such as `ResolveTask` resolve to `notAllow` for **every** user, including legitimate assignees.

Any client that gates task actions (approve / reject / resolve) on this endpoint therefore gets the wrong answer for everyone.

## Root cause
`PermissionsResource` builds the resource context for the by-id and by-name handlers as:

```java
ResourceContext<?> resourceContext = new ResourceContext(resource, id, null);
```

`ResourceContext.resolveEntity()` only requests `owners`, `tags`, `domains`, and `reviewers` fields. For a `Task`, `assignees` and `createdBy` are never loaded, so `RuleEvaluator.isTaskAssignee()` / `isTaskFiler()` (which read `task.getAssignees()` / `task.getCreatedBy()`) see `null` and return `false`.

The seeded `TaskAuthorPolicy` relies on exactly these conditions (allow `ResolveTask` when `isTaskAssignee() || isTaskReviewer()`, deny when `isTaskFiler()`), so it can never be reflected through this endpoint.

Note: `TaskResource.delete` already worked around the same limitation by constructing an explicit `TaskResourceContext(task)` before authorizing — the permissions endpoint was simply never given the same treatment.

## Steps to reproduce
1. Create a task and assign user A (who is not the task creator).
2. Ensure a policy grants `ResolveTask` when `isTaskAssignee()` (e.g. the seeded
   `TaskAuthorPolicy` via the DataConsumer role).
3. As user A, call `GET /v1/permissions/task/{taskId}`.
4. **Observed:** `ResolveTask` access is `notAllow`.
5. **Expected:** `ResolveTask` access is `allow` (A is an assignee).

Similarly, the task creator should get `deny` (via `isTaskFiler()`), but with the generic context they also get `notAllow` — the deny rule never fires.

## Expected behavior
`GET /permissions/task/{id}` must evaluate task-scoped conditions against the
actual task, so:
- assignee / reviewer who did not file the task → `ResolveTask = allow`
- filer (even if also an assignee) → `ResolveTask = deny` (deny overrides allow)

## Proposed fix
In `PermissionsResource`, build a `TaskResourceContext` for the `task` resource (loading `assignees,reviewers,watchers,about,domains,createdBy`) on both the by-id and by-name permission handlers; keep the generic `ResourceContext` for all other resources. This mirrors the existing `TaskResource.delete` handling.

anuj-kumary:
### Describe your changes:

Fixes #30850
Issue - Task permission only loads owners/tags/domains/reviewers — not assignees or createdBy. So when the policy condition isTaskAssignee() runs, task.getAssignees() is null → returns false → ResolveTask is never granted through this endpoint, for anyone. Every approve button hides.
Build a TaskResourceContext for the `task` resource on both the by-id and by-name permission endpoints (loading assignees/reviewers/watchers/about/ domains/createdBy).

Adds TaskPermissionsIT: verifies GET /permissions/task/{id} returns ResolveTask=ALLOW for a non-filer assignee and DENY for the filer (who is also an assignee), exercising the seeded TaskAuthorPolicy semantics.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### High-level design:

N/A — small change. 

#
#### Manual testing performed

#

Not applicable. 

#
### Checklist:


openmetadata-service/src/main/java/org/openmetadata/service/resources/permissions/PermissionsResource.java
```diff
@@ -175,8 +182,22 @@ public ResourcePermission getPermission(
...
+    return authorizer.getPermission(
+        securityContext, user, buildResourceContext(resource, null, name));
+  }
+
+  private ResourceContextInterface buildResourceContext(String resource, UUID id, String name) {
+    ResourceContextInterface resourceContext;
+    if (Entity.TASK.equals(resource)) {
+      Task task =
+          id != null
+              ? Entity.getEntity(Entity.TASK, id, TASK_AUTH_FIELDS, Include.NON_DELETED)
```
yan-3005: Entity.getEntity(Entity.TASK, id, TASK_AUTH_FIELDS, Include.ALL) We need to pass ALL here

yan-3005 (approved):

PR #30867, opened 2026-08-03
perf(ui): fetch database in parallel with permission on DatabaseDetailsPage (W1 template)

harsh-vador:
Fixes [5445](https://github.com/open-metadata/openmetadata-collate/issues/5445)

## Description

The database detail page paid **two serial round-trips** on every mount: a permission fetch, then a database `useQuery` gated on the permission resolving (`enabled: hasViewBasicPermission && !permissionsLoading`). This fires the database GET **in parallel** with the permission fetch instead.

This is the **template** for the W1 request-waterfall fix — the audit lists ~7 primary-render pages (SearchIndexDetailsPage, Database/DatabaseSchema/Version pages, StoredProcedurePage, APICollectionPage) with the identical permission-gates-fetch shape. Reference parallel pattern already in the tree: `PlatformLineage.tsx`.

### Preserving the no-permission UX (the subtle part)

Because the GET now runs speculatively, a no-permission user's GET returns 403. Three coordinated changes keep the prior behaviour exactly:

1. `enabled` is gated only on the FQN → the query fires immediately, in parallel.
2. Render checks `!hasViewBasicPermission` **before** `isError` → a no-permission 403 shows the inline **PERMISSION** placeholder, not the generic error placeholder.
3. The `/forbidden` redirect fires **only on a genuine permission desync** (`hasViewBasicPermission` true but the backend still denied), never for a normal no-permission user.

## Type of change

- [x] Performance (request waterfall → parallel)

## Tests

- **Added a regression unit test**: a no-permission user gets the PERMISSION placeholder (no generic error, no forbidden redirect) **and** the database GET is still fired in parallel.
- All **5** DatabaseDetailsPage unit tests pass; eslint clean.

## Verification note

The 403 / permission UX is guarded by the new unit test, but is best also confirmed in a **running app / Playwright** (Jest mocks permissions and does not exercise the real backend 403 path). Recommend a manual/Playwright pass on: (a) happy path, (b) no-permission user sees the inline placeholder and is **not** redirected to /forbidden.

Ref: open-metadata/openmetadata-collate#5442

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/pages/DatabaseDetailsPage/DatabaseDetailsPage.test.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/pages/DatabaseDetailsPage/DatabaseDetailsPage.test.tsx
@@ -15,12 +15,23 @@ import { findByTestId, findByText } from '@testing-library/react';
 import React from 'react';
 import { MemoryRouter } from 'react-router-dom';
 import PageLayoutV1 from '../../components/PageLayoutV1/PageLayoutV1';
+import { usePermissionProvider } from '../../context/PermissionProvider/PermissionProvider';
 import {
   getDatabaseDetailsByFQN,
   patchDatabaseDetails,
 } from '../../rest/databaseAPI';
 import { renderWithQueryClient } from '../../test/unit/test-utils';
 import DatabaseDetailsPage from './DatabaseDetailsPage';
+
+const FULL_PERMISSION = {
+  Create: true,
+  Delete: true,
+  ViewAll: true,
+  EditAll: true,
+  EditDescription: true,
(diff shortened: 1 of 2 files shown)


harsh-vador:
Closing after reconsidering the approach. `PermissionProvider.getEntityPermissionByFqn` caches results (`entitiesPermission[fqn]`), so the serial permission→data path is usually not a real waterfall — on in-app navigation the permission resolves instantly from cache. Firing the entity GET in parallel therefore only helps the narrow cold-cache (deep-link/refresh) case for permitted users, while adding a throwaway 403 for every no-permission user (backend log/metric noise). The narrow benefit doesn't justify the doomed calls, so backing this out. Ref: open-metadata/openmetadata-collate#5442

PR #30869, opened 2026-08-03
Fixes 30868: stop Auto Classification tagging plain code columns as PII

### LINKED ISSUE #30868: Auto Classification tags ordinary code columns as PII.Sensitive
## Affected module
Ingestion — Auto Classification / PII recognizers

## Describe the bug
A column named `scenario_code` holding values like `SCN-125`, `SCN-113`, `SCN-100` is auto-classified as PII (`PII.Sensitive` in OSS, `General.CreditCardNumber` on Collate) with a confidence of 1.00.

Two independent defects in the CVV recognizer path combine to produce this:

**1. The `cvv_pattern` regex is not anchored.** `piiTagsWithRecognizers.json` defines it as `\b\d{3,4}\b`, which matches a 3-4 digit run *inside* a value — the `125` in `SCN-125` matches. Real CVV columns hold a bare `123`, so the pattern should match the whole value.

**2. Context words are matched as substrings, not tokens.** `presidio_utils.enhance_using_context` tests each recognizer context word with `ctx_word.lower() in " ".join(context).lower()`, where `context` is the column name split into parts. `"cid" in "acid level"` is true, so a column named `acid_level` (or `incident_count`, or `decoder_ring` via `code`) matches the CVV context list.

Either defect alone is enough to tag a column, because OpenMetadata's `enhance_using_context` sets a matching result's score straight to `MAX_SCORE` rather than applying Presidio's additive `+0.35`. So the weakest pattern in the set (score 0.5, "any 3-4 digits") becomes a certain match — past the 0.6 classification minimum and past the default workflow confidence of 80.

## To Reproduce
1. Create a table with a column `scenario_code` of `VARCHAR`, populated with `SCN-100` … `SCN-125`.
2. Run the Auto Classification workflow with default settings (`confidence: 80`, `classificationLanguage: en`).
3. The column is tagged `PII.Sensitive` with score 1.00.

Renaming the same column to `scenario_id` produces no tag at all — the column name alone flips the result.

Also affected, with bare 3-digit values: `error_code`, `acid_level`, `incident_count`, `decoder_ring`.

## Expected behavior
Only columns that actually hold a CVV/CVC value should match the CVV recognizer. A column name that merely *contains the letters* of a context word should not.

## Version
- OS: any
- Python version: 3.10 / 3.11
- OpenMetadata version: 2.0.0-SNAPSHOT (`main`)
- OpenMetadata Ingestion package version: `main`

pmbrull:
### Describe your changes:

Fixes #30868

A column named `scenario_code` holding `SCN-125`, `SCN-113`, … was auto-classified as `PII.Sensitive` at confidence 1.00. Two independent defects combine to cause it, and either one alone is enough to tag the column.

**1. The `cvv_pattern` regex was not anchored.** `\b\d{3,4}\b` matches a 3-4 digit run *inside* a value, so the `125` in `SCN-125` matched. Changed to `\A\d{3,4}\Z`.

Not `^\d{3,4}$` — the recognizer sets `regexFlags.multiline`, under which `^..$` still matches a single line of a multi-line value. `\A..\Z` was the only anchor that closed that case in testing.

**2. Context words were matched as substrings, not tokens.** `presidio_utils.enhance_using_context` tested each context word with `ctx_word.lower() in " ".join(context).lower()`, where `context` is the column name split into parts. `"cid" in "acid level"` is true, so `acid_level`, `incident_count` and `decoder_ring` all matched the CVV context list. Extracted the comparison into `context_matches()`, which compares whole tokens; multi-word entries (`"indian passport"`) keep substring semantics since they have no single token to compare against.

These stack because OpenMetadata's `enhance_using_context` sets a matching result's score straight to `MAX_SCORE` instead of Presidio's additive `+0.35`. That turns the weakest pattern in the set (0.5, "any 3-4 digits") into a certain match — past the 0.6 classification minimum and past the default workflow confidence of 80. I left that behaviour alone: changing it affects every recognizer, and both defects above are fixable without it.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered
- A `scenario_code` / `error_code` column of `SCN-125`-style values is no longer tagged PII
- An `acid_level` / `incident_count` / `decoder_ring` column of bare 3-digit values is no longer tagged PII
- A `cvv`, `security_code` or `card_verification_code` column of bare 3-digit values is **still** tagged PII
- A `credit_card_number` column is still tagged PII
- A multi-line value whose second line is 3 digits is no longer tagged PII

#### Unit tests
- Files updated: `ingestion/tests/unit/pii/algorithms/test_presidio_utils.py`
  - new `TestContextMatches` — whole-token matches, substring non-matches (`acid`/`incident`/`decoder`), multi-word entries, case-insensitivity
  - `TestEnhanceUsingContext::test_context_word_that_is_only_a_substring_does_not_boost`
  - `TestEnhanceUsingContext::test_score_below_minimum_is_not_boosted` — first coverage of the `MIN_SCORE_FOR_ENHANCEMENT` branch, which had none
- `ingestion/tests/unit/pii` + `ingestion/tests/unit/metadata/pii`: **235 passed** (222 before, +13 new), 0 failures.

- [x] Not applicable — covered by unit tests above plus the end-to-end scoring run below.

#### Playwright (UI) tests

#### Manual testing performed
Replayed the real scoring path (`TagAnalyzer` → `ScoreTagsForColumnService` → `ConflictResolver`, Presidio + `en_core_web_md`) offline against the shipped recognizer definitions:

| column | values | before | after |
|---|---|---|---|
| `scenario_code` | `SCN-100` … `SCN-125` | 1.00 ❌ | none ✅ |
| `error_code` | `ERR-200` … | 1.00 ❌ | none ✅ |
| `acid_level` | bare 3-digit | 1.00 ❌ | none ✅ |
| `decoder_ring` | bare 3-digit | 1.00 ❌ | none ✅ |
| `incident_count` | bare 3-digit | 1.00 ❌ | none ✅ |
| `note_code` | `"line one\n125"` | 1.00 ❌ | none ✅ |
| `cvv` | bare 3-digit | 1.00 ✅ | 1.00 ✅ |
| `security_code` | bare 3-digit | 1.00 ✅ | 1.00 ✅ |
| `card_verification_code` | bare 3-digit | 1.00 ✅ | 1.00 ✅ |
| `credit_card_number` | Luhn-valid PANs | 0.90 ✅ | 0.90 ✅ |

Also ran a 20-column regression table covering every auto-classifiable tag (Person, Email, PhoneNumber, Address, Location, NRP, Gender, BirthDate, DateTime, CreditCardNumber, BankNumber, IBANCode, Crypto, IPAddress, URL, NationalID, VATCode, DriverLicense, MedicalLicense) at the default confidence of 80: **0 regressions**.

Migration verified against a real tag row on **PostgreSQL 15** and **MySQL 8**:
- updates exactly 1 row, only the intended tag
- recognizer count, recognizer order and the context list are preserved
- re-running updates 0 rows

Note on the MySQL migration: locating the recognizer with `EXISTS (SELECT ... FROM JSON_TABLE(json, ...))` in the `WHERE` clause of an `UPDATE` does **not** correlate reliably — it evaluated correctly in a `SELECT` but silently matched 0 rows in the `UPDATE` as soon as the table had more than one row, which would have shipped a no-op migration. It uses a join instead, which is where the idempotency gate lives.

#

Not applicable.

#
### Checklist:
- [x] For UI changes: I attached a screen recording and/or screenshots above.

Bug fix

Migration re-verified by running the committed statement against a live OpenMetadata database (the
dev PostgreSQL 15.18 instance, connected with the `conf/openmetadata.yaml` default credentials) rather
than only a restored fixture:

- `UPDATE 1` on the first run, `UPDATE 0` on a second run
- recognizer count, an md5 of the recognizer-name ordering, and the context-list length are all
  byte-identical before and after
- `tag` confirmed as the only table holding recognizer config; `tag_usage` stores applied labels and
  `classification` stores only classification-level `autoClassificationConfig`
- exactly two rows in the whole database carry a `CvvRecognizer`: `PII.Sensitive` (this PR) and
  `General.CreditCardNumber` (the Collate PR), so between the two migrations there is no overlap and
  no row is missed

After migrating, re-scoring a live `scenario_code` column of `SCN-100` … `SCN-125` against the
database's own tag definitions returns no tag, while `cvv` and `security_code` still score 1.00.


pmbrull:
Thanks — half of this is right, and I've pushed a fix for that half in 2d9423669a.

**The crash risk is not real, on either line.**

Line 483 (the `enhance_using_context` decorator) already returned early on a falsy `rec.context`, and that guard predates this PR:

```python
if not rec.context or not context:
    return results
```

So `rec.context` was never reached in an empty state there.

Line 349 (`ContextAwareUsBankRecognizer`) has no such guard, but `self.context` cannot be `None` or empty. Presidio's `UsBankRecognizer.__init__` coerces it:

```python
context = context if context else self.CONTEXT
```

Verified against the actual factory:

```
factory default (context omitted)  -> self.context=['check', 'account', 'account#', 'acct', 'bank', 'save', 'debit']
factory context=None               -> self.context=['check', 'account', 'account#', 'acct', 'bank', 'save', 'debit']
direct  context=None               -> self.context=['check', 'account', 'account#', 'acct', 'bank', 'save', 'debit']
direct  context=[]                 -> self.context=['check', 'account', 'account#', 'acct', 'bank', 'save', 'debit']
```

Even hypothetically it would not be a regression from this PR: the code being replaced iterated the same attribute — `any(ctx_word.lower() in context_lower for ctx_word in self.context)`.

**"Compute the match once per call" is a fair point, and I've applied it.**

`context_matches()` was called inside the per-result loop, re-joining and re-tokenizing the column name for every result. The match depends only on the recognizer's context list and the column name, neither of which changes inside the loop. Both sites now resolve it once and bail out early:

```python
if not rec.context or not context or not context_matches(rec.context, context):
    return results
```

That also removes a level of nesting from the loop body. I added the explicit `self.context` guard on the UsBank path while I was there — it cannot trigger, but it makes the precondition local rather than inherited from a base class two libraries away.

No behaviour change: 235 unit tests still pass, and the false-positive set (`scenario_code`, `error_code`, `acid_level`, `decoder_ring`, `incident_count`, multi-line `note_code`), the true-positive set (`cvv`, `security_code`, `card_verification_code`, `credit_card_number`) and the 19-tag regression table are all identical before and after.

IceS2 (approved):

pmbrull:
picked to 2.0

PR #30870, opened 2026-08-03
perf(ui): fetch entity in parallel with permission on 3 more detail pages (W1 rollout)

harsh-vador:
## Description

Applies the [DatabaseDetailsPage template (#30867)](https://github.com/open-metadata/OpenMetadata/pull/30867) to the other react-query permission-gated detail pages. Each entity `useQuery` was `enabled` only after the permission fetch resolved (`enabled: ...viewPermission && !permissionsLoading`), forcing **two serial round-trips** on mount. Each now fires the GET **in parallel** with the permission fetch.

- **SearchIndexDetailsPage** — `enabled` gated on FQN only. Render already checks `!viewPermission` before the error state and there is no forbidden redirect, so no other change is needed.
- **StoredProcedurePage** — `enabled` gated on FQN only; the `/forbidden` redirect now fires only on a genuine permission desync (`viewBasicPermission` true), so a no-permission user falls through to the inline PERMISSION placeholder.
- **APICollectionPage** — `enabled` gated on FQN only; the 403 branch is nested so a no-permission 403 redirects only on desync and **never** falls through to the error toast. Render already checks permission before `isError`.

## Type of change

- [x] Performance (request waterfall → parallel)

## Tests

- Updated the two tests that asserted the old *"no permission ⇒ no fetch"* waterfall behaviour to assert the new intended behaviour: the fetch fires in parallel and render still gates on permission (placeholder shown).
- All **16** unit tests across the three pages pass; eslint clean.

## Verification note

Same as the template — the 403 / permission UX is best also confirmed in a **running app / Playwright** (Jest mocks permissions, not the real backend 403). Check: happy path loads; a no-permission user sees the inline placeholder and is **not** redirected to /forbidden.

Ref: open-metadata/openmetadata-collate#5442

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/pages/APICollectionPage/APICollectionPage.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/pages/APICollectionPage/APICollectionPage.tsx
@@ -141,11 +141,11 @@ const APICollectionPage: FunctionComponent = () => {
       decodedAPICollectionFQN,
       API_COLLECTION_DEFAULT_FIELDS
     ),
-    enabled: Boolean(
-      decodedAPICollectionFQN &&
-        viewAPICollectionPermission &&
-        !isPermissionsLoading
-    ),
+    // Fire the API-collection fetch in parallel with the permission fetch
+    // rather than gating it on permission — removes a serial round-trip on
+    // mount. A no-permission user's speculative GET 403s and render shows the
+    // inline PERMISSION placeholder (checked before the error state below).
+    enabled: Boolean(decodedAPICollectionFQN),
   });
 
   const isError = useMemo(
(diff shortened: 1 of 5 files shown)


harsh-vador:
Closing after reconsidering the approach. `PermissionProvider.getEntityPermissionByFqn` caches results (`entitiesPermission[fqn]`), so the serial permission→data path is usually not a real waterfall — on in-app navigation the permission resolves instantly from cache. Firing the entity GET in parallel therefore only helps the narrow cold-cache (deep-link/refresh) case for permitted users, while adding a throwaway 403 for every no-permission user (backend log/metric noise). The narrow benefit doesn't justify the doomed calls, so backing this out. Ref: open-metadata/openmetadata-collate#5442

PR #30901, opened 2026-08-03
Custom Connector per service type, with integration tests

### LINKED ISSUE #30899: Custom Connectors have no test coverage for any service type
Custom Connectors — a service whose connection `type` starts with `custom` and which carries a `sourcePythonClass` — are supported for seven service types, but no implementation or test ships in the repo. There is no example to copy, and no test that would catch a regression in the dynamic-import path (`metadata/utils/importer.py`, `metadata/workflow/ingestion.py`).

Because a custom connector reads no external system, a test for one needs only a running OpenMetadata server: no testcontainers, no credentials. That makes this the cheapest connector coverage available, and it currently does not exist.

Two defects surfaced while investigating:

- `import_from_module` catches `ModuleNotFoundError` and `ImportError`, but `getattr` raises `AttributeError` when the module imports and the name is absent. A typo in the class part of `sourcePythonClass` therefore escapes as a bare `AttributeError` and bypasses the `DynamicImportException` branch that exists specifically to surface the real cause for custom connectors.
- `shouldTestConnection` in the UI denies seven `Custom*` service types but omits `CustomSearch`, so Custom Search alone shows a Test Connection button.

IceS2:
### Describe your changes:

Fixes #30899

Adds one Custom Connector for each service type whose connection schema carries `sourcePythonClass`, plus `CustomDrive`, and an integration test suite covering all eight.

A custom connector reads no external system, so these tests need a running OpenMetadata server and nothing else — no testcontainers. The whole suite runs in about 9 seconds.

`ingestion/tests/integration/custom_connectors/`

| Module | Service type | Produces |
|---|---|---|
| `custom_database.py` | CustomDatabase | database, 2 schemas, 3 tables with columns, a classification and tag, a stored procedure, table-to-table lineage |
| `custom_dashboard.py` | CustomDashboard | 2 charts, a data model, a dashboard linking both |
| `custom_messaging.py` | CustomMessaging | 2 topics with a message schema and sample data |
| `custom_mlmodel.py` | CustomMlModel | a model with features, hyper parameters and a store |
| `custom_pipeline.py` | CustomPipeline | a pipeline with 3 chained tasks and an execution status |
| `custom_search.py` | CustomSearch | 2 search indexes with typed fields |
| `custom_storage.py` | CustomStorage | a root container and a child carrying a data model |
| `custom_drive.py` | CustomDrive | a directory with a file, a spreadsheet with a worksheet |

Each subclasses `metadata.ingestion.api.steps.Source` and yields `Either(right=<CreateXRequest>)`, which is the documented recipe. Between them they exercise the sink paths a minimal connector never reaches: `OMetaTagAndClassification`, `OMetaFQNLineageRequest` behind a `Barrier`, `CreateDashboardDataModelRequest`, `OMetaTopicSampleData` and `OMetaPipelineStatus`. Every entity, column, field and task carries a description, and each connector builds its own `CreateXServiceRequest` so the service gets a description and displayName — `get_create_service_from_source` sets neither.

Two fixes the work surfaced:

- `import_from_module` caught `ModuleNotFoundError` and `ImportError` but not the `AttributeError` `getattr` raises when a module imports and the name is absent. A typo in the class part of `sourcePythonClass` escaped raw and bypassed the `DynamicImportException` branch that exists to surface it for custom connectors.
- `shouldTestConnection` denied seven `Custom*` service types but omitted `CustomSearch`, so Custom Search alone offered a Test Connection button.

Note on `CustomDrive`: `customDriveConnection.json` does not declare `sourcePythonClass`. It resolves only because the schema sets `additionalProperties: true`, so the connector works from the CLI and API but cannot be configured from the UI. Left as-is here; worth a separate schema change.

#
### Type of change:
- [x] Improvement
- [x] Bug fix

### DIFF
--- a/ingestion/src/metadata/utils/importer.py
+++ b/ingestion/src/metadata/utils/importer.py
@@ -130,7 +130,9 @@ def import_from_module(key: str, log_traceback: bool = True) -> Type[Any]:  # no
     try:
         obj = getattr(importlib.import_module(module_name), obj_name)
         return obj  # noqa: RET504, TRY300
-    except (ModuleNotFoundError, ImportError) as err:
+    # AttributeError covers an importable module missing the requested name, e.g. a
+    # typo in a custom connector's `sourcePythonClass`.
+    except (ModuleNotFoundError, ImportError, AttributeError) as err:
         if log_traceback:
             logger.debug(traceback.format_exc())
         raise DynamicImportException(module=module_name, key=obj_name, cause=err)  # noqa: B904
--- a/ingestion/src/metadata/utils/importer.py
+++ b/ingestion/src/metadata/utils/importer.py
@@ -128,11 +128,18 @@ def import_from_module(key: str, log_traceback: bool = True) -> Type[Any]:  # no
     logger.debug("Importing: %s", key)
     module_name, obj_name = key.rsplit(MODULE_SEPARATOR, 1)
     try:
-        obj = getattr(importlib.import_module(module_name), obj_name)
+        module = importlib.import_module(module_name)
+    except (ModuleNotFoundError, ImportError) as err:
+        if log_traceback:
+            logger.debug(traceback.format_exc())
+        raise DynamicImportException(module=module_name, key=obj_name, cause=err)  # noqa: B904
+
+    # Scoped to the lookup alone: an AttributeError raised by the module's own top-level
+    # code must keep its original traceback rather than be reported as a missing name.
+    try:
+        obj = getattr(module, obj_name)
         return obj  # noqa: RET504, TRY300
-    # AttributeError covers an importable module missing the requested name, e.g. a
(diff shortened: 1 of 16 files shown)


IceS2:
Both review points were correct and are addressed in 35f7e5e72f.

**`import_from_module` wrapping import-time `AttributeError`** — right, `importlib.import_module` executes the module body, so the broadened `except` could have reported a genuine bug inside a connector as "Cannot import <module>.<name>". The import and the attribute lookup are now separate `try` blocks, so only a missing name is wrapped and an `AttributeError` from the module's own code keeps its traceback. Added a unit test asserting it propagates unwrapped.

**Fixture cleanup ordering** — also right. The service is now registered for teardown before the workflow runs, since a run that fails partway may still have created it.

pmbrull (approved):

ShaileshParmar11 (approved):

PR #30902, opened 2026-08-03
Honour the service filter when listing tables, schemas and stored procedures

### LINKED ISSUE #30900: service filter is silently ignored when listing tables, database schemas and stored procedures
`GET /v1/tables?service=X`, `GET /v1/databaseSchemas?service=X` and `GET /v1/storedProcedures?service=X` return every entity in the instance instead of the ones under `X`.

The three resources never declare a `service` `@QueryParam`. JAX-RS discards an undeclared query parameter without an error, so the caller gets a 200 with unfiltered results — indistinguishable from a filter that matched everything.

The filtering itself was never missing: `ListFilter.getCondition()` calls `getServiceCondition()` unconditionally for every entity, and it resolves to `fqnHash LIKE '<hash(service)>.%'`, which is correct for these entities because their FQNs are service-prefixed. The Java SDK compounds the surprise — `ListParams.setService()` has always put `service` in the query string, so `client.tables().list(params.setService(x))` is a documented call that quietly returns everything.

Measured on a local instance:

```
tables               service=388    all=388    -> IGNORED
databaseSchemas      service=101    all=101    -> IGNORED
storedProcedures     service=17     all=17     -> IGNORED
databases            service=1      all=79     -> HONOURED
topics               service=2      all=19     -> HONOURED
dashboards / charts / mlmodels / pipelines / searchIndexes /
containers / drives.directories                -> HONOURED
```

The three affected endpoints are exactly the database-hierarchy levels below `databases`, which works.

IceS2:
### Describe your changes:

Fixes #30900

`GET /v1/tables?service=X`, `/v1/databaseSchemas?service=X` and `/v1/storedProcedures?service=X` returned every entity in the instance rather than the ones under `X`.

The three resources never declared a `service` `@QueryParam`, and JAX-RS discards an undeclared query parameter silently — so the caller got a 200 with unfiltered results, which is indistinguishable from a filter that matched everything.

The filtering machinery was already there. `ListFilter.getCondition()` calls `getServiceCondition()` unconditionally for every entity, resolving to `fqnHash LIKE '<hash(service)>.%'`, which is correct for these entities because their FQNs are service-prefixed. Only the plumbing that feeds it was missing. `ListParams.setService()` in the Java SDK has always sent the parameter, so `client.tables().list(params.setService(x))` was a documented call that quietly returned everything.

The fix is two lines per resource: declare the parameter, forward it into the filter. `ListFilter` is unchanged.

Adds a regression IT per resource, and unit tests asserting the condition and bind value are generated for `table_entity`, `database_schema_entity` and `stored_procedure_entity`.

Verified against a local instance by running the exact generated SQL: 3 of 388 tables, 2 of 101 schemas and 1 of 17 stored procedures matched — precisely the entities under the target service.

#
### Type of change:
- [x] Bug fix

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DatabaseSchemaResourceIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DatabaseSchemaResourceIT.java
@@ -20,7 +20,6 @@ import org.apache.commons.csv.CSVRecord;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
-import org.openmetadata.it.factories.DatabaseSchemaTestFactory;
 import org.openmetadata.it.factories.DatabaseServiceTestFactory;
 import org.openmetadata.it.util.SdkClients;
 import org.openmetadata.it.util.TestNamespace;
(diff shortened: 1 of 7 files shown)


IceS2:
Good catch on `DatabaseSchemaResourceIT`, fixed in d6f617ea63.

`ns.prefix(base)` is fixed for a test method, so `DatabaseSchemaTestFactory.createSimple(ns, service)` really did reuse the same database and schema names on the second call — the test would not have produced the two schemas the assertion expects. It now creates one database and two schemas with distinct names.

pmbrull (approved):

PR #30933, opened 2026-08-04
fix(test): simplify SearchExport tray step and use fresh auth for AUT polling

anuj-kumary:
## Summary

Fixes two failures in the AUT 1.x → 2.0 nightly for `SearchExport.spec.ts:411 › Export queues a background job and downloads from the jobs tray`.

- **Auto-open tray**: The tray now opens automatically when a job starts or reaches a terminal state. The previous conditional `launcherButton.or(trayPopover)` + gated click logic is no longer needed — and is racy when the job completes before the test reaches the step. Replaced with a direct `expect('.csv-jobs-tray-popover').toBeVisible()`.

- **401 → TypeError**: `waitForExportJobCompleted` was called with `page.request`, which carries the page's cookies but not the bearer token. After a migration the server returns `{code:401, message:"Not Authorized"}` (an object, not an array), surfacing as `TypeError: jobs.find is not a function`. Two fixes:
  1. Added an `Array.isArray` guard that throws a descriptive error (status + body) instead of a cryptic TypeError.
  2. Replaced `page.request` with `createAdminApiContext()` so every poll uses a fresh bearer token, immune to stale cached worker contexts.

## Test plan
- [ ] AUT 2.0 nightly: `SearchExport.spec.ts › Export queues a background job and downloads from the jobs tray` passes
- [ ] Other SearchExport tests unaffected

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/SearchExport.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/SearchExport.spec.ts
@@ -427,52 +427,20 @@ test.describe('Search Export', { tag: ['@Features', '@Discovery'] }, () => {
 
     const jobId = await startAsyncExport(page);
 
-      await test.step('Jobs tray surfaces the export job', async () => {
-        // PR #30615 added a useEffect that calls setOpen(true) when a job
-        // reaches a terminal state, auto-opening the tray and hiding the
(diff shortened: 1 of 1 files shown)


ShaileshParmar11 (approved):

anuj-kumary:
Reopening from upstream branch to fix fork CI restriction

PR #30968, opened 2026-08-05
test(ui): let core-component mocks fall through to the real module

chirag-madlani:
## Why

`jest.mock('@openmetadata/ui-core-components', factory)` replaces the module **wholesale**, so any export the factory omits comes back `undefined`.

**124** test files mock it that way; only **6** spread `jest.requireActual`. That makes every component sweep a tripwire for suites that never referenced the new component.

Typography and Button escaped it only because those names were already listed in the mocks. The layout sweep introduces `Grid`/`Box`, and `Grid.Item` immediately throws `Cannot read properties of undefined (reading 'Item')`:

| | failing |
|---|---|
| before | **22 suites / 267 tests** |
| after | **4 suites / 5 tests** |

## What

A new codemod (`transforms/core-mock-require-actual.js`, 6 tests) spreads `jest.requireActual` as the **first** property, so explicit overrides still win while unlisted exports resolve to the real implementation. Applied across **121** files.

It deliberately leaves alone: mocks already spreading `requireActual`, mocks of other modules, and factories that don't return an object literal.

## Also included

Merges a duplicate `CoreTypography` import in `AdminPermissionDebugger`. esbuild and tsc both accept it — TypeScript tolerates re-importing the same symbol from the same module — so nothing was broken. But jscodeshift's spec-compliant parser rejects the file, which blocks every codemod run over it.

## Verification

- 121 files transformed, prettier + eslint clean
- Full suite: **1154/1155 suites, 13,641 tests passing**
- The single failure (`ExploreTree`) is test-isolation flake, not this change: it passes standalone both with and without the fix (17/17 each), contains no core-components mock at all, and wasn't among the failures in either earlier full-suite run.

## Why it's split out

This is independently valuable and self-contained. The layout sweep that surfaced it needs more work — core `Box` is always a flex container, so mapping `Col` → `Box` makes every column its own flex container. That's being reworked separately.

Collate has the same defect in **188** files; the same transform applies there once Collate unparks.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItemV2/DestinationFormItemV2.test.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItemV2/DestinationFormItemV2.test.tsx
@@ -99,6 +99,8 @@ jest.mock('@openmetadata/ui-core-components', () => {
   Grid.Item = GridItem;
 
   return {
+    ...jest.requireActual('@openmetadata/ui-core-components'),
+
     Button: ({
       onPress,
       children,
@@ -114,9 +116,11 @@ jest.mock('@openmetadata/ui-core-components', () => {
         {children}
       </button>
     ),
+
     Card,
     Divider: () => <hr />,
     Grid,
+
     Input: ({
       onChange,
       value,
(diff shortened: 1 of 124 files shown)


chirag-madlani:
Closing while the migration is on hold. Branch `antd-migration/core-mock-requireactual` is preserved.

Worth keeping in view when this reopens: it is the fix that unblocks every remaining sweep. 124 test files mock `@openmetadata/ui-core-components` with a partial factory and only 6 spread `requireActual`, which is why the layout sweep produced **267** failures — reduced to 5 by this change. Collate has the same defect in 188 files.

PR #30974, opened 2026-08-05
Fixes #30973: key BigQuery dataset and table object caches per schema

### LINKED ISSUE #30973: BigQuery: dataset descriptions and labels are ingested one dataset off
### Affected module
Ingestion — BigQuery connector

### Describe the bug

BigQuery dataset descriptions and dataset labels are stored against the **wrong schema** in OpenMetadata: schema *i* receives schema *i-1*'s description and labels. Reported by a customer running daily single-way ingestion with **Override Metadata** enabled.

Observed vs. BigQuery source:

| schema | BigQuery description | Ingested description |
|---|---|---|
| clean_identity_resolution | Clean Identity Resolution data | *(empty)* |
| clean_mac | Clean MAC data | Clean Identity Resolution data |
| clean_recon | Clean recon data | Clean MAC data |
| clean_transunion | Clean TransUnion data | Clean recon data |
| curated_dimension | Curated dimension data | Clean TransUnion data |

The same shift applies to dataset labels, so schemas end up tagged with tags that do not exist on them in BigQuery.

### Root cause

`BigquerySource.get_dataset_obj()` is documented as "per-schema caching" but is a single-slot cache with **no key** — once `_current_dataset_obj` is set it ignores `schema_name` and returns the previously fetched dataset:

```python
def get_dataset_obj(self, schema_name: str):
    """Get dataset object with per-schema caching"""
    if self._current_dataset_obj is None:
        database = self.context.get().database
        self._current_dataset_obj = self.client.get_dataset(f"{database}.{schema_name}")
    return self._current_dataset_obj
```

The only invalidation is in `query_table_names_and_types()` — the **table** node's producer. The `databaseSchema` topology node runs `yield_database_schema_tag_details` and `yield_database_schema` *before* its table child node, so the reset always fires after that schema's description and tags were already emitted, and the table/stored-procedure stage then refills the slot with the current schema. The next schema therefore reads the previous one's dataset object.

Introduced in #25471 (BigQuery Ingestion Optimization), which replaced the previous `BIGQUERY_SCHEMA_DESCRIPTION` SQL lookup with this cache.

Two aggravating factors:

- **`overrideMetadata` gates visibility.** `RESTRICT_UPDATE_LIST` covers `description` and `tags`, so with override off only ADD operations apply and the corruption is largely masked. With it on, REPLACE is allowed and every daily run rewrites the wrong value.
- **The `databaseSchema` node is `threads=True`.** `_current_dataset_obj` and `_current_schema_tables` are unsynchronized instance attributes, so at `threads > 1` the contamination is arbitrary rather than a clean off-by-one. `_current_schema_tables` is additionally keyed by bare `table_name`, so same-named tables in concurrently-processed schemas can swap label sets.

### To Reproduce

1. Configure a BigQuery service over a project with several datasets that have descriptions and labels.
2. Enable `overrideMetadata` and `includeTags` (any of `includeDDL`, lifecycle, or stored procedures being enabled is what refills the cache).
3. Run metadata ingestion.
4. Compare schema descriptions in OpenMetadata against `INFORMATION_SCHEMA.SCHEMATA_OPTIONS` — each is off by one dataset.

### Expected behavior

Each database schema is ingested with its own dataset's description and labels.

### Version
- OS: any
- Python version: 3.10 / 3.11
- OpenMetadata version: present on `main`; regression since #25471
- OpenMetadata Ingestion package version: same

ulixius9:
### Describe your changes:

Fixes #30973

I keyed the BigQuery `get_dataset_obj` and `get_table_obj` caches by fully qualified name because both were effectively unkeyed, so every database schema was ingested with the *previous* dataset's description and labels.

`get_dataset_obj` was documented as "per-schema caching" but held a single slot with no key — once `_current_dataset_obj` was set it ignored `schema_name` entirely. Its only invalidation lived in `query_table_names_and_types()`, the **table** node's producer, which runs *after* the `databaseSchema` node has already emitted that schema's tags and description; the table/stored-procedure stage then refilled the slot with the current schema, so the next schema read the previous one's dataset object. Regression from #25471.

Both caches now use the existing thread-safe bounded `LRUCache` (`metadata/utils/lru_cache.py`), and the manual invalidations are gone: they caused the off-by-one and, being unsynchronized instance attributes, could wipe a concurrent thread's entry while the schema node runs `threads=True`. `get_table_obj` is keyed by `project.dataset.table` (was bare `table_name`), so same-named tables in different schemas no longer share a cached object. Fixing the keying rather than moving the reset earlier is deliberate — moving it would leave the thread-safety hole. The per-call API savings the original optimization added are preserved: `get_dataset` is still called exactly once per schema.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change (2 files, one connector).

#
#### Use cases covered
- Ingesting a BigQuery project with multiple datasets stores each schema's own description, instead of the previous dataset's
- Ingesting with `includeTags` stores each schema's own dataset labels, instead of inheriting tags that do not exist on that dataset in BigQuery
- Two schemas in flight concurrently (the schema node is multi-threaded) keep their own descriptions
- Two same-named tables in different schemas keep their own labels
- Repeat lookups for the same schema/table still hit the cache (no extra BigQuery API calls)

#### Unit tests
- Files updated: `ingestion/tests/unit/topology/database/test_bigquery.py` — new `TestBigqueryPerSchemaCaching` (6 tests)
- Each of the 4 bug tests was confirmed failing before the fix, with the exact off-by-one shift, e.g. `clean_mac` → `"Clean Identity Resolution data"` and `clean_recon` → `"Clean MAC data"`. The 2 remaining tests guard the caching behaviour itself so a future "just drop the cache" fix cannot regress the API-call savings.
- Also updated 6 pre-existing tests that set `_current_dataset_obj = None` to call `_dataset_obj_cache.clear()`; left alone they would have become silent no-ops that stop invalidating.
- Coverage: **100% of the changed lines** (`__init__` cache setup, `get_dataset_obj`, `get_table_obj`) are covered — verified with `--cov-report=term-missing`. Module-wide coverage of `bigquery/metadata.py` is 38%, which is pre-existing and unchanged by this PR.
- `test_bigquery.py`: 32 passed. All four BigQuery test files: 81 passed.

Not added — the fix is in per-schema cache keying, which the unit tests cover deterministically by driving the real `BigquerySource` through the topology call order. An integration test would need a live multi-dataset BigQuery project.

#### Playwright (UI) tests

#### Manual testing performed
No live BigQuery run — I do not have access to the reporter's project. Verification was a standalone reproduction script that drives the real `BigquerySource` through the topology's call order (schema stages → table producer) against a mocked `bigquery.Client`:

1. Before the fix: 4/5 schemas received the wrong description, reproducing the reporter's screenshot value-for-value.
2. After the fix: 0/5 wrong, and `client.get_dataset` is called exactly once per schema.
3. Also confirmed the buggy behaviour returns by temporarily reinstating the unkeyed lookup, so the new tests are known to fail against the old code.

#

Not applicable.

#
### Checklist:

#
### Notes for the reviewer

Two related items I deliberately left out of this PR, happy to fold either in:

- `_table_ddl_cache` has the identical defect — keyed by bare `table_name` and `clear()`ed per schema — so at `threads > 1` it can serve one schema's DDL for a same-named table in another. Outside the reported symptoms; same one-line keying change.
- `BIGQUERY_SCHEMA_DESCRIPTION` in `bigquery/queries.py` is the correct pre-#25471 SQL lookup and is now dead code.

Deploying this fix stops new corruption, but existing wrong values need one re-ingestion with Override Metadata enabled to be rewritten. The phantom tags will **not** clear themselves — `overrideMetadata` deliberately skips REMOVE operations on `tags`, so already-attached wrong tags need a manual cleanup pass.

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/bigquery/metadata.py
@@ -113,9 +113,15 @@ from metadata.utils import fqn
 from metadata.utils.credentials import GOOGLE_CREDENTIALS
 from metadata.utils.helpers import retry_with_docker_host
 from metadata.utils.logger import ingestion_logger
+from metadata.utils.lru_cache import LRUCache
 from metadata.utils.sqlalchemy_utils import is_complex_type
 from metadata.utils.tag_utils import get_ometa_tag_and_classification, get_tag_label
 from metadata.utils.tag_utils import get_tag_labels as fetch_tag_labels_om
+
+# The databaseSchema node runs multi-threaded, so these caches are shared across schemas
+# being processed concurrently and must be keyed by the fully qualified name.
+DATASET_OBJ_CACHE_SIZE = 512
+TABLE_OBJ_CACHE_SIZE = 2048
 
 _bigquery_table_types = {
     "BASE TABLE": TableType.Regular,
@@ -241,8 +247,8 @@ class BigquerySource(LifeCycleQueryMixin, CommonDbSourceService, MultiDBSource):
         self.incremental = incremental_configuration
         self.incremental_table_processor: Optional[BigQueryIncrementalTableProcessor] = None  # noqa: UP045
 
-        self._current_schema_tables = {}
-        self._current_dataset_obj = None
+        self._table_obj_cache: LRUCache = LRUCache(capacity=TABLE_OBJ_CACHE_SIZE)
+        self._dataset_obj_cache: LRUCache = LRUCache(capacity=DATASET_OBJ_CACHE_SIZE)
         self._policy_tag_cache = {}
         self._taxonomy_cache = {}
         self._taxonomy_to_tags = {}
@@ -366,8 +372,6 @@ class BigquerySource(LifeCycleQueryMixin, CommonDbSourceService, MultiDBSource):
         database = self.context.get().database
         dataset_ref = f"{database}.{schema_name}"
 
-        self._current_schema_tables.clear()
-        self._current_dataset_obj = None
         self._prefetch_table_ddls(schema_name)
         clear_constraint_cache_for_schema(database, schema_name)
 
@@ -416,11 +420,19 @@ class BigquerySource(LifeCycleQueryMixin, CommonDbSourceService, MultiDBSource):
         return super().get_table_description(schema_name=schema_name, table_name=table_name, inspector=inspector)
 
     def get_dataset_obj(self, schema_name: str):
-        """Get dataset object with per-schema caching"""
-        if self._current_dataset_obj is None:
-            database = self.context.get().database
-            self._current_dataset_obj = self.client.get_dataset(f"{database}.{schema_name}")
-        return self._current_dataset_obj
+        """Get dataset object with per-schema caching.
+
+        Keyed by `project.dataset`: the schema node emits tags and the schema entity before
+        its table child node runs, so a cache that is not keyed hands one schema's
+        description and labels to the next one.
+        """
+        dataset_ref = f"{self.context.get().database}.{schema_name}"
+        if dataset_ref in self._dataset_obj_cache:
+            return self._dataset_obj_cache.get(dataset_ref)
+
+        dataset_obj = self.client.get_dataset(dataset_ref)
+        self._dataset_obj_cache.put(dataset_ref, dataset_obj)
+        return dataset_obj
 
     def yield_life_cycle_data(self, _) -> Iterable[Either[OMetaLifeCycleData]]:
         """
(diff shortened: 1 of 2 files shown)


ulixius9:
Thanks both — @copilot-pull-request-reviewer and @gitar-bot independently flagged the same check-then-get race, and it's a valid finding. Fixed in d100bdbba0.

Confirming the mechanism: `LRUCache.__contains__` and `LRUCache.get` each acquire `self.lock` separately, and `get` calls `OrderedDict.move_to_end(key)`, which raises `KeyError` when the key is absent. So a concurrent `put` that trips `popitem(last=False)` in that window makes the read raise. It matters because `yield_database_schema` calls `get_dataset_obj` outside a try block, so a `KeyError` there would fail the entire schema rather than degrade.

Both methods now fetch once and catch `KeyError`:

```python
try:
    return self._dataset_obj_cache.get(dataset_ref)
except KeyError:
    pass
```

I kept the fix local rather than adding a `get_or_none`/`get_or_default` to `metadata/utils/lru_cache.py` — that util is shared with the lineage mixin, Trino profiler and Databricks ownership, so widening its API surface for this seemed like the larger change. Happy to move it there if a maintainer would rather have the atomic accessor available generally, since the same contains-then-get pattern exists in `LRUCache.wrap`.

Added a regression test per cache (`test_dataset_obj_survives_entry_evicted_before_read`, `test_table_obj_survives_entry_evicted_before_read`). Real threads can't hit this window reliably, so the tests substitute a cache subclass whose `get` raises `KeyError` to drive the eviction window deterministically; both fail with `KeyError` against the previous commit and pass now. `test_bigquery.py` is at 34 passing.

One note on severity for anyone reading the thread: the window is narrower than it looks, because `__contains__` itself does `move_to_end(key)`, promoting the key to most-recently-used, while eviction pops the least-recently-used. Evicting it would have required ~512 intervening puts inside that window. Still worth fixing — it's a real regression risk versus the single-slot code it replaced, and the fix also removes one lock acquisition per cache hit.

mohittilala (approved):

Issue #30941, opened 2026-08-04
Bulk edit silently drops tags after the antd Button migration


chirag-madlani:
## Summary

After the antd `Button` → `ui-core-components` migration (#30810), a tag selected
in the **bulk edit** grid is never applied. The import runs and reports success,
but the tag is missing from the entity afterwards.

**User impact:** silent data loss. Someone bulk-editing assets sets a tag, sees no
error, and the tag simply isn't there.

## Evidence

From the Playwright trace of `DataAssetRulesDisabled.spec.ts` → *Bulk Edit Actions
→ Database service*:

- The bulk edit **does** run — both `importAsync?dryRun=true` (preview) and
  `importAsync?dryRun=false` (apply) fire.
- `Sensitive` appears **only** in a JS bundle and the tag-search response. It
  **never appears in the import payload**.
- The post-edit screenshot shows the service with empty Tags and Tier.

So the tag is chosen in the dropdown but never committed into the grid cell, and
the CSV is built without it.

## Confirmed a regression, not flake

Same shards on #30780 (Typography, merged green) — same pipeline, same seeded
data, one commit earlier:

| shard | #30780 |
|---|---|
| data-asset-rules-01 | pass (10m56s) |
| ingestion-01 | pass (15m14s) |
| ingestion-02 | pass (17m59s) |
| reindex-01 | pass (19m32s) |

They fail on the Button branch consistently across two reruns of the same commit,
and pass on unrelated PRs (e.g. #30469, full 105-check run). One shard,
`chromium-01`, was genuinely flaky and now passes.

## Ruled out (tested, not reasoned)

- **Core `Button` swallows `onClick`** (react-aria `usePress` stops propagation —
  the #30804 mechanism). Headless probe: `onClick` fires, including with
  `force: true`.
- **Core `Button` defaults to `type="submit"`** — it defaults to `"button"`.
- **antd `Tooltip` wrapping a core `Button` crashes** — renders fine.
- **`InlineEdit` prop precedence** — was a real bug (converted props emitted after
  the `{...saveButtonProps}` spread, clobbering caller overrides) and is fixed in
  #30810, but it did not change the shard results.

## Where to look next

The commit path between selecting the tag and the grid cell holding it —
`fillTagDetails` → `clickInlineSave` (`inline-save-btn`) in
`playwright/utils/importUtils.ts`, against `InlineEdit.component.tsx` and the
bulk-edit grid's tag cell editor.

[...]

PR #30999, opened 2026-08-05
refactor(ui): antd → ui-core-components, wave 1 (Typography + Button)

chirag-madlani:
Wave 1 of the antd → `@openmetadata/ui-core-components` migration. **607 files.**

Draft until the two blockers below are cleared.

## Scope

| Component | before | after |
|---|---|---|
| `Typography` | 442 | **0** |
| `Button` | 310 | **0** |

QueryBuilder is excluded throughout — owned by #29849.

## Blockers before this can merge

- **#30941 — bulk edit silently drops tags.** A real data-loss bug from the Button sweep: the CI trace shows both `importAsync?dryRun=true` and `dryRun=false` firing, but the selected tag never reaches the import payload. Three e2e shards (`data-asset-rules-01`, `ingestion-01`, `reindex-01`) are red for this. They pass on the Typography-only commit and on unrelated PRs, so it is ours.
- **#30968** — core-component mock fix, should land in wave-1 first.

## Product bugs found and fixed along the way

None of these were visible to jest — jsdom computes no layout — so all three were caught by CI, a browser probe, or review:

1. **`div.prose` block wrapper.** Core `Typography` wrapped everything in a block-level `<div>`, turning every former inline `Typography.Text` into a block element and producing invalid `<div>`-inside-`<span>` nesting. Fixed in core by dropping the wrapper for `span`/`div`, where `.prose`'s descendant selectors target nothing — verified against `typography.css` rather than assumed.
2. **`data-icon` zero-sized icons.** Icons passed as elements got no sizing at all, collapsing every icon-only button. Measured in Chrome: **36×36 with the fix, 16×16 with a 0×0 icon without.** This would have shipped invisible edit buttons (edit-tier, edit-owner, edit-description) across the product.
3. **`InlineEdit` prop precedence.** The codemod emitted converted props *after* `{...saveButtonProps}`, inverting precedence so caller overrides were silently clobbered.

## Manual validation focus

Automated coverage is green (13,641 jest tests; visual baselines refreshed and each diff compared against its predecessor). What it cannot cover:

- **Inline text flow** — bold segments mid-sentence in activity feeds, form labels beside links. The `prose` wrapper fix changes these deliberately.
- **Icon-only buttons** at every size — the `data-icon` fix is the highest-risk change here.
- **Buttons inside antd containers** — `Tooltip`, `Dropdown`, `Space`, table action columns.
- **Bulk edit** — known broken, see #30941.

## Not in this wave

`Row`/`Col`/`Space` (~1,600 sites) is parked. Core `Box` is always a flex container, so mapping `Col` → `Box` would make every column its own flex container when antd's `Col` is a flex *item*. Needs rework before sweeping.

`.ant-*` selector occurrences still stand at **2103** — component sweeps do not remove them, and they are what actually blocks dropping the antd package.


chirag-madlani:
Closing while the migration is on hold. Nothing is lost — `antd-migration/wave-1` still carries the full Typography + Button sweep (both at 0 antd imports upstream) and the branch is untouched. Reopen when the wave resumes.

Outstanding against this branch when it does: **#30941** (bulk edit silently drops tags — a real regression from the Button sweep) and the unadopted `add-team-form` visual baseline.

PR #31001, opened 2026-08-05
chore(ui): remove four unreferenced modules (263 lines)

chirag-madlani:
First pass of the dead-code cleanup that runs alongside the antd migration.

## Removed — 263 lines

knip reported eight unused files. Only these four survived verification as genuinely unreferenced across `src`, tests **and** `playwright`:

- `src/utils/DataAssetsHeaderExtraInfo.tsx`
- `src/utils/DataAssetsHeaderTypeUtils.ts`
- `src/utils/EntityTagUtils.ts`
- `src/rest/queries/glossaryQuery.ts`

## Deliberately kept — knip is wrong about these

**`SSOUtilClassBase.ts`** — production code no longer imports it, but `SettingsSso.test.tsx` still `jest.mock`s the path. knip ignores test files so it can't see that; deleting the module would break the mock. Wants the stale mock removed in the same change.

**Three `index.ts` barrels** (`ColorPicker`, `atoms/ProfilePicture`, `atoms/TagChip`) — knip flags the barrel because consumers import the component file directly rather than through it. `ProfilePicture` alone has **62** references. Needs per-consumer checking, not a blanket delete.

## A note on the verification

My first check reported "0 references" for all eight, which would have been a green light to delete the lot. It was wrong: I grepped for `DataAssetsHeaderExtraInfo.tsx` **including the extension**, which import specifiers never contain. Re-running against the module specifier is what surfaced `ProfilePicture`'s 62 consumers.

Worth recording because the same trap applies to the rest of this cleanup: knip's output is a list of *candidates*, not a list of deletions.

## Verification

Full suite: **1155 suites / 13,642 tests, no failures.**

## Not in this PR

604 unused exports and 357 unused types remain. Those need a more careful pass — that's where knip's false-positive rate really bites, since an export used only by a test, a lazy import, or a barrel looks identical to a dead one.


chirag-madlani:
Closing while the migration is on hold. Branch `antd-migration/dead-code-1` is preserved.

Content stands: 4 verified-unreferenced modules, 263 lines, full suite green. The other 4 knip candidates were deliberately excluded — `SSOUtilClassBase` is still referenced by a test that mocks it, and the three `index.ts` barrels have consumers importing the component directly (`ProfilePicture` alone has 62).

PR #31017, opened 2026-08-05
fix(workflows): eliminate null-businessKey race + silence non-OM log noise

yan-3005:
## Summary

Two independent fixes that showed up together in the Aug-5 AUT PostgreSQL 1.11.13 → 1.13 nightly (Loki namespace `aut-30964402587`): the `MainWorkflowTerminationListener` was NPE-ing every ~10s at ERROR, and the `[STAGE_SKIP]` / `[STAGE_UPDATE_NO_ID]` pair was flooding logs at WARN/ERROR on Flowable-internal transitions.

Both share the same trigger: process instances with a null `businessKey`. The root cause is in this PR (commit 2). Commit 1 handles the legacy tail and log noise.

## Root cause (commit 2 — `fd80276dc8`)

The periodic + legacy fallback paths in `WorkflowHandler.triggerWorkflow` called `startProcessInstanceByKey(key)` with no `businessKey`. Design relied on `WorkflowInstanceListener.updateBusinessKey` firing at the trigger's `startEvent` to backfill a fresh UUID.

Two ways that broke:
1. **Race:** if `WorkflowInstanceListener.execute` threw before reaching line 108 (`Entity.getEntityTimeSeriesRepository` transient failure, DB glitch, etc.), the trigger ran to completion with null `businessKey`. Its `CallActivity` (`EventBasedEntityTrigger.java:170`, `inheritBusinessKey=true`) spawned the MainWorkflow inheriting null. Every downstream listener that did `UUID.fromString(businessKey)` NPE-d.
2. **Silent overwrite:** the listener unconditionally re-assigned `UUID.randomUUID()`, discarding caller-supplied `businessKey`s from `TaskRepository.triggerByKey` (task id) and `MigrationUtil.triggerByKey` (task id). Task-workflow correlation drifted.

**Fix**
- `WorkflowHandler.triggerWorkflow:1745,1759`: pass `UUID.randomUUID().toString()` on both `startProcessInstanceByKey` calls. Trigger instance has a `businessKey` from the moment it exists — no race window.
- `WorkflowInstanceListener.addWorkflowInstance`: only call `updateBusinessKey` when `businessKey` is absent. Preserves caller-supplied UUIDs and closes the race.

## Log-noise fixes (commit 1 — `758b9768ba`)

Belt-and-suspenders for any legacy pre-inheritance process instance still sitting in Flowable (upgrade rows, force-cancelled test teardowns).

- `MainWorkflowTerminationListener.execute`: null-guard `businessKey`, log at DEBUG and return instead of NPE-ing the catch block at ERROR.
- `WorkflowInstanceStageListener`: `[STAGE_SKIP]` WARN → DEBUG; `[STAGE_UPDATE_NO_ID]` ERROR → DEBUG. Both fire on the same non-OM-managed instances flagged by the businessKey check.
- `WorkflowFailureListener`: add `Workflow definition deleted` to `INTENTIONAL_CANCELLATION_CAUSES`. On a workflow-definition hard-delete with N running instances, we were logging N × PROCESS_CANCELLED WARNs — now silent, matching the other intentional teardown causes already whitelisted.

## Behavior change

- OM-managed workflows: no behavior change. Every existing `triggerByKey` caller already passed a non-null `businessKey`; the listener's overwrite is now skipped for them, which is what the caller expected.
- Periodic/legacy `triggerWorkflow` paths: now have a valid `businessKey` from the start, so the `WorkflowInstance` row and its stage records get written correctly on the FIRST listener firing instead of racing.
- Non-OM Flowable-internal instances (pre-inheritance legacy rows, force-cancelled test teardowns): no ERROR/WARN spam.

## Loki evidence (aut-30964402587, 01:20-02:00 UTC 2026-08-05)

- `MainWorkflowTerminationListener` NPE: 50+ hits in a 40-min window (all `UUID.fromString(null)`)
- `[STAGE_SKIP]` / `[STAGE_UPDATE_NO_ID]`: paired hits from `WorkflowInstanceStageListener` on the same non-OM instances
- PROCESS_CANCELLED × 17 on a single `pw-user-approval-owners-table-*` workflow deletion (test teardown emits one per running instance)

## Test plan

- [x] `mvn spotless:apply` clean on the 4 touched files
- [ ] `mvn test -pl openmetadata-service -Dtest='WorkflowHandler*Test'` — run in CI
- [ ] Manual: trigger a periodic workflow, verify `WorkflowInstance` and `WorkflowInstanceState` rows land on the first firing (previously required a retry when the listener raced)
- [ ] Manual: delete a WorkflowDefinition with running instances, verify no `PROCESS_CANCELLED` WARN in logs
- [ ] Backport to `1.13` once merged

----
## Summary by Gitar

- **Entity hard-delete workflow cleanup:**
    - Added `cancelInstancesForEntities` batching mechanism and called it from `EntityRepository` hard-delete and bulk hard-delete paths
    - Whitelisted `Entity deleted` and `Workflow definition deleted` as intentional cancellation causes in `WorkflowFailureListener`
- **FilterEntityImpl robustness:**
    - Short-circuits `FilterEntityImpl` with `passesFilter=false` when `relatedEntity` variable is absent or blank


yan-3005:
Workflow Test passed: https://github.com/open-metadata/openmetadata-collate/actions/runs/31038417019

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowInstanceListener.java
```diff
@@ -94,6 +94,19 @@ private void updateBusinessKey(String processInstanceId) {
     WorkflowHandler.getInstance().updateBusinessKey(processInstanceId, workflowInstanceBusinessKey);
   }
 
+  private static boolean isUuid(String value) {
+    boolean valid = false;
+    if (value != null && !value.isBlank()) {
+      try {
+        UUID.fromString(value);
```
sonika-shah, line 101: nit : can we have value != null && !value.isBlank() in UUID.fromString(value); instead of new minor method

sonika-shah (approved):

PR #31019, opened 2026-08-05
fix(search): apply minimum_should_match floor to column index search

mohityadav766:
### Describe your changes:

Fixes open-metadata/openmetadata-collate#3851

Column search (`index=tableColumn`) routes through `buildColumnSearchBuilderV2`, which passed `"0"` as the `multi_match` **fuzziness** argument. `minimum_should_match` was only ever applied inside the `!fuzziness.equals("0")` branch, so the `"0"` sentinel silently dropped the floor — leaving a bare `best_fields`/OR `multi_match` that matches any column sharing a **single** token (e.g. a common parent-name token in `fullyQualifiedName` / `table.name`).

As a result, a specific-FQN column search returned the **entire** column index, while the `dataAsset` aggregation (strict builder, `minimum_should_match(1)`) counted only the true matches — so the Explore **Columns** tab count and the results list diverged (observed: tab count **1**, results **7,381**).

**Fix:** pass the existing `MINIMUM_SHOULD_MATCH` (`"2<70%"`) constant to the `multi_match` in both the OpenSearch and Elasticsearch source-builder factories. The 9-arg overload applies `minimum_should_match` unconditionally (independent of fuzziness). Exact-name searches (1–2 tokens) still match; long FQN-style queries now require ~70% token overlap instead of a single shared token.

Changed:
- `openmetadata-service/.../search/opensearch/OpenSearchSourceBuilderFactory.java` — `buildColumnSearchBuilderV2`
- `openmetadata-service/.../search/elasticsearch/ElasticSearchSourceBuilderFactory.java` — `buildColumnSearchBuilderV2`

#
### Type of change:
- [x] Bug fix

#
#### Unit tests
- [x] Added `SearchSourceBuilderFactoryTest.testColumnIndexSearchAppliesMinimumShouldMatchFloor` — builds the column query for both OpenSearch and Elasticsearch and asserts the serialized query carries `"minimum_should_match":"2<70%"`. Fails on the previous code (floor absent), passes with the fix.
- `SearchSourceBuilderFactoryTest`: 18/18 pass locally.

- [x] Not applicable (no API changes). Existing `ColumnSearchIndexIT` asserts a column is findable by name, which this change preserves — and makes more robust, since the target column is no longer buried beyond `size(10)` among thousands of permissive matches.

#
Not applicable (backend search-query change).

#
### Checklist:

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchSourceBuilderFactory.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchSourceBuilderFactory.java
@@ -381,6 +381,8 @@ public class ElasticSearchSourceBuilderFactory
       queryBuilder = Query.of(q -> q.matchAll(m -> m));
     } else {
       Map<String, Float> fields = ColumnSearchIndex.getFields();
+      // Require a minimum token overlap so a single shared parent token (e.g. in
+      // fullyQualifiedName or table.name) does not match every column in the index.
       queryBuilder =
           ElasticQueryBuilder.multiMatchQuery(
               query,
@@ -388,7 +390,10 @@ public class ElasticSearchSourceBuilderFactory
               TextQueryType.BestFields,
               Operator.Or,
               String.valueOf(DEFAULT_TIE_BREAKER),
-              "0");
+              "0",
+              MINIMUM_SHOULD_MATCH,
+              null,
+              null);
     }
     Highlight hb = buildHighlightsV2(List.of("name", "displayName", "description"));
     return searchBuilderV2(queryBuilder, hb, from, size);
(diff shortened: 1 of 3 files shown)


chirag-madlani (approved):

mohityadav766:
Superseded by #31093.

Tightening `buildColumnSearchBuilderV2` with a `minimum_should_match` floor regressed exact-FQN column lookups to **0 hits** (Playwright `SearchRBAC › Table Column`): a `best_fields` multi_match needs 70% of the query's tokens in a *single* field, but an FQN query's tokens are split across `name` / `table.name` / `fullyQualifiedName`, so no field qualifies.

The real issue isn't matching — it's counting. The Explore tab badge is read from the cross-index `dataAsset` aggregation while the list comes from the tab's own index; #31093 fixes the badge to use the active tab's own results total instead. Closing this in favor of that.

PR #31030, opened 2026-08-05
feat(user): persist per-user UI preferences on the User entity (starting with appMode)

chirag-madlani:
## Summary

Persist the user's app-mode preference (AI vs Classic) — and any future opaque per-user UI preference — on the backend `User` entity, so the choice follows the user across devices, browsers, and profiles instead of only living in localStorage.

## What changed

**Schema — `openmetadata-spec/.../teams/user.json`**
- New optional `preferences: Map<String, Object>` field, opaque to the server, stored inside the existing `user_entity.json` JSON blob. No DB migration. `existingJavaType` directive ensures the Java shape is `Map<String, Object>` rather than a generated wrapper POJO.

**Backend — `openmetadata-service`**
- `UserRepository.clearFields` gates `preferences` behind `fields=preferences`. The field is serialised only when explicitly requested — never in list responses, and never in other users' GETs.
- `UserResource.getCurrentLoggedInUser` forces `preferences` into its fetched field set via `EntityUtil.addField`, so the UI can hydrate on bootstrap without every caller needing to remember the query param.
- Swagger `FIELDS` example updated. `allowedFields` is auto-derived from `@JsonPropertyOrder` on generated `User.java`, so no manual allow-list edit was needed.

**UI — `openmetadata-ui/.../hooks/currentUserStore/useCurrentUserStore.ts`**
- New `BACKEND_SYNCED_KEYS = new Set(['appMode'])` — the whitelist of preference keys that also live on the backend. All other `UserPreferences` fields remain purely local (persisted by the existing `persist` middleware, unchanged).
- `hydrateBackendSyncedPreferences(user)` is called from `AuthProvider` right after `getLoggedInUser()` resolves (both the returning-session bootstrap and the interactive-login paths). It:
  - **Server wins** when the server has a value: local slice is overwritten.
  - **One-shot migration** when the server has no value but local does: emits a JSON-Patch `add` up to the backend.
- `setPreference(patch)` — the consumer-facing hook — is unchanged. Under the hood, writes to whitelisted keys additionally enqueue a 300ms-debounced `updateUserDetail(userId, ops)` JSON-Patch:
  - `null` → `remove`, key not previously on server → `add`, otherwise → `replace`. Coalesces rapid writes into one PATCH.
- On PATCH failure: local state rolls back to the last server-confirmed value and `showErrorToast` fires. No silent divergence between client and server.
- `beforeunload` flushes any pending debounced patch best-effort.
- `persist` middleware is **not** removed — sidebar collapse state, recently-viewed, etc. still live on localStorage as before.

**Types**
- `openmetadata-ui/.../generated/entity/teams/user.ts` regenerated with the new `preferences?` field.

## Testing

- **Java IT tests (`openmetadata-integration-tests/.../UserResourceIT.java`)** — 4 tests added:
  - `patch_preferences_add_appMode_returnsIt` — JSON-Patch `add` round-trip.
  - `get_user_without_fields_omitsPreferences` — field-gating works both directions (without `fields=preferences`, the response omits it; with it, the response carries it).
  - `patch_preferences_replaceOp_persists` — `replace` op semantics.
  - `patch_preferences_removeOp_clearsKey` — `remove` op semantics.
  - `list_users_omitsPreferencesForAllEntries` — list-response omission across a page.

  These tests were **drafted but not executed locally** — they need Docker/MySQL/ES containers that weren't available in the authoring environment. CI on this PR is the first real run.

- **UI Jest tests (`useCurrentUserStore.test.ts`)** — 9 new tests, all passing locally:
  - Hydration overrides local `appMode` when server has a value (server wins).
  - Hydration leaves local alone when neither has a value.
  - `setPreference({ appMode })` fires a debounced PATCH.
  - Rapid writes coalesce into a single PATCH with the last value.
  - Non-whitelisted keys do **not** PATCH but still write locally.
  - `appMode: null` emits a `remove` op.
  - PATCH failure rolls back local state and toasts.
  - Local `appMode` migrates to backend when the server has none.
  - No migration when the server already has a value.

  Full store test suite: 15/15 passing. Pre-existing `AuthProvider.test.tsx` (12 tests) still passing after wiring the bootstrap call.

## Design notes / decisions

- **`preferences` is opaque server-side.** No validation of individual keys or values — the schema/enum lives in the UI. This keeps future preferences (like `connectionsViewMode`, or Collate-only keys) landable without schema changes.
- **P1 whitelist scope.** Only `appMode` is backend-synced in this PR. Other `UserPreferences` fields (sidebar collapse state, recently viewed, marketplace history, etc.) intentionally stay per-device. A future PR can categorize each and expand `BACKEND_SYNCED_KEYS`.
- **Field-gating avoids leaking preferences across users.** Without `fields=preferences`, `GET /users/{id}` and list responses omit the field entirely. Only the loggedInUser endpoint opts in by default.
- **`existingJavaType` on the schema** prevents `jsonschema2pojo` from generating a wrapper `Preferences` POJO with `@JsonAnyGetter`/`@JsonAnySetter`, which would have made the Java handling inconsistent with the "opaque bag" design.
- **Debounce = 300ms**, coalesced per key. Enough to swallow rapid toggle bursts; short enough that a user tabbing away sees the write land.

## Follow-ups (out of scope for this PR)

- Categorize other `UserPreferences` fields and expand the backend-synced whitelist (Collate's `AppModeSwitcher` shipped `appMode` first, other fields will follow if they cross device).
- Consider `preferences` versioning if the shape ever needs breaking changes on the wire.

## Test plan for reviewers

- [ ] CI passes (both the Java IT tests and the UI Jest tests).
- [ ] Manual smoke: log in on browser A, toggle app mode → PATCH `/v1/users/{id}` fires within ~300ms with `op: add|replace|remove` on `/preferences/appMode`.
- [ ] Log in on browser B with same user → `getLoggedInUser` returns `preferences.appMode`; AppModeSwitcher reflects the choice.
- [ ] `GET /v1/users/{id}` (without `fields=preferences`) returns `preferences: null`.
- [ ] `GET /v1/users` list returns every user with `preferences: null`.


mohityadav766 (commented):
This needs more work, we should have another table user_preferences. 
Where each user can have various types of preferences.

chirag-madlani:
Closing in favor of the v2 architecture per reviewer feedback (Pablo + Mohit). Per-user preference storage will move from a field on the User entity to a dedicated `user_preferences` table with its own lightweight resource. The runtime behavior also reshapes: no hard force, soft default with fallback chain user preference → persona → app-level default → constant.

New combined PR forthcoming — will link when opened. Branch `feat/user-preferences-backend` stays on the remote for reference.

PR #31057, opened 2026-08-05
fix(data-quality): pass Azure SQL connection parameters to data diff

### LINKED ISSUE #31044: Table Diff fails for Azure SQL as secondary source with missing connection parameters
### Affected module

Data Quality / Profiler

### Describe the bug

The Table Diff test is expected to support comparing tables across different data sources (e.g., Snowflake ↔ Azure SQL). However, when Azure SQL is configured as one of the sources, the test fails during connection initialization.

### To Reproduce

Error
```
TypeError: MsSQL.__init__() missing 4 required positional arguments:
'host', 'port', 'user', and 'password'
```
Stack trace:
```
File ".../tableDiff.py", line 339, in get_incomparable_columns
table2 = data_diff.connect_to_table(...)
File ".../data_diff/databases/_connect.py", line 210, in connect_to_uri
db = cls(thread_count=thread_count, **kw, **kwargs)
TypeError: MsSQL.__init__() missing 4 required positional arguments:
'host', 'port', 'user', and 'password'
```

### Expected behavior

Table Diff should successfully initialize the Azure SQL connection and perform the comparison between the two data sources.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

1.13.3

_No response_

### Additional context

_No response_

### Pre-submission checklist

IceS2:
Fixes #31044

Azure SQL builds its connection URL with an ODBC connection string, so
data-diff cannot read host, port, user and password from it and fails to
construct the MsSQL client.

The connection is now handed to data-diff as an explicit connection dict —
the same path Trino and Oracle already use — with the per-table database and
schema applied.

Also drops the Snowflake password from the data-diff URL when a private key is
configured, since data-diff rejects receiving both.


ingestion/src/metadata/ingestion/source/database/azuresql/connection.py
```diff
@@ -44,6 +44,8 @@
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
 from metadata.utils.constants import THREE_MIN
 
+DEFAULT_SQL_SERVER_PORT = 1433
```
TeddyCr, line 48: I don't think we would need that right. The connection should fail earlier in the code if the service_connection port section is null right?

ingestion/src/metadata/ingestion/source/database/snowflake/data_diff/data_diff.py
```diff
@@ -34,4 +39,15 @@ def get(
...
+            url = make_url(table_param.serviceUrl)
+            # data-diff refuses a password and a private key at once. The private key wins, as it
+            # does for ingestion, where the connector passes it as a connection argument.
+            if url.password:
+                logger.warning(
+                    "[Data Diff]: Service %s is configured with both a password and a private key. "
+                    "Using the private key.",
+                    service.name.root,
+                )
+                table_param.serviceUrl = url.set(password="").render_as_string(hide_password=False)
```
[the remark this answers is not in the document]
TeddyCr: @IceS2 that sounds relevant

ingestion/src/metadata/ingestion/source/database/snowflake/data_diff/data_diff.py
```diff
@@ -34,4 +39,15 @@ def get(
         connection_config = cast(SnowflakeConnection, service.connection.config)  # noqa: TC006
         table_param.privateKey = connection_config.privateKey
         table_param.passPhrase = connection_config.snowflakePrivatekeyPassphrase
+        if table_param.privateKey and isinstance(table_param.serviceUrl, str):
+            url = make_url(table_param.serviceUrl)
+            # data-diff refuses a password and a private key at once. The private key wins, as it
```
TeddyCr, line 44: Do you see any side effects here to default to private key?

ingestion/src/metadata/ingestion/source/database/snowflake/data_diff/data_diff.py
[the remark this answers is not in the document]
IceS2: The suggested change does not do what it says: SQLAlchemy's `URL.set()` treats `None` as "leave this component unchanged", so `set(password=None)` keeps the password in the url.

```
set(password="")        -> snowflake://my_user:@my_account/...           dsn.password=''
set(password=None)      -> snowflake://my_user:my_password@my_account/...  dsn.password='my_password'
_replace(password=None) -> snowflake://my_user@my_account/...            dsn.password=None
```

The concern behind it is fair though — the empty component was relying on data-diff's Snowflake constructor tolerating `""`, and the test only re-parsed with SQLAlchemy. Both are addressed: the url is now rebuilt with `URL.create` so the password component is absent entirely, and the test asserts through `dsnparse` + `CustomParseResult`, the parser data-diff itself uses.

ingestion/src/metadata/ingestion/source/database/snowflake/data_diff/data_diff.py
IceS2: From what I saw this is similar to what the SQLAlchemy dialect we use for the Connection does, prioritising Private Key over password.

I don't see any regression, it is a choice

ingestion/src/metadata/ingestion/source/database/azuresql/connection.py
IceS2: From what I tested, it actually could accept None and technically we could pass it and it would strip it and works (It only seems to break if we don't pass the port as it receives it as positional arg)

I like the explicit nature here and think about this more as defense in depth

TeddyCr (approved):

PR #31063, opened 2026-08-05
test(e2e): fix 12 chronically flaky Playwright tests

ShaileshParmar11:
Fixes 12 chronically flaky Playwright tests. Every one failed its first attempt and passed on retry, so none of them ever blocked a run — they simply eroded confidence in the suite.

11 files, +273/-48 — **all under `playwright/`, no application code**.

Two of these are already **confirmed fixed in nightlies**: `Toggle fullscreen mode` and `navigation blocker after saving changes` each went from flaky in 16 of 16 runs to passing on the first attempt, on **both** Postgres and MySQL. Both are flaky 9/9 on `main`, so the same result is expected here.

---

## Tests fixed

| Test | Flaky before | Commit |
|---|---|---|
| `InputOutputPorts › Toggle fullscreen mode` | 16/16 | 1 |
| `InputOutputPorts › Remove last port shows empty state` | 16/16 | 1, 6 |
| `NavigationBlocker › should not show navigation blocker after saving changes` | 16/16 | 2 |
| `PlatformLineage › Verify domain platform view` | 15/16 | 3 |
| `PlatformLineage › Verify platform view switching` | 13/16 | 3 |
| `BulkImport › Range selection` | 11/16 | 4 |
| `Entity › Domain Propagation` | 8/16 | 5 |
| `LargeGlossaryPerformance › should search and filter glossary terms` | 6/16 | 7 |
| `CustomProperties › … in right panel` (4 tests × 18 entity types) | 6/16 | 8 |
| `PersonaFlow › Set default persona for team` | 6/16 | 10 |
| `ObservabilityAlerts › Table alert` (+3 siblings) | 5/16 | 11 |
| `ContextCenterDocument › scrolling the bulk "Move" dropdown` | 4/16 | 9 |

---

## Commits

1. **`expandLineageSection` waits on state instead of proxies** — the accordion header is a toggle, but the helper clicked it unconditionally; calling it on an expanded section collapsed it, and since no `/portsView` follows a collapse the subsequent `waitForResponse` hung to the timeout. Also disambiguates the response matcher (the count probe and the lineage fetch share an endpoint) and waits for a terminal render state rather than a loader count that can pass vacuously.
2. **Match the docStore update** — tests here share one persona and run sequentially, so an earlier test creates the layout and this save is an **update**. The wait used a bare `'/api/v1/docStore'`, an exact URL match, which never matches `PUT /api/v1/docStore/{id}`. Note `*` does not cross a `/` in Playwright globs.
3. **Navigate before driving the global search box** — the `page` fixture hands out `browser.newPage()`, which sits on `about:blank`; `TableClass.visitEntityPage` only navigates on its direct-navigation branch, so the fallback drove a search box on a page that had none. Probes for the box rather than checking `page.url()`, since a URL test cannot tell an app page from any other http page.
4. **Re-press arrow keys RDG drops while settling** — react-data-grid discards the keypress outright while re-rendering; focus never moves, so waiting longer cannot help. Re-presses until focus lands, checking the destination first so a registered press is never doubled.
5. **Stop a slow search response aborting `waitForSearchResult`'s poll** — each iteration registered a `waitForResponse` with its own hard 15s timeout inside an `expect.poll` with a 45s budget. `expect.poll` does not swallow exceptions, so one slow search aborted the whole poll.
6. **Reload when a port row is missing** — the ports endpoint sources rows and total from separate queries and drops unresolvable records without adjusting the total, so it can answer with no rows beside a non-zero total; the list only fetches on mount, so waiting can never recover it but a reload can.
7. **Poll the glossary row count instead of sampling once** — `waitForResponse` proves the bytes arrived, not that the rows rendered; a single `.count()` has no retry budget and read the stale search results.
8. **Wait for virtualised filter options** — the entity filter is a virtualised list holding ~11 rows, and the helper probed with non-retrying `isVisible()` then **threw inside an `expect.poll`**, aborting a 90s budget after 224ms. Every flaking entity sits at index ≥ 11; none inside the initial window ever flaked.
9. **Scope the moved-document assertion to its folder** — the reload refetches page 1 of an `updatedAt DESC` list; parallel uploads evict the moved row, and the test never scrolls.
10. **Assert the team-persona invariant, not global default state** — the backend resolves an unset user default to the *system* default persona, which a sibling describe sets and clears concurrently.
11. **Settle the alerts table before deciding to paginate** — `isHidden()` does not wait while `isEnabled()` does, so the helper sampled an unpainted table, walked past the alert, and (being forward-only) could never return.

---

## Notes for review

- **No application code.** Every change is under `playwright/`. A component-level fix was drafted for the ports list and deliberately dropped in favour of the test-side reload in commit 6.
- **Shared helpers touched:** `utils/domain.ts`, `utils/entity.ts`, `utils/common.ts`, `utils/entityPanel.ts`, `utils/alert.ts`. Blast radius is wide by design — several of these fix a whole class rather than one test.
- **Verification limits stated honestly per commit.** Most of these tests pass locally *before* the fix, because the trigger only appears under CI load; local green therefore shows no regression rather than proving the flake is gone. Only nightlies confirm that.
- Commit 10 could not be verified locally at all — the spec fails earlier on a missing local search index, upstream of the change.

## Related work not included

Three product bugs surfaced during this investigation and are worth separate issues: `getPaginatedPorts` reporting a count its payload cannot support; `SIZE.X_LARGE` — a pixel-dimension enum — used as an Elasticsearch bucket size, silently truncating the browse tree past 166 services; and `CsvJobsTray` auto-opening for jobs the user never started.


ShaileshParmar11:
Holding this PR until the same fix is validated on `2.0` (#31062).

The fix is opened here so the change isn't lost, but it should **not** be merged until the next few 2.0 nightlies confirm that `Toggle fullscreen mode` and `Remove last port shows empty state` come back clean. Both tests currently pass locally even *without* the fix, so CI is the only signal that matters.

Marked as draft until then.

siddhant1 (approved):

Rohit0301 (approved):

PR #31072, opened 2026-08-06
Fixes collate#3919: sort the agent list by the name the Name column renders

siddhant1:
Fixes open-metadata/openmetadata-collate#3919

## Describe your changes

The Name column renders `displayName ?? name`, but `GET /v1/services/ingestionPipelines` has **no sort parameter** and cursor-pages on `ORDER BY name, id`. So the column could only be sorted client-side, over the loaded page. Agents created from the UI get a machine-generated `name` — Automations use `OpenMetadata_application_<random8>` — so ordering by it is arbitrary on screen.

The reason it reads as *"clicking does nothing"* rather than *"sorted wrong"*: the server **already** ordered by `name`, so the first (ascending) click reproduced the order that was already displayed.

### Backend

- New generated column `displayNameSort` = `COALESCE(NULLIF(displayName,''), name)` — the SQL equivalent of `getEntityName`'s `displayName || name`, empty string included — plus a `(deleted, displayNameSort, id)` index. MySQL `VIRTUAL` (indexable, no table rewrite), Postgres `STORED`; follows the `impersonatedBy` (1.11.0) and `glossary_term_entity.displayName` (1.9.6) precedents.
- `sortField` (allowlisted to `displayName`) + `sortOrder` on the list endpoint. Without `sortField` the endpoint is byte-identical to today.
- **Keyset pagination is preserved** — the cursor becomes the `(displayNameSort, id)` tuple, modelled on `ContextFileRepository.listByUpdatedAt`. No offset, so deep paging stays an index range scan and clients keep using `before`/`after` unchanged.
- The sort column is deliberately **not** case-folded: `ORDER BY` and the keyset comparison then share the column's own collation, so the cursor value is carried to Java verbatim. Lowercasing in SQL would force Java to reproduce SQL's case-folding, and any disagreement silently skips or duplicates a row at a page boundary.
- The sorted path re-runs cursor validation, authorization and the domain filter, so it cannot bypass what `listInternal` enforces.

### Frontend

- `IngestionListTable`'s Name column opts into server ordering via `onSortChange`; **consumers that don't opt in keep the client-side comparator**, so `TestSuitePipelineTab` and `MetadataAgentsWidget` are unchanged.
- Drive-by fix: sorting silently cleared the Type filter in the pipelines list. AntD reports sort, filter and pagination through one `onChange`, and the handler read `filters.pipelineType` on *every* action — `undefined` on a sort.

## Type of change

- [x] Bug fix

## Screenshots

Three agents whose `name` order (`agent-1/2/3`) is unrelated to their display names.

**Before** — clicking Name orders by the hidden `name`, so the visible order is arbitrary:

before

**After** — ordered by the name actually shown:

after

## Tests

Written test-first — each failed before the change, for the right reason.

**Backend — `IngestionPipelineSortIT` (new, 6 tests).** RED was `expected: [aaa-first…, zzz-last…] but was: [zzz-last…, aaa-first…]`; `sortField` was silently ignored.

| Test | Guards |
|---|---|
| `_ascending` / `_descending` | orders by display name, not `name` |
| `_fallsBackToNameWhenDisplayNameAbsent` | sort key matches `getEntityName` semantics |
| `_pagesWithoutGapsOrDuplicates` | keyset cursor forward **and** backward across a page boundary |
| `_listWithoutSortField_isUnchangedAndOrdersByName` | the default path is untouched |
| `_listWithUnsupportedSortField_isRejected` | allowlist — no column injection |

**Regression:** existing `IngestionPipelineResourceIT` — 230 tests, 0 failures, 0 errors (20 skipped are K8s-gated).

**Cross-dialect:** the IT harness runs Postgres only, so the MySQL DDL was verified separately on a scratch `CREATE TABLE … LIKE` copy — `displayName` set → used; `displayName=''` → falls back to `name`; absent → falls back to `name`; ordering case-insensitive under `utf8mb4_0900_ai_ci`, matching `localeCompare`.

**Frontend — `IngestionListNameSorting.spec.ts` (new).** RED: the header issued no request carrying `sortField` (`waitForResponse` timed out), re-confirmed afterwards by stashing the diff. The spec asserts the two halves of the UI's contract: the click issues `sortField=displayName&sortOrder=asc|desc`, and the rendered rows mirror that response's order. Ordering *correctness* is the endpoint's contract, covered by the IT above — splitting it this way is what makes the UI assertion deterministic, since the pipelines tab lists every database agent and other rows can occupy page 1.

`mvn spotless:check` clean; `tsc --noEmit` clean for all touched files.

## Manual testing

Ran a local stack from this branch (Postgres + Elasticsearch), seeded three agents whose `name` order differs from their display names, then clicked the Name header on `Settings → Services → Databases → Pipelines` — the screenshots above are that run. Also verified directly against the endpoint: `sortField=displayName` → 200, `sortField=bogus` → `400 Invalid sortField 'bogus'. Supported values: displayName`.

## Checklist

- [x] I have read the CONTRIBUTING document
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective (they fail without it)
- [x] Backend integration tests added
- [x] Playwright test added
- [ ] Documentation — N/A, no user-facing docs describe this ordering

## Follow-up (not in this PR)

- The Collate `AutomatorListPage` change is ready but lands in `openmetadata-collate`; it needs this OM change first (submodule bump). It also sorts the search path via `displayName.keyword`.
- `AutomatorListPage` requests only `fields=owners`, never `pipelineStatuses`, so the Automations **Count column is always 0/0/0** — visible in the original issue screenshot. Separate bug.
- `testTableSorting` (`playwright/utils/common.ts`) only asserts the first cell *changed* between clicks, never that the order is right. That false-positive-prone helper currently "covers" a pipelines table and is why this shipped.

The PR adds server-side display-name sorting for ingestion pipelines while preserving cursor pagination and wires the ingestion table’s Name column to that endpoint.
- Adds display-name-aware keyset queries and optional sorting parameters to the ingestion-pipeline list API.
- Preserves sort order in the UI URL so cursored pages remain reloadable.
- Keeps pipeline-type filters during sorting and adds backend, component, and Playwright regression coverage.
</details>

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java | Adds shared service-type filtering and dialect-specific display-name keyset queries; no eligible follow-up defect was established. |
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java | Implements ascending and descending forward/backward pagination using effective display-name cursors. |
| openmetadata-service/src/main/java/org/openmetadata/service/resources/services/ingestionpipelines/IngestionPipelineResource.java | Adds optional allowlisted display-name sort parameters while retaining the common authenticated listing path. |
| openmetadata-ui/src/main/resources/ui/src/components/Settings/Services/Ingestion/IngestionPipelineList/IngestionPipelineList.component.tsx | Persists sorting in URL state, resets stale cursors, and consolidates fetching around current paging, filtering, and sorting inputs. |
| openmetadata-ui/src/main/resources/ui/src/components/Settings/Services/Ingestion/IngestionListTable/IngestionListTable.tsx | Delegates Name sorting to the server for opted-in consumers while preserving local sorting elsewhere. |
| openmetadata-ui/src/main/resources/ui/src/rest/ingestionPipelineAPI.ts | Forwards the optional sort field and direction through the existing shared API client. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/IngestionPipelineSortIT.java | Covers ordering, filtering, cursor compatibility, keyset boundaries, and fallback behavior. |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/IngestionListNameSorting.spec.ts | Verifies browser-side delegation, response ordering, pagination, and reload persistence. |

</details>

```mermaid
sequenceDiagram
  participant User
  participant Table as Ingestion List
  participant API as Ingestion Pipeline API
  participant Repo as Repository
  participant DB
  User->>Table: Sort Name ascending/descending
  Table->>API: "GET list(sortField=displayName, sortOrder, cursor)"
  API->>Repo: listAfter/listBefore with ListFilter
  Repo->>DB: Keyset query by effective display name and id
  DB-->>Repo: Ordered page plus lookahead row
  Repo-->>API: Data and before/after cursors
  API-->>Table: Ordered ingestion pipelines
  Table-->>User: Render displayName ?? name
```
</details>

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into frontend-only-s..."](https://github.com/open-metadata/openmetadata/commit/6fb638eee82f8bfe5c50e7f9e4742659bd75ecb8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50695917)</sub>

**Context used:**


siddhant1:
@copilot resolve the merge conflicts in this pull request

yan-3005:
@gitar-bot good catch on the cursor truncation — confirmed a real bug.

In the earlier revision the sort key was a generated column defined as `LEFT(COALESCE(NULLIF(displayName,''), name), 256)`, so both the SQL sort key and the Java cursor were capped at 256 and stayed consistent. This branch replaced the generated column with an **inline, un-capped** expression `COALESCE(NULLIF(displayName,''), name)`, but the Java cursor was still truncated to 256 — so for a displayName longer than 256 chars the cursor disagreed with the SQL sort key and the keyset comparison could skip/duplicate a row at the page boundary.

Fixed in `d695267590`: removed `truncateToSortWidth` / `DISPLAY_NAME_SORT_MAX_CHARS` so the cursor carries the full value, matching the un-truncated expression. Added a regression test (`IngestionPipelineSortIT#test_listSortedByDisplayName_pagesAcrossAnUntruncatedLongDisplayName`) that pages across a 300-char displayName — it fails with the old truncation and passes now. Verified locally: 12/12 IT + 16/16 unit.

openmetadata-service/src/main/java/org/openmetadata/service/resources/services/ingestionpipelines/IngestionPipelineResource.java
```diff
@@ -333,9 +389,17 @@ public ResultList<IngestionPipeline> list(
...
             .addQueryParam("provider", provider == null ? null : provider.value());
-    ResultList<IngestionPipeline> ingestionPipelines =
-        super.listInternal(
-            uriInfo, securityContext, fieldsParam, filter, limitParam, before, after);
+    ResultList<IngestionPipeline> ingestionPipelines;
+    if (isDisplayNameSort(sortField)) {
+      KeysetPageParams page =
+          new KeysetPageParams(limitParam, before, after, isAscending(sortOrder));
+      ingestionPipelines =
+          listSortedByDisplayName(uriInfo, securityContext, fieldsParam, filter, page);
```
sonika-shah: @yan-3005 these 2 code paths are almost identical. If someone later changes listInternal (adds a filter, tweaks auth, etc.), the normal list gets it but the display-name sort silently misses out.The two will drift apart.
 
Suggestion — keep one path instead of two:
    -  we can pass new param sortField / sortOrder along on the ListFilter (same way service, pipelineType are already passed).
    - Override listAfter / listBefore inside IngestionPipelineRepository so it picks the display-name query when asked, and falls back to the normal one otherwise.
    - The resource then just keeps its single super.listInternal(...) call — no fork, nothing to keep in sync.

yan-3005: Thanks @sonika-shah — the drift concern is fair, but I dug into the three ways to collapse it and each costs more than the ~5 duplicated lines it removes, so I've left the fork in place. Reasoning:

**1. Pass `sortField`/`sortOrder` on `ListFilter` + override `listAfter`/`listBefore` (your suggestion).** The blocker is the cursor. The generic `EntityRepository.listAfter/listBefore` parse the keyset cursor as `{name, id}` (`cursorMap.get("name")`), while the display-name sort mints a `{displayNameSort, id}` cursor. Since the sort param is lenient (an unrecognised/absent `sortField` falls through to the default listing), a client holding a display-name cursor that drops `sortField` would land in `super.listAfter`, read `name` → `null`, and silently page from the wrong place — no 400. There's also a smaller issue: `ListCountCache.hashFilter` keys on the `queryParams` map, so putting `sortField`/`sortOrder` on the filter splits the count cache into separate sorted/unsorted entries for the same data. Making this safe needs a cursor discriminator in the generic keyset path (cursor tags which column it was minted for; the default path rejects a foreign one) — a framework change across all entities, bigger than the drift it prevents.

**2. Extract the shared preamble (`getFields → validateCursors → authorize → addDomainQueryParam`) so both paths call it.** That preamble lives in `EntityResource.listInternal`, i.e. the base class every entity resource extends — so extracting it touches the listing path of ~60 entities for one list's sort. Blast radius isn't worth it.

**3. Keep the fork.** ~5 duplicated lines in one resource, zero blast radius, and the sorted path already reuses cursor validation via its own `parseDisplayNameCursor` (which throws a clean 400 instead of silently mispaging). The preamble it duplicates is stable core plumbing; if it ever changes, it's one obvious spot to update.

Given OM does arbitrary-field sort through search (`SearchSortFilter`) rather than the DB list anywhere else, a true generic DB-list sort would be its own framework effort. For this PR I think the fork is the least-bad option — happy to revisit if you feel strongly.

sonika-shah: hi @yan-3005 

 There's a cleaner way that avoids both problems you raised:
  - Resource keeps one super.listInternal(...) call.
  - Put sortField/sortOrder as fields on ListFilter (not in the hashed params) → no count-cache split.
  - Override listAfter/listBefore in IngestionPipelineRepository to use the display-name query when asked → no base-class change.


If we still prefer the current changes that's fine but please add a test that checks both paths (with and without sortField) enforce the same auth, domain filter, and cursor validation. That way they can't drift silently later.

yan-3005: Implemented your refined approach in `434723205c`:

- `sortField`/`sortOrder` now live as **fields on `ListFilter`** (not `addQueryParam`), so `ListCountCache.hashFilter` is unchanged — sorted and unsorted share one count-cache entry, no split.
- `IngestionPipelineRepository` **overrides `listAfter`/`listBefore`**: when `filter.getSortField()` is set it runs the display-name keyset query, otherwise it falls straight through to `super`. No base-class change.
- The resource dropped the fork — it now just does `filter.withSort(...)` then a single `super.listInternal(...)`, so authorization, the domain filter and cursor validation are shared with the default listing and the two orderings can't drift.

Verified: `mvn` build clean, `IngestionPipelineSortCursorTest`/`IngestionPipelineSortConditionTest` (16) + `IngestionPipelineSortIT` (12) green, and I ran it against a live server — default (no `sortField`) name ordering + paging, `sortField=displayName` asc/desc with cross-page keyset, and an unsupported `sortField` falling through to the default listing all behave correctly. Thanks for pushing on this.

sonika-shah (dismissed):

karanh37 (approved):

yan-3005:
@gitarbot the `recordQueuedPipelineStatus` finding isn't from this PR — it comes from main (introduced in #31606 "add agentType filter and stop polling orchestrators for queued status", commit 6819167bf7). It only shows in the diff because main was merged into this branch. Out of scope here.

sonika-shah (approved):

yan-3005:
maven collate passed: https://github.com/open-metadata/openmetadata-collate/actions/runs/32119643900/job/95656929817

PR #31073, opened 2026-08-06
fix(ui): derive teams page default tab from the rendered tab list

harsh-vador:
### Describe your changes:

The teams page picked its default tab from `teamType` alone (`Group` → Users, otherwise Teams), while the tab *list* keys on the team **name** for the root Organization team. A team with inconsistent name/teamType data (e.g. a root team mistyped as `Group` — observed on a live tenant) therefore defaulted to the Users tab, which is not rendered for Organization, leaving the page blank until a tab was clicked manually.

This PR derives the default tab from the first entry of the actual tab list (`getTabs(...)`), so the default can never point at a tab that does not exist and bad data degrades gracefully.

Guardrails (per review discussion):
- An explicit `?activeTab=` in the URL is always honored as-is — asynchronously registered plugin tabs are never redirected away from while tabs populate.
- The fallback only reads `currentTeam.name`/`teamType`, which are loaded before `TeamDetailsV1` mounts (page-level loader gates rendering), so the tab list cannot flip mid-load.

https://github.com/user-attachments/assets/e2c3f8d2-126a-45cf-804c-543f7a10b225

#
### Type of change:
- [x] Bug fix

#
### Checklist:
- [x] I have read the [CONTRIBUTING](https://docs.open-metadata.org/developers/contribute) document.
- [x] My change requires a change to the documentation. → No
- [x] I have added unit tests covering the change:
  - blank-page regression (name=Organization + teamType=Group defaults to first existing tab)
  - Group team still defaults to Users tab
  - explicit `?activeTab=` never overridden even when absent from the tab list
- [x] E2E considered and intentionally skipped: the corrupted-data state cannot be created safely via API (Group type change is irreversible server-side and would permanently mutate the shared Organization team in the e2e environment); the healthy default-tab path is already covered by `Teams.spec.ts`.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/Settings/Team/TeamDetails/TeamDetailsV1.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/Settings/Team/TeamDetails/TeamDetailsV1.tsx
@@ -168,8 +168,20 @@ const TeamDetailsV1 = ({
       return activeTab;
     }
 
-    return isGroupType ? TeamsPageTab.USERS : TeamsPageTab.TEAMS;
-  }, [activeTab, isGroupType]);
+    // Derive the default from the actual tab list so a team with inconsistent
+    // name/teamType data (e.g. root team mistyped as Group) can't default to a
+    // tab that isn't rendered, which would leave the page blank.
+    const [defaultTab] = getTabs(
+      currentTeam,
+      isGroupType,
+      isOrganization,
+      0,
+      0,
+      false
+    );
+
+    return defaultTab?.key ?? TeamsPageTab.TEAMS;
+  }, [activeTab, currentTeam, isGroupType, isOrganization]);
   const [deletingUser, setDeletingUser] = useState<{
     user: UserTeams | undefined;
     state: boolean;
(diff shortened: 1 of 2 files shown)


openmetadata-ui/src/main/resources/ui/src/components/Settings/Team/TeamDetails/TeamDetailsV1.tsx
```diff
@@ -168,8 +168,20 @@ const TeamDetailsV1 = ({
...
+    const [defaultTab] = getTabs(
+      currentTeam,
+      isGroupType,
+      isOrganization,
+      0,
+      0,
+      false
+    );
+
+    return defaultTab?.key ?? TeamsPageTab.TEAMS;
```
chirag-madlani, line 183: If we do not pass anything and pass prop like defaultFirstActive that will work too no?

harsh-vador: No — we control the active tab from the URL, and other parts of the page need to know which tab is active (highlight, content). If the tabs component picked its own default internally, the rest of the page wouldn't know and would go out of sync. Picking the first tab ourselves does the same thing without that problem.

chirag-madlani (approved):

PR #30911, opened 2026-08-04
fix(security): move ingestion images to Python 3.12, drop build tooling, upgrade datamodel-code-gen.

harshsoni2024:
### Describe your changes:

Follow-up to #30751. That PR patched PySpark's vulnerable jars in place; this one
addresses the frozen base layer underneath them and the build toolchain.

`python:3.10-slim-bookworm` is EOL and no longer rebuilt, so its Debian layer and bundled
Python tooling are frozen. All four ingestion Dockerfiles move to
`python:3.12-slim-bookworm` / `apache/airflow:3.2.2-python3.12`.

**Key changes**

- **Base bump to Python 3.12** across `ingestion/Dockerfile{,.ci}` and
  `ingestion/operators/docker/Dockerfile{,.ci}`. 3.12's `ensurepip` ships pip only — no
  setuptools, no wheel — which drops the vendored `wheel-0.45.1` behind CVE-2026-24049.

- **`datamodel-code-generator` 0.25.6 → 0.64.0** — clears 8 HIGH (CVE-2026-54621, -54653,
  -54654, -54690, -54691, -55389, -55391, -55415). This was a real API migration, not a
  version bump; see below.

- **Bug fix:** `ingestion/Dockerfile` pinned the Airflow **Python 3.10** constraints file
  while building on a python3.12 base → now `constraints-3.12.txt`.

- **Regenerated `airflow-constraints-3.2.2.txt`** from upstream's 3.12 variant. Both files
  come from the same 2026-05-26 generation run, so this is the matching sibling: it drops
  the `py<3.11` backports (`backports.strenum`, `exceptiongroup`, `tomli`) and refreshes
  numpy/scipy/scikit-learn. The vendored file is byte-identical to upstream — verifiable
  with a single `diff`.

- **pip pinned to `>=26.2,<27` on both sides of the `USER` switch.** The pre-existing
  `pip install --upgrade pip` ran *after* that switch, so it only reached `~/.local` and
  left the base image's copy on disk for scanners (CVE-2025-8869, CVE-2026-3219,
  CVE-2026-6357, CVE-2026-8643 — all fixed by 26.1.2). Python 3.12 alone is not enough:
  its bundled pip is 25.0.1. The `<27` ceiling keeps a major bump out while letting 26.x
  patch releases in, since every pip CVE so far has been fixed within a minor line.

#
### Type of change:
- [x] Bug fix
- [x] Improvement

#
### High-level design:

[...]

PR #31090, opened 2026-08-06
fix(ingestion): bump datamodel-code-generator to 0.72.1 for CVE fixes

Khairajani:
## What

Bumps the pinned `datamodel-code-generator` from `0.25.6` to `0.72.1` and adapts `scripts/datamodel_generation.py` to the API move that came with it.

## Why

`0.25.6` is affected by eight advisories surfaced by an AWS Inspector scan of the ingestion image:

| CVE | Fixed in |
|---|---|
| CVE-2026-54621 | 0.60.1 |
| CVE-2026-54653 | 0.60.2 |
| CVE-2026-54654 | 0.60.2 |
| CVE-2026-54690 | 0.61.0 |
| CVE-2026-54691 | 0.61.0 |
| CVE-2026-55389 | 0.62.0 |
| CVE-2026-55391 | 0.63.0 |
| CVE-2026-55415 | 0.64.0 |

`0.64.0` is the earliest release that clears all eight; this pins the current `0.72.1`.

## The script change

`0.30` removed the pydantic v1 model package entirely, so this no longer resolves:

```python
from datamodel_code_generator.model import pydantic as pydantic_model
pydantic_model.types.IMPORT_SECRET_STR = ...
```

`IMPORT_SECRET_STR` now lives in `datamodel_code_generator.model.pydantic_v2.imports` and is re-exported into `pydantic_v2.types`. The type map reads it as a module global when it is built, not at import time, so rebinding it on `pydantic_v2.types` keeps the redirect working. Without this the script fails outright with `ImportError`.

## Verification

Generated the full model tree under both `0.25.6` and `0.72.1` from the same schemas and diffed:

- **`CustomSecretStr` still applied to the same 125 files** — the secrets-manager redirect is intact.
- **Zero occurrences of bare `pydantic.SecretStr`** in the output (a leak here would silently break secret resolution).
- The whole tree byte-compiles (`compileall`, exit 0).
- **885 of 1048 files change beyond the timestamp header.** All of it is `0.72`'s typing modernisation, three mechanical shapes:
  - `Optional[X]` → `X | None`
  - defaults moved out of `Field(default, ...)` into assignment (`= 3`)
  - `typing_extensions.Annotated` → `typing.Annotated`
- One file added: `schema/__init__.py`.

Sample:

```diff
-from typing import Optional
-from typing_extensions import Annotated
+from typing import Annotated
     jwtTokenExpiryTime: Annotated[
-        Optional[int],
-        Field(3600, description='Jwt Token Expiry time for login in seconds'),
-    ]
+        int | None, Field(description='Jwt Token Expiry time for login in seconds')
+    ] = 3600
```

`generated/` is gitignored, so none of that churn lands in the diff — but it is real, and reviewers should expect the ingestion test suite to exercise it. CI running the full suite against the regenerated models is the gate here.

## Note

`0.72.1` emits a `FutureWarning` that the default `black`/`isort` formatters will become opt-in. Not addressed here; worth a follow-up to pin `formatters=` explicitly before that flips.

### DIFF
--- a/ingestion/Dockerfile.ci
+++ b/ingestion/Dockerfile.ci
@@ -123,7 +123,7 @@ WORKDIR /home/airflow/ingestion
 # Pre-install dialect packages that declare SQLAlchemy<2 in their metadata
 # but work fine at runtime with SQLAlchemy 2.0 (unmaintained packages).
 RUN pip install --no-deps "sqlalchemy-redshift==0.8.14" "sqlalchemy-ibmi==0.9.3" "pydoris-custom==1.1.0"
-RUN pip install "datamodel-code-generator==0.25.6"
+RUN pip install "datamodel-code-generator==0.72.1"
 RUN mkdir -p /home/airflow/ingestion/src/metadata/generated
 RUN python /home/airflow/scripts/datamodel_generation.py
 RUN wget -q https://repo1.maven.org/maven2/org/antlr/antlr4/4.9.2/antlr4-4.9.2-complete.jar -O /tmp/antlr.jar \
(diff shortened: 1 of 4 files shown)


Khairajani:
Already taken care in here: https://github.com/open-metadata/OpenMetadata/pull/30911

PR #31093, opened 2026-08-06
fix(ui): show active Explore tab count from its own index, not the aggregate

mohityadav766:
### Describe your changes:

Fixes open-metadata/openmetadata-collate#3851

The Explore tab badges are populated from a single `index=dataAsset` aggregation (its `entityType` buckets), while each tab's result list is fetched from that tab's **own** index (e.g. `tableColumn`, `databaseSchema`). The two paths run through different query builders, so the number on the tab you're viewing can diverge from the list it shows:

- a column search where the `dataAsset` aggregate counted **1** but the `tableColumn` index returned **7,381**
- a schema search where the aggregate counted **87** but only **~15** were actually listed

**Fix:** for the **active** tab, use its own results total (`searchResults.hits.total.value`) as the badge count, so the number always matches the list in view. Inactive tabs keep the aggregate estimate (their lists aren't visible to verify anyway), and the active tab stays visible even when its own total is 0.

Supersedes #31019 — tightening the column search builder there regressed exact-FQN column lookups to 0 hits (Playwright `SearchRBAC › Table Column`); the divergence is a UI counting concern, not a matching one.

#
### Type of change:

- [x] Bug fix

#
#### Unit tests
- `generateTabItems`: the active tab reports its own results total (not the aggregate), and falls back to the aggregate when no active total is available.

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/pages/ExplorePage/ExplorePageV1.component.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/pages/ExplorePage/ExplorePageV1.component.tsx
@@ -299,11 +299,22 @@ const ExplorePageV1: FC<unknown> = () => {
 
   // Use the utility function to generate tab items
   const tabItems = useMemo(() => {
-    const items = generateTabItems(tabsInfo, searchHitCounts, searchIndex);
+    const items = generateTabItems(
+      tabsInfo,
+      searchHitCounts,
+      searchIndex,
+      searchResults?.hits?.total?.value
+    );
 
     return searchQueryParam
       ? items.filter((tabItem) => {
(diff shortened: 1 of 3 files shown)


mohityadav766:
Superseded by the backend fix in #REPLACE.

Discussed with the team: the aggregate's count of **1** is correct — searching a column's FQN should return that one column. The 7,381 is the bug (the permissive `tableColumn` builder OR-matched every column sharing a parent-name token). Making the badge show the inflated 7,381 (this PR) fixes the *symptom* in the wrong direction. The real fix routes column search through the structured builder so `tableColumn` returns 1, and the existing `dataAsset` aggregate already matches. Closing in favor of that.

mohityadav766:
(correction: superseded by #31106)

PR #31099, opened 2026-08-06
feat(appConfig): add tenant-level defaultAppMode override via yaml

chirag-madlani:
## Summary

Add a **yaml-configured tenant-level force** for the app mode. Operators can pin every user in a deployment to a specific mode (`ai` or `classic`) via `openmetadata.yaml` — overriding the per-user preference. When unset, the current per-user behavior wins.

Complements #31030 (per-user preference persistence). Independent of it — this PR branches off `main` and can land in either order.

## Fallback chain at boot

1. `appConfiguration.defaultAppMode` (tenant force from yaml)
2. `user.preferences.appMode` (per-user preference, from #31030)
3. `DEFAULT_APP_MODE` constant

## What changed

**Backend (yaml → Java → REST):**
- New JSON schema `openmetadata-spec/.../api/configuration/appConfiguration.json` with a nullable `defaultAppMode` field (enum: `"ai" | "classic" | null`).
- `OpenMetadataApplicationConfig.appConfiguration` field, defaulted to `new AppConfiguration()` so `getDefaultAppMode()` is null-safe when yaml is absent.
- New endpoint `GET /v1/system/config/appConfig` in `ConfigResource` — mirrors the `/authorizer` pattern (reads from `openMetadataApplicationConfig`, not `SettingsCache`). Standard `JwtFilter` auth.
- `conf/openmetadata.yaml` gains:
  ```yaml
  appConfiguration:
    defaultAppMode: ${APP_DEFAULT_MODE:-null}
  ```

**UI:**
- `getAppConfig()` in `rest/miscAPI.ts`.
- New `useAppModeConfig` Zustand store with `{ isForced, forcedMode, setForced }`.
- `hydrateAppModeConfig(config)` — called from both bootstrap paths in `AuthProvider.tsx` (returning-session AND fresh-login), via `Promise.all([getLoggedInUser(), getAppConfig()])` to avoid a waterfall. When `defaultAppMode` is non-null, pins runtime via `writeAppMode` and sets `isForced=true`.
- **Hard force**: both `writeAppMode(...)` (in `useAppMode.ts`) and `setPreference({ appMode })` (in `useCurrentUserStore.ts`) short-circuit when `isForced` is true. The `setPreference` guard is per-key — other preference keys still write through normally.
- **Correctness fix discovered during implementation**: added a `CONFIG_MODE_TO_RUNTIME` translation (`'classic'` → `DEFAULT_APP_MODE` which is `'default'`). Without this, forcing "classic" would have set an unrecognized runtime string, and `useIsAiMode()` would incorrectly return `true` for a classic-forced tenant. Mirrors the existing `APP_MODE_ENUM_TO_RUNTIME` pattern in `useResolvedAppMode.ts`.

## Testing

- **Java**: `ConfigResourceTest` (new, 3/3 pass) — Mockito unit test in the same style as `IndexResourceTest.java`. No infra needed; runs via plain `mvn test`. Covers: default returns null; configured value is echoed; null-safe when `getAppConfiguration()` returns `null`.
- **Jest**: 54/54 pass across `useAppModeConfig` (new suite), `useCurrentUserStore` (+2 guard tests), `useAppMode` (+2 guard tests). Pre-existing `AuthProvider.test.tsx` still 12/12 (no regression).
- **Lint + tsc**: clean on touched files.
- `mvn spotless:apply` clean on `openmetadata-service` / `openmetadata-spec`.

## Design decisions locked

- **Yaml-only** — no DB storage, no `SettingsCache`, no admin UI page, no runtime mutation. Ops change the value by editing yaml + restart, same as `authorizerConfiguration`.
- **Boot-only enforcement** — no mid-session flip when yaml changes.
- **Hard force** — user can't fight the admin's choice. `AppModeSwitcher` (Collate-side follow-up) will hide when forced.
- **Per-key guard** — only the `appMode` key is guarded on `setPreference`; other user prefs pass through.

## Not in scope

- Admin UI settings page — future PR if ever needed. For now ops set via env var / yaml.
- Live enforcement / SSE push of yaml changes — would need a new subsystem.
- `AppModeSwitcher` visibility guard — lives in Collate, tracked separately.

## Test plan for reviewers

- [ ] CI passes (Java unit tests + Jest suite).
- [ ] Deploy with `APP_DEFAULT_MODE=ai` → every user lands in AI mode at login; user toggle attempts are no-ops.
- [ ] Deploy with `APP_DEFAULT_MODE=classic` → same in reverse (verifies the runtime translation).
- [ ] Deploy without setting the env → behavior is identical to today (user preference wins, or `DEFAULT_APP_MODE` if no preference).
- [ ] `GET /api/v1/system/config/appConfig` returns `{ defaultAppMode: null }` on a default deployment; returns the configured value when set.

### DIFF
--- a/conf/openmetadata.yaml
+++ b/conf/openmetadata.yaml
@@ -854,6 +854,13 @@ rdf:
   dataset: ${RDF_DATASET:-"openmetadata"}
   inferenceEnabled: ${RDF_INFERENCE_ENABLED:-false}
 
+# App Configuration
+# Tenant-wide, yaml/env-only UI configuration. Not backed by the database.
+appConfiguration:
+  # Non-null pins every user to this mode at boot regardless of their per-user
+  # preference (e.g. "ai" or "classic"); null means the user's own preference wins.
+  defaultAppMode: ${APP_DEFAULT_MODE:-null}
+
 # Cache Configuration
 # Caching layer for entity metadata, relationships, and tag usage to reduce database load
 # Default: none (no external dependency). Set CACHE_PROVIDER=redis to enable the
--- a/openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplicationConfig.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/OpenMetadataApplicationConfig.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import lombok.Getter;
 import lombok.Setter;
 import org.openmetadata.DefaultOperationalConfigProvider;
+import org.openmetadata.schema.api.configuration.AppConfiguration;
 import org.openmetadata.schema.api.configuration.dataQuality.DataQualityConfiguration;
 import org.openmetadata.schema.api.configuration.events.EventHandlerConfiguration;
 import org.openmetadata.schema.api.configuration.pipelineServiceClient.PipelineServiceClientConfiguration;
@@ -191,6 +192,10 @@ public class OpenMetadataApplicationConfig extends Configuration {
 
   @JsonProperty("rdf")
   private RdfConfiguration rdfConfiguration = new RdfConfiguration();
+
+  @JsonProperty("appConfiguration")
+  @Valid
+  private AppConfiguration appConfiguration = new AppConfiguration();
 
   @JsonProperty("cache")
   private org.openmetadata.service.cache.CacheConfig cacheConfig;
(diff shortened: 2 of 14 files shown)


ShaileshParmar11 (approved):

conf/openmetadata.yaml
```diff
@@ -854,6 +854,13 @@ rdf:
   dataset: ${RDF_DATASET:-"openmetadata"}
   inferenceEnabled: ${RDF_INFERENCE_ENABLED:-false}
 
+# App Configuration
```
mohityadav766, line 857: let's remove it from here , we can keep this only in the database

mohityadav766 (commented):
Needs some work. Use SettingsCache to load the setting into the database if not existing, else don't . 
We don't need to store the settings in opemetadata.yaml.

chirag-madlani:
Closing in favor of the v2 architecture per reviewer feedback (Pablo + Mohit). `appConfiguration` moves from yaml-only hard force to a DB-backed `SettingsCache` setting (yaml seeded on first boot only), admin-mutable via API, and becomes a **soft default** — user and admin can always override.

New combined PR forthcoming — will link when opened. Branch `feat/app-mode-tenant-override` stays on the remote for reference.

PR #31103, opened 2026-08-06
fix(auth): recover from expired-token 401s instead of hanging the UI

### LINKED ISSUE #31102: Auth: expired-token 401 hangs the UI on a spinner instead of silently refreshing the session
### Describe the bug

When a user's access token (JWT) expires mid-session, the first API request after expiry returns `401`. The frontend is meant to silently refresh the token and retry the request, but under common conditions the failed request is **parked indefinitely** — the UI shows a spinner and never recovers until an unrelated background `loggedInUser` poll happens to force a `logout → login` redirect (up to ~a minute later). The session itself is still valid on the server; only the client-side 401 recovery fails.

### Root cause

1. **Orphaned request queue (primary).** The Axios 401 response interceptor (`AuthProvider.tsx`) queues the failed request in a module-level `pendingRequests` array, but only the request that *starts* the refresh ever drains it. Whether "a refresh is already in progress" is read from a **`localStorage` flag shared across tabs** (`TokenServiceUtil`). If that flag is set by another tab — or left stale by a prior refresh — every 401 takes the "just queue it" branch and nothing ever drains it. The request promise never settles → infinite spinner, no redirect.
2. **`refreshToken()` returned `undefined` to concurrent callers**, which the interceptor read as "refresh failed" and logged the user out even while a refresh was actually succeeding.
3. **Logout used `GET /api/v1/auth/logout`**, but the endpoint is `POST`-only (`AuthLogoutServlet` returns `405` + `Allow: POST` on GET), so server-side logout never completed.
4. **`renewToken()` had dead `status === 302` handling.** Browsers (XHR) auto-follow 3xx, so a refresh that redirected an expired session to re-login came back as a truthy body with an empty `accessToken` — a false success that retried with an empty token instead of redirecting.
5. **Unknown JWT `kid` returned HTTP 500, not 401.** `MultiUrlJwkProvider.get()` wrapped `SigningKeyNotFoundException` in `UnhandledServerException` (500). The UI only recovers on 401, so IdP key-rotation / stale-token cases surfaced as a hard error the client couldn't classify as an auth failure.

### Expected behavior

A 401 from an expired token should trigger a single token refresh and transparently retry the in-flight request. If the refresh cannot produce a token, the app should redirect to login cleanly — never hang on a spinner.

### Fix (this PR)

- **Interceptor:** a single in-tab refresh *driver* that **always** drains the queue — retry every queued request on a new token, otherwise reject them and log out. Nothing is parked behind a cross-tab flag.
- **`TokenService.refreshToken()`:** coalesce concurrent refreshes into one in-flight promise; wait for a sibling tab's refresh instead of racing it or returning `undefined`.
- **Logout** via `POST`.
- **`renewToken()`:** throw when the refresh response carries no access token; remove the dead 302 branch.
- **Backend:** return `401` (`AuthenticationException`) for an unknown signing-key id; keep `500` only for genuine server errors (e.g. a JWKS fetch failure).

### Affected files

- `openmetadata-ui/.../components/Auth/AuthProviders/AuthProvider.tsx`
- `openmetadata-ui/.../utils/Auth/TokenService/TokenServiceUtil.ts`
- `openmetadata-ui/.../rest/LoginAPI.ts`
- `openmetadata-ui/.../components/Auth/AppAuthenticators/GenericAuthenticator.tsx`
- `openmetadata-service/.../security/MultiUrlJwkProvider.java`

mohityadav766:
## Describe your changes

Fixes #31102.

When an access token (JWT) expires mid-session, the first API request after expiry returns `401`. The UI is supposed to silently refresh the token and retry the request, but the failed request could be **parked indefinitely** — a spinner with no redirect — until an unrelated background `loggedInUser` poll forced a `logout → login` up to a minute later. The server session was still valid the whole time; only the client-side 401 recovery failed.

### Root cause
- **Orphaned request queue (primary):** the Axios 401 interceptor queued the failed request in a module-level `pendingRequests` array, but only the request that *starts* the refresh drains the queue — and the "refresh already in progress?" gate read a **cross-tab `localStorage` flag**. When that flag was set by another tab, or left stale by a prior refresh, every 401 took the "just queue it" branch and nothing ever drained it, so the request promise never settled.
- `refreshToken()` returned `undefined` to concurrent callers, which the interceptor treated as a failure and logged the user out even while a refresh was succeeding.
- Logout used `GET /api/v1/auth/logout`, but the endpoint is `POST`-only (`AuthLogoutServlet` returns `405`), so server-side logout never completed.
- `renewToken()`'s `status === 302` branch was dead code — browsers (XHR) auto-follow 3xx — so a refresh that redirected an expired session came back as a truthy body with an empty `accessToken` (false success → retry loop instead of clean redirect).
- The backend returned **500**, not **401**, for an unknown JWT `kid` (`MultiUrlJwkProvider` wrapped `SigningKeyNotFoundException` in `UnhandledServerException`). The UI only recovers on 401, so IdP key-rotation / stale-token cases surfaced as a hard error.

### Changes
- **`AuthProvider.tsx`** — a single in-tab refresh *driver* that **always** drains `pendingRequests` (retry every queued request on a new token, otherwise reject them and log out). Removes the cross-tab-flag branch that orphaned requests. Hardened the `loggedInUser` body check with optional chaining.
- **`TokenServiceUtil.ts`** — `refreshToken()` coalesces concurrent refreshes into one in-flight promise, and waits (bounded) for a sibling tab's refresh instead of racing it or returning `undefined`.
- **`LoginAPI.ts`** — logout via `POST`; `renewToken()` throws when the refresh response has no `accessToken` (dead 302 branch removed).
- **`GenericAuthenticator.tsx`** — drop the now-unused `navigate` argument.
- **`MultiUrlJwkProvider.java`** — return `401` (`AuthenticationException`) for an unknown signing-key id; keep `500` only for genuine server errors (e.g. a JWKS fetch failure).

## Type of change

## How was this patch tested
- **UI (Jest)** — affected suites green (33 tests):
  - `TokenServiceUtil.test.ts`: concurrent-refresh coalescing + sibling-tab wait (replaces the test that asserted the buggy `undefined` return).
  - `AuthProvider.test.tsx`: a normal 401 now refreshes the token and retries the request instead of hanging.
- **Backend (JUnit)** — `MultiUrlJwkProviderTest` now asserts an unknown `kid` throws `AuthenticationException` (401), not `UnhandledServerException` (500).

## Checklist
- [x] I have added unit tests that fail without this change
- [x] UI lint (ESLint + Prettier + organize-imports) and backend `spotless:apply` run clean


mohityadav766:
Thanks for the reviews — addressed the three code findings in a16c38f417:

- **gitar-bot** (null on a concurrent/locally-valid 401): the 401 interceptor now calls `refreshToken(true)` — a forced refresh skips the local-expiry short-circuit, so a token the server has rejected (clock skew / revocation) is actually refreshed instead of triggering a spurious logout.
- **Copilot** (`TokenServiceUtil.ts`, void-returning silent renew): `performRefresh` now determines success by a **persisted token change**, not the renewer's return value — OIDC `signInSilently` resolves `void` and writes the token via an iframe side effect, which was being misread as a failure.
- **Copilot** (`MultiUrlJwkProvider.java`): the underlying cause is now chained on both the `AuthenticationException` (401) and `UnhandledServerException` (500) paths, so genuine JWKS-fetch failures stay debuggable.

Added unit tests for the void-renew and forced-401 cases (`TokenServiceUtil.test.ts`); `MultiUrlJwkProviderTest` still green (6/6).

mohityadav766:
Addressed a second, deeper review (thanks!) covering every authenticator — fixed in 6b282e0476. Three of these could otherwise still cause a spurious logout on public clients:

- **Dropped the `force` flag + "skip if still valid" short-circuit** (`performRefresh`). Every caller (proactive timer, visibility, SSE stream, 401 interceptor) already refreshes only when needed, so the short-circuit only blocked a genuinely-needed refresh — and let a non-force in-flight refresh drag a coalesced 401 (`useServiceProgressStream` racing an HTTP 401) into a `null` → logout.
- **Success = the renewer resolved without throwing**, not "stored token changed". MSAL/Okta/Auth0 silent renew can re-issue the *same* `id_token` (`forceRefresh` only refreshes the access token) → the change-check was a regression to logout. Void-returning OIDC silent renew still handled via a bounded persistence wait.
- **`waitForTokenPersistence` checks at t=0** before sleeping, and the cross-tab wait **falls through to a local refresh** on timeout instead of logging out.
- **`loggedInUser` 401 recovery** now keys on `REFRESHABLE_AUTH_ERRORS` (`Expired token!` **or** `Token signing key not found`) so the IdP key-rotation case recovers on the polled endpoint.
- **`AuthenticationException.getInvalidTokenException(reason, e)` chains the cause** (`super(msg, cause)`), so `%ex` logs keep the `SigningKeyNotFoundException` stack.
- Cleared `refreshInProgress` when no renewer is configured (flag-leak guard); removed a stray `// ponytail:` comment.

Verification: 65 UI tests across 7 suites (TokenService, AuthProvider, all six authenticators) green — new cases cover same-`id_token` success, void-renew success, refresh-on-locally-valid, and the unknown-`kid` `loggedInUser` refresh. `MultiUrlJwkProviderTest` 6/6 (now also asserts the chained cause).

openmetadata-ui/src/main/resources/ui/src/utils/Auth/TokenService/TokenServiceUtil.ts
```diff
@@ -80,67 +79,63 @@ class TokenService {
...
       }
     }
 
     this.setRefreshInProgress();
     const renewResult = await this.fetchNewToken();
     if (renewResult === null) {
-      // Explicit failure: the renewer threw, or none is configured.
+      // The renewer threw, or none is configured — a genuine refresh failure.
+      // fetchNewToken only clears the flag when a renewer actually ran.
+      this.clearRefreshInProgress();
```
chirag-madlani, line 131: In can of public OIDC, we are getting renewResult as null 
Since we will be getting an updated token in a callback
Refer here: https://github.com/open-metadata/OpenMetadata/blob/5c831f44b6d33e86a6dcc08a70dcf776e591313d/openmetadata-ui/src/main/resources/ui/src/components/Auth/AppAuthenticators/OidcAuthenticator.tsx#L140

mohityadav766: Good catch — fixed in 379cc6dbbf. You're right: `signInSilently` returns null/void and the token arrives via the silent-callback iframe, so judging success from `renewResult` logged public-OIDC users out on a *successful* renew.

`performRefresh` now treats **any falsy** renew result (null or void) as "token comes out of band" and decides success by whether a token actually lands in storage (bounded ~1s wait, checked at t=0):

```ts
const refreshedToken = renewResult || (await this.waitForTokenPersistence(oldToken));
if (!refreshedToken) { this.clearRefreshInProgress(); return null; }
```

Value renewers (MSAL/Okta/Auth0/Basic/Generic) still succeed immediately from their return value (incl. a re-issued identical id_token); only a falsy result with no token landing is treated as a real failure. Added a unit test covering the null-renew + callback-delivered-token path.

openmetadata-ui/src/main/resources/ui/src/constants/Auth.constants.ts
```diff
@@ -41,3 +41,13 @@ export const UN_AUTHORIZED_EXCLUDED_PATHS = [
...
+
+// 401 messages on the polled `/users/loggedInUser` that a silent token refresh
+// can actually fix (so the interceptor should refresh rather than log out).
+// Kept as substrings to match the backend's `Not Authorized! ...` wrapping.
+export const REFRESHABLE_AUTH_ERRORS = [
+  'Expired token!',
+  // Unknown JWT signing-key id (IdP key rotation / stale token) — a refresh
+  // mints a token signed with a currently-published key.
+  'Token signing key not found',
+];
```
siddhant1, line 53: @mohityadav766 these are magic strings, changing anything in the backend will break these, is this the best way to do it?

chirag-madlani (approved):

sonika-shah (approved):

PR #31121, opened 2026-08-06
Fixes #31120: strip NUL byte from pipeline status before Postgres jsonb write

### LINKED ISSUE #31120: Pipeline status ingestion returns 500 on PostgreSQL when a stack trace contains a NUL byte (\u0000)
### Affected module

Backend

### Describe the bug

Publishing an ingestion **pipeline status** fails on **PostgreSQL** when any string in the payload (typically a task `stackTrace`/`error`) contains a raw NUL character (`U+0000`). The `POST /api/v1/services/ingestionPipelines/.../pipelineStatus` request returns **500** and the status is never persisted.

Jackson serializes the NUL character as its JSON escape `\u0000`. PostgreSQL's `jsonb` type **rejects** that escape when converting to text, whereas MySQL accepts it — so this is a **PostgreSQL-only** failure.

Observed exception:

```
org.jdbi.v3.core.statement.UnableToExecuteStatementException:
  org.postgresql.util.PSQLException: ERROR: unsupported Unicode escape sequence
  Detail: \u0000 cannot be converted to text.
  Where: JSON data, line 1: ... Unknown error: Wrapping \u0000...
  [statement:"/* EntityExtensionTimeSeriesDAO.updateExtensionByKeyInternal */
   UPDATE entity_extension_time_series SET json = (:json :: jsonb) WHERE ..."]
    at org.openmetadata.service.jdbi3.EntityTimeSeriesDAO.updateExtensionByKey(EntityTimeSeriesDAO.java:707)
    at org.openmetadata.service.jdbi3.IngestionPipelineRepository.addPipelineStatus(IngestionPipelineRepository.java:604)
    at org.openmetadata.service.resources.services.ingestionpipelines.IngestionPipelineResource.addPipelineStatus(IngestionPipelineResource.java:1207)
```

The NUL byte leaks in from binary sources captured in ingestion diagnostics — in the reported case a Kubernetes "Pod Diagnostics" dump embedded in a `failures[].stackTrace` field.

### To Reproduce

1. Deploy OpenMetadata against a **PostgreSQL** backend.
2. Run (or simulate) an ingestion whose run reports a status where a `stackTrace`/`error` string contains a NUL byte (`U+0000`). This happens in practice when pod/subprocess diagnostics are captured into the error text.
3. The server attempts to store the status into `entity_extension_time_series` via `... SET json = (:json :: jsonb)`.
4. PostgreSQL raises `ERROR: unsupported Unicode escape sequence — \u0000 cannot be converted to text`; the request returns 500 and the status is lost.

Minimal equivalent at the DB layer:

```sql
SELECT '{"stackTrace":"foo\u0000bar"}'::jsonb;  -- ERROR: unsupported Unicode escape sequence
```

### Expected behavior

The pipeline status should be persisted successfully on PostgreSQL. A NUL byte is never meaningful in stored text and should be stripped before the JSON is written to a `jsonb` column, rather than causing a 500.

### OpenMetadata version

2.0 (reproducible on all versions using a PostgreSQL backend; the code path is long-standing)

### Additional context

The same `jsonb` write pattern (serialize a status object carrying a `stackTrace` → store in `entity_extension_time_series`) exists for both the **ingestion pipeline** status (`IngestionPipelineRepository`) and the **data pipeline** entity status (`PipelineRepository`), so both are affected on PostgreSQL.

Infrastructure identifiers (pod name, namespace, node IP) from the original report have been redacted.

mohityadav766:
### Describe your changes:

Fixes #31120

I made ingestion **pipeline status** writes safe for PostgreSQL `jsonb` because a raw NUL character (`U+0000`) in a stack trace was returning **500** and dropping the status.

**Root cause.** Jackson serializes a NUL character as its JSON escape. PostgreSQL's `jsonb` rejects that escape on text conversion (`ERROR: unsupported Unicode escape sequence — ... cannot be converted to text`), while MySQL accepts it — so the failure is PostgreSQL-only. The NUL leaks in from binary sources captured in ingestion diagnostics (the report had a Kubernetes "Pod Diagnostics" dump inside `failures[].stackTrace`).

**Fix.** Added `JsonUtils.pojoToJsonPostgresSafe(o)` (`= pojoToJson(o)` with the NUL escape stripped) and routed every pipeline-status writer through it — the stack-trace-bearing status object is the realistic NUL source:
- `IngestionPipelineRepository` — `addPipelineStatus` (insert + update) and `updatePipelineStatusByRunId`.
- `PipelineRepository` — `addPipelineStatus` (insert + update) and the bulk batch insert (same `entity_extension_time_series` `jsonb` write, so it would 500 identically on PostgreSQL).

Stripping is applied unconditionally (harmless on MySQL — a NUL is never meaningful in stored text), so there is no DB-type branching.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change. The only new surface is one helper in `openmetadata-spec` `JsonUtils`; the repository changes are one-line call-site swaps.

#
#### Use cases covered
- Publishing an ingestion pipeline status whose `stackTrace`/`error` contains a NUL byte persists successfully on PostgreSQL instead of returning 500.
- Normal status payloads (no NUL) are unaffected; surrounding text is preserved.

#### Unit tests
- [x] I added a unit test for the new logic.
- Files added/updated: `openmetadata-service/.../util/JsonUtilsTest.java` — `testPojoToJsonPostgresSafeStripsNulChar` proves a real NUL char serializes to the escape PostgreSQL rejects, that the helper strips it while preserving surrounding text, and null-in → null-out.
- Result: full `JsonUtilsTest` class **17/17 pass**.

- [x] Not applicable — no new or changed API endpoint or contract; this is a serialization/data-sanitization fix on an existing write path, covered by the unit test above.

#### Playwright (UI) tests

#### Manual testing performed
1. `mvn -pl openmetadata-spec install -DskipTests` then `mvn -pl openmetadata-service test -Dtest=JsonUtilsTest` → **17/17 pass**.
2. `mvn spotless:check compile -pl openmetadata-service` → **BUILD SUCCESS**.
3. Confirmed the PostgreSQL behavior the fix targets — a NUL escape inside a JSON string cannot be cast to `jsonb` (`SELECT '{"k":"a<NUL>b"}'::jsonb` raises `unsupported Unicode escape sequence`) — which the unit test reproduces at the serialization layer.

#

Not applicable.

#
### Checklist:
- [x] My PR is linked to a GitHub issue via `Fixes #31120` above.
- [x] For JSON Schema changes: not applicable (no schema change).

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
@@ -606,7 +606,7 @@ public class IngestionPipelineRepository extends EntityRepository<IngestionPipel
               pipelineStatus.getRunId(),
               ingestionPipeline.getFullyQualifiedName(),
               PIPELINE_STATUS_EXTENSION,
-              JsonUtils.pojoToJson(pipelineStatus));
+              JsonUtils.pojoToJsonPostgresSafe(pipelineStatus));
     } else {
       daoCollection
           .entityExtensionTimeSeriesDao()
@@ -614,7 +614,7 @@ public class IngestionPipelineRepository extends EntityRepository<IngestionPipel
               ingestionPipeline.getFullyQualifiedName(),
               PIPELINE_STATUS_EXTENSION,
               PIPELINE_STATUS_JSON_SCHEMA,
-              JsonUtils.pojoToJson(pipelineStatus));
+              JsonUtils.pojoToJsonPostgresSafe(pipelineStatus));
     }
     ChangeDescription change =
         addPipelineStatusChangeDescription(
@@ -772,7 +772,7 @@ public class IngestionPipelineRepository extends EntityRepository<IngestionPipel
   public void updatePipelineStatusByRunId(String fqn, PipelineStatus pipelineStatus) {
     IngestionPipeline ingestionPipeline = findByName(fqn, Include.NON_DELETED);
     String pipelineFqn = ingestionPipeline.getFullyQualifiedName();
-    String json = JsonUtils.pojoToJson(pipelineStatus);
+    String json = JsonUtils.pojoToJsonPostgresSafe(pipelineStatus);
     PipelineStatus storedPipelineStatus =
         JsonUtils.readValue(
             daoCollection
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/PipelineRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/PipelineRepository.java
@@ -342,14 +342,14 @@ public class PipelineRepository extends EntityRepository<Pipeline> {
           .update(
               pipeline.getFullyQualifiedName(),
               PIPELINE_STATUS_EXTENSION,
-              JsonUtils.pojoToJson(pipelineStatus),
+              JsonUtils.pojoToJsonPostgresSafe(pipelineStatus),
               pipelineStatus.getTimestamp());
     } else {
       storeTimeSeries(
           pipeline.getFullyQualifiedName(),
           PIPELINE_STATUS_EXTENSION,
           "pipelineStatus",
-          JsonUtils.pojoToJson(pipelineStatus));
+          JsonUtils.pojoToJsonPostgresSafe(pipelineStatus));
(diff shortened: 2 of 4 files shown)


mohityadav766:
closing in favor of https://github.com/open-metadata/OpenMetadata/pull/30956

PR #16017, opened 2024-04-24
Issue #14812: Add support for empty string as missing count

ayush-shah:
### Describe your changes:

Fixes #14812 
Add support for empty string as missing count

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

Issue #14812, opened 2024-01-22
empty string cant be matched as a null value


Ianzhii:
**Affected module**
Data Quaalty
**Describe the bug**
on the ui, add a column quality test which test type choose 'column values missing count'. empty string such as '', '      ' cant  be match

**To Reproduce**
![image](https://github.com/open-metadata/OpenMetadata/assets/19724748/274feaa6-dac9-44e4-acda-396632845e52)
![image](https://github.com/open-metadata/OpenMetadata/assets/19724748/fc7bf3cd-4c40-47e6-b78d-21c5db57e351)
![image](https://github.com/open-metadata/OpenMetadata/assets/19724748/8f50cb64-d564-4435-80e8-7ac49fdfe075)

**Expected behavior**
as the description shows 'By default match all null and empty values to be missing. ', '' or '    ' should be matched as empty values
**Version:**
 - OS: [e.g. centos7.9]
 - Python version:3.8.5
 - OpenMetadata version: [e.g. 1.2.2]
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

TeddyCr:
This is [the test](https://docs.open-metadata.org/v1.3.x/connectors/ingestion/workflows/data-quality/tests#column-values-missing-count-to-be-equal) to check if setting `""` validate empty string

PR #31130, opened 2026-08-06
Fixes 21482: only compare a column against '' when the type can hold it

### LINKED ISSUE #21482: Some test case errors
Dear all,

I faced some troubles deploying test cases:

1) Column test type "Column values missing count(1,10)":
- `Pipeline: success, test: aborted` for string/character datatype in both clickhouse and postgres
- Clickhouse, datatype int: `Could not run test case Test_int: Orig exception: Code: 32. DB::Exception: Attempt to read after eof: while converting '' to Int32. (ATTEMPT_TO_READ_AFTER_EOF) (version 24.8.14.10459.altinitystable (altinity build))`
- Postgres, datatype integer: `Could not run test case test: (psycopg2.errors.InvalidTextRepresentation) invalid input syntax for type integer: "" LINE 2: ...SUM(CAST(CASE WHEN (id IS NULL) THEN 1 WHEN (id = '') THEN 1... ^ [SQL: /* {"app": "OpenMetadata", "version": "1.6.5.1"} */ SELECT SUM(CAST(CASE WHEN (id IS NULL) THEN %(param_1)s WHEN (id = %(id_1)s) THEN %(param_2)s ELSE %(param_3)s END AS NUMERIC)) AS "nullCount" FROM public.testtable LIMIT %(param_4)s] [parameters: {'param_1': 1, 'id_1': '', 'param_2': 1, 'param_3': 0, 'param_4': 1}] (Background on this error at: https://sqlalche.me/e/14/9h9h)`
- Clickhouse, datatype date: `Could not run test case Test_date: Orig exception: Code: 38. DB::Exception: Cannot parse date: value is too short: while converting '' to Date. (CANNOT_PARSE_DATE) (version 24.8.14.10459.altinitystable (altinity build))`
- Postgres, datatype timestamp: `Could not run test case test: (psycopg2.errors.InvalidDatetimeFormat) invalid input syntax for type timestamp with time zone: "" LINE 2: ...EN (created_at IS NULL) THEN 1 WHEN (created_at = '') THEN 1... ^ [SQL: /* {"app": "OpenMetadata", "version": "1.6.5.1"} */ SELECT SUM(CAST(CASE WHEN (created_at IS NULL) THEN %(param_1)s WHEN (created_at = %(created_at_1)s) THEN %(param_2)s ELSE %(param_3)s END AS NUMERIC)) AS "nullCount" FROM public.testtable LIMIT %(param_4)s] [parameters: {'param_1': 1, 'created_at_1': '', 'param_2': 1, 'param_3': 0, 'param_4': 1}] (Background on this error at: https://sqlalche.me/e/14/9h9h)`

2) Column test type "Column Values to Match Regex Pattern(10)":
- Clickhouse, datatype string: `Could not run test case Test_String: Orig exception: Code: 46. DB::Exception: Function with name 'MATCH' does not exist. In scope SELECT count(id) AS valuesCount, SUM(accurateCastOrNull(multiIf(MATCH(id, '10'), 1, 0), 'BIGINT')) AS regexCount FROM Test_Projects.testtable LIMIT 1. Maybe you meant: ['match','path']. (UNKNOWN_FUNCTION) (version 24.8.14.10459.altinitystable (altinity build))`

3) Column test type "Column Values to be Between(1,10)":
- Postgres, datatype timestamp: `Could not run test case test: can't compare offset-naive and offset-aware datetimes`

Am I doing something wrong or there is some advice how to use these tests?

I am using OpenMetadata 1.6.5 and Ingestion 1.6.5.1

Thank you for any help!

TeddyCr:
### Describe your changes:

Fixes #21482

The `columnValuesMissingCount` ("Column Values Missing Count") test always compared the column against the empty string, whatever its type:

```sql
SUM(CAST(CASE WHEN (id2 IS NULL) THEN 1 WHEN (id2 = '') THEN 1 ELSE 0 END AS NUMERIC)) AS "nullCount"
```

On PostgreSQL an `INTEGER` column makes this fail outright — `psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type integer: ""` — so the test aborts. On MySQL it does not fail: `''` is coerced to `0`, so **every zero row is counted as missing** and the test silently reports a wrong number.

The empty-string branch comes from `NullMissingCount` (added in #16017 for #14812, which only considered string columns). It is now emitted only for types that can actually hold `''`. The predicate fails open: quantifiable, date/time, boolean, UUID, interval, enum and IP types are excluded; anything else — including dialect-specific string types we don't recognize — keeps counting empty strings exactly as before.

`CustomTime` (the MariaDB `TIME` mapping) is listed in the metric's exclusion tuple rather than added to `registry.is_date_time`, because that helper also drives `min`/`max`/`mean`/`stddev` and `columnValuesToBeBetween`; widening it would change those paths too, which is out of scope for this fix.

`missingValueMatch` is unaffected: user-supplied missing values go through the separate `countInSet` metric.

The pandas/DataFrame path is deliberately left alone. A CSV/datalake column typed as `INT` genuinely can carry `""` for a missing value, so guarding there would drop real missing values; the SQL path has no such case because the engine enforces the type.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered
- `Column Values Missing Count` on a PostgreSQL `INTEGER` column runs and reports only the NULLs, instead of aborting with `invalid input syntax for type integer: ""`
- The same test on a MySQL `INT`/`BOOLEAN` column no longer counts `0` / `false` rows as missing
- The same test on a MariaDB `TIME` column no longer counts `'00:00:00'` rows as missing
- The same test on `DATE`, `BOOLEAN`, `UUID`, `INTERVAL`, `ENUM` and `IPV4`/`IPV6` columns no longer errors on PostgreSQL (and on `DATE` for MySQL, which raised `Incorrect DATE value: ''` in strict mode)
- String, text and unrecognized/custom string-like columns keep counting `''` as missing — no behavior change

#### Unit tests
- Files added: `ingestion/tests/unit/observability/profiler/sqlalchemy/test_null_missing_count_metric.py`
- Compiles the real metric expression against both the PostgreSQL and MySQL dialects and asserts on the emitted SQL per column type: no `= ''` for `Integer`, `SmallInteger`, `BIGINT`, `Numeric`, `DECIMAL`, `Float`, `Date`, `DateTime`, `Time`, `CustomTimestamp`, `CustomTime`, `Boolean`, `UUIDString`, `Interval`, `Enum`, `CustomIP`; `= ''` preserved for `String`, `VARCHAR`, `TEXT`, `CHAR`, `NVARCHAR` and the unrecognized-type fail-open case.
- A data-driven case walks `CommonMapTypes._TYPE_MAP` for the 18 OM `DataType`s that cannot hold `''`, so a converter change that reintroduces the bug fails the suite.

- [x] I added ingestion integration tests.
- Files added: `ingestion/tests/integration/profiler/test_null_missing_count_metric.py`
- Runs the metric against real PostgreSQL 15 and MySQL 8.4 testcontainers over a table with 3 NULLs, 2 zeros, 2 false and 2 empty strings, asserting integer/date/boolean columns count 3 and the string column counts 5.

#### Playwright (UI) tests

#### Manual testing performed

Verified the generated SQL before/after, and the engine behavior on throwaway containers with a table holding 3 NULLs and 2 zeros:

| | before | after |
|---|---|---|
| PostgreSQL `integer` | `ERROR: invalid input syntax for type integer: ""` | 3 |
| MySQL `int` | 5 (zeros counted as missing) | 3 |
| MySQL `boolean` | 4 (false rows counted) | 2 |
| MariaDB `time` | 2 (`'00:00:00'` counted) | 1 |
| PostgreSQL `date` / `boolean` / `uuid` / `interval` / `inet` / `enum` | `ERROR: invalid input syntax for type …: ""` | works |
| MySQL `date` | `ERROR 1525: Incorrect DATE value: ''` | works |

Full `ingestion/tests/unit/observability/profiler` + `ingestion/tests/unit/observability/data_quality` suites plus the new integration file pass (819 passed, 1 skipped). The new tests were confirmed RED against the unpatched metric.

#
### Note for reviewers — behavior change on MySQL / MariaDB

On MySQL, integer and boolean columns previously had their `0` / `false` rows counted as missing; on MariaDB, `TIME` columns had their `'00:00:00'` rows counted. Anyone who calibrated `missingCountValue` against those inflated numbers will see their test go from green to red after this fix. The new number is the correct one, but it is worth a release note.

Blast radius is limited to this DQ test: `NullMissingCount` is a distinct metric from the profiler's `NullCount` (`static/null_count.py`), which never had the `= ''` branch, so stored profiles and `nullProportion` history are untouched.

One deliberate trade-off: MySQL permits `''` as a real `ENUM` member, so an enum column with an empty label will no longer have those rows counted. PostgreSQL native enums raise on the comparison, so avoiding the hard error was preferred over that edge case. Anyone affected can add `""` explicitly via `missingValueMatch`, which is routed through `countInSet` and unaffected by this change.

The predicate fails open by design, so six types in `supportedDataTypes` that are not in `CommonMapTypes._TYPE_MAP` (`TIMESTAMPZ`, `VARIANT`, `GEOMETRY`, `POINT`, `POLYGON`, `LOWCARDINALITY`) still emit the comparison. That is intentional — `LOWCARDINALITY(String)` genuinely can hold `''`, and `TIMESTAMPZ` is unreachable from SQL ingestion because `column_type_parser` normalizes `TIMESTAMPTZ` to `TIMESTAMP`. If one of the exotic ones ever surfaces a `point = ''` report, it belongs in `NON_EMPTY_STRING_TYPES`.

#

Not applicable.

#
### Checklist:


ingestion/src/metadata/profiler/metrics/static/null_missing_count.py
```diff
@@ -32,6 +38,26 @@
...
+
+def can_hold_empty_string(col_type: TypeEngine) -> bool:
+    """Whether comparing a column of this type against '' is meaningful.
+
+    Types we don't recognize fall through to True so dialect-specific string types
+    keep counting empty values.
+    """
+    if is_quantifiable(col_type) or is_date_time(col_type):
+        return False
+    return not isinstance(col_type, NON_EMPTY_STRING_TYPES)
```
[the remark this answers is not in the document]
TeddyCr: deliberate choice to fail open (vs closed)

IceS2 (approved):

IceS2 (dismissed):

ingestion/src/metadata/profiler/metrics/static/null_missing_count.py
```diff
@@ -32,6 +38,26 @@
 logger = profiler_logger()
 
 
+# Types that cannot hold '', beyond the quantifiable/date-time families. Postgres
+# and ClickHouse raise on the comparison; MySQL/MariaDB coerce instead and count
+# real values as missing ('' matches 0, false and, on MariaDB TIME, '00:00:00').
+# CustomTime is listed here rather than in registry.is_date_time because that
+# helper also drives min/max/mean/stddev and columnValuesToBeBetween, which we are
+# not changing here. TODO(#21482): close that registry gap separately.
+NON_EMPTY_STRING_TYPES = (Boolean, UUIDString, CustomIP, CustomTime, Interval, SqlEnum)
```
Khairajani: JSON and array columns can hit this too — `col = ''` errors on both Postgres and MySQL, so `sqlalchemy.JSON`, `sqlalchemy.ARRAY` and `CustomArray` could go in here as well.

Binary types are fine as-is though — an empty blob really does match `''`, so excluding those would drop real counts.

TeddyCr: Good catch — fixed in f3b0dc1, along with tests.

`sqlalchemy.JSON`, `sqlalchemy.ARRAY` and `CustomArray` are all in `NON_EMPTY_STRING_TYPES` now. `CustomArray` has to be named explicitly: it's a `TypeDecorator`, so `isinstance(CustomArray(...), ARRAY)` is `False` and the bare `ARRAY` entry alone would miss the type the ORM converter actually builds for `DataType.ARRAY`.

Validated end to end against Postgres 16, ingesting a table into a local OM and running `columnValuesMissingCount` on each column:

| column | Postgres type | before | after |
|---|---|---|---|
| `col_json` | `json` | Aborted — `operator does not exist: json = unknown` | Success `nullCount=1` |
| `col_jsonb` | `jsonb` | Aborted — `invalid input syntax for type json` | Success `nullCount=1` |
| `col_int_arr` | `integer[]` | **Failed — `nullCount=2` vs expected 1** | Success `nullCount=1` |
| `col_text_arr` | `text[]` | **Failed — `nullCount=3` vs expected 2** | Success `nullCount=2` |
| `col_text` | `text` | Success `nullCount=3` | Success `nullCount=3` (unchanged) |

The two `Failed` rows are the silent ones — a wrong count written to the test result, no error anywhere in the logs.

Agreed on binary, left alone: an empty `bytea` matches `''` and counts, which is the behaviour we want.

Tests: the existing "`= ''` is absent from the compiled SQL" assertion can't catch an array, since the comparison renders as `= ARRAY[]` — that assertion passes on the broken code. So the unit tests now also pin the CASE branch count (one branch = NULL-only), and `DataType.JSON`/`DataType.ARRAY` were added to the converter-mapping test. The silent miscount is only visible against a real engine, so the integration test gained a JSON column and a Postgres-only array table with an empty-array row. Reverting just the guard tuple fails 12 unit cases and 3 integration cases; with the fix, 104 pass.

Unrecognized types still fail open, as before.

Khairajani (approved):

PR #31134, opened 2026-08-06
fix: special charcater decoding from datadiff username

### LINKED ISSUE #31124: Double encoding in data diff
### Affected module

Data Quality / Profiler

### Describe the bug

Service URI is sent encoded to the data diff package which encodes it again

### To Reproduce

n/a

### Expected behavior

n/a

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

_No response_

_No response_

### Additional context

_No response_

### Pre-submission checklist

TeddyCr:
### Describe your changes:

  Fixes #31124

  The `serviceUrl` we hand to `data-diff` is a canonical SQLAlchemy URL, and SQLAlchemy
  percent-encodes the username when it renders one. `data-diff`'s URI parser only decodes the
  password, host and query string — never the userinfo username — so an encoded username reaches
  the driver still encoded and `user@corp.com` authenticates as `user%40corp.com`.

  `render_url_for_data_diff()` renders the URL so every component survives exactly one
  encode/decode round trip: the username is handed over decoded (except for `:/?#`, which must
  stay encoded or the authority no longer parses, and which we log a warning about since
  `data-diff` will not decode them back), while the password stays encoded because `data-diff`
  does decode it. `TableParameter.data_diff_service_url` applies this at the single point where
  the URL leaves us; connection dicts pass through untouched. The stored `serviceUrl` is
  unchanged and remains a valid SQLAlchemy URL.

  Two related items rode along:

  - **Duplicate key detection.** `data-diff`'s result model assumes the key is unique — it folds
    rows into a `{key: sign}` map and asserts each key appears at most twice. A non-unique key
    therefore fails deep inside the library (`ValueError: Duplicate primary keys` on joindiff, a
    bare `AssertionError` on hashdiff), or silently inflates counts through the join fan-out.
    None of that tells the user which column is at fault. `_validate_key_uniqueness()` now runs a
    count/count-distinct on the key columns first and aborts with the table, the column(s), the
    row/distinct-key counts and up to 5 offending values. The check fails open: if it cannot run,
    we log and continue rather than failing an otherwise valid test. The test's WHERE clause is
    applied (it may itself make the key unique); sampling is not (a key unique only within one
    random sample is not a key).
  - **Build hygiene.** `make build-ingestion-base-slim-local` was defined under a duplicate
    `build-ingestion-base-local` target name and so was unreachable. `scripts/datamodel_generation.py`
    now silences the `format of 'X' not understood` warnings for the custom `format` vocabulary
    OpenMetadata owns (`queryBuilder`, `utc-millisec`, …) — only those, so a genuinely new or
    misspelled format still surfaces — and passes `--formatters black isort` explicitly, since
    external formatters are becoming opt-in upstream and the post-processing depends on black's
    output shape.

  #
  ### Type of change:
  - [x] Bug fix

  #
  ### High-level design:

  N/A — small change.

  #
  ### Tests:

  #### Use cases covered
  - Table diff on a Snowflake service whose username is an email (`user@corp.com`) authenticates
    successfully, with both password and private-key auth
  - A user-overridden `serviceUrl` with special characters is decoded the same way
  - A table diff configured with a non-unique key column aborts with a message naming the table,
    the column and sample duplicated values instead of failing with an opaque `AssertionError`
  - A key-uniqueness check that cannot run (permissions, unsupported dialect) does not fail the
    test — the diff proceeds

  #### Unit tests
  - Files added:
    - `ingestion/tests/unit/observability/data_quality/validations/test_data_diff_url.py` —
      round-trips every URL component through `data-diff`'s own parser, covers the double-encoding
      regression, reserved-character handling, the warning path, dict pass-through, and end-to-end
      Snowflake `serviceUrl` construction
    - `ingestion/tests/unit/observability/data_quality/validations/table/sqlalchemy/test_table_diff.py` —
      asserts `connect_to_table` receives a decoded username on both diff paths without mutating the
      stored URL, plus `DuplicateKeyError` message shapes, `_validate_key_uniqueness` ordering/skip
      behaviour, and the fail-open paths
  - Run with `make unit_ingestion` / `pytest ingestion/tests/unit/observability/data_quality`

  - [x] Not applicable — covered by unit tests against `data-diff`'s real URI parser.

  - A user-overridden `serviceUrl` with special characters is decoded the same way
  - A table diff configured with a non-unique key column aborts with a message naming the table,
    the column and sample duplicated values instead of failing with an opaque `AssertionError`
  - A key-uniqueness check that cannot run (permissions, unsupported dialect) does not fail the
    test — the diff proceeds

  #### Unit tests
  - Files added:
    - `ingestion/tests/unit/observability/data_quality/validations/test_data_diff_url.py` —
      round-trips every URL component through `data-diff`'s own parser, covers the double-encoding
      regression, reserved-character handling, the warning path, dict pass-through, and end-to-end
      Snowflake `serviceUrl` construction
    - `ingestion/tests/unit/observability/data_quality/validations/table/sqlalchemy/test_table_diff.py` —
      asserts `connect_to_table` receives a decoded username on both diff paths without mutating the
      stored URL, plus `DuplicateKeyError` message shapes, `_validate_key_uniqueness` ordering/skip
      behaviour, and the fail-open paths
  - Run with `make unit_ingestion` / `pytest ingestion/tests/unit/observability/data_quality`

  - [x] Not applicable — covered by unit tests against `data-diff`'s real URI parser.

  #### Playwright (UI) tests

  #### Manual testing performed
  1. Configured a Snowflake service with an email-style username and ran a `tableDiff` test —
     previously failed authentication, now connects.
  2. Ran a `tableDiff` on a table with a duplicated key column — test result is `Aborted` with the
     column name and sample duplicate values in the result message.
  3. Ran `make generate` — no `format not understood` warnings, generated models unchanged.
  4. Ran `make build-ingestion-slim-local` — builds the slim image.

  #

  Not applicable.

  #
  ### Checklist:
  - [x] My PR is linked to a GitHub issue via `Fixes #31124` above.

  
  - [x] I have added a test that covers the exact scenario we are fixing (`test_it_stops_double_encoding_the_username`, referencing #31124).

  Suggested PR title: Fixes #31124: stop double-encoding the username in the data-diff service URL

  Two things to verify before posting — I inferred them from the diff rather than running anything: the manual test steps above (I did not run a live Snowflake
  diff), and whether you want the duplicate-key detection in this PR at all, since it's a distinct behaviour change from the encoding fix and might read better as
  its own issue/PR.


ingestion/src/metadata/data_quality/validations/utils.py
```diff
@@ -54,6 +64,44 @@ def get_bool_test_case_param(
...
+
+    userinfo = _encode_username_for_data_diff(url.username)
+    if userinfo != url.username:
+        logger.warning(
+            "[Data Diff]: The username contains characters reserved by the connection URI (%s). "
+            "data-diff does not decode them, so authentication may fail.",
+            ", ".join(sorted(set(url.username) & set(USERNAME_RESERVED_CHARACTERS))),
+        )
+    if url.password is not None:
+        userinfo += f":{quote(str(url.password), safe=' +')}"
```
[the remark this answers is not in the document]
TeddyCr: Not an issue in practice. data-diff parses with `dsnparse.parse(..., parse_class=CustomParseResult)`, which `unquote`s the password (`data_diff/databases/_connect.py:36`), and `urlsplit` accepts a space in the netloc. The space also sits after the `:` and before the `@`, so it can't move the userinfo/authority boundary. Both forms round-trip identically:

```
'postgresql://user:pa ss@host:5432/db'   -> user / 'pa ss' / host / 5432
'postgresql://user:pa%20ss@host:5432/db' -> user / 'pa ss' / host / 5432
```

This is already covered by the `spaces-in-credentials` round-trip case in `test_data_diff_url.py`.

Worth noting `safe` isn't there for the space — it's overriding `quote`'s default `safe='/'`. Dropping it entirely would leak a literal `/` from the password and break dsnparse's credential regex (guarded by `test_it_keeps_the_password_encoded`).

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
```diff
@@ -260,36 +294,37 @@ def _run(self) -> TestCaseResult:
...
-                # depending on the data, this require scanning a lot of data
-                # so we only log the sample in debug mode. data can be sensitive
-                # so it is masked by default
-                for s in islice(
-                    self.safe_table_diff_iterator(),
-                    10 if logger.level <= logging.DEBUG else 0,
-                ):
-                    logger.debug("%s", str([s[0]] + [masked(st) for st in s[1]]))
-            test_case_result = self.get_row_diff_test_case_result(
+        with self._duplicate_keys_named(table_diff_iter):
```
IceS2, line 299: The with block extends to the end of _run, so it also covers _compute_row_count (line 318) and both
get_row_diff_test_case_result calls. Any AssertionError raised by those gets run through
_table_with_duplicate_keys, and if result_list happens to contain a repeated (sign, key) it will be reported to the user as a duplicate-key problem it isn't.

ingestion/src/metadata/data_quality/validations/table/sqlalchemy/tableDiff.py
```diff
@@ -586,6 +621,60 @@ def _validate_dialects(self):
...
+        already failed, and neither queries anything.
+        """
+        try:
+            yield
+        except ValueError as exc:
+            if DUPLICATE_KEY_MESSAGE not in str(exc):
+                raise
+            # joindiff does not say which of the two tables it found the duplicates in.
+            raise DuplicateKeyError(self._diff_key_columns()) from exc
+        except AssertionError as exc:
```
IceS2, line 649: AssertionError originates in DiffResultWrapper._get_stats, which only runs on the not threshold or computePassedFailedRowCount branch at line 298. On the other branch (threshold set, computePassedFailedRowCount false) calculate_diffs_with_limit never calls _get_stats, so hashdiff duplicates still silently inflate the count exactly as they do today.

IceS2 (dismissed):

PR #31135, opened 2026-08-06
fix(mssql): foreign keys, temporal period columns and query-log truncation visibility

IceS2:
Three independent fixes found while testing the MSSQL connector on 2.0.

### Foreign keys were silently dropped

MSSQL's `get_foreign_keys` never returned `referred_database`, so `common_db_source`
built the referred table FQN with a `None` database, failed to look the table up and
discarded the constraint. No FK ever reached the catalogue.

`MSSQL_GET_FOREIGN_KEY` now selects `DB_NAME() AS referred_database`, mirroring how
`POSTGRES_FETCH_FK` returns `con_db_name`. It rides the query already being issued, so
there is no extra roundtrip, and it stays correct after a `USE`.

### Temporal tables lost their period columns

The dialect skipped every column with `generated_always_type in (1, 2)`, so the
`SysStartTime`/`SysEndTime` pair of a system-versioned table was missing from the
catalogue entirely. The columns are now reflected; the sampler excludes them from
sample-data queries instead, since `SELECT` against a period column is what the skip was
guarding.

`get_temporal_column_names` is shared by the MSSQL and AzureSQL samplers — AzureSQL had
its own private copy.

### Query-log truncation was invisible

`usage_source` and `lineage_source` logged "Query log file is truncated" at `debug`, so a
run that silently processed a fraction of the queries looked complete. Raised to
`warning` and reworded to say the results may be incomplete.


IceS2:
Looked at the temporal-column query finding. Keeping it as is, for two reasons.

`fetch_sample_data` is called once per sampler in `sampler_interface.generate_sample_data`, and a
sampler is built per table — so caching per `(schema, table)` can never hit. The cache would only
ever be written, never read.

The other suggestion, carrying a `generated_always` flag on the reflected columns, doesn't reach
this code: the sampler filters the OM `Column` entities it was handed, not live SQLAlchemy
reflection output, so the flag would have to be persisted on the entity — a spec change well
outside this fix.

What remains is one indexed `sys.columns` lookup per sampled table, issued immediately before a
`TABLESAMPLE` read of that same table. Worth revisiting if sample-data ingestion ever shows up in a
profile, but not on the evidence here.

Khairajani (commented):
Replacing this with inline comments on the specific lines.

ingestion/src/metadata/sampler/sqlalchemy/mssql/sampler.py
```diff
@@ -47,3 +50,12 @@ def get_sample_query(self, static: StaticSamplingConfig, *, column=None) -> CTE:
...
         return query.cte(f"{self.get_sampler_table_name()}_sample")
+
+    def fetch_sample_data(self, columns: Optional[List[Column]] = None) -> TableData:  # noqa: UP006, UP045
+        """Period columns are catalogued but kept out of the sample: they are row-validity
+        bookkeeping, not user data, and classifying them adds noise (issue #21329).
+        """
+        if columns:
+            temporal_cols = get_temporal_column_names(self)
+            columns = [col for col in columns if col.name not in temporal_cols]
+        return super().fetch_sample_data(columns)
```
Khairajani: If the filter removes *every* column, `columns` becomes `[]` — and the base method reads an empty list as "caller gave me nothing, use all of them":

```python
if not columns:
    sqa_columns = [col for col in inspect(ds).c if col.name != RANDOM_LABEL]  # everything
```

So "drop these columns" silently flips to "keep all of them", period columns included.

```suggestion
        if not columns:
            return TableData(columns=[], rows=[])
        return super().fetch_sample_data(columns)
```

ingestion/src/metadata/sampler/sqlalchemy/azuresql/sampler.py
```diff
@@ -82,7 +61,7 @@ def get_sample_query(self, static: StaticSamplingConfig, *, column=None) -> CTE:
     def fetch_sample_data(self, columns: Optional[List[Column]] = None) -> TableData:  # noqa: UP006, UP045
         sqa_columns = []
         if columns:
-            temporal_cols = self._get_temporal_column_names()
+            temporal_cols = get_temporal_column_names(self)
```
Khairajani: Same empty-list problem here, and this one is reachable in practice.

`sqa_columns if columns is not None else None` still hands `[]` to the base method, which treats it as "use all columns". `NOT_COMPUTE_PYODBC` feeds this same list — so a table whose columns are *all* `SQASGeography`/`UndeterminedType` un-filters straight back into the pyodbc error #27960 was written to prevent.

Worth guarding both samplers the same way: return empty `TableData` when nothing survives the filter.

ingestion/src/metadata/ingestion/source/database/mssql/utils.py
```diff
@@ -228,9 +226,6 @@ def get_columns(self, connection, tablename, dbname, owner, schema, **kw):  # py
     cols = []
     for row in cursr.mappings():
         name = row[columns.c.column_name]
```
Khairajani, line 228: This removes the period-column skip but nothing tests the new behaviour. Worth a `get_columns` test asserting a period column comes back — `test_mssql.py` already has mocked-cursor dialect tests to copy.

ingestion/src/metadata/ingestion/source/database/usage_source.py
```diff
@@ -152,9 +152,9 @@ def yield_table_queries(self) -> Iterable[TableQuery]:
                     logger.info(f"Processed {row_count} query log entries for usage")
                     result_limit = getattr(self.source_config, "resultLimit", None)
                     if isinstance(result_limit, int) and row_count >= result_limit:
-                        logger.debug(
+                        logger.warning(
```
Khairajani: This sits inside `for days in ...` and `for engine in ...`, and multi-DB sources yield one engine per database — so 20 databases over a 30-day window emits it up to 600 times per run. `lineage_source.py` is the same, once per engine. A run-scoped flag so it fires once would keep the signal without the flood.

ingestion/src/metadata/sampler/sqlalchemy/mssql/sampler.py
IceS2: Good catch, fixed in b7c8210.

`fetch_sample_data` now returns an empty `TableData` when the filter leaves nothing, rather than handing `[]` to the base method.

Worth flagging that the AzureSQL sampler had a test on `main` pinning the old behaviour — `test_all_columns_filtered_passes_empty_list_not_original` asserted the empty list *must* reach `super()`. That is precisely the inversion, so I rewrote it as `test_all_columns_filtered_returns_no_sample`, asserting the base is never reached and the returned `TableData` is empty.

ingestion/src/metadata/sampler/sqlalchemy/azuresql/sampler.py
IceS2: Fixed in b7c8210 — same guard, and I added `test_all_columns_unsupported_by_pyodbc_returns_no_sample` for exactly the case you describe: a table of only `SQASGeography` columns now returns an empty `TableData` instead of un-filtering back into the type -151 crash.

The `sqa_columns if columns is not None else None` dance is gone too; `None` still means "sample everything", anything else is filtered and an empty result short-circuits.

ingestion/src/metadata/ingestion/source/database/mssql/utils.py
IceS2: Added in b7c8210 — `TestMssqlTemporalPeriodColumns` in `test_mssql.py`, built on the mocked-cursor pattern as suggested.

`test_period_columns_are_reflected` feeds `get_columns` a row set of `id` / `SysStartTime` / `SysEndTime` where the period columns carry `generated_always_type` 1 and 2, and asserts all three come back. I checked it is not vacuous: restoring the old skip makes it fail with `assert ["id"] == ["id", "SysStartTime", "SysEndTime"]`.

ingestion/src/metadata/ingestion/source/database/usage_source.py
IceS2: You are right about the volume. Fixed in b7c8210.

The notice moved to `QueryParserSource.warn_if_query_log_truncated`, behind a run-scoped `_result_limit_warned` flag, and both sources call it. Unifying also removes the copy that had already drifted between the two files.

`test_the_notice_fires_once_per_run` pins it. Removing the flag makes it fail with **12** notices for usage (3 engines x 4 days) and **3** for lineage, which matches the arithmetic you described.

ingestion/src/metadata/sampler/sqlalchemy/mssql/sampler.py
```diff
@@ -47,3 +50,12 @@ def get_sample_query(self, static: StaticSamplingConfig, *, column=None) -> CTE:
...
         query = self.get_client().query(rnd)
         return query.cte(f"{self.get_sampler_table_name()}_sample")
+
+    def fetch_sample_data(self, columns: Optional[List[Column]] = None) -> TableData:  # noqa: UP006, UP045
+        """Period columns are catalogued but kept out of the sample: they are row-validity
+        bookkeeping, not user data, and classifying them adds noise (issue #21329).
+        """
+        if columns:
+            temporal_cols = get_temporal_column_names(self)
+            columns = [col for col in columns if col.name not in temporal_cols]
```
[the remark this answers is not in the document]
IceS2: The round-trip count is right, but caching per `(schema, table)` would not remove it. `generate_sample_data` calls `fetch_sample_data(self.columns)` once per table entity (`sampler_interface.py:173`), and the sampler is constructed per table — so every key is looked up exactly once and the cache would never hit. It would add memory and staleness for no saved queries.

The other suggestion is the real fix: carry the `generated_always` flag through reflection so the sample path can filter without asking again. That is a dialect + reflection change reaching past this PR, so I have left it out here rather than half-doing it.

For context on the cost: this is a `sys` catalogue lookup on an already-open session, alongside the sample query, row count, and metric queries the profiler already issues per table.

pmbrull (approved):

Khairajani (approved):
LGTM

PR #31136, opened 2026-08-06
fix(ui): stop a superseded Test Connection run from clobbering a newer result

IceS2:
Hitting **Test Connection** again while a run is still in flight leaves the first run's
timers alive. When they fire they overwrite the newer run's state, so the dialog can show
a stale success for a connection that has since been changed and now fails — or the
reverse.

`TestConnection` now keeps its timers in a ref and tags each run with an id. Callbacks
from a superseded run are ignored, and every timer is cleared on unmount, which also stops
the component polling a workflow after the dialog is closed.


IceS2:
Good catch — fixed in 9567a26a16.

The reachable path is the run's own expiry: the poll callback awaits `getWorkflowData`, the
2-minute timeout fires during that await and re-enables the button, and a run started from there
was then clobbered by the stale callback, which also cleared the *new* run's timers and left it
polling nothing. Re-checking `runId` after the await closes it.

Covered by a test that drives exactly that sequence; verified to fail with the new guard removed.

aniketkatkar97 (approved):

PR #31078, opened 2026-08-06
Fixes #31077: escape query_string reserved chars in test case/suite search

mohityadav766:
### Describe your changes:

Fixes #31077

I escaped Lucene `query_string` reserved characters in the free-text search value of the Test Cases and Test Suites lists, because pasting a value containing `:` or `/` (e.g. a URL) built an invalid `*<value>*` query and OpenSearch returned a 500 (`query_shard_exception: all shards failed`). The raw input is now run through the existing `escapeESReservedCharacters` util before being wrapped in wildcards, so the `*…*` contains-search still works while the inner text is matched literally.

#
### Type of change:
- [x] Bug fix

#
### High-level design:
N/A — small change.

#
#### Use cases covered
- Searching the Test Cases / Test Suites list by a term containing `:` `/` (e.g. a pasted URL) no longer 500s; the wildcard contains-search still matches normal terms.

#### Unit tests
- [x] Added a regression test in `useTestCaseList.test.ts` using the reported URL; verified it fails without the fix and passes with it.
- Files: `useTestCaseList.test.ts` (+ fix in `useTestCaseList.ts` and `useTestSuitesData.ts`).
- `useTestCaseList` + `useTestSuitesData` suites: 27 tests pass. (Coverage % not captured — this worktree has no installed node_modules; tests were run with a borrowed copy.)

#### Backend integration tests
- [x] Not applicable (no backend changes).

#### Ingestion integration tests
- [x] Not applicable (no ingestion changes).

#### Playwright (UI) tests
- [x] Not applicable — the exact regression is covered by the unit test above.

#### Manual testing performed
Automated: ran the Jest suites above (27 pass; new test is red before the fix, green after).
Reproduction (UI): Data Quality → Test Cases (and Test Suites) → type a value with `:`/`/` such as `https://<host>/test-case-results` → before: list shows a search error; after: search returns results/empty with no 500.

#
### UI screen recording / screenshots:
Not applicable — non-visual fix (removes a 500 on search; no layout change).

#
### Checklist:
- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [x] My PR title is `Fixes <issue-number>: <short explanation>`
- [x] My PR is linked to a GitHub issue via `Fixes #31077` above.
- [x] I have commented on my code, particularly in hard-to-understand areas.
- [x] For UI changes: N/A — non-visual fix.
- [x] I have added tests and listed them above.
- [x] I have added a test that covers the exact scenario we are fixing.

[...]

PR #31142, opened 2026-08-06
Fixes 31077: parse free-text data quality search as literal text, not Lucene

### LINKED ISSUE #31077: Test case / test suite search returns 500 (query_shard_exception) for terms with Lucene reserved characters (e.g. pasted URL)
### Affected module
Data Quality — Test Cases and Test Suites list search (UI).

### Description
Searching the Test Cases or Test Suites list with a term that contains Lucene `query_string` reserved characters — most easily reproduced by pasting a URL — makes the list fail with a 500.

The UI builds the search as a "contains" query by wrapping the raw input in wildcards (`` `*${searchValue}*` ``) and sends it as the `q` param. When `searchValue` contains characters like `:` or `/`, the resulting `q` (e.g. `*https://host/test-case-results*`) is an invalid `query_string`, so OpenSearch/Elasticsearch throws `query_shard_exception` and the request 500s.

### Steps to reproduce
1. Go to **Data Quality → Test Cases** (or **Test Suites**).
2. In the search box, enter a value containing `:` or `/`, e.g. `https://<host>/observability/test-case/<fqn>/test-case-results`.
3. The list errors out instead of returning results.

### Server-side error
```
org.openmetadata.sdk.exception.SearchException: Search failed due to Request failed: [search_phase_execution_exception] all shards failed | Root cause: [query_shard_exception: Failed to parse query [*https://.../test-case-results*]]
    at org.openmetadata.service.resources.dqtests.TestCaseResource.listFromSearch(TestCaseResource.java:491)
    ...
```

### Root cause
Free-text search input is wrapped in wildcards and passed straight into a `query_string` query with no escaping of Lucene reserved characters. Any `q` containing `: / ( ) [ ] { } ^ " ~ \` etc. that doesn't form valid `query_string` syntax fails to parse.

### Expected behavior
Reserved characters in the user's search term should be treated literally (escaped) so the wildcard contains-search works and never 500s.

TeddyCr:
### Describe your changes:

Fixes #31077

Reopening the scope of #31077: it was closed by #31078, which escaped the search term **in the UI**. The server was never changed, so the 500 is still reachable by any direct API caller on current `main`:

```bash
curl -i -H "Authorization: Bearer $TOKEN" --get \
  --data-urlencode 'q=https://localhost:8585/table/orders' \
  http://localhost:8585/api/v1/dataQuality/testCases/search/list
# HTTP/1.1 500 — query_shard_exception: Failed to parse query [...]
```

This fixes it at the layer that owns the contract, and removes the UI-side escaping, which becomes actively harmful once the server treats `q` as literal text.

**Why not escape.** I started with a server-side escaper (a `QUERY_STRING_SYNTAX_CHARACTERS` denylist mirroring Elasticsearch's documented reserved set) and abandoned it. A denylist cannot be made correct here:

- **`<` and `>` cannot be escaped at all.** ES's own docs: *"Note that `<` and `>` can't be escaped at all. The only way to prevent them from attempting to create a range query is to remove them from the query string entirely."* A backslash in front of them does nothing.
- **Reserved *words* are not characters.** `?q=NOT` or a trailing `foo AND` still raises a ParseException. No character list addresses that.
- **The escaper cannot tell "already escaped" from "contains a backslash."** `entityName` permits everything except `>`, `"` and control chars (`^((?!::)[^>"\x00-\x1f])*$`), and `fullyQualifiedEntityName` has **no pattern at all** — so `\` is legal in names and common in SQL-Server-derived ones. An escape-aware escaper passes `db\schema` through unchanged, ES reads `\s` as a literal `s`, and the term silently becomes `dbschema`: no error, wrong answer.

The input alphabet is effectively all of Unicode; a fixed 19-character denylist against it is a mitigation, not a fix.

**What this does instead.** `simple_query_string` is built for untrusted input: its parser discards syntax it cannot read rather than throwing, and `flags=NONE` disables every operator so the input is literal text. There is no character list to keep in sync with the query language.

Only the **fuzzy** branch of `buildSearchQueryBuilderV2` could ever throw — the non-fuzzy branch is a `multi_match`, which does not parse Lucene. That one branch switches, gated by a `freeText` flag threaded from the `*/search/list` entry point, following the existing `includeExplain` / `includeAggregations` threading convention in the same factory.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

`buildSearchQueryBuilderV2` serves two endpoint families with opposing `q` contracts, so the choice has to come from the caller:

| Endpoint | `q` contract | Parser |
|---|---|---|
| `GET /v1/dataQuality/testCases/search/list` | free text | `simple_query_string` |
| `GET /v1/dataQuality/testSuites/search/list` | free text | `simple_query_string` |
| `GET /v1/dataQuality/testCases/testCaseResults/search/{list,latest}` | free text | `simple_query_string` |
| `GET /v1/search/query`, `/v1/search/export` | documented Lucene syntax | `query_string` |

`freeText` is only read by the data quality branches, so `*/search/list` on every other index is unaffected. Tests pin **both** directions — literalising `/v1/search/query` would turn a field-scoped query into 0 hits with HTTP 200, which is a silent wrong answer on a published contract.

**Substring matching moved server-side.** It previously came from the UI wrapping the term in `*…*`. With `q` literal there is no wildcard, and the existing `*.ngram` fields are `edge_ngram` — token prefixes only, so `alues` would stop matching `column_values_to_be_between`. New `name.substring` / `displayName.substring` fields back true mid-token matching. They use a plain `ngram` tokenizer (`min_gram: 3`, `max_gram: 4`, within the default `max_ngram_diff` of 1) rather than the `wildcard` field type, which **OpenSearch only added in 2.15** while OpenMetadata supports 2.6+. Scoped to `TestCaseIndex.getFields()` rather than `SearchIndex.getDefaultFields()` so no other index pays the index-size cost.

**No migration file.** `IndexMappingVersionTracker` hashes mapping content, so `SmartReindexPlan` detects the change and reindexes only `testCase` / `testSuite` on upgrade.

**One defect fixed along the way.** `ES_RESERVED_CHARACTERS` (`StringUtils.ts`) is looked up by the character class `/[\\[\]#+=&|><!(){}^"~*?:/-]/g`, which matches one character at a time — so the `'&&'` and `'||'` keys are **unreachable**, a pair can never be looked up as a pair. With no `'|'` key, a lone pipe fell through to `?? char` and was replaced with itself, leaving `a||b` to lex as a boolean OR at all 18+ call sites. The helper had no direct test coverage; it now has it. This is kept because the helper still legitimately serves `/v1/search/query` callers.

#
#### Use cases covered
- Reserved characters return **200** from both DQ list endpoints — including `count < 10`, `count > 10`, `NOT` and `orders AND`, the four cases escaping could never have covered.
- A reserved-character term still **matches** the entity it names, not merely "does not 500".
- A mid-token substring (`alues` → `column_values_to_be_between`) matches with no caller-supplied wildcard.
- `/v1/search/query` on a DQ index still honours Lucene syntax.
- The picker lists its suite's test cases and excludes another table's.
- `escapeESReservedCharacters` escapes the full reserved set including a lone `|`.

#### Unit tests
- [x] Added.
- Java: `SearchSourceBuilderFactoryTest` — asserts on the **serialized query document**, pinning `simple_query_string` + `flags=NONE` for free text, `query_string` for `/v1/search/query`, and the presence of the `*.substring` fields.
- UI: `useTestCaseList`, `useTestSuitesData`, `AddTestCaseList`, `AddToBundleSuiteModal`, `AddTestSuitePipeline`, `StringUtils` (+53, the helper's first direct coverage).
- `mvn test -pl openmetadata-service -Dtest='org.openmetadata.service.search.**'` → **2156 run, 0 failures**
- UI Jest, DataQuality + StringUtils → **982 passed, 2 skipped, 67 suites**

**Both new backend assertions were verified RED before GREEN, by mutation:**

| Mutation | Result |
|---|---|
| `freeText` ignored in the fuzzy branch (always `query_string`) | *"Free-text q must use the parser that never throws"* fails — output reverts to `query_string` |
| `*.substring` fields dropped from `TestCaseIndex.getFields()` | *"testDataQualitySearchQueriesTheSubstringFields"* fails — the fields vanish from the `multi_match` |

- [x] Added, against a real Elasticsearch container.
- `TestCaseResourceIT`, `TestSuiteResourceIT`: reserved characters return 200; the term still matches; mid-token substring matching; picker scope + exclusion; the `/v1/search/query` Lucene pin; both `testCaseResults` search endpoints.
- **NOT EXECUTED in this session** — see Known limitations.

#### Playwright (UI) tests
- [x] Updated — the specs that asserted the escaped-and-wrapped `q` now assert the raw term; the pasted-URL case asserts a 200.
- **NOT EXECUTED** — needs the full Docker stack.

#### Manual testing performed

**NOT VERIFIED by hand.** Steps for a reviewer with a live stack:

1. Start the stack from **this branch** — the backend must come from here.
2. Data Quality → Test Cases. Paste `https://localhost:8585/table/orders` → no error toast, `search/list` returns **200**.
3. Repeat with `orders (v2)`, `count < 10`, `a||b`, `NOT`, `trailing\` → 200 each time. On `main`, several of these still 500.
4. Search `alues` → `column_values_to_be_between` still appears (substring matching survived).
5. Open a test suite pipeline → "Add test cases": the list is populated with the search box **empty** (regression guard: `*` is no longer a match-all).
6. Server-only, proving it is not a UI-only fix: the `curl` above → **200**.
7. `/v1/search/query?q=name.keyword:<existing test case>&index=testCase` → ≥ 1 hit. If 0, free-text handling has leaked into the shared builder.

#
### Checklist:
- [x] My PR is linked to a GitHub issue via `Fixes #<issue-number>` above (reusing #31077 — the UI half shipped in #31078, the server half is this PR).
- [x] For JSON Schema changes: index mapping change only; reindex is detected automatically by mapping hash, no migration needed.
- [ ] For UI changes: screen recording pending.

#
### Known limitations

- **This removes the UI escaping merged in #31078.** Once the server parses `q` as literal text, escaping in the client is wrong: `foo:bar` would arrive as `foo\:bar` and be matched *with* the backslash, causing silent misses. The escaping helper itself is kept — `/v1/search/query` callers still need it.
- **Mid-token matching is now `ngram(3,4)`**, so terms shorter than 3 characters no longer match via that field, and it costs index size on two indexes. `min_gram`/`max_gram` were chosen to stay within the default `max_ngram_diff` of 1.
- **The ITs and Playwright specs were not executed here.** The Java unit tests and UI Jest were, with the RED checks above.
- **OpenSearch is covered at unit level only** — the ITs bootstrap Elasticsearch. The OS builder is exercised by the serialized-document assertions in `SearchSourceBuilderFactoryTest`.
- **A reindex is required on upgrade** for `testCase` and `testSuite`. It is triggered automatically, but on a large catalog it is not free.
- `escapeESReservedCharacters` now escapes `|`. 8 of its call sites mock the helper in their own tests and so could not catch a regression.

The PR moves Data Quality free-text parsing into the server while preserving Lucene syntax for general search endpoints.
- Uses `simple_query_string` with operators disabled for Data Quality list and test-result searches.
- Adds substring-indexed name fields to preserve mid-token matching without UI-supplied wildcards.
- Removes Data Quality UI escaping and wildcard wrapping, updates sample data, and expands backend and frontend regression coverage.
- Corrects lone-pipe escaping for remaining Lucene-query callers.
</details>

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/search/SearchSourceBuilderFactory.java | Introduces explicit free-text overloads while retaining Lucene-compatible defaults for existing callers. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchSourceBuilderFactory.java | Selects literal `simple_query_string` parsing for free-text Data Quality searches and preserves `query_string` otherwise. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OpenSearchSourceBuilderFactory.java | Mirrors the Elasticsearch parser-selection behavior for OpenSearch. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TestCaseIndex.java | Adds substring-indexed name and display-name fields to Data Quality search queries. |
| openmetadata-spec/src/main/resources/elasticsearch/en/test_case_index_mapping.json | Defines the bounded ngram analyzer and substring fields used for mid-token matching. |
| openmetadata-ui/src/main/resources/ui/src/utils/StringUtils.ts | Makes lone-pipe escaping reachable while removing unreachable multi-character replacement entries. |
| openmetadata-ui/src/main/resources/ui/src/components/DataQuality/TestCases/useTestCaseList.ts | Sends Data Quality search terms according to the server’s new literal free-text contract. |
| openmetadata-service/src/test/java/org/openmetadata/service/search/SearchSourceBuilderFactoryTest.java | Pins parser selection and inclusion of substring search fields in serialized query documents. |

</details>

```mermaid
flowchart LR
  DQ[Data Quality search/list] -->|freeText=true| Factory[Search source builder]
  General[Search query/export] -->|freeText=false| Factory
  Factory -->|Data Quality free text| Simple[simple_query_string flags NONE]
  Factory -->|Documented Lucene syntax| Lucene[query_string]
  Simple --> Fields[name/displayName substring fields]
  Fields --> Backend[(Elasticsearch or OpenSearch)]
  Lucene --> Backend
```
</details>

<sub>Reviews (11): Last reviewed commit: ["fix(tests): remove two shared-estate fla..."](https://github.com/open-metadata/openmetadata/commit/acf6f4b9b8f76bed27b3026e7a3f6065d1cf1fa8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53402980)</sub>


ShaileshParmar11 (dismissed):

openmetadata-ui/src/main/resources/ui/src/components/DataQuality/AddTestCaseList/AddTestCaseList.component.tsx
```diff
@@ -269,10 +268,11 @@ export const AddTestCaseList = ({
...
         setIsLoading(true);
-        const globalSearch = searchText ? `*${searchText}*` : WILD_CARD_CHAR;
-        const q = testCaseFilters
-          ? `${globalSearch} && ${testCaseFilters}`
-          : globalSearch;
+        // `q` must stay free text: /v1/dataQuality/testCases/search/list parses it as a literal
+        // term, so scoping filters travel as first-class params via `testCaseParams`. With no
+        // search text `q` is omitted rather than sent as `*` — a lone asterisk is no longer a
+        // match-all wildcard, it is a literal asterisk that matches nothing.
+        const q = searchText || undefined;
```
manerow, line 275: **What changes.** Before this PR, `q` was parsed by `query_string`, where a bare `*` is a wildcard that matches every document. With `flags=NONE` it becomes a literal asterisk. I confirmed it in the serialized query at head:

```
before:  {"query_string":       {"query":"*","fields":[...]}}                -> matches all
after:   {"simple_query_string":{"query":"*","flags":"NONE","fields":[...]}} -> 0 hits
       + {"multi_match":        {"query":"*","operator":"and",...}}          -> 0 hits
```

**What breaks.** `q=*` was the idiom for "no search term, return everything". Callers using it now get an empty list with HTTP 200 and no error to surface. This file and the other three UI callers are updated in the PR. Collate is not: `collate-ui/src/main/resources/ui/src/components/Widgets/DataQualityWidget/DataQualityWidget.component.tsx:92` on `main` still calls `getListTestCaseBySearch({ q: '*', ... })`, and no open Collate PR changes it. The Data Quality landing-page widget will render empty as soon as Collate bumps the submodule. Any external API caller doing the same gets silent zero rows instead of a 400.

**Ask.** A companion Collate PR dropping `q: '*'` should land with this, and the server should either treat a bare `*` as match-all for backwards compatibility, or this should be listed as a breaking change in the release notes.

openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TestCaseIndex.java
```diff
@@ -148,6 +150,13 @@ public static Map<String, Float> getFields() {
     fields.put("testSuite.description", 1.0f);
     fields.put("entityLink", 3.0f);
     fields.put("entityFQN", 10.0f);
+    // The DQ list endpoints treat q as literal text, so a user's mid-token substring
+    // ("alues" in column_values_to_be_between) has no wildcard to fall back on. The
+    // *.ngram fields are edge_ngram, which only matches token prefixes; these are the
+    // only fields backing true substring matching, so they are scoped to the DQ
+    // indexes rather than added to SearchIndex.getDefaultFields().
+    fields.put(FIELD_NAME_SUBSTRING, 1.0f);
```
manerow, line 158: `name.substring` and `displayName.substring` are new multi-fields. Adding them to the mapping does not backfill existing documents, only reindexed ones get them, and `useTestCaseList.ts:117` stops sending `*...*` at the same time.

So between an upgrade and the testCase/testSuite reindex finishing, mid-token search returns nothing rather than degrading: `alues` stops finding `column_values_to_be_between`, which is the exact case the new fields exist to cover.

`IndexMappingVersionTracker` does detect the drift and `SystemRepository.computeMappingDrift` surfaces it, but detection is not application. The reindex still has to be run, and it is not free on a large catalog. The "No migration file, reindex is detected automatically by mapping hash" line in the description reads as "nothing operational required", which is not the case here. Reindexing should be made an explicit upgrade note.

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchSourceBuilderFactory.java
```diff
@@ -226,16 +231,22 @@ public Query buildSearchQueryBuilderV2(String query, Map<String, Float> fields)
...
-            query,
-            fuzzyFields,
-            Operator.And,
-            "1",
-            10,
-            3,
-            DEFAULT_TIE_BREAKER,
-            TextQueryType.MostFields);
+        freeText
+            ? ElasticQueryBuilder.simpleQueryStringQuery(query, fuzzyFields, Operator.And)
```
manerow, line 240: **What changes.** The `query_string` call this replaces passes `fuzziness:"1"`, `fuzzy_prefix_length:3`, `fuzzy_max_expansions:10` (lines 242-246 below). `simple_query_string` has no equivalent parameter, and `flags=NONE` also disables the `~` operator, so there is no way to keep fuzziness on this branch. Serialized query at head:

```
before: {"query_string":{"query":"...","fuzziness":"1","fuzzy_prefix_length":3,"fuzzy_max_expansions":10,...}}
after:  {"simple_query_string":{"query":"...","flags":"NONE",...}}
```

**What breaks.** Typo tolerance disappears from the Test Cases and Test Suites search boxes. A one-character typo after a 3-character prefix, which is what those parameters allowed, no longer matches. Same in the OpenSearch mirror at `OpenSearchSourceBuilderFactory.java:268`.

**Ask.** This is a separate regression from the 500 being fixed, and it is not in Known limitations. Either state it there explicitly, or keep a `match` clause with fuzziness on `name`/`displayName` in the free-text branch so the trade is "no Lucene operators" rather than "no operators and no fuzziness".

openmetadata-ui/src/main/resources/ui/src/components/DataQuality/AddTestCaseList/AddTestCaseList.component.tsx
TeddyCr: Correct, we'll need this one in to have the other PR up.

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchSourceBuilderFactory.java
TeddyCr: `query_string`'s `fuzziness` only applies to terms carrying the `~` operator — Lucene's parser calls `getFuzzyQuery()` only on `TERM~`; a bare term goes through `getFieldQuery()`, which ignores it. The param sets the default edit distance for `~` terms, it isn't implicit typo tolerance.

Verified against ES 9.3.0 with this branch's `test_case_index_mapping.json`, building both queries exactly as main and head build them (same field lists, boosts, `bool.should` + `minimum_should_match:1`), over `column_values_to_be_between` and `table_row_count_to_equal`:

| query | main (`query_string`, fuzziness=1) | head (`simple_query_string` + substring) |
|---|---|---|
| `column_values_to_be_between` | hit | hit |
| `column_val` (prefix) | hit | hit |
| `alues` (mid-token substring) | **0 hits** | **hit** |
| `betwene` (1-char typo) | **0 hits** | 0 hits |
| `column_valuse_to_be_between` (typo after 3-char prefix) | **0 hits** | 0 hits |
| `betwene~` (explicit `~`) | hit | 0 hits |
| `a AND (` | `Failed to parse query` | 0 hits, no error |

So the scenario described — "a one-character typo after a 3-character prefix, which is what those parameters allowed" — returns zero hits on main as well. It never matched. Meanwhile `alues` returns 0 hits on main and hits on head, via the new `name.substring` ngram fields (`TestCaseIndex.java:158`).

There's no typo-tolerance regression to document: head matches a strict superset of main's free-text inputs, minus the `~` operator, which is already covered by "no Lucene operators" in Known limitations.

Adding a fuzzy `match` clause on `name`/`displayName` would be *introducing* typo tolerance the DQ search boxes never had, not restoring it — happy to do it, but as a separate enhancement rather than as a fix for this PR.

openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TestCaseIndex.java
TeddyCr: reindex is re-run on upgrade

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchSourceBuilderFactory.java
manerow: You're right, verified. Withdrawing.

manerow (dismissed):

PR #31153, opened 2026-08-07
Fixes #31152: emit Snowflake Sink lineage in the KafkaConnect connector

### LINKED ISSUE #31152: KafkaConnect: Snowflake Sink connectors produce no lineage
## Describe the bug

The `kafkaconnect` connector emits **no lineage at all** for Snowflake Sink connectors (both Confluent Cloud fully-managed `SnowflakeSink` and self-managed `SnowflakeSinkConnector`), in every realistic configuration.

Two independent defects stack:

**1. No datasets are emitted.** The Snowflake Sink derives its target table from the topic name; no config key names the table directly unless `snowflake.topic2table.map` is set. The generic key-list dataset parsing therefore matches nothing, `found_values["table"]` is never set, and zero datasets are produced — so nothing downstream ever runs. When `snowflake.topic2table.map` *is* set, the parser keeps only the table side of each pair and discards the topic side, after which topic/table pairing degrades to positional index plus name equality — which fails by construction, since renaming is the whole point of the map.

**2. The Snowflake service cannot be resolved.** `CONNECTOR_CLASS_TO_SERVICE_TYPE` and `SERVICE_TYPE_HOSTNAME_KEYS` have no Snowflake entry, so `snowflake.url.name` cannot be matched to an OpenMetadata Snowflake service. The fallback builds a literal `"*.DB.SCHEMA.TABLE"` FQN that always 404s. Additionally, hostname matching probes only `hostPort`/`host`, and `SnowflakeConnection` has neither — it exposes `account` — so even with the constants added, matching cannot succeed without also teaching the probe about account-style identifiers.

Net effect: `lineageInformation.dbServiceNames` is effectively mandatory and undocumented, and even setting it does not help, because defect 1 means no dataset is ever produced.

## To Reproduce

1. Create a Confluent Cloud fully-managed Snowflake Sink with `input.data.format: AVRO`, `snowflake.enable.schematization: true`, one topic per table, and **no** `snowflake.topic2table.map` (the common 1:1 case).
2. Ingest the Kafka messaging service (with Schema Registry credentials) and the Snowflake database service into OpenMetadata.
3. Create a KafkaConnect pipeline service pointed at `https://api.confluent.cloud/connect/v1/environments/{env-id}/clusters/{lkc-id}` with a Cloud API key.
4. Run metadata ingestion.

**Expected:** `Topic → Pipeline → Snowflake Table` lineage, with column-level detail.
**Actual:** no lineage edges. Logs show `0 datasets` parsed and `connector.class 'SnowflakeSink' -> service type mapping: None`.

## Additional context

- The managed plugin reports `connector.class` as the **short** plugin name `SnowflakeSink`, not a Java FQCN — so registering only an FQCN-derived name misses every managed connector.
- The Confluent Cloud Connect v1 API **omits defaulted properties** entirely, so `snowflake.enable.schematization` and `snowflake.ingestion.method` are frequently absent. Resolution logic must not branch on their presence.
- Snowflake sanitises topic names that are not valid identifiers, appending `abs(javaStringHashCode(topic))` — e.g. topic `om-lineage-test` lands in table `OM_LINEAGE_TEST_702890019`.
- Snowflake schematization creates one column per **top-level** Avro field, uppercased; a nested record becomes a single `VARIANT` column rather than being flattened. Underscore-joined columns come only from a `Flatten` SMT.

ulixius9:
### Describe your changes:

Fixes #31152

I worked on making the `kafkaconnect` connector emit `Topic → Pipeline → Snowflake Table` lineage, with column-level detail, for Snowflake Sink connectors — because today it emits **nothing** in every realistic configuration. Two defects stacked: the Snowflake Sink derives its table from the topic name, so the generic key-list parser matched no table key and produced zero datasets (and when `snowflake.topic2table.map` *was* set, the topic side of each pair was discarded and pairing degraded to positional index plus name equality, which fails by construction); and the Snowflake service could not be resolved at all, because there was no service-type mapping and hostname matching probes only `hostPort`/`host`, which `SnowflakeConnection` does not have. Rather than add another branch to an already `# noqa: C901` function, I introduced a connector-class-keyed sink resolver registry, moved the existing key-list logic into a `DefaultResolver` verbatim, and gave Snowflake its own resolver.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

**Architecture.** New `ingestion/src/metadata/ingestion/source/pipeline/kafkaconnect/sinks/` package:

- `base.py` — `SinkDatasetResolver` ABC (`resolve_datasets` / `match_topic` / `column_mappings`), a `sink_resolver_registry` built on the repo's shared `enum_register()` primitive, `get_resolver()`, and `DefaultResolver`.
- `snowflake.py` — `java_string_hashcode()`, `snowflake_table_name()`, and `SnowflakeSinkResolver`, registered under **both** `SnowflakeSink` (the Confluent Cloud managed short plugin name) and `SnowflakeSinkConnector` (self-managed). Registering only one silently falls back to the default resolver with no error.

`metadata.py` now delegates instead of branching inline, so it **shrinks**. This follows the Factory/Registry pattern already documented in `docs/design-patterns.md`, and keeps connector-specific logic out of shared code — `metadata.py` contains zero occurrences of the string "snowflake"; all Snowflake knowledge lives in `constants.py` and `sinks/`.

**`DefaultResolver` is a verbatim move, which is what makes this safe.** Debezium/CDC, JDBC, and S3/GCS/Azure storage sinks all depend on that logic. It was differentially fuzzed against the original across 6,701 cases (6,365 `resolve_datasets` configs + 336 `match_topic` cases) with zero divergence in return values *or* logger call sequences. The pre-existing 1,586-line `test_kafkaconnect.py` passes untouched at every commit.

**Two model fields carry the fix** (`models.py`, hand-written Pydantic — not schema-generated, so no `make generate` is owed):
- `source_topic` — the topic↔table pairing was previously thrown away at parse time and guessed back positionally. Carrying it makes `match_topic` an O(1) exact lookup and deletes the guessing.
- `fully_qualified` — needed because `get_dataset_entity`'s Priority 1 hardcoded `database_name=None, schema_name=dataset.database`. That shape is CDC-specific (for Debezium, `database` holds `topic.prefix`, a logical server name, not a real database). A naive generalisation such as `database if schema else None` silently breaks Debezium configured with `table.include.list = inventory.orders`, so the resolver declares qualification explicitly instead and the 4-part FQN is built only for datasets that opt in.

**Service resolution** adds `SERVICE_CONNECTION_HOST_ATTRIBUTES` and `SERVICE_TYPE_HOST_DOMAIN_SUFFIXES` as data in `constants.py`, so hostname matching can also probe `account` and strip a service-type-keyed domain suffix. This is what lets `snowflake.url.name` match a real Snowflake service and makes `dbServiceNames` genuinely optional.

**Alternatives rejected.** Inlining the Snowflake special case was cheaper but the generic key-list machinery cannot express "derive the table from `topics` when no mapping key exists", and every sink added that way makes the next harder. A dedicated Confluent Cloud connector was rejected outright — it would fork ~2,800 lines of topic resolution, service matching, and lineage plumbing for one sink type, and Confluent Cloud is already a first-class path in the existing client.

**Backward compatibility.** Both new model fields default to preserving current behaviour, and no existing construction site passes either. `DefaultResolver` applies to every connector class without a dedicated resolver, so non-Snowflake paths are unchanged. `SnowflakeSinkResolver` is a strict superset of `DefaultResolver`: it accepts the `snowflake.database`/`snowflake.schema` variation keys and falls back to `DefaultResolver` when no topics are discoverable, so self-managed sinks cannot lose lineage they had.

**Deliberately out of scope:** migrating the storage-sink branch into the registry (drags S3/GCS/Azure regression risk into a customer-facing fix), and `ReplaceField` SMT support.

**One behaviour change does reach the storage-sink path, and it is intentional.** Sink topic discovery is now unified in `_search_topics_by_patterns`, which matches `topics.regex` with `pattern.fullmatch()`; the old `_search_topics_by_regex` used `pattern.match()`, a prefix match. Kafka subscribes a pattern with `Pattern.matcher(topic).matches()` — a full match — so the prefix behaviour was discovering topics the connector is not subscribed to and emitting lineage for them: `topics.regex=orders` claimed `orders_v1` and `orders_v2`. S3/GCS/Azure sinks route through this same call, so **an existing storage-sink config that relied on the prefix match will now discover fewer topics**. The lineage it loses is lineage that was wrong, but the change is visible, so it is called out here rather than left to be discovered. Restricting the full match to the new sink path was considered and rejected: it would preserve the defect on the path that has more users.

#
#### Use cases covered
- Confluent Cloud managed Snowflake Sink, one topic per table, **no** `snowflake.topic2table.map` → one dataset per topic, table name derived from the topic
- Topic name that is not a valid Snowflake identifier → sanitised and suffixed with `abs(javaStringHashCode(topic))` (`om-lineage-test` → `OM_LINEAGE_TEST_702890019`)
- `snowflake.topic2table.map` with renaming, including a **partial** map where some topics are mapped and others derive
- `snowflake.topic2table.map` regex keys, `$1`/`${1}` replacement under `snowflake.topic2table.map.regex.replacement`, double-quoted identifiers (including ones containing `,` and `:`), and literal keys whose dots must *not* act as wildcards
- `SnowflakeStreamingSinkConnector` and its `snowflake.compatibility.enable.*` flags for autogenerated table names and column-identifier normalisation
- Nullable Avro arrays (`UNION<null,ARRAY<…>>`) treated as leaf columns, and Avro records whose producer omits the named-record wrapper
- Snowflake service resolved from `snowflake.url.name` with `lineageInformation.dbServiceNames` left empty
- Column-level lineage for flat Avro and for nested Avro (nested record → one `VARIANT` column; `RECORD_METADATA` correctly yields no edge)
- `Flatten` SMT configured → nested leaf paths mapped to flattened columns with the configured delimiter
- Debezium/CDC and JDBC sinks unchanged

#### Unit tests
- Files added: `ingestion/tests/unit/topology/pipeline/test_kafkaconnect_snowflake_sink.py` (2,181 lines)
- **249 passing** across `test_kafkaconnect_snowflake_sink.py`, `test_kafkaconnect.py`, `test_kafkaconnect_service_discovery.py`, and `tests/unit/source/pipeline/test_kafkaconnect.py`.
- Coverage on new/changed files (`--cov=metadata.ingestion.source.pipeline.kafkaconnect`): **`sinks/snowflake.py` 94%**, **`sinks/base.py` 93%**, `constants.py` 100%, `models.py` 96%, `client.py` 70%, `metadata.py` 81% (large pre-existing file), overall 83%.
- Tests were **mutation-tested** rather than assumed. Several vacuous assertions were found and fixed this way — including a CDC regression guard that passed even with the `fully_qualified` gate removed entirely, and a "column lineage" suite in which an early `return None` inside `build_column_lineage` killed zero tests. Assertions now target `fqn.build`'s keyword arguments and real `ColumnLineage` edges.

Covered by the unit suite above plus the live end-to-end run below. Fixtures are captured verbatim from a real Confluent Cloud `GET /connectors/{name}` response and a real `DESC TABLE`, including their quirks (a leading space in `snowflake.url.name`, masked-not-omitted secrets, and defaulted properties absent from the response).

#### Playwright (UI) tests

#### Manual testing performed
Verified end-to-end against **live** Confluent Cloud and Snowflake, not only unit tests:

1. Created three Confluent Cloud Datagen sources (flat Avro, nested Avro with an `address` record, and a hyphenated topic name) plus a managed Snowflake Sink with `input.data.format: AVRO`, `SNOWPIPE_STREAMING`, `snowflake.enable.schematization: true`, and no `topic2table.map`.
2. Ingested the Kafka messaging service with Schema Registry credentials (6 topics with Avro schemas) and the Snowflake service.
3. Created a KafkaConnect pipeline service against the Connect v1 endpoint and ran metadata ingestion, with **`lineageInformation.dbServiceNames` left empty**.
4. Confirmed via `GET /api/v1/lineage/getLineage` that all three tables have `Topic → Pipeline → Table` edges with column-level detail:
   - `order_events_flat → ORDER_EVENTS_FLAT` — 6 column edges
   - `order_events_nested → ORDER_EVENTS_NESTED` — 4 column edges, `address → ADDRESS` as a single `VARIANT` column
   - `om-lineage-test → OM_LINEAGE_TEST_702890019` — 4 column edges (the hash-derivation path)
5. Confirmed no edge targets `RECORD_METADATA`, and that the Snowflake service resolved purely by matching `snowflake.url.name` against the service's `account`.

#

Not applicable.

#
### Checklist:
- [x] For JSON Schema changes: not applicable — the changed models are hand-written Pydantic in the connector package, not generated from `openmetadata-spec/`, so no regeneration or migration is owed.

<h2><a href="https://app.greptile.com/api/retrigger?id=51070538"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because Snowflake table derivation still disagrees with Java for non-BMP topic names and the Integer.MIN_VALUE hash edge case.

<h2>Findings</h2>

1. P1&nbsp;**Java hash semantics diverge** <a href="https://github.com/open-metadata/OpenMetadata/pull/31153#discussion_r4036762722">▶</a>

This PR adds connector-specific sink resolution so Kafka Connect Snowflake sinks can emit topic-to-pipeline-to-table lineage, including column mappings and Snowflake service discovery.
- Introduces a resolver registry while retaining the historical generic sink behavior as the default.
- Derives Snowflake target tables from topic configuration and explicit topic-to-table mappings.
- Adds qualified target identity and exact topic pairing to lineage resolution.
- Expands nested topic-field handling for column-level lineage.
- Adds extensive Snowflake sink tests and prunes obsolete logging-lint baseline entries.
</details>

```mermaid
flowchart LR
    C[Kafka Connect sink configuration] --> R[Sink resolver registry]
    R -->|Snowflake sink| S[Snowflake sink resolver]
    R -->|Other sink| D[Default resolver]
    S --> T[Resolve subscribed topics]
    S --> M[Derive topic-to-table mapping]
    T --> L[Kafka topic entity]
    M --> F[Qualified Snowflake table lookup]
    L --> P[Kafka Connect pipeline]
    P --> F
    S --> X[Column mappings]
    X --> F
```
</details>

<sub>Reviews (3) · Last reviewed commit: ["chore(kafkaconnect): prune the G004 base..."](https://github.com/open-metadata/openmetadata/commit/2c3bd45d268db29e9a05babfe2ee4e237a8cac5d)</sub>


ingestion/src/metadata/ingestion/source/pipeline/kafkaconnect/sinks/snowflake.py
```diff
@@ -0,0 +1,283 @@
...
+            return datasets
+
+        database = self._first_configured(config, ConnectorConfigKeys.SNOWFLAKE_DATABASE_KEYS)
+        schema = self._first_configured(config, ConnectorConfigKeys.SNOWFLAKE_SCHEMA_KEYS)
+        self._warn_on_partial_qualification(config, database, schema)
+        mapping = self._topic2table_map(config)
+
+        return [
+            KafkaConnectDatasetDetails(
+                table=mapping.get(topic) or snowflake_table_name(topic),
```
IceS2: Can you double check the following?

The two branches of this expression fold case differently: snowflake_table_name() uppercases its result, while mapping.get(topic) returns the configured value verbatim.

That matters because the connector creates the mapped table with an unquoted identifier, so Snowflake stores it uppercased. A map written order_events:orders therefore lands in ORDERS in Snowflake and is ingested into OpenMetadata as ORDERS, but this dataset carries orders — and Priority 1 in get_dataset_entity builds an exact FQN from it, which won't match. The ES pre-search inside fqn.build may recover it, but that's the path this PR was specifically fixing so it shouldn't be load-bearing. 

test_topic2table_map_overrides_derivation uses ORDERS, so a lowercase map value is currently untested. Applying the same folding to mapped values (upper unless explicitly quoted) would make both branches agree.

ingestion/src/metadata/ingestion/source/pipeline/kafkaconnect/sinks/snowflake.py
```diff
@@ -0,0 +1,283 @@
...
+    DefaultResolver,
+    SinkDatasetResolver,
+    sink_resolver_registry,
+)
+from metadata.utils.logger import ingestion_logger
+
+logger = ingestion_logger()
+
+# Snowflake unquoted identifiers: letter or underscore first, then alphanumerics, _ or $.
+VALID_SNOWFLAKE_IDENTIFIER = re.compile(r"^[A-Za-z_][A-Za-z0-9_$]*$")
```
IceS2: This regex is a character looser than the check it's reproducing. The connector's isValidSnowflakeObjectIdentifier is ^([_a-zA-Z]{1}[_$a-zA-Z0-9]+)$ — the + means it requires two or more characters, so a single-character topic is invalid upstream and gets the sanitise-plus-hash treatment.

With * here, topic a is treated as valid and derives table A, where the real connector writes A_97.
Single-character topic names are rare enough that this is cosmetic, but since the whole point of java_string_hashcode is exact upstream parity, matching the regex too seems worth the one character.

ingestion/src/metadata/ingestion/source/pipeline/kafkaconnect/sinks/snowflake.py
```diff
@@ -0,0 +1,476 @@
...
+    if VALID_SNOWFLAKE_IDENTIFIER.match(topic):
+        return topic.upper()
+
+    sanitized = "".join(char if (char.isascii() and char.isalnum()) or char in "_$" else "_" for char in topic)
+    if not re.match(r"^[A-Za-z_]", sanitized):
+        sanitized = f"_{sanitized}"
+    return f"{sanitized.upper()}_{abs(java_string_hashcode(topic))}"
+
+
+class SnowflakeSinkResolver(SinkDatasetResolver):
```
IceS2, line 95: This resolver reimplements Snowflake’s topic2table behavior, so we need to guarantee that it produces the same destination table as the actual connector.

Could you align the implementation with the versioned Snowflake contract and add compatibility tests based on Snowflake’s official resolver tests? . The tests should verify exact matching, regex fallback, numbered/named replacements, quoting/case handling, and no-match fallback.

ingestion/src/metadata/ingestion/source/pipeline/kafkaconnect/metadata.py
```diff
@@ -159,6 +165,16 @@ def _extract_hostname(self, host_string: str) -> str:
...
+    def _strip_domain_suffix(host: str, domain_suffixes: List[str]) -> str:  # noqa: UP006
+        """
+        Drop a known domain suffix so a bare host and its fully qualified form compare equal.
+        """
+        for suffix in domain_suffixes:
+            if host.endswith(suffix):
+                return host[: -len(suffix)]
+        return host
+
     def find_database_service_by_hostname(self, service_type: str, hostname: str) -> Optional[str]:  # noqa: UP045
```
IceS2, line 178: Multiple OpenMetadata services can share the same hostname, no? 🤔

harshsoni2024 (approved):

PR #31174, opened 2026-08-07
fix(glossary): honor the global domain filter on the glossary list endpoint

### LINKED ISSUE #31173: Glossary listing does not honor the global Domain filter
### Affected module
Backend — `GlossaryResource`

### Description
On the **Govern → Glossary** page, the left-panel glossary list is **not** scoped by the global **Domain** selector (the navbar globe dropdown). With domain *Manufacturing* selected, a glossary that belongs to a different domain (e.g. *Customer Domain*) is still listed. Reported by a customer (Experian).

### Steps to reproduce
1. Create two domains, e.g. `Manufacturing` and `Customer Domain`.
2. Create a glossary and assign it to `Customer Domain`.
3. In the navbar domain dropdown, select `Manufacturing`.
4. Open **Govern → Glossary**.

**Expected:** the `Customer Domain` glossary is not listed.
**Actual:** it is still listed (open it → right-panel *Domains = Customer Domain* while the navbar shows *Manufacturing*).

### Root cause
The UI's `withDomainFilter` axios interceptor appends `?domain=<fqn>` to `GET /api/v1/glossaries`, but `GlossaryResource.list()` does not declare `@QueryParam("domain")` and builds a bare `ListFilter`, so the parameter is dropped and the list is returned unfiltered. Service/DataProduct list endpoints already honor the param by resolving it into the `ListFilter` `domainId` condition.

The only domain scoping applied today is the RBAC `DomainOnlyAccessRole` restriction (`EntityUtil.addDomainQueryParam`), which filters to the user's **assigned** domains — not the navbar selection — so admins/stewards see no filtering at all.

### Environment
Reproduced on a local `2.0.0-rc1` stack; the gap exists on `main` as well.

sonika-shah:
### Describe your changes:

Fixes #31173

`GlossaryResource.list()` did not honor the global **Domain** filter. The UI's `withDomainFilter` interceptor appends `?domain=<fqn>` to `GET /api/v1/glossaries`, but the endpoint declared no `@QueryParam("domain")` and built a bare `ListFilter`, so the parameter was silently dropped and the Govern → Glossary left panel showed glossaries from every domain regardless of the selected domain.

This change adds the `domain` query parameter and resolves it into the `ListFilter` `domainId` condition — the same pattern already used by `DataProductResource` and the service resources (`ServiceEntityResource`). The RBAC `DomainOnlyAccessRole` filter (`EntityUtil.addDomainQueryParam`, applied inside `listInternal`) still runs afterwards and overwrites `domainId` for domain-restricted users, so the navbar selection can never widen a restricted user's scope.

#
### Type of change:
- [x] Bug fix

#
### Frontend Preview (Screenshots)

N/A — backend fix. The domain selector UI is unchanged.

#
### Checklist:
- [x] My PR includes unit/integration tests for the fix.

#
### Tests

**Backend integration test** — `GlossaryResourceIT.list_filterByDomain_scopesResults`: a glossary assigned to domain A appears when listing `?domain=A` and is absent when listing `?domain=B`. Fails without the fix.

**Playwright E2E** — `GlossaryDomainFilter.spec.ts`: the admin navbar-selector case plus three `DomainOnlyAccessRole` cases (single-domain isolation for two users, and a multi-domain user where the navbar must narrow within the assigned domains).

#
### Manual testing

Reproduced and verified on a local `2.0.0-rc1` Docker stack:

- **Before:** `GET /glossaries?domain=<A>` and `?domain=<B>` returned the identical set (the param was ignored) — a "Customer Domain" glossary appeared while "Manufacturing" was selected, matching the customer report.
- `GET /databaseServices?domain=<B>` filtered correctly, confirming the gap was specific to the glossary endpoint.
- In-browser (current UI) the glossary stayed visible in the left panel after switching the navbar domain; the two `DomainOnlyAccessRole` isolation cases passed (RBAC filtering works).

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/GlossaryResourceIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/GlossaryResourceIT.java
@@ -182,6 +182,36 @@ public class GlossaryResourceIT extends BaseEntityIT<Glossary, CreateGlossary> {
   // ===================================================================
 
   @Test
+  void list_filterByDomain_scopesResults(TestNamespace ns) {
+    // Reported bug: GlossaryResource.list ignored the global `?domain=` filter, so a
+    // glossary belonging to one domain still showed up when a different domain was
+    // selected. The listing must be scoped to the requested domain.
+    Glossary glossary =
+        createEntity(
+            new CreateGlossary()
+                .withName(ns.prefix("domainScopedGlossary"))
+                .withDescription("Glossary assigned to a single domain")
+                .withDomains(List.of(testDomain().getFullyQualifiedName())));
+
+    ListParams sameDomain =
+        new ListParams().setLimit(1000).setDomain(testDomain().getFullyQualifiedName());
+    assertTrue(
+        listContainsGlossary(sameDomain, glossary.getId()),
+        "Glossary must appear when filtering by its own domain");
+
+    ListParams otherDomain =
+        new ListParams().setLimit(1000).setDomain(testSubDomain().getFullyQualifiedName());
+    assertFalse(
+        listContainsGlossary(otherDomain, glossary.getId()),
+        "Glossary must not appear when filtering by a different domain");
+  }
+
+  private boolean listContainsGlossary(ListParams params, UUID glossaryId) {
+    return listEntities(params).getData().stream()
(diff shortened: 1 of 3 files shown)


sonika-shah:
Superseded by #31180, which carries the same fix. Closing this one to consolidate on that PR.

PR #31176, opened 2026-08-07
fix(playwright): retry swallowed antd menu clicks in alert helpers

ShaileshParmar11:
## Describe your changes

Fixes the `Flow/ObservabilityAlerts.spec.ts:281` **"Table alert"** failure — the only failure in an 815-test 2.0 nightly shard ([run 31157132424](https://github.com/open-metadata/OpenMetadata/actions/runs/31157132424), `868a9786`).

### Root cause

Both attempts of that test failed, in two different places, with the same shape: **a click on an antd menu item is dispatched successfully but React's handler never runs**, so the app never changes state and the one-shot wait that follows spends the entire test budget on something that was never going to happen.

The antd `Dropdown`/`Menu` opens with a `slide-up` motion. Playwright's `stable` actionability check clears once the bounding box holds for two animation frames, which the tail of the cubic-bezier ease satisfies while the transform is still running — so `mousedown` and `mouseup` land on different nodes and the browser never synthesises a `click`. Critically, `locator.click()` retries *actionability* but never the *effect*, so a lost click is never redone.

| | Attempt 1 | Attempt 2 (retry) |
|---|---|---|
| Helper | `inputBasicAlertInformation` | `visitObservabilityAlertPage` |
| Lost click | source dropdown option | sidebar → Alerts |
| Symptom | `source-select` never rendered → `toHaveText` burned 15s | never navigated → `beforeEach` hung 57s, hit the 60s timeout |

The DOM snapshot at failure confirms it: the **"Add Source" button is still rendered**, the menu is **still open** with all 8 options, and Filters/Trigger/Destination are all still `disabled` — `resources` was never set.

**Timing evidence.** Seven invocations of `inputBasicAlertInformation` in that one run, measured menu-open → click-complete:

| Test | open + click | Result |
|---|---:|---|
| **Table alert** | **149 + 192 = 341ms** | ❌ |
| Data Contract Name filter | 94 + 312 = 406ms | ✅ |
| Alert operations (also `table-option`) | 253 + 212 = 465ms | ✅ |
| Test Suite alert | 58 + 445 = 503ms | ✅ |
| Pipeline Alert | 105 + 481 = 586ms | ✅ |
| Test case alert | 152 + 529 = 681ms | ✅ |
| Ingestion Pipeline alert | 331 + 442 = 773ms | ✅ |

The single failure is the fastest sequence in the run, 65ms below the next fastest — same helper, same page, same worker.

### The fix

Retry the *selection* and the *navigation* themselves, in `expect(...).toPass()` loops, instead of hoping one click lands. Two details:

- The `isVisible()` probes are **branch conditions inside the retry loop**, not assertions — they don't reintroduce the non-waiting-assertion anti-pattern. The assertions are `toBeVisible` / `waitForURL`.
- The dropdown retry gates on the **option** rather than on `drop-down-menu`, because that testid is also used by `SearchDropdown.tsx`; probing it unscoped would be a strict-mode hazard.
- In `visitObservabilityAlertPage` the `waitForResponse` promise stays armed **once, outside** the loop (re-arming per attempt would miss a response that already landed), with a timeout that outlives the retry budget.

## Type of change

## Tests

Validated against a local 2.0 server on `:8585` with a fault-injection harness that deterministically reproduces the swallowed click (a document-capture listener that eats the first click on a given selector). Each pair is a controlled experiment — the pre-fix sequence is inlined and must keep failing, proving the injected fault is still active and still lethal:

| Stage | Source dropdown | Sidebar nav |
|---|---|---|
| Repro reproduces the CI error | ✅ identical (`toHaveText` → `element(s) not found`, 15s) | ✅ identical |
| Pre-fix sequence, fault injected | ❌ fails 18.3s | ❌ fails 16.7s |
| **Fixed helper, same fault** | ✅ **passes 8.2s** | ✅ **passes 13.1s** |

Regression, no fault injected — `ObservabilityAlerts.spec.ts` + `NotificationAlerts.spec.ts` (all six callers of `inputBasicAlertInformation`):

```
16 passed, 0 failed, 0 flaky   (222.7s)
  ✓ Table alert                 13380ms
  ✓ Pipeline Alert              27558ms
  ✓ Ingestion Pipeline alert    15696ms
  ✓ Test case alert             17870ms
  ✓ Test Suite alert            14436ms
  ✓ Single Filter Alert         36017ms
  ✓ Multiple Filters Alert      24632ms
  ... (+9)
```

`tsc`, `eslint`, and `prettier` clean on both touched files.

The repro harness itself is deliberately **not** included — it lives outside `testDir`, so committed it would be documentation that no gate enforces.

## Note

This test is not new-flaky: it shows as flaky in 2 of 3 sampled 2.0 runs and 1/14 in the nightly baseline, single fingerprint — so this one fix should retire it.


Rohit0301 (dismissed):

ShaileshParmar11:
### Follow-up: the first CI run never exercised this change

@ShaileshParmar11 spotted that [run 31170086001](https://github.com/open-metadata/OpenMetadata/actions/runs/31170086001?pr=31176) went green without running either alert spec. That is real, and it is a selector gap rather than a fluke.

`select_playwright_tests.py` classifies `playwright/utils/**` as `sharedInfrastructure`, which adds the fixed **smoke + canary** sets (the 21 selectors in that run). It then walks `mappings` for impact-based coverage. The mapping that owns the alert specs lists app sources plus exactly one playwright helper:

```
openmetadata-ui/src/main/resources/ui/playwright/utils/incidentManager.ts
```

`alert.ts`, `observabilityAlert.ts` and `notificationAlert.ts` were missing, so the two changed files matched **zero** mappings. And because `sharedInfrastructure` already marks a path as "mapped", the `unmappedChange` canary fallback did not fire either. The run's own selection artifact confirms it:

```json
{ "sharedInfrastructureChanged": true,
  "unmappedFiles": [],
  "directChangedSpecs": [],
  "changedFiles": ["...playwright/utils/alert.ts",
                   "...playwright/utils/observabilityAlert.ts"] }
```

Added in `096575a`: the three helpers now sit alongside the existing `incidentManager.ts` entry. Verified by running the selector locally over this PR's changed files:

```
mode: targeted | selected: 21 -> 53
  + playwright/e2e/Flow/ObservabilityAlerts.spec.ts
  + playwright/e2e/Flow/NotificationAlerts.spec.ts
  + playwright/e2e/Features/DataQuality/**            (26 specs)
  + playwright/e2e/Features/IncidentManager*.spec.ts
  + playwright/e2e/Pages/TestSuite*.spec.ts
```

Two notes for reviewers:

- **Cost.** That mapping bundles DataQuality + IncidentManager + alerts because they share backend surface, so any future PR touching these helpers now pulls ~32 extra specs. That is the existing mapping's design — this change only adds sources to it — but it is a deliberate trade worth naming.
- **This PR was never going to merge unvalidated.** `full_event = args.event_name in {"merge_group", "schedule"}`, so the merge queue runs the complete suite regardless. The gap was that the *PR check itself* gave no signal, and that every future PR touching these helpers would have inherited the same blind spot.

This commit is self-validating: it changes `.github/playwright/**`, so this run re-selects using the new map and should now show both alert specs executing.

Rohit0301 (approved):

PR #31180, opened 2026-08-07
fix: honor the global domain filter on glossary, metric, classification and tag list endpoints

### LINKED ISSUE #31173: Glossary listing does not honor the global Domain filter

sonika-shah:
### Describe your changes:

Fixes #31173

Several REST `list()` endpoints did not honor the global **Domain** filter. The UI's `withDomainFilter` interceptor appends `?domain=<fqn>` to non-search `GET`s, but an endpoint only applies it if its `list()` declares `@QueryParam("domain")` and maps it into the `ListFilter` `domainId` condition. Endpoints that didn't silently dropped the param and returned entities from **every** domain regardless of the domain selected in the navbar.

Reported for the **Glossary** left panel (by a customer). Verified the same gap on a running 2.0 instance for other domain-scoped catalog entities:

| Endpoint | Honors `?domain=` before | after |
|---|---|---|
| `/glossaries` | ❌ | ✅ |
| `/metrics` | ❌ | ✅ |
| `/classifications` | ❌ | ✅ |
| `/tags` | ❌ | ✅ |
| `/dataProducts` | ✅ (unchanged) | ✅ |

This adds the `domain` query parameter to `GlossaryResource`, `MetricResource`, `ClassificationResource` and `TagResource`, resolving it into the `ListFilter` `domainId` condition — the same pattern already used by `DataProductResource#list` and the service resources. The condition flows through the entity-agnostic `ListFilter.getDomainCondition` (matching `entity_relationship` rows with `relation=10`), so it scopes each list to the selected domain. The `DomainOnlyAccessRole` filter applied inside `listInternal` still runs afterwards, so the navbar selection can never widen a restricted user's scope.

#
### Type of change:
- [x] Bug fix

#
### Tests

**Backend integration tests** — `test_listWithDomainFilter` added to `GlossaryResourceIT`, `MetricResourceIT`, `ClassificationResourceIT` and `TagResourceIT`: entities assigned to domain A are returned for `?domain=A` and absent for `?domain=B`, and vice-versa. Each fails without the fix.

**Playwright E2E** (`playwright/e2e/Pages/`) — `GlossaryDomainFilter.spec.ts`, `MetricDomainFilter.spec.ts`, `ClassificationDomainFilter.spec.ts`: selecting a domain in the navbar scopes the Glossary / Metrics / Classifications list pages. The glossary spec additionally covers the `DomainOnlyAccessRole` paths (single-domain isolation for two users, and a multi-domain user narrowing within assigned domains) so the global filter and the RBAC filter are told apart.

#
### Checklist:
- [x] My PR includes integration and E2E tests for the fix.


openmetadata-service/src/main/java/org/openmetadata/service/resources/metrics/MetricResource.java
```diff
@@ -142,6 +150,11 @@ public ResultList<Metric> list(
           @DefaultValue("non-deleted")
           Include include) {
     ListFilter filter = new ListFilter(include);
+    if (!nullOrEmpty(domain)) {
+      EntityReference domainReference =
```
harshach, line 154: do we need this call agian? why are we not passing FQN for filtering?

sonika-shah:
Updated approach - https://github.com/open-metadata/OpenMetadata/pull/31538

PR #31186, opened 2026-08-07
fix(ui-core-components): make Dialog.Content scroll instead of clipping overflow

### LINKED ISSUE #32084: fix(ui-core-components): make Dialog.Content scroll instead of clipping overflow
## Problem
Dialog content is being clipped when it overflows instead of scrolling, making it impossible for users to view or interact with content that extends beyond the viewport.

## Solution
Update the Dialog.Content component in ui-core-components to enable scrolling when content overflows rather than clipping it.

## Related PRs
- OpenMetadata: https://github.com/open-metadata/OpenMetadata/pull/31186
- OpenMetadata-Collate: https://github.com/open-metadata/openmetadata-collate/pull/6157

## Impact
This fix ensures users can access all dialog content regardless of viewport size or content length.

shrabantipaul-collate:
## Problem

\`Dialog\`'s content pane has no height bound or overflow handling of its own. The wrapping \`rounded-2xl overflow-hidden\` div silently clips anything that doesn't fit inside the modal's available height — including, in one real case, the footer's Save/Cancel buttons going out of view entirely, with no scrollbar or any other indication content was missing.

This isn't a hypothetical: **at least 13 separate consumers** across \`openmetadata-ui\` and \`collate-ui\` already hand-roll \`className="tw:max-h-[NNvh] tw:overflow-y-auto"\` on \`Dialog.Content\` to work around exactly this (inconsistently, at 60vh/68vh/70vh depending on who wrote it). That's evidence the fix belongs in the shared component, not in every consumer that happens to render enough content to hit it.

## Fix

One line: \`DialogContent\` now defaults to \`max-h-[60vh] overflow-y-auto\`, matching the value/pattern already proven across those 13 call sites. \`Dialog.Footer\` sits below it, outside the scrolling pane, so it's unaffected either way. A consumer can still override the cap via \`className\` for the rare case 60vh isn't right for them.

(An earlier version of this PR also restructured \`DialogBase\`'s wrapper, \`DialogHeader\`, and \`DialogFooter\` into a flex column so \`Content\`'s height would self-compute relative to \`Header\`/\`Footer\` instead of using a fixed value. That solved a problem nobody reported — inconsistent vh values across the 13 existing workarounds — at the cost of real complexity across four places instead of one, for no proven benefit over the pattern already working in production. Dropped in favor of this minimal version.)

## Follow-up fix (from review)

A bot review correctly caught a real regression risk: \`cx\` here is \`tailwind-merge\`, which treats \`overflow-*\` (shorthand) and \`overflow-y-*\` (axis-specific) as **separate conflict groups** — so a consumer's plain \`overflow-hidden\` does not strip the new \`overflow-y-auto\` default. Both classes survive in the merged string, and the winner is decided by generated-CSS order, not JSX order. Two real consumers deliberately opt out of \`Dialog.Content\` scrolling this way (each manages its own internal scroll region):

- \`AssetSelectionModal.tsx\`
- \`AiFormModal.tsx\` (hint-column layout — has its own comment explaining why it clips instead of scrolls)

Swapped both to \`overflow-x-hidden overflow-y-hidden\`, which *is* in the same conflict group as the new default and correctly overrides it. Swept the rest of the codebase for any other \`Dialog.Content\` + shorthand \`overflow-hidden\` combination — these two were the only ones.

## Consumer cleanup pass

With the default now shipping from the core component, removed all redundant hardcoded height/overflow workarounds from the remaining consumers:

| File | Change |
|---|---|
| \`QuickLinkFormModal.tsx\` | Removed \`tw:max-h-[60vh] tw:overflow-y-auto\` — now inherited from core |
| \`UploadDocumentModal.component.tsx\` | Removed \`tw:max-h-[60vh] tw:overflow-y-auto\` — now inherited from core |
| \`ODCSImportModal.component.tsx\` | Removed \`tw:max-h-[70vh]\` — now inherited from core |
| \`AssetSelectionModal.tsx\` | Removed \`tw:max-h-[70vh]\` — kept \`tw:overflow-x-hidden tw:overflow-y-hidden\` (internal scroll region) |
| \`AiFormModal.tsx\` | Removed both calc-based \`max-h\` values — hint branch keeps layout overrides; no-hint branch uses core defaults |
| \`AddToBundleSuiteModal.component.tsx\` | Retained \`tw:min-h-45\` (minimum height for Select dropdown); \`min-h\` and \`max-h\` are independent constraints and do not conflict |

**Intentionally excluded:** \`BulkImportVersionSummary\` — uses a virtualized data grid (\`LazyDataGrid\`) that requires an explicit fixed-height container (\`style={{ height: '60vh', maxHeight: 700 }}\`) on its inner wrapper. The grid measures row virtualization from this anchor; removing it would collapse the grid to zero height or break virtualization entirely.

## Verification

No unit tests exist for this component (Storybook-only). TypeScript/JSX syntax verified (\`tsc --noEmit\`, no errors) on all changed files. Could not run a full \`vite build\` in this environment (fresh clone, no \`node_modules\` installed) — worth a quick visual check in Storybook before merge, particularly the \`AiFormModal\` hint-column layout and \`AssetSelectionModal\`.

The PR adds a shared 60vh scrolling boundary to Dialog.Content and removes redundant consumer-level height and overflow classes.
- Long dialog content now scrolls inside the shared content pane while the footer remains outside it.
- Existing modal consumers are migrated to the shared default.
- The AiFormModal hint-column override is simplified, but its previously reported nested-scroll regression remains.
</details>

The PR is not yet safe to merge because the previously reported AiFormModal hint-column nested-scrolling regression remains.

The hint-column branch does not override Dialog.Content's outer overflow-y-auto, while HookForm panel mode retains independently scrolling form and documentation panes, so long content can still produce nested scrolling and move the two-column layout together.

**Files Needing Attention:** openmetadata-ui/src/main/resources/ui/src/components/common/atoms/drawer/AiFormModal.tsx
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui-core-components/src/main/resources/ui/src/components/application/modals/modal.tsx | Adds max-h-[60vh] and overflow-y-auto as the shared Dialog.Content defaults. |
| openmetadata-ui/src/main/resources/ui/src/components/common/atoms/drawer/AiFormModal.tsx | Removes the hint-column overflow override, leaving the previously reported outer scroller active around independently scrolling columns. |
| openmetadata-ui/src/main/resources/ui/src/components/DataAssets/AssetsSelectionModal/AssetSelectionModal.tsx | Replaces the modal-specific height and clipping behavior with the shared Dialog.Content scrolling default. |
| openmetadata-ui/src/main/resources/ui/src/components/ContextCenter/UploadDocumentModal/UploadDocumentModal.component.tsx | Removes redundant local max-height and overflow classes in favor of the shared default. |
| openmetadata-ui/src/main/resources/ui/src/components/DataContract/ODCSImportModal/ODCSImportModal.component.tsx | Replaces its 70vh content cap with the shared 60vh scrolling default. |
| openmetadata-ui/src/main/resources/ui/src/components/KnowledgeCenter/QuickLinkFormModal/QuickLinkFormModal.tsx | Removes redundant local content scrolling classes in favor of the shared default. |

</details>

<sub>Reviews (11): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/dialog-cont..."](https://github.com/open-metadata/openmetadata/commit/31c058ff58ec1a45aaef1e3f0b065698f8a9de65) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51159994)</sub>


openmetadata-ui-core-components/src/main/resources/ui/src/components/application/modals/modal.tsx
```diff
@@ -79,10 +79,16 @@ interface DialogContentProps {
...
+// instead of being silently clipped by the panel's overflow-hidden rounding
+// (Dialog.Footer sits below this, outside it, so it's unaffected either
+// way). Matches the value/pattern already duplicated across ~13 call sites
+// in openmetadata-ui/collate-ui that worked around this by hand; a consumer
+// can still override via className for a case that genuinely needs more.
 const DialogContent = ({ children, className }: DialogContentProps) => (
   <div
     className={cx(
-      'tw:flex tw:flex-col tw:justify-start tw:gap-4 tw:px-4 tw:pt-5 tw:sm:px-6',
+      'tw:flex tw:max-h-[60vh] tw:flex-col tw:justify-start tw:gap-4 tw:overflow-y-auto tw:px-4 tw:pt-5 tw:sm:px-6',
```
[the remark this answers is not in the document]
shrabantipaul-collate: Confirmed this is correct — the two real consumers using the shorthand `overflow-hidden` on `Dialog.Content` (`AssetSelectionModal.tsx`, `AiFormModal.tsx`'s hint-column layout) both deliberately opt out of Dialog.Content scrolling and would have hit this. Fixed in 092bebad by switching both to `overflow-x-hidden overflow-y-hidden`, which is in the same tailwind-merge conflict group as the new `overflow-y-auto` default and correctly overrides it. Swept the rest of the codebase for the same pattern — these two were the only ones. Thanks for the catch.

openmetadata-ui-core-components/src/main/resources/ui/src/components/application/modals/modal.tsx
```diff
@@ -145,21 +155,27 @@ const DialogBase = ({
...
+      // flex + max-h-[85dvh] here (not a % height — Modal above only ever
+      // sets max-height, never a definite height, so a percentage would have
+      // no definite ancestor to resolve against and silently no-op) bounds
+      // the panel so DialogContent's flex-1/min-h-0/overflow-y-auto below
+      // has an actual constraint to shrink against instead of growing
+      // unbounded and being clipped by the rounded-2xl overflow-hidden pane.
       <div
         className={cx(
-          'tw:relative tw:w-full tw:rounded-2xl tw:bg-primary tw:shadow-xl',
+          'tw:relative tw:flex tw:max-h-[85dvh] tw:w-full tw:flex-col tw:rounded-2xl tw:bg-primary tw:shadow-xl',
```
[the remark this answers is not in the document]
shrabantipaul-collate: The 85dvh-on-the-wrapper approach this flagged was dropped entirely in 877839c9 in favor of a much smaller fix (just `max-h-[60vh] overflow-y-auto` on `Dialog.Content` itself, matching the pattern already used by ~13 existing consumers) — so this specific code no longer exists in the PR. Should be moot now, but flagging in case the bot re-runs against the latest diff.

openmetadata-ui/src/main/resources/ui/src/components/common/atoms/drawer/AiFormModal.tsx
```diff
@@ -268,11 +268,16 @@ export const AiFormModal: FC<AiFormModalProps> = ({
...
+                Same reasoning for overflow-x-hidden/overflow-y-hidden instead
+                of the overflow-hidden shorthand: Dialog.Content now defaults
+                to overflow-y-auto, and tailwind-merge treats overflow-* and
+                overflow-y-* as separate groups, so the shorthand wouldn't
+                reliably override that default. */}
             <Dialog.Content
               className={
                 hasHintColumn
-                  ? 'tw:max-h-[calc(88vh-141px)] tw:flex-row tw:gap-0 tw:overflow-hidden tw:p-0 tw:sm:p-0'
+                  ? 'tw:max-h-[calc(88vh-141px)] tw:flex-row tw:gap-0 tw:overflow-x-hidden tw:overflow-y-hidden tw:p-0 tw:sm:p-0'
```
ShaileshParmar11: Why we are making this change? its one and the same no?

ShaileshParmar11 (dismissed):

openmetadata-ui/src/main/resources/ui/src/components/common/atoms/drawer/AiFormModal.tsx
shrabantipaul-collate: @ShaileshParmar11 Makes sense. So can we remove the max-height property? We already have it defined in core components? Or we need it in this case?

ShaileshParmar11: lets not do anything in AIFormModal Ill validate if something needs to be done
It has combination of 2 thing,
1. Form
2. Dynamic Document

openmetadata-ui/src/main/resources/ui/src/components/common/atoms/drawer/AiFormModal.tsx
```diff
@@ -268,12 +268,17 @@ export const AiFormModal: FC<AiFormModalProps> = ({
...
             <Dialog.Content
               className={
                 hasHintColumn
-                  ? 'tw:max-h-[calc(88vh-141px)] tw:flex-row tw:gap-0 tw:overflow-hidden tw:p-0 tw:sm:p-0'
-                  : 'tw:max-h-[calc(100vh-260px)] tw:overflow-y-auto'
+                  ? 'tw:flex-row tw:gap-0 tw:overflow-x-hidden tw:overflow-y-hidden tw:p-0 tw:sm:p-0'
+                  : undefined
               }>
               {children}
             </Dialog.Content>
```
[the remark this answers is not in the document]
shrabantipaul-collate: We are removing those so they inherit the max-h-[60vh] from Dialog.Content in the core component. The “shrink to 60vh” is the intended behaviour

openmetadata-ui/src/main/resources/ui/src/components/common/atoms/drawer/AiFormModal.tsx
shrabantipaul-collate: Validated these scenarios. Will work fine with 60vh max height also

openmetadata-ui/src/main/resources/ui/src/components/common/atoms/drawer/AiFormModal.tsx
```diff
@@ -268,12 +268,17 @@ export const AiFormModal: FC<AiFormModalProps> = ({
...
+                of the overflow-hidden shorthand: Dialog.Content now defaults
+                to overflow-y-auto, and tailwind-merge treats overflow-* and
+                overflow-y-* as separate groups, so the shorthand wouldn't
+                reliably override that default. */}
             <Dialog.Content
               className={
                 hasHintColumn
-                  ? 'tw:max-h-[calc(88vh-141px)] tw:flex-row tw:gap-0 tw:overflow-hidden tw:p-0 tw:sm:p-0'
-                  : 'tw:max-h-[calc(100vh-260px)] tw:overflow-y-auto'
+                  ? 'tw:flex-row tw:gap-0 tw:overflow-x-hidden tw:overflow-y-hidden tw:p-0 tw:sm:p-0'
```
chirag-madlani: Why we need overflow hidden here

shrabantipaul-collate: Fixed

chirag-madlani (approved):

PR #31204, opened 2026-08-07
fix: resolve nested column paths in FieldPathUtils dot-form field paths

### LINKED ISSUE #31325: Approving a description suggestion on a deeply nested column writes to the wrong column
Approving a Description suggestion for a column nested two or more levels deep writes the text onto an ancestor column instead of the intended leaf, overwriting the ancestor's existing description. The request returns 200 with no warning.

Shreyansh100704:
### Describe your changes:

Approving a description suggestion for a column nested two or more levels deep wrote the suggested text onto an **ancestor** column instead of the intended leaf. The request returned HTTP 200 and logged no warning, so the corruption was silent and destructive: the ancestor's own description was overwritten and the leaf never received its text.

The dot branch of `FieldPathUtils.parseFieldPath` split the field path into at most three segments:
  
  ```java
  String[] parts = fieldPath.split("\\.", 3);
```

For columns.profile.personal.full_name.description that yields fieldName=profile and property=personal.full_name.description. navigateAndSetDescription never reads property, so the remainder of the path was discarded and the write landed on the top-level profile column. The three-way split also broke quoted paths apart, so columns."a.b".description did not resolve at all.

I parse the property as the final segment and treat everything between the container and it as the field name, using the existing quote-aware FullyQualifiedName.split / unquoteName rather than hand-rolling a parser (ChangeSummarizer already uses the same helpers for the same field-name shape). No traversal code changed: setDescriptionInList already recurses through children for dotted names — that is how the :: form has always resolved nested columns. The fix simply feeds that existing recursion the full dotted name it was written to consume.

The FQN parser uses BailErrorStrategy, so malformed input now throws instead of silently producing garbage components. The catch falls through to the existing return null, which routes malformed paths to updateFieldDescription's log-and-abort at lines 63-66 — converting silent corruption into a loud no-op.

#
### Type of change:

- [x] Bug fix

#
### High-level design:

N/A — small change. 

#
#### Use cases covered

- Approving a description suggestion on a column nested 2, 3, or 4 levels deep updates that column and leaves its ancestors and siblings untouched
- Approving a suggestion whose fieldPath uses the quoted form (columns."a.b".description) resolves correctly
- A column whose name literally contains a dot (risk.score) still wins over nested traversal, via exact-match-first
- A malformed or unresolvable path aborts with a warning instead of writing to an ancestor
- Flat, array-index, and all :: paths behave exactly as before

#### Unit tests

- Files added/updated:
  - openmetadata-service/src/test/java/org/openmetadata/service/util/FieldPathUtilsTest.java: 7 parse cases (depth 2/3 unquoted, depth 3 quoted, nested tags, three-segment property-last, literal dotted name, malformed empty segment) and 3 behavioural cases asserting updateFieldDescription writes to the leaf, leaves ancestors untouched, and returns false without patching for unresolvable paths.
  - openmetadata-service/src/test/java/org/openmetadata/service/tasks/TaskWorkflowHandlerTest.java: approving a depth-3 Description suggestion updates only that column.

- [x] Not applicable (no API surface changed — same endpoints, same request/response shapes).

#### Playwright (UI) tests

#### Manual testing performed

#

Before the fix: Profile column has desc of phone number column
Screenshot 2026-08-07 at 8 10 37 PM

After fix: Profile column has correct description
Screenshot 2026-08-07 at 8 36 55 PM

#
### Checklist:


pmbrull (approved):

Shreyansh100704:
Cherry-picked to 2.0

PR #31232, opened 2026-08-08
Fixes #31231: Consolidate migrations into a 2.0 baseline and remove Flyway

### LINKED ISSUE #31231: Consolidate pre-2.0 migrations into a 2.0 baseline and remove Flyway
### Problem

A fresh install replays the entire migration history before it can serve a request: 16 legacy Flyway scripts, 85 native version directories and 148 Java migration classes. The server also re-parses that whole tree on every boot to decide whether anything is pending.

Separately, there is no upgrade-order enforcement at all. A 1.13 (or older) database can be migrated directly by a 2.1 build, which runs 2.x DDL against a schema those migrations never anticipated.

### Proposal

1. Freeze everything below 2.0.0 into a generated, per-dialect baseline that an empty database installs once; `native/2.0.0+` continue to apply incrementally on top.
2. Require a database to be on 2.0.x before this release will migrate it, and make the check immune to `--force`.
3. Make the migration history self-describing — record what kind of step each row is and whether it completed — so an operator can see where a run stopped.
4. Delete the pre-2.0 SQL and Java that the baseline replaces, and drop the Flyway dependency (retained only for SQL statement splitting).

### Notes

The baseline carries schema only. Migrations write rows to repair databases that already exist; a new database gets its data from the application's own boot-time seeding, so freezing a copy of those rows would pin them to whatever they looked like the day the baseline was cut.

Upgrades from 1.13 or older become a two-step path: upgrade to the latest 2.0.x, run its migrations, then upgrade.

harshach:
### Describe your changes:

Fixes #31231

I consolidated every migration below 2.0.0 into a generated baseline, added an upgrade gate that
refuses databases older than 2.0, made the migration history describe its own steps, and deleted the
pre-2.0 SQL and Java plus the Flyway dependency — because a fresh install was replaying 16 Flyway
scripts, 85 native version directories and 148 Java migration classes (and re-parsing all of it on
every boot), while nothing prevented a 1.13 database from being migrated straight to 2.1.

#
### Type of change:

Upgrading from 1.13 or older now requires going through the latest 2.0.x first.

#
### High-level design:

**Baseline.** `bootstrap/sql/migrations/baseline/{mysql,postgres}/schema.sql` is the schema produced
by running Flyway `v000`–`v015` plus native `1.1.0`–`1.13.4` on an empty database. An empty database
installs it once and records a single `2.0.0-baseline` step; `native/2.0.0+` then apply
incrementally. `2.0.0` and `2.1.0` stay live directories, so 2.0.x patches cherry-picked into `2.0.0`
still reach both upgraded and baseline-installed databases through the existing release-train
reprocessing.

**Schema only, deliberately.** Migrations write rows to repair databases that already exist; a new
database has nothing to repair, and the application seeds its own data at boot from `json/data/**`.
Freezing a copy of those rows would pin them to whatever they looked like the day the baseline was
cut, with no migration left to update them. The one exception is the `task_sequence` counter row,
which is schema bootstrap — the live 2.0.0 migration initializes its replacement the same way. This
surfaced a latent gap: the system Data Insight charts were built only by the deleted `v150`/`v170`
Java, and `DataInsightSystemChartResource.initialize()` globs a `json/data/dataInsight/custom/`
directory that never existed. They are now seed JSON in that directory, so the application owns them
for fresh installs and existing clusters keep theirs (`initializeEntity` is create-if-missing).

**Gate and floor.** The gate runs before the force branch, so `--force` — which every development
compose passes on each boot — cannot skip a release. On a baseline-managed database every version
below the floor is filtered out of the available set, so a stray pre-2.0 directory can never replay
on top of the baseline. Both are keyed off the presence of a migration configuration, which keeps
engine-level callers driving their own migration sets unaffected.

**History as steps.** `SERVER_CHANGE_LOG` gained `migrationType`
(`BASELINE`/`NATIVE`/`EXTENSION`/`FLYWAY`) and `status` (`STARTED`/`COMPLETED`/`FAILED`), added and
backfilled in place by the runner rather than by a migration — the runner writes history rows for
every version it processes, so the columns must exist before the first of those writes. Only
`COMPLETED` counts as applied, so a crash leaves a visible marker and the version stays pending.
`openmetadata-ops.sh info` prints the steps, and `repair` — which had queried a `status` column that
never existed — now works and clears unfinished rows.

**Flyway removal.** `SqlStatementSplitter` replaces the parsers (the only thing Flyway was still used
for). `MigrationValidationClient` compares only versions this release still ships, so upgraded
clusters keep their historical rows without `/v1/system/status` reporting drift. `flywayPath` is
retained as `@Deprecated` because Dropwizard rejects unknown configuration properties.

**Alternative rejected:** snapshotting statement hashes into `SERVER_MIGRATION_SQL_LOGS` so the old
directories would self-skip. That keeps 40k lines alive forever and still replays Java data
migrations.

#
#### Use cases covered
- Fresh install: empty database installs the baseline, then `2.0.0` and `2.1.0`, recording exactly those three steps
- Upgrade from 2.0.x: proceeds normally, pre-2.0 history left intact and no longer reported as drift
- Upgrade from 1.13 or older: refused with instructions, including under `--force`
- Pre-native (Flyway-era) database: refused
- Crashed mid-step: leaves a `STARTED` row, the version stays pending, `repair` clears it
- Operator inspects history: `info` shows Version / Type / Status / Installed On

#### Unit tests
- [x] Added unit tests for the new logic.
- Added: `MigrationVersionUtilTest`, `SqlStatementSplitterTest`, `BaselineFilesTest`, `BaselineWorkflowTest`
- Updated: `MigrationWorkflowTest`, `MigrationWorkflowReprocessingTest`
- Result: 162 tests green (`mvn test -pl openmetadata-service -Dtest='org.openmetadata.service.migration.**'`)
- Coverage %: not captured — the JaCoCo agent is not enabled in the default profile for this module,
  so no execution data is produced. The new classes are covered by the unit tests above plus the
  integration tests below, which exercise them against real MySQL and PostgreSQL.
- A temporary parity test compared `SqlStatementSplitter` against the Flyway parsers across all 276
  migration SQL files and required byte-identical output; it is removed with the dependency.

- [x] Added integration tests in `openmetadata-integration-tests/`.
- Added: `MigrationGateIT`, `BaselineFreshInstallIT`, and the `BaselineScratchSupport` /
  `SchemaEquivalenceDiff` / `BaselineArtifactWriter` / `BaselineDataDump` /
  `DataInsightChartSeedExporter` harness behind `scripts/generate_migration_baseline.sh`
- Result: 48 tests green on **both PostgreSQL and MySQL** (`MigrationGateIT`,
  `BaselineFreshInstallIT`, `ContinuousMigrationIT`, `SystemResourceIT`)
- A `BaselineEquivalenceIT` proved a baseline install indistinguishable from a full chain install —
  tables, columns, indexes, foreign keys, check constraints, sequences and seeded rows — on both
  dialects. It is deleted with the chain it compared against, because that proof can only be run
  while both paths exist.

#### Playwright (UI) tests

#### Manual testing performed
1. `mvn -pl openmetadata-service dependency:tree` — zero `org.flywaydb` entries
2. `make harness-check` — no dead references to any deleted path, no new warnings
3. Regenerated the baseline twice via `scripts/generate_migration_baseline.sh` and confirmed the
   MySQL and PostgreSQL runs produce byte-identical Data Insight chart seeds

#

Not applicable.

#
### Checklist:
- [x] My PR is linked to a GitHub issue via `Fixes #31231` above.
- [x] For JSON Schema changes: not applicable — no schema changes.
- [ ] I have added the tag `Backward-Incompatible-Change`.


openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/migration/BaselineDataDump.java
```diff
@@ -0,0 +1,143 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.openmetadata.it.tests.migration;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.HexFormat;
+import java.util.List;
```
harshach, line 18: why are we importing this - test

harshach, line 18: why are we importing this - test

scripts/generate_migration_baseline.sh
```diff
@@ -0,0 +1,36 @@
...
+#   mvn test -pl openmetadata-integration-tests -Dtest=BaselineEquivalenceIT -DdatabaseType=mysql
+
+set -euo pipefail
+cd "$(dirname "$0")/.."
+
+echo "==> Generating PostgreSQL baseline (bootstrap/sql/migrations/baseline/postgres/)"
+mvn test -pl openmetadata-integration-tests \
+  -Dtest=BaselineGeneratorHarness -Dbaseline.generate=true -DfailIfNoTests=false
+
+echo "==> Generating MySQL baseline (bootstrap/sql/migrations/baseline/mysql/)"
```
harshach, line 57: test

openmetadata-service/src/main/resources/json/data/dataInsight/custom/total_data_assets.json
```diff
@@ -0,0 +1,16 @@
+{
+  "chartDetails" : {
+    "excludeGroups" : [ "tag", "glossaryTerm" ],
+    "groupBy" : "entityType.keyword",
+    "includeGroups" : [ ],
+    "metrics" : [ {
+      "formula" : "count(k='id.keyword')"
+    } ],
```
manerow: The header on these files pins the export to [`b07117a`](https://github.com/open-metadata/OpenMetadata/commit/b07117a765466d3fd12c3179ac800bc734de0a5f) (8 Aug). #31759 landed on main on 27 Aug and is already merged into this branch, and it realigns exactly these 14 charts, so the export predates it.

After #31759 each metric should carry [`DATA_ASSET_FILTER`](https://github.com/open-metadata/OpenMetadata/blob/f705d54c9f0663ea7af74eb707774b3aebd232b9/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataInsightSystemChartRepository.java#L106), whose key part is `must: [{"exists": {"field": "entityType"}}]`, and `excludeGroups` should be `["tag", "glossaryTerm", "dataProduct"]`.

None of the 31 files in this directory has the `exists` clause. Six of the fourteen have no metric `filter` at all, this one included (lines 6-8), and `excludeGroups` here is missing `dataProduct` (line 3). Without `exists`, documents with no `entityType` (test case results, which the `di-data-assets-*` pattern also resolves) are counted by the summary card and dropped by the breakdown, which is #31478.

Nothing repairs it afterwards. 1.13.5 sorts below `MINIMUM_SUPPORTED_MIGRATION_VERSION`, so [`applyBaselineFloor`](https://github.com/open-metadata/OpenMetadata/blob/f705d54c9f0663ea7af74eb707774b3aebd232b9/openmetadata-service/src/main/java/org/openmetadata/service/migration/api/MigrationWorkflow.java#L249-L259) drops it on every baseline-installed database, and `initializeEntity` is create-if-missing. No 2.0.0 / 2.0.1 / 2.1.0 migration touches these charts either. Clusters that upgrade do get the fix, so fresh and upgraded databases would diverge.

Would it make sense to regenerate the seeds on top of #31759, or to call [`alignDataAssetChartScope()`](https://github.com/open-metadata/OpenMetadata/blob/f705d54c9f0663ea7af74eb707774b3aebd232b9/openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v1135/DataInsightChartMigration.java#L68) from the 2.1.0 migration?

scripts/generate_migration_baseline.sh
```diff
@@ -0,0 +1,36 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+# Regenerates the consolidated migration baseline under bootstrap/sql/migrations/baseline/
+# for BOTH dialects by chain-installing every pre-2.0 migration into scratch databases and
+# dumping the result. Requires Docker (testcontainers) and a full local Maven build.
+#
+# After regenerating: review the diff, then prove fidelity with the equivalence check:
+#   mvn test -pl openmetadata-integration-tests -Dtest=BaselineEquivalenceIT
+#   mvn test -pl openmetadata-integration-tests -Dtest=BaselineEquivalenceIT -DdatabaseType=mysql
```
manerow: Both halves of this script look like they stop working once this merges.

Lines 19-20 point at `BaselineEquivalenceIT`, which this PR deletes.

Line 26 runs `BaselineGeneratorHarness`, whose input comes from [`buildPreTwoZeroNativeRoot`](https://github.com/open-metadata/OpenMetadata/blob/f705d54c9f0663ea7af74eb707774b3aebd232b9/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/migration/BaselineScratchSupport.java#L224-L245): it copies the sub-2.0.0 directories out of `bootstrap/sql/migrations/native/`. This PR removes all of them except `1.13.4` (only `postDataMigrationSQLScript.sql` survives) and `1.13.5` (four empty files), and the Flyway scripts are gone too, so a regeneration run would dump an almost empty schema.

That leaves the baseline SQL and the 31 chart seeds as artifacts we can neither reproduce nor re-check. I think that is how the stale seeds got in (see my comment on `total_data_assets.json`): #31759 landed on main, regeneration was not possible, and nothing failed.

Either of these would close the gap:

- pin the generator to a revision that still has the chain, e.g. `git archive` [`b07117a`](https://github.com/open-metadata/OpenMetadata/commit/b07117a765466d3fd12c3179ac800bc734de0a5f) into the temp root instead of copying the working tree
- or a cheaper guard: fail the build if any directory under `native/` sorts below `MigrationVersionUtil.MINIMUM_SUPPORTED_MIGRATION_VERSION`

The guard would also catch what is still in the branch today: `native/1.13.4`, `native/1.13.5`, and the `migration/{mysql,postgres}/v1135` and `migration/utils/v1135` classes, which run on upgrades but are skipped on fresh installs.

openmetadata-service/src/main/java/org/openmetadata/service/util/OpenMetadataOperations.java
```diff
@@ -431,23 +431,27 @@ public Integer repair() {
...
+              "Clearing {} unfinished migration(s): {}", failedVersions.size(), failedVersions);
 
-          // Remove failed migrations from SERVER_CHANGE_LOG
+          // Remove them so the next migrate run treats those versions as pending again
           jdbi.useHandle(
               handle ->
                   handle
-                      .createUpdate("DELETE FROM SERVER_CHANGE_LOG WHERE status = 'FAILED'")
+                      .createUpdate(
+                          "DELETE FROM SERVER_CHANGE_LOG WHERE status IN ('FAILED', 'STARTED')")
```
manerow: I think this leaves a crashed baseline install unrecoverable.

[`execute()`](https://github.com/open-metadata/OpenMetadata/blob/f705d54c9f0663ea7af74eb707774b3aebd232b9/openmetadata-service/src/main/java/org/openmetadata/service/migration/baseline/BaselineWorkflow.java#L156-L165) writes one row, `2.0.0-baseline` with status STARTED, before any DDL runs. If it dies partway you are left with partial tables plus that row. `repair` deletes it here, and on the next `migrate` `SERVER_CHANGE_LOG` is empty, so [`resolveActionForEmptyHistory`](https://github.com/open-metadata/OpenMetadata/blob/f705d54c9f0663ea7af74eb707774b3aebd232b9/openmetadata-service/src/main/java/org/openmetadata/service/migration/baseline/BaselineWorkflow.java#L132-L141) sees `entity_relationship` already created and returns `ABORT`.

That row is the only input that selects [`RESUME`](https://github.com/open-metadata/OpenMetadata/blob/f705d54c9f0663ea7af74eb707774b3aebd232b9/openmetadata-service/src/main/java/org/openmetadata/service/migration/baseline/BaselineWorkflow.java#L149), the path that wipes and reinstalls behind the `WIPE_GUARD_TABLES` check, so `repair` turns a recoverable state into "restore from a backup or drop-create". Given the description presents `repair` as the way to clear unfinished rows, it is probably the first thing someone tries after a failed fresh install.

Could we exclude `MigrationVersionUtil.BASELINE_VERSION` from the `SELECT` on line 441 and from this `DELETE`?

openmetadata-service/src/main/java/org/openmetadata/service/migration/api/MigrationWorkflow.java
```diff
@@ -496,10 +648,25 @@ public void runMigrationWorkflows(boolean computeAllContext) {
...
+   * Schema work the runner owns, done before any migration executes: bring an older history table
+   * up to shape, and install the consolidated baseline when this is an empty database. Only the
+   * migrate path reaches this — the server's startup validation loads and checks without writing.
+   */
+  private void prepareDatabase() {
+    if (baselineWorkflow != null) {
+      historyTableUpgrader.ensureSchema();
+      baselineWorkflow.runIfRequired();
+    }
+  }
```
manerow, line 661: Minor, not blocking. When the baseline files are not found, nothing ends up creating `SERVER_CHANGE_LOG`.

`baselineWorkflow` is non-null whenever a migration configuration exists, so this branch is taken, but with no baseline files `resolveAction()` returns `DISABLED` and `runIfRequired()` does nothing, and `ensureSchema()` no-ops on its `if (historyTableExists())`. The first migration then reaches [`markStepStarted`](https://github.com/open-metadata/OpenMetadata/blob/f705d54c9f0663ea7af74eb707774b3aebd232b9/openmetadata-service/src/main/java/org/openmetadata/service/migration/api/MigrationWorkflow.java#L759-L769), guarded only by `baselineWorkflow != null`, and fails on the missing table.

Flyway `v000` used to create it, so an empty database always had it. It is reachable in the `BaselineGeneratorHarness` run too, where the derived baseline path lands in a temp dir with no `baseline/` beside it. A "baseline files not found at X and the database is empty" message would be easier to act on than `Table 'SERVER_CHANGE_LOG' doesn't exist`.

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/migration/FlowableSchemaCustomizationIT.java
```diff
@@ -0,0 +1,80 @@
...
+  private static final String EVENT_SUBSCRIPTION_TABLE = "ACT_RU_EVENT_SUBSCR";
+  private static final String ACTIVITY_ID_COLUMN = "ACTIVITY_ID_";
+
+  /** Governance workflow activity ids exceed Flowable's stock varchar(64). */
+  private static final int REQUIRED_LENGTH = 255;
+
+  @Test
+  void flowableActivityIdIsWideEnoughForGovernanceWorkflows() {
+    Optional<Integer> length = activityIdLength();
+    assertNotNull(length);
```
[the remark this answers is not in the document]
harshach: Fixed in 30c357d2b7. Removed the redundant Optional null assertion and its import; the test now relies on the meaningful presence and width assertions.

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/migration/BaselineDataDump.java
harshach: ScratchDatabase is the nested record declared by BaselineScratchSupport. BaselineDataDump uses that type for its database field and constructor parameter, so the import avoids repeating the enclosing class name and is not unused.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java
```diff
@@ -494,9 +494,19 @@ public void deploy(Workflow workflow) {
...
+            .processDefinitionKey(wf.getName())
+            .list()
+            .isEmpty();
+    boolean triggerWorkflowDeployed =
+        !repositoryService
+            .createProcessDefinitionQuery()
+            .processDefinitionKeyLike(getTriggerWorkflowId(wf.getName()) + "%")
+            .list()
+            .isEmpty();
+    return mainWorkflowDeployed && triggerWorkflowDeployed;
```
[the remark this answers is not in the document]
harshach: Fixed in 30c357d2b7. isDeployed now queries both the main and trigger definitions with the workflow fully qualified name. Added a regression test where name and FQN differ.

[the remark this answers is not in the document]
harshach: Fixed in 30c357d2b7. Both existence probes now use count greater than zero, and the regression test verifies that list is never called.

openmetadata-service/src/main/resources/json/data/dataInsight/custom/total_data_assets_live.json
```diff
@@ -0,0 +1,17 @@
+{
+  "chartDetails" : {
+    "excludeGroups" : [ "testSuite", "testCase" ],
+    "groupBy" : "entityType",
+    "includeGroups" : [ ],
```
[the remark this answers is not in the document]
harshach: Fixed in 30c357d2b7. total_data_assets_live now groups by entityType.keyword, with regression coverage for the committed seed.

openmetadata-service/src/main/resources/json/data/dataInsight/custom/total_data_assets.json
harshach: Fixed in 30c357d2b7. Refreshed all 14 data-asset chart seeds to use the current DATA_ASSET_FILTER, including the entityType existence clause and dataProduct exclusion. A test now loads every committed seed and enforces parity with the upgrade transform.

scripts/generate_migration_baseline.sh
harshach: Fixed in 30c357d2b7. The generator builds an isolated checkout of pinned revision b07117a765466d3fd12c3179ac800bc734de0a5f, which retains the complete Flyway and native chain, then overlays compatible generator helpers. Deleted equivalence and seed-export references were removed. I verified the pinned reactor and all 446 historical integration-test sources compile.

openmetadata-service/src/main/java/org/openmetadata/service/util/OpenMetadataOperations.java
harshach: Fixed in 30c357d2b7. Repair is now transactional and excludes 2.0.0-baseline from both selection and deletion while removing SQL logs only for other unfinished versions. A real H2 regression test preserves a STARTED baseline row and its log.

openmetadata-service/src/main/java/org/openmetadata/service/migration/api/MigrationWorkflow.java
harshach: Fixed in 30c357d2b7. When baseline files are absent, the workflow rechecks database state: an empty or interrupted database fails immediately with the resolved baseline directory and migrationConfiguration.baselinePath, while an existing upgraded database continues. Both paths have tests.

PR #31294, opened 2026-08-10
fix(ingestion): apply CA cert and mask password in OpenLineage Kafka connection

### LINKED ISSUE #31296: OpenLineage Kafka connection ignores CA cert and exposes SASL password unmasked
### Bug

When configuring the **OpenLineage** pipeline connector with a **Kafka** broker, two problems occur:

1. **The CA certificate is not used.** A CA cert (and optional client cert/key) entered under SSL config is never actually applied to the Kafka consumer at connect time.
2. **The SASL password is not masked.** The password field renders as a plain-text input and is returned by the API unmasked.

### Environment

- OpenMetadata `main` (checked against `2064dcec45`)

### Expected behavior

- An SSL-configured broker connects with the uploaded CA bundle honored.
- The SASL password input is masked on the UI and returned masked in the service-connection payload.

### Actual behavior

- The broker is never handed the CA bundle, and the SASL password is shown/returned in plain text.

### Suggested fix (PR in progress)

PR: open-metadata/OpenMetadata#31294
- Materialize the cert content to temp files via `SSLManager` and pass those paths to `confluent_kafka`'s `ssl.*.location` keys (it expects file paths, not content).
- Add `format: "password"` to `saslPassword` in `saslClientConfig.json` so masking is applied server-side and the field renders as a password input on the UI.

yalindogusahin:
Fixes #31296

### Describe your changes

Fixes two bugs in the **OpenLineage** pipeline connector when configuring a **Kafka** broker:

1. **The CA cert (and optional client cert/key) was never applied to the broker.** The SSL config fields hold the cert **content**, but `confluent_kafka`'s `ssl.ca.location` / `ssl.certificate.location` / `ssl.key.location` expect **file paths**, so a real CA bundle was never handed to the consumer. The connector now materializes the content to temp files via the existing `SSLManager` (the same path every other Kafka-ish connector uses) and passes the temp-file paths, registering `cleanup_temp_files` on connection close.

2. **The SASL password was not masked on the UI or in API responses.** `saslPassword` in `saslClientConfig.json` was a plain string, so it rendered as a plain-text input and was returned unmasked. Adding `format: "password"` drives the `@PasswordField` annotation (server-side masking), the `PasswordWidget` (UI), and the Python `CustomSecretStr` (read via `get_secret_value()`). `saslClientConfig.json` is referenced only by `openLineageConnection.json`, so the schema change is isolated.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Checklist before requesting a review

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added corresponding unit tests for all new or changed code (Python tests in `test_connection.py`; Java masking round-trip test in `TestEntityMasker`)
- [x] If this is a bugfix, a test that reproduces the bug has been added — the Python SSL-temp-file and SASL-secret tests fail without the fix
- [ ] UI build / i18n / styling verified (no UI code changed)


IceS2:
Hey @yalindogusahin, thanks for the contrib. 

You will need to add the .ts files after generating those since you changed some schemas if I am not mistaken, can you check that?

Cheers,

yalindogusahin:
Good catch — the generated .ts files for the new split-out `kafkaBrokerConfig.json`/`kinesisBrokerConfig.json` schemas were missing. Regenerated via `json2ts-generate-all.sh` and pushed in b85a7c54a8. `openLineageConnection.ts` itself came back unchanged (quicktype dereferences self-contained, so its output was already correct).

IceS2 (approved):

chirag-madlani (approved):

PR #31297, opened 2026-08-10
Fix failing py-cli-e2e-tests

akashverma0786:
## Describe your changes:

mssql — the "deleted table" test failed because after dropping its one table, the schema looked completely empty, and a safety check treats an empty scan as "connector might have crashed" and skips deleting anything. Fix: added a second table that's never dropped, so the schema is never empty.
dbt_redshift — a test checking two columns together was getting grouped with a different test that only checks one column, so the server rejected it. Fix: keep those two kinds of tests separate.
redshift — a routine, harmless warning (Redshift dropping its own bad constraint) started counting as a real warning after an unrelated change, breaking a test expecting zero. Fix: turned it back into a plain log message.
quicksight — not a bug. The shared test account just has more dashboards now than the test expected. Fix: updated the expected number.
tableau — not a bug either. A recent fix correctly gives some columns their real type now instead of a placeholder, but the test still expected the old placeholder everywhere. Fix: relaxed the check to match.

## Type of change:
- [x] Bug fix

----
## Summary by Gitar

- **Fixed bugs:**
    - Fixed `Status.as_string()` breaking literal_eval by making truncation notices and BaseModel failures literal-safe
- **Database ingestion fixes:**
    - Updated DBT metadata processing to scope test definitions by entity type and asserted expected test definition count
- **E2E test updates:**
    - Adjusted MSSQL, Tableau, Redshift, and Quicksight e2e tests for schema compatibility and expected counts


IceS2:
Few things to double-check:

  1. filtered / failures over 1000 items. The harness does Status.model_validate(literal_eval(...)). Those two fields
  aren't List[Any]. The leading header string fails validation instead of parsing. Same crash, different exception.
  2. List counts are off by one when truncated. Header counts as an element, so len(records) reads 1001. E2E asserts
  on those counts. Both go away if the notice lives outside the list — e.g. a sibling 'records_total_items': 1500 key (Status ignores extras on validate).
  3. [verify] debug logging in dbt_service.py / dbt/metadata.py — needs removing; the dbt_service.py:364 one runs a
  dict comprehension on every dbt ingestion regardless of log level.

ingestion/tests/unit/test_status.py
IceS2, line 44: Shouldn't this be _total_items?

IceS2, line 61: Shouldn't this be _total_items?

IceS2 (approved):

PR #31301, opened 2026-08-10
fix(ui): stop deep links 404ing on cold boot and fix incident table column sizing

ShaileshParmar11:
## Problem

Three defects found while reviewing the AI-mode Observability pages. All three are in shared UI — the last two fix Classic mode too.

### 1. Deep links and reloads land on `/404`

`AppRouter` resolves routes as `ModeRoutes ?? AuthenticatedRoutes`. A non-default app mode's routes are registered by the plugin that owns the mode, in an effect gated on `applications` loading — so for the first renders the registry is legitimately empty. Falling through to the default routes in that window mounts their `path="*"` catch-all:

```tsx
<Route element={<Navigate to={ROUTES.NOT_FOUND} />} path="*" />
```

which navigates to `/404` and destroys the requested URL before the real routes ever mount. It is a race, so it reproduces intermittently — roughly half of hard loads locally.

### 2. A long test case name stretches the Incident Manager's name column

The cell declares `w-72`, but the table lays out `auto` and `overflow-wrap: break-word` does **not** reduce a word's min-content contribution. The column grew to the full width of the longest name — 900px on a 1033px container.

### 3. The Assignee column collapses when every row is unassigned

antd's `.ant-typography` sets `word-break: break-word`, which drops the "No Assignee" placeholder's min-content contribution to a single character. Owner names already render nowrap + ellipsis, so once every row is unassigned (e.g. filtering to `New`) nothing holds the column open and the placeholder stacks one letter per line.

## Solution

**Routing.** `useResolvedAppMode` already computes a `registrySettled` flag for exactly this window — it deliberately refuses to clear a valid session before the flag flips. It now returns that flag, and `AppRouter` holds a loader while a non-default mode is active but unregistered. Once settled and still unregistered (plugin genuinely uninstalled) it falls back to the default routes exactly as before.

The loader sits **inside** `AuthenticatedApp` deliberately: `ApplicationsProvider` lives there and is what flips `applicationsLoaded`. Short-circuiting above it deadlocks — the routes could never register, so the loader would never clear. There is a comment at the call site to stop that being "simplified" later.

**Table.** `wrap-anywhere` on the name link (it *does* shrink min-content) with the floor pinned to the same 18rem the cell declares, so auto-layout cannot then collapse the column. `whitespace-nowrap` on the assignee cell, matching the Last Updated cell in the same file.

## Measurements

| | before | after |
|---|---|---|
| Name column, 116-char name | 900px | 288px |
| Assignee column, all unassigned | 103px (placeholder 15×108px) | 166px (72×18px) |

## Tests

Two new `AppRouter` cases covering the registration window. Both were confirmed **failing** with the router change reverted (`2 failed, 5 passed`) and passing with it, so they are real regression coverage.

- `yarn jest src/components/AppRouter src/hooks/useResolvedAppMode src/hooks/useAppMode` — 11 suites, 112 passed
- `yarn jest src/components/IncidentManager/IncidentManagerTable.component.test.tsx` — 10 passed
- prettier + eslint clean on all changed files

## Type of change

- [x] Bug fix

----
## Summary by Gitar

- **DataQualityProvider enhancements:**
    - Added `isActive` prop to freeze URL tracking and avoid background refetches against shared query parameters in cached routes
- **Router and incident table bug fixes:**
    - Prevented deep links from 404ing on cold boot by holding a loader until non-default app mode routes are settled
    - Fixed incident table column sizing by updating cell width constraints and wrapping behaviors for name and assignee columns


ShaileshParmar11:
Added a fourth commit: `fix(ui): let a host freeze DataQualityProvider's URL tracking`.

**Problem.** `DataQualityProvider` derives its filters from the query string, which is global:

```js
const filterParams = omit(QueryString.parse(location.search), ['currentPage','pageSize','searchValue'])
useEffect(() => { fetchTestSummary(filterParams) }, [filterKey])
```

Its filter keys — `testPlatforms`, `tags`, `serviceName`, `dataQualityDimension`, `testCaseStatus`, `testCaseType` — overlap with the Test Library's. A host that keeps the page mounted while routing elsewhere (AI mode caches visited routes) therefore leaves a backgrounded Data Quality page re-deriving its filters from whatever route now owns the URL.

Captured off the wire — the *hidden* Data Quality page issuing an ES query built from the Test Library's filter:

```json
{"query":{"bool":{"must":[{"terms":{"testPlatforms":"GreatExpectations"}}, …
```

(Note `terms` with a string where an array is expected — malformed enough to 400 on a stricter index, which is what surfaces the toast from the generic `catch { showErrorToast(error) }`.)

That call also flips `isTestCaseSummaryLoading`, which unmounts and remounts the dashboard — measured at **26 stray requests per filter click**.

**Change.** An optional `isActive` prop (default `true`, so hosts that unmount on navigation are unaffected) short-circuits the effect while the page is backgrounded. Re-running on re-activation is intentional — it revalidates against the page's real filters.

Measured on the same repro after wiring it up downstream: **28 requests → 1**, no remount, and navigating back still revalidates.

shah-harshit (approved):

chirag-madlani (approved):

PR #31317, opened 2026-08-11
fix(playwright): eliminate flaky/failing ColumnBulkOperations tests

anuj-kumary:
## Summary

- **Replace `waitForResponse(status===200)` with `waitForGridRequest`** for URL-navigation and filter-apply steps so tests no longer depend on HTTP response status codes or a timing race between the API call and listener registration.
- **Add explicit `{ timeout: 30000 }` to `visitColumnBulkOperationsPage`** so a failing `beforeEach` fails fast (in 30 s) rather than silently consuming the full test budget.
- **Add `test.setTimeout(90000)` to the Filters & Search describe block** to give each test enough CI headroom without relying on the 60 s default.
- **New `waitForGridRequest` helper** (`page.waitForRequest`) resolves the moment the request is sent — no response-status dependency.

### Root-cause analysis

| Failing test | Root cause | Fix |
|---|---|---|
| `should clear individual filter and update URL` (always fails) | `waitForResponse(status===200)` for `page.goto('…?metadataStatus=MISSING')` — if the API returns non-200 for that filter value the promise never resolves | Switch to `waitForGridRequest` |
| `should filter by metadata status and verify API param` (flaky) | Listener registered after checkbox clicks; if filter fires on click (not update-btn) it is missed | Register listener before `update-btn.click()`, switch to `waitForGridRequest` |
| `should show pending progress spinner` beforeEach timeout | No timeout on `waitForResponse` — on slow CI it waits the full 120 s | Explicit `{ timeout: 30000 }` in `visitColumnBulkOperationsPage`; `test.setTimeout(90000)` for the group |

## Test plan

- [ ] Run `ColumnBulkOperations.spec.ts` in CI on the 2.0 / main branch — all 3 previously failing/flaky tests should now pass consistently
- [ ] Confirm no other tests in the file were regressed (the `searchColumn` helper still uses `waitForResponse` with a retry loop — unchanged)

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ColumnBulkOperations.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ColumnBulkOperations.spec.ts
@@ -29,17 +29,22 @@ const METADATA_STATUS_FILTER_TESTID = 'search-dropdown-Has / Missing Metadata';
 const GRID_API_URL = '/api/v1/columns/grid';
 const BULK_UPDATE_API_URL = '/api/v1/columns/bulk-update-async';
 
-async function waitForGridResponse(page: Page) {
-  return page.waitForResponse(
-    (r) => r.url().includes(GRID_API_URL) && r.status() === 200
-  );
+async function waitForGridRequest(page: Page, timeout = 15000) {
+  return page.waitForRequest((r) => r.url().includes(GRID_API_URL), {
+    timeout,
+  });
 }
 
 async function visitColumnBulkOperationsPage(page: Page) {
   await redirectToHomePage(page);
-  const dataRes = waitForGridResponse(page);
+  // Register before sidebarClick so the listener is active before navigation fires.
+  const responsePromise = page.waitForResponse(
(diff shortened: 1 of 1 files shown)


openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/ColumnBulkOperations.spec.ts
```diff
@@ -29,17 +29,22 @@ const METADATA_STATUS_FILTER_TESTID = 'search-dropdown-Has / Missing Metadata';
...
+    { timeout }
   );
 }
 
 async function visitColumnBulkOperationsPage(page: Page) {
   await redirectToHomePage(page);
-  const dataRes = waitForGridResponse(page);
+  // Register before sidebarClick so the listener is active before navigation fires.
+  const responsePromise = page.waitForResponse(
+    (r) => r.url().includes(GRID_API_URL) && r.status() === 200,
```
Rohit0301: don't check 200 status code here, add that after await

Rohit0301 (dismissed):

chirag-madlani (dismissed):

chirag-madlani (approved):

PR #31322, opened 2026-08-11
fix(workflows): reduce workflow polling delays on 1.13

yan-3005:
## Changes
- Reduces Flowable async-job acquisition from 60 seconds to 10 seconds and timer-job acquisition to 5 seconds.
- Reduces `WorkflowEventConsumer` polling from 10 seconds to 1 second.
- Migrates existing MySQL and PostgreSQL `WorkflowEventConsumer` subscriptions to the 1-second interval.
- Updates the generated TypeScript workflow-setting types.

## Scope
- Six files, three commits.
- This branch is rebased directly on `1.13`; it contains no unrelated commits.

## Validation
- Verified the JSON configuration, migration statements, and whitespace checks.
- Executed each subscription migration twice against disposable MySQL and PostgreSQL databases; only `WorkflowEventConsumer` changed from 10 to 1.

### DIFF
--- a/bootstrap/sql/migrations/native/1.13.4/mysql/postDataMigrationSQLScript.sql
+++ b/bootstrap/sql/migrations/native/1.13.4/mysql/postDataMigrationSQLScript.sql
@@ -12,3 +12,19 @@ UPDATE entity_extension
 SET json = JSON_SET(JSON_REMOVE(json, '$.appConfiguration'), '$.allowConfiguration', CAST('false' AS JSON))
 WHERE extension LIKE 'app.version.%'
   AND json->>'$.name' = 'McpApplication';
+
+-- Restore faster Flowable job-acquisition on 1.13.
+-- The 1.10.5 migration set asyncJobAcquisitionInterval/timerJobAcquisitionInterval to 60000 to
+-- reduce perceived idle polling, but 60s pickup starves the workflow engine under load: the
+-- governance change-event consumer backlogs by minutes and approval chains time out (2.0, which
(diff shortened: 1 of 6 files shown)


yan-3005:
@gitar-bot Both findings are on the `entity_relationship` `tableColumn`/`page` cleanup, which is **not part of this PR's change set**. That `DELETE` is a pre-existing `1.13` migration (authored separately); it only appears in this diff because the branch was synced with `1.13`. This PR's actual change is the `WorkflowEventConsumer` `pollInterval` (10→1) and the Flowable job-acquisition interval tuning.

On the flagged `DELETE` direction — this is a **false positive**. The migration author verified on a real instance that `fromEntity = 'tableColumn' AND toEntity = 'page'` matches and removes the intended rows; the stored direction is exactly as written. The review inferred the opposite direction from the comment prose, not the actual data.

Safe to dismiss both.

yan-3005:
> @gitar-bot Both findings are on the `entity_relationship` `tableColumn`/`page` cleanup, which is **not part of this PR's change set**. That `DELETE` is a pre-existing `1.13` migration (authored separately); it only appears in this diff because the branch was synced with `1.13`. This PR's actual change is the `WorkflowEventConsumer` `pollInterval` (10→1) and the Flowable job-acquisition interval tuning.
> 
> On the flagged `DELETE` direction — this is a **false positive**. The migration author verified on a real instance that `fromEntity = 'tableColumn' AND toEntity = 'page'` matches and removes the intended rows; the stored direction is exactly as written. The review inferred the opposite direction from the comment prose, not the actual data.
> 
> Safe to dismiss both.

@gitar-bot see this

PR #31324, opened 2026-08-11
Fixes #31323: Stream ingestion logs from any backend, not just S3

### LINKED ISSUE #31323: Live ingestion log streaming only works with S3 log storage; returns 404 on a default deployment
### Describe the bug

`GET /v1/services/ingestionPipelines/logs/{fqn}/stream/{runId}` is meant to tail a pipeline run's logs over SSE, but it only streams in one opt-in configuration, and is broken even there.

`IngestionPipelineRepository.streamLogs()` branches three ways:

```java
if (isS3LogStorageEnabled())      -> SSE
else if (isLogStorageEnabled())   -> getTraditionalLogs()   // one JSON blob
else                              -> 404 "Log storage is not configured"
```

| Deployment | Actual behaviour |
|---|---|
| No `logStorageConfiguration` in `openmetadata.yaml` (stock OSS + Airflow) | `setLogStorageConfiguration(null)` is called and `LogStorageFactory.create(null, …)` succeeds, so the fallback `catch` never runs and `logStorageConfiguration` stays null → `isLogStorageEnabled()` is false → **404** |
| `logStorageConfiguration.type: default` | `getTraditionalLogs()` returns `Response.ok(logs)` — a **JSON map, sent once, connection closes** — under `@Produces("text/event-stream")`. A `data:`-parsing client renders nothing |
| `logStorageConfiguration.type: s3` + bucket | Real SSE, but see below |

**The root gap:** there is no path to stream logs from the pipeline service. Object storage is the only source that can stream, so Airflow — where most deployments' logs actually live — has no live tail at all, only the paginated `/logs/{id}/last`.

The UI reflects this: `useLogStream.ts` exists but is imported by nothing except its own test, and `LogViewerModal.interface.ts` carries the comment *"reserved for re-enabling a self-fetching SSE stream (via `useLogStream`) once the backend endpoint is available again"*.

### Additional problems in the S3 path

- **Single-node only.** Live lines come from `s3Storage.registerLogListener(...)`, an in-JVM listener. Behind a load balancer, a viewer on node A never sees appends POSTed to node B — it gets the historical file and then silence forever.
- **History is deduped by count.** It sends 100 cached recent lines, then reads the whole object skipping `skipLines = recentLogs.size()`. Correct only if those 100 happen to be the file's first 100; otherwise it duplicates or drops lines.
- **Never terminates.** `while (!Thread.currentThread().isInterrupted()) { sleep(30000); }` — the stream stays open after the run finishes, and the client has no end-of-stream signal.
- **Pins a request thread.** The `StreamingOutput` body blocks a Jetty worker for the lifetime of every connection.
- **Reads scale with viewers.** Each connection opens its own full read of the object.
- **No cursor.** There is no way to resume after a dropped connection, and no way for a client to tell "the run finished" from "the socket dropped".

### To Reproduce

1. Start OpenMetadata with the default config (Airflow, no `logStorageConfiguration`).
2. Trigger any ingestion pipeline and note its `runId`.
3. `curl -N -H "Authorization: Bearer $OM_TOKEN" "http://localhost:8585/api/v1/services/ingestionPipelines/logs/<fqn>/stream/<runId>"`

**Expected:** log lines streamed as the run produces them.
**Actual:** `404 Log storage is not configured`.

### Expected behavior

One SSE endpoint that tails a run's logs on every deployment — reading from object storage when streamable logs are enabled for the pipeline and from the pipeline service otherwise — carrying a resume cursor, ending itself with an explicit event, sharing one backend reader across viewers, and holding no request thread.

### Version

- OS: any
- Python version: n/a
- OpenMetadata version: main

pmbrull:
### Describe your changes:

Fixes #31323

I worked on the ingestion log SSE endpoint because it only ever streamed when S3 log storage was configured — on a stock Airflow deployment `GET /logs/{fqn}/stream/{runId}` returns `404 Log storage is not configured`, and with `type: default` it returns a one-shot JSON body under `text/event-stream`. This rewrites the endpoint behind a shared streaming engine that reads from whichever backend actually holds the run's log, keeping the same URL rather than adding a second one. **Backend only** — no UI components are touched; the sole frontend file in this diff is the generated `logStreamEvent.ts` type.

#
### Type of change:

**Breaking:** frames are now one JSON `LogStreamEvent` each instead of one raw log line. A client parsing `data:` as log text must read `event.logs`. Nothing shipped consumes this endpoint today — `useLogStream.ts` is imported by nothing but its own test, beside a comment reading *"reserved for re-enabling a self-fetching SSE stream … once the backend endpoint is available again"* — so no user-visible behaviour changes here. Updating that hook to the new contract is deliberately left as follow-up UI work.

#
### High-level design:

**One endpoint, source chosen by where the bytes are.** `IngestionLogStreamFactory` picks `StorageLogTailSource` (object storage, cursor = line offset) when streamable logs are on for the pipeline and the run is keyed by UUID, otherwise `PipelineServiceLogTailSource` (Airflow/Argo, cursor = `chunk:charsDelivered`). The two cursor formats are not interchangeable, so the choice follows the bytes rather than the pipeline's configuration. Airflow's last chunk keeps growing between calls, so a chunk-index cursor alone would re-deliver it every poll — only the growth past the offset is emitted, and the source never asks for a chunk index the backend did not itself report (Airflow answers an out-of-range chunk with a 400).

**One reader per run, not per viewer.** `IngestionLogStreamManager` reference-counts an `IngestionLogTailer` per `(backend, pipeline, run)`. Ten tabs on the same run create ten SSE connections but one read loop against S3 or Airflow; the last viewer to leave stops it, so an unwatched run is not read at all. Polling runs on a shared scheduler sized from `maxActiveRuns` (one thread per 25 runs, clamped 2–16) because a poll is a blocking network read — a fixed pool of two would turn the poll delay into a queueing delay for every other run.

**Non-blocking.** The resource method is `void` with an injected `SseEventSink`, so the request thread returns immediately instead of being pinned for the connection's lifetime (the old `StreamingOutput` + `Thread.sleep` loop leaked one Jetty worker per viewer). `SseConnectionRegistry` — extracted from the existing `ProgressSseManager`, which now delegates to it — owns the heartbeat, the connection cap and the disconnect sweep.

**Everything is bounded.** `LogStreamSettings` caps reads and bytes per tick, total stream bytes and lifetime, idle time, the per-run replay buffer, unacknowledged bytes per viewer, concurrent runs and concurrent connections. A request past a cap gets an `error` event and a closed stream rather than being queued.

**Correctness decisions worth calling out.** A run with no status row is `UNKNOWN`, not finished — only the last five runs keep a status row, so "no row" also describes a run triggered seconds ago; treating it as terminal reported a just-triggered pipeline as already over ~10 s in, so it now gets a longer grace. `poll()` never throws, because an escaping exception silently cancels the scheduled task, leaking the run's tail slot and hanging its viewers. A joining viewer's `?after=` cursor is honoured against the shared replay buffer, and a cursor the server cannot place is answered with `truncated: true` rather than a silent gap.

**Alternatives rejected.** A second `/logs/{id}/last/stream` endpoint alongside the legacy one — dropped, since it left two shapes and a format enum to maintain for no consumer. Keeping raw-line frames for compatibility — dropped, since there is no working consumer to protect and cursors/end-of-stream have nowhere to live in that shape. Widening the path param to id-or-FQN and the `runId` to `String` is additive, so the URL itself is unchanged.

#
#### Use cases covered
- A user opens logs for a running pipeline on a stock Airflow deployment and sees lines appear live (previously a 404).
- A user opens logs on an S3 streamable-logs deployment and sees them live from any node behind the load balancer.
- The stream closes itself with `complete`/`runFinished` once the run ends, instead of holding the connection open.
- A dropped connection reconnects with `?after=<cursor>` and neither re-reads nor skips content.
- Ten viewers of the same run cost one backend reader.
- A just-triggered run whose status row has not landed is not reported as finished.
- A deployment with no log backend gets a single `error` event explaining why, not a hang.
- Airflow's `scheduled__<timestamp>` run id is accepted (previously 404'd by the `UUID` path param).

#### Unit tests
- [x] Added unit tests for the new logic. **47 tests, all passing** (`mvn test -pl openmetadata-service`).
- Files added: `IngestionLogTailerTest` (18), `IngestionLogStreamFactoryTest` (11), `PipelineServiceLogTailSourceTest` (7), `StorageLogTailSourceTest` (5), `IngestionLogStreamManagerTest` (3), plus `RecordingSseEventSink` / `TestSse` fakes.
- Time is driven by an injected clock, so timeouts are asserted exactly rather than slept through; no `Thread.sleep` in tests.
- Coverage: not measured with `jacoco:report` — the new `logstorage/stream` package is covered by the suites above, and the reviewer should treat the % as unverified rather than take a number I did not run.

- [x] Added `openmetadata-integration-tests/.../IngestionPipelineLogStreamIT.java` — asserts the response is committed immediately rather than buffered until the run ends, that every event names its type and run, that a resume cursor is accepted, that pipeline Id and FQN both resolve, that a non-UUID run id is accepted, that an unknown pipeline 404s, and that a deployment with no log backend refuses on the stream instead of hanging.

#### Playwright (UI) tests
- [x] Not applicable — no UI components are changed in this PR.

#### Manual testing performed
Not yet run against a live stack — verification here is the automated suites plus the integration tests above. Reviewer sanity check:
1. `docker compose -f docker/development/docker-compose.yml up -d`, trigger any metadata ingestion.
2. `curl -N -H "Authorization: Bearer $OM_TOKEN" ".../logs/<fqn>/stream/<runId>"` and confirm JSON frames arrive while the run is in progress, then a `complete` frame with `reason: runFinished`.
3. Re-run with `?after=<the last cursor>` and confirm nothing already delivered is repeated.

#

Not applicable — no UI components are changed. The only file under `openmetadata-ui/` is the generated type `src/generated/entity/services/ingestionPipelines/logStreamEvent.ts`, regenerated from the new schema via `json2ts-generate-all.sh`.

#
### Checklist:
- [x] My PR is linked to a GitHub issue via `Fixes #31323` above.
- [x] For JSON Schema changes: no migration needed — `logStreamEvent.json` describes a wire event, not a stored entity, so nothing is persisted.
- [x] For UI changes: not applicable, no UI components changed.

- [x] I have added the tag `Backward-Incompatible-Change`.


pmbrull:
Thanks — one applies, one no longer does.

**2. Pending-byte ceiling counts characters — valid, fixed** in 3a11133.

`maxPendingBytesPerClient` bounds what has been handed to the container on its way to a socket, so it is now charged in encoded UTF-8 length. Counted rather than measured: `getBytes(UTF_8).length` allocates a second copy of every frame purely to learn its size, on a path that runs per event per viewer.

Added `LogStreamSubscriberTest.chargesMultiByteContentByWhatItWeighsEncoded`, which fails on the old code (`expected: <false> but was: <true>`) and passes now.

The finding is broader than the one call site, so I addressed the rest deliberately rather than silently: `LogChunk.size()` feeds `maxBytesPerTick`, `maxStreamBytes` and `maxReplayBytes` with the same character count. Those stay in characters — exact for the ASCII log output overwhelmingly is, and the closer proxy for `maxReplayBytes`, whose subject is heap held in UTF-16 strings rather than bytes on the wire. Making them byte-exact would mean re-scanning every chunk on each access on a hot path, for ceilings where erring long is acceptable. Both units are now spelled out on `LogStreamSettings` so they cannot be mistaken for each other.

**1. UI reconnect backoff — does not apply to this diff.**

`useLogStream.ts` is unchanged from `origin/main` here; `git diff origin/main -- .../LogViewerModal/` is empty and the file contains no reconnect loop. A cursor-resuming version briefly existed on this branch and was reverted, since this PR is backend-only — the sole file under `openmetadata-ui/` is the generated `logStreamEvent.ts` type.

The underlying point is right, though, and applies to whoever wires the hook up next, so it is captured in the client contract in `docs/ingestion-log-streaming.md`: reconnect causes are usually persistent, each reconnect re-fetches the replay backlog, so use a capped exponential delay and give up after a few consecutive failures.

Backend suite is 49/49 green; `spotless:check` clean.

chirag-madlani (approved):

PR #31347, opened 2026-08-11
fix(profiler): emit valid Databricks SQL for unquoted identifiers

IceS2:
### Describe your changes:

Profiling a Databricks or Unity Catalog asset fails with `[PARSE_SYNTAX_ERROR] ... SQLSTATE: 42601` on every column, because the generated SQL has unbalanced identifier quoting:

```sql
-- generated
SELECT my_schema`.`my_table`.`my_col FROM my_schema.my_table
-- expected
SELECT my_schema.my_table.my_col FROM my_schema.my_table
```

`DatabricksProfilerInterface.visit_column` is monkeypatched onto `DatabricksDialect.statement_compiler` to re-quote STRUCT children (`` `db`.`sc`.`tbl`.`col.child` `` → `` …`col`.`child` ``). It assumed the compiler always emits backtick-quoted identifiers, but `DatabricksIdentifierPreparer` declares `legal_characters = ^[A-Z0-9_]+$`, so SQLAlchemy quotes an identifier only when it is mixed-case, reserved, or contains illegal characters. An all-lowercase `schema.table.column` therefore arrives **unquoted**, still satisfies the `dot_count > 1` guard, and gets rejoined with `` `.` `` separators while the branch that would supply the leading backtick is skipped — so the outer backticks are lost. Every column-level metric becomes unparseable on any table whose schema, table and column names are all lowercase, which is the common case.

The fix returns early when the compiled reference carries no backticks: there is no struct path to repair, so there is nothing to rewrite.

A second defect sits behind the first on the same code path. `ModuloFn` has no Databricks compiler and falls through to the generic `f"{value} %% {base}"`. The `%%` is an escape for `format`/`pyformat` paramstyle drivers; `databricks-sql-connector` uses `named` paramstyle, so nothing unescapes it and Spark receives a literal `%%`. Registering `Dialects.Databricks` on the existing `MOD(...)` group fixes it — without this the backtick fix alone still leaves profiling broken.

Separately, the progress and operation-metrics mixins logged only `str(exc)`, which reduces a server rejection to `400 Client Error: Bad Request for url: ...` and discards the server's reason. They now report a truncated response body alongside it, so this class of failure is diagnosable from agent logs.

#### Blast radius

This needs no opt-in. It reproduces on the **stock profiler defaults** (`randomizedSample: false`, `profileSampleConfig: DYNAMIC` with `smartSampling`), because that configuration still routes through `get_sample_query` and emits the `_rnd` sample CTE that forces three-part column references.

#### Verified against a live Databricks workspace

A/B on OM `2.0.0-rc1` against a real Unity Catalog service — same pipeline, same data, only these files differing:

| | Profiler records | errors | run state |
|---|---|---|---|
| unpatched | 12 | **90** | failed |
| patched | 102 | **0** | success |

Unpatched, both defects in one line:

```
[PARSE_SYNTAX_ERROR] Syntax error at or near 'customers'. SQLSTATE: 42601 (line 3, pos 16)
(SELECT sales`.`customers`.`ssn AS ssn, ABS(RANDOM()) * 100 %% :`param_3` AS random
FROM sales.customers),
```

Patched: 6/6 tables and 30/30 columns carry profiles, with real values.

New `test_databricks_profiler_sql.py` compiles against the real `DatabricksDialect` instead of a mocked compiler. The pre-existing `test_visit_column.py` only ever fed pre-quoted strings to a mocked `visit_column`, which is why the real preparer's behaviour was never observed; it gains the unquoted case.

#### Backports

Needs cherry-picking to `1.13` and `2.0` after merge. Both branches carry these files unchanged from `main`, so the picks apply clean.

#
### Type of change:
- [x] Bug fix

#
### Manual test steps:
1. Configure a Databricks or Unity Catalog service whose schema, table and column names are all lowercase.
2. Run metadata ingestion, then a profiler pipeline on default settings.
3. Before: every column metric fails with `[PARSE_SYNTAX_ERROR] Syntax error at or near '<table>'. SQLSTATE: 42601`, and only table-level `rowCount` succeeds. After: column metrics compute normally.

### DIFF
--- a/ingestion/src/metadata/ingestion/ometa/mixins/progress_mixin.py
+++ b/ingestion/src/metadata/ingestion/ometa/mixins/progress_mixin.py
@@ -25,6 +25,16 @@ from metadata.utils.logger import ometa_logger
 
 logger = ometa_logger()
 
+RESPONSE_BODY_LOG_LIMIT = 500
+
+
+def error_detail(exc: Exception) -> str:
+    """Truncated server response body for a failed request, empty when unavailable."""
+    body = getattr(getattr(exc, "response", None), "text", None)
+    if not body:
+        return ""
+    return f" - response: {body.strip()[:RESPONSE_BODY_LOG_LIMIT]}"
+
 
 class OMetaProgressMixin:
     """
@@ -53,7 +63,7 @@ class OMetaProgressMixin:
                 update.model_dump_json(exclude_none=True),
             )
         except Exception as exc:
-            logger.debug(f"Failed to send progress update: {exc}")
+            logger.debug("Failed to send progress update: %s%s", exc, error_detail(exc))
 
     def send_operation_metrics_batch(self, pipeline_fqn: str, run_id: str, batch: OperationMetricsBatch) -> None:
         """
@@ -71,7 +81,7 @@ class OMetaProgressMixin:
                 batch.model_dump_json(exclude_none=True),
             )
         except Exception as exc:
-            logger.debug(f"Failed to send operation metrics batch: {exc}")
+            logger.debug("Failed to send operation metrics batch: %s%s", exc, error_detail(exc))
 
     def get_progress_state(self, pipeline_fqn: str, run_id: str) -> Optional[ProgressUpdate]:  # noqa: UP045
         """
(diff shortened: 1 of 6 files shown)


pmbrull (approved):

ingestion/src/metadata/ingestion/ometa/mixins/progress_mixin.py
```diff
@@ -25,6 +25,16 @@
 
 logger = ometa_logger()
 
+RESPONSE_BODY_LOG_LIMIT = 500
+
+
+def error_detail(exc: Exception) -> str:
```
TeddyCr, line 31: Do you know why we need to truncate it?

TeddyCr (approved):

PR #31352, opened 2026-08-11
Fixes #30477: Add support for 24h format in the UI

### LINKED ISSUE #30477: Add support for 24h format in the UI

For us europeans, 24h formatting is the common way of displaying time. 16:30 instead of 4:30 PM

Two settings would be a good approach. One global default, simply by an ENV variable like TIMEFORMAT that have values 12 or 24. The second would be that an individual could set their own preference, overriding the global default

_No response_

### Additional context

_No response_

### Pre-submission checklist

jaya6400:
## Summary

Adds support for 24-hour time formatting in the UI (Fixes #30477). 24h formatting is the common way of displaying time, and previously the UI always rendered 12h.

Resolution precedence is: **user preference → server-side global default → `12h` fallback**

- A root-level `timeFormat` setting (`${TIME_FORMAT:-12h}`) in `openmetadata.yaml` gives admins a tenant-wide default.
- Each user can override it with an on-device preference (same category as `isSidebarCollapsed` / `globalPageSize`).

### Frontend API & Store
- `settingConfigAPI.ts` – `getSystemConfig` returns `timeFormat: '12h' | '24h'`.
- `store.interface.ts` / `useApplicationStore.ts` – global `timeFormat` state + `setTimeFormat` (default `'12h'`).
- `AppRoot.tsx` – seeds the store with the server default during the boot-time config fetch.
- `useCurrentUserStore.ts` – optional `timeFormat?: '12h' | '24h'` in `UserPreferences` for the per-user override.

### Frontend Components & Utilities
- `UserProfileIcon.component.tsx` – Time Format group (12-hour / 24-hour radio) in the profile dropdown.
- `DateTimeUtils.ts` – `getActiveTimeFormat` (user → global → 12h) + `getMappedTimeFormat`; `formatDateTime`, `formatDateTimeLong`, `customFormatDateTime`, `formatDateTimeWithTimezone`, `getScheduleDescriptionTexts` honor the resolved format.
- `en-us.json` – labels for the new setting.

## Design notes

After `25fa4d637f` reshaped server-persisted user preferences into a discriminated-union list, `timeFormat` was deliberately **kept as a local-only preference** and **not** added to `BACKEND_SYNCED_KEYS`. This avoids backend schema changes and DB migrations while still fully satisfying the per-user override requirement.

## Testing

- Recording:
> 
[screen-capture.webm](https://github.com/user-attachments/assets/d2ceb441-23f5-40bf-827d-8ee388c46145)

Screenshots:
1. Toggle:
> 
Capture

2. 12hr format:
>
12-hr-format

3. 24 hr format:
> 24-hr-format

- [x] `yarn test src/utils/date-time/DateTimeUtils.test.ts`  (new tests added for the time format changes - verified locally)
- 
  Capture

- [x] `make ui-checkstyle-changed` (fixed locally)
- 
  ui-checkstyle

- [x] `yarn i18n` (locale files changed after addition of new variables)
-  Added

- [x] `mvn spotless:apply` (verified locally)
-  
  spotless
  
- [x] `yarn ui-checkstyle`  (verify by CI CD as requires more than 8GB RAM device)
- [x] Manual: toggling 12h/24h in the profile menu updates timestamps across the UI; without a user preference the `TIME_FORMAT` env default applies; without either, the UI falls back to 12h.

## Related

Fixes #30477

<h2><a href="https://app.greptile.com/api/retrigger?id=52282679"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The functional changes appear sound, but the explicit localization requirement must be satisfied before merging because the new strings remain English in every non-English locale.

<h2>Findings</h2>

1. P2&nbsp;**Translations Remain English** <a href="https://github.com/open-metadata/OpenMetadata/pull/31352#discussion_r4052595887">▶</a>

This PR adds tenant-wide and per-user 12/24-hour display preferences and updates date-time formatting to honor the resolved preference.
- Adds a nullable tenant default to the application configuration schema and an administrator settings page.
- Adds a local user preference in the profile menu with precedence over the tenant default.
- Makes timestamp formatting respond to the selected hour cycle.
- Separates configuration requests during startup so one failed request does not discard successful configuration data.
- The changes since the previous review remove obsolete translation entries, but the previously reported untranslated replacement values remain outstanding.
</details>

```mermaid
flowchart LR
  U[User time-format preference] --> R{User preference set?}
  R -->|Yes| A[Active time format]
  R -->|No| T{Tenant default set?}
  T -->|Yes| A
  T -->|No| F[12-hour fallback]
  F --> A
  A --> D[Date-time formatting utilities]
  D --> UI[Rendered UI timestamps]
```
</details>

<sub>Reviews (9) · Last reviewed commit: ["chore(i18n): clean up unused time format..."](https://github.com/open-metadata/openmetadata/commit/4b6798d9f8ee3be29d97d54a258d5219019850f1)</sub>


jaya6400:
Hi, can maintainers add the `"safe to test"` label, for CI workflows to execute? @ShaileshParmar11 @aniketkatkar97

conf/openmetadata.yaml
```diff
@@ -20,6 +20,8 @@ assets:
 
 basePath: ${BASE_PATH:-/}
 
+timeFormat: ${TIME_FORMAT:-12h}
+
```
chirag-madlani: We can avoid it for yaml config

chirag-madlani (commented):
@jaya6400 can you help us with screenshot or recording so that we can review changes on the UI too?

conf/openmetadata.yaml
jaya6400: thanks for the review, i reverted the changes for backend/yaml file.

jaya6400:
@chirag-madlani thanks for the review. Added the screen recording & screenshots in the PR description, pls check.
- 24-hr-format
- 12-hr-format
[screen-capture.webm](https://github.com/user-attachments/assets/ce8097d5-76aa-492b-98e1-77e88435e032)

If further changes needed, do let me know.

jaya6400:
Hi @chirag-madlani, I've resolved the merge conflicts with main and rebased the branch. I've also addressed the `gitar-bot` four review comments. could you add the `"safe to test"` label to start the CI checks?

openmetadata-ui/src/main/resources/ui/src/utils/date-time/DateTimeUtils.ts
```diff
@@ -45,36 +48,74 @@ const loadCronstrue = (): Promise<CronstrueModule> => {
...
+export const getActiveTimeFormat = (): '12h' | '24h' => {
+  const { currentUser, timeFormat: globalTimeFormat } =
+    useApplicationStore.getState();
+
+  const userTimeFormat = currentUser?.name
+    ? usePersistentStorage.getState().preferences[currentUser.name]?.timeFormat
+    : undefined;
+
+  return userTimeFormat ?? globalTimeFormat ?? '12h';
+};
```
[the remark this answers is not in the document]
jaya6400: Resolved in the latest commit

ShaileshParmar11:
@jaya6400 Please check the CI failure, let us know if any help needed, Thank you!

openmetadata-ui/src/main/resources/ui/src/AppRoot.tsx
```diff
@@ -86,9 +103,7 @@ const AppRoot: FC = () => {
   return (
     <div className="main-container">
       <div className="content-wrapper" data-testid="content-wrapper">
-        <BrowserRouter
-          basename={getBasePath()}
-          useTransitions={!isPlaywrightEnv()}>
+        <BrowserRouter basename={getBasePath()}>
```
chirag-madlani: Is this change related?

jaya6400: thanks for pointing it, that wasn't intentional. Looks like `useTransitions={!isPlaywrightEnv()}` got dropped during a rebase. Restoring it now, unrelated to this PR's changes. 
Restored and pushed the changes.

jaya6400:
@ShaileshParmar11 thanks, I've fixed the UI CI failures errors, I see bot removed the `safe to test` label, could u please re add it.

openmetadata-ui/src/main/resources/ui/src/components/Settings/Users/UserProfileIcon/UserProfileIcon.component.tsx
```diff
@@ -233,18 +240,28 @@ export const UserProfileIcon = () => {
     setIsDropdownOpen(false);
   }, []);
 
+  const timeFormatLabelRenderer = useCallback(
+    (format: '12h' | '24h', label: string) => (
+      <div
+        className="w-full d-flex items-center justify-between cursor-pointer"
+        data-testid={`time-format-${format}`}
+        onClick={() => handleTimeFormatChange(format)}>
```
ShaileshParmar11, line 248: This row is mouse-only — it's a `<div onClick>` with a display-only `<Radio checked>`, so the 12h/24h options can't be focused or toggled by keyboard. ESLint flags it (this file goes from 0 errors on `main` to 3):

```
245:7  error  Visible, non-interactive elements with click handlers must have at least one keyboard listener
245:7  error  Avoid non-native interactive elements...
```

Worth fixing rather than silencing. `handleKeyboardActivation` is already imported in this file, or an Ant `Radio.Group` (matching how the persona rows work) would handle it for free.

openmetadata-ui/src/main/resources/ui/src/components/Settings/Users/UserProfileIcon/UserProfileIcon.component.tsx
```diff
@@ -446,7 +492,6 @@ export const UserProfileIcon = () => {
         data-testid="dropdown-profile"
         icon={
           isImgUrlValid ? (
-            // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions -- onError load fallback
             <img
```
ShaileshParmar11, line 495: The `// eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions -- onError load fallback` comment above this `` got dropped, which brings back a third ESLint error here (`495:13`). Restoring the line clears it — looks like incidental rebase fallout rather than intentional.

openmetadata-ui/src/main/resources/ui/src/components/Settings/Users/UserProfileIcon/UserProfileIcon.component.tsx
```diff
@@ -24,17 +24,18 @@ import { ReactComponent as LogoutIcon } from '../../../../assets/svg/logout.svg'
...
 import { EntityReference } from '../../../../generated/entity/type';
+import { useCurrentUserPreferences } from '../../../../hooks/currentUserStore/useCurrentUserStore';
 import { useApplicationStore } from '../../../../hooks/useApplicationStore';
 import { getEntityName } from '../../../../utils/EntityNameUtils';
 import { handleKeyboardActivation } from '../../../../utils/KeyboardUtil';
 import navbarUtilClassBase from '../../../../utils/NavbarUtilClassBase';
 import {
-  getImageWithResolutionAndFallback,
-  ImageQuality,
+    getImageWithResolutionAndFallback,
```
ShaileshParmar11, line 33: Small thing, but this import block got reindented to 4 spaces with the trailing comma dropped — same pattern in `useApplicationStore.ts`, `store.interface.ts` and `DateTimeUtils.ts`. Looks like an editor auto-format that didn't pick up the repo Prettier config.

All four files are prettier-clean on `main` and dirty on the branch (45 / 10 / 8 / 6 differing lines). `make ui-checkstyle-changed` fixes them all in one pass.

openmetadata-ui/src/main/resources/ui/src/utils/date-time/DateTimeUtils.ts
```diff
@@ -45,13 +46,57 @@ const loadCronstrue = (): Promise<CronstrueModule> => {
...
+ * 3. `12h` fallback.
+ * 
+ * NOTE: This function reads the time format non-reactively via .getState().
+ * Components using utility formatters directly will not automatically re-render 
+ * when the user toggles the time format preference. They will reflect the new 
+ * format on their next render cycle (e.g., via navigation or unrelated state changes).
+ * For fully reactive time formatting, components should subscribe to the time 
+ * format preference via useCurrentUserPreferences() and pass it to customFormatDateTime.
+ */
+export const getActiveTimeFormat = (): '12h' | '24h' => {
```
ShaileshParmar11, line 72: This is the gitar/greptile comment that's still open — the `hour12` one you fixed is genuinely resolved, but this one isn't yet, and the JSDoc above documents it.

Because both reads go through `.getState()`, nothing re-renders when the preference changes, so from a user's point of view the toggle appears to do nothing until they navigate away and back. For the headline interaction of the feature that's worth fixing rather than documenting — a small `useTimeFormat()` hook that subscribes to `timeFormat` where timestamps render is one option.

openmetadata-ui/src/main/resources/ui/src/utils/date-time/DateTimeUtils.ts
```diff
@@ -45,13 +46,57 @@ const loadCronstrue = (): Promise<CronstrueModule> => {
...
+  return userTimeFormat ?? globalTimeFormat ?? '12h';
+};
+
+/**
+ * Maps the time tokens of a luxon format string to the requested
+ * 12h/24h representation; date tokens are left untouched.
+ * e.g. getMappedTimeFormat("MMM dd, yyyy, hh:mm a '(UTC'ZZ')'", '24h')
+ *  => "MMM dd, yyyy, HH:mm '(UTC'ZZ')'"
+ */
+export const getMappedTimeFormat = (
```
ShaileshParmar11, line 88: The mapping itself is correct — I ran it against all four exported constants and they convert cleanly in both directions.

Two edge cases worth a guard or at least a comment, neither of which any caller hits today:
- the regex needs `a` immediately after `mm`, so `hh:mm:ss a` stays 12h under a 24h preference
- in the other direction `HH:mm:ss` becomes `hh:mm a:ss` -> renders as `01-05-2026 07:04 PM:32`

This is also the main thing that needs a test: a table over the four constants x `12h`/`24h`, plus `getActiveTimeFormat` precedence (user set -> wins, user unset -> global, neither -> `12h`). `DateTimeUtils.test.ts` is currently byte-identical to `main`, so the `yarn test DateTimeUtils.test.ts` checkbox in the description ran an unchanged file.

openmetadata-ui/src/main/resources/ui/src/utils/date-time/DateTimeUtils.ts
```diff
@@ -551,4 +543,4 @@ export const getScheduleDescriptionTexts = (scheduleInterval: string) => {
   } catch {
     return { descriptionFirstPart: '', descriptionSecondPart: '' };
   }
-};
+};
```
ShaileshParmar11, line 546: Tiny one: the trailing newline at end of file got stripped here (it's there on `main`). `make ui-checkstyle-changed` will put it back.

ShaileshParmar11 (commented):
Thanks for sticking with this one, @jaya6400 — and for patiently working through the bot rounds. I went through the branch at `88e20e54` in detail. The design is right: user → tenant → `12h` is the correct precedence, and I verified your `getMappedTimeFormat` mapping against the real format constants through Luxon — all four convert cleanly:

```
MMM dd, yyyy, hh:mm a '(UTC'ZZ')'   24h ->  Jan 05, 2026, 19:04 (UTC+05:30)   OK
MMMM dd, yyyy, h:mm a '(UTC'ZZ')'   24h ->  January 05, 2026, 19:04 (...)     OK
ccc d'th' MMMM, yyyy, hh:mm a       24h ->  Mon 5th January, 2026, 19:04      OK
```

So the core is solid. I've left inline notes on the specific spots — summary here.

### First, please ignore the red CI, it isn't you :)

The `safe to test` label isn't on this PR, so every red job stopped at the label gate (`Error! This pull request does not contain any of the valid labels: ['safe to test']`) — nothing in your code failed. The `Playwright impact-map drift` failure is a fork-checkout artifact, also not yours. I'll add `safe to test` so the real jobs can run on your next push.

One thing to be aware of: **the green `ui-checkstyle` check is misleading.** It's a summary job, and the `checkstyle` job it summarises was skipped (0 steps run). So the lint/format/i18n gate has never actually run on this branch — which is why none of the items below were surfaced to you earlier. Sorry about that.

### Blockers

1. **Locale files** — only `en-us.json` has the new keys; the other 19 have none. `yarn i18n` + translate the placeholders (inline note on the file).
2. **Prettier + ESLint** — `make ui-checkstyle-changed`. Four files are prettier-clean on `main` and dirty here, and `UserProfileIcon` goes from **0 ESLint errors on main to 3**. One of those is a real keyboard-accessibility gap on the new rows (inline).
3. **Tests** — `DateTimeUtils.test.ts` is byte-identical to `main`, so the `yarn test` checkbox ran an unchanged file. `getMappedTimeFormat` and `getActiveTimeFormat` both want a small table test (inline).
4. **The global/ENV tier is currently dead code** — makes sense that the yaml/Java came out on @chirag-madlani's review, but the UI-side reader stayed behind (inline on `AppRoot.tsx`). Needs a decision either way.
5. **The toggle doesn't update timestamps already on screen** — still open from the bot round (inline on `DateTimeUtils.ts`).

### Minor

- `RUN_STARTED_AT_FORMAT` in `agentsDataMapper.ts:41` is `'MMM dd, yyyy · HH:mm'` — our only always-24h surface. With this PR it renders as `Jan 05, 2026 · 07:04 PM` on default settings. Probably intended, just flagging it's a visible change to Service Agents.
- Could you revert the unrelated whitespace/comment deletions in `DateTimeUtils.ts` and `UserProfileIcon`? A few explanatory comments went with them (the cronstrue lazy-load note, the Ant Design group-label rationale). Keeping the diff to just the feature makes re-review much quicker.

### To clear the last red check

`Validate PR Metadata` is red because #30477 isn't on the **Shipping** project — adding it there and pushing will clear it.

Really nice work on the core logic here — it's mostly hygiene plus the one open design question. Ping me when it's updated and I'll take another look.

openmetadata-ui/src/main/resources/ui/src/AppRoot.tsx
```diff
@@ -34,28 +34,54 @@ import { getThemeConfig } from './utils/ThemeUtils';
...
+        setApplicationConfig({
+          ...themeData,
+          customTheme: getThemeConfig(themeData.customTheme),
+        });
+      }
 
-      setRdfEnabled(systemConfig.rdfEnabled || false);
+      if (systemConfig) {
+        setRdfEnabled(systemConfig.rdfEnabled || false);
+        setTimeFormat(systemConfig.timeFormat || '12h');
```
ShaileshParmar11, line 83: This is the one that needs a decision rather than a fix.

It makes sense that the yaml/Java came out on @chirag-madlani's review — but the UI-side reader stayed behind. `ConfigResource.getConfig()` returns only `basePath` and `rdfEnabled`, and there's no `timeFormat` anywhere in `openmetadata-service/` or `conf/`, so this line always resolves to `'12h'` and the tenant tier of the documented precedence is effectively dead.

Since #30477 asks for a global ENV default *and* a per-user override, could you and @chirag-madlani settle which way to go? Either wire the backend field back, or drop this reader plus the type on `settingConfigAPI.ts` and we track the global default as a follow-up issue. Right now it reads as supported but silently isn't.

openmetadata-ui/src/main/resources/ui/src/rest/settingConfigAPI.ts
```diff
@@ -140,6 +140,7 @@ export const getSystemConfig = async () => {
   const response = await axiosClient.get<{
     basePath: string;
     rdfEnabled: boolean;
+    timeFormat?: '12h' | '24h';
```
ShaileshParmar11, line 143: Pairs with the note on `AppRoot.tsx` — the `/system/config/rdf` endpoint never returns this field today, so it's declared but always `undefined`. Whichever way the global-default question goes, this line and the `AppRoot` reader should move together.

openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
```diff
@@ -3,6 +3,8 @@
     "-with-colon": "{{text}}:",
     "1-day": "1 day",
     "1-hr": "1 hour",
+    "12-hour": "12-hour",
+    "24-hour": "24-hour",
```
ShaileshParmar11, line 7: These three keys only landed in `en-us.json` — the other 19 locale files have none of them:

```
en-us : 3/3        de-de fr-fr ja-jp es-es zh-cn pt-br ru-ru ... : 0/3
```

```bash
cd openmetadata-ui/src/main/resources/ui && yarn i18n
```

That propagates the keys, but it copies the English string in as a placeholder — please replace those with real translations. Our locales carry genuine ones (e.g. `dark-mode` is `Dunkelmodus` / `Mode sombre`), and every recent PR touching `en-us.json` updates all 20 files, so CI will expect it. `yarn i18n` is part of `yarn ui-checkstyle`, which is why the skipped gate didn't catch this.

openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
```diff
@@ -4668,6 +4671,7 @@
     "tier-coverage-widget-description": "Tier coverage for all data assets in the service. <0>learn more.</0>",
     "tier-distribution-description": "Tier captures the business importance of the data.",
     "tier-distribution-widget-description": "Overview of all tier distribution for all data assets in the service. <0>learn more.</0>",
+    "time-format-description": "Choose how times are displayed across the application on this device.",
```
ShaileshParmar11, line 4674: This key isn't referenced anywhere in the diff — the menu only uses `label.time-format`, `label.12-hour` and `label.24-hour`. Either wire it in as a description under the group label, or drop it so it doesn't ship as a dead string across 20 locales.

jaya6400:
Hi! I've addressed all the review comments:
✅ Accessibility: Time format rows now have proper keyboard support (Enter/Space activation)
✅ Edge cases: Fixed regex to prevent HH:mm:ss → hh:mm a:ss bug
✅ Reactivity: Added optional timeFormat param to formatters for components that need live updates
✅ Tests: Added 15 new tests covering all 4 format constants + edge cases + precedence chain in `DateTimeUtils.test.ts`
✅ i18n: Removed unused time-format-description key, Sync i18n keys (12-hour, 24-hour, time-format) across all locales
✅ ESLint: Restored suppression comment for profile image

Checkstyle: make ui-checkstyle-changed passed with 0 errors 
Tests: All 117 tests passing (updated the PR description with screenshots of the same)

pls add the `safe to test` label for CI workflows to run. do let me know if further changes required. 
@ShaileshParmar11 @chirag-madlani 

- Capture

jaya6400:
Hi maintainers, Maven Collate Tests and Postgresql PR Playwright E2E are both failing at `yarn install --frozen-lockfile` with a registry 503 (`registry.yarnpkg.com` ... `503 Service Unavailable`), before any PR code is even built. 

Not related to the diff - could you re-run those two?

Playwright impact-map drift is the fork-checkout artifact already flagged.

Also, UI Checkstyle now shows 0 errors, 7 warnings- confirming that's a pass ? before I look at SonarCloud coverage.

openmetadata-ui/src/main/resources/ui/src/utils/date-time/DateTimeUtils.ts
jaya6400: Added 15 new tests covering all 4 format constants + edge cases + precedence chain in DateTimeUtils.test.ts

openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
jaya6400: i18n: Removed unused time-format-description key, Sync i18n keys (12-hour, 24-hour, time-format) across all locales

openmetadata-ui/src/main/resources/ui/src/locale/languages/en-us.json
jaya6400: i18n: Removed unused time-format-description key, Sync i18n keys (12-hour, 24-hour, time-format) across all locales

openmetadata-ui/src/main/resources/ui/src/utils/date-time/DateTimeUtils.ts
jaya6400: Fixed, `make ui-checkstyle-changed` now shows 0 errors.

openmetadata-ui/src/main/resources/ui/src/components/Settings/Users/UserProfileIcon/UserProfileIcon.component.tsx
jaya6400: Fixed

openmetadata-ui/src/main/resources/ui/src/components/Settings/Users/UserProfileIcon/UserProfileIcon.component.tsx
jaya6400: Fixed.

openmetadata-ui/src/main/resources/ui/src/rest/settingConfigAPI.ts
jaya6400: Added optional timeFormat param to formatters for components that need live updates

openmetadata-ui/src/main/resources/ui/src/components/Settings/Users/UserProfileIcon/UserProfileIcon.component.tsx
jaya6400: Time format rows now have proper keyboard support (Enter/Space activation)
✅ Edge cases: Fixed regex to prevent HH:mm:ss → hh:mm a:ss bug

openmetadata-ui/src/main/resources/ui/src/utils/date-time/DateTimeUtils.ts
jaya6400: I wanted to clarify the architectural decision here:
Per the original review feedback (feedback point 5 regarding reactive time-format behavior), I avoided making non-React utility functions dependent on React state, and I avoided a massive refactor of existing components to keep the diff minimal and focused on Issue https://github.com/open-metadata/OpenMetadata/issues/30477.

Instead, I implemented useActiveTimeFormat as an opt-in hook and added an optional timeFormat parameter to the shared formatters (formatDateTime, customFormatDateTime, etc.). This provides the reactive foundation for components to update live without forcing a high-risk, application-wide refactor of every timestamp consumer in this specific PR.

Could you please review and let me know if this scoped approach is acceptable, or if you would prefer I proactively update specific high-visibility components to use the new hook in this PR?

openmetadata-ui/src/main/resources/ui/src/AppRoot.tsx
jaya6400: `timeFormat` is kept across UI components, `AppRoot.tsx`, settingsConfigAPI.ts, hooks like `useApplicationStore`, and tsx files `UserProfileIcon.component.tsx` , for consistency, readability and removed from backend Java/yaml files as making it frontend related changes in the first version. If later further backend support needed follow up PR can be made.

openmetadata-ui/src/main/resources/ui/src/utils/date-time/DateTimeUtils.ts
```diff
@@ -45,25 +48,96 @@ const loadCronstrue = (): Promise<CronstrueModule> => {
...
+ * format on their next render cycle (e.g., via navigation or unrelated state changes).
+ * For fully reactive time formatting, use the {@link useActiveTimeFormat} hook
+ * and pass its result to customFormatDateTime / formatDateTime.
+ */
+export const getActiveTimeFormat = (): '12h' | '24h' => {
+  const { currentUser, timeFormat: globalTimeFormat } =
+    useApplicationStore.getState();
+  const userTimeFormat = currentUser?.name
+    ? useCurrentUserPreferences.getState().preferences[currentUser.name]
+        ?.timeFormat
```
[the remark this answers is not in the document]
jaya6400: fixed in the latest commit. https://github.com/open-metadata/OpenMetadata/pull/31352/commits/88d881075380d6db0e07ab53ac408b251b1c9fac
fix(ui): restore usePersistentStorage.getState() in getActiveTimeFormat
- Revert getActiveTimeFormat to use usePersistentStorage.getState() for non-reactive reads, fixing a TypeError crash for logged-in users.
- Add useActiveTimeFormat hook using useCurrentUserPreferences() for components requiring reactive time format updates.
- Update test mocks to correctly distinguish between the Zustand store (usePersistentStorage) and the React hook (useCurrentUserPreferences).

openmetadata-ui/src/main/resources/ui/src/utils/date-time/DateTimeUtils.ts
```diff
@@ -45,25 +48,96 @@ const loadCronstrue = (): Promise<CronstrueModule> => {
...
+    : undefined;
+
+  return userTimeFormat ?? globalTimeFormat ?? '12h';
+};
```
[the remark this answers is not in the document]
jaya6400: fixed in the latest commit. https://github.com/open-metadata/OpenMetadata/pull/31352/commits/88d881075380d6db0e07ab53ac408b251b1c9fac

jaya6400:
@ShaileshParmar11 @chirag-madlani PR update: I've replied & addressed all review comments. Requesting maintainer's review on it & adding `safe to test` label again.

I noticed Greptile flagged the PR as 4/5 because the new `useActiveTimeFormat` hook isn't being consumed by existing components yet. 

I wanted to clarify the architectural decision here: 
Per the original review feedback (feedback point 5 regarding reactive time-format behavior), I avoided making non-React utility functions dependent on React state, and I avoided a massive refactor of existing components to keep the diff minimal and focused on Issue #30477. 

Instead, I implemented `useActiveTimeFormat` as an opt-in hook and added an optional `timeFormat` parameter to the shared formatters (`formatDateTime`, `customFormatDateTime`, etc.). This provides the reactive foundation for components to update live without forcing a high-risk, application-wide refactor of every timestamp consumer in this specific PR. 

Could you please review and let me know if this scoped approach is acceptable, or if you would prefer I proactively update specific high-visibility components to use the new hook in this PR? 

✅ **Playwright Fix:** Fixed the `Users.spec.ts` strict mode violation by scoping the checked radio button assertion to the specific persona element.

Thanks!

karanh37:
Hi @jaya6400, You are storing it in memory which we shouldn't be doing. We should be driving this decision from Settings > Preferences.

Backend already have mechanism. Global admin-set config = Settings entity: enum key + JSON value, DB-backed, admin-mutable.

extend appConfiguration

`openmetadata-spec/src/main/resources/json/schema/api/configuration/appConfiguration.json`

- Purpose-built: “App-wide UI configuration. Seeded from yaml/env on first boot; DB-backed and admin-mutable at runtime.”
- Already holds defaultAppMode (tenant default, user/persona override wins) — exact pattern you want.
- Already wired: SettingsType.appConfiguration, SystemRepository, SettingsCache, read by UI via useApplicationStore.

Add field:

`"defaultTimeFormat": { "type": "string", "enum": ["12", "24"], "default": "12" },`

Settings UI page under system settings preferences.
Consume via `useApplicationStore`

karanh37 (changes requested):
Refer https://github.com/open-metadata/OpenMetadata/pull/31352#issuecomment-5740327712

openmetadata-ui/src/main/resources/ui/src/pages/Settings/DefaultTimeFormatPage/DefaultTimeFormatPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+const OPTIONS: TimeFormatOption[] = [
+  { value: NO_DEFAULT_VALUE, labelKey: 'label.no-default' },
+  { value: '12h', labelKey: 'label.12-hour-format' },
+  { value: '24h', labelKey: 'label.24-hour-format' },
+];
+
+const DefaultTimeFormatPage: React.FC = () => {
+  const { t } = useTranslation();
+  const { setTimeFormat } = useApplicationStore((state) => ({
```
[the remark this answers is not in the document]
jaya6400: fixed in commit: https://github.com/open-metadata/OpenMetadata/pull/31352/commits/e009c8fdd66b7ca895733adb5c83c894d01aef2b

openmetadata-ui/src/main/resources/ui/src/pages/Settings/DefaultTimeFormatPage/DefaultTimeFormatPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+          ? null
+          : (currentValue as '12h' | '24h');
+          
+      await patchAppConfiguration({ defaultTimeFormat });
+      setInitialValue(currentValue);
+      
+      // Update the global store immediately so the UI reflects the change reactively
+      if (defaultTimeFormat) {
+        setTimeFormat(defaultTimeFormat);
+      }
```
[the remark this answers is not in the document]
jaya6400: Resolved: [fix(ui): reactively update store when tenant time format default is cleared](https://github.com/open-metadata/OpenMetadata/pull/31352/changes/2a5cec6e1ad60073f434f954af9e0c733ef05cb9)

openmetadata-ui/src/main/resources/ui/src/locale/languages/ar-sa.json
```diff
@@ -3,6 +3,8 @@
     "-with-colon": "{{text}}:",
     "1-day": "يوم واحد",
     "1-hr": "ساعة واحدة",
+    "12-hour": "12-hour",
+    "24-hour": "24-hour",
```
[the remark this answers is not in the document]
jaya6400: resolved: chore(i18n): clean up unused time format keys and sync locales
commit : https://github.com/open-metadata/OpenMetadata/pull/31352/commits/4b6798d9f8ee3be29d97d54a258d5219019850f1

jaya6400:
@karanh37 Thanks for the direction on using `AppConfiguration`. 
I've refactored the PR to follow the exact same pattern as `defaultAppMode`.
Here is a summary of the changes implemented based on your feedback:
- Architecture & Cleanup: Added defaultTimeFormat to the `appConfiguration.json` schema and ran the necessary codegen. Removed the dead timeFormat read from `/system/config/rdf` in `settingConfigAPI.ts`.
- Store Wiring: Updated AppRoot.tsx to fetch `getAppConfiguration` on boot and dispatch the tenant default to the global store. The precedence chain `(User Preference > Tenant Default > Hardcoded '12h')` in `DateTimeUtils.ts` is fully preserved.
- Settings UI: Created `DefaultTimeFormatPage` (mirroring DefaultAppModePage) and registered it under Settings > Preferences.
- Edge Case Fix: Addressed a P1 issue flagged by Greptile where clearing the tenant default (selecting "No Default" / null) left stale state in the UI. I updated the save handler to use `setTimeFormat(defaultTimeFormat ?? '12h')` to reactively fallback to the hardcoded default.
- i18n: Added the required translation keys to `en-us.json` and synced the locales.

Could you please take a look and run the CI checks if the code structure seems fine?

openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Users.spec.ts
```diff
@@ -857,7 +857,7 @@ test.describe('User Profile Dropdown Persona Interactions', () => {
 
       // Verify the selected persona is checked
       const checkedRadio = adminPage.locator('input[type="radio"]:checked');
-
+
```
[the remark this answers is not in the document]
jaya6400: removed this file change as not needed after design change

openmetadata-ui/src/main/resources/ui/src/generated/api/configuration/appConfiguration.ts
```diff
@@ -21,6 +21,7 @@ export interface AppConfiguration {
      * Null means no tenant default is configured.
      */
     defaultAppMode?: DefaultAppMode | null;
+    defaultTimeFormat?: string | null;
```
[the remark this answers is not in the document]
jaya6400: null value removed in this commit: https://github.com/open-metadata/OpenMetadata/pull/31352/changes/1d8aa696344a1ceac62fc3180d08068481db0a79

openmetadata-ui/src/main/resources/ui/src/components/Settings/Users/UserProfileIcon/UserProfileIcon.component.tsx
```diff
@@ -86,35 +84,36 @@ const renderLimitedListMenuItem = ({
...
-            : null,
-        ],
+              },
+            ]
+          : []),
       ];
 };
 
 export const UserProfileIcon = () => {
   const { currentUser, selectedPersona, setSelectedPersona } =
```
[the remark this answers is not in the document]
jaya6400: [UserProfileIcon.component.tsx](https://github.com/open-metadata/OpenMetadata/pull/31352/files/82cef5b13fc57dd2b735290a834bf8e450e2367e#diff-445f2cf57f3c078df25ec0bfc76e99e21d77a0c5233e37f0f418ab3175ef4d87) is reverted back to old code after the change in design decision to `Settings->Preferences` flow.

openmetadata-ui/src/main/resources/ui/src/rest/settingConfigAPI.ts
```diff
@@ -164,3 +173,4 @@ export const updateGlossaryTermRelationSettings = async (
 
   return response.data;
 };
+
```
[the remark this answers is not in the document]
jaya6400: removed in commit: https://github.com/open-metadata/OpenMetadata/pull/31352/changes/d400cfa28f32afb57908a8b9fce2bc74f8d70fe0

jaya6400:
Addressed the feedback regarding the per-user override. 

- Updated the PR description with a `new recording demonstrating` the revised **Settings > Preferences** flow (tenant-wide only).
- Refactored `[DateTimeUtils.ts](http://datetimeutils.ts/)` and `[DateTimeUtils.test.ts](http://datetimeutils.test.ts/)` to remove the user-preference checks, simplifying the resolver to rely solely on the tenant-wide default.
- Updated the schema description and admin page copy to reflect that this is now strictly a tenant-wide setting without a user-level override.
- 1Capture
- Capture

jaya6400:
@karanh37 Thanks for the merge commit. could a maintainer pls add `safe to test` label on this PR as well. I can resolve the [minor quality related gitar-bot](https://github.com/open-metadata/OpenMetadata/pull/31352#discussion_r4089474428) comment if needed, after checking the CI checks.

@chirag-madlani @ShaileshParmar11

PR #31375, opened 2026-08-12
fix(drive): add deleteStale endpoint and set File column FQNs on write

### LINKED ISSUE #31381: Drive connectors: every ingestion run fails, and re-runs cannot update files
**Affected module:** Backend / Ingestion (Drive entities — SFTP, GoogleDrive, SharePoint)

### Describe the bug

Two defects in the Drive entity layer make Drive connectors unusable in normal operation.

**1. `deleteStale` is not implemented for Drive.** Every ingestion run ends with three
`HTTP 405 Method Not Allowed` on `/v1/drives/{directories,files,spreadsheets}/deleteStale`. The
endpoint exists for tables, topics and containers, but on none of the Drive resources.

So `markDeletedDirectories` / `markDeletedFiles` are silent no-ops — entities deleted at the source
stay in the catalogue forever — and **every** run reports Failed even when extraction succeeded.

**2. Re-ingesting a file with columns fails with a server NPE.**

```
Cannot invoke "String.equals(Object)" because the return value of
"org.openmetadata.schema.type.Column.getFullyQualifiedName()" is null
```

`FileRepository` sets column FQNs on the read path but not on write, so columns are persisted with a
null FQN and the update comparison NPEs. Metadata is frozen at whatever the first run captured.

### To Reproduce

1. Create a Drive service pointing at a source with at least one CSV, and run a metadata agent.
2. The run ingests correctly but reports **Failed** — the log shows the three 405s.
3. Delete a file at the source, keep "Mark Deleted Files" on, re-run — the file is still in the
   catalogue, not soft-deleted.
4. Re-run again unchanged — every file with columns now fails with the NPE (0 errors on run 1, 17 on
   run 2 in our testing).

### Expected behavior

- Entities missing from the source are soft-deleted when `markDeleted*` is on
- A run that extracts successfully reports Success
- Re-runs are idempotent and can update existing files

### OpenMetadata version

2.0.0 — also reproduces on `main`. Both reproduce through the CLI too, so neither is agent-specific.

Khairajani:
Fixes #31381

Two defects found while testing the SFTP connector for the 2.0 release. Both sit in the Drive entity
layer, so they affect **GoogleDrive and SharePoint identically** — not just SFTP.

## 1. `deleteStale` was never exposed on the Drive resources

Every ingestion run ended with three `HTTP 405 Method Not Allowed`:

```
PUT /v1/drives/directories/deleteStale
PUT /v1/drives/files/deleteStale
PUT /v1/drives/spreadsheets/deleteStale
```

`PUT /v1/tables/deleteStale` returns 400 (route exists, body rejected) — 405 vs 400 is the tell. The
endpoint is declared on `TableResource`, `TopicResource`, `ContainerResource` and five others, but was
missing from all four Drive resources.

**Impact:** `markDeletedDirectories` / `markDeletedFiles` were silent no-ops — a file deleted at the
source stayed in the catalogue forever. It also made **every** run report Failed even when extraction
succeeded, so a real failure was indistinguishable from the permanent one.

**Fix:** declare the endpoint on `Directory`, `File`, `Spreadsheet` and `Worksheet`. No new logic —
`EntityResource.deleteStaleEntities` is fully generic, and `resolveScopeEntityType` already maps
`directory`/`file` to `driveService` via `ENTITY_SERVICE_TYPE_MAP`.

## 2. `FileRepository` never set column FQNs on the write path

The first ingestion of a service was clean; **every run after it** failed on each file with columns:

```
Cannot invoke "String.equals(Object)" because the return value of
"org.openmetadata.schema.type.Column.getFullyQualifiedName()" is null
```

Observed 0 errors on run 1, then 17 and 18 on runs 2 and 3, reproduced on three separate services and
through the CLI.

**Impact:** metadata was frozen at whatever the first run captured — a column added at the source
could never reach OpenMetadata.

**Cause:** `setColumnFQN` was called in `setFields` (the read path) but not in
`setFullyQualifiedName` (the write path), so columns were persisted with a null FQN and the update
comparison NPE'd. `WorksheetRepository` already does this correctly inside `setFullyQualifiedName`;
`FileRepository` now matches, guarded with a null check because non-tabular files (images, PDFs,
parquet) have no columns and `ColumnUtil.setColumnFQN` does not accept null.

## Testing

Found against a live SFTP server on release-2-0, driving real ingestion through both a hybrid runner
and the CLI. `mvn compile` and `spotless:apply` clean.

Full test notes: 17 bugs found across phases 0–4; these are the two blockers that need server-side
changes. The connector's own remaining issues are being fixed separately.


openmetadata-service/src/main/java/org/openmetadata/service/resources/drives/FileResource.java
```diff
@@ -326,6 +328,32 @@ public Response bulkCreateOrUpdate(
     return processBulkRequest(uriInfo, securityContext, createRequests, mapper, async);
   }
 
+  @PUT
```
mohityadav766: It should be @DELETE

mohityadav766 (commented):
Need changes

openmetadata-service/src/main/java/org/openmetadata/service/resources/drives/FileResource.java
mohityadav766: Done in cb23f131fb — but flipped on **all 18 resources**, not just the four Drive ones.

Flipping only Drive would have left them unreachable: `OpenMetadata.delete_stale_entities` is shared by every connector and sent `PUT`, so Drive would have gone straight back to `405`. So the verb moved everywhere the endpoint is declared (Table, Topic, Container, Database, DatabaseSchema, StoredProcedure, APICollection, APIEndpoint, Chart, Dashboard, DashboardDataModel, MlModel, Pipeline, SearchIndex + the 4 new Drive ones), along with both callers:

- `ometa_api.delete_stale_entities` → `client.delete(url, json=request)`; `REST.delete` gained a `json=` passthrough
- `BulkApi.deleteStaleRaw` → `.method("DELETE", …)` (`Builder.DELETE()` takes no body)
- `docs/generated/api-reference.md` regenerated

Two things worth flagging:

**It stays DELETE-with-a-body.** A scope can hold thousands of `seenFqns`, so the set can't move to the query string. That's legal and Jersey/Jetty handle it, but a body on DELETE is the kind of thing an aggressive proxy can strip. Nothing in our own path does, and rc1 is pre-GA so there's no compatibility cost to the change — just noting the reason PUT was probably picked originally.

**Old-server fallback still works.** On a server without the endpoint, `DELETE /v1/tables/deleteStale` now falls into `deleteById`, whose UUID path param rejects `"deleteStale"` with a `404` — exactly the status `delete_stale_entities` already keys on for the legacy paginate-and-diff path, so that behaviour is unchanged.

Routing is safe: Jersey sorts literal segments ahead of templates, so `/deleteStale` wins over `/{id}` — the same resolution 48 existing literal-vs-template sibling routes already rely on (`GET /v1/tables/entityRelationship` vs `GET /v1/tables/{id}`, etc.). `BulkDeleteStaleIT` exercises it end to end and would 404 loudly if it went the other way.

Verified: `mvn compile` + `test-compile` + `spotless:check` clean on both modules; confirmed `requests` preserves the JSON body across a DELETE against a local server. The Python unit tests (`test_delete_stale.py`, updated to assert on `client.delete`) I did not run locally — this worktree has no venv and they need the full `make generate` + `install_dev` bootstrap; CI covers them.

IceS2 (approved):

PR #31380, opened 2026-08-12
Fixes #31379: stop one orphaned test case from 404ing the whole test case listing

### LINKED ISSUE #31379: Data Quality test case listing returns 404 when any listed test case points at a deleted table
### Affects

`main` and `2.0` (the relevant code is byte-identical on both).

### What happened

`GET /api/v1/dataQuality/testCases/search/list` answers `404` for the **whole page** whenever any single row on it references a table that no longer exists:

```
GET /api/v1/dataQuality/testCases/search/list?sortType=desc&sortField=testCaseResult.timestamp\
&limit=15&includeAllTests=true&fields=testCaseResult,testSuite,incidentId,incidentStatus&offset=0

{
  "code": 404,
  "message": "Entity not found: table <service>.<database>.<schema>.<table>"
}
```

This surfaced as a `DataQuality.spec` Playwright failure on the 2.0 AUT, but it is not test-only: the request has no `entityLink` filter, so it is the **global** Data Quality page. One leftover test case anywhere in the instance takes that page down for every user until the row is swept.

### Why a list endpoint 404s on one row

`PIIMasker.getTestCases` runs over every test case listing response and resolves the linked table once per row to decide column masking:

```java
table = Entity.getEntityByName(
    Entity.TABLE, testCaseLink.getEntityFQN(), "owners,tags,columns", Include.ALL);
```

`EntityNotFoundException` is allowed to escape, so a single dangling `entityLink` fails the entire response rather than that one row.

Orphaned test cases are a **designed steady state** in this codebase, not a corrupt-data edge case — `DataRetention` reaps them via `OrphanTestCaseCleanup` / `OrphanTestCaseRelationshipCleanup`, and `EntityRepository.setFieldsForDelete` explicitly documents tolerating "an orphaned test case whose table and testDefinition were hard-deleted out from under it". Test case rows legitimately outlive their table between the delete and the next retention run, so every read path over test cases has to tolerate a missing table.

### Why orphans are being produced in the first place

`EntityRepository.bulkEntitySpecificCleanup` loops the **no-arg** overload:

```java
protected void bulkEntitySpecificCleanup(List<T> entities) {
  for (T entity : entities) {
    entitySpecificCleanup(entity);
  }
}
```

but `TableRepository` overrides only the **`deletedBy`-aware** variant:

```java
@Override
protected void entitySpecificCleanup(String deletedBy, Table table) {
  deleteResidualTestCases(table, deletedBy);
  deleteResidualExecutableTestSuite(table, deletedBy);
}
```

That override is the safety net added in #28159 *specifically* to stop orphaned test cases from breaking listing. The no-arg base is an empty no-op, so the sweep never runs on the bulk path.

It regressed when #29322 set `descendantsCoveredByAncestorCascade = true` on `TableRepository` and routed recursive deletes through `bulkHardDeleteSubtree`. Any table hard-deleted as part of an ancestor cascade — e.g. `DELETE /api/v1/services/databaseServices/name/{fqn}?recursive=true&hardDelete=true`, which is what Playwright's `TableClass.delete()` issues — has skipped the sweep ever since.

### Steps to reproduce

1. Create a database service with a table, and add a test case to that table.
2. Hard-delete the service recursively: `DELETE /api/v1/services/databaseServices/name/{fqn}?recursive=true&hardDelete=true`.
3. Before `DataRetention` next runs, open the Data Quality page (or `GET /api/v1/dataQuality/testCases/search/list`).

**Expected:** the page lists the surviving test cases.
**Actual:** `404 Entity not found: table ...` for the entire page.

### Notes

Fixing only the delete path is not sufficient — orphans are expected to exist transiently by design, so the listing must degrade gracefully regardless. Both need fixing.

mohityadav766:
### Describe your changes:

Fixes #31379

`GET /api/v1/dataQuality/testCases/search/list` answers `404 Entity not found: table <fqn>` for the **whole page** whenever any single row on it points at a table that no longer exists. Reported as a `DataQuality.spec` failure on the 2.0 AUT, but the request carries no `entityLink` filter — it is the global Data Quality page, so one leftover row anywhere in the instance takes that page down for every user.

Two independent defects combine to produce it, and both are fixed here.

**1. `PIIMasker.getTestCases` rethrows on a dangling `entityLink`.** It resolves `Entity.getEntityByName(TABLE, ...)` once per row to decide column masking and lets `EntityNotFoundException` escape, failing the entire response instead of that one row. Orphaned test cases are a *designed* steady state — `DataRetention` reaps them via `OrphanTestCaseCleanup`, and `EntityRepository.setFieldsForDelete` already documents tolerating "an orphaned test case whose table and testDefinition were hard-deleted out from under it" — so a read path over test cases has to tolerate a missing table.

Resolve the table leniently, and when it is gone fall back to the caller's PII authorization rather than exposing the orphan's results: the column tags that would have decided masking died with the table, so masking is the safe default for a caller not authorized for PII.

**2. `EntityRepository.bulkEntitySpecificCleanup` dispatched to the wrong overload.** It looped the no-arg `entitySpecificCleanup(entity)`, but `TableRepository` overrides only `entitySpecificCleanup(String deletedBy, Table)` — its residual test case / test suite sweep, added in #28159 precisely to stop orphaned test cases from breaking listing. The no-arg base is an empty no-op, so that safety net has been dead for every table hard-deleted through an ancestor cascade since #29322 set `descendantsCoveredByAncestorCascade` on `TableRepository` and routed recursive deletes through `bulkHardDeleteSubtree`. Playwright's `TableClass.delete()` deletes the *databaseService* with `recursive=true&hardDelete=true`, which is exactly that path.

Dispatch to the `deletedBy`-aware overload so subclass hooks actually run. On a healthy cascade the two extra indexed lookups per table return empty, because the `table -> test suite -> test case` CONTAINS chain has already removed them; the sweep only does work when that chain is broken, which is what it was written for.

Fixing only the delete path would not be sufficient — orphans are expected to exist transiently by design, so the listing has to degrade gracefully regardless.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change, 2 production files.

#
#### Use cases covered
- A user opens the Data Quality page while an orphaned test case (table hard-deleted, not yet swept by `DataRetention`) exists anywhere in the instance — the page lists the surviving test cases instead of returning 404.
- A user not authorized for PII sees an orphaned test case masked rather than having its results, parameter values and description exposed.
- An admin hard-deletes a database service recursively — `TableRepository`'s residual test case / test suite sweep now runs for each contained table, so the cascade stops leaving orphans behind.

#### Unit tests

Both were verified **failing before the fix and passing after** (behaviour reverted with the signatures kept, so the RED run is a real behavioural failure and not a compile error):

```
[ERROR] PIIMaskerTest.getTestCasesTolerateTestCasesWhoseTableWasHardDeleted:277
        » EntityNotFound Entity not found: table service.db.schema.deleted_orders
[ERROR] EntityRepositoryRestoreTest.bulkHardDeleteSubtree_callsBulkEntitySpecificCleanupAndAdditionalChildrenHooks:523
        expected: <[user, user]> but was: <[]>
```

The first reproduces the reported production error string exactly. After the fix:

```
[INFO] Tests run: 9,  Failures: 0, Errors: 0, Skipped: 0 -- PIIMaskerTest
[INFO] Tests run: 17, Failures: 0, Errors: 0, Skipped: 0 -- EntityRepositoryRestoreTest
[INFO] Tests run: 26, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

`mvn spotless:check -pl openmetadata-service` clean. Re-verified after rebasing onto latest `main`.

#### Integration tests
Not added. The dispatch fix is covered at the unit level; a service-delete → no-orphan-test-cases assertion would fit naturally in `BaseServiceIT.recursiveHardDelete_serviceSubtree_leavesNoOrphansAndSearchClean` (added by #29322) and I am happy to add it if reviewers want that coverage.

#### Playwright tests
Not added — this restores existing `DataQuality.spec` coverage rather than adding a new user-facing surface. The spec creates and hard-deletes several tables across `describe` blocks while other blocks load the global Data Quality page, which is what surfaced the bug.

#
### Backport

The relevant code is **byte-identical on `main` and `2.0`**, so this applies to both. Raising against `main` only; happy to cherry-pick to `2.0` on request.


mohityadav766:
Added the integration test that was flagged as missing: `TestCaseDeleteResilienceIT.recursiveServiceHardDelete_withBrokenSuiteChain_sweepsResidualTestCase`.

It stages the exact state `TableRepository`'s residual sweep exists for — it drops the basic `testSuite -> testCase` CONTAINS edge so the cascade's subtree walk cannot reach the test case, leaving it reachable only through its `entityLink` — then hard-deletes the owning database service recursively. It drops that edge rather than the `table -> testSuite` one so the cascade still reaps the basic suite and never leaves one whose table is gone, which would break concurrent `GET /testSuites` listings.

Verified against a real stack (Postgres + OpenSearch), reverting only the dispatch line with the signature intact so the RED is behavioural and not a compile error:

```
without the fix   ConditionTimeout ... test_case row removed ==> expected: <0> but was: <1>
with the fix      Tests run: 6, Failures: 0, Errors: 0, Skipped: 0  (1.8s)   BUILD SUCCESS
```

**One thing worth recording, because it nearly went the other way.** My first draft also asserted the listing side here, by deleting the table row through the DAO and then calling `search/list`. It passed against unfixed code — a raw row delete leaves the entity caches warm, so `getEntityByName` never threw and the test proved nothing. I caught it because the same run exercised a stale `openmetadata-service` jar, which made the sweep test fail while the listing test passed; that split is what exposed the false pass.

I removed it rather than ship a green test that asserts nothing. Staging a genuine orphan at IT level needs the table gone from those caches too, and once this sweep runs no API sequence leaves one behind — which is precisely the point of fix (2). The listing's tolerance for a dangling `entityLink` stays covered by `PIIMaskerTest.getTestCasesTolerateTestCasesWhoseTableWasHardDeleted`, which reproduces the reported error string exactly and was verified failing before the fix.

yan-3005 (approved):

PR #31388, opened 2026-08-12
Fixes #31376: ingest dbt source freshness results from sources.json

### LINKED ISSUE #31376: dbt source freshness ingestion fails for sources.json v3: Results1 has no message field
## Connector

dbt

## Feature area

Profiler / Data Quality

## Describe the bug

Ingesting dbt source-freshness results from a valid `sources.json` v3 artifact fails while processing the freshness test result. The production parser returns `collate_dbt_artifacts_parser.parsers.sources.sources_v3.Results1`, which does not define a `message` field, but `DbtSource.add_dbt_test_result()` accesses `dbt_test_result.message` directly.

The resulting exception is caught by the ingestion code, so the workflow may continue while the source-freshness result is silently omitted from OpenMetadata.

This report contains a reproduced and commit-traced RCA. Maintainers should independently validate the conclusion across supported dbt artifact versions and freshness statuses before implementing or merging a fix.

## Actual behavior

Given a valid dbt `sources.json` v3 freshness result with `status: pass`, OpenMetadata logs:

```text
AttributeError: 'Results1' object has no attribute 'message'

Failed to capture tests results for node: <source>_freshness
'Results1' object has no attribute 'message'
```

No freshness `TestCaseResult` is written to OpenMetadata.

The parsed result has these fields:

```text
adapter_response
criteria
execution_time
max_loaded_at
max_loaded_at_time_ago_in_s
snapshotted_at
status
thread_id
timing
unique_id
```

It has no `message` field, consistent with the dbt `sources.json` v3 schema.

## Expected behavior

- A valid source-freshness result should be ingested without requiring a `message` field.
- `pass` should create a successful `TestCaseResult` with its execution timestamp.
- `warn`, `error`, and runtime-error freshness results should be mapped according to OpenMetadata's intended status semantics without raising an attribute error.
- The existing protection against compiled-only `run_results.json` entries (`status=success`, `message=null`) should remain intact.

## Steps to reproduce

1. Configure a dbt source with freshness criteria in `sources.yml`.
2. Run:

   ```bash
   dbt source freshness
   ```

3. Confirm `target/sources.json` declares schema v3 and contains at least one freshness result.
4. Configure dbt ingestion with the generated manifest and sources artifact:

   ```yaml
   dbtConfigSource:
     dbtConfigType: local
     dbtManifestFilePath: /path/to/target/manifest.json
     dbtSourcesFilePath: /path/to/target/sources.json
   ```

5. Run the dbt ingestion workflow.
6. Observe the `Results1.message` exception and the missing freshness result in OpenMetadata.

## Validated reproduction

The issue was reproduced with:

- dbt Core 1.11.7
- `https://schemas.getdbt.com/dbt/sources/v3.json`
- Python 3.12
- Snowflake-backed dbt project
- The production `parse_sources()` parser
- The real `DbtSource.add_dbt_test_result()` processing path

Minimal validation:

```python
freshness_result = parse_sources(sources_payload).results[0]

assert freshness_result.status.value == "pass"
assert not hasattr(freshness_result, "message")

# Current production path logs AttributeError here.
source.add_dbt_test_result(dbt_test_with(freshness_result))
```

### Original source-freshness support

PR #18730 added source-freshness ingestion and placed parsed `sources.json` results into the same processing collection used by normal dbt test results. At that time, `add_dbt_test_result()` did not read `.message`, so the differing artifact shapes were compatible.

### Regression introduction

PR #26812, merge commit `964ac7578271e3df6013691305d53dca2371aba5`, added this compiled-only guard:

```python
if not dbt_test_result.message:
    return
```

That guard was intended for `run_results.json`, where compiled-only nodes can have `status=success` and `message=null`. However, it was added to the shared handler without narrowing it to run-result objects. Every `sources.json` v3 freshness result therefore attempts to access an attribute that does not exist.

Git history validation:

- The parent of #26812 contains no `dbt_test_result.message` access in this handler.
- `git log -S 'if not dbt_test_result.message'` identifies #26812 as the first introduction of the failing guard.
- `1.12.8-release` does not contain the merge commit.
- `1.13.0-release` does contain the merge commit.

This supports #26812 as the regression point, but maintainers should independently validate this conclusion against the release build and supported parser package matrix.

### Related changes that do not fully fix this case

PR #29828 narrows the guard to `message` being empty and `status == success`:

```python
if not dbt_test_result.message and dbt_test_result.status.value == "success":
```

This fixes executed `run_results.json` entries with `status=pass, message=null`, but it still evaluates the missing `.message` attribute first. A `sources_v3.Results1` object therefore raises the same exception.

PR #29650 introduced another direct `.message` access when populating failure details. Once the first guard is made safe, non-success freshness results should also be checked against this second access.

Open PR #31138 centralizes the compiled-only predicate but currently still reads `dbt_test_result.message` directly, so it does not appear to address the `sources.json` object shape.

## Why existing tests did not catch it

- The #26812 tests construct mocked run-result objects that always expose `.message`.
- They do not parse a real `sources.json` fixture.
- The #29828 parser-backed tests cover `run_results.json`, not `sources.json`.
- The original source-freshness tests do not exercise a parsed freshness result through `add_dbt_test_result()`.

## Suggested fix direction

Normalize optional fields before applying run-result-specific rules:

```python
result_status = dbt_test_result.status.value
has_message_field = hasattr(dbt_test_result, "message")
result_message = getattr(dbt_test_result, "message", None)

if (
    has_message_field
    and not result_message
    and result_status == DbtTestSuccessEnum.SUCCESS.value
):
    # Compiled-only run_results.json entry.
    return
```

All later accesses should use `result_message` rather than directly reading `.message`.

This is a suggested direction, not a prescribed implementation. The engineer should validate whether explicit result-type normalization or a shared internal result protocol is more appropriate.

## Acceptance criteria

- Add a parser-backed `sources.json` v3 fixture test.
- Freshness `pass` without `message` is ingested as Success.
- Freshness `warn`, `error`, and runtime error without `message` do not crash and map to the agreed statuses.
- `run_results.json` `status=success, message=null` remains skipped as compiled-only.
- Executed `run_results.json` `status=pass, message=null` remains ingested.
- Failed/warned normal dbt tests with a populated message retain their result details.
- A local end-to-end ingestion using `manifest.json` plus `sources.json` creates a visible freshness result in OpenMetadata.

## Related issues and PRs

- #18730 — added dbt freshness checks
- #26812 — probable regression introduction
- #29650 — populates test-result failure messages
- #29824 / #29828 — related null-message regression for `run_results.json`
- #31138 — related open PR for multiple run-results artifacts
- #24538 — general dbt source-freshness integration tracker

## Pre-submission checklist

- [x] I searched existing issues and PRs for `Results1`, `sources.json`, source freshness, and `dbt_test_result.message`.
- [x] I removed credentials, hostnames, emails, and other sensitive data.
- [x] The behavior was reproduced with a real parser-backed dbt artifact.
- [x] The RCA is presented as validated evidence with an explicit request for maintainer revalidation.

ulixius9:
### Describe your changes:

Fixes #31376

I worked on **dbt source freshness ingestion** because every freshness result has been failing with `AttributeError: 'Results1' object has no attribute 'message'` since 1.13.0. `sources.json` results are a different artifact shape from `run_results.json` — neither the executed variant (`Results1`) nor the runtime-error variant (`Results`) has a `message` field, and the runtime-error variant has no `timing` either — but both flow through `add_dbt_test_result()`, whose compiled-only guard from #26812 reads `.message` unconditionally. While fixing that I found a second, independent break: a dbt source node has no `depends_on`, so resolving upstreams the way a test node does always returned `[]`, leaving the freshness test case with no table to attach to — meaning freshness results have never actually reached OpenMetadata. Both are fixed here.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

**Artifact-shape dispatch.** `add_dbt_test_result()` is split into `_build_freshness_test_case_result` and `_build_run_result_test_case_result`, sharing a `_map_dbt_test_status` / `_resolve_dbt_test_timestamp` pair and the existing FQN/emit tail. The run-results builder is a verbatim move — **the compiled-only guard expression is byte-identical**, so #26812/#29828 semantics are untouched and this does not collide with open PR #31138.

Dispatch is an explicit `DbtCommonEnum.IS_FRESHNESS` key set by the producer in `add_dbt_sources()`, not a `hasattr` check on the parsed object. Rejected `hasattr(manifest_node, "freshness")` because a `MagicMock` answers `True` to every `hasattr`, which would have silently rerouted every existing mocked run-results test through the freshness builder — the same blind spot that let #26812 ship green.

**Freshness semantics.** `sources.json` has no `message`, so `result` is populated from the actual staleness signal (`max_loaded_at_time_ago_in_s` vs `criteria.warn_after`/`error_after`), or the adapter error for runtime errors. A new `sources_generate_time` context value gives the runtime-error variant a timestamp fallback; without it the crash would simply be replaced by a silent drop at the unparseable-timestamp guard on any freshness-only run.

**Status mapping is unchanged** — `warn` and `runtime error` keep the pre-1.13.0 fall-through to `Aborted`, so this stays a pure regression fix. `TestCaseStatus` has no `Warning` value; adding one is a separate schema-first change.

**Upstream resolution.** `add_dbt_sources()` now builds the source's own table FQN via the same `fqn.build` construction `yield_data_models` uses (metadata.py:887), gated on `_get_table_entity`, so the freshness test attaches to the entity dbt ingestion already created. A source missing from OpenMetadata is warned and skipped rather than failing downstream. The now-unused `manifest_entities` parameter was dropped (single caller, no overrides).

#
#### Use cases covered
- `dbt source freshness` with status `pass` is ingested as a Success test case result
- Status `warn` and `error` are ingested with the staleness detail (age vs configured thresholds) in `result`
- Status `runtime error` (no `timing`, no `message`) is ingested with the adapter error, timestamped from `sources.json`'s `generated_at`
- A freshness test case resolves to an entity link on the source's own table, so it is created and its result is written
- A dbt source absent from OpenMetadata is skipped with a warning instead of failing
- `run_results.json` compiled-only entries (`status=success, message=null`) are still skipped; executed `status=pass, message=null` entries are still ingested

#### Unit tests
- Files added/updated: `ingestion/tests/unit/test_dbt.py` (new `TestAddDbtSourceFreshnessResults`, 10 tests), `ingestion/tests/unit/resources/dbt_ingest/sources_v3.json` (new parser-backed fixture covering all four freshness statuses)
- Coverage: **93.1% on the changed functions** (94/101 statements; `make unit_ingestion` equivalent run with `--cov=metadata.ingestion.source.database.dbt`). The 7 uncovered statements are pre-existing guard branches (doubly-unparseable timestamp, missing-result early return).
- **196 passed** across `test_dbt.py` + `test_dbt_ingest.py`.

Every freshness test drives the real `collate_dbt_artifacts_parser`, not a mock — #26812's tests passed only because their mocked results answered to any attribute. Two existing harnesses (`TestAddDbtTestResultSkipsCompiledOnly`, `TestAddDbtTestResultFailureReporting`) were switched from `MagicMock(spec=DbtSource)` to a real instance so their assertions exercise the builders instead of mock wiring.

- [x] Not applicable — covered by parser-backed unit tests against real `sources.json` v3 artifacts.

#### Playwright (UI) tests

#### Manual testing performed
1. Reproduced the reported failure against the real parser: all four freshness statuses raised `AttributeError` and were recorded as `status.failed`.
2. Verified the regression window — `git log -S 'if not dbt_test_result.message'` identifies #26812 (`964ac75782`); `git tag --contains` shows it in `1.13.0-release` onward and absent from 1.12.x.
3. RED/GREEN for both bugs: reverting the dispatch reproduces the exact reported `AttributeError` in 6 of 8 tests; forcing `UPSTREAM = []` (the old upstream resolution) fails the wiring test. Both pass after the fix.
4. Ran the full production path — real `manifest_v12.Sources` node → `add_dbt_sources()` → `add_dbt_test_result()` — and confirmed each status produces a test case FQN (`svc.RAW_DB.RAW.orders.orders_freshness`), an entity link, and a result, with `status.failed` never called.
5. `make py_format_check` clean; `basedpyright --baselinemode=discard` reports 0 errors on both changed source files with the baseline unmodified.

**Not done:** a live end-to-end run (real `dbt source freshness` against a warehouse → ingestion → freshness result visible in the UI). I did not have a stack for it. Two things worth confirming in that run: (a) `get_dbt_test_definition_name` falls back to the node name for source nodes, producing one `TestDefinition` per source table rather than one shared "freshness" definition — pre-existing and untouched here, but it contradicts the shared-per-type convention from #28927; (b) sources where dbt's `identifier` differs from `name` build an FQN from `get_dbt_model_name` (alias-or-name), consistent with `yield_data_models` but wrong for both together.

#

Not applicable.

#
### Checklist:
- [x] For JSON Schema changes: not applicable — no schema changes.
- [x] For UI changes: not applicable — no UI changes.

Bug fix
- [x] I have added a test that covers the exact scenario we are fixing. Issue #31376 is referenced in the `TestAddDbtSourceFreshnessResults` docstring for future reference.

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
@@ -688,7 +688,7 @@ class DbtSource(DbtServiceSource):
                 manifest_entities, manifest_node
             )
 
-    def add_dbt_sources(self, key: str, manifest_node, manifest_entities, dbt_objects: DbtObjects) -> None:
+    def add_dbt_sources(self, key: str, manifest_node, dbt_objects: DbtObjects) -> None:
         """
         Method to append dbt test cases based on sources file for later processing
         In dbt manifest sources node name is table/view name (not test name like with test nodes)
@@ -701,17 +701,34 @@ class DbtSource(DbtServiceSource):
             (item for item in dbt_objects.dbt_sources.results if item.unique_id == key),
             None,
         )
+        if not freshness_test_result:
+            return
 
-        if freshness_test_result:
-            upstream_nodes = self.parse_upstream_nodes_with_names(manifest_entities, manifest_node)
-            self.context.get().dbt_tests[key + "_freshness"] = {DbtCommonEnum.MANIFEST_NODE.value: manifest_node_new}
-            self.context.get().dbt_tests[key + "_freshness"][DbtCommonEnum.UPSTREAM.value] = [
-                node.fqn for node in upstream_nodes
-            ]
-            self.context.get().dbt_tests[key + "_freshness"][DbtCommonEnum.UPSTREAM_BY_NAME.value] = (
-                build_upstream_name_map(upstream_nodes)
-            )
-            self.context.get().dbt_tests[key + "_freshness"][DbtCommonEnum.RESULTS.value] = freshness_test_result
+        # A freshness check tests the source table itself. Sources are graph roots with no
+        # depends_on, so resolving upstreams the way a dbt test node does always yields an
+        # empty list and leaves the test case with no table to attach to.
+        source_fqn = fqn.build(
+            self.metadata,
+            entity_type=Table,
+            service_name=self.config.serviceName,
+            database_name=get_corrected_name(manifest_node.database),
+            schema_name=get_corrected_name(manifest_node.schema_),
+            table_name=get_dbt_model_name(manifest_node),
+        )
+        if not source_fqn or not self._get_table_entity(table_fqn=source_fqn):
+            logger.warning(f"Table entity not found for dbt source {key}, skipping its freshness test")
+            return
+
+        self.context.get().dbt_tests[key + "_freshness"] = {DbtCommonEnum.MANIFEST_NODE.value: manifest_node_new}
+        self.context.get().dbt_tests[key + "_freshness"][DbtCommonEnum.UPSTREAM.value] = [source_fqn]
+        self.context.get().dbt_tests[key + "_freshness"][DbtCommonEnum.UPSTREAM_BY_NAME.value] = (
+            build_upstream_name_map([build_upstream_node(manifest_node, source_fqn)])
+        )
+        self.context.get().dbt_tests[key + "_freshness"][DbtCommonEnum.RESULTS.value] = freshness_test_result
+        # sources.json results share the dbt_tests collection with run_results.json
+        # results but are a different artifact shape, so tag them at the producer
+        # instead of sniffing the parsed object downstream
+        self.context.get().dbt_tests[key + "_freshness"][DbtCommonEnum.IS_FRESHNESS.value] = True
 
     def _get_table_entity(self, table_fqn) -> Optional[Table]:  # noqa: UP045
         def search_table(fqn_search_string: str) -> Optional[Table]:  # noqa: UP045
@@ -792,11 +809,14 @@ class DbtSource(DbtServiceSource):
             self.context.get().table_domains = {}
             self.context.get().table_custom_properties = {}
             self.context.get().run_results_generate_time = None
+            self.context.get().sources_generate_time = None
 
             # Since we'll be processing multiple run_results for a single project
             # we'll only consider the first run_results generated_at time
             if dbt_objects.dbt_run_results and dbt_objects.dbt_run_results[0].metadata.generated_at:
                 self.context.get().run_results_generate_time = dbt_objects.dbt_run_results[0].metadata.generated_at
+            if dbt_objects.dbt_sources and dbt_objects.dbt_sources.metadata.generated_at:
+                self.context.get().sources_generate_time = dbt_objects.dbt_sources.metadata.generated_at
             dbt_project_name = getattr(dbt_objects.dbt_manifest.metadata, "project_name", None)
             for key, manifest_node in manifest_entities.items():
                 try:
@@ -820,7 +840,6 @@ class DbtSource(DbtServiceSource):
                         self.add_dbt_sources(
                             key,
                             manifest_node=manifest_node,
-                            manifest_entities=manifest_entities,
                             dbt_objects=dbt_objects,
                         )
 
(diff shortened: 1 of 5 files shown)


ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
```diff
@@ -701,17 +701,34 @@ def add_dbt_sources(self, key: str, manifest_node, manifest_entities, dbt_object
...
+        # A freshness check tests the source table itself. Sources are graph roots with no
+        # depends_on, so resolving upstreams the way a dbt test node does always yields an
+        # empty list and leaves the test case with no table to attach to.
+        source_fqn = fqn.build(
+            self.metadata,
+            entity_type=Table,
+            service_name=self.config.serviceName,
+            database_name=get_corrected_name(manifest_node.database),
+            schema_name=get_corrected_name(manifest_node.schema_),
+            table_name=get_dbt_model_name(manifest_node),
```
ayush-shah: Please resolve a dbt source with its physical identifier, falling back to name, before building the table FQN. get_dbt_model_name() resolves alias-or-name, but parsed source nodes have identifier and no alias. I reproduced this with name=logical_orders and identifier=physical_orders: the code looked up snowflake_svc.RAW_DB.RAW.logical_orders, logged that the table was missing, and queued zero freshness tests, even though the warehouse table was ...RAW.physical_orders. This is a valid and common dbt configuration: https://docs.getdbt.com/reference/resource-properties/identifier. Please add a parser-backed regression test where identifier differs from name.

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
```diff
@@ -701,17 +701,34 @@ def add_dbt_sources(self, key: str, manifest_node, manifest_entities, dbt_object
...
+        )
+        if not source_fqn or not self._get_table_entity(table_fqn=source_fqn):
```
ayush-shah: Please retain and use the canonical FQN from the Table returned by _get_table_entity(). The lookup can find a table under another service when searchAcrossDatabases is enabled, but this code currently uses the returned entity only as a boolean and then queues the original source_fqn. I reproduced a lookup returning actual_svc.RAW_DB.RAW.orders while UPSTREAM was still configured_svc.RAW_DB.RAW.orders, so TestCase creation and result attachment target the wrong entity. Assign the returned Table and use model_str(table_entity.fullyQualifiedName) for UPSTREAM and the entity link. A test covering cross-service resolution would prevent this regression.

ayush-shah:
Independent validation summary: the artifact-shape split fixes the reported sources.json v3 Results1.message crash. I ran the focused parser/result suite (21 passed), Ruff lint/format, and our real dbt 1.11.7 sources.json artifact; the parsed Results1 has no message field and this change produced OpenMetadata Success, value 1, with the exact execute timestamp. The compiled-only success/null boundary and executed pass/null behavior also remained intact. Before merge, the issue acceptance criteria still need one live OpenMetadata E2E run: ingest manifest.json plus sources.json, verify the freshness TestDefinition/TestCase/result through the API and UI, and rerun to confirm idempotency. The PR description currently marks that live E2E as not done. Also, this PR targets main; applying it directly to the current 2.0 branch produces a metadata.py conflict, so release 2.0 will need a deliberate backport after the inline correctness findings are resolved.

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
ulixius9: Confirmed and fixed in 4f214da. You're right on both counts — and the reproduction was exact.

Added `get_source_physical_name()` in `dbt_utils.py`, which prefers `identifier` and falls back to `get_dbt_model_name()`; `add_dbt_sources` now builds the FQN with it. Verified the underlying cause: source nodes have no `alias` field at all, so `get_dbt_model_name` fell straight through to `name`:

```
has alias attr : False
identifier     : physical_orders
get_dbt_model_name -> logical_orders   <-- FQN was built from this
```

Regression test added: `test_source_fqn_uses_the_physical_identifier_not_the_logical_name`, parser-backed with `name=logical_orders` / `identifier=physical_orders`. Reverting the fix makes it fail with "freshness test was dropped - the FQN used the logical name". The test case keeps its logical name (`logical_orders_freshness`) — only the table FQN uses the identifier.

Scoped the change to source nodes rather than touching `get_dbt_model_name` itself, since models legitimately resolve through `alias` and that helper is used across the connector.

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
ulixius9: Confirmed and fixed in 4f214da. `_get_table_entity` retries with `service_name="*"` when `searchAcrossDatabases` is on (metadata.py:757-770), so it can legitimately return a table under another service — and I was discarding that and queueing the FQN I had guessed.

Now binds the entity and takes its canonical FQN:

```python
table_entity = self._get_table_entity(table_fqn=source_fqn) if source_fqn else None
if not table_entity:
    logger.warning(...); return
source_fqn = model_str(table_entity.fullyQualifiedName)
```

Regression test added: `test_source_fqn_follows_the_entity_found_across_services` — configured service `configured_svc`, table resolving to `actual_svc.RAW_DB.RAW.orders`, asserting both UPSTREAM and the resulting test case FQN follow the entity. Reverting reproduces your exact symptom: `assert ['configured_svc...'] == ['actual_svc...']`.

One thing worth flagging for anyone writing similar tests here: both regression tests were initially **vacuous**. `fqn.build` for a `Table` runs its own ES lookup via `search_table_from_es`, so a mock with a flat `return_value` echoes the mocked entity back for whatever table name it was asked — the tests passed even with both bugs reintroduced. They now use a search-string-aware `es_search_from_fqn` (`_es_lookup`) that only resolves FQNs the warehouse actually has, which is what makes the RED checks above real.

ingestion/src/metadata/ingestion/source/database/dbt/metadata.py
```diff
@@ -1966,7 +1985,153 @@ def patch_dbt_test_case_description(self, test_case: TestCase, description: Opti
                 force=bool(self.source_config.dbtUpdateDescriptions),
             )
 
-    def add_dbt_test_result(self, dbt_test: dict):  # noqa: C901
+    @staticmethod
+    def _map_dbt_test_status(status_value: str) -> tuple[TestCaseStatus, int]:
```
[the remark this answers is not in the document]
ulixius9: Leaving this as `tuple[TestCaseStatus, int]`.

The Python 3.8 import-time concern doesn't apply — the ingestion package declares `requires-python = ">=3.10"` and ruff is configured with `target-version = "py310"` (`ingestion/pyproject.toml:16,85`), so PEP 585 subscription is always available.

On consistency: the `List`/`Optional` in this file carry `# noqa: UP006` / `# noqa: UP045`, and those suppressions exist precisely because ruff's pyupgrade rules want the modern builtin form. `tuple[...]` is what the linter prefers and needs no suppression; switching to `typing.Tuple` would mean adding a `# noqa: UP006`. The surrounding style is the legacy one being suppressed, not the target.

ayush-shah (approved):

PR #31402, opened 2026-08-12
Fixes 31401: Handle unnamed and quoted ROW fields in Trino reflection

### LINKED ISSUE #31401: Trino: tables with unnamed ROW fields ingest with zero columns, and struct children are named with literal quotes
## Description

Two defects in the Trino connector's ROW/struct type parsing (`ingestion/src/metadata/ingestion/source/database/trino/metadata.py`). Both reproduced against a live **Trino 483** instance.

### 1. Tables with unnamed ROW fields are ingested with zero columns

Trino allows ROW field names to be omitted, which is what any CTAS over a `ROW(...)` constructor produces:

```sql
CREATE TABLE memory.omtest.anon_struct AS
SELECT 1 AS id, ROW(1, 'x') AS s, ARRAY[ROW(1, 2.0)] AS arr;
```

`SHOW COLUMNS` then reports `row(integer, varchar(1))` and `array(row(integer, decimal(2,1)))`.

`parse_row_data_type` unpacks every field as `name type`:

```python
attr_name, attr_type_str = datatype.aware_split(data_type.strip(), delimiter=" ", maxsplit=1)
```

With no name there is only one token, so this raises `ValueError: not enough values to unpack (expected 2, got 1)`. `sql_column_handler.py` catches it and falls back to `columns = []`, so **the table is published with no columns at all**, logged only as a WARNING while the pipeline still reports success.

The driver's own `trino.sqlalchemy.datatype.parse_sqltype` has the same defect and is called first, so it fails on the same input.

Observed end-to-end:

```
omtest.named_struct: 2 column(s)
omtest.anon_struct:  0 column(s)   <-- all columns lost
```

### 2. Struct child columns are named with literal double quotes

Trino renders named ROW fields **quoted**, and preserves their case:

```
SHOW COLUMNS -> row("a" bigint, "b" varchar)
SHOW COLUMNS -> row("MyField" bigint)
```

The parser copies the name verbatim, producing `system_data_type = struct<"a":bigint,"b":varchar>`, and the resulting OpenMetadata children are named `"a"` / `"b"` instead of `a` / `b`:

```
column s  dataType=STRUCT  dataTypeDisplay='struct<"a":bigint,"b":varchar>'  children=['"a"', '"b"']
```

This breaks nested-column FQNs, column-level lineage matching, and nested-column test cases / profiling. The type string is additionally lower-cased wholesale, so `row("MyField" bigint)` becomes `myfield`.

## Why this was not caught

Every fixture in `ingestion/tests/unit/test_trino_complex_types.py` uses an **unquoted** form (`row(a int, b string)`) that Trino never actually emits, so the suite passes while production is wrong.

## Expected behaviour

- A table with unnamed ROW fields ingests all of its columns; unnamed fields get a stable positional name.
- Named struct children are `a` / `MyField`, not `"a"` / `"myfield"`.

## Environment

- Trino 483 (`trinodb/trino:latest`), `trino[sqlalchemy]`, SQLAlchemy 2.0.51

harshsoni2024:
### Describe your changes:

Fixes #31401

I fixed two defects in the Trino ROW/struct type parser, both reproduced and then verified against a live **Trino 483** instance.

**1. Tables with unnamed ROW fields lost every column.** Trino allows a ROW field name to be omitted — `CREATE TABLE t AS SELECT ROW(1,'x') AS s` reports `row(integer, varchar(1))`, which any CTAS over a `ROW(...)` constructor produces. `parse_row_data_type` unpacked each field as `name type` and raised `ValueError`. `sql_column_handler` swallows that and falls back to an empty column list, so the table was published with **zero columns** behind a WARNING while the pipeline still reported success.

The driver's own `datatype.parse_sqltype` has the same defect and runs first, so it is guarded as well. Since the OpenMetadata data type is derived from `system_data_type` for complex columns, falling back to `NULLTYPE` there keeps the column and its children intact. Unnamed fields are named positionally to match Trino's 1-based field access, so `s[1]` reads as `field1`.

**2. Struct child columns were named with literal quotes.** Trino renders named fields quoted (`row("a" bigint)`, escaping an embedded quote as `""`), so children were `"a"` instead of `a` — breaking nested-column FQNs and column-level lineage matching. The type string is also no longer lower-cased wholesale, which preserves the case of quoted field names in `dataTypeDisplay`.

A leading quote is what distinguishes a named field from an unnamed one whose type itself contains spaces (`timestamp(3) with time zone`, `interval day to second`); both are now covered.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change, 1 source file.

#
#### Use cases covered
- A Trino table with an unnamed ROW field (`row(integer, varchar(1))`) ingests **all** of its columns instead of none.
- A ROW mixing named and unnamed fields (`row(bigint, "b" bigint)`) ingests as `struct<field1:bigint,b:bigint>`.
- Named struct children are `a` / `b`, not `"a"` / `"b"`.
- Unnamed fields whose type contains spaces (`timestamp(3) with time zone`, `interval day to second`) are not misread as a field named `timestamp(3)`.
- Field names containing an escaped quote (`row("we""ird" bigint)`) resolve to `we"ird`.
- Quoted field-name case is preserved in `dataTypeDisplay` (`struct<MyField:bigint>`).
- Existing unquoted/legacy type strings continue to parse unchanged.

#### Unit tests
- Files added/updated:
  - `ingestion/tests/unit/source/database/trino/test_complex_types.py` (new, pytest style) — 24 cases across `split_row_field`, `parse_row_data_type`, and `_get_columns`, including explicit regression tests that a table keeps all its columns and that children are unquoted.
  - `ingestion/tests/unit/test_trino_complex_types.py` — added 14 fixtures that are **verbatim `SHOW COLUMNS` output captured from Trino 483**. The pre-existing fixtures used an unquoted form (`row(a int, b string)`) that Trino never emits, which is why this went unnoticed; those are retained for backward compatibility.

Result: `61 passed` across the Trino unit tests (35 before this PR).

- [x] Not applicable — covered by unit tests against captured real-Trino output. The existing `ingestion/tests/integration/trino/` suite requires the full OM stack and was not extended; happy to add a case there if reviewers prefer.

#### Playwright (UI) tests

#### Manual testing performed
1. Ran `trinodb/trino:latest` (Trino 483) and created 15 tables covering named / unnamed / nested / array-of-row / map-in-row / multi-word-type / escaped-quote / mixed struct shapes.
2. Captured real `SHOW COLUMNS` output for each and drove them through the real dialect override plus `SqlColumnHandlerMixin.get_columns_and_constraints` to get actual OM `Column` models.
3. Before the fix: `anon_struct` → **0 columns**; `named_struct` → children `['"a"', '"b"']`.
4. After the fix: all 15 tables ingest their columns, e.g.

```
anon_simple        struct<field1:integer,field2:varchar(1)>              children=['field1', 'field2']
anon_and_named     struct<field1:bigint,b:bigint>                        children=['field1', 'b']
anon_tstz          struct<field1:timestamp(3) with time zone>            children=['field1']
array_of_anon      array<struct<field1:integer,field2:decimal(2,1)>>     children=['field1', 'field2']
named_simple       struct<a:bigint,b:varchar>                            children=['a', 'b']
nested_named       struct<outer:struct<inner:bigint>>                    children=['outer']
named_with_quote   struct<we"ird:bigint>                                 children=['we"ird']
```

5. `ruff check` + `ruff format --check` clean on all changed files.

#

Not applicable.

#
### Checklist:

**Bug fix**
- [x] I have added a test that covers the exact scenario we are fixing (see the `test_anonymous_row_still_yields_all_columns` and `test_named_row_children_are_not_quoted` regression tests, issue #31401).

#
### Note for reviewers

Two adjacent things I deliberately left **out of scope**:

1. **Child column names are still lower-cased and space-stripped** by the shared `ColumnTypeParser` (`column_type_parser.py:357-358` does `.lower().strip()` and `.replace(" ", "")`). So `row("MyField" bigint)` yields `dataTypeDisplay=struct<MyField:bigint>` (fixed here) but a child named `myfield`. That normalization is generic and affects every connector with struct support, so changing it needs a broader review than this bug fix.
2. `get_type_name_and_opts` returns `sqltypes.NULLTYPE` (not a 2-tuple) when its regex does not match, so callers raise `TypeError: cannot unpack non-iterable NullType object`. Unrelated to this issue; happy to fix separately.

This PR updates Trino reflection to retain unnamed ROW fields and normalize quoted child names while preserving quoted-name case.
- Adds positional, collision-resistant names for unnamed ROW fields.
- Parses quoted and multi-word ROW field forms.
- Falls back to `NULLTYPE` when Trino's SQLAlchemy parser rejects a reflected type.
- Adds regression coverage using representative Trino `SHOW COLUMNS` output.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/source/database/trino/metadata.py | Adds ROW-field splitting, positional-name resolution, case-preserving complex-type parsing, and resilient SQLAlchemy type reflection; no eligible blocking follow-up issue was established. |
| ingestion/tests/unit/source/database/trino/test_complex_types.py | Adds focused regression tests for unnamed, quoted, nested, collision-prone, and multi-word ROW fields. |
| ingestion/tests/unit/test_trino_complex_types.py | Extends existing parser fixtures with captured Trino ROW and ARRAY type representations. |

</details>

<sub>Reviews (4): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/trino-row-t..."](https://github.com/open-metadata/openmetadata/commit/a17c57b3c23485d04d02d47899417e0fa78f62c2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52560960)</sub>

**Context used:**


harshsoni2024:
> **Edge Case: Positional field name can collide with an explicit `fieldN` name**

Good catch — this is real and reachable, so I've fixed it in 64e1165.

I checked it against Trino 483 rather than assuming, and it's a bit less unlikely than it reads. Trino accepts the DDL:

```sql
CREATE TABLE memory.col.mix_field1 AS SELECT CAST(ROW(1,'x') AS row(bigint, "field1" varchar)) AS s;
-- SHOW COLUMNS -> row(bigint, "field1" varchar)
```

and it propagated all the way through with no validation error anywhere in the ingestion path:

```
mix_field1: children=['field1', 'field1']  duplicate=True
mix_field2: children=['field2', 'field2']  duplicate=True
```

(`mix_field2` is `row("field2" varchar, bigint)` — worth noting the collision doesn't need the user to pick an adversarial name at the *same* position; any struct with a field named `fieldN` plus an unnamed sibling anywhere can hit it, and `field1`/`field2` are not exotic names in real data.)

### Fix

Naming moved out of `split_row_field` and into a new `resolve_field_names`, because uniqueness can't be decided one field at a time:

- `split_row_field` now returns `None` for an unnamed field, which also mirrors how the driver's own `ROW` type models it (`List[Tuple[Optional[str], SQLType]]`).
- `resolve_field_names` sees the whole ROW, collects the explicit names first, then suffixes each positional name until it's unique. Scope is one ROW level, since nested rows are separate child namespaces.

```
row(bigint, "field1" varchar)  ->  struct<field1_1:bigint,field1:varchar>
row("field2" varchar, bigint)  ->  struct<field2:varchar,field2_1:bigint>
```

The loop also dodges an explicit `fieldN_1`, so `row(bigint, "field1" varchar, "field1_1" varchar)` resolves to `field1_2`, `field1`, `field1_1` — it can't be defeated by adding more collisions.

### Tests

Added `TestResolveFieldNames` (4 cases, each asserting no duplicates) and two `parse_row_data_type` cases using the verbatim Trino output above. Re-ran the full end-to-end check over all 15 real-Trino struct shapes plus the two collision tables: **0 tables with zero columns or duplicate children**, no regressions. Trino unit tests now `67 passed` (61 before this comment, 35 before the PR).

mohittilala (approved):

PR #31410, opened 2026-08-12
Fixes #5526: harden multi-pod session, SSO and store-outage handling

mohityadav766:
### Describe your changes:

Fixes open-metadata/openmetadata-collate#5526

I worked on the open findings from the **2.0.0 multi-pod platform test run** (3 pods behind a
non-sticky LB, MySQL + Postgres, Redis and JDBC session stores, plus real Google / Okta / Auth0
tenants) because every one of them is a defect that only appears with more than one server pod or
against a real IdP — so a green single-pod PR run never surfaced them.

Fourteen of the sixteen open items are addressed here. Two are not, deliberately:

- **HR-26** (a relayed command that times out is not cancelled on the runner) lives in
  `io.collate.service.hybrid.relay.*` in the collate repo, not in this tree.
- **SM-34** (no guardrail for a mismatched `CACHE_PROVIDER` across pods) is addressed by the
  documentation half only. Real cross-pod detection needs a shared startup registry — i.e. a
  migration — because a JDBC-only pod has nowhere to look for its Redis-backed peers. The issue's
  acceptance criterion allows either detection *or* documentation.

One item needed no change: **NEW-7** (tampered `state` on `/callback` returned 500) was already
fixed on `main` by #31181 on 2026-08-11, six days after the test run. `/callback` now redirects to
`/signin`.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

**Session lifecycle — revoke where all callers already route.**
`UserRepository.postDelete` fires for both soft and hard delete and already invalidates bot tokens,
so session revocation goes there rather than in the resource layer; `SessionService` gained a
bounded `revokeSessionsForUser` built on the existing `findByUserIdAndStatus` + `revokeSession`
primitives. Revocation already notifies the WebSocket and cross-pod listeners, so peers drop the
user's sockets for free.

**Provider swap — validate, don't sweep.**
`UserSession` already records the provider that authenticated it, so `JwtFilter` rejects a session
whose provider is no longer the configured one. This needs no store query, no new endpoint, and no
cluster-wide sweep: each pod checks against its own current config, so it holds everywhere. The
alternative — enumerating and revoking every session on provider change — needs a new
"find all active sessions" store method on both backends and still races with the config reload.

**Cross-pod security config — reuse the bus that already exists.**
`CacheInvalidationPubSub` is already the cluster's invalidation channel (session revocation rides
it as `type=session, op=revoke`). `reloadSecuritySystem()` now splits into `reloadLocally()` +
`broadcastReload()`, and peers handle `type=securityConfig, op=reload` by invalidating
`SettingsCache` and reloading. Invalidate-then-reload ordering matters: the reload reads through
`SettingsCache`, so a stale entry would silently reload the old config. The remote path calls
`reloadLocally()` only, so there is no broadcast echo loop. When no bus is configured
(`cache.provider != redis`) the server logs a WARN naming the all-pods restart requirement, which
is the documented fallback.

**Store outage — one classifier, two transports.**
The API surface gets an `ExceptionMapper<RedisException>` → `503 + Retry-After`. The servlet auth
paths have no JAX-RS mapper, so their `catch (Exception) → 500 + e.getMessage()` catch-alls now
route through a single `SecurityUtil.writeFailureResponse` that maps a rejected credential to its
own status, a store outage to a retryable 503, and only the remainder to 500. That one helper also
fixes SM-20's secondary finding (login for a deleted user returned 500) and, as a side effect, the
same catch-all reporting a wrong password as 500 — `BasicAuthenticator` throws
`security.AuthenticationException`, which carries a 401 `Response` but is **not** a
`WebApplicationException`, so both branches are needed.

**SSO — the SPA has to be told, then has to listen.**
`ConfigResource.getAuthConfig` builds a fresh `AuthenticationConfiguration` and never copied
`responseType`, so the POJO's `id_token` default was served regardless of what was configured or
stored. Fixing the server alone is not enough: `getUserManagerConfig` never passed `response_type`
to oidc-client, which also defaults to implicit. Both halves are needed for a `code`-configured
public client to actually get code+PKCE.

**Backward compatibility.** Two changes tighten live behaviour and are worth calling out in review:
swapping `AUTHENTICATION_PROVIDER` now logs everyone out (previously old tokens stayed valid up to
7 days), and a WebSocket handshake with no session at all is refused on deployments that mint
server-side sessions. The second is gated on
`SecurityConfigurationManager.createsServerSideSessions()` so public-client OIDC deployments — which
have no server-side session to check — keep their existing behaviour. Everything else is either
additive or a wire-code correction.

#
#### Use cases covered

Security
- An admin soft-deletes a user → that user's existing session-bound tokens stop working
  immediately, on every pod, instead of at token expiry (7d default). Hard delete unchanged.
- A soft-deleted user attempting a fresh login gets `401`, not `500`.
- `aliceJWT + bobCookie` is rejected instead of resolving as alice. A token with no session claim
  (PAT, bot token, public-client id_token) is unaffected — there is nothing to bind it to.
- A WebSocket handshake with neither an `OM_SESSION` cookie nor a session-bound token is refused on
  session-based deployments; public-client deployments are untouched.
- After `AUTHENTICATION_PROVIDER` is swapped (e.g. Google → basic), tokens minted under the
  decommissioned provider are rejected on every pod.
- At `LOG_LEVEL=DEBUG` the logs no longer contain bearer tokens or `OM_SESSION` values.

Availability
- With Redis down, requests and logins answer `503` + `Retry-After: 5` and a generic message
  instead of `500 {"error":"Command timed out after 300 millisecond(s)"}`. Recovery stays
  unattended.

Multi-pod configuration
- An admin changes SSO config on pod A → pods B and C reload from the database, so a round-robin
  LB cannot serve a login from a pod still on the previous provider.
- With no invalidation bus, the update still applies locally and the server logs that every other
  pod needs a restart.

SSO
- `AUTHENTICATION_RESPONSE_TYPE=code` with `clientType=public` produces `response_type=code` +
  PKCE instead of `id_token` (which Google 400s as `unsupported_response_type`).
- Okta public clients no longer send `offline_access`, which Okta rejects outright per OAuth 2.1.
- `AUTHENTICATION_AUTHORITY` with Auth0's spec-conformant trailing slash no longer builds
  `.../oauth2//authorize` (Auth0 404).
- Clicking "Sign in" with an unreachable IdP shows an error and re-enables the page instead of
  spinning forever.

#### Unit tests

Backend — 9 new tests, all verified to fail without the fix:
- `openmetadata-service/src/test/java/org/openmetadata/service/security/JwtFilterTest.java` —
  cookie/token mismatch rejected, matching cookie accepted, session from a decommissioned provider
  rejected
- `.../security/SecurityUtilTest.java` — `writeFailureResponse` maps a Redis outage to
  `503 + Retry-After` without leaking the driver string, a missing user to `401`, a rejected
  credential to its own `401`, and anything else to `500`
- `.../security/session/SessionServiceTest.java` — `revokeSessionsForUser` revokes every active
  session; no-ops on a null user id
- `.../socket/SocketAddressFilterTest.java` — sessionless handshake refused on a session-based
  deployment, allowed on a public-client one

Frontend — 4 new tests, all verified to fail without the fix, in
`openmetadata-ui/src/main/resources/ui/src/utils/AuthProviderUtil.test.ts`: authority
trailing-slash normalization, Okta scopes without `offline_access`, `response_type` passthrough,
and its `id_token` default.

Results:
```
mvn test -pl openmetadata-service \
  -Dtest='org.openmetadata.service.security.**.*Test,org.openmetadata.service.socket.*Test,\
          org.openmetadata.service.cache.*Test,org.openmetadata.service.resources.system.*Test'
  → Tests run: 702, Failures: 0, Errors: 0, Skipped: 0

yarn test --testPathPattern "(AuthProvider|Authenticator|SignInPage)"
  → Test Suites: 9 passed, Tests: 115 passed
```

Coverage on changed classes, measured with `mvn -Pstatic-code-analysis test jacoco:report` over
that same 702-test unit scope (**not** the full module — figures are a floor, not a ceiling):

| Class | Line coverage |
|---|---|
| `SecurityUtil` | 94.3% (266/282) |
| `SocketAddressFilter` | 80.0% (100/125) |
| `SessionService` | 77.5% (365/471) |
| `JwtFilter` | 76.5% (163/213) |
| `LdapAuthServletHandler` | 45.5% (90/198) |
| `SamlAuthServletHandler` | 39.2% (140/357) |
| `BasicAuthServletHandler` | 36.2% (68/188) |
| `SecurityConfigurationManager` | 26.2% (34/130) |
| `CacheBundle` | 0% (0/104) |
| `ConfigResource` | 0% (0/48) |
| `SessionStoreUnavailableExceptionMapper` | 0% (0/9) |
| `UserRepository` | 1.1% (8/710) |

Being straight about this: the four security classes carrying the behaviour changes are at 76–94%,
but the servlet handlers, `UserRepository`, `ConfigResource`, `CacheBundle` and the new mapper are
below the 90% target and the last four have no unit coverage in this scope at all — they are
exercised by ITs in `openmetadata-integration-tests` and by `SessionMultiNodeIT` /
`SessionRedisMultiNodeIT`, which need a live stack this run did not start. The *lines this PR adds*
to `UserRepository`, `ConfigResource` and `CacheBundle` are each a handful of delegating
statements; the logic behind them is unit-tested where it lives (`SessionService`,
`SecurityConfigurationManager`, `SecurityUtil`). Happy to add ITs for the delete-revokes-sessions
and cross-pod-reload paths if reviewers want them gated here rather than in the multi-node suite.

- [ ] Not added.
- No new or changed REST endpoints — the changes are inside existing filters, repositories and the
  auth servlets. The behaviours worth an IT (delete revokes sessions across pods, cross-pod config
  reload) belong in the existing `SessionMultiNodeIT` / `SessionRedisMultiNodeIT` multi-node suite;
  say the word and I will add them there.

#### Playwright (UI) tests

- [ ] Not added.
- The UI changes are in the SSO bootstrap path: `response_type` / scope / authority construction and
  login-failure handling. Exercising them needs a real or mocked IdP that returns the provider-side
  error, which is what the `playwright/utils/sso-providers` project and the mock-IdP harness are
  for, not a page-level spec. The pure-construction logic is covered by the 4 new Jest tests above.

#### Manual testing performed

Automated verification run for this branch:
1. `mvn spotless:check -pl openmetadata-service` → clean.
2. 702 backend unit tests + 115 UI Jest tests green (above).
3. Confirmed every new test fails without its fix by stashing the source file and re-running —
   4/4 UI tests fail, and each backend assertion is on a throw that does not happen without the
   change.
4. UI CI gate reproduced locally on the changed files: `organize-imports-cli` → `eslint --fix` →
   `prettier --write` is idempotent (second pass produces no diff), `yarn license-header-check`
   passes, and `yarn i18n` reports the locales in sync.
5. New i18n key translated in all 19 non-English locales (one inserted line per file — no
   reformatting).

Not performed: an end-to-end run against live Google / Okta / Auth0 tenants and a 3-pod cluster.
The original findings came from that rig (see the linked issue for per-scenario evidence), and
re-running it is the natural verification step for this PR.

#

**TODO — not attached.** This PR is open as a **draft** for that reason.

The UI diff is auth plumbing rather than visuals: two of the five touched files are only comments
and a type widening, and the remaining changes alter the `/authorize` URL the SPA builds. The one
user-visible change is a toast replacing the infinite spinner when the IdP is unreachable, which
needs a blocked IdP (`/etc/hosts` entry per the linked issue's NEW-5 repro) to demonstrate. Mark
ready for review once that recording is attached, or say the word if a recording is not warranted
for this change and I will flip it.

#
### Checklist:
- [x] For JSON Schema changes: not applicable — no schema changes, so no migration is needed.
- [ ] For UI changes: I attached a screen recording and/or screenshots above. — **TODO, draft PR**

Bug fix:
- [x] I have added a test that covers the exact scenario we are fixing. Each finding ID from the
  linked issue maps to a test or, where it does not, the gap is stated above.


mohityadav766:
Thanks — all three findings were real and are fixed in 5cf8f82.

**`pr-pr.json` got Portuguese instead of Persian** (both reviewers). Correct call, and my mistake: I read `pr-pr` as a Portuguese variant when Portuguese is already `pt-br` / `pt-pt` and `pr-pr` is Persian/Farsi. Replaced with a Persian translation; the PR body's "translated in all 19 non-English locales" claim now actually holds.

**`revokeSessionsForUser` double-counting and re-looping** (both reviewers). Confirmed: `revoked++` fired per attempt, so a session that lost every compare-and-set was counted once per pass *and* read back on all 20 passes. Now only sessions that reach a terminal state are counted, and a pass that reads back the same batch stops the loop — reusing the no-progress guard `expireSessionsInBatches` already uses in that class rather than inventing a second one. Also dropped the `cache.invalidate` that was evicting the `REVOKED` entry `revokeSession` had just cached. New test `revokeSessionsForUser_stopsAndCountsNothingWhenSessionsCannotBeRevoked` pins both: returns `0`, and hits the store twice instead of twenty times.

**`writeFailureResponse` leaking `failure.getMessage()` on an unclassified 500** (@copilot-pull-request-reviewer). This was a regression I introduced — the LDAP handler previously returned a generic `"Authentication service error"` for internal errors and my refactor replaced it with the raw exception text. The fallback is generic again; the classified branches keep their intentional client-facing messages, and every call site already logs the exception, so nothing is lost for debugging. `SecurityUtilTest` now asserts the exception text does not reach the response body.

`mvn spotless:check` clean; 143 backend unit tests and 115 UI Jest tests green after the change.

One thing I did **not** change, in case it looks like an omission: the `Session is required` tightening on the WebSocket handshake stays gated on `createsServerSideSessions()`. Public-client OIDC mints no server-side session, so requiring one there would break every public-client deployment's socket.

mohityadav766:
Release-risk follow-up, self-reported: **408921807** fixes a defect in this PR that neither reviewer caught and that would have been an incident, not a nuisance.

The provider check I added for the session-invalidation-on-swap finding compared `UserSession.provider` to the configured provider as a plain string. But `basic` and `openmetadata` are two historical names for the *same* native-password authenticator — `SecurityConfigurationManager.isNativePasswordProvider` treats them interchangeably in six places and `BasicAuthServletHandler` serves both. So renaming one to the other, or a stored config that normalises one to the other across an upgrade, would have rejected **every live session on the deployment** for a change that does nothing.

Verified by stashing the fix and running the new test:

```
org.openmetadata.service.security.AuthenticationException:
  Not Authorized! Session was issued by a provider that is no longer configured.
```

Now compared through that equivalence, pinned by
`sessionIssuedUnderTheOtherNativePasswordProviderNameIsAccepted`. A genuine swap (google → basic,
okta → saml) still invalidates the decommissioned provider's sessions, which is the point of the
original finding.

For anyone assessing this PR against a release: the three review findings were **not** functionally
critical — one wrong-language string, one inflated log counter on a contention path, and one
information-disclosure regression on the login 500. This one was the only change in the PR that
could have logged out a whole cluster, and it is the kind of thing that only shows up on an upgrade
or a config normalisation rather than in a test run.

mohityadav766:
**Split for the release.** The three release-blocking findings now live in #31423, which is scoped to exactly those and nothing else — 14 files, backend and config only, no UI:

- **SM-20** soft delete did not revoke live sessions (7-day window on an off-boarded user)
- **SM-30** bearer tokens and `OM_SESSION` cookies written to the log at `LOG_LEVEL=DEBUG`
- **SSO-27** swapping the auth provider did not invalidate sessions minted under the old one

That PR is **not** a draft — with no UI changes the screen-recording gate does not apply, so it is ready for review as-is.

This PR keeps the remaining eleven findings: cookie↔JWT binding (SM-15), the WebSocket handshake tightening (SM-23), Redis-outage 503s (RS-04/SM-33), cross-pod SSO config propagation (SSO-26), the `CACHE_PROVIDER` documentation (SM-34), the observability nits (SM-19/SM-27), and the SSO/UI set (NEW-1/2/3/5/6). None of them are release blockers: the SSO items affect public-client and Auth0/Okta configurations that are already broken today rather than working ones, and SM-15/SM-23 are hardening of a boundary that is documented-but-unenforced rather than an escalation.

**Merge order matters** — #31423 shares `SessionService`, `JwtFilter`, `SecurityUtil`, `UserRepository` and the auth servlet handlers with this branch. Once #31423 lands I will rebase this one onto it; expect conflicts if it is merged the other way round. Happy to hold this PR entirely until after the release if that is easier to reason about.

PR #31412, opened 2026-08-12
fix: return ingestion pipeline logs on Argo and stop wiping owners on reindex

### LINKED ISSUE #31334: Reporting a pipeline run status wipes owners from the ingestion pipeline search index
## Problem

When an ingestion pipeline reports a run status, its document in `ingestion_pipeline_search_index` loses its owners.

- `IngestionPipelineRepository.addPipelineStatus` loads the pipeline with only `getFields("service")` — `owners` and `domains` are never loaded, and owner inheritance from the parent service is never applied.
- It then calls `searchRepository.updateEntityIndex(ingestionPipeline)`. When the scripted partial update does not apply, the full search document is rebuilt from that partially-loaded entity, so it is written with `owners: []`.
- Every run status report repeats this, so exactly the pipelines that run regularly end up with no owners in the search index until the next full reindex.

## Impact

Any owner-filtered search over `ingestion_pipeline_search_index` returns incomplete results — e.g. "failing pipelines owned by me" comes back empty even when the user owns a failing pipeline (directly or via the parent service).

## Suggested fix

Load the fields the index document needs (`owners,domains`) in `addPipelineStatus` before calling `updateEntityIndex`, or use a partial update that only touches `pipelineStatuses`.

### LINKED ISSUE #31411: Ingestion pipeline logs always return empty on Argo deployments
Fetching logs for an ingestion pipeline returns HTTP 200 with an empty body for every pipeline on an Argo deployment.

- `DefaultLogStorage.getLogs` builds a stub `IngestionPipeline` carrying only the name. Argo builds its workflow label selector from `service.name`, so the lookup throws, and a broad `catch (Exception)` converts that into "no logs".
- The same method reads a literal `"logs"` key, but runners key the content by task name (`ingestion_task`, `lineage_task`, ...) as built in `IngestionLogHandler.buildLogResponse`. Nothing ever writes a `"logs"` key, so even a successful fetch yields `""`.
- `IngestionPipelineRepository.getLogsFromPipelineService` has both defects as well.

Vishnuujain:
Fixes #31411
Fixes #31334

- Loads the pipeline together with its `service` before asking the runner for logs. Argo locates a workflow by `service.name`, so a name-only stub found nothing and the failure was swallowed into an empty result.
- Reads log content from the task-keyed field the runners actually return (`ingestion_task`, `lineage_task`, ...) instead of a literal `"logs"` key that is never written.
- Loads `owners`, `domains` and `followers` before reindexing a pipeline on a run status report, so they are not wiped from the search index when the full document is rebuilt.
- An unknown pipeline FQN now surfaces as not-found rather than as a pipeline with no logs.

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/logstorage/DefaultLogStorage.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/logstorage/DefaultLogStorage.java
@@ -23,6 +23,7 @@ import org.openmetadata.schema.entity.services.ingestionPipelines.PipelineStatus
 import org.openmetadata.schema.type.Include;
 import org.openmetadata.sdk.PipelineServiceClientInterface;
 import org.openmetadata.service.Entity;
+import org.openmetadata.service.exception.UnhandledServerException;
 
 /**
  * Default implementation of LogStorageInterface that delegates to the existing
@@ -64,10 +65,8 @@ public class DefaultLogStorage implements LogStorageInterface {
   @Override
   public Map<String, Object> getLogs(
       String pipelineFQN, UUID runId, String afterCursor, int limit) {
-    // Load the real pipeline: Argo builds its workflow label selector from service.name, so a
-    // name-only stub finds nothing. Airflow needs only the name, which is why a stub sufficed.
-    // Outside the catch below: an unknown pipeline is the caller's error, and reporting it as
-    // "this run has no logs" would send them looking for a runner problem that does not exist.
+    // Load the real pipeline with its service: a runner may locate the run by service, not by
+    // pipeline name alone, so a name-only stub finds nothing.
     IngestionPipeline pipeline =
         Entity.getEntityByName(Entity.INGESTION_PIPELINE, pipelineFQN, "service", Include.ALL);
     try {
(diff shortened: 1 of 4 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/logstorage/DefaultLogStorage.java
```diff
@@ -60,25 +64,24 @@ public InputStream getLogInputStream(String pipelineFQN, UUID runId) {
   @Override
   public Map<String, Object> getLogs(
       String pipelineFQN, UUID runId, String afterCursor, int limit) {
+    // Load the real pipeline: Argo builds its workflow label selector from service.name, so a
```
pmbrull: clean messaging, no Argo

openmetadata-service/src/main/java/org/openmetadata/service/logstorage/DefaultLogStorage.java
```diff
@@ -100,13 +103,30 @@ public Map<String, Object> getLogs(
...
 
+  /**
+   * Pulls the log text out of a client response. Runners key it by task name (see {@code
+   * PipelineServiceClientInterface.TYPE_TO_TASK}), e.g. {@code lineage_task}, so only the paging
+   * keys are fixed. Take the remaining entries instead of guessing the task name. Today a response
+   * carries exactly one task, but iterate in key order so more than one stays deterministic.
+   */
+  public static String extractLogContent(Map<String, String> clientLogs) {
+    if (clientLogs == null) {
+      return "";
```
pmbrull, line 105: if no client shouldn't we respond with a 500 or smt? this seems like masking something

pmbrull (changes requested):

pmbrull (approved):

PR #31423, opened 2026-08-12
Fixes open-metadata/openmetadata-collate#5526: revoke sessions on user delete and provider swap, stop logging tokens at DEBUG

mohityadav766:
### Describe your changes:

Fixes open-metadata/openmetadata-collate#5526

I worked on the three findings from the **2.0.0 multi-pod platform test run** that gate the release,
split out of #31410 so they can ship on their own. #31410 keeps the remaining eleven — cookie↔JWT
binding, the WebSocket handshake tightening, Redis-outage 503s, cross-pod SSO config propagation, the
`responseType` / `offline_access` / authority-normalization SSO fixes, and the observability nits —
none of which are release blockers.

**SM-20 — soft delete did not revoke live sessions.** Soft delete is the normal off-boarding action
in the UI, so a removed user kept full API access until their token expired: 7 days by default,
across every pod. Verified in the original run on all 9 session×pod combinations.

**SM-30 — bearer tokens and session cookies were written to the log at DEBUG.** `LOG_LEVEL` sets the
ROOT logger and nothing pinned `org.eclipse.jetty`, so its `HttpParser` logged every request header
verbatim: 14 raw JWTs, 18 `OM_SESSION` values and 14 `Bearer` tokens per pod in a normal session
workload. DEBUG is exactly what support asks customers to enable, and those logs get attached to
tickets and shipped to aggregators carrying credentials valid for up to a week.

**SSO-27 — swapping the auth provider did not invalidate old sessions.** After moving IdPs, tokens
minted under the decommissioned provider stayed valid until natural expiry, so off-boarding a user by
moving IdPs did not cut their access. Same shape as SM-20, one layer up.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

**Revoke where every caller already routes.** `UserRepository.postDelete` fires for both soft and
hard delete and already invalidates bot tokens, so revocation belongs there rather than in the
resource layer — a guard in the shared hook is smaller than a guard in each delete endpoint, and it
cannot be bypassed by a caller that takes a different route in. `SessionService.revokeSessionsForUser`
is built on the existing `findByUserIdAndStatus` + `revokeSession` primitives and is bounded: each
pass reads at most `CLEANUP_BATCH_SIZE`, and a pass that reads back an identical batch stops the
loop — reusing the no-progress guard `expireSessionsInBatches` already uses in that class rather
than inventing a second one. Revocation already notifies the WebSocket and cross-pod listeners, so
peer pods drop the user's sockets without new plumbing.

**Validate the provider, don't sweep sessions.** `UserSession` already records the provider that
authenticated it, so `JwtFilter` rejects a session whose provider is no longer configured. No store
query, no new endpoint, no cluster-wide sweep: each pod checks against its own current config, so it
holds everywhere. Enumerating and revoking every session on a config change would need a new
"find all active sessions" method on both store backends and would still race the reload.

`basic` and `openmetadata` are deliberately treated as the same provider. They are two historical
names for one native-password authenticator — `isNativePasswordProvider` treats them interchangeably
in six places and `BasicAuthServletHandler` serves both — so renaming one to the other, or a stored
config normalizing one to the other across an upgrade, must not reject every live session. Without
that equivalence this change would log out a whole deployment on a no-op rename; there is a test
pinning it.

**One classifier for the servlet auth paths.** Those paths have no JAX-RS exception mapper, so their
`catch (Exception) → 500 + e.getMessage()` catch-all reported a rejected credential as a server fault
and put the exception text on the wire. `SecurityUtil.writeFailureResponse` classifies instead.
Three unrelated hierarchies carry an intended status and none share a supertype — JAX-RS
`WebApplicationException`, `security.AuthenticationException`, and the SDK's `WebServiceException`
(the parent of `CustomExceptionMessage`, which is what the basic and LDAP authenticators actually
throw for a rejected credential). `EntityNotFoundException` is checked *first* because it also
extends the SDK class with `NOT_FOUND`, and answering 404 on a login endpoint tells an
unauthenticated caller which accounts exist. Anything unclassified stays a 500 with a generic
message; callers already log the exception.

**Backward compatibility.** SSO-27 is the one behaviour change with a blast radius worth naming: a
genuine provider swap now logs everyone out where it previously left old tokens working. That is the
finding. The `basic`↔`openmetadata` equivalence above keeps a no-op rename from triggering it. SM-20
and SM-30 are additive.

#
#### Use cases covered
- An admin soft-deletes a user → that user's session-bound tokens stop working immediately on every
  pod, instead of at token expiry. Hard delete is unchanged.
- A login attempt by a deleted user answers the 4xx the authenticator intended, not `500`.
- A login with a wrong password answers its own 4xx — same catch-all was reporting that as `500` too.
- At `LOG_LEVEL=DEBUG` the logs no longer contain `Authorization` or `OM_SESSION` values;
  `JETTY_LOG_LEVEL` still allows opting back in when someone genuinely needs wire-level Jetty debug.
- After `AUTHENTICATION_PROVIDER` is swapped, tokens minted under the old provider are rejected on
  every pod.
- Renaming `basic` → `openmetadata` (or back) does **not** invalidate sessions.

#### Unit tests

- [x] I added unit tests for the new/changed logic. 8 new tests, each verified to fail without its
      fix by stashing the source file and re-running.

- `openmetadata-service/src/test/java/org/openmetadata/service/security/session/SessionServiceTest.java`
  — `revokeSessionsForUser` revokes every active session; no-ops on a null user id; stops after two
  store reads instead of twenty when a session cannot be revoked, and counts it as zero
- `.../security/JwtFilterTest.java` — a session from a decommissioned provider is rejected; a session
  under the other native-password provider name is accepted
- `.../security/SecurityUtilTest.java` — `writeFailureResponse` keeps the status of a
  `CustomExceptionMessage` (the deleted-user login path), maps a missing user to `401` rather than a
  user-enumerating `404`, keeps the status of a rejected credential, and falls back to `500` without
  putting the exception text in the body

```
mvn test -pl openmetadata-service \
  -Dtest='org.openmetadata.service.security.**.*Test,org.openmetadata.service.socket.*Test'
  → Tests run: 641, Failures: 0, Errors: 0, Skipped: 0

mvn spotless:check -pl openmetadata-service → clean
```

Coverage on changed classes, `mvn -Pstatic-code-analysis test jacoco:report` over the unit scope
above — `SecurityUtil` 94.3%, `SessionService` 77.5%, `JwtFilter` 76.5%. `UserRepository`'s added
lines are a delegating call whose logic is unit-tested in `SessionService`; the class itself is
covered by ITs that need a live stack.

- [ ] Not added.
- No new or changed REST endpoints. The behaviours worth an IT — delete revokes sessions across pods,
  and a provider swap invalidating them — belong in the existing `SessionMultiNodeIT` /
  `SessionRedisMultiNodeIT` multi-node suite. Happy to add them there if reviewers want them gated
  on this PR rather than the multi-node run.

#### Playwright (UI) tests

- [x] Not applicable (no UI changes — this PR is backend and config only).

#### Manual testing performed
1. `mvn spotless:check -pl openmetadata-service` → clean.
2. 641 backend unit tests across the security and socket packages → green.
3. Confirmed each new test fails without its fix by stashing the source file and re-running. The
   `basic`/`openmetadata` case fails with
   `AuthenticationException: Not Authorized! Session was issued by a provider that is no longer configured.`
   — i.e. without the equivalence it would log out the deployment.
4. Traced the deleted-user login path to confirm the fix actually lands: `lookUserInProvider` throws
   `CustomExceptionMessage(BAD_REQUEST, ...)`, which is not a `WebApplicationException`, which is why
   the first version of this classifier still returned 500.

Not performed: a 3-pod cluster run against live IdPs. The findings came from that rig (per-scenario
evidence in the linked issue) and re-running it is the natural verification step. The two scenarios
to re-check are SM-20 (9 session×pod combinations after a soft delete) and SSO-27 (Google→basic swap,
then the old JWT across all pods).

#

Not applicable — no UI changes in this PR.

#
### Checklist:
- [x] For JSON Schema changes: not applicable — no schema changes, so no migration is needed.
- [x] For UI changes: not applicable — no UI changes.

Bug fix:
- [x] I have added a test that covers the exact scenario we are fixing, for each of the three
  findings.

### DIFF
--- a/conf/openmetadata-h2-test.yaml
+++ b/conf/openmetadata-h2-test.yaml
@@ -268,6 +268,8 @@ cacheMemory:
 logging:
   level: ${LOG_LEVEL:-INFO}
   loggers:
+    org.eclipse.jetty:
+      level: ${JETTY_LOG_LEVEL:-INFO}
     org.openmetadata.service.util.jdbi.OMSqlLogger:
       level: ${SQL_LOG_LEVEL:-INFO}
       additive: false
--- a/conf/openmetadata.yaml
+++ b/conf/openmetadata.yaml
@@ -242,6 +242,12 @@ cacheMemory:
 logging:
   level: ${LOG_LEVEL:-INFO}
   loggers:
+    # LOG_LEVEL sets the ROOT logger, and at DEBUG Jetty's HttpParser prints every request header
+    # verbatim — including `Authorization: Bearer <jwt>` and `Cookie: OM_SESSION=<id>`. DEBUG is
+    # exactly what support asks customers to enable and those logs get attached to tickets, so pin
+    # the framework logger independently of LOG_LEVEL.
+    org.eclipse.jetty:
+      level: ${JETTY_LOG_LEVEL:-INFO}
     org.openmetadata.service.util.jdbi.OMSqlLogger:
       level: ${SQL_LOG_LEVEL:-INFO}
       additive: false
--- a/docker/development/distributed-test/local/server1.yaml
+++ b/docker/development/distributed-test/local/server1.yaml
@@ -152,6 +152,11 @@ server:
 # Set LOG_FORMAT=json for structured logs. The default text format preserves legacy output.
 logging:
   level: ${LOG_LEVEL:-INFO}
+  loggers:
+    # ROOT at DEBUG makes Jetty's HttpParser log every header verbatim, including Authorization
+    # bearer tokens and OM_SESSION cookies. Pin it so LOG_LEVEL=DEBUG stays safe to share.
+    org.eclipse.jetty:
+      level: ${JETTY_LOG_LEVEL:-INFO}
   appenders:
     - type: console
       threshold: INFO
--- a/docker/development/distributed-test/local/server2.yaml
+++ b/docker/development/distributed-test/local/server2.yaml
@@ -152,6 +152,11 @@ server:
 # Set LOG_FORMAT=json for structured logs. The default text format preserves legacy output.
 logging:
   level: ${LOG_LEVEL:-INFO}
+  loggers:
+    # ROOT at DEBUG makes Jetty's HttpParser log every header verbatim, including Authorization
+    # bearer tokens and OM_SESSION cookies. Pin it so LOG_LEVEL=DEBUG stays safe to share.
+    org.eclipse.jetty:
+      level: ${JETTY_LOG_LEVEL:-INFO}
   appenders:
     - type: console
       threshold: INFO
--- a/docker/development/distributed-test/local/server3.yaml
+++ b/docker/development/distributed-test/local/server3.yaml
@@ -152,6 +152,11 @@ server:
 # Set LOG_FORMAT=json for structured logs. The default text format preserves legacy output.
 logging:
   level: ${LOG_LEVEL:-INFO}
+  loggers:
+    # ROOT at DEBUG makes Jetty's HttpParser log every header verbatim, including Authorization
+    # bearer tokens and OM_SESSION cookies. Pin it so LOG_LEVEL=DEBUG stays safe to share.
+    org.eclipse.jetty:
+      level: ${JETTY_LOG_LEVEL:-INFO}
   appenders:
     - type: console
       threshold: INFO
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/UserRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/UserRepository.java
@@ -95,12 +95,14 @@ import org.openmetadata.service.search.InheritedFieldEntitySearch.InheritedField
 import org.openmetadata.service.search.InheritedFieldEntitySearch.InheritedFieldResult;
 import org.openmetadata.service.secrets.SecretsManager;
 import org.openmetadata.service.secrets.SecretsManagerFactory;
+import org.openmetadata.service.security.AuthServeletHandlerRegistry;
 import org.openmetadata.service.security.SecurityUtil;
 import org.openmetadata.service.security.auth.BotTokenCache;
 import org.openmetadata.service.security.auth.SecurityConfigurationManager;
 import org.openmetadata.service.security.auth.UserActivityTracker;
 import org.openmetadata.service.security.policyevaluator.SubjectCache;
 import org.openmetadata.service.security.policyevaluator.SubjectContext;
+import org.openmetadata.service.security.session.SessionService;
 import org.openmetadata.service.util.AsyncService;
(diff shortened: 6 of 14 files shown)


harshach (dismissed):

mohityadav766:
Thanks @copilot-pull-request-reviewer — two of the three were real and are fixed in 528d77f. The third I'm pushing back on.

### 1. `REFRESHING` sessions were not revoked — real, and the most serious thing on this PR

Confirmed and fixed. A session holding a refresh lease sits in `REFRESHING`, which an ACTIVE-only lookup cannot see, and both `completeRefresh` and `releaseRefreshLease` put it back to `ACTIVE` afterwards. So a user deleted inside the 15s (`REFRESH_LEASE_MILLIS`) refresh window kept a live session for the full token lifetime — which is precisely the hole SM-20 exists to close. Narrow window, but it defeats the guarantee on the exact code path the fix is about.

Now revoking `ACTIVE` and `REFRESHING`. `PENDING` stays out deliberately: a login in flight has no user attached yet (`createPendingSession` sets no `userId`), so it can never be returned by a per-user lookup. Both stores support the lookup for any status — `RedisSessionStore.INDEXED_STATUSES` covers every non-terminal status per user, `JdbcSessionStore` filters in SQL — so this needed no store change.

Verified the hole by stashing the fix and running the new test:

```
SessionServiceTest.revokeSessionsForUser_alsoRevokesASessionThatIsMidRefresh
  expected: <1> but was: <0>
```

`revokeSessionsForUser_alsoRevokesASessionThatIsMidRefresh` also asserts `refreshLeaseUntil` is cleared, so the lease cannot be picked back up.

### 2. Misleading comment — correct

The comment described the counting behaviour from *before* the review round on #31410, not what the code does now. `revoked++` is conditional on a terminal status, so the double-counting it warned about can't happen; the real cost is burning iterations on rows that won't move. Reworded.

### 3. Mismatched backticks in `conf/openmetadata.yaml` — false positive, leaving as-is

That line has four backticks in two balanced pairs:

```
# verbatim — including `Authorization: Bearer <jwt>` and `Cookie: OM_SESSION=<id>`. DEBUG is
```

`` `Authorization: Bearer <jwt>` `` and `` `Cookie: OM_SESSION=<id>` `` both open and close on the same line — verified by counting (`backticks=4` on line 246, `0` on every neighbouring line). No change needed.

---

@harshach — you approved before these two commits landed. The `REFRESHING` one is a behaviour fix inside `revokeSessionsForUser` rather than new surface, but it does change what the PR does, so flagging it rather than letting it ride on the earlier approval. 642 backend unit tests green, `spotless:check` clean.

yan-3005 (approved):

akash-jain-10 (approved):

PR #31430, opened 2026-08-12
fix(migration): adopt pre-2.0 incidents that never had a task thread

IceS2:
The v200 cutover converts feed threads into `Task` entities, but an incident that never reached `Ack`/`Assigned` had no thread, so it comes out of a 1.13 → 2.0 upgrade with **no Task at all**.

Those incidents cannot be acked, assigned or resolved from the task-first UI — `GET`/`POST /v1/tasks/{incidentId}` 404 — while the legacy resolution endpoint still returns 200 and advances their status with no task and no workflow behind it (`storeInternal` falls through to a direct TCRS insert). The next failing run then opens a fresh incident and silently supersedes the old one, taking any ack/assign with it. Incidents that were never acked also show no history on the test case page, since the Incident tab lists tasks by `aboutEntity`.

The cutover now creates a Task for each ongoing incident chain that has no task, **reusing the chain's `stateId` as the task id**, so the `stateId == task.id` identity the task-first model assumes holds retroactively — no mapping needed, and historical results that carry that id resolve. Chains that already drive a task (directly, or via a migrated thread's `payload.testCaseResolutionStatusId`) and resolved chains are left untouched. Workflow instantiation and status replay are inherited from `backfillOpenTasksToWorkflowInstances`, which runs immediately after, so a chain sitting at `Ack`/`Assigned` lands on the right stage.

Stored resolution records drop `testCaseReference`, so the owning test case is recovered by joining the indexed `entityFQNHash`.

Verified on a real 1.13.3 → 2.0 upgrade with all five pre-2.0 incident shapes seeded:

- never acked → Task created under the legacy stateId, workflow attached, live `incidentId` unchanged, one task for the test case; `ack` transition accepted and status advances
- ack'd / assigned before the upgrade → untouched, no second task minted under the legacy id
- resolved before the upgrade (both variants) → untouched
- adopted task is indistinguishable from one created natively on 2.0: `stage=new`, workflow instance, transitions `['ack','assign','resolve']`
- idempotent: `migrate --force` re-run reports `Adopted 0`, no duplicate tasks


openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
```diff
@@ -2242,6 +2242,120 @@ private MigrationStats migrateLegacyThreadTasks() {
...
+     * task-first model assumes holds for these chains too. Workflow start and status replay are
+     * left to {@link #backfillOpenTasksToWorkflowInstances()}, which runs after this.
+     */
+    private int adoptOrphanIncidentChains() {
+      int adopted = 0;
+      try {
+        for (OrphanIncidentCandidate candidate : listLatestIncidentPerTestCase()) {
+          TestCaseResolutionStatus latest = candidate.latest();
+          UUID stateId = latest.getStateId();
+          if (stateId == null
```
harshach: instead of continue , make it !

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
```diff
@@ -2242,6 +2242,120 @@ private MigrationStats migrateLegacyThreadTasks() {
...
+              || latest.getTestCaseResolutionStatusType() == TestCaseResolutionStatusTypes.Resolved
+              || incidentChainHasTask(stateId)) {
+            continue;
+          }
+          Task task = buildTaskFromIncidentChain(latest, candidate.testCaseId());
+          if (task == null) {
```
harshach: we can use emptyOrNull and same make it !

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
IceS2: Done

openmetadata-service/src/main/java/org/openmetadata/service/migration/utils/v200/MigrationUtil.java
IceS2: Inverted the logic, but keep it without emptyOrNull. Trying to understand if that is a pattern we use for objects but it seems we do it for strings and collections mainly. If that is not actually the case I can change all to use emptyOrNull

harshach (approved):

PR #31460, opened 2026-08-13
fix(service): reconcile seeded tags on startup

pmbrull:
## Problem

`TagResource.initialize()` seeds tags through `EntityRepository.initializeEntity()`, which is create-if-absent:

```java
T existingEntity = findByNameOrNull(entity.getFullyQualifiedName(), ALL);
if (existingEntity != null) {
  LOG.debug("{} {} is already initialized", entityType, entity.getFullyQualifiedName());
  return;
}
```

The tag row survives a bad write, so the seed is never re-applied. Two ways a seeded tag drifts and then stays drifted across every restart:

- **`CreateTag` defaults `recognizers` to empty**, so a PUT that never mentions them is indistinguishable from one clearing them. #31205 stopped new wipes via `preserveRecognizerConfigOnPut()`, but nothing repairs a tag already emptied.
- **`CreateTag.provider` is a bare `$ref` to `providerType`, whose schema default is `user`**, so the same PUT downgrades `provider` from `system`. That also drops the guard at `EntityRepository.java:8182` that stops a system entity being deleted.

Same gap means recognizers added to the seed in a later release never reach installs that already have the tag — the only paths that re-apply recognizer seed data are the `v1110`/`v1120`/`v1122` migrations, which do not re-run once recorded.

Observed on a 2.0.0 install: `PII.Sensitive` went from 39 recognizers / `provider: system` at v0.1 to 0 recognizers / `provider: user` / `autoClassificationEnabled: false` / priority 100→50 at v0.2, and multiple restarts since did not restore any of it.

## Fix

Reconcile the seed after the create pass in `TagResource.initialize()`:

- Missing recognizers are matched **by name** and appended with a fresh id.
- A recognizer that is still present is **never touched** — user edits to its `confidenceThreshold`, `exceptionList` or `enabled` flag survive.
- A `provider` that no longer matches the seed is restored. Unlike `autoClassificationEnabled`, provider is not a user preference; it records origin.
- Recognizers are inert while `autoClassificationEnabled` is false, so a tag that lost **every** seeded recognizer — the signature of a wipe — also gets that flag and its priority back. A tag merely short a newly shipped recognizer keeps its own settings.
- A tag that has not drifted is not written at all.

Gating is on the seed recognizer's `isSystemDefault`, not on the stored tag's `provider`: the same PUT that wipes recognizers also downgrades provider, so a provider gate would skip exactly the tags that need healing.

Provider reconciliation covers all seeded tags — all five seed classifications (PII, Tier, Certification, PersonalData, KnowledgeCenter) declare `provider: system` and drift identically. Recognizer reconciliation only applies to tags whose seed defines them.

## Tests

`TagRepositoryUnitTest`, 21 passing (6 new):

- wipe → recognizers, `autoClassificationEnabled`, priority and provider all restored
- user-edited recognizer → same instance retained, `enabled: false` and threshold `0.95` survive; tag settings untouched
- provider drifted alone → provider restored, recognizers and autoClassification left as the user had them
- nothing drifted → `verify(never()).store(...)`
- non-`isSystemDefault` seed recognizer → ignored
- `missingSystemRecognizers` deep-copies the seed and mints fresh ids

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TagRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TagRepository.java
@@ -224,6 +224,91 @@ public class TagRepository extends EntityRepository<Tag> {
     }
   }
 
+  /**
+   * Heals seeded Tags: system recognizers the Tag lost - or that a later release added - are put
+   * back by name, and a provider knocked off the seeded value is restored. A recognizer that is
+   * still there is never touched, so user edits to it survive.
+   */
+  public void reconcileSeededTags(List<Tag> seedTags) {
+    for (Tag seedTag : listOrEmpty(seedTags)) {
+      try {
+        reconcileSeededTag(seedTag);
+      } catch (Exception e) {
+        LOG.warn("Failed to reconcile seeded tag {}", seedTag.getFullyQualifiedName(), e);
+      }
+    }
+  }
+
+  private void reconcileSeededTag(Tag seedTag) {
+    // A Tag missing altogether was just created from this same seed by initializeEntity()
+    Tag stored = findByNameOrNull(seedTag.getFullyQualifiedName(), ALL);
+    if (stored == null) {
+      return;
+    }
+
+    List<Recognizer> seeded = systemRecognizers(seedTag);
+    List<Recognizer> missing = missingSystemRecognizers(stored.getRecognizers(), seeded);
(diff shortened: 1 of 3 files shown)


IceS2 (approved):

pmbrull:
picked to 2.0

pmbrull:
validated in 2.0 instance, tags are back to life

PR #31461, opened 2026-08-13
Fixes #31459: actionable errors when the OMeta client cannot start

### LINKED ISSUE #31459: Workflow init fails with "'Response' object is not subscriptable" instead of naming the bad hostPort or version mismatch
### Describe the bug

Every ingestion workflow starts by calling `GET /system/version`. When that call does not return the JSON the client expects, the failure surfaces as a `TypeError` several frames away from the actual misconfiguration, with no hint of what is wrong:

```
[ERROR] Error initializing metadata: 'Response' object is not subscriptable
For more information, please visit: https://docs.open-metadata.org/connectors/ingestion/workflows/metadata
```

The message is identical whether the cause is a wrong `hostPort`, an expired token, a 404, or a client/server version mismatch. It gives the user nothing to act on.

### To Reproduce

1. Point a workflow at a server without the `/api` suffix:
   ```yaml
   workflowConfig:
     openMetadataServerConfig:
       hostPort: https://<your-server>/
   ```
2. Run `metadata ingest -c <config>.yaml`.

The request lands on `https://<your-server>/v1/system/version`. The OpenMetadata UI serves `index.html` for any route it does not recognise, so the client gets **HTTP 200 with `text/html`**:

```
$ curl -s -o /dev/null -w "%{http_code} %{content_type}\n" https://<server>/v1/system/version
200 text/html;charset=utf-8
$ curl -s -o /dev/null -w "%{http_code} %{content_type}\n" https://<server>/api/v1/system/version
200 application/json
```

### Why it fails this way

Four layers each drop information:

| Layer | Behaviour |
|---|---|
| `ometa/client.py` `_one_request` | JSON decode fails, so it returns the raw `requests.Response` **as if it were the decoded body**, logging only at DEBUG |
| `ometa/mixins/server_mixin.py` `get_server_version` | `self.client.get("/system/version")["version"]` guards only `KeyError`, so the `TypeError` escapes |
| `ometa/ometa_api.py` `health_check` | repeats the same unguarded subscript |
| `ometa/client_utils.py` `create_ometa_client` | `raise ValueError(exc)` drops the exception class; logs `"Wild error initialising the OMeta Client"` |

`basedpyright` already flags both subscripts (`reportIndexIssue`, `reportOptionalSubscript`) — they sit in `ingestion/.basedpyright/baseline.json`.

The same dead-end message appears for the other failure modes of that endpoint: 404, 401/403, a non-JSON body, and a client/server version mismatch (e.g. a 2.0 client against a 1.13 server), which is arguably the most common one and the one users most need spelled out.

### Expected behavior

Each failure mode should name what to fix — the URL actually called, the status, and the remediation (`hostPort` missing `/api`, expired token, install the matching client version, …).

### Version

- OpenMetadata client: 2.0.0.0.dev0 (reproduced against a 1.13.x server, but the code path is version-independent)

pmbrull:
### Describe your changes:

Fixes #31459

Every workflow starts with `GET /system/version`, and when that call did not return the expected JSON the client returned the raw `requests.Response` **as if it were the decoded body**, so users got `TypeError: 'Response' object is not subscriptable` several frames away from the real problem — identical whether the cause was a `hostPort` missing `/api`, an expired token, a 404, or a client/server version mismatch. I made each failure mode name what to fix, because that message cost real debugging time and told the user nothing.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change (4 source files, one call path).

For the record, the shape of the fix: an HTML body is now an explicit `HtmlResponseError` raised by the client itself, since no endpoint legitimately answers HTML — the OpenMetadata UI serves `index.html` for unknown routes, which is exactly why a `hostPort` without `/api` returns `200 text/html`. CSV and ODCS-YAML exports genuinely answer non-JSON, so they keep receiving their `Response`; only HTML is promoted to an error. `get_server_version` then reads the raw response and classifies the rest (401/403, 404, other non-2xx, non-JSON, missing `version`, exhausted retry budget), each naming the URL actually called. `create_ometa_client` keeps the exception class, host and `__cause__` instead of flattening to `ValueError(str)`; `OMetaClientInitError` subclasses `ValueError` so existing `except ValueError` callers are unaffected. `health_check` reuses `get_server_version` rather than repeating the same unguarded subscript.

The two subscripts this fixes were already sitting in `ingestion/.basedpyright/baseline.json` as `reportIndexIssue` / `reportOptionalSubscript`; those entries are dropped.

#
#### Use cases covered
- `hostPort` missing `/api` → error names the UI catch-all and the `/api` suffix, instead of `TypeError`
- 2.0 client against a 1.13 server → error reports both versions and the `pip install` remedy
- Wrong `apiVersion` (404) → error names the URL and the config keys to check
- Expired/invalid token (401/403) → error points at `securityConfig` and the bot user
- Unreachable host → transport error with the DNS/connection cause
- 5xx, non-JSON non-HTML body, missing `version` field, exhausted retry budget → each named
- CSV / ODCS-YAML exports keep working (regression guard — they rely on the non-JSON path)

#### Unit tests
- [x] Added unit tests for the new/changed logic.
- File added: `ingestion/tests/unit/ometa/test_client_error_messages.py` (17 tests)
- Coverage: **98% (42/43) of added runtime statements** — `make unit_ingestion` scoped to `metadata.ingestion.ometa.{client,client_utils,mixins.server_mixin,ometa_api}`. The one miss is `health_check`'s single delegating line, which `ingestion/tests/unit/conftest.py` stubs globally for all unit tests; it is covered by the manual runs below. Import/`class`/`def` statements are excluded from the ratio: `conftest.py` imports `metadata` before pytest-cov starts, so they are reported missing even for unchanged code.
- Full suite around the change: 555 passed, 1 skipped (`test_dataframe_validator.py`, pre-existing missing `pandas` locally).

- [x] Not applicable — the change is in the OMeta client, not a connector. Covered by unit tests plus the manual end-to-end runs below.

#### Playwright (UI) tests

#### Manual testing performed
Ran real ingestions with `metadata ingest -c redshift.yaml`, varying only `hostPort`:

1. **Regression check** — against a local 2.0 server: full Redshift ingestion, `Workflow Success %: 100.0`, 15 records processed, 0 errors.
2. **`hostPort: https://<1.13-server>/`** (no `/api`):
   ```
   Error initializing metadata: Could not initialize the OpenMetadata client against
   [https://<server>/]: HtmlResponseError: Got an HTML page instead of JSON from
   [https://<server>//v1/system/version] (HTTP 200). The request reached the OpenMetadata UI,
   not the REST API - the UI answers index.html for unknown routes. Check that `hostPort`
   points at the API (e.g. https://<host>/api, ending in `/api`) and that no proxy or login
   page is intercepting the call.
   ```
3. **`hostPort: https://<1.13-server>/api`** with a 2.0 client:
   ```
   Error initializing metadata: Could not initialize the OpenMetadata client against
   [https://<server>/api]: VersionMismatchException: Server version is 1.13.4 vs. Client
   version 2.0.0.0. Major and minor versions should match. Either install the matching client
   with `pip install 'openmetadata-ingestion~=1.13.0'` or point `hostPort` at a 2.0.0.0 server.
   ```
4. Bad `apiVersion` (404) and an unresolvable host, both returning their specific messages.

Also confirmed the server-side behaviour the fix keys off:
```
$ curl -s -o /dev/null -w "%{http_code} %{content_type}\n" https://<server>/v1/system/version
200 text/html;charset=utf-8
$ curl -s -o /dev/null -w "%{http_code} %{content_type}\n" https://<server>/api/v1/system/version
200 application/json
```

`ruff check`, `ruff format --check` and `basedpyright` are clean on all changed files.

#

Not applicable.

#
### Checklist:
- [x] My PR is linked to a GitHub issue via `Fixes #31459` above.

- [x] I have added a test that covers the exact scenario we are fixing (`TestHtmlResponseIsNamed`, `TestGetServerVersionErrors`, `TestVersionMismatchMessage` — all fail before the fix with `TypeError: 'Response' object is not subscriptable`).

### DIFF
--- a/ingestion/src/metadata/ingestion/ometa/client.py
+++ b/ingestion/src/metadata/ingestion/ometa/client.py
@@ -90,17 +90,21 @@ def is_html_body(resp: requests.Response) -> bool:
     return head.startswith("<") and "<html" in head.lower()
 
 
-def _decode_body(resp: requests.Response, url: object):
+def _decode_body(resp: requests.Response, url: object, raise_on_html: bool = False):
     """Decode a successful response body.
 
-    JSON when it parses; the ``Response`` itself for the text payloads some
-    endpoints answer with (CSV and ODCS-YAML exports); an ``HtmlResponseError``
-    when it is a UI page, since no endpoint legitimately answers HTML.
+    JSON when it parses; otherwise the ``Response`` itself, for the text payloads
+    some endpoints answer with (CSV and ODCS-YAML exports).
+
+    ``raise_on_html`` turns an HTML page into an ``HtmlResponseError`` instead. It
+    is opt-in because callers disagree on what HTML means: the OpenMetadata API
+    never answers it, but connectors share this client and some deliberately
+    tolerate a non-JSON reply on their ingestion path.
     """
     try:
         return resp.json()
     except JSONDecodeError as json_decode_error:
-        if is_html_body(resp):
+        if raise_on_html and is_html_body(resp):
             raise HtmlResponseError(url, resp.status_code) from json_decode_error
         logger.debug(
             "Non-JSON response (%s) from [%s] with content type [%s] returned as-is: %s",
@@ -190,6 +194,10 @@ class ClientConfig(ConfigModel):
     user_agent: Optional[str] = None  # noqa: UP045
     raw_data: Optional[bool] = False  # noqa: UP045
     allow_redirects: Optional[bool] = False  # noqa: UP045
+    # Treat an HTML body as an error rather than handing the caller the raw
+    # Response. Off by default: connectors share this client against third-party
+    # APIs, and some tolerate a non-JSON reply on purpose.
+    raise_on_html: bool = False
     auth_token_mode: Optional[str] = "Bearer"  # noqa: UP045
     verify: Optional[Union[bool, str]] = None  # noqa: UP007, UP045
     cookies: Optional[Any] = None  # noqa: UP045
@@ -370,7 +378,7 @@ class REST:
                 return resp
 
             if resp.text != "":
-                return _decode_body(resp, url)
+                return _decode_body(resp, url, self.config.raise_on_html)
 
         except HTTPError as http_error:
             # retry if we hit Rate Limit
--- a/ingestion/src/metadata/ingestion/ometa/client.py
+++ b/ingestion/src/metadata/ingestion/ometa/client.py
@@ -58,25 +58,26 @@ class RestTransportError(Exception):
 class HtmlResponseError(Exception):
     """An HTML page came back where the API answers JSON.
 
-    The OpenMetadata UI serves `index.html` for any route it does not recognise, so
-    this almost always means the request never reached the REST API: `hostPort` is
-    missing the `/api` suffix, or a proxy/login page intercepted it. Raised instead
-    of handing the caller a ``Response`` it would try to subscript.
+    The body is a web page, not an API response: the request reached a UI, a login
+    page or a proxy rather than the endpoint. Raised instead of handing the caller a
+    ``Response`` it would try to subscript.
+
+    ``REST`` is generic - connectors use it against third-party APIs too - so the
+    message stays provider-neutral. Callers that know which API they were talking to
+    pass a ``hint`` with the advice specific to it.
     """
 
-    def __init__(self, url: object, status_code: int) -> None:
+    def __init__(self, url: object, status_code: int, hint: Optional[str] = None) -> None:  # noqa: UP045
         super().__init__(
             f"Got an HTML page instead of JSON from [{url}] (HTTP {status_code})."
-            " The request reached the OpenMetadata UI, not the REST API - the UI answers"
-            " index.html for unknown routes. Check that `hostPort` points at the API"
-            " (e.g. https://<host>/api, ending in `/api`) and that no proxy or login page"
-            " is intercepting the call."
+            " The endpoint served a web page, not an API response - check the configured"
+            " host/URL and that no proxy or login page is intercepting the call." + (f" {hint}" if hint else "")
         )
         self.url = url
         self.status_code = status_code
 
 
-def _is_html_body(resp: requests.Response) -> bool:
+def is_html_body(resp: requests.Response) -> bool:
     """Whether a non-JSON body is an HTML page.
 
     Content type first; some proxies mislabel index.html as text/plain, so fall back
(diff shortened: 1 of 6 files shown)


ingestion/src/metadata/ingestion/ometa/client.py
```diff
@@ -55,6 +55,66 @@ def __init__(self, method: str, url: object, cause: BaseException) -> None:
...
+    of handing the caller a ``Response`` it would try to subscript.
+    """
+
+    def __init__(self, url: object, status_code: int) -> None:
+        super().__init__(
+            f"Got an HTML page instead of JSON from [{url}] (HTTP {status_code})."
+            " The request reached the OpenMetadata UI, not the REST API - the UI answers"
+            " index.html for unknown routes. Check that `hostPort` points at the API"
+            " (e.g. https://<host>/api, ending in `/api`) and that no proxy or login page"
+            " is intercepting the call."
```
IceS2: This mentions OpenMetadata but raises on REST, used by connectors

pmbrull: Good catch — you're right, and it's worse than the wording: ~25 connectors (Superset, PowerBI, Databricks, NiFi, Airbyte, Redash, Atlas, dbt Cloud, …) import this same `REST` class to call third-party APIs. An HTML login page from any of them would have told the user to add `/api` to `hostPort`, which is just wrong.

Fixed in 1779e837e6: the message raised by the client is now provider-neutral ("the endpoint served a web page, not an API response — check the configured host/URL…"), and `HtmlResponseError` takes an optional `hint`. `server_mixin`, which *does* know it is talking to an OpenMetadata server, supplies the `hostPort`/`/api` advice. Added `test_message_stays_provider_neutral` to assert the client message never names OpenMetadata or `hostPort`.

ingestion/src/metadata/ingestion/ometa/mixins/server_mixin.py
```diff
@@ -73,24 +78,71 @@ def get_server_version(self) -> str:
...
-                + " If running the server in DEV mode locally, make sure to `mvn clean install`."
+            payload = response.json()
+        except JSONDecodeError as exc:
+            # `get_raw` skips the client's JSON handling, so classify the body here.
+            if "html" in response.headers.get("Content-Type", "").lower() or response.text.lstrip().startswith("<"):
+                raise HtmlResponseError(url, response.status_code) from exc
+            raise VersionNotFoundException(
+                f"The response from [{url}] is not JSON"
+                f" (content type [{response.headers.get('Content-Type', 'unknown')}]): {response.text[:500]}"
+            ) from exc
```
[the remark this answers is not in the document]
pmbrull: Confirmed and fixed in 1779e837e6 — the divergence was real. `_read_version` sniffed with a bare `startswith("<")` while the client also required an `<html` tag, so an XML/SVG error body (e.g. an S3-style `<?xml …><Error>`) got the misleading "points at the UI, add `/api`" hint instead of the accurate "not JSON" message.

Both paths now share the helper, promoted from `_is_html_body` to `is_html_body` since it is crossing module boundaries. Added `test_xml_body_is_not_mistaken_for_a_ui_page` covering exactly the `<?xml` case you described.

IceS2 (approved):

pmbrull:
picked to 2.0

PR #31473, opened 2026-08-13
fix(e2e): handle dry-run modal in DomainDataProductsWidgets remove test (1.13)

anuj-kumary:
## Summary

- The `Domain asset count should update when assets are removed` test in `DomainDataProductsWidgets.spec.ts` was failing on the 1.13 nightly CI run ([#31655001287](https://github.com/open-metadata/openmetadata-nightly/actions/runs/31655001287/job/94313571811))
- Root cause: the test used a wildcard response matcher `'**/assets/remove'` which resolved immediately on the **dry-run** request. The confirmation modal was never dismissed, so the actual removal never fired, and `checkAssetsCount(page, 1)` failed because assets remained at 2.
- Fix: add the two-phase wait pattern — await the dry-run response, click the `domain-dry-run-modal` save button, then await the actual remove response — matching what the `main` branch already does.

## Test plan
- [ ] Nightly 1.13 CI run: `DomainDataProductsWidgets.spec.ts > Domain and Data Product Asset Counts > Domain asset count should update when assets are removed` passes


anuj-kumary:
Closing — the actual failure is not the dry-run modal but the waitForResponse race condition and stale widget poll. Fixed correctly in #31477 (1.13) and #31476 (main).

PR #31484, opened 2026-08-13
Fixes 31482: Send the service FQN as the deleteStale scope

### LINKED ISSUE #31482: [BUG] Stale deletion never fires when the service name requires FQN quoting
### Affected module
Ingestion Framework

### Describe the bug

`markDeletedDashboards`, and every other `markDeleted*` option that is scoped by service, has no effect when the service name contains a character that forces its FQN to be quoted — in practice, a dot (so any service with a version number in its name).

The connector passes the service **name** as the deletion scope. `deleteStale` resolves the scope against the stored **FQN**, which for a service is its *quoted* name. The scope never resolves, so the server finds nothing in it and returns success having deleted nothing. The run reports 100% success.

### Root cause

`ingestion/src/metadata/ingestion/api/delete.py` forwards the caller's scope dict straight to `deleteStale`:

```python
result = metadata.delete_stale_entities(
    entity=entity_type,
    scope_params=params,      # {"service": "Looker- 2.0 Test"} — raw name, not FQN
    ...
)
```

Server side, `EntityRepository.bulkDeleteStaleEntities` gates on `scopeExists` → `findByNameOrNull(scopeFqn)` and then lists descendants via `listDescendantIdFqnByPrefixNonDeleted(scopeFqn)`, whose prefix is `FullyQualifiedName.buildHash(scopeFqn) + ".%"`. Both key off the stored FQN hash, so an unquoted name misses on both.

**How it regressed.** 1.13.3 fed the same dict to a list call:

```python
entity_state = metadata.list_all_entities(entity=entity_type, params=params)
```

so `{"service": "Looker- 2.0 Test"}` became `?service=Looker- 2.0 Test`, and list endpoints match a service by **name** — which works. 2.0 keeps the identical dict but passes it to `deleteStale` as `scopeFqn`, which resolves against FQNs. The producer still emits a name, the consumer now needs an FQN, and nothing between them converts it. For any service whose name needs no quoting the two strings are identical, which is why it went unnoticed.

### Evidence

Traced call from a real run, filter excluding 3 of 4 dashboards:

```
DELETE_STALE CALLED entity=Dashboard          scope={'service': 'Looker- 2.0 Test'} n_live=1
DELETE_STALE RESULT processed=0 passed=0
DELETE_STALE CALLED entity=DashboardDataModel scope={'service': 'Looker- 2.0 Test'} n_live=4
DELETE_STALE RESULT processed=0 passed=0
DELETE_STALE CALLED entity=Chart              scope={'service': 'Looker- 2.0 Test'} n_live=6
DELETE_STALE RESULT processed=0 passed=0
```

`n_live=1` is correct — one dashboard survived the filter, three should have been soft-deleted, zero were.

What OpenMetadata stores:

```
service name : 'Looker- 2.0 Test'
service FQN  : '"Looker- 2.0 Test"'
dashboards   : "Looker- 2.0 Test".1 ... .4
```

Controlled A/B against the same endpoint with the same `seenFqns` (3 of 4) and `dryRun: true`. Only the scope string differs:

```
scopeFqn = 'Looker- 2.0 Test'     ->  processed=0  passed=0   (what the connector sends)
scopeFqn = '"Looker- 2.0 Test"'   ->  processed=1  passed=1   (the stored FQN)
```

The endpoint is fine. It is being handed a scope that does not exist.

### Scope of impact

The bug is in the shared `delete_entity_from_source` helper, so it is not Looker-specific. Every service-scoped stale deletion is affected:

| File | Entity |
|---|---|
| `database_service.py` | Database (`markDeletedDatabases`) |
| `dashboard_service.py` | Dashboard, DashboardDataModel, Chart |
| `pipeline_service.py` | Pipeline |
| `messaging_service.py` | Topic |
| `search_service.py` | SearchIndex |
| `storage_service.py` | Container |
| `mlmodel_service.py` | MlModel |
| `api_service.py` | APICollection |
| `drive_service.py` | Directory, File, Spreadsheet |

Non-service scopes are unaffected — `{"database": ...}`, `{"databaseSchema": ...}` and `{"spreadsheet": ...}` are already built as FQNs via `fqn.build` / `_get_filtered_schema_names(return_fqn=True)`.

### To Reproduce

1. Create a dashboard service whose name contains a dot, e.g. `Looker- 2.0 Test`.
2. Ingest several dashboards.
3. Re-run with a filter matching only one of them, and `markDeletedDashboards: true`.
4. Query the dashboards with `include=all`.

**Expected:** the excluded dashboards are soft-deleted.
**Actual:** all present with `deleted: false`, and the run reports 100% success.

Rename the service without a dot and the same test passes.

### Version

- OpenMetadata version: 2.0.0 (regression from 1.13.3)
- Ingestion: 2.0.0

ulixius9:
### Describe your changes:

Fixes #31482

I worked on **service-scoped stale deletion silently deleting nothing** because `deleteStale` resolves its scope against the stored FQN, while every caller passes the raw service **name** — so for any service whose name needs quoting (in practice, one containing a dot, e.g. `Looker- 2.0 Test`) the scope never resolved and the server returned success having deleted nothing. The fix converts the service scope to its FQN at the `deleteStale` boundary in the shared `delete_entity_from_source` helper, which repairs all service-scoped `markDeleted*` options at once rather than only the dashboard ones.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

The conversion is done in one place — `_scope_params_as_fqn` in `ingestion/api/delete.py` — and applied **only** to the `delete_stale_entities` call, not to the caller's dict:

- **Only the `service` key is converted.** Every other scope key (`database`, `databaseSchema`, `spreadsheet`) already arrives as a full FQN; quoting a multi-part FQN would break a scope that currently works.
- **The legacy fallback keeps the raw name.** On a 404 from an older server, `_delete_stale_entities_legacy` calls `list_all_entities(params=params)`, and `?service=` matches a service by *name*. This is why the obvious fix — passing the FQN from the 13 call sites — is wrong: it would fix 2.0 and break the fallback.
- **`quote_name` is idempotent** (`'A.B'` → `'"A.B"'`, `'"A.B"'` → `'"A.B"'`, `'AB'` → `'AB'`), and is exactly what `fqn._build` does for a one-part FQN, so a caller already passing an FQN is unaffected. It raises `ValueError` on names containing `"`, so that is caught and the scope passes through unchanged rather than aborting the delete stage.
- **Key order is preserved**, since `delete_stale_entities` reads the scope from the first entry of the dict.

Alternatives rejected: converting at the 13 call sites (breaks the legacy fallback, and every new caller can reintroduce the bug); converting inside `OpenMetadata.delete_stale_entities` (puts ingestion-shaped knowledge in the transport layer and makes its documented "value is the scope FQN" contract conditional on the key); making the server resolve a `service` scope by name (a Java change that does not help anyone on an older server, when the client is the side that is wrong).

No schema, generated-model, Java or UI changes. Backward compatible in both directions: new connector against old server takes the legacy path with the raw name, new connector against new server sends the FQN.

#
#### Use cases covered
- Re-running ingestion for a service whose name contains a dot, with `markDeleted*` enabled, now soft-deletes the entities the filter excluded (previously zero were deleted and the run reported 100% success).
- Fixes this for every service-scoped stale deletion, not just dashboards: Database, Dashboard, DashboardDataModel, Chart, Pipeline, Topic, SearchIndex, Container, MlModel, APICollection, Directory, File, Spreadsheet.
- Services whose names need no quoting keep working exactly as before.
- Connectors running against a pre-`deleteStale` server keep working via the legacy client-side diff.

#### Unit tests
- Files updated: `ingestion/tests/unit/topology/test_delete_stale.py` — 9 new tests in `TestServiceScopeIsSentAsFqn`.
- Three encode the fix: a dotted name is sent quoted, `service` stays the first key, and the quoted FQN reaches the actual request body end-to-end through the real client method. Six guard against regressions the fix could cause: dotless names unchanged, already-quoted not double-quoted, `database`/`databaseSchema` FQNs never quoted, the caller's dict not mutated, unquotable names passed through instead of raising, and the legacy fallback still listing by the raw name.
- The three fix tests fail before the change (`3 failed, 13 passed`) and pass after (`16 passed`).
- Coverage: every line and branch of the new `_scope_params_as_fqn` is exercised. The file-level number is 48%, which includes the untouched `delete_entity_by_name` and a pre-existing untested `except` branch.

- [x] Not applicable (no backend API changes — the server side is already correct).

- Unit tests only. Regression net for the 13 untouched call sites: `tests/unit/topology/database/test_postgres.py`, which asserts on `delete_entity_from_source` call args. Full run of `test_postgres.py` + `tests/unit/topology/dashboard` + `test_delete_stale.py`: **587 passed, 1 skipped**.

#### Playwright (UI) tests

#### Manual testing performed
1. `ruff check` + `ruff format --check` on both changed files — clean.
2. `basedpyright --baselinemode=discard src/metadata/ingestion/api/delete.py` — 0 errors, 0 warnings, 0 notes.
3. The `dryRun` A/B that isolates the bug is recorded in #31482: same endpoint, same `seenFqns`, only the scope string differs — `'Looker- 2.0 Test'` → `processed=0`, `'"Looker- 2.0 Test"'` → `processed=1`.

**Not yet done:** the full end-to-end re-run against a live Looker service named `Looker- 2.0 Test` (ingest four dashboards, re-run with a filter matching one, confirm the other three come back `deleted: true`). Worth a reviewer confirming on a real stack before merge.

#

Not applicable.

#
### Checklist:


ingestion/src/metadata/ingestion/api/delete.py
```diff
@@ -23,15 +23,42 @@
...
+    passes an FQN is unaffected. Key order is preserved because the scope is read from the first
+    entry.
+    """
+    if not params or SERVICE_SCOPE_KEY not in params:
+        return params
+    service = params[SERVICE_SCOPE_KEY]
+    try:
+        return {**params, SERVICE_SCOPE_KEY: quote_name(service)}
+    except ValueError:
+        logger.warning(f"Cannot build the FQN of service '{service}'; sending the delete scope unchanged")
```
IceS2: Should not use f-string for logs

ingestion/tests/unit/topology/test_delete_stale.py
```diff
@@ -200,3 +200,104 @@ def test_falls_back_to_legacy_when_endpoint_missing(self):
...
+
+        assert metadata.list_all_entities.call_args.kwargs["params"] == {"service": "Looker- 2.0 Test"}
+        deleted_fqns = {r.right.entity.fullyQualifiedName.root for r in results if not isinstance(r.right, Barrier)}
+        assert deleted_fqns == {"svc.d2"}
+
+    def test_quoted_fqn_reaches_the_wire(self):
+        """End to end through the real client: the request body carries the service FQN."""
+        metadata = MagicMock()
+        metadata.get_suffix.return_value = "/dashboards"
+        metadata.client.put.return_value = {
```
IceS2: Current main changed delete_stale_entities to use client.delete in [#31375](https://github.com/open-metadata/OpenMetadata/pull/31375).

This test still stubs client.put and reads it again on line 301, so the conflict-free merge calls delete once, put zero times, and fails there.

Please update both references to delete and rerun after merging the latest main.

IceS2 (changes requested):

IceS2 (approved):

PR #31496, opened 2026-08-13
fix(service): key data product ports by entity id, not request index

### LINKED ISSUE #31495: Data Product portsView returns ports keyed by list index, so rows can be misattributed or silently dropped
### Affected module

Backend — `DataProductRepository.getPaginatedPorts` (`/dataProducts/{id|name}/portsView`, `/inputPorts`, `/outputPorts`)

### Describe the bug

`getPaginatedPorts` correlates bulk-fetched entities with the relationship records that requested them **by list index**:

```java
List<EntityInterface> entitiesOfType =
    Entity.getEntities(entry.getValue(), fieldsToFetch, NON_DELETED);
for (int i = 0; i < entitiesOfType.size(); i++) {
  entitiesById.put(
      entry.getValue().get(i).getId(), new EntityWithType(entitiesOfType.get(i), entityType));
}
```

That assumes the bulk fetch returns **one row per requested ref, in request order**. It does neither. The call resolves to `EntityDAO.findByIds`:

```sql
SELECT id, json FROM <table> WHERE id IN (<ids>) <cond>
```

There is no `ORDER BY`, so row order is whatever the plan produces — index vs sequential scan, buffer state, concurrency. And `<cond>` is the `NON_DELETED` filter, so the result can be **shorter** than the request.

Two distinct failures follow:

1. **Reordered result → relationship order is not preserved.** With equal counts the mapping is a permutation, so `entities[j]` ends up being `dbResult[j]`: the returned *set* is correct but the rows come back in database order rather than the order the ports were added. (An earlier revision of this issue claimed the entity itself could be misattributed to another port — that is **not** correct, since a row *is* the entity.)
2. **Short result → trailing ids never mapped.** Only the first `dbResult.size()` input ids get keyed, so the remainder are dropped by the `if (entity != null)` guard further down. Meanwhile `total` comes from an independent `countFindTo` query and still counts them, so **the payload contradicts its own count**. This is the failure with user-visible impact.

### To Reproduce

Failure 2 is visible in the AUT nightly. The page snapshot from a failing run of `InputOutputPorts › Toggle fullscreen mode` shows:

```
tab   "Input/Output Ports 1" [selected]
      "Input Ports (1)"
group "Ports Lineage":
      "To view lineage, assign assets to this Data Product and add them as input/output ports."
```

The tab badge reads `1` — that is `paging.total` from the count probe — while the lineage panel rendered its empty branch, because the same `/portsView` response carried **zero rows**. `PortsLineageView` derives `hasAnyPorts` from the row array:

```ts
const hasAnyPorts =
  dataProduct && (inputPortsData.length > 0 || outputPortsData.length > 0);
```

Because ordering depends on the query plan, it is intermittent rather than constant, and it is engine-agnostic — the AUT data shows it at the same rate on MySQL and PostgreSQL.

### Expected behavior

A port's row carries that port's entity, and the page's `total` is consistent with the rows it returns.

### Version

- OS: Kubernetes (AUT nightly, EKS/RDS)
- OpenMetadata: `2.0` and `main` (identical code)
- Observed in AUT nightlies from Aug 9–13

### Additional context

Open question for maintainers: should `total` be reconciled with the rows actually returned (count only resolvable relationships), or is a dangling relationship a data-integrity problem to fix at its source? The keying fix below does not change `total`, so a page can still report more ports than it returns.

This was found while root-causing AUT nightly flakes; the UI-test side is handled in #31494, which mitigates the symptom but explicitly does not fix this.

ShaileshParmar11:
Fixes #31495

## What

`getPaginatedPorts` paired the bulk-fetched entities with the relationship records that requested them **by list index**:

```java
List<EntityInterface> entitiesOfType =
    Entity.getEntities(entry.getValue(), fieldsToFetch, NON_DELETED);
for (int i = 0; i < entitiesOfType.size(); i++) {
  entitiesById.put(
      entry.getValue().get(i).getId(),
      new EntityWithType(entitiesOfType.get(i), entityType));
}
```

That holds only if the bulk fetch returns one row per requested ref, in request order. It does neither — the call reaches `EntityDAO.findByIds`:

```sql
SELECT id, json FROM <table> WHERE id IN (<ids>) <cond>
```

No `ORDER BY`, so row order is whatever the plan produces (index vs sequential scan, buffer state, concurrency); and `<cond>` is the `NON_DELETED` filter, so the result can be shorter than the request.

## Two failures this caused

**Reordered result → relationship order is not preserved.** With equal counts the mapping is a permutation, so `entities[j]` ends up being `dbResult[j]` — the returned *set* is correct, but rows come back in database order rather than the order the ports were added.

**Short result → trailing ids never mapped.** Only the first `dbResult.size()` input ids get keyed, so the remainder are dropped by the `if (entity != null)` guard below, while `total` comes from an independent `countFindTo` and still counts them. The payload then contradicts its own count. **This is the one with user-visible impact.**

> Correction: an earlier revision of this description claimed a port could render with another asset's name and href. That is not right — a row *is* the entity, so there is no separate port identity to mismatch. The order and the count contradiction are the real defects.

## Evidence

The second failure is visible in the AUT nightly. The page snapshot from a failing run of `InputOutputPorts › Toggle fullscreen mode` shows:

```
tab   "Input/Output Ports 1" [selected]
      "Input Ports (1)"
group "Ports Lineage":
      "To view lineage, assign assets to this Data Product and add them as input/output ports."
```

The tab badge reads `1` — that is `paging.total` from the count probe — while the lineage panel rendered its **empty** branch, because the same `/portsView` response carried zero rows. `PortsLineageView` derives `hasAnyPorts` from the row array, and its empty branch has no fullscreen button at all, so the test clicked a control that did not exist for 55s.

Because ordering depends on the query plan it is intermittent rather than constant, and engine-agnostic: the AUT data shows the same rate on MySQL and PostgreSQL (that test flaked 19/20, split 9 MySQL / 10 PostgreSQL).

## The change

**1. Key each entity by its own id** — correct regardless of order or count. The entity already carries it, since `jsonToEntity` deserializes it from the stored JSON. This also restores relationship ordering, which the index pairing silently lost.

**2. Make `total` consistent with the rows.** `countFindTo` counts rows in `entity_relationship`, which knows nothing about whether the target still exists — so a soft-deleted port stayed in the count while `NON_DELETED` kept it out of the page. When, and only when, a relationship fails to resolve, the count is recomputed over the full relationship set using **reference-only** reads (`getReferences`), far cheaper than the page's entity fetch. The normal case — everything resolves — costs nothing extra, and the hot path is untouched.

The drop is also logged once per request with the count and ids, so a data product with many dangling references leaves a trace without flooding the log.

## Testing

⚠️ **Not compiled locally.** Two attempts failed for reasons unrelated to this change and I could not get past them:

- Full reactor build (`-pl openmetadata-service -am`) fails in the **`common`** module — a module this diff does not touch — with `ExceptionInInitializerError: com.sun.tools.javac.code.TypeTag :: UNKNOWN`, a Lombok/JDK incompatibility on the local JDK 21.0.10.
- Building the module alone fails on missing generated sources from `openmetadata-spec`.

`DataProductRepository.java` appears in neither error set. The change is a five-line rewrite of one loop plus a log statement, using `EntityInterface.getId()` (already used elsewhere in the file's dependencies) — but **CI is the compile gate here, not me.** Please don't merge on my say-so that it builds.

**Regression test added:** `test_softDeletedPortAssetIsNotCountedInTotal`, following the conventions of the tests around it.

Why the suite missed this:

- `test_getPortsViewCombined` asserts counts only, and nothing in its fixture is unresolvable, so counts always matched.
- `test_deletingAssetRemovesItFromPorts` **hard**-deletes, which cascades the `entity_relationship` row away — so the count drops naturally and the inconsistency never arises.

A **soft** delete is the case that breaks: the relationship row survives and is counted, while the entity is filtered out of the rows. The new test asserts the page returns the surviving port and that `total` matches what was returned.

Not covered by a test: the ordering half of the fix. Asserting insertion order only fails when the database happens to return a divergent order, so as a gate it is a coin flip rather than a guard. Open to suggestions if you'd rather see it pinned some other way.

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DataProductResourceIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DataProductResourceIT.java
@@ -3203,6 +3203,48 @@ public class DataProductResourceIT extends BaseEntityIT<DataProduct, CreateDataP
   }
 
   @Test
+  void test_softDeletedPortAssetIsNotCountedInTotal(TestNamespace ns) throws Exception {
+    Domain domain = getOrCreateDomain(ns);
+
+    CreateDataProduct create =
+        new CreateDataProduct()
+            .withName(ns.prefix("dp_soft_deleted_port"))
+            .withDescription("Data product for soft-deleted port count consistency")
+            .withDomains(List.of(domain.getFullyQualifiedName()));
+    DataProduct dataProduct = createEntity(create);
+
+    Table softDeletedTable = createTestTable(ns, "soft_del_input", domain);
+    Table survivingTable = createTestTable(ns, "soft_del_surviving", domain);
+
+    bulkAddInputPorts(
+        dataProduct.getFullyQualifiedName(),
+        new BulkAssets()
+            .withAssets(
+                List.of(
+                    softDeletedTable.getEntityReference(), survivingTable.getEntityReference())));
+
+    assertEquals(2, getInputPorts(dataProduct.getId(), 10, 0).getPaging().getTotal());
+
+    // Soft delete, unlike the hard delete in test_deletingAssetRemovesItFromPorts:
+    // the entity_relationship row survives, so countFindTo still counts this port
+    // while the NON_DELETED filter keeps the entity out of the returned rows. The
+    // page then advertised a port it could not carry -- the ports tab rendered
+    // "Input Ports (1)" above an empty list.
+    SdkClients.adminClient()
+        .tables()
+        .delete(softDeletedTable.getId().toString(), Map.of("hardDelete", "false"));
+
+    ResultList<Map<String, Object>> inputPorts = getInputPorts(dataProduct.getId(), 10, 0);
+    assertEquals(1, inputPorts.getData().size());
+    assertEquals(survivingTable.getId(), getEntityId(inputPorts.getData().get(0)));
+    assertEquals(inputPorts.getData().size(), inputPorts.getPaging().getTotal());
+
+    DataProductPortsView portsView = getPortsView(dataProduct.getId(), 10, 0, 10, 0);
+    assertEquals(1, portsView.getInputPorts().getData().size());
+    assertEquals(1, portsView.getInputPorts().getPaging().getTotal());
+  }
+
+  @Test
   void softDeletedExpert_notReturnedInSingleGet(TestNamespace ns) {
     OpenMetadataClient client = SdkClients.adminClient();
     Domain domain = getOrCreateDomain(ns);
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataProductRepository.java
@@ -643,22 +643,80 @@ public class DataProductRepository extends EntityRepository<DataProduct> {
       String entityType = entry.getKey();
       List<EntityInterface> entitiesOfType =
           Entity.getEntities(entry.getValue(), fieldsToFetch, NON_DELETED);
-      for (int i = 0; i < entitiesOfType.size(); i++) {
-        entitiesById.put(
-            entry.getValue().get(i).getId(), new EntityWithType(entitiesOfType.get(i), entityType));
+      // Key each entity by its own id. Pairing the result with the request list
+      // by index assumed the bulk fetch answered in request order and returned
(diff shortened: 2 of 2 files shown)


ShaileshParmar11:
Closing this — the flaky tests are being triaged and assigned to their owning teams instead of fixed in one batch. The RCA for each is preserved in the PR description and commit message above, and issue #31495 stays open for the `getPaginatedPorts` backend defect.

PR #31502, opened 2026-08-14
Fix search relevance: exact names losing to near matches

harshach:
An exactly-named asset could rank below assets that merely start with the query. Two causes: the text ranking stages were a multi_match boosted by their weight, so a stage scored BM25 x weight and was unbounded, while exactName is a constant_score capped at 100 -- under dis_max a stage configured at 24 measured over 105 and won. And om_compound_analyzer split "customers7" into [customers7, customers, 7], so with fuzziness the query matched two terms in the sibling and Lucene's blended-freq rewrite summed them.

Auditing the rest of the query surface against a real corpus turned up six more defects, all of which broke matching or ordering before ranking could help:

- om_analyzer lowercased before word_delimiter ran, so split_on_case_change never saw a case boundary and CustomerAddress indexed as one opaque token.
- kstem leaves orders, sessions, transactions and regions unstemmed while stemming customers and items, so a singular query missed a quarter of the plural names in a catalog. om_plural_stemmer closes those cases behind it.
- om_compound_analyzer did not stem, so it disagreed with its base field about the same document.
- displayName falls back to name and name.compound now analyses like name, so the identity stages restated one piece of evidence up to four times and a tie_breaker turned field count into a relevance signal.
- om_ngram has a three character minimum and encodes no notion of "starts with", so partially typed names ranked arbitrarily and anything shorter matched nothing. Added a prefixName stage using match_bool_prefix.
- No entity-type prior existed, and the column index carries two orders of magnitude more documents than the table index.
- stageFieldWeights normalised against the stage's own maximum, so a field's weight depended on what it shared a stage with, and the cross-entity merge let one entity type's niche field set the ceiling for everyone.

Also incorporates PR #31228, which stops the fuzzy stage widening recall on identifier lookups.

The mapping edits change the content hash, so IndexMappingVersionTracker triggers a reindex on upgrade; that is required for the analyzer fixes to take effect.

### Describe your changes:

Fixes #<issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Documentation

#
### High-level design:

N/A — small change. 

#
#### Manual testing performed

#

Not applicable. 

#
### Checklist:


chirag-madlani (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRankingHelper.java
```diff
@@ -485,4 +617,186 @@ private static boolean isSignificantToken(String token, Set<String> stopWords) {
...
+      SearchSettings searchSettings,
+      int from,
+      int size,
+      SearchPass<R> pass,
+      Function<R, Stream<String>> identifiersOf)
+      throws IOException {
+    if (!hasPrunableFuzzyStage(searchSettings)) {
+      return pass.run(searchSettings, from, size);
+    }
+    boolean requestedWindowIsTheProbe = from == 0 && size >= IDENTITY_PROBE_SIZE;
```
tomasmontielp: shall we add a `&& request has no search_after`? cursor requests also arrive with from=0, so this shortcut would treat their scrolled window as the top-10 check, the pruned/widened decision can still flip between pages for API clients paging with search_after. The dedicated probe would also need search_after cleared so it reads the actual top

tomasmontielp (dismissed):

PR #31504, opened 2026-08-14
fix(test): drop getByRole chain in EditClassification — testid is on input

anuj-kumary:
## Summary

- `getByTestId('name').getByRole('textbox')` timed out with **element(s) not found** in both `EditClassification` tests
- Root cause: Playwright's `getByRole` searches **descendants only** — the `<input data-testid="name">` element has no children, so the chained `getByRole('textbox')` finds nothing
- Fix: drop the redundant `.getByRole('textbox')` chain — the testid is already on the `<input>` element, and `toHaveValue` / `toBeDisabled` work directly on it

## Test plan

- [ ] `Edit a user classification from the manage button` — name field has expected value
- [ ] `System classification name is disabled in the edit drawer` — name field is disabled for system classifications


anuj-kumary:
Tests already pass on main — the issue is specific to 1.13 where TEXT_MUI puts data-testid inside inputProps (on the raw input). Closing this PR.

PR #31505, opened 2026-08-14
fix(test): drop getByRole chain in EditClassification — testid is on input

anuj-kumary:
## Summary

- `getByTestId('name').getByRole('textbox')` timed out with **element(s) not found** in both `EditClassification` tests
- Root cause: Playwright's `getByRole` searches **descendants only** — the `<input data-testid="name">` element has no children, so the chained `getByRole('textbox')` finds nothing
- Fix: drop the redundant `.getByRole('textbox')` chain — the testid is already on the `<input>` element, and `toHaveValue` / `toBeDisabled` work directly on it

Backport of #31504


anuj-kumary:
Superseded by #31506 — the real fix is moving data-testid from inputProps to top-level props in tagFormFields.tsx, not changing the test.

PR #31529, opened 2026-08-14
fix(agents): distinguish loading from an unreachable pipeline service

aniketkatkar97:
### Describe your changes:

The agents lists are gated on `GET /services/ingestionPipelines/status`, but neither the window while
that call is in flight nor an unreachable pipeline service was represented in the UI. Both showed a
terminal state that was factually wrong.

**1. The status read as "known and unavailable" before it was ever fetched.**
`AirflowStatusProvider` seeded `isLoading = false` and `isAirflowAvailable = false`, so for the paint
before its effect fired, every consumer saw "not loading, Airflow is down" and rendered its error
state. It now seeds `isLoading = true`, and the `!currentUser?.id` branch stays loading rather than
reporting a status it never asked for — the effect re-runs once the user resolves.

**2. The whole agents tab was replaced by the Airflow setup guide during the status call.**
`Ingestion.component` returned `ErrorPlaceHolderIngestion` on `!isAirflowAvailable` with no
`isFetchingStatus` check, so the "Install Airflow API" guide flashed on every load — including over
the Collate AI sub-tab, which does not depend on Airflow at all. It now waits for the status to
actually answer before showing the guide, and hides `DeploymentSummaryCard` while the list is still
unknown (its counts otherwise read as "0 agents").

**3. An unreachable pipeline service was reported as "No agents set up yet".**
When the service is down the pipeline fetch is skipped entirely, so the list is empty because nothing
was fetched — not because the service has no agents. `MetadataAgentsView` and `IngestionListTable` now
render `ErrorPlaceHolderIngestion` in that case, which already branches on platform: the managed-Airflow
guide for Airflow deployments, and "The Ingestion Scheduler is unable to respond" for everything else.
Note `getErrorPlaceHolder`'s only signal was `platform === DISABLED`, and an unreachable Airflow reports
`'Airflow'` or `'unknown'`, so even its existing disabled-platform copy never fired for this case.

**4. First loads now show placeholder cards.**
`AgentGroup` gains `isLoading` / `skeletonCount` and renders `AgentCardSkeleton` rows in place of the
empty placeholder, keeping its header mounted so the refresh control cannot vanish mid-flight.
`ServiceDetailsPage` seeds its pipeline and Collate AI loading flags to match, and releases them on the
paths that never fetch (deleted service, non-DB service).

**No new i18n keys** — every string on the unavailable path already existed in `ErrorPlaceHolderIngestion`.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

The branch is deliberately kept inside the two existing empty-state call sites rather than extracted
into a shared component — it is three lines each, and both components already read the Airflow status
they need (`MetadataAgentsView` via `useAirflowStatus()`, `IngestionListTable` via its
`airflowInformation` prop).

Reusing `ErrorPlaceHolderIngestion` rather than adding new copy means the platform branch (Airflow vs.
Argo/other) lives in exactly one place and stays correct for both. `getErrorPlaceHolder` and
`ErrorPlaceHolderIngestion` keep their existing signatures, so no other caller is affected.

`Ingestion.component.tsx` intentionally keeps its whole-tab early return — it already handled the
unavailable case correctly; only the missing `isFetchingStatus` guard was added.

#
#### Use cases covered
- Loading a service's agents tab with the pipeline service reachable → placeholder cards, then the real list
- Loading it with the pipeline service unreachable → the scheduler/Airflow message, not "No agents set up yet"
- Airflow platform vs. Argo/other → the two branches of `ErrorPlaceHolderIngestion` copy
- Status call still in flight → neither the "no agents" nor the "unreachable" state is shown
- Service with no agents and a healthy pipeline service → the existing "No agents set up yet" copy, unchanged

#### Unit tests
- [x] Added/updated for the new logic.
- Files added: `AgentCardSkeleton.component.tsx` + `AgentCardSkeleton.test.tsx`
- Files updated: `AgentGroup.test.tsx`, `MetadataAgentsView.test.tsx`, `Ingestion.test.tsx`, `IngestionListTable.test.tsx`
- `npx jest src/components/ServiceAgents src/components/Settings/Services/Ingestion src/utils/IngestionUtils.test.tsx src/context/AirflowStatusProvider src/pages/ServiceDetailsPage src/components/common/ErrorWithPlaceholder src/components/common/AirflowMessageBanner`
  → **29 suites / 398 tests passing**

#### Playwright (UI) tests
- Not added. The change is a rendering branch on an existing API response and is fully covered by the
  unit tests above; reproducing it E2E needs the pipeline service stopped, which the Playwright stack
  does not currently model.

#### Manual testing performed
1. Throttled `GET /api/v1/services/ingestionPipelines/status` in DevTools and reloaded a service's
   agents tab — placeholder cards for the whole window, no "Install Airflow API" flash.
2. Stopped the pipeline service on a service that has agents — the agents tab now explains the
   scheduler is unreachable instead of claiming there are none.

#

Not attached — flagging this so a reviewer can ask for one. The visible deltas are: placeholder cards
during first load, and the scheduler/Airflow message replacing "No agents set up yet" when the pipeline
service is down.

#
### Checklist:
- [x] For JSON Schema changes: I updated the migration scripts or explained why it is not needed. (No schema changes.)

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/ServiceAgents/components/AgentCardSkeleton.component.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/ServiceAgents/components/AgentCardSkeleton.component.tsx
@@ -1,0 +1,57 @@
+/*
+ *  Copyright 2026 Collate.
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import { Box, Card, Skeleton } from '@openmetadata/ui-core-components';
+import { FC } from 'react';
+
+/**
+ * Placeholder row for {@link AgentCard}. The outer chrome and the three column
+ * widths are copied from the real card so the list does not reflow when the
+ * agents arrive.
+ */
+const AgentCardSkeleton: FC = () => (
+  <Card
+    className="tw:rounded-2xl tw:border tw:border-secondary tw:bg-primary tw:px-4.5 tw:py-4 tw:shadow-xs"
+    data-testid="agent-card-skeleton"
+    variant="ghost">
+    <Box align="center" className="tw:gap-3.5">
+      {/* identity */}
+      <Box
+        align="start"
+        className="tw:w-[30%] tw:min-w-[300px] tw:max-w-[520px] tw:shrink-0 tw:gap-3">
+        <Skeleton
+          className="tw:size-9.5 tw:shrink-0 tw:rounded-xl"
+          variant="rectangular"
+        />
+        <div className="tw:min-w-0 tw:flex-1">
+          <Skeleton height={14} width="60%" />
+          <Skeleton className="tw:mt-1.5" height={10} width="35%" />
+        </div>
+      </Box>
+
+      {/* live status zone */}
+      <div className="tw:min-w-0 tw:flex-1">
+        <Skeleton height={22} variant="rounded" width={96} />
+        <Skeleton className="tw:mt-2" height={10} width={180} />
+      </div>
+
+      {/* actions */}
+      <Box align="center" className="tw:shrink-0 tw:gap-2">
+        <Skeleton height={32} variant="rounded" width={88} />
+        <Skeleton className="tw:size-8" variant="rounded" />
+      </Box>
+    </Box>
+  </Card>
+);
+
+export default AgentCardSkeleton;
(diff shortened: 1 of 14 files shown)


aniketkatkar97:
Confirmed and fixed in 4aec032a52 — thanks, this was a real gap and the finding is exactly right.

`AirflowStatusProvider.tsx` was reverted to `main` while removing unrelated local scaffolding just
before I pushed, so the seed the PR description promised never made it into the diff. As you note,
that left all three guards evaluating true on first paint, which is worse than the original bug: the
agents list would have claimed the pipeline service was unreachable on every load.

The commit seeds `isLoading = true` and keeps it true on the `!currentUser?.id` early return, matching
your suggested patch.

Worth calling out why the existing tests stayed green: the effect calls `setIsLoading(true)`
synchronously before awaiting, so `AirflowStatusProvider.test.tsx`'s initial
`isFetchingStatus === 'true'` assertion passes under either seed, and the consumer tests all mock
`useAirflowStatus` / pass `airflowInformation` explicitly, so none of them exercise the real provider.
Added a test for the window the seed actually governs — `currentUser` unresolved, status never
requested — which fails against `useState(false)` and passes with the seed. Verified both directions
locally.

chirag-madlani (approved):

PR #31537, opened 2026-08-14
fix(test): eliminate race conditions in DataContractsSemanticRules

shrabantipaul-collate:
## Summary

Fixes 31 race conditions in DataContractsSemanticRules.spec.ts where tests manually trigger contract validation and immediately reload without waiting for the async validation to complete.

## The Problem

The test pattern was:
```typescript
const runNowResponse = page.waitForResponse('/api/v1/dataContracts/*/validate');
await page.getByTestId('contract-run-now-button').click();
await runNowResponse;  // ← resolves when HTTP *request* is accepted, not when validation *finishes*
await page.reload();   // ← reloads before result is ready
```

`runNowResponse` resolves as soon as the backend accepts the trigger request, not when the async validation completes. The page reloads while validation is still running, so the UI shows stale latestResult from the previous step.

## The Fix

Replace all 31 manual runNow + reload patterns with:
```typescript
await triggerContractValidation(page, contractId);  // polls API until validation reaches terminal state
await page.reload();
```

The existing `triggerContractValidation` helper (already in dataContracts.ts) calls `pollContractStatus` which waits until latestResult.status reaches a terminal state (Success, Failed, Aborted, or PartialSuccess) before returning.

## Impact

- **Fixes 2 confirmed flaky tests** in the nightly workflow:
  - Validate Description Rule Contains (line 646)
  - Validate Description Rule Is_Not_Set (line 912)
- **Prevents the same race in 29 other test steps** across the entire suite
- **Better determinism under load** — test steps that modify entity state can now safely re-validate and see the updated result

## Files Changed

- `openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContractsSemanticRules.spec.ts`
  - Added `triggerContractValidation` to imports
  - Added `let contractId` hoist to 30 test blocks
  - Captured contract ID from `saveAndTriggerDataContractValidation` return value
  - Replaced 31 racy runNow + reload blocks with `triggerContractValidation(page, contractId)`

This PR makes semantic data-contract Playwright tests wait for validation completion before reloading.
- Captures the created contract identifier from the initial save-and-validation helper.
- Replaces repeated run-now response waits with the shared validation polling helper.
- Removes duplicated UI interaction sequences from 31 validation steps.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContractsSemanticRules.spec.ts | Replaces racy manual validation triggers with contract-status polling across the semantic-rule test suite. |

</details>

<sub>Reviews (2): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/semantic-ru..."](https://github.com/open-metadata/openmetadata/commit/af3f62c65a7d87e9392499f632c376f1346a16d8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53306213)</sub>


shrabantipaul-collate:
Thanks for catching that @gitar-bot! I've addressed the issue by adding `expect(contractId).toBeTruthy()` assertions after every `contractId` capture.

Now if `saveAndTriggerDataContractValidation` ever returns a shape without `id` or returns undefined, the test will fail loudly with an explicit assertion error instead of silently degrading to the fire-and-forget pattern.

This is applied to all 30 test blocks that capture contractId, ensuring the fix is robust against future changes to the response shape.

shrabantipaul-collate:
Updated: Also added a fix for Bug 1 (Elasticsearch indexing lag in the `Contains` test).

Before the first validation runs, the test now polls the search API (`/api/v1/search/query`) to verify the table has been indexed in Elasticsearch. Only after the table appears in search results does the test proceed with the validation.

This eliminates the race where the validation engine queries ES for the 'contains' match and finds nothing because the table hasn't been indexed yet.

The polling:
- Checks every 1-3 seconds for up to 30 seconds
- Searches for the table by name in the `table_search_index`
- Passes once the table appears in results
- Gives ES time to catch up without modifying backend validation logic

This is applied only to the 'Contains' test step that validates the 'contains' operator, which is the one affected by ES lag.

shrabantipaul-collate:
Good catch @gitar-bot! The ES wait was in the wrong place.

**Issue:** I was waiting for indexing *after* validation had already run and persisted its result. By then it was too late.

**Fix:** Moved the ES-indexing poll to *before* the `saveAndTriggerDataContractValidation` call. Now:
1. Test polls `/api/v1/search/query` until the table appears in results (confirms ES indexing complete)
2. **Then** calls `saveAndTriggerDataContractValidation` so validation runs against an indexed table
3. The 'contains' rule now reliably finds the table in ES

This ensures the validation runs with the table already indexed, eliminating the ES lag race that caused the original flakiness.

shrabantipaul-collate:
Extended the Elasticsearch indexing fix to cover all text-search operators:

**Text-search operators affected by ES lag:**
- ✅ `contains` — Fixed for 'Description with contains condition should passed'
- ✅ `not_contains` — Fixed for both 'Description with not_contains condition should failed' and 'Description with not_contains condition should passed'

**Other operators are NOT affected** (they read from entity properties, not ES):
- `is`, `is_not`, `is_set`, `is_not_set` — Read properties directly
- `any_in`, `not_in`, `between`, `not_between` — Read properties directly  
- `less`, `greater`, `less_equal`, `greater_equal` — Read properties directly

So the ES indexing wait is now applied to all places where it's needed (contains/not_contains), preventing the ES lag race for all text-search validations.

shrabantipaul-collate:
Refactored to reduce duplication:

**Created reusable helper function** in `dataContracts.ts`:
```typescript
export const waitForTableIndexing = async (
  page: Page,
  tableName: string,
  timeoutMs = 30_000
): Promise<void>
```

**Benefits:**
- Single source of truth for ES indexing logic
- Easier to maintain and extend
- No code duplication across tests
- Clear, self-documenting function name

**Usage in tests is now simple:**
```typescript
await waitForTableIndexing(page, table.entity.name);
await saveAndTriggerDataContractValidation(page, true);
```

This applies to all text-search operators (`contains`, `not_contains`) without repeating the polling logic in each test step.

shrabantipaul-collate:
Fixed UI Checkstyle issues in `dataContracts.ts`:
- Reformatted long lines to stay within line-length limits
- Improved code readability with better line breaks

The Playwright ESLint check should now pass.

shrabantipaul-collate:
Reverted the helper function approach. The ES indexing wait logic is now inlined directly in the test file to avoid lint/format issues in the utilities file. This keeps the fixes focused and eliminates unnecessary abstractions.

chirag-madlani (approved):

PR #31543, opened 2026-08-14
fix(backend): column description updation with overrideMetadata=true

### LINKED ISSUE #31544: Trino: empty column comments ingested as "", and updated column comments ignored with overrideMetadata=true
### Connector

Trino

### Feature area

Metadata ingestion

### Describe the bug

Two defects that together cause Trino column descriptions to silently stop syncing.

**1. Columns with no comment are ingested with an empty-string description**

Trino's `SHOW COLUMNS` reports a column with no comment as the empty string, never `NULL`. `_get_columns` in `ingestion/src/metadata/ingestion/source/database/trino/metadata.py` passes `record.Comment` straight through to `description`, so every uncommented column lands in OpenMetadata with `"description": ""` rather than no description at all. An explicit `COMMENT ''` produces the same value, so the two cases are indistinguishable at that layer.

**2. Updated column comments are not ingested even with `overrideMetadata: true`**

Once a column has a non-empty description, a re-run never updates it. The entity-level `updateDescription()` in `EntityRepository.java` honours the `overrideMetadata` flag, but the column-level `updateColumnDescription()` does not — it unconditionally reverts a bot PUT whenever the stored column description is non-empty. `updateColumnDisplayName()` has the same gap.

The two defects mask each other. A column with no comment is stored as `""`, which is `nullOrEmpty`, so its *first* comment does land — making the sync look functional right up until a second update to an already-described column is silently dropped.

### To Reproduce

**Part 1 — blank descriptions**

1. In Trino, create a table mixing all three comment states:
   ```sql
   CREATE TABLE memory.testdb.student_data (
       id BIGINT,
       age INTEGER COMMENT '',
       name VARCHAR COMMENT 'Name of the student'
   );
   ```
2. Run metadata ingestion for the Trino service.
3. Fetch the table's columns from the API. `id` and `age` both come back as `"description": ""` instead of having no description.

**Part 2 — updates dropped**

4. Create a table where one column has a comment and another does not:
   ```sql
   CREATE TABLE memory.default.fruits (
       type VARCHAR,
       price DOUBLE COMMENT 'price of fruit'
   );
   ```
5. Run ingestion. `type` is stored as `""`, `price` as `price of fruit`.
6. Change both comments in Trino:
   ```sql
   COMMENT ON COLUMN memory.default.fruits.type IS 'first comment on fruit type';
   COMMENT ON COLUMN memory.default.fruits.price IS 'second comment on fruit type';
   ```
7. Re-run ingestion with `overrideMetadata: true`.
8. `type` updates correctly, but `price` keeps the stale `price of fruit`.

As a control, ingesting the same Trino source into a **brand-new** service produces the correct description for both columns — that path is an insert, so no merge guard runs. This confirms the connector reads the updated comments correctly and the loss is in the server-side column merge, not in extraction.

### Expected behavior

1. A column with no comment in Trino should be ingested with **no description**, not an empty string, so it is not shown as described and does not interfere with later updates.
2. With `overrideMetadata: true`, a changed source comment should overwrite the stored column description, matching the behaviour already provided for the entity-level description.

```yaml
source:
  type: trino
  serviceName: <redacted>
  serviceConnection:
    config:
      type: Trino
      hostPort: <redacted>
      username: <redacted>
      authType:
      catalog: memory
  sourceConfig:
    config:
      type: DatabaseMetadata
      overrideMetadata: true
sink:
  type: metadata-rest
  config: {}
workflowConfig:
  loggerLevel: DEBUG
  openMetadataServerConfig:
    hostPort: <redacted>/api
    authProvider: openmetadata
    securityConfig:
      jwtToken: <redacted>
```

### OS

macOS 26.5.1

### Python version

3.11.15

### OpenMetadata version

2.0.0-SNAPSHOT (main)

### Additional context

Root causes located at:

- `ingestion/src/metadata/ingestion/source/database/trino/metadata.py` — `_get_columns` passes the empty-string comment through unchanged. `TRINO_TABLE_COMMENTS` filters `is not null` but not `COMMENT ''`, so the table comment has the same gap.
- `openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java` — `ColumnEntityUpdater.updateColumnDescription` / `updateColumnDisplayName` omit the `!overrideMetadata` condition that `updateDescription()` and `updateDisplayName()` both have.

Note that fixing only the ingestion side is not sufficient: it stops new blank descriptions but leaves existing non-empty column descriptions permanently un-updatable by ingestion. Both sides are needed.

I have a fix ready and will open a PR linking this issue.

### Pre-submission checklist

harshsoni2024:
### Describe your changes:

Fixes #31544

I worked on the backend half of #31544: once a table column had a non-empty
description, ingestion could never update it again, even with
`overrideMetadata: true`.

The entity-level `updateDescription()` in `EntityRepository` consults the
`overrideMetadata` flag that the bulk path sets from
`PUT /v1/tables/bulk?overrideMetadata=true`. The column-level
`updateColumnDescription()` did not — it unconditionally reverted a bot PUT
whenever the stored column description was non-empty, so a changed source
comment could never reach an existing column. `updateColumnDescription()` now
consults the flag, matching the entity-level behaviour.

The guard keeps one condition that the entity-level version does not:
`overrideMetadata` may **replace** a column description but never **blank** it.

```java
&& (!overrideMetadata || nullOrEmpty(updatedColumn.getDescription()))
```

A connector that finds no comment on a column omits the field from the payload
entirely, so the incoming description arrives as `null`. Without that clause, an
override run would read the absence as a deletion and wipe hand-written column
documentation on every uncommented column. Clearing a column description
therefore still requires a PATCH (UI edit or API) — ingestion is replace-only.

`overrideMetadata` deliberately does **not** extend to the column `displayName`.
Connectors rarely send one, so honouring the flag there would blank curated
column display names on nearly every override run. `updateColumnDisplayName()` is
left as-is and covered by a test that pins that asymmetry.

The `overrideMetadata` field moves from `private` to `protected` because
`ColumnEntityUpdater` is a subclass of `EntityUpdater` and a private superclass
field does not resolve by simple name.

**Companion PR:** the connector-side half of #31544 (Trino reporting an unset
column comment as `''`, so uncommented columns were ingested with
`description: ""`) is on branch `trino_empty_column_comment_fix` and ships
separately. Both are needed to close the issue: this PR lets an updated comment
reach an already-described column, and the connector PR stops blank descriptions
being written in the first place.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered
- Re-running ingestion with `overrideMetadata: true` propagates an updated source
  comment onto a column that already has a non-empty description
- Re-running ingestion **without** `overrideMetadata` still preserves a
  user-curated column description (no regression to the existing protection)
- An `overrideMetadata` run that supplies no description for a column does not
  blank the stored one
- A curated column `displayName` survives an `overrideMetadata` run

#### Unit tests
- [x] Not applicable — the change is a merge-path condition on the bulk PUT
  endpoint, exercised end-to-end by the integration tests below rather than by a
  unit test. `EntityUpdater` is an inner class reached through the repository, so
  a unit test here would assert mock wiring rather than behaviour.

- [x] I added integration tests in `openmetadata-integration-tests/`.
- File updated:
  `openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BulkOverrideMetadataIT.java`
  — four column-level cases alongside the existing table-level ones:
  - `test_botCannotOverwriteColumnDescription_withoutOverride`
  - `test_botOverwritesColumnDescription_withOverride`
  - `test_overrideDoesNotBlankColumnDescription`
  - `test_columnDisplayNamePreserved_evenWithOverride`

- [x] Not applicable (no ingestion changes on this branch).

#### Playwright (UI) tests

#### Manual testing performed

Local Trino (`memory` catalog) against a locally built OpenMetadata stack, with
`overrideMetadata: true` in the ingestion config. The companion connector fix was
also applied locally, since reproducing the original report end to end needs
both.

1. Table `memory.default.fruits` with a `price` column carrying the comment
   `'price of fruit'`. Ran ingestion; OpenMetadata showed
   `price → 'price of fruit'`.
2. Changed the comment in Trino:
   ```sql
   COMMENT ON COLUMN memory.default.fruits.price IS 'second comment on fruit type';
   ```
3. Re-ran ingestion with `overrideMetadata: true` against the **existing**
   service. Before this change, `price` kept the stale `'price of fruit'`.
4. As a control, ingested the same source into a brand-new service. All comments
   were correct there — that path is an insert, so no merge guard runs. This
   confirmed the loss was in the server-side column merge, not in extraction.
5. Rebuilt the backend with this change, redeployed, re-ran on the existing
   service: `price` now updates to `'second comment on fruit type'`.

#

Not applicable.

#
### Checklist:
- [x] My PR is linked to a GitHub issue via `Fixes #31544` above.
- [x] I have added tests (integration) and listed them above.

This PR allows bot-driven bulk ingestion with `overrideMetadata=true` to replace existing column descriptions while preserving them when the incoming description is absent or empty.
- Exposes the updater’s override flag to column-level update logic.
- Adds integration coverage for override, non-override, absent-description, and display-name behavior.
</details>

The PR appears safe to merge, with no concrete blocking or non-blocking defects identified.

The override flag is set on a fresh updater for each bulk entity, remains disabled on ordinary PUTs, and the revised guard preserves existing descriptions for absent or empty incoming values while permitting explicit non-empty replacements.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java | Correctly scopes column-description replacement to bot bulk updates with override enabled while retaining the no-blank safeguard. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BulkOverrideMetadataIT.java | Adds synchronous persisted-state integration coverage for all intended column-level override behaviors. |

</details>

<sub>Reviews (1): Last reviewed commit: ["test(it): pin that overrideMetadata does..."](https://github.com/open-metadata/openmetadata/commit/9fd5369d2494280908099c5d69dc6c2670de0ee4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53326454)</sub>

### DIFF
--- a/ingestion/src/metadata/ingestion/api/delete.py
+++ b/ingestion/src/metadata/ingestion/api/delete.py
@@ -23,13 +23,40 @@ from metadata.ingestion.api.models import Either
 from metadata.ingestion.models.barrier import Barrier
 from metadata.ingestion.models.delete_entity import DeleteEntity
 from metadata.ingestion.ometa.ometa_api import OpenMetadata, T
+from metadata.utils.fqn import quote_name
 from metadata.utils.logger import utils_logger
 
 logger = utils_logger()
 
+SERVICE_SCOPE_KEY = "service"
+
 
 def _default_dispatch_async() -> bool:
     return ingestion_settings.delete_async
+
+
+def _scope_params_as_fqn(params: Optional[Dict[str, str]]) -> Optional[Dict[str, str]]:  # noqa: UP006, UP045
+    """
+    Return ``params`` with the service scope expressed as an FQN instead of a raw name.
+
+    ``deleteStale`` resolves the scope against the stored FQN, which for a service is its quoted
+    name - so a service whose name needs quoting (in practice, one containing a dot) never
+    resolves and nothing is ever deleted. The legacy fallback still needs the raw name, since
+    ``?service=`` matches a service by name, so the original dict is left untouched.
+
+    Only the service scope is converted: every other scope key already arrives as a full FQN, and
+    quoting a multi-part FQN would break it. Quoting is idempotent, so a caller that already
+    passes an FQN is unaffected. Key order is preserved because the scope is read from the first
+    entry.
+    """
+    if not params or SERVICE_SCOPE_KEY not in params:
+        return params
+    service = params[SERVICE_SCOPE_KEY]
+    try:
+        return {**params, SERVICE_SCOPE_KEY: quote_name(service)}
+    except ValueError:
+        logger.warning("Cannot build the FQN of service '%s'; sending the delete scope unchanged", service)
+        return params
 
 
 def delete_entity_from_source(
@@ -55,7 +82,9 @@ def delete_entity_from_source(
     :param entity_type: Pydantic Entity model
     :param entity_source_state: FQNs of the entities produced by the connector this run
     :param recursive: When True, the soft-delete cascades to child entities
-    :param params: single-key scope dict, e.g. {"database": fqn} / {"databaseSchema": fqn}
+    :param params: single-key scope dict, e.g. {"database": fqn} / {"databaseSchema": fqn}. A
+        {"service": name} scope may be passed as the raw service name; it is converted to the
+        service FQN for the server call — see :func:`_scope_params_as_fqn`.
     :param dispatch_async: For the legacy fallback path, route the sink delete through the
         server-side async endpoint (returns 202 + jobId, runs cascade on the server's
         executor) so ingestion does not block on large hierarchies — see issue #4003. The
@@ -69,7 +98,7 @@ def delete_entity_from_source(
     try:
         result = metadata.delete_stale_entities(
             entity=entity_type,
-            scope_params=params,
+            scope_params=_scope_params_as_fqn(params),
             live_fqns=entity_source_state,
             recursive=recursive,
         )
--- a/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
+++ b/ingestion/src/metadata/ingestion/lineage/sql_lineage.py
@@ -259,9 +259,15 @@ def get_table_fqn_from_query_name(
     """
     Method to extract database, schema and table name
     from raw table name used in query
+
+    The split is quote-aware: connectors are free to put a `.` inside a single name
+    component (Dremio flattens nested folder paths into `folder.subfolder`), in which
+    case that component reaches us quoted. Splitting such a name on every `.` would
+    tear the quoted identifier in half and yield components with unbalanced quotes,
+    which `fqn.quote_name` then rejects with `Invalid name "folder`.
     """
 
-    split_table = table_name.split(".")
+    split_table = fqn.split_raw_name(table_name)
     empty_list: List[Any] = [None]  # Otherwise, there's a typing error in the concat  # noqa: UP006
 
     if len(split_table) > 3:
--- a/ingestion/src/metadata/ingestion/source/database/snowflake/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/database/snowflake/metadata.py
@@ -111,7 +111,7 @@ from metadata.ingestion.source.database.snowflake.queries import (
     SNOWFLAKE_GET_STORED_PROCEDURES_AND_FUNCTIONS,
     SNOWFLAKE_GET_STREAM,
     SNOWFLAKE_LIFE_CYCLE_QUERY,
-    SNOWFLAKE_SESSION_TAG_QUERY,
+    set_session_tag_query,
 )
 from metadata.ingestion.source.database.snowflake.semantic_view_metrics import (
     build_metric_request,
@@ -324,7 +324,7 @@ class SnowflakeSource(
             @event.listens_for(self.engine, "connect")
             def _set_query_tag(dbapi_connection, connection_record):
                 cursor = dbapi_connection.cursor()
-                cursor.execute(SNOWFLAKE_SESSION_TAG_QUERY.format(query_tag=query_tag))
+                cursor.execute(set_session_tag_query(query_tag))
                 cursor.close()
 
     def set_partition_details(self) -> None:
--- a/ingestion/src/metadata/ingestion/source/database/snowflake/queries.py
+++ b/ingestion/src/metadata/ingestion/source/database/snowflake/queries.py
@@ -78,7 +78,16 @@ SNOWFLAKE_SQL_STATEMENT = textwrap.dedent(
     """
 )
 
-SNOWFLAKE_SESSION_TAG_QUERY = 'ALTER SESSION SET QUERY_TAG="{query_tag}"'
+SNOWFLAKE_SESSION_TAG_QUERY = "ALTER SESSION SET QUERY_TAG='{query_tag}'"
+
+
+def set_session_tag_query(query_tag: str) -> str:
+    """Return the ALTER SESSION statement setting QUERY_TAG to the given value."""
+    # Snowflake reads backslash escapes inside the literal, so those double first;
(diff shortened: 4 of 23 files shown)


IceS2 (changes requested):
This fixes two separate bugs that surface as one Trino symptom. They have different blast radii and different backport targets.

We should split this in two PRs.

### The Bugs

#### Bug 1 - Python

Trino's SHOW COLUMNS never returns NULL for a column with no comment. It returns an empty string, and it returns that same empty string for an explicit COMMENT ''. We pass it straight to description, so an uncommented column is stored as description: "".

That matters because "" isn't "no description", it's a value. It looks blank in the UI, but the merge logic asks "is there already something here?", gets yes, and treats it as text a person may have written.

#### Bug 2 - Java

`updateColumnDescription()` never consults `overrideMetadata` , so a bot PUT can't update a column description that isn't empty.

-------------------------------------------------------------

### The Python fix on 1.13

  ```
┌─────┬───────────────┬──────────────────────────┬──────────────────────────────┬─────────────────────────────┐
  │  #  │  Comment in   │     Description in       │        Today on 1.13         │     With the Python fix     │
  │     │     Trino     │       OpenMetadata       │                              │                             │
  ├─────┼───────────────┼──────────────────────────┼──────────────────────────────┼─────────────────────────────┤
  │ 1   │ none          │ nothing yet — first      │ saved as "" ❌                │ saved as nothing ✅          │
  │     │               │ ingestion                │                              │                             │
  ├─────┼───────────────┼──────────────────────────┼──────────────────────────────┼─────────────────────────────┤
  │ 2   │ none          │ "" from an earlier run   │ force-sync rewrites "" over  │ left alone                  │
  │     │               │                          │ ""                           │                             │
  ├─────┼───────────────┼──────────────────────────┼──────────────────────────────┼─────────────────────────────┤
  │ 3   │ none          │ text a user typed        │ force-sync overwrites it     │ kept ✅                      │
  │     │               │                          │ with "" ❌                    │                             │
  ├─────┼───────────────┼──────────────────────────┼──────────────────────────────┼─────────────────────────────┤
  │ 4   │ comment added │ "" from an earlier run   │ only lands with force-sync   │ unchanged — still           │
  │     │               │                          │ ⚠                            │ force-sync only             │
  ├─────┼───────────────┼──────────────────────────┼──────────────────────────────┼─────────────────────────────┤
  │ 5   │ comment added │ nothing                  │ lands normally ✅             │ lands normally ✅            │
  ├─────┼───────────────┼──────────────────────────┼──────────────────────────────┼─────────────────────────────┤
  │ 6   │ comment       │ earlier text             │ only lands with force-sync   │ unchanged                   │
  │     │ changed       │                          │                              │                             │
  └─────┴───────────────┴──────────────────────────┴──────────────────────────────┴─────────────────────────────┘
```

### Why we should split the Java bit into another PR

- `EntityUpdater` is the shared merge core for every entity type. A connector two-liner and a change to core merge semantics shouldn't share a review thread or a revert.

- The connector fix is wanted on 1.13 and cherry-picks cleanly. The Java fix is inert on 1.13. Ingestion there PATCHes rather than PUTs.

- It needs one more guard first. Relaxing the check also opens a delete path on main. A column with a user-written description whose source has no comment is protected today by the guard being relaxed. After the change, a force-sync sends null and `recordChange` records a `fieldDeleted`.
 
Worth adding `|| nullOrEmpty(updatedColumn.getDescription())` plus an IT that pins it, so `overrideMetadata` can update but not delete.
 
That's exactly the rule 1.13 enforces client-side: ADD always, REPLACE with override, REMOVE never.

IceS2 (approved):

Issue #31407, opened 2026-08-12
CustomizeWidgets Playwright spec is flaky: parallel tests clobber the shared persona landing-page layout


aniketkatkar97:
### Affected module
Playwright E2E (UI)

### Describe the bug

`playwright/e2e/Flow/CustomizeWidgets.spec.ts` is flaky. Tests fail with a widget that is simply absent from the landing page, e.g.:

```
Flow/CustomizeWidgets.spec.ts:472 › Total Data Assets Widget

Error: expect(locator).toBeVisible() failed
Locator: getByTestId('KnowledgePanel.TotalAssets')
Expected: visible
Timeout: 15000ms
Error: element(s) not found
```

The page snapshot at failure time shows only 4 of the 7 default widgets rendered — `MyData`, `KPI` and `TotalAssets` are missing from the layout entirely (not scrolled out of view, not loading).

### Root cause

Every test in the file shares a single `persona` and a single `adminUser` created in `beforeAll`, and every test ends with a `Test widget customization` step that runs
`removeAndVerifyWidget()` → save → `addAndVerifyWidget()` → save.

Each save is a full rewrite of the `persona.<name>` docStore document via `PUT /api/v1/docStore`, i.e. last-write-wins over the whole widget layout.

`playwright.config.ts` sets `fullyParallel: true`, so the tests in this file run concurrently on different workers. Two workers each read the layout, each drop their own widget from their own stale copy, and each write the whole document back. Widgets belonging to the other test are silently lost, permanently, and any test that later expects one of them fails with "element(s) not found".

Making the persona per-test is not sufficient on its own: the landing page resolves its layout from `currentUser.defaultPersona` (`src/hooks/useApplicationStore.ts`), which is a single field on the user — so a shared user just moves the race onto that field.

### To Reproduce

Run the spec with more than one worker:

```bash
cd openmetadata-ui/src/main/resources/ui
npx playwright test playwright/e2e/Flow/CustomizeWidgets.spec.ts --project=chromium
```

### Expected behavior

Each test owns its own landing-page layout, so concurrent widget add/remove in one test cannot remove widgets another test depends on.

### Version
- OS: macOS
- Python version: N/A
- OpenMetadata version: main
- OpenMetadata Ingestion package version: N/A

PR #31551, opened 2026-08-14
Chore(UI): Fix the CustomizeWidget CustomizeLandingPage tests

aniketkatkar97:
This pull request refactors the landing page widget test utilities in `customizeLandingPage.ts` to improve reliability and clarity when interacting with deferred (lazy-loaded) widgets. The main changes focus on robustly revealing, asserting, and verifying the presence or absence of widgets that may not be immediately present in the DOM due to deferred rendering.

**Improvements to widget detection and interaction:**

* Added `getLandingPageWidgetSlot` to accurately locate widget slots by matching both exact and suffixed layout keys, handling cases where widgets are added dynamically and may have unique IDs.
* Introduced `revealLandingPageWidget` to ensure that widgets are scrolled into view only if their slot or widget DOM node is present, preventing unnecessary or stalled scroll actions.

**Simplification and reliability of widget visibility checks:**

* Refactored `isLandingPageWidgetVisible` to use `revealLandingPageWidget` and a direct visibility assertion, removing manual scrolling and polling logic for more reliable checks.
* Updated `waitForLandingPageWidget` to always reveal the widget before asserting its visibility and waiting for any skeleton loaders to disappear, ensuring that loading is complete before proceeding.

**Correctness of widget removal assertions:**

* Changed `removeAndVerifyWidget` to assert on the absence of the deferred slot rather than the widget itself, ensuring that the test correctly distinguishes between a widget being removed and merely being out of view (deferred).

**Test flow improvements:**

* Simplified `addAndVerifyWidget` to rely on `waitForLandingPageWidget` for web-first assertions, removing redundant polling and making the flow more straightforward and reliable.
Screenshot 2026-08-14 at 9 00 01 PM

### DIFF
--- a/.github/playwright/impact-map.json
+++ b/.github/playwright/impact-map.json
@@ -534,6 +534,25 @@
         "playwright/e2e/Features/ActivityFeedTabBadge.spec.ts",
         "playwright/e2e/Flow/CustomizeWidgets.spec.ts"
       ]
+    },
+    {
+      "_comment": "The customize-landing-page and widget-filter playwright helpers are shared by every landing-page/customization spec. Without this entry a helper-only change (e.g. a wait fix in customizeLandingPage.ts or widgetFilters.ts) counted as an unmapped file, so CustomizeWidgets/CustomizeLandingPage and the other specs that execute the changed helpers never ran on the PR. Specs listed are the exact importers of these modules; DataInsight.spec.ts runs in the Data Insight project.",
+      "sources": [
+        "openmetadata-ui/src/main/resources/ui/playwright/utils/customizeLandingPage.ts",
+        "openmetadata-ui/src/main/resources/ui/playwright/utils/widgetFilters.ts"
+      ],
+      "projects": ["chromium", "Data Insight"],
+      "specs": [
+        "playwright/e2e/Features/ActivityFeed.spec.ts",
+        "playwright/e2e/Features/CuratedAssets.spec.ts",
+        "playwright/e2e/Features/LandingPageWidgets/**/*.spec.ts",
+        "playwright/e2e/Features/NavigationBlocker.spec.ts",
+        "playwright/e2e/Features/SettingsNavigationPage.spec.ts",
+        "playwright/e2e/Flow/CustomizeLandingPage.spec.ts",
+        "playwright/e2e/Flow/CustomizeWidgets.spec.ts",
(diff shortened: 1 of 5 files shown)


ShaileshParmar11 (dismissed):

aniketkatkar97:
Pushed two commits (PR was dequeued to allow the push; auto-merge re-enabled, it will re-queue once checks pass):

1. **bb57f8d** — `fix(e2e)`: consolidates #31566 here. `widgetFilters.ts` waited on `widget.locator('entity-list-skeleton')` — a tag-name selector that never matches (the skeleton renders `data-testid="entity-list-skeleton"`), so all 15 settle-waits were no-ops and the sort/filter flows raced the widget refetches. Now uses `getByTestId` like the already-correct call sites.

2. **5250991** — `ci(playwright)`: the reason CustomizeWidgets/CustomizeLandingPage specs didn't run on this PR — `playwright/utils/customizeLandingPage.ts` (and `widgetFilters.ts`) were **unmapped** in `.github/playwright/impact-map.json`, so a helper-only change ran smoke+canary only. Added a mapping from both helper modules to their exact importer specs (12 specs: CustomizeWidgets, CustomizeLandingPage, PersonaFlow, ActivityFeed, CuratedAssets, LandingPageWidgets/**, NavigationBlocker, SettingsNavigationPage, DataInsight). Verified locally with `select_playwright_tests.py` — this PR's changed files now select all 12 with zero unmapped files, so this very PR's CI run exercises the fixed helpers.

aniketkatkar97:
Pushed **5fad241** — `fix(e2e)`: isolate persona and user per test in CustomizeLandingPage spec (same class as #31407/#31408, root-caused from AUT run 31867679974):

- The file shared one persona/user across tests and only *Add, Remove and Reset* set the default persona (mid-test, via UI). Under `fullyParallel` the file's tests land in different workers (all 4 started within 3 s in that run), so e.g. *Widget drag and drop reordering* ran with `defaultPersona: null` — it saved the persona layout correctly (POST body contains `KnowledgePanel.MyData`) but `/my-data` rendered the stock layout and the visibility assertions failed.
- Ported the per-test `testUser`/`persona` fixtures from #31408; the persona fixture attaches `personas/0` + `defaultPersona` via API PATCH before login. Per-test personas also remove last-write-wins races on the shared layout document.

After this merges to main I'll cherry-pick the squashed commit to `2.0`.

ShaileshParmar11 (approved):

PR #31556, opened 2026-08-14
Fixes 27041: purge profiler data on table hard delete, and stop the orphan sweep deleting live column profiles

### LINKED ISSUE #27041: [BUG] Hard delete of Table does not clean up profiler data from profiler_data_time_series — stale profiles resurface on re-creation

TeddyCr:
### Describe your changes:

Fixes #27041

I fixed two independent defects in the lifecycle of `profiler_data_time_series`, both rooted in the fact that the table is keyed by **FQN hash**, not by entity id.

**(A) The filed bug — hard delete leaks profiler rows, and a re-created table inherits them.**

`TableRepository` never touched `profiler_data_time_series` on delete. Nothing else in the hard-delete path does either: `EntityRepository.cleanup` deliberately skips time-series children and leaves them to the offline `DataRetention` sweep. So hard-deleting a table left every one of its `table.tableProfile`, `table.systemProfile` and `table.columnProfile` rows behind, and because the table is keyed by FQN hash, the next table created at the same FQN (re-ingestion, rename-and-recreate, IaC replay) silently adopted them — `GET /v1/tables/{fqn}/tableProfile/latest` returned the dead table's row counts and column statistics for a brand-new table.

**(B) A second defect found while fixing (A) — the weekly sweep was deleting live column profiles. Please do not skim this one.**

`CollectionDAO.ProfilerDataTimeSeriesDAO.deleteOrphanedRecords` deleted any row where `NOT EXISTS (SELECT 1 FROM table_entity te WHERE te.fqnHash = profiler_data_time_series.entityFQNHash)`, with **no extension filter**. `FullyQualifiedName.buildHash` hashes each FQN *part* (MD5 hex) and joins with `.`, so a table's `fqnHash` always has exactly 4 dot-joined segments (`service.database.schema.table`) while a **column** profile's `entityFQNHash` has 5 or more. The two can never be equal. Every `table.columnProfile` row of every **live, healthy** table was therefore classified as an orphan and deleted on each `DataRetention` run — a weekly cron (`0 0 * * 0`). That is silent data loss of column-level profiling history, and it is arguably more severe than the filed bug.

It went unnoticed because `OrphanedTimeSeriesCleanupIT.profilerDataOrphans` only ever seeded a table-level row.

The fix resolves a column-profile row to its parent table by taking the **first four hash segments** — `SUBSTRING_INDEX(entityFQNHash, '.', 4)` on MySQL, four chained `split_part(...)` on PostgreSQL. Stripping exactly *one* segment would be wrong: nested/struct columns have FQNs like `service.db.schema.table.parent.child`, so their hashes carry 6+ segments and would still fail to match. Non-column extensions keep the previous raw-hash comparison, and genuinely orphaned rows are still swept.

This second defect was **folded into this PR with the maintainer's explicit approval** rather than filed separately, because it lives in the same table, is caused by the same hashing property, and the two fixes interact: (A)'s correctness depends on (B) no longer mis-classifying live rows.

#
### Type of change:

- [x] Bug fix

#
### High-level design:

**(A) Where the purge runs, and the two properties that make it safe.**

The purge is issued from `TableRepository.postDelete`, on the hard-delete branch only, through
`AsyncService.executeDatabaseTask(DatabaseOperation.PROFILER_CLEANUP, …)`. It is **one unbounded
`DELETE` per extension** — no batching, no iteration cap.

It does not run inside `entitySpecificCleanup`, which executes inside `EntityRepository.cleanup`'s
transaction. A table with a long profiling history can own millions of profiler rows; draining them
there holds every row lock for the life of the request, and batching inside a single transaction
buys nothing — the locks are held to the end either way, so it is the same peak cost spread over
~1,700 round-trips instead of one statement.

Two ordering properties make the asynchronous purge safe, and they cover different failures:

1. **It runs after the delete has committed.** A rolled-back delete therefore cannot purge a live
   table's history — those rows legitimately predate the attempt.
2. **It is bounded to profiles recorded at or before the delete** (`timestamp <= :deletedAt`).
   `profiler_data_time_series` carries no table id, so a purge that lands after the FQN has been
   reused cannot otherwise tell the dead table's rows from its successor's. Everything a successor
   records happens after the delete, so the watermark makes the purge safe to run at *any* later
   point, and idempotent when it runs more than once.

Property 2 is what removes the need for a cap: the statement is unbounded because it no longer needs
to be bounded to be correct. An earlier revision of this PR ran the purge synchronously and capped it
at 1,001,000 rows; that cap silently truncated large tables, and truncated rows regain a live parent
if the FQN is reused — reintroducing the exact symptom of this issue at scale. Both are gone.

**Why a prefix delete is required, not three exact deletes.**

Table and system profiles are written under the *table* FQN, so an exact-match delete suffices.
Column profiles are written under each **column** FQN (`TableRepository.addTableProfileData`, and
`EntityProfileRepository.addProfileData` for the standalone endpoint), including nested columns.
Those rows are not enumerable from the table's current column list either — dropping a column does
not delete its profiler rows, so the only complete way to reach them is the FQN-hash prefix
`'<table hash>.%'`. `delete_hardDeletePurgesProfilesOfDroppedColumns` pins that.

**(B) Why the orphan predicate resolves the parent from the outer row.**

Correlating the other way — `entityFQNHash LIKE te.fqnHash || '.%'` — computes the LIKE pattern from
the **inner** row, which leaves no equijoin key, so both planners degrade to a nested loop over
`table_entity` for every profiler row. Resolving the parent from the outer row keeps a single
equality and preserves the hash anti-join.

Measured on a fixture reproducing the real column types, collation and indexes (5,000 tables /
105,250 profiler rows / 250 planted orphans), issuing the exact statement the DAO issues:

| | PostgreSQL 15 | MySQL 8.4 |
|---|---|---|
| `LIKE`-correlated (nested loop) | 32,389 ms | 75,637 ms |
| parent-hash equality (hash anti-join) | **106 ms** | **247 ms** |

Both return the same 250 orphans, with an empty symmetric difference in both directions, on both
engines. The PostgreSQL plans are `Nested Loop Anti Join` with `Rows Removed by Join Filter:
263,697,500` versus `Hash Anti Join`. Cost is O(profiler_rows × tables), so the nested-loop form
stops completing rather than merely slowing on a large catalog — which matters directly here,
because this PR ships no backfill and relies on the sweep to reclaim historical orphans.

#
### Database migration:

Appended to `bootstrap/sql/migrations/native/2.0.0/{postgres,mysql}/schemaChanges.sql`.

**PostgreSQL** — adds `idx_profiler_data_time_series_fqnhash_pattern`, a `text_pattern_ops` index on `entityFQNHash`, built `CONCURRENTLY` so the migration takes no write lock.

Without it, the prefix delete cannot use an index. The only persistent index with `entityFQNHash` leading is the 1.1.5 unique constraint `(entityFQNHash, extension, operation, timestamp)`; it uses the default operator class and the column inherits the database default collation, neither of which qualifies the planner for `LIKE 'prefix%'`. (1.9.9 *does* create `idx_pdts_entityFQNHash_prefix`, but the same script drops it again at the end — it is a migration-time helper, not a persistent index.) So every table hard delete cost at least one sequential scan of `profiler_data_time_series`, and a recursive service delete cost one per table.

Measured on `postgres:15` (`lc_collate=en_US.utf8`, matching managed Postgres/RDS) against the real 1.1.5 DDL with 300k rows / 211 MB, issuing the exact statement the DAO issues with the prefix bound as text the way JDBC binds it:

| Case | Before | After |
|---|---|---|
| Terminal statement, prefix matches nothing | `Parallel Seq Scan`, **24.174 ms**, **11112 buffers** | `Index Scan`, **0.056 ms**, **3 buffers** |
| Prefix matches 3000 rows | `Seq Scan`, 16.579 ms, 3701 buffers | `Bitmap Heap Scan`, 8.647 ms, 1006 buffers |

Index size **2776 kB** against a 211 MB table.

**Why `text_pattern_ops` and not `varchar_pattern_ops`.** `entityFQNHash` is `VARCHAR(768)`, so `varchar_pattern_ops` is the type-matched choice on paper. But the planner normalises `varchar LIKE text` — what every JDBC `setString` bind produces — by casting the column, and the filter in the measured plan is literally `(entityfqnhash)::text ~~ …`. `1.13.0/postgres/schemaChanges.sql` records this repo hitting a production environment where `varchar_pattern_ops` was consequently unused and the table seq-scanned, and the existing `idx_*_fqnhash_pattern` family settled on `text_pattern_ops` for that reason. To be plain: **on PostgreSQL 15 both opclasses planned identically as an `Index Scan` in my fixture**. This is a consistency-and-robustness choice across versions and environments, not a difference this benchmark demonstrates.

**MySQL — deliberately statement-free**, with the reasoning in the file rather than left as an unexplained empty script. `profiler_data_time_series.entityFQNHash` has been `VARCHAR(768) CHARACTER SET ascii COLLATE ascii_bin` since 1.1.5 — a binary collation — so InnoDB already answers a `LIKE` prefix predicate with a range scan on the leading column of the existing unique index. Adding a second index on the same leading column would be write overhead with no read benefit.

**Operator runbook — interrupted `CREATE INDEX CONCURRENTLY`.** An interrupted build leaves an INVALID index behind, and `IF NOT EXISTS` then no-ops against it forever. This is a known pattern-level gap shared with the 1.11.0 and 1.13.0 index migrations. The full detection query and remediation are in the migration file header; in short:

```sql
SELECT c.relname FROM pg_class c JOIN pg_index i ON i.indexrelid = c.oid
 WHERE NOT i.indisvalid AND c.relname = 'idx_profiler_data_time_series_fqnhash_pattern';
DROP INDEX CONCURRENTLY idx_profiler_data_time_series_fqnhash_pattern;
DELETE FROM server_migration_sql_logs
 WHERE version = '2.0.0' AND sqlstatement LIKE '%idx\_profiler\_data\_time\_series\_fqnhash\_pattern%' ESCAPE '\';
```

The version above is **2.0.0**, not 2.1.1 — `fcece67` moved this migration onto the 2.0 train. A
cleanup keyed on 2.1.1 matches nothing.

No data migration is included. With (B) corrected, the existing weekly sweep reclaims genuinely-orphaned historical rows on its own while preserving live column profiles.

#
#### Use cases covered

- Hard-deleting a table purges its table, system and column profiles; a table re-created at the same FQN starts clean instead of serving the dead table's profile.
- **Profiles recorded after the delete survive the purge** — the property that makes the asynchronous purge race-free against FQN reuse.
- Soft-deleting a table preserves all three, and `GET /v1/tables/{fqn}/tableProfile/latest` keeps serving them (it resolves with `Include.ALL`).
- Recursively hard-deleting a database schema purges the profiler data of the tables it contained (bulk cascade path).
- Hard delete purges the profiles of columns that were **dropped from the table before the delete**.
- Hard delete drains a large column-profile history in full.
- The weekly `DataRetention` orphan sweep **preserves** column profiles of live tables — including nested-column profiles — while still deleting rows whose parent table no longer exists.
- The 2.0.0 migration leaves a valid prefix index on PostgreSQL; on MySQL the premises that make the statement-free script correct hold.

#### Unit tests

Not applicable — no unit-testable logic was added. The change is JDBI DAO SQL plus repository wiring, whose behaviour is only meaningful against a real database; asserting it with mocks would test mock wiring rather than whether rows are actually deleted. Covered by integration tests instead.

- [x] Added, in `openmetadata-integration-tests/`.

`TableResourceIT.java`:

| Test | Asserts |
|---|---|
| `delete_hardDeletePurgesProfilerData` | 0 rows for all three extensions after hard delete; a table re-created at the same FQN returns `profile == null`, all `columns[*].profile == null`, and an empty `systemProfile` `data` array |
| `delete_hardDeletePurgeSpareProfilesRecordedAfterTheDelete` | **new** — one row seeded before the watermark and one after; the purge removes only the earlier row, and the survivor is the later one. Seeding the future-dated row up front pins the boundary deterministically instead of racing the async purge |
| `delete_softDeletePreservesProfilerData` | all three row groups still present after a soft delete, and still served by the API |
| `delete_hardDeleteOfSchemaPurgesTableProfilerData` | recursive schema hard delete purges the contained table's rows (bulk cascade path) |
| `delete_hardDeletePurgesLargeColumnProfileHistory` | 1001 seeded rows are drained in full by the single statement |
| `delete_hardDeletePurgesProfilesOfDroppedColumns` | a column-only profile, whose column is then dropped via `PUT /v1/tables`, is still purged on hard delete; asserts its own precondition that dropping a column leaves the rows behind |

`OrphanedTimeSeriesCleanupIT.java`:

| Test | Asserts |
|---|---|
| `profilerDataOrphans` | after `deleteOrphanedRecords`, a live table's table / system / column / **nested-column** rows all survive, while table-level and column-level rows under a non-existent table are removed |
| `profilerPrefixIndexIsAvailable` | PostgreSQL: the migration's index exists and `pg_index.indisvalid`. MySQL: `entityFQNHash` is the leading column of the unique index **and** its collation is `ascii_bin` — the two facts that make the statement-free MySQL script correct |

Every assertion is a row count in `profiler_data_time_series` or a field of an HTTP response body — no mocks, no `verify()`, no `Thread.sleep`. The purge is now asynchronous, so the delete-path assertions await with `Awaitility` rather than asserting inline.

**Results** — MySQL 8 + Elasticsearch, run in both the `isolated-tests` and `parallel-tests` failsafe lanes (each applies the whole native migration chain from scratch against a fresh testcontainer, so this also exercises the migration end to end):

```
TableResourceIT (profiler tests)   isolated   Tests run: 6, Failures: 0, Errors: 0
OrphanedTimeSeriesCleanupIT        isolated   Tests run: 2, Failures: 0, Errors: 0
TableResourceIT (profiler tests)   parallel   Tests run: 6, Failures: 0, Errors: 0
OrphanedTimeSeriesCleanupIT        parallel   Tests run: 2, Failures: 0, Errors: 0
mvn spotless:check -pl openmetadata-service,openmetadata-integration-tests   BUILD SUCCESS
```

**RED check on the new watermark test.** Neutralising the predicate (`AND timestamp <= :deletedAt` → an always-true bound) and re-running `delete_hardDeletePurgeSpareProfilesRecordedAfterTheDelete` fails for exactly the right reason — the purge takes the successor's row too:

```
Purge must remove the row recorded before the delete ==> expected: <1> but was: <0>
```

Not applicable — no ingestion changes.

#### Playwright (UI) tests

#### Manual testing performed ✅

Re-run end to end against the asynchronous design on a local Docker stack (MySQL 8 + Elasticsearch),
server built from this branch. The instance was returned to its exact starting state afterwards
(195 tables, 3 profiler rows, no leftover fixtures).

**A — hard delete purges profiler data.** Create service/database/schema/table with columns `id` and
`name`, then `PUT /v1/tables/{id}/tableProfile` with a table, column and system profile.

1. `SELECT extension, COUNT(*) …` — one row per extension; `GET /v1/tables/{fqn}/tableProfile/latest?includeColumnProfile=true`
   returns `profile.rowCount = 42` and `columns[0].profile.uniqueCount = 42`.
2. `DELETE /v1/tables/{id}?hardDelete=true` → all three counts drain to 0.
3. Re-create the table at the identical FQN → `profile: null`, every `columns[*].profile` null, and
   `GET /v1/tables/{fqn}/systemProfile?startTs=0&endTs={now}` returns `"data": []`.
4. Soft-delete regression on a second table → all three rows retained and `rowCount = 42` still served.
5. Cascade regression on a third table → `DELETE /v1/databaseSchemas/{id}?hardDelete=true&recursive=true`
   drains its rows (the bulk path reaches `postDelete` per entity via `bulkHardDeleteSubtree`).

> **The purge is asynchronous, so poll rather than checking once.** At 300k rows the counts were still
> fully present the instant the DELETE returned. This is the one behavioural change from the previous
> revision of this PR.

**Scale.** A single table seeded with **1,200,000** column-profile rows — deliberately above the
1,001,000 ceiling the removed cap imposed:

```
DELETE /v1/tables/{id}?hardDelete=true   →  200, request returned in 0.24s
[om-async-6] TableRepository - Purged 1200000 profiler row(s) for hard-deleted table …
```

One statement, no truncation, on the bounded `AsyncService` executor. Under the capped design this
would have stopped at 1,001,000 and stranded ~199,000 rows — unreclaimable by the sweep if the FQN
were reused. Timing the raw statement separately: **300,000 rows in 7.27s**, so ~29s at 1.2M, all off
the request path. That delete is one transaction, so it holds those row locks for its duration; a
concurrent count over the same rows blocks until it completes.

**B — the weekly sweep no longer eats live column profiles.** Seeded 297,509 profiler rows across the
instance's 198 live tables (including a nested-column row, 6 hash segments) plus 500 planted orphans,
then triggered `POST /v1/apps/trigger/DataRetentionApplication` with `-H 'Content-Type: application/json' -d '{}'`:

```
before: 297,509 rows / 500 orphans     after: 297,009 rows / 0 orphans   (2.2s)
live-table rows preserved: 297,009
```

Exactly the 500 orphans removed, every live row kept. Both predicates timed on that same real data,
returning the **same 500 orphans**:

| Predicate | Time |
|---|---|
| This PR (parent-hash equality) | **344 ms** |
| LIKE-correlated | **6,748 ms** |

19.6× at only 198 tables; the gap grows with table count.

> A bodyless `POST /v1/apps/trigger/…` returns **500**, not 415. The `Content-Type` header and `{}`
> body are required either way.

**C — the index migration.** The migration ships in **2.0.0** (moved there by `fcece67`); any runbook
or check referencing 2.1.1 is stale, including the `server_migration_sql_logs` cleanup predicate.

On MySQL the premises behind the statement-free script were confirmed on a migrated instance: only
the 1.1.5 unique index and `combined_id_ts` are present (nothing added), `entityFQNHash` is
`varchar(768) ascii ascii_bin`, `server_migration_sql_logs` holds 93 statements for 2.0.0 and zero
for the index, and `EXPLAIN` on the purge predicate gives `type=range`,
`key=profiler_data_time_series_unique_hash_extension_ts`, `key_len=1796`, `Using where; Using index`
— index-driven, not a scan.

**Not verified by hand:** the PostgreSQL branch (`idx_profiler_data_time_series_fqnhash_pattern` and
its `indisvalid` check) — the manual stack and the IT runs were both MySQL. It is covered by
`OrphanedTimeSeriesCleanupIT.profilerPrefixIndexIsAvailable` on a Postgres CI lane.

#
### Residual risk:

Stated plainly rather than hedged, so reviewers can weigh them:

- **A crash between the delete committing and the purge completing loses that purge.** `AsyncService` is an in-process executor, so a server restart in that window drops the task. If the table is not re-created, the weekly orphan sweep reclaims the rows normally. If it **is** re-created at the same FQN before the sweep runs, the leftovers regain a live parent and the sweep can never reclaim them — only age-based retention (`profileDataRetentionPeriod`, default 1440 days) will. The window is the seconds between commit and drain, versus the permanent leak this PR replaces. A durable queue (`background_jobs`) would close it; that was judged not worth the machinery for a window this narrow, given the watermark already makes a late purge harmless.
- **The watermark uses the profile's logical timestamp, not the row's write time.** So it mis-classifies only if a successor table records a profile *backdated* to before its predecessor's deletion — a `PUT .../tableProfile` backfill with an explicit historical timestamp, or clock skew — **and** the purge is still pending. Closing this fully would need a write-time column on a very large table, which is not worth the migration.
- **A retention behaviour change for dropped columns.** The sweep is table-granular: it can tell whether a column row's parent *table* exists, but not whether that specific *column* still does. `table.columnProfile` rows for columns dropped from a still-live table are therefore retained until age-based retention removes them. Before this PR they were deleted weekly — but only as collateral damage of defect (B), which deleted every live column's profiles along with them. The trade-off is deliberate.
- **The interrupted-`CONCURRENTLY` runbook is documented, not automated.** Same gap as 1.11.0 and 1.13.0.
- **Historical rows stranded by pre-upgrade hard deletes are not back-filled.** The corrected weekly sweep reclaims them, which can take up to a week — except where a table was already re-created at that FQN before upgrading, where the same permanence described above applies.

#

#
### Checklist:
- [x] For JSON Schema changes: N/A — no schema change. A database migration **is** included (2.0.0) and is described above.
- [x] For UI changes: N/A — no UI changes.
- [x] I have added a test that covers the exact scenario we are fixing, with the issue number referenced in this PR and in the migration file header.

The PR asynchronously purges profiler history after table hard deletion and corrects orphan cleanup so live column profiles resolve to their parent table.
- Adds timestamp-bounded cleanup for table, system, and nested column profiles.
- Changes orphan detection to compare profiler rows with the parent table hash.
- Adds PostgreSQL prefix-index migration support and integration coverage for both database engines.
</details>

The PR does not yet appear safe to merge because a recreated table can still lose backdated successor profiles to the predecessor's delayed purge.

The purge identifies ownership using a reused FQN and the profile's caller-controlled logical timestamp, so a valid historical profile written by the successor before cleanup runs remains indistinguishable from predecessor data and is deleted.

**Files Needing Attention:** openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java, openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java | Schedules profiler cleanup after hard-delete processing, but the detached purge still relies on a caller-controlled logical timestamp to distinguish a recreated table. |
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java | Adds profiler purge statements and parent-table hash orphan detection; the outstanding timestamp ownership limitation remains. |
| bootstrap/sql/migrations/native/2.0.0/postgres/schemaChanges.sql | Adds the PostgreSQL pattern-operator index needed by column-profile prefix deletion. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java | Covers hard/soft deletion, recursive deletion, large histories, dropped columns, and the post-watermark successor case. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/OrphanedTimeSeriesCleanupIT.java | Verifies preservation of live table, system, column, and nested-column profiles while deleting genuine orphans. |

</details>

```mermaid
sequenceDiagram
  participant Client
  participant TableRepo as TableRepository
  participant Async as AsyncService
  participant DB
  Client->>TableRepo: Hard-delete table
  TableRepo->>DB: Delete table entity
  TableRepo->>Async: Queue profiler purge(FQN, deletedAt)
  TableRepo-->>Client: Delete response
  Async->>DB: "Delete table/system rows <= deletedAt"
  Async->>DB: "Delete column-prefix rows <= deletedAt"
```
</details>

<sub>Reviews (7): Last reviewed commit: ["refactor(profiler): collapse the orphan ..."](https://github.com/open-metadata/openmetadata/commit/1d5c4ea3047998575363a6b0731adc0e0d6614fe) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53403171)</sub>

**Context used:**

- Knowledge Base — [Bootstrap SQL and the Database Migration Runner](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/bootstrap-migrations.md)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -10184,6 +10185,84 @@ void deleteReportDataTypeAtDate(
...
+
+    private int drainBatches(IntSupplier deleteBatch) {
+      int totalDeleted = 0;
+      int deleted;
+      do {
+        deleted = deleteBatch.getAsInt();
+        totalDeleted += deleted;
+      } while (deleted == DELETE_BATCH_SIZE);
+      return totalDeleted;
+    }
```
manerow: `DELETE_BATCH_SIZE = 1000` bounds one statement, not how many statements run. The loop continues for as long as every statement returns a full batch, with no iteration cap. It terminates, but nothing bounds total rows or wall clock. Using the scale example from the description itself (~1200 columns profiled hourly for 60 days, ~1.73M rows), that is ~1,730 sequential `DELETE`s issued from the request thread.

On the direct hard-delete path they all run in one transaction: `deleteInternal` calls `cleanup(updated)` (`EntityRepository.java:4687`), which wraps its body in `Entity.getJdbi().inTransaction(...)` and calls `entitySpecificCleanup` inside it (`EntityRepository.java:4846-4851`). Nothing commits between batches, so the usual reasons to batch do not apply: row locks taken in batch 1 are still held during batch 1,730, none of the deleted rows can be vacuumed until the whole delete commits, and any failure rolls back every batch. The statements get smaller, the peak cost does not.

The cascade path differs but is not better. `bulkHardDeleteSubtree` is documented as non-transactional, and `bulkEntitySpecificCleanup` loops the per-entity hook (`EntityRepository.java:7065`), so a recursive schema or service hard delete runs one uncapped drain per table, sequentially, holding a single connection for the walk.

This is also the risk `deleteChildren` cites a few lines up when it keeps time-series children off this path: "millions-of-rows lock risk, orphans cleaned offline by DataRetention" (`EntityRepository.java:4820-4822`).

The description still specifies the missing piece: `DESCENDANT_DELETE_MAX_BATCHES` (1000), `DescendantPurge(rowsDeleted, capReached)`, and a cap `WARN` via `logColumnProfilePurge`. Those were added in `bbac400054` and removed in `2b554818b`, so the 1,001,000-row ceiling described under Residual risk is not the current behaviour. Restoring the cap and the WARN is the minimum. Committing the drain outside the entity transaction would be better, but note `ef1c47f8f` moved this off the async `postDelete` path specifically to avoid a recreation race, so it would need to commit before the FQN can be reused rather than simply going async again.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -10252,6 +10331,10 @@ default void deleteEntityProfileData(ListFilter filter, Long timestamp) {
                 + "WHERE NOT EXISTS ("
                 + "  SELECT 1 FROM table_entity te "
                 + "  WHERE te.fqnHash = profiler_data_time_series.entityFQNHash"
+                + "    OR (profiler_data_time_series.extension = '"
+                + TABLE_COLUMN_PROFILE_EXTENSION
+                + "' AND profiler_data_time_series.entityFQNHash "
+                + "LIKE CONCAT(te.fqnHash, '.%'))"
                 + ") "
                 + "LIMIT :limit",
```
manerow, line 10335: This applies to the Postgres branch below as well.

`te.fqnHash = entityFQNHash OR entityFQNHash LIKE CONCAT(te.fqnHash, '.%')` has two different join keys, and the LIKE pattern is computed from the inner row, so no equijoin key remains to hash on. Both planners fall back to a nested loop over `table_entity` for every profiler row. A column-profile row can never satisfy the equality branch, so it is resolved only by scanning `table_entity` until its parent turns up, and all of it when there is no parent. `LIMIT` stops bounding the work as well: on a healthy catalog nearly every row is live, so the executor must evaluate the anti-join over the whole table to discover the handful of orphans. The 2.0.0 index does not offset this. It is on `profiler_data_time_series.entityFQNHash`, which is the side being scanned in full, and the LIKE branch cannot use `table_entity`'s index either since its pattern is not constant.

`DataRetention.cleanOrphanedTimeSeriesRows` runs this with `BATCH_SIZE = 10_000` and repeats until a call returns 0 (`MAX_ORPHAN_CLEANUP_ITERATIONS = 1000`), so a healthy instance pays at least two of these full passes per weekly run.

Measured on a fixture reproducing the real column types, collation and indexes, with only 5,000 tables and 105,300 profiler rows, issuing the exact statement the DAO issues:

| | Postgres 15 | MySQL 8 |
|---|---|---|
| this PR | 27,968 ms | 46.7 s |
| `822fe4c42` | 63.9 ms | 0.14 s |

Both return the same 250 orphans. The Postgres plan for this PR is `Nested Loop Anti Join` with `Rows Removed by Join Filter: 263,698,725` against a materialised 5,000-row inner relation; `822fe4c42` plans as `Hash Anti Join`. Cost is O(profiler_rows x tables), so at 50k tables and 10M rows one statement is on the order of 10^11 comparisons and the sweep stops completing rather than merely slowing. That matters directly here: this PR ships no backfill and relies on the sweep to reclaim historical orphans.

`822fe4c42` had the shape that works, resolving a column row to its parent by hash-prefix equality (`SUBSTRING_INDEX(entityFQNHash, '.', 4)` on MySQL, chained `split_part` on Postgres), and `2b554818b` replaced it. It keeps a single equality so the hash anti-join survives, and it is behaviourally identical including the cases that motivated the change: on the same fixture extended with nested-column rows under live tables (6 hash segments) and column rows under dead tables (5 segments), both predicates return the same 250 rows with an empty symmetric difference in both directions.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -10184,6 +10184,55 @@ void deleteReportDataTypeAtDate(
...
+        @Bind("deletedAt") long deletedAt);
+
+    @SqlUpdate(
+        "DELETE FROM <table> WHERE entityFQNHash LIKE :hashPrefix AND extension = :extension "
+            + "AND timestamp <= :deletedAt")
+    int deleteByFqnHashPrefix(
+        @Define("table") String table,
+        @Bind("hashPrefix") String hashPrefix,
+        @Bind("extension") String extension,
+        @Bind("deletedAt") long deletedAt);
```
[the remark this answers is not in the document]
TeddyCr: intentional

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
TeddyCr: Agreed on all counts, and fixed in b8f1a99 — though not by restoring the cap.

You're right that `DELETE_BATCH_SIZE` bounded a statement and nothing bounded the loop, and right that on the direct path it all runs in one transaction. That last point is what decided the fix: if the locks are held to the end either way, the batching bought nothing — it was the same peak cost spread over ~1,700 round-trips instead of one statement. So the batching is gone, and so is the cap.

The cap isn't restored because it traded an availability problem for a silent-correctness one. At the 1.73M-row scale from the description it truncates, and truncated rows regain a live parent if the FQN is reused — which is exactly this issue's symptom, permanently, since the sweep correctly declines to reclaim them. The PR description's "Residual risk" said as much.

What makes the purge safe to run unbounded and off the request thread is a deletion watermark. Every statement now carries `AND timestamp <= :deletedAt`, so it cannot touch a successor table's profiles no matter when it lands — they're all recorded after the delete. That gives the property `ef1c47f8f` went synchronous to obtain, without holding the request:

- **Runs after the delete commits** → a rolled-back delete can't purge a live table's history.
- **Bounded to profiles predating the delete** → FQN reuse is a non-event; the purge is also idempotent on retry.

It now dispatches from `TableRepository.postDelete` via `AsyncService.executeDatabaseTask(PROFILER_CLEANUP, …)`, so it's bounded by the configured DB-task limit rather than unbounded — that facility arrived in #31583 after this branch forked, which is also why the earlier revision used the plain unbounded `AsyncService.execute`. `bulkHardDeleteSubtree` calls `postDelete` per entity, so the cascade path you flagged is covered; pinned by `delete_hardDeleteOfSchemaPurgesTableProfilerData`.

Measured on a local stack with a table carrying **1,200,000** column-profile rows — deliberately above the 1,001,000 ceiling the cap imposed:

```
DELETE /v1/tables/{id}?hardDelete=true  →  200, request returned in 0.24s
[om-async-6] TableRepository - Purged 1200000 profiler row(s)
```

One statement, no truncation. Timing the raw statement separately: 300,000 rows in 7.27s, so ~29s at 1.2M — all off the request path. It is one transaction, so it does hold those row locks for its duration; that's the deliberate cost of not batching, and it's now paid in the background rather than in the request.

`delete_hardDeletePurgeSpareProfilesRecordedAfterTheDelete` pins the watermark: one row seeded either side of it, only the earlier one purged. Neutralising the predicate fails it with `expected: <1> but was: <0>`.

The description is updated — `DESCENDANT_DELETE_MAX_BATCHES`, `DescendantPurge` and `logColumnProfilePurge` are gone from both the code and the text, and the 1,001,000-row ceiling is no longer listed under Residual risk. Thanks for catching that drift.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
TeddyCr: You're right, and this was a self-inflicted regression — `822fe4c42` earlier in this same PR had the shape that works, and `2b554818b` replaced it as a drive-by while reworking the batching, with no stated reason. Reverted in b8f1a99.

I reproduced your analysis independently before reverting rather than taking the numbers on faith. On a fixture with the real column types, collation and indexes (`ascii_bin` on MySQL, the 1.1.5 unique index, plus the `text_pattern_ops` index this PR ships), 5,000 tables / 105,250 rows / 250 planted orphans, issuing the exact statement the DAO issues:

| | Postgres 15 | MySQL 8.4 |
|---|---|---|
| this PR (`LIKE`-correlated) | 32,389 ms | 75,637 ms |
| `822fe4c42` (parent-hash equality) | **106 ms** | **247 ms** |

Same plan shapes you describe: `Nested Loop Anti Join` with `Rows Removed by Join Filter: 263,697,500` versus `Hash Anti Join`. My Postgres number came out worse than your 27,968 ms, so if anything the report understated it.

Behavioural equivalence holds — both return the same 250 orphans with an empty symmetric difference in both directions, on both engines. I checked the assumption underneath `SUBSTRING_INDEX(…, '.', 4)` too: `FullyQualifiedName.buildHash` hashes each `split()` part and joins with `.`, and each segment is MD5 hex containing no dots, so a table's `fqnHash` is always exactly 4 segments. `OrphanedTimeSeriesCleanupIT.profilerDataOrphans` already covers both directions that matter — a nested-column row under a live table (6 segments) and a column row under a dead one.

Also confirmed on a real instance, 297,509 profiler rows across 198 live tables plus 500 planted orphans:

| Predicate | Time |
|---|---|
| parent-hash equality | **344 ms** |
| `LIKE`-correlated | **6,748 ms** |

Both find the same 500. 19.6× at only 198 tables, and the gap grows with table count. Running the real `DataRetentionApplication` sweep removed exactly those 500 in 2.2s and preserved all 297,009 live-table rows.

Your point about this mattering more than it looks is the one I'd underline: the PR ships no backfill and leans on the sweep to reclaim historical orphans, so a sweep that stops completing means the migration path quietly never finishes. The rationale is now a comment above the predicate with the measurements attached, so it doesn't get flattened back into the `LIKE` form.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java
```diff
@@ -10307,38 +10307,30 @@ default void deleteEntityProfileData(ListFilter filter, Long timestamp) {
...
-            + "|| split_part(pdts.entityFQNHash, '.', 2) || '.' "
-            + "|| split_part(pdts.entityFQNHash, '.', 3) || '.' "
-            + "|| split_part(pdts.entityFQNHash, '.', 4)";
+    // Resolving the parent from the outer row keeps a single equality, so both planners serve this
+    // as a hash anti-join. Correlating the other way (entityFQNHash LIKE te.fqnHash || '.%') builds
+    // the pattern from the inner row, leaving no equijoin key, and degrades to a nested loop over
+    // table_entity per profiler row: 32s vs 106ms on Postgres 15, 76s vs 247ms on MySQL 8 over 5k
+    // tables / 105k rows. 1.9.9/postgres/postDataMigrationSQLScript.sql resolves the same parent
+    // for the same reason, noting it is "much faster than LIKE".
+    String PARENT_TABLE_HASH = "LEFT(pdts.entityFQNHash, 131)";
```
TeddyCr, line 10322: This relies on the hash function being constant (i.e. each segment of the fqn is an MD5 hash separated by dots.)

manerow (approved):

PR #31565, opened 2026-08-14
Fixes 31564: pass the extended config when formatting a single-value rule

### LINKED ISSUE #31564: Persona AI Context: a completed filter condition is rejected as "unfinished" and the rule cannot be saved
### Affects

`main` (`2.0.0-SNAPSHOT`). The validation was introduced in #30716; no released branch contains it yet.

### What happened

In **Settings → Personas → \<persona\> → AI Context → Add Rule**, a filter condition that is completely
filled in is rejected on save with:

> Finish or remove the unfinished condition before saving. An unfinished condition is ignored, which
> would widen this rule to match everything.

There is no unfinished condition. The match preview in the same drawer resolves the filter and reports
the correct entity count, and "View in Explore" returns the same assets — so the filter is valid and
the emitted query is fine. The rule simply cannot be saved from the UI.

### Steps to reproduce

1. Settings → Personas → pick any persona → **AI Context** tab → **Add Rule**.
2. Name the rule, leave Entity Type as **Table**.
3. Add one condition: **Service** / **Is** / *(pick any service)*.
4. Observe the preview banner correctly reports the matched entity count.
5. Click **Save Rule**.

**Expected:** the rule is saved.
**Actual:** the "unfinished condition" error appears and the drawer stays open.

### Scope

Any condition using a single-value operator (`Is`, `Is not`, `Contains`, `Is null`, `Between`, …) on a
normal field triggers it. Two cases happen to escape:

- multiselect operators (`Any in`, `Not any in`) — they take a different branch, so a filter built only
  out of those saves fine;
- custom-property (`extension.*`) fields — they return before the affected code.

### Root cause

`elasticSearchFormat` (`QueryBuilderElasticsearchFormatUtils.js`) extends the incoming config once at
the top and hands `extendedConfig` to every branch — the group branch, and the multiselect half of the
rule branch — except the single-value half, which passes the raw `config` through to `buildEsRule`.

`buildEsRule` resolves the field's widget through that config (`getWidgetForFieldOp` →
`config.widgets[widget]`). A raw config resolves no widget, so it returns `undefined`.

That difference is invisible while the whole tree is formatted, because a rule is then always reached
*through* its group and so always receives an extended config. It becomes visible as soon as
`elasticSearchFormat` is called on a rule node directly — which is exactly what `hasUnfinishedRule`
does to decide whether a row was left unfinished. It sees `undefined`, concludes the row produces no
constraint, and blocks the save.

Minimal demonstration — same tree, same config, opposite answers:

```js
elasticSearchFormat(tree, config)
// => {"bool":{"must":[{"term":{"service.displayName.keyword":"banking-bigquery"}}]}}

elasticSearchFormat(firstRuleOf(tree), config)
// => undefined

hasUnfinishedRule(tree, config)
// => true
```

### Workaround

Create the rule through the API: `POST /api/v1/personas/{id}/aiContext/rules`.

pmbrull:
Fixes #31564

## Problem

In **Settings → Personas → \<persona\> → AI Context → Add Rule**, a filter condition that is fully
filled in is rejected on save:

> Finish or remove the unfinished condition before saving. An unfinished condition is ignored, which
> would widen this rule to match everything.

Nothing is unfinished. The match preview in the same drawer resolves the filter and reports the right
count, and "View in Explore" returns the same assets. The rule just cannot be saved from the UI.

It hits any condition using a single-value operator (`Is`, `Is not`, `Contains`, `Is null`,
`Between`, …) on a normal field. Two cases escape: multiselect operators (`Any in`, `Not any in`),
and custom-property (`extension.*`) fields.

## Root cause

`elasticSearchFormat` extends the incoming config once at the top and hands `extendedConfig` to every
branch — the group branch, and the multiselect half of the rule branch — except the single-value half,
which passed the raw `config` straight to `buildEsRule`.

`buildEsRule` resolves the field's widget through that config (`getWidgetForFieldOp` →
`config.widgets[widget]`). A raw config resolves no widget, so it returns `undefined`.

The difference is invisible while a whole tree is formatted, because a rule is then always reached
*through* its group and so always receives an extended config. It surfaces the moment
`elasticSearchFormat` is called on a rule node directly — which is exactly what `hasUnfinishedRule`
(added in #30716) does to decide whether a row was left unfinished. It saw `undefined`, concluded the
row constrained nothing, and blocked the save.

Same tree, same config, opposite answers before this change:

```js
elasticSearchFormat(tree, config)
// {"bool":{"must":[{"term":{"service.displayName.keyword":"banking-bigquery"}}]}}

elasticSearchFormat(firstRuleOf(tree), config)
// undefined

hasUnfinishedRule(tree, config)
// true
```

## Fix

Pass `extendedConfig` to `buildEsRule` in the single-value branch, matching what the multiselect
branch beside it and the group branch already do. Formatting a whole tree is unaffected — the config
reaching `buildEsRule` there was already extended.

## Tests

`QueryBuilderElasticsearchFormatUtils.test.js`, four new cases. The existing `hasUnfinishedRule`
coverage all used `extension.*` fields, which return through `buildExtensionQuery` before the widget
is ever resolved, so it could not see this; the new cases use a plain `select` field and a real
`QbUtils.loadTree`/`checkTree` tree.

- a rule builds the same clause whether reached through its group or on its own
- an entered single-value condition is accepted (fails without the fix)
- an entered multiselect condition is accepted (the path that already worked, pinned)
- a single-value condition with no value entered is still reported — the #30716 behaviour this must
  not regress

```
Tests: 22 passed, 22 total
```

Also green: `QueryBuilder*`, `AdvancedSearch*`, `ContextRule*`, `PersonaAIContext*`, `Explore*`,
`CuratedAssets*` — 487 tests. ESLint clean, no new warnings.

---

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/utils/QueryBuilderElasticsearchFormatUtils.js
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/QueryBuilderElasticsearchFormatUtils.js
@@ -1058,7 +1058,17 @@ export function elasticSearchFormat(tree, config, syntax = ES_6_SYNTAX) {
           },
         };
       } else {
-        return buildEsRule(field, value, operator, config, valueSrc, syntax);
+        // extendedConfig, as in every other branch: buildEsRule resolves the field's widget
+        // through the config it is given, and a raw one resolves none — so a fully entered
+        // condition builds no clause at all when this runs on a rule node directly.
+        return buildEsRule(
+          field,
+          value,
+          operator,
+          extendedConfig,
+          valueSrc,
+          syntax
+        );
       }
     }
 
--- a/openmetadata-ui/src/main/resources/ui/src/utils/QueryBuilderElasticsearchFormatUtils.test.js
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/QueryBuilderElasticsearchFormatUtils.test.js
@@ -11,7 +11,10 @@
  *  limitations under the License.
  */
 
-import { AntdConfig } from '@react-awesome-query-builder/antd';
+import {
+  AntdConfig,
+  Utils as QbUtils,
+} from '@react-awesome-query-builder/antd';
 import {
   elasticSearchFormat,
   hasUnfinishedRule,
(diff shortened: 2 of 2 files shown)


harshach (approved):

pmbrull:
picked by hand to 2.0

PR #31566, opened 2026-08-15
fix(e2e): make widget skeleton waits target the real data-testid

aniketkatkar97:
## What

`playwright/utils/widgetFilters.ts` waited for widget data to settle with
`widget.locator('entity-list-skeleton')` — a **tag-name** selector that matches nothing (the
skeleton renders `data-testid="entity-list-skeleton"`). All 15 such waits resolved instantly,
so the sort/filter flows clicked dropdown → item → dropdown → item with zero settle time between
widget refetches.

## Why

Root-caused from Collate 1.13 nightly AUT run 31801028071: `CustomizeWidgets.spec.ts` › My Data
Widget hung 180 s waiting for the Z-to-A `search/query` request that never fired — the test raced
the widget through three sort changes with no real wait in between. Three call sites in the same
file already use `getByTestId('entity-list-skeleton')` correctly; this aligns the rest.

## Change

Mechanical: `.locator('entity-list-skeleton')` → `.getByTestId('entity-list-skeleton')`
(15 occurrences, one file). Prettier-clean.

1.13 counterpart: #31567

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/utils/widgetFilters.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/utils/widgetFilters.ts
@@ -111,7 +111,7 @@ export const verifyDataFilters = async (
   );
   await page.getByRole('menuitem', { name: 'A to Z' }).click();
   await aToZFilter;
-  await widget.locator('entity-list-skeleton').waitFor({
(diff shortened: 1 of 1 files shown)


aniketkatkar97:
Consolidated into #31551 (commit bb57f8d2d9) per review — closing. The 1.13 counterpart #31567 stays open.

PR #31572, opened 2026-08-15
Fixes 31571: show the real Queries tab count, with a skeleton while it loads

### LINKED ISSUE #31571: Queries tab count on the Table details page shows 0 until the tab is opened
### Describe the bug

On the Table details page the **Queries** tab badge renders `0` on first paint, even for tables that have queries. Clicking the tab flips the badge to the real count.

### Root cause

`queryCount` is initialised to `0` and the count fetch is deferred until the Queries tab becomes active:

- `openmetadata-ui/src/main/resources/ui/src/pages/TableDetailsPageV1/TableDetailsPageV1.tsx` — `useState(0)`, then `useDeferredTabData(EntityTabs.TABLE_QUERIES, activeTab, fetchQueryCount, [...])`, plus a reset back to `0` on FQN change.
- `openmetadata-ui/src/main/resources/ui/src/components/common/TabsLabel/TabsLabel.component.tsx` renders the badge whenever `!isNil(count)`, so `0` renders as a real count rather than "unknown".

So the badge shows a value that is wrong until the user opens the tab.

Secondary: `useDeferredTabData` compares `activeTab === tabKey` strictly, so column-scoped routes whose `tab` param is not literally `tableQueries` never trigger the fetch at all, contradicting the comment at the call site.

### To Reproduce

1. Open a table that has at least one query, landing on the **Columns** tab.
2. The Queries tab shows `Queries 0`.
3. Click the Queries tab — the badge becomes the real count (e.g. `Queries 1`).

### Expected behavior

The Queries tab badge shows the true count on first paint, without requiring the tab to be opened.

### Version

- OS: macOS
- OpenMetadata version: main

harsh-vador:
### Describe your changes:

Fixes #31571

The **Queries** tab badge on the Table details page rendered `0` on first paint even when the table had queries; it only flipped to the real count once the tab was clicked.

`queryCount` was initialised to `0` while the fetch was deferred to tab activation via `useDeferredTabData`, and `TabsLabel` renders the badge whenever `!isNil(count)` — so `0` was displayed as a genuine count rather than "not yet known". The deferral also silently never fired on column-scoped routes, whose `tab` param is never literally `tableQueries`.

Changes:
- **Fetch the count with the entity, not on tab activation.** `TableDetailsPageV1` now reads it through `useQuery` (`tableQueryCountKey` / `tableQueryCountFn`, added alongside the existing table query plumbing in `rest/queries/tableQuery.ts`). `useQuery` rather than a fetch effect plus a `useState` flag is load-bearing: `isFetching` is already true on the render that starts the request, whereas an effect-driven boolean flips one render too late and flashes a `0` first.
- **Show a placeholder while it is in flight.** The new `isQueryCountLoading` prop is forwarded to the Queries `TabsLabel` as `isLoading`.
- **Make that placeholder a skeleton rather than a spinner.** `TabsLabel`'s `isLoading` branch rendered `<Loader size="small" />` — a spinner in place of a small static pill draws more attention than the number it stands in for. It is now a `Skeleton.Button` sized to the badge so the label does not reflow when the count lands. The Queries tab is the first production consumer of `isLoading`, so no other tab changes appearance.
- **Deleted `hooks/useDeferredTabData.ts` and its test** — `TableDetailsPageV1` was its only consumer, so it is now dead code.

https://github.com/user-attachments/assets/27bc3aae-8028-4c3b-8928-44f89a1733e5

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered
- A user landing on the Columns tab of a table that has queries sees the true count in the Queries tab badge, without opening the tab.
- While the count is in flight the badge shows a skeleton, never a placeholder `0`.
- Navigating to a different table updates the badge rather than showing the previous table's count.
- A failing count request leaves the badge at `0` instead of breaking the page.

#### Unit tests
- `openmetadata-ui/src/main/resources/ui/src/pages/TableDetailsPageV1/TableDetailsPageV1.test.tsx` — new `Queries tab count` block: the count is fetched on mount while on the schema tab (`getQueriesList` called as `{ limit: 0, entityId }`, badge receives the resolved count); **every** render before the request resolves keeps the badge in its loading state, which is what rules out the `0` flash rather than just asserting the end state; and the fallback to `0` when the request rejects.
- `openmetadata-ui/src/main/resources/ui/src/utils/TableUtils.test.tsx` — fixture updated for the new required prop.
- `hooks/useDeferredTabData.test.ts` removed along with the hook.
- `yarn test src/pages/TableDetailsPageV1 src/utils/TableUtils.test.tsx src/components/common/TabsLabel` → **167 passed, 6 suites**.
- `npx tsc --noEmit` → no new errors against the pre-change baseline (550 both before and after, same per-file distribution). `yarn token-audit` → the only error is pre-existing in `styles/components/om-rdg.less`, untouched here.

#### Playwright (UI) tests
- [x] Not applicable — covered by the unit tests above; no new UI surface was added.

#### Manual testing performed
1. Opened a BigQuery table with one query, landing on the **Columns** tab.
2. Confirmed the Queries badge shows a skeleton and then the real count on first paint, with a single `GET /api/v1/queries?limit=0&entityId=<id>` in the Network tab and no duplicate request on clicking the tab.
3. Navigated to a second table and confirmed the badge updates rather than retaining the first table's count.

#

Before: the Queries tab reads `Queries 0` while on the Columns tab, and only reads `Queries 1` for the same table after the tab is opened. After: a skeleton in the badge, then the correct count, without opening the tab.

#
### Checklist:
- [x] For UI changes: described above.


harsh-vador:
Follow-up pushed: eager fetching alone still let the badge render `0` for the first frames before the count landed. The count now goes through `useQuery` and the badge shows `TabsLabel`'s existing skeleton while it is in flight.

- `rest/queries/tableQuery.ts` — added `tableQueryCountKey` / `tableQueryCountFn` alongside the existing table query plumbing, so the count gets its own cache slot.
- `TableDetailsPageV1.tsx` — replaced the `queryCount` state and `fetchQueryCount` effect with `useQuery`, and pass `isFetching` down as `isQueryCountLoading`. Using `useQuery` matters here: the flag is already `true` on the render that starts the request, whereas an effect-driven boolean flips one render too late and flashes a `0` first.
- `TableClassBase.ts` / `TableTabsUtils.tsx` — new `isQueryCountLoading` prop, forwarded to the Queries `TabsLabel` as `isLoading`.

New test: `should render the skeleton instead of a count while the request is in flight` asserts that *every* render before the request resolves kept the badge in its loading state, so the placeholder `0` can never flash. `yarn test src/pages/TableDetailsPageV1 src/utils/TableUtils.test.tsx src/components/common/TabsLabel` → 167 passed. `npx tsc --noEmit` shows no new errors against the pre-change baseline.

chirag-madlani (approved):

PR #31580, opened 2026-08-15
perf(aicontext): strip render-only markup from context documents

pmbrull:
## What

Descriptions are stored as block-editor HTML and sanitized for XSS on write by `DescriptionSanitizer`, which deliberately keeps what a browser needs: `class` attributes, sizing hints, and inline `data:` images. A model can see none of that, so in an AI context document those bytes buy nothing — and they are charged twice:

1. **Against the persona-context character budget.** Knowledge rules are mandatory: `PersonaContextMarkdown.render` subtracts `mandatoryKnowledgeChars` from the budget *before* any asset rule gets a share.
2. **On every LLM call** that carries the document in its prompt.

One customer had a single inline base64 architecture diagram worth **320,000 characters** in a knowledge article — **63% of a 505,000-character persona context**. It consumed the whole 400,000-character budget, so the rule that should have rendered 151 table schemas rendered none (`151 matched, 0 rendered, truncated: true`), and it added roughly 80,000 tokens to every prompt built from that document.

## How

New `aicontext/PromptText.forPrompt(String)` applies a second, stricter OWASP policy at every description site:

- `PersonaContextMarkdown` — article summary and full body, metric and glossary definitions, shared-knowledge content.
- `AIContextMarkdown` — asset description, frontmatter summary, column cells, and knowledge content. In the last case it runs **before** `truncate()`, so an excerpt is never cut out of the middle of a base64 blob.

The policy allows neither the `data` URL protocol nor the `img` element, so an inline binary cannot survive by any route, and whitelists no presentational attribute, which drops `class`, `style`, `id` and sizing without naming them one by one. What survives is the structure a model reads: headings, lists, tables with `colspan`/`rowspan`, code, http(s) links, and `data-fqn` / `data-entitytype` on mention anchors — those name something the agent can look up.

A tag-shaped-run guard returns plain text untouched. Without it the sanitizer HTML-escapes a bare `<`, turning a note reading `WHERE a < b` into `a &lt; b`.

`DescriptionSanitizer.sanitizeWith(policy, text)` is extracted so the `<#E::...>` entity-link preservation is reused rather than copied.

**The storage policy is unchanged — images still render in the UI.** This only affects what agents are sent.

## Impact

Measured on the persona context that motivated this:

| | before | after |
|---|---|---|
| document | 504,885 chars | **159,454 chars** (−68.4%) |
| `tokens_est` | 126,221 | **39,863** |

Larger than the image alone predicts, because attribute stripping compounds — `class="om-list-disc"`-style attributes were another 23,752 characters.

## Tests

`PromptTextTest` (9 new) covers base64 removal, URL and `blob:` images, `data:` URIs on non-image attributes, presentational-attribute stripping, structure preservation, entity mentions, entity-link tokens, the plain-text guard, and null/blank.

`PersonaContextMarkdownTest.anInlineImageInAnArticleDoesNotStarveTheAssetRulesOfBudget` is the regression guard for the budget defect: it asserts the asset rule now renders in full with an empty manifest. Confirmed to fail with the fix reverted.

`mvn -pl openmetadata-service spotless:check test` — 67 tests green, including the 26 existing `DescriptionSanitizerTest` and 21 `AIContextMarkdownTest` unchanged.

## Notes

`PersonaContextCache` is keyed on persona id + definition hash with a TTL, so a deploy does not invalidate cached documents; they refresh within one TTL.


pmbrull:
picked to 2.0

PR #31590, opened 2026-08-16
fix(governance): persist workflow task stage after Flowable commit to stop spurious 409

yan-3005:
## Describe your changes

A `/resolve` for the next user-task transition can arrive while the workflow is still running an **async `automatedTask` between two user tasks**. The canonical case is Collate's DataAccessRequest workflow:

```
TaskReview --approve--> PolicyAgent (async automatedTask) --> ApprovedAccess --markAsGranted--> GrantedAccess
```

`CreateTask.notify()` runs as a Flowable `TaskListener` inside the command that creates the next runtime user task, but on the **update path** it persisted the entity's `workflowStageId` / `availableTransitions` via `taskRepository.update(...)` — which commits on JDBI's own connection **mid-command**, milliseconds before Flowable commits the `ACT_RU_TASK` row and the `customTaskId` process variable.

A client polling `availableTransitions` could therefore see the next transition (e.g. `markAsGranted`) advertised and resolve into a **not-yet-committed** runtime task: the resolve found no task, completed with **no variables**, and the outgoing gateway condition `${<node>_result == '<transition>'}` evaluated an **unbound variable** → `FlowableException` → surfaced to the caller as a spurious **409 CONFLICT** under concurrency.

### Fix
- Defer the **update-path** persist to a Flowable **post-commit transaction listener** (`TransactionState.COMMITTED`) so the entity's stage / `availableTransitions` are only advanced once the runtime task they advertise is durable.
- This is **non-blocking** — the write is *reordered*, never waited on. No `Thread.sleep`, no polling, no thread blocked. `COMMITTED` listeners run synchronously inside the command's commit **before** a synchronous `taskService.complete()` returns, so synchronous stage advances (e.g. incident userTask→userTask) still reflect the new stage in their response. On rollback the listener never fires and the entity correctly keeps the prior stage.
- The **create path stays synchronous** so a persist failure still aborts the workflow via `BpmnError`.
- `TaskWorkflowHandler.resolveWorkflowTask` guards against a `null` namespace map (treats it as a failed resolve) so a genuinely stale resolve never completes a task with no variables.

## Type of change

## How was this tested

- `CreateTaskPostCommitPersistTest` — source-level regression guard that the update-path persist is deferred to a `TransactionState.COMMITTED` listener (not persisted inline), that it never blocks (`Thread.sleep`-free), and that the null-namespace guard is present.
- Behavioural acceptance via the Collate DAR integration tests (`DarStatusGroupIT`, `DataAccessRequestValidationIT`, `DataAccessRequestIT`), which run the real PolicyAgent async node: the two CI-failing tests (`darListEndpointMultiSelectStatus`, `testDarClose_onGranted_returns400`) pass, and the full DAR suite is green across repeated concurrent runs.

## Checklist

- [x] My PR title follows the repo conventions.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective.

The PR defers update-path task persistence until Flowable commits and prevents task completion with a null namespaced-variable map.
- Registers task-entity updates on Flowable's committed transaction state.
- Re-reads the latest task row before applying the deferred stage update.
- Treats a missing runtime-task namespace as a failed workflow resolution.
</details>

The PR does not yet appear safe to merge because a terminal resolve can still update the task entity directly while its Flowable process is asynchronously advancing and has not materialized the next runtime task.

The null-map guard prevents completing a Flowable task without variables, but a null lookup also makes the active-task check false and leaves the existing terminal direct-resolution fallback reachable while the workflow may still be live.

**Files Needing Attention:** openmetadata-service/src/main/java/org/openmetadata/service/tasks/TaskWorkflowHandler.java
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/userTask/CreateTask.java | Defers existing-task stage persistence to a Flowable committed-transaction listener while retaining synchronous creation behavior. |
| openmetadata-service/src/main/java/org/openmetadata/service/tasks/TaskWorkflowHandler.java | Rejects null namespaced variables before invoking Flowable, but the previously reported terminal fallback remains reachable when no runtime task is visible. |
| openmetadata-service/src/test/java/org/openmetadata/service/governance/workflows/elements/nodes/userTask/CreateTaskPostCommitPersistTest.java | Adds focused coverage for committed-listener registration, execution timing, fallback execution, and exception containment. |

</details>

```mermaid
sequenceDiagram
  participant Client
  participant TaskHandler as TaskWorkflowHandler
  participant Flowable
  participant Listener as CreateTask listener
  participant TaskDB as Task repository
  Client->>TaskHandler: Resolve transition
  TaskHandler->>Flowable: Complete current user task
  Flowable->>Listener: Create next user task
  Listener->>Flowable: Register COMMITTED callback
  Flowable->>Flowable: Commit runtime task and variables
  Flowable->>Listener: Run COMMITTED callback
  Listener->>TaskDB: Persist workflow stage and transitions
  Flowable-->>TaskHandler: Completion returns
```
</details>

<sub>Reviews (4): Last reviewed commit: ["test(governance): make CreateTask post-c..."](https://github.com/open-metadata/openmetadata/commit/37d026c8732a821a83e007bc5888dc68c65866ee) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53757378)</sub>

**Context used:**

- Knowledge Base — [Governance Workflows and Applications in OpenMetadata Service](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/openmetadata-service-workflows-apps.md)

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/userTask/CreateTask.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/userTask/CreateTask.java
@@ -582,9 +582,19 @@ public class CreateTask implements TaskListener {
       // the post-commit listener, which runs synchronously on this thread inside the command's
       // commit (before a synchronous resolve returns), and is simply skipped if Flowable rolls
       // back.
-      final Task baseline = currentTask;
       final Task desired = updatedTask;
-      registerPostCommitPersist(() -> taskRepository.update(null, baseline, desired, updatedBy));
+      final UUID persistTaskId = updatedTask.getId();
+      registerPostCommitPersist(
+          () -> {
+            // Re-read the committed row as the update baseline: between building `desired` and this
+            // post-commit callback the stored task may have advanced, and updating against a stale
+            // snapshot would lose the optimistic-version check. On a hard persist failure the
+            // entity
+            // stays one stage behind the (already committed) Flowable runtime and self-heals on the
+            // next stage advance, which re-reads the current task.
+            Task latest = taskRepository.get(null, persistTaskId, taskRepository.getFields("*"));
+            taskRepository.update(null, latest, desired, updatedBy);
+          });
       return updatedTask;
     }
 
--- a/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/userTask/CreateTask.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/userTask/CreateTask.java
@@ -38,6 +38,9 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.flowable.common.engine.api.FlowableObjectNotFoundException;
 import org.flowable.common.engine.api.delegate.Expression;
+import org.flowable.common.engine.impl.cfg.TransactionContext;
+import org.flowable.common.engine.impl.cfg.TransactionState;
+import org.flowable.common.engine.impl.context.Context;
 import org.flowable.engine.RuntimeService;
 import org.flowable.engine.delegate.BpmnError;
 import org.flowable.engine.delegate.TaskListener;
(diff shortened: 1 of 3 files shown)


yan-3005:
Addressing the review findings:

**On the current commit (post-commit-listener approach):**

- **Post-commit persist failure / stale baseline** (gitar-bot, `CreateTask.java`) — fixed in `0671ad0a60`: the deferred persist now re-reads the committed task row as the update baseline inside the `COMMITTED` listener, so a concurrent advance can't cause an optimistic-version conflict. On a genuine hard persist failure the entity is one stage behind the already-committed Flowable runtime and self-heals on the next stage advance (which re-reads the current task) — this is the inherent, documented trade-off of the JDBI↔Flowable datasource split, and is strictly safer than the prior ordering (entity leading the runtime).

- **`CreateTaskPostCommitPersistTest` asserts source text, not behaviour** (gitar-bot) — intentional and consistent with the existing `WorkflowHandlerLegacyThreadUpdateTest` in the same package, which the repo uses precisely because a behavioural repro needs a live Flowable engine with a saturated async executor to force the race (flaky). The **behavioural** acceptance is the Collate DAR integration tests (`DarStatusGroupIT`, `DataAccessRequestValidationIT`, `DataAccessRequestIT`, which drive the real PolicyAgent async node) — green across 5/5 repeated concurrent runs locally, including the two previously-failing CI tests.

**Stale (reviewed the earlier `4d6bdc2c39` commit):** the findings about `findUserTaskAwaitingAsyncAdvance` / `sleepQuietly` / `ASYNC_ADVANCE_MAX_WAIT_MS` / the double 5s wait / `WorkflowHandlerAsyncAdvanceTest` reference a bounded-wait implementation that was **replaced** by the non-blocking post-commit-listener approach — that code no longer exists on the branch.

yan-3005:
Follow-up on the test finding: `CreateTaskPostCommitPersistTest` is now **behavioural** (37d026c873) — it mocks only the Flowable `TransactionContext` (framework boundary) and asserts the observable ordering: the persist does not run inline, is registered as a `TransactionState.COMMITTED` listener, runs exactly once when that listener fires, runs inline when there is no transaction context, and a listener-side persist failure is contained. No more source-text assertions.

sonika-shah (approved):

PR #31602, opened 2026-08-17
fix(e2e): isolate delete tests to adminUser to prevent cross-worker toast pollution

anuj-kumary:
## Root cause

All parallel Playwright workers log in as the same global \`admin\` user. OpenMetadata broadcasts entity change events via WebSocket to **all browser sessions of the same user**. So when any worker's delete test removes a glossary through an admin browser, the backend emits a \`deleteEntityChannel\` WebSocket event that appears as a \`"… deleted successfully!"\` toast in **every other worker's admin browser page** simultaneously.

The toast sits at \`z-[1000]\` and intercepts the pointer event on \`save-glossary\` in tests that are creating a glossary concurrently — causing an intermittent 180-second timeout in \`Glossary.spec.ts › Approve and reject glossary term from Glossary Listing\`.

## Fix

Switch the three UI-deletion tests (**Delete Modal**, **Async single**, **Async multi**) from \`performAdminLogin\` → \`performUserLogin(browser, adminUser)\`.

\`adminUser\` is a unique per-describe-block user (created in \`beforeAll\`, deleted in \`afterAll\`) with admin role. No other parallel worker is ever logged in as \`adminUser\`, so WebSocket events from its deletions are isolated to its own session and never reach other workers' admin browsers.

The two mocked-WebSocket tests are intentionally left on global admin because they deliberately intercept the global WebSocket channel.

## Test plan

- [ ] \`Glossary.spec.ts › Approve and reject glossary term from Glossary Listing\` passes without timeout across parallel workers
- [ ] \`Delete Glossary and Glossary Term using Delete Modal\` still verifies the delete UI flow correctly
- [ ] \`Async Delete - single delete success\` and \`Async Delete - multiple deletes all succeed\` pass as before


openmetadata-ui/src/main/resources/ui/playwright/utils/glossary.ts
```diff
@@ -440,6 +440,13 @@ export const createGlossary = async (
     }
   }
 
+  // A toast from a prior test's teardown sits at z-[1000] and can intercept the save click.
+  // Only stall if one is actually visible right now.
+  const alertBar = page.locator('[data-testid="alert-bar"]').first();
+  if (await alertBar.isVisible()) {
+    await alertBar.waitFor({ state: 'hidden', timeout: 15_000 });
+  }
+
```
ShaileshParmar11: this wont work if there are multiple notification

ShaileshParmar11 (approved):

PR #31606, opened 2026-08-17
feat(ingestion-pipelines): add agentType filter and stop polling orchestrators for queued status

pmbrull:
## Describe your changes

Two changes to the ingestion pipeline APIs behind the service **Agents** page, plus a crash found while testing them.

### 1. `agentType` list filter

`GET /v1/services/ingestionPipelines` had no way to ask for *"the metadata agents"*, so every caller enumerated the seven `pipelineType` values that make one up. The UI kept that list in `SERVICE_INGESTION_PIPELINE_TYPES`, Collate kept its own copy, and a caller that passed nothing got AI automations (`pipelineType=application`) mixed into the Metadata tab:

```
/api/v1/services/ingestionPipelines?fields=owners,pipelineStatuses&service=banking-bigquery&serviceType=databaseService
```

Now:

| filter | expands to |
|---|---|
| `agentType=metadata` | `metadata, usage, lineage, profiler, autoClassification, dbt, policyAgent` |
| `agentType=application` | `application` |

It is an **allow-list, not a negation**. `TestSuite`, `dataInsight` and `elasticSearchReindex` belong to neither group; implementing `metadata` as "everything that is not an application" would leak them onto the page. Setting both `agentType` and `pipelineType` intersects them, so a caller can narrow a group to one type; an empty intersection matches nothing.

The filter is **derived, not stored** — `AgentTypeResolver` expands it into the existing `pipelineType IN (...)` clause. No new entity field, no migration, and nothing that can drift out of sync with `pipelineType`.

### 2. Queued status without an orchestrator round trip

`IngestionPipelineRepository.listPipelineStatus` called `pipelineServiceClient.getQueuedPipelineStatus` on **every** read. For the K8s client that is a `listNamespacedJob` against the API server each time run history is opened, including paged historical reads that cannot contain a queued run. The Airflow call has a 10s connect timeout and no read timeout, so a slow scheduler pins the Jetty thread.

`PipelineServiceClientResponse` now carries `runId`:

```java
PipelineServiceClientResponse response = pipelineServiceClient.runPipeline(ingestionPipeline, service);
repository.recordQueuedPipelineStatus(uriInfo, ingestionPipeline.getFullyQualifiedName(), response.getRunId());
```

Orchestrators that mint the run ID when triggering report it back, and the server records the `queued` status itself. The K8s client already generated that ID and injects it into the pod as `pipelineRunId`, so the worker's later status **upserts onto the same row** (`addPipelineStatus` keys on `runId`) — no phantom run. Airflow mints its run ID inside the task (`openmetadata_managed_apis/workflows/ingestion/common.py`) and cannot report one, so it returns `null` and keeps the live-polling path completely unchanged.

**No engine branching anywhere** — the presence of a `runId` selects the behaviour. Collate's `ArgoServiceClient` gets the same treatment in a companion PR.

Recorded on three trigger paths, not just the manual one: `IngestionPipelineResource` (Run button), `AppResource` (Applications page) and `RunIngestionPipelineImpl` (AutoPilot, whose runs render on the very Agents page this change is about). Without them those runs would lose the indicator along with the poll, since `listPipelineStatus` is the single read path for queued status regardless of who triggered the run.

Deliberately **not** recorded on `RunAppImpl` or `DataContractRepository`. Those are workflow-internal: `waitForCompletion` treats any non-terminal status as still running, so the queued row buys their own polling nothing while costing an upsert plus a search reindex per run.

### 3. Crash fix

`getOperationForPipelineType` derived the workflow type **outside** its own try block, so an `application` pipeline with no `appConfig` threw `NullPointerException` before authorization ran and returned 500, instead of falling back to the generic create permission the way every other unrecognized type already does.

#
### Type of change:
- [x] Improvement

#
### High-level design:

**Why `agentType` is a query-param preset rather than an entity field.** It carries zero information that is not already derivable from `pipelineType`. Storing it would mean a schema field, a migration, a backfill, and two sources of truth that can disagree. Expanding it in `ListFilter`'s existing `pipelineType IN (...)` path costs nothing and cannot drift.

**Why the run ID lives on the response instead of a capability flag.** The alternatives were an `instanceof` check in the repository, or a `supportsStoredQueuedStatus()` method on the client interface. Both make the repository know about engines. A nullable `runId` on the response is self-selecting: an orchestrator that knows its run ID says so, one that does not stays on the old path, and a future engine opts in by populating one field.

**Rejected: correlating by writing the queued row under a server-generated ID.** For Airflow the server never learns the ID the worker will use, so the queued row would never resolve and run history would show a permanently-pending phantom next to the real run. Passing a `run_id` through `trigger.py` into `pipelineRunId` would fix that, but it spans three repos and needs a version gate for older `openmetadata-airflow-apis` that ignore the parameter. Not worth it to save one HTTP call on the engine that is least affected.

**Trade-off — a run that is accepted but never starts.** The live poll was self-cleaning: it re-derived queued state from the orchestrator each read. A stored row has no such feedback, so an unschedulable job (no capacity, ImagePullBackOff, quota) would stay `queued` forever. Queued rows older than `queuedStatusTimeoutSeconds` (new config, default 1h) are therefore hidden on read. The default is deliberately generous so a genuinely long queue does not disappear from the UI.

**Known gap — scheduled runs.** They generate their run ID outside the trigger path: CronOMJob via a reconciler placeholder, plain CronJob via the pod UID (Downward API). They get no queued row and appear when the worker posts its first status. Also lost: queued runs OpenMetadata did not trigger, e.g. a DAG started directly in the Airflow UI. Both were judged acceptable against the per-read cost.

**Backward compatibility.** `agentType` is optional and additive; omitting it preserves today's behaviour exactly. `runId` is an optional response field. `queuedStatusTimeoutSeconds` defaults to 3600. Hybrid deployments are untouched — `HybridServiceClient` overrides `getQueuedPipelineStatus` and its runner response carries no `runId`, so those pipelines keep their existing path.

#
#### Use cases covered
- The service Agents page lists only metadata agents, with AI automations excluded, without the client enumerating pipeline types
- A `pipelineType` that belongs to no agent group (`elasticSearchReindex`) is excluded from **both** groups rather than falling into `metadata`
- `agentType` combined with `pipelineType` narrows to the intersection; a disjoint combination returns nothing
- Opening run history no longer issues a Kubernetes API call
- A manually triggered K8s run reports its run ID so the server can record `queued`, and the worker's status collapses onto that same row
- A run accepted but never started stops showing as queued after the timeout
- Creating an `application` pipeline with no `appConfig` succeeds instead of returning 500

#### Unit tests
- [x] Added — `AgentTypeResolverTest` (7), `IngestionPipelineStaleQueuedStatusTest` (4), plus updated `K8sPipelineClientTest`
- `AgentTypeResolverTest` includes a guard asserting the exact set of pipeline types **excluded** from `metadata`, so a newly added `PipelineType` fails the build until someone classifies it rather than silently vanishing from the page.
- Affected suites run green: `AgentTypeResolverTest`, `IngestionPipelineStaleQueuedStatusTest`, `K8sPipelineClientTest`, `AirflowRESTClientTest`, `LogStorageTest`, `PipelineServiceClientTest`, `RunIngestionPipelineImplTest`, `IngestionPipelineRepositoryTest`, `DataContractFieldSupportTest`, `AppResourceRetryQueueTest`, `AppResourceSnapshotRedactionTest` — **156 tests, 0 failures**.

#### Integration tests
- [x] Added — `IngestionPipelineAgentTypeIT` (2 tests, real server + MySQL + Elasticsearch via Testcontainers): green.
- [x] Added — `IngestionPipelineResourceIT#test_createApplicationPipelineWithoutAppConfig`, a regression test for the NPE. **Verified RED** by stashing only `IngestionPipelineResource.java` and rebuilding: fails with the exact `NullPointerException` at `getOperationForPipelineType`. Green with the fix.
- Full `IngestionPipelineResourceIT`: **233 tests, 0 failures** (20 pre-existing skips).

#### Frontend tests
- [x] `ServiceDetailsPage` Jest suite: **54 tests, 0 failures**.

#### Manual test steps
The `agentType` filter is exercised end-to-end by `IngestionPipelineAgentTypeIT` against a real server rather than by hand. The trigger → `queued` → worker-status transition needs a live K8s or Argo orchestrator, which neither the IT stack (Airflow client pointed at nothing) nor Jest can provide; it is covered structurally instead — `K8sPipelineClientTest` asserts the run ID is reported on the response and that the queued read costs no API call, and `addPipelineStatus`'s existing upsert-by-`runId` is what collapses the rows. Worth a look on a real cluster before release.

#
### UI screen recording / screenshots

No visual change. The only UI diff swaps the query the Agents tab sends:

```diff
-  pipelineType: SERVICE_INGESTION_PIPELINE_TYPES,
+  agentType: AgentType.Metadata,
```

The rendered list is identical — same pipelines, same order. `SERVICE_INGESTION_PIPELINE_TYPES` stays for its two non-filter uses (building the "Add Agent" menu, and filtering pipelines discovered over SSE in `useMetadataAgents`), which cannot move server-side.

#
### Checklist:
- [x] Generated TypeScript types regenerated from the schema changes (`json2ts-generate-all.sh`) — 4 files, no unrelated drift
- [x] `mvn spotless:check` clean
- [x] UI ESLint + Prettier + organize-imports clean (0 errors)
- [x] `tsc --noEmit` reports nothing in the changed files
- [x] No `.github/workflows/**` changes
- [ ] Linked issue — none; happy to open one if you want the discussion recorded

The PR adds server-side agent-type filtering and replaces Kubernetes queued-status polling with persisted queued statuses keyed by orchestrator run ID. It also adds expiration for stale queued statuses, extends trigger paths to record queued runs, and handles application pipelines without app configuration.
- Adds the `metadata` and `application` agent groups and updates the Agents page query.
- Propagates Kubernetes run IDs and records queued status after manual, application, and workflow-triggered runs.
- Adds configurable stale-queued visibility and coverage for filtering, status handling, and authorization fallback.
</details>

The PR is not yet safe to merge because stale queued rows can still displace valid runs from limited history responses.

The database limit is applied before stale queued statuses are filtered, and the repository does not fetch replacement rows afterward, so valid older history entries can be omitted.

**Files Needing Attention:** openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/IngestionPipelineRepository.java | Adds persisted queued-status recording and timeout-based filtering across status response paths. |
| openmetadata-service/src/main/java/org/openmetadata/service/clients/pipeline/k8s/K8sPipelineClient.java | Returns the generated run ID from triggers and removes Kubernetes API polling from queued-status reads. |
| openmetadata-service/src/main/java/org/openmetadata/service/resources/services/ingestionpipelines/AgentTypeResolver.java | Defines explicit agent-group allow-lists and intersection behavior for pipeline-type filtering. |
| openmetadata-service/src/main/java/org/openmetadata/service/resources/services/ingestionpipelines/IngestionPipelineResource.java | Exposes the agentType filter, records queued statuses after triggers, and safely falls back to generic create authorization. |
| openmetadata-ui/src/main/resources/ui/src/pages/ServiceDetailsPage/ServiceDetailsPage.tsx | Replaces the client-maintained metadata pipeline-type list with the new server-side agentType filter. |

</details>

```mermaid
sequenceDiagram
  participant UI as Agents UI
  participant API as Ingestion Pipeline API
  participant Repo as Pipeline Repository
  participant K8s as Kubernetes Client
  participant DB as Status Store
  UI->>API: List pipelines with agentType
  API->>Repo: Resolve agentType to pipelineType set
  Repo-->>UI: Filtered pipelines
  UI->>API: Trigger pipeline
  API->>K8s: runPipeline
  K8s-->>API: response with runId
  API->>Repo: recordQueuedPipelineStatus(runId)
  Repo->>DB: Upsert queued status
  Note over DB: Worker updates the same runId later
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(ingestion-pipelines): apply the stal..."](https://github.com/open-metadata/openmetadata/commit/d7279e0ca6f5fd4ce0a59277843013d94a75f2cd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53908407)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [Governance Workflows and Applications in OpenMetadata Service](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/openmetadata-service-workflows-apps.md)
- Knowledge Base — [openmetadata-spec: JSON Schema specification and multi-language code generation](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/openmetadata-spec.md)
</details>

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/IngestionPipelineAgentTypeIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/IngestionPipelineAgentTypeIT.java
@@ -1,0 +1,133 @@
+package org.openmetadata.it.tests;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.time.Instant;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+import org.openmetadata.it.factories.DashboardServiceTestFactory;
+import org.openmetadata.it.util.SdkClients;
+import org.openmetadata.it.util.TestNamespace;
+import org.openmetadata.it.util.TestNamespaceExtension;
+import org.openmetadata.schema.api.services.ingestionPipelines.CreateIngestionPipeline;
+import org.openmetadata.schema.entity.services.DashboardService;
+import org.openmetadata.schema.entity.services.ingestionPipelines.AirflowConfig;
+import org.openmetadata.schema.entity.services.ingestionPipelines.IngestionPipeline;
+import org.openmetadata.schema.entity.services.ingestionPipelines.PipelineType;
+import org.openmetadata.schema.metadataIngestion.ApplicationPipeline;
+import org.openmetadata.schema.metadataIngestion.DashboardServiceMetadataPipeline;
+import org.openmetadata.schema.metadataIngestion.SourceConfig;
+import org.openmetadata.schema.utils.ResultList;
+import org.openmetadata.sdk.client.OpenMetadataClient;
+import org.openmetadata.sdk.network.HttpMethod;
+
+/**
+ * Integration tests for the {@code agentType} list filter on {@code
+ * GET /v1/services/ingestionPipelines}, which expands to the set of {@code pipelineType} values
+ * that make up an agent group so clients do not have to enumerate them.
+ */
+@Execution(ExecutionMode.CONCURRENT)
+@ExtendWith(TestNamespaceExtension.class)
+public class IngestionPipelineAgentTypeIT {
+
+  private static final Date START_DATE = Date.from(Instant.parse("2022-06-10T15:06:47Z"));
+  private static final String LIST_PATH = "/v1/services/ingestionPipelines";
+
+  @Test
+  void test_agentTypeMetadata_returnsOnlyMetadataAgents(TestNamespace ns) {
+    OpenMetadataClient adminClient = SdkClients.adminClient();
+    DashboardService service = DashboardServiceTestFactory.createMetabase(ns);
+    String serviceFqn = service.getFullyQualifiedName();
+
+    try {
+      String metadataPipeline = createPipeline(ns, service, "agentMetadata", PipelineType.METADATA);
+      String lineagePipeline = createPipeline(ns, service, "agentLineage", PipelineType.LINEAGE);
+      String reindexPipeline =
+          createPipeline(ns, service, "agentReindex", PipelineType.ELASTIC_SEARCH_REINDEX);
+      String applicationPipeline =
+          createPipeline(ns, service, "agentApplication", PipelineType.APPLICATION);
+
+      Set<String> metadataAgents = listNames(adminClient, serviceFqn, "agentType=metadata");
+      assertTrue(
+          metadataAgents.containsAll(Set.of(metadataPipeline, lineagePipeline)),
+          "agentType=metadata must return every metadata pipelineType");
+      assertTrue(
+          Set.of(reindexPipeline, applicationPipeline).stream().noneMatch(metadataAgents::contains),
+          "agentType=metadata must not fall back to 'everything that is not an application'");
+
+      Set<String> applicationAgents = listNames(adminClient, serviceFqn, "agentType=application");
+      assertEquals(Set.of(applicationPipeline), applicationAgents);
+    } finally {
+      adminClient
+          .dashboardServices()
+          .delete(service.getId().toString(), Map.of("hardDelete", "true", "recursive", "true"));
+    }
+  }
+
+  @Test
+  void test_agentTypeIsIntersectedWithPipelineType(TestNamespace ns) {
+    OpenMetadataClient adminClient = SdkClients.adminClient();
+    DashboardService service = DashboardServiceTestFactory.createMetabase(ns);
+    String serviceFqn = service.getFullyQualifiedName();
+
+    try {
+      createPipeline(ns, service, "narrowMetadata", PipelineType.METADATA);
+      String lineagePipeline = createPipeline(ns, service, "narrowLineage", PipelineType.LINEAGE);
+      createPipeline(ns, service, "narrowApplication", PipelineType.APPLICATION);
+
+      assertEquals(
+          Set.of(lineagePipeline),
+          listNames(adminClient, serviceFqn, "agentType=metadata&pipelineType=lineage"),
+          "Both filters set must intersect, not union");
+      assertTrue(
+          listNames(adminClient, serviceFqn, "agentType=metadata&pipelineType=application")
+              .isEmpty(),
+          "An empty intersection must match no pipeline");
+    } finally {
+      adminClient
+          .dashboardServices()
+          .delete(service.getId().toString(), Map.of("hardDelete", "true", "recursive", "true"));
+    }
+  }
+
+  private String createPipeline(
+      TestNamespace ns, DashboardService service, String name, PipelineType pipelineType) {
+    SourceConfig sourceConfig =
+        PipelineType.APPLICATION.equals(pipelineType)
+            ? new SourceConfig()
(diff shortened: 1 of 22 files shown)


chirag-madlani (dismissed):

chirag-madlani (approved):

mohityadav766 (approved):

pmbrull:
picked to 2.0

PR #31614, opened 2026-08-17
fix(security): name the token's own claims when principal resolution finds none

### LINKED ISSUE #31615: Not getting the configured claims
Related to JWT claims:
While Authentication, not getting the configured claims could blocker the end user about the missing pieces.

Solution:
Add list of both expected and observed claims so that it's clear for user's what to add

Khairajani:
Fixes #31615 

When a token carries none of the claims a deployment resolves principals from, `SecurityUtil.getFirstMatchJwtClaim` reported only what was expected:

```
Invalid JWT token, none of the following claims are present [email]
```

That names the configuration but not the token, so an administrator cannot tell what to set `jwtPrincipalClaims` *to* — only that the current value is wrong.

It matters because of where this fails. `JwtFilter` runs this on every request, so the symptom is a total lockout: every API call 401s and the UI logs the user out with no explanation. The server log is the only diagnostic available, and it withheld the one fact that resolves the incident.

Now both sides are named:

```
Invalid JWT token, none of the configured principal claims [email] are present
-- the token carries [aud, exp, iat, iss, oid, preferred_username, sub, tid].
Set jwtPrincipalClaims to a claim this identity provider issues.
```

**Claim names only** — values identify people and must not reach a log; a test pins that.

This is a common shape rather than an edge case. Azure AD / Entra emits `email` only for accounts with a populated `mail` attribute, so a deployment configured for `email` locks out exactly those users who lack one while working for everyone else. One line from the message above identifies it; without it the investigation is largely guesswork.

Tests: the existing assertion is updated to the new wording, plus one new test covering that the token's own claim names appear and its claim *values* do not. Verified to fail without the change. No other test or source in the repo asserts the old string.

The PR improves JWT principal-resolution diagnostics by naming both the configured principal claims and the claim names carried by the token, while deliberately excluding claim values.
- Sorts token claim names for deterministic exception output.
- Updates the existing assertion and adds coverage ensuring claim names appear without exposing claim values.
</details>

The PR appears safe to merge with no blocking or independently actionable issues identified.

The changed failure path preserves authentication behavior, emits only claim names from a signature-verified token, and has focused tests preventing claim-value disclosure.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/security/SecurityUtil.java | Enhances the missing-principal-claim AuthenticationException with a deterministic list of token claim names and configuration guidance; no actionable defect found. |
| openmetadata-service/src/test/java/org/openmetadata/service/security/SecurityUtilTest.java | Updates the expected diagnostic and verifies that claim names, but not identifying claim values, are included. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(security): name the token&#39;s own clai..."](https://github.com/open-metadata/openmetadata/commit/6ffa4e9f9ba95f33f32769f4cf7198177eb9224d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53918454)</sub>

**Context used:**

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/security/SecurityUtil.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/security/SecurityUtil.java
@@ -41,6 +41,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
@@ -246,8 +247,11 @@ public final class SecurityUtil {
         .orElseThrow(
             () ->
                 new AuthenticationException(
-                    "Invalid JWT token, none of the following claims are present "
-                        + jwtPrincipalClaimsOrder));
+                    String.format(
+                        "Invalid JWT token, none of the configured principal claims %s are present"
+                            + " -- the token carries %s. Set jwtPrincipalClaims to a claim this"
+                            + " identity provider issues.",
+                        jwtPrincipalClaimsOrder, new TreeSet<>(claims.keySet()))));
   }
 
   /**
--- a/openmetadata-service/src/test/java/org/openmetadata/service/security/SecurityUtilTest.java
+++ b/openmetadata-service/src/test/java/org/openmetadata/service/security/SecurityUtilTest.java
@@ -381,7 +381,30 @@ class SecurityUtilTest {
             AuthenticationException.class,
             () -> SecurityUtil.getFirstMatchJwtClaim(List.of("email"), Map.of("sub", "1234")));
 
-    assertTrue(exception.getMessage().contains("none of the following claims are present"));
+    assertTrue(exception.getMessage().contains("none of the configured principal claims"));
+  }
+
+  @Test
+  void testGetFirstMatchJwtClaimNamesTheClaimsTheTokenActuallyCarries() {
+    // Naming only what was expected leaves an admin guessing what to configure instead: the token's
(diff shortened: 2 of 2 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/security/SecurityUtil.java
```diff
@@ -246,8 +247,11 @@ public static String getFirstMatchJwtClaim(
         .orElseThrow(
             () ->
                 new AuthenticationException(
-                    "Invalid JWT token, none of the following claims are present "
-                        + jwtPrincipalClaimsOrder));
+                    String.format(
+                        "Invalid JWT token, none of the configured principal claims %s are present"
+                            + " -- the token carries %s. Set jwtPrincipalClaims to a claim this"
+                            + " identity provider issues.",
+                        jwtPrincipalClaimsOrder, new TreeSet<>(claims.keySet()))));
```
mohityadav766, line 254: We cannot expose the claims @Khairajani

mohityadav766 (approved):

Khairajani:
Closing — the objection is correct and the remainder isn't worth reviewer time.

`AuthenticationException` serialises its message into the 401 response body, so this would have sent the token's claim names to the caller. I should have checked that before opening it. With the claim names removed, what's left is a rewording plus a pointer to `jwtPrincipalClaims`, which isn't worth the churn.

For anyone who finds this later: the failure it targeted is a total lockout — `JwtFilter` runs `getFirstMatchJwtClaim` on every request, so a `jwtPrincipalClaims` value matching none of the token's claims 401s every call and the UI logs the user out with no explanation. The server log is the only diagnostic, and it names the configured claims without naming what the token actually carried. If that's ever worth improving, the claim names belong in a `LOG.warn` (`SecurityUtil` already has `@Slf4j`), not in the exception message.

PR #31634, opened 2026-08-17
fix(alerts): redact destination config from the testDestination response

### LINKED ISSUE #31633: Backend: testDestination response echoes submitted destination config, including webhook auth secrets
## Problem

`POST /api/v1/events/subscriptions/testDestination` (the **Test Destinations** button on the alert create/edit page) echoes the submitted destinations back in the 200 response verbatim, `config` included.

For an external webhook destination using OAuth2 auth, the response therefore contains the plaintext `config.authType.clientSecret` and `clientId`; with bearer auth it contains `secretKey`. Custom `headers` and `queryParams`, which routinely carry `Authorization` values, are echoed as well.

```json
"config": {
  "authType": {
    "type": "oauth2",
    "scope": "...",
    "clientId": "...",
    "tokenUrl": "...",
    "clientSecret": "..."
  }
}
```

Reported against 1.13.3 by an admin who inspected the browser network response after triggering a test.

The caller already holds the config it just submitted, so nothing in the response needs to carry it back. Echoing it puts credentials into HAR captures, proxy and APM logs, screen shares, and support bundles.

## Root cause

`EventSubscriptionResource.sendTestMessageAlert` adds the request's own `SubscriptionDestination` objects to `resultDestinations` and returns them. The publishers mutate `statusDetails` onto those same instances, so the response body is literally `request + statusDetails`.

The schema already declares intent that does not reach this path: `entity/events/webhook.json` marks `authType` with `"mask": true`, and the values are Fernet-encrypted at rest, but neither applies here because `SubscriptionDestination.config` is an untyped `Object` and this response goes through the ordinary JAX-RS mapper.

### Secondary bug in the same method

A destination whose publisher throws `EventPublisherException` (connection refused, unknown host) is logged and **dropped** from the response array. The user gets no entry and therefore no status at all for the destination that actually failed, and the response length silently stops matching the request.

## Proposed fix

Return one entry per requested destination, in request order, with `config` redacted and `statusDetails` populated, and include destinations whose delivery threw as `Failed` with a reason instead of dropping them.

On the UI side, the destination rows currently locate their result by deep-equality on `{type, category, config}` against live form state, so they must take only `statusDetails` from the response and keep the config they submitted. That deep-equality doubles as a staleness guard (editing a field after a test correctly clears the banner), so it is worth preserving rather than replacing with index-based props.

## Known gaps in the same area, not covered by this issue

- `webhookOAuth2Config.json` (`clientSecret`, `clientId`) and `webhookBearerAuth.json` (`secretKey`) declare no `"format": "password"`, so `PasswordAnnotator` never emits `@PasswordField` and `PasswordEntityMasker` is a no-op on webhook auth. GET, list, and `{id}/listDestinations` return the stored value unmasked.
- `Fernet.encryptWebhookSecretKey` gates on `type == WEBHOOK`, so `Slack`, `MsTeams`, and `GChat` destinations, which use the same `Webhook` config POJO, store and return their `authType` in plaintext.
- `TestDestinationStatus.entity` carries the remote endpoint's response body, which an echo-style endpoint could use to reflect a credential back.

Happy to file these as a separate hardening issue.

manerow:
### Describe your changes:

Fixes #31633

`POST /v1/events/subscriptions/testDestination` returned the caller's own `SubscriptionDestination` objects, so the 200 response echoed `config` back verbatim, including the plaintext webhook `authType.clientSecret` / `clientId` (OAuth2), `secretKey` (bearer), and any credential-bearing `headers` or `queryParams`. The caller already holds that config, so nothing in the response needs to carry it.

The endpoint now returns one entry per requested destination, in request order, with `config` redacted and only `statusDetails` populated. Destinations whose delivery threw `EventPublisherException` were previously logged and dropped from the array; they are now returned with a `Failed` status and a reason, which both fixes a real bug (a connection failure produced no status banner at all) and makes the response length and order match the request.

On the UI side the destination rows locate their result by deep-equality on `{type, category, config}` against live form state, so they can no longer rely on the echoed config. The three call sites now keep the config they submitted and take only `statusDetails` from the response, via a new shared helper `getDestinationsWithTestStatus` in `AlertsUtilPure`.

#
### Type of change:

- [x] Bug fix

#
### High-level design:

The UI's deep-equality match is not only correlation, it is also a **staleness guard**: each row re-reads its values from live form state, so editing a field after a test correctly clears the banner. Attaching status by index prop instead would have regressed that behaviour and touched three child components plus their tests. Rebuilding `destinationsWithStatus` from the submitted array preserves child behaviour bit-for-bit while removing the dependency on the wire config entirely, since the submitted array is exactly what the server used to echo.

Redaction rather than masking: masking the secret fields would still break the deep-equality match, and precisely for the destinations that carry secrets. A `SubscriptionDestination` without `config` is the typed result this endpoint should have returned in the first place.

The response shape stays `SubscriptionDestination[]` (the field is simply absent), so no schema change and no client typing change is required.

Deliberately out of scope, listed in the issue: webhook auth fields still lack `"format": "password"` so `PasswordEntityMasker` is a no-op on them; `Fernet.encryptWebhookSecretKey` still gates on `type == WEBHOOK` leaving Slack/MsTeams/GChat auth in plaintext at rest; `TestDestinationStatus.entity` still carries the remote response body.

#
#### Use cases covered
- Admin clicks **Test Destinations** with an OAuth2 webhook: the response carries the status but no `config`, and the success banner still renders on the correct row.
- A destination that cannot be reached (connection refused) now returns a `Failed` status with a reason instead of vanishing from the response.
- Editing a destination field after a test still clears its banner (staleness guard preserved).

#### Unit tests
- Added `openmetadata-ui/src/main/resources/ui/src/utils/Alerts/AlertsUtilPure.test.ts` (3 tests) covering config retention, positional pairing, and a short/missing result array.
- Updated the mocked response in `DestinationFormItem.test.tsx` to the new config-less contract.
- `yarn test src/components/Alerts/DestinationFormItem src/components/Alerts/DestinationFormItemV2` → 65 passed.

- Added `EventSubscriptionResourceIT#test_testDestinationRedactsConfigAndReportsFailures`, which posts a reachable OAuth2 webhook plus an unreachable one against an in-JVM `HttpServer`, and asserts the raw response body contains neither the sentinel secret nor `clientSecret` nor `authType`, that both entries come back in request order with `config == null`, and that the unreachable one reports `Failed` with a reason.
- Verified as a genuine regression guard: against the unfixed server it fails on `Response echoed the webhook client secret`; against the fixed server it passes.
- Full suite: `mvn test -pl openmetadata-integration-tests -Dtest=EventSubscriptionResourceIT` → 243 tests, 0 failures.

Not applicable (no ingestion changes).

#### Playwright (UI) tests
- Extended `playwright/e2e/Flow/NotificationAlerts.spec.ts` with `expect(testResult.config).toBeUndefined()` to pin redaction end-to-end. The spec's existing length-2 assertion also becomes deterministic, since a transport failure no longer removes an entry.

#### Manual testing performed
Verification so far is the automated coverage above (backend IT run locally against the built server, jest, and the Playwright assertion); I have not yet driven this through a browser session end-to-end. Happy to attach a recording if reviewers would like one before merge.

#

No visual change to the success path: the same banner renders in the same place, sourced from the submitted config instead of the echoed one. The one user-visible improvement is that an unreachable destination now renders a `Failed` banner where it previously rendered nothing.

#
### Checklist:

- [x] My PR is linked to a GitHub issue via `Fixes #31633` above.
- [x] For JSON Schema changes: not needed, no schema change (the response field is simply omitted).
- [x] I have added a test that covers the exact scenario we are fixing, and confirmed it fails without the fix.

The PR prevents the destination-test endpoint from echoing credential-bearing destination configuration and preserves request-order status correlation in the UI.
- Returns a config-free result for every tested destination, including delivery failures.
- Reconstructs UI results from the submitted destination snapshots and returned status details.
- Adds backend integration, frontend unit, and Playwright coverage for redaction and failure reporting.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/resources/events/subscription/EventSubscriptionResource.java | Redacts destination configuration, preserves positional results, and converts publisher exceptions into failed status entries. |
| openmetadata-service/src/main/java/org/openmetadata/service/events/subscription/AlertUtil.java | Adds a status-construction overload used to report delivery exceptions without exposing destination configuration. |
| openmetadata-ui/src/main/resources/ui/src/utils/Alerts/AlertsUtilPure.ts | Recombines submitted destination snapshots with config-free response statuses in request order. |
| openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItem.component.tsx | Adopts the shared status-merging helper for the original destination form. |
| openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItemV2/DestinationFormItemV2.component.tsx | Adopts the shared status-merging helper for the V2 destination form. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/EventSubscriptionResourceIT.java | Verifies credential redaction, ordered result cardinality, and failed delivery reporting. |
| openmetadata-ui/src/main/resources/ui/src/utils/Alerts/AlertsUtilPure.test.ts | Covers config retention, positional status pairing, and missing response entries. |

</details>

```mermaid
sequenceDiagram
  participant UI as Alerts UI
  participant API as testDestination API
  participant Dest as Destination Publisher
  UI->>API: POST submitted destinations with config
  loop Each destination in request order
    API->>Dest: Send test message
    Dest-->>API: Status or delivery exception
    API->>API: Build statusDetails and remove config
  end
  API-->>UI: Config-free ordered results
  UI->>UI: Combine submitted config with returned statusDetails
```
</details>

<sub>Reviews (5): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/test-destin..."](https://github.com/open-metadata/openmetadata/commit/7d4cdfac29f7157f66530e7a2678ade22f699e49) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53965956)</sub>


manerow:
Collate-side consumer update (private repo): openmetadata-collate#5840. It should merge after this PR lands and the submodule pin is bumped.

ShaileshParmar11 (approved):

yan-3005 (approved):

manerow:
manually picked into 1.13

PR #31647, opened 2026-08-17
Fix flaky UserDetails Playwright spec: visitUserProfilePage hangs on an undispatched search

harsh-vador:
### Describe your changes:

I worked on the flaky `UserDetails.spec.ts` test because it fails intermittently on the nightly run ([2.0 nightly run 31918043673](https://github.com/open-metadata/openmetadata-nightly/actions/runs/31918043673)) with a 49s hang inside `visitUserProfilePage`, not inside the spec itself.

**Failing test:** `Create team with domain and verify visibility of inherited domain in user profile after team removal`

**Trace:** `Wait for event "response" — playwright/utils/user.ts:139` → **49.1s**, followed by `Timed out waiting for user <name> to become visible in the user list`. The failure screenshot shows the user row rendered correctly — nothing was actually missing.

#### Root cause

`visitUserProfilePage` retried its user-list search inside an `expect.poll` by clearing the search box and re-typing the same username:

```ts
const searchRequest = page.waitForResponse('/api/v1/search/query*'); // unbounded
await searchBar.fill('');
await searchBar.fill(userName);
await searchRequest;
```

That retry can never dispatch a second search:

- `SearchBar.component.tsx:131` sets `typingInterval: 1000`, so `fill('')` and `fill(userName)` (~20ms apart) collapse into a **single** debounced `handleSearch(userName)`.
- `UserListPageV1.tsx:230` — `handleSearch` only calls `setFilters({ user: value })`.
- `UserListPageV1.tsx:247-250` — the fetch effect keys on `[searchValue, currentPage, isDeleted]`. Re-entering the term already held in `searchValue` leaves it unchanged, so **the effect never re-runs and no request is made**.

The first iteration works (empty box → `'' → userName` is a real change). Every later iteration deadlocks on an unbounded `waitForResponse` for a request that will never be sent.

Two aggravating factors:

1. The poll budget was `60000`, equal to the test timeout in `playwright.config.ts:386`. The poll could never exhaust its own retries, so the test died mid-poll and reported the raw wait rather than the poll's diagnostic message.
2. The sibling test `Admin user can edit teams from the user profile` calls the same helper but declares `test.slow()`. This one did not, so it ran on the bare 60s budget.

Why it surfaces on nightly and not locally: CI runs `workers: 3`, which lengthens ES index lag and makes the helper reach retry 2+ far more often. Locally the first iteration usually succeeds and the latent deadlock never fires.

#### The fix

Replace the retry loop with a single deterministic search. The search box is empty on arrival, so one `fill(userName)` is always a real value change and always dispatches:

```
settingClick → loader detached → arm response → fill → await → loader detached → expect visible → click
```

- Response predicate is scoped to the username, so a stray `search/query` from another component cannot satisfy the wait and mask an unpopulated list.
- Dropped two dead waits (`userResponse`, created and never meaningfully awaited; `loaderPromise`, created once outside the loop and re-awaited each iteration despite being already settled).
- Added a comment recording *why* re-typing cannot work, so the retry is not reintroduced.
- Added `test.slow()` to the failing test, matching its sibling that uses the same helper.

Net **−12 lines**; the helper now matches the shape of `searchUserByEmail` in the same file.

The removed poll was there to absorb ES index lag, but since its retry was inert the helper has in practice always been a single search — and it passes nearly always, which is itself the evidence that index lag is not a real factor on this path. If a user genuinely is not indexed, `expect(userRow).toBeVisible()` now fails with a clear message pointing at a real bug rather than a loop hiding it.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered
- Admin navigates to Settings → Members → Users, searches for a user, and opens that user's profile page (`visitUserProfilePage`, used by `UserDetails.spec.ts` and other specs).

#### Unit tests
Not applicable — this PR only changes Playwright test code.

#### Playwright (UI) tests
This PR *is* a Playwright fix; no product code changed, so no new test is added.
- Files updated:
  - `openmetadata-ui/src/main/resources/ui/playwright/utils/user.ts`
  - `openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/UserDetails.spec.ts`

#### Manual testing performed
Static verification only — I do not have a live stack available:

1. `npx tsc --noEmit -p playwright/tsconfig.json` — no new errors. The two remaining `TS6133 'displayName'` reports are pre-existing, in `softDeleteUserProfilePage` / `hardDeleteUserProfilePage`, untouched by this PR.
2. `npx eslint playwright/utils/user.ts playwright/e2e/Pages/UserDetails.spec.ts` — 0 errors. The 2 warnings on the spec are pre-existing `browser.newPage()` notices that the rule itself documents as expected for multi-user tests.
3. `npx prettier --check` — clean.
4. Root cause confirmed by reading the product code rather than by reproduction: the debounce interval, `handleSearch`, and the fetch effect's dependency array are cited with line numbers above.

**Reviewer note:** the flake is a race, so please run the spec against a live stack before merging to confirm — `yarn playwright:run playwright/e2e/Pages/UserDetails.spec.ts --repeat-each=5`.

#

Not applicable — no product code changed, test-only.

#
### Checklist:
- [x] For UI changes: not applicable, test-only change.

- [x] The change is itself a test fix; the repaired `visitUserProfilePage` covers the exact scenario that was hanging.

The PR simplifies `visitUserProfilePage` to issue one deterministic user search and increases the timeout for the previously flaky scenario.
- Removes the ineffective polling and repeated response waits from the shared Playwright helper.
- Waits for the user-list loader around the search request before opening the matching row.
- Marks the inherited-domain user-profile test as slow.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/playwright/utils/user.ts | Replaces the deadlocking retry loop with a single synchronized search, loader wait, visibility assertion, and row click. |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/UserDetails.spec.ts | Marks the affected end-to-end scenario as slow to provide an appropriate execution budget. |

</details>

<sub>Reviews (3): Last reviewed commit: ["address gitar"](https://github.com/open-metadata/openmetadata/commit/2617a66ab017d991b6c27b66410fde9196a400bd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53997749)</sub>


harsh-vador:
Good catch — accepted and fixed in `2617a66`. The concern is real, and the divergence is actually wider than encoder mismatch: the frontend never percent-encodes `q` at all.

`rawSearchQuery` interpolates the term straight into the URL string rather than passing it through axios `params`:

```ts
// src/rest/searchAPI.ts:187
const apiUrl = `/search/query?q=${apiQuery}${filters ?? ''}`;
```

So there are two independent ways the predicate could miss:

| Source | Frontend emits | `encodeURIComponent` produces |
|---|---|---|
| `searchAPI.ts:187` — raw interpolation, no encoding | literal `@` `:` `,` `$` | `%40` `%3A` `%2C` `%24` |
| `StringUtils.ts:90` — `getQueryWithSlash` backslash-escapes quotes | `\"` | `%22`, no backslash |

Either one means `response.url().includes(...)` never matches and the wait hangs to the default timeout — exactly the failure class this PR removes, so it would have been a self-inflicted regression.

Fixed as suggested, matching the URL shape rather than the term:

```ts
const searchResponse = page.waitForResponse(
  '/api/v1/search/query?q=*&index=user&from=0&size=*'
);
```

Kept `index=user` rather than `searchUserByEmail`'s `index=*` so the wait stays pinned to the user-list search instead of matching any search on the page — costs nothing and is encoding-independent. Correctness of *which* user came back is still covered by the `expect(userRow).toBeVisible()` assertion right after, per your suggestion.

`from=0` is safe to pin: `handleSearch` resets paging to `INITIAL_PAGING_VALUE` before the fetch effect runs.

ShaileshParmar11 (approved):

PR #31656, opened 2026-08-17
fix(cache): invalidate persona AI context caches across pods

pmbrull:
## Problem

`PersonaContextCache.local` (Caffeine) and `SubjectCache.USER_CONTEXT_CACHE` (Guava) are per-JVM and were never reached by the `CacheInvalidationPubSub` subscriber — neither implements `Invalidatable` and neither is passed to `CacheBundle.registerInvalidatable`. On a multi-replica deployment a persona change is visible on one pod and stale on the others for up to that cache's TTL (30 min / 15 min), intermittently, depending on which pod the load balancer picks.

Symptoms, all reported against Ask Collate:

- **Admin "regenerate persona context" appears broken.** `refresh()` rewrites Redis and the calling pod's local entry only. The definition hash doesn't change, so peers keep serving the identically-keyed stale entry until their own TTL.
- **A persona switch intermittently doesn't take.** After assigning a persona, peers keep a `User` without it, so `SubjectContext.getActivePersona()` silently discards the `X-OpenMetadata-Persona` header and falls back to the default persona (`Requested persona '<id>' is not assigned to user '<user>'`).
- **Same user, same question, different answers depending on routing.**

Reported downstream as open-metadata/openmetadata-collate#5847, which has the full analysis.

## Changes

- **`PersonaContextCache implements Invalidatable`**, registered with `CacheBundle`. A peer can't know the definition hash the local entry is keyed by (`personaId:definitionHash`), so it drops by `personaId:` prefix and clears `generationStates`.

  The drop is **local only**, deliberately: the publishing pod already wrote (or deleted) the authoritative Redis copy, and a peer deleting those keys would throw away a document that was just rebuilt and make every pod re-run the ES-heavy build.

- **`refresh()` publishes explicitly.** It rewrites Redis but mutates no entity, so nothing broadcast it before. It goes out under a non-entity type (`CacheInvalidationPubSub.TYPE_PERSONA_CONTEXT`) so peers drop the document without bumping the persona write epoch or evicting entity caches for a change that touched no entity. The handler in `CacheBundle` gates the entity-cache work on that type; the `Invalidatable` fan-out still runs.

- **`SubjectCache` registers an `Invalidatable` too.** A `user` write drops that user's context; a `persona` or `team` write drops all of them (a team's `defaultPersona` reaches users as `inheritedPersonas`, via membership or the parent hierarchy).

  The all-drop is blunt on purpose — the affected user set isn't derivable from the message, since assignments change through the persona, the user's `defaultPersona`, and a team's default alike. Persona writes are rare admin actions and the reload is a cached entity read. A context `refresh` publishes under `TYPE_PERSONA_CONTEXT` and so does not land here.

  The user match is case-insensitive: a user's FQN is the lower-cased quoted name, while the cache is keyed by the principal name as the request presented it.

- **`PersonaRepository.postCreate` publishes.** `EntityRepository.postCreate` calls `CacheBundle.invalidateEntity` (local fan-out) but never publishes — only the update and delete paths do. So a persona created with users already assigned never reached peers at all.

- **Local context TTL capped at 60 s while Redis is up.** Nothing invalidates when a *referenced* asset changes (a retagged table, an edited KB article, a new asset matching a rule), and that drift has no single entity to key an invalidation off — the cap is what bounds it, and it also bounds any dropped pub/sub message. Expiry then costs one `mget` of the shared copy, not a rebuild.

  With Redis down the local entry *is* the cache, so the definition's `cacheTtlMinutes` still stands there — capping unconditionally would make every pod re-run the build every minute.

## Testing

`PersonaContextCacheTest` (2 new) and `SubjectCacheTest` (3 new) cover the remote-invalidate path: a message in, local entry gone, next read rebuilds; unrelated entity types leave the entry warm.

Both were confirmed non-vacuous by short-circuiting the two invalidators and watching all four positive tests fail (`expected: <MISS> but was: <HIT>`), then restoring.

- `mvn spotless:check` clean, `compile` clean.
- 315 tests green across `aicontext`, `cache`, `policyevaluator`, and `PersonaRepositoryTest`.

## Not covered

- No Redis-backed multi-pod integration test — the wiring is exercised at the `Invalidatable` boundary only.
- `generationStates` is still pod-local, so `GET /v1/personas/{id}/context/status` can still report `GENERATING` on one pod and `FRESH`/`STALE` on another. Invalidation now clears it, but making the status genuinely shared needs Redis-backed state.
- Invalidating on referenced-asset changes remains a design question rather than wiring; the TTL cap is the interim answer.

Fixes open-metadata/openmetadata-collate#5847

The PR adds distributed invalidation for persona AI context and user-context caches so persona changes converge across service pods.
- Registers persona and subject cache invalidators with the shared cache bundle.
- Publishes persona-context refresh and persona-create events to peer pods.
- Handles persona, team, and user invalidations with cache-specific eviction behavior.
- Caps local persona-context freshness while Redis is available.
- Adds focused tests for remote invalidation behavior.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/aicontext/PersonaContextCache.java | Adds local-only distributed invalidation, explicit refresh publication, and a Redis-aware local TTL cap. |
| openmetadata-service/src/main/java/org/openmetadata/service/cache/CacheBundle.java | Registers the new invalidators and separates non-entity persona-context signals from entity-cache eviction. |
| openmetadata-service/src/main/java/org/openmetadata/service/cache/CacheInvalidationPubSub.java | Defines the non-entity persona-context invalidation message type. |
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/PersonaRepository.java | Publishes persona creation so peer pods invalidate contexts for initially assigned users. |
| openmetadata-service/src/main/java/org/openmetadata/service/security/policyevaluator/SubjectCache.java | Clears peer user contexts for persona and team changes and selectively evicts users for USER messages. |
| openmetadata-service/src/test/java/org/openmetadata/service/aicontext/PersonaContextCacheTest.java | Exercises persona-context remote invalidation and unrelated-message behavior. |
| openmetadata-service/src/test/java/org/openmetadata/service/security/policyevaluator/SubjectCacheTest.java | Covers persona, team, user, and unrelated invalidation behavior for cached user contexts. |

</details>

```mermaid
sequenceDiagram
  participant Admin
  participant Writer as Writing pod
  participant Redis
  participant Peer as Peer pod
  participant Subject as SubjectCache
  Admin->>Writer: Update team/persona or refresh context
  Writer->>Redis: Update or invalidate shared cache
  Writer->>Redis: Publish invalidation message
  Redis-->>Peer: TEAM / USER / PERSONA / personaContext
  Peer->>Subject: Fan out through registered Invalidatable
  Subject->>Subject: Evict affected local user contexts
  Note over Peer: Next request reloads current shared or persisted state
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(cache): drop peer user contexts on t..."](https://github.com/open-metadata/openmetadata/commit/c059194c57ade45531098543b2b5651a9bed9548) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54102105)</sub>

**Context used:**

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/aicontext/PersonaContextCache.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/aicontext/PersonaContextCache.java
@@ -31,16 +31,24 @@ import org.openmetadata.schema.type.PersonaContext;
 import org.openmetadata.schema.type.PersonaContextDefinition;
 import org.openmetadata.schema.type.personaContext.PersonaContextCacheState;
 import org.openmetadata.schema.utils.JsonUtils;
+import org.openmetadata.service.Entity;
 import org.openmetadata.service.cache.CacheBundle;
 import org.openmetadata.service.cache.CacheConfig;
+import org.openmetadata.service.cache.CacheInvalidationPubSub;
 import org.openmetadata.service.cache.CacheKeys;
 import org.openmetadata.service.cache.CacheProvider;
+import org.openmetadata.service.cache.Invalidatable;
 
 /** Two-level cache and distributed single-flight coordinator for persona context documents. */
 @Slf4j
-public class PersonaContextCache {
+public class PersonaContextCache implements Invalidatable {
   public static final String CACHE_HEADER = "X-Cache";
   private static final int MAX_CACHEABLE_CHARS = 8_000_000;
+  // The local Caffeine entry is a front for the shared Redis copy, so it must not outlive the
+  // cluster's view of the document by much. Nothing invalidates when a *referenced* asset changes
+  // (a retagged table, an edited KB article), and that drift has no entity to key an invalidation
+  // off — this cap is what bounds it, and it also bounds any pub/sub message that gets dropped.
+  private static final long LOCAL_TTL_CAP_SECONDS = 60;
   private static final Duration BUILD_LEASE = Duration.ofSeconds(120);
   private static final int POLL_ATTEMPTS = 30;
   private static final long POLL_MILLIS = 500;
@@ -62,6 +70,11 @@ public class PersonaContextCache {
           },
           new ThreadPoolExecutor.AbortPolicy());
   private static volatile PersonaContextCache instance;
+  // Registered with CacheBundle instead of the instance itself: getInstance() swaps the singleton
+  // whenever the cache provider changes, and a registration holding the old object would silently
+  // stop invalidating anything. This delegate resolves the current instance on every call.
+  private static final Invalidatable INVALIDATOR =
+      (type, id, fqn) -> getInstance().invalidate(type, id, fqn);
 
   private final CacheProvider provider;
   private final CacheKeys keys;
@@ -97,6 +110,11 @@ public class PersonaContextCache {
       }
     }
     return current;
+  }
+
+  /** The {@link Invalidatable} to register with {@code CacheBundle} for this cache. */
+  public static Invalidatable invalidator() {
+    return INVALIDATOR;
   }
 
   public CachedResult get(Persona persona, boolean refresh) {
@@ -197,10 +215,28 @@ public class PersonaContextCache {
     generationStates.remove(original.getId());
   }
 
+  /**
+   * Drop this pod's local copy of a persona's context. Called for a local entity write through
+   * {@code CacheBundle.invalidateEntity} and for a remote one through the pub/sub subscriber.
+   *
+   * <p>Local only, deliberately: the pod that published already wrote the authoritative Redis
+   * copy (or deleted it). A peer deleting those keys here would throw away a document that was
+   * just rebuilt and make every pod re-run the ES-heavy build.
+   */
+  @Override
+  public void invalidate(String type, UUID id, String fqn) {
+    boolean persona =
+        Entity.PERSONA.equals(type) || CacheInvalidationPubSub.TYPE_PERSONA_CONTEXT.equals(type);
+    if (persona && id != null) {
+      invalidateLocal(id);
+    }
+  }
+
   public CachedResult refresh(Persona persona) {
     try {
       CachedResult result = get(persona, true);
       markFresh(persona);
+      publishRefresh(persona);
       return result;
     } catch (RuntimeException exception) {
       markFailed(persona, exception);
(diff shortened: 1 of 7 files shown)


pmbrull:
Thanks — went through all three. One is a real gap and is fixed in c059194; the other two I'm pushing back on, with the reasoning below.

### ✅ greptile: team updates bypass peer eviction — fixed

Correct, and it's the same bug class as the one this PR targets. A team's `defaultPersona` reaches users as `inheritedPersonas` (`UserRepository.getInheritedPersonas` resolves Team→Persona `HAS` rows for the user's teams), and membership or parent-hierarchy edits change which teams contribute one. All of those publish `type=team`, which the invalidator ignored. `TeamRepository:1348` already calls `SubjectCache.invalidateAll()` locally on hierarchy/policy change — this was missing the peer half.

Applied the suggestion, plus a test (`testRemoteTeamWriteDropsAllUserContexts`).

I kept it scoped to user contexts rather than `invalidateAll()`. `USER_POLICIES_CACHE` has its own 2-minute TTL, and peers relying on that is pre-existing behaviour I'd rather not widen inside a persona-context fix.

### ❌ gitar: peer reload may still read stale assignments — not applicable

The premise is that the reload resolves personas "through the user-keyed relationship/entity caches (`CachedRelationshipDao`/`CachedEntityDao`)". Neither caches persona assignments:

- **`CachedRelationshipDao` only caches owners, domains, and `CONTAINS` containers.** `getFromEntityRef` gates the cache on `fromEntityType == null && relationship == CONTAINS` (`EntityRepository.java:7458-7461`), with the comment right above it: *"Other relationship types (OWNS, HAS, FOLLOWS, ...) change per-write and must always hit the DB so downstream inheritance sees the freshest record."* `APPLIED_TO` and `DEFAULTS_TO` are not cached.
- **`Entity.USER` is in `UNCACHED_ENTITY_TYPES`** (`EntityRepository.java:3947`), so the Redis L2 read/write for users is skipped entirely — `isCacheableEntityType` gates both the read-through at `:10860` and the populate at `:10922`.
- **The L1 `CACHE_WITH_NAME` holds the raw DB row**, which doesn't carry relationship-derived fields. Personas are re-resolved on every load by the field fetchers, which go straight to JDBI: `fetchAndSetPersonas` → `relationshipDAO().findFromBatch(userIds, APPLIED_TO, ...)` (`UserRepository.java:1059`), `fetchAndSetDefaultPersona` → `findFromBatch(..., DEFAULTS_TO, ...)` (`:1086`), `getPersonas` → `findFrom` → `findFromRecords` → `relationshipDAO().findFrom` (`EntityRepository.java:7315-7323`).

So dropping `USER_CONTEXT_CACHE` on a peer repopulates from the database and does see the new assignments. Happy to be corrected if I've missed a path.

### ❌ gitar: O(n) scan in `invalidateUserContextByFqn` — kept deliberately, now documented

The exact-key alternative would silently reintroduce the bug. `SecurityUtil.getUserName` is `principal.getName().split("[/@]")[0]` (`SecurityUtil.java:64-67`) — no case-folding. An IdP emitting `John.Doe@corp.com` keys this cache under `John.Doe`, while the user's FQN is `quoteName(name.toLowerCase())` = `john.doe` (`UserRepository.java:182`). Keys also arrive from `createdBy`/`updatedBy` strings via `SubjectContext.getSubjectContext(...)` in the CSV import and test-case paths, which carry `user.getName()` case. Neither exact nor lower-cased lookup finds `John.Doe`.

On cost: the scan is bounded by the cache's `maximumSize`, and it only runs on user *writes* — logins and profile edits. The genuinely hot path, per-request activity tracking, updates the row through `userDAO().updateLastActivityTime` (a raw `JSON_SET`, `UserRepository.java:336`) which never reaches `invalidateCachesAfterStore` and so publishes nothing.

Added a comment at the call site so this doesn't get "optimised" into a correctness bug later.

mohityadav766 (approved):

pmbrull:
picked to 2.0

PR #31689, opened 2026-08-18
Fixes 31688: show a skeleton in the Queries tab badge while the count loads

### LINKED ISSUE #31688: 1.13: Queries tab badge shows 0 while the query count is still loading
### Describe the bug

On the Table details page the **Queries** tab badge renders a real-looking `0` for the first frames of every page load, then switches to the true count once the request resolves.

`queryCount` is initialised to `0` (`TableDetailsPageV1.tsx`) and the count request only fires after the entity resolves, while `TabsLabel` renders the badge whenever `!isNil(count)` — so the placeholder `0` is displayed as if it were the real count.

`TabsLabel` already has an `isLoading` branch for exactly this case, but nothing on this page passes it, and that branch renders a spinner rather than a skeleton.

### Scope note

This is the 1.13-only remainder of #31571. On `main`/`2.0` the count fetch was additionally deferred until the Queries tab was activated (via `useDeferredTabData`), so there the badge stayed at `0` until the user clicked the tab. That deferral was never on 1.13 — 1.13 already fetches eagerly when the entity resolves:

```tsx
useEffect(() => {
  if (tableDetails) {
    fetchQueryCount();
    getTestCaseFailureCount();
  }
}, [tableDetails?.fullyQualifiedName]);
```

So only the loading-state half of #31571 applies here, and the `main` fix cannot be cherry-picked — that page differs by ~330 lines between the branches and 1.13 has no React Query on it.

### To Reproduce

1. Open a table that has at least one query.
2. Watch the Queries tab badge on first paint: it reads `Queries 0`, then flips to the real count.

### Expected behavior

While the count is in flight the badge shows a loading placeholder, never a `0` that looks like a real count.

### Version

- OS: macOS
- OpenMetadata version: 1.13

harsh-vador:
### Describe your changes:

Fixes #31688

The **Queries** tab badge on the Table details page rendered a real-looking `0` for the first frames of every table page load, then switched to the true count. `queryCount` starts at `0`, the count request only fires once the entity resolves, and `TabsLabel` renders the badge whenever `!isNil(count)` — so the placeholder was displayed as if it were the count.

**This is the 1.13-only remainder of #31571, not a backport of #31572.** On `main`/`2.0` the count fetch was *also* deferred until the Queries tab was activated (`useDeferredTabData`), so there the badge stayed at `0` until the tab was clicked. That deferral was never on 1.13 — this branch already fetches eagerly when the entity resolves — so only the loading-state half applies here. The `main` fix could not be cherry-picked regardless: `TableDetailsPageV1.tsx` differs by ~330 lines between the branches and 1.13 has no React Query on that page, so the loading state is a `useState` flag here rather than `useQuery`'s `isFetching`.

Changes:
- `TableDetailsPageV1.tsx` — added `isQueryCountLoading`, set around `fetchQueryCount` (including the no-id early return, so tour mode doesn't leave the skeleton spinning). The flag is re-armed in the same effect that clears `tableDetails`, so the new table's first render cannot briefly show the previous table's count.
- `TableClassBase.ts` / `TableTabsUtils.tsx` — new `isQueryCountLoading` prop, forwarded to the Queries `TabsLabel` as `isLoading`.
- `TabsLabel.component.tsx` — that `isLoading` branch rendered `<Loader size="small" />`; a spinner in place of a small static pill draws more attention than the number it stands in for, so it now renders the core-components `Skeleton`. The resulting file is byte-identical to the same file on `main` after #31572, so the branches stay in sync.

Note the `isLoading` branch had no production consumer on this branch before now — only its own unit test — so no other tab's appearance changes.

https://github.com/user-attachments/assets/7a2b846b-ee2f-48f9-b28b-17644b24a578

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered
- While the count request is in flight the Queries badge shows a skeleton, never a placeholder `0`.
- Once it resolves the badge shows the true count.
- Navigating to another table shows the skeleton again rather than the previous table's count.
- A failing count request settles the badge at `0` instead of leaving it loading forever.

#### Unit tests
- `openmetadata-ui/src/main/resources/ui/src/pages/TableDetailsPageV1/TableDetailsPageV1.test.tsx` — new `Queries tab count` block. The in-flight test asserts that **every** render before the request resolves kept the badge in its loading state, which is what actually rules out the `0` flash; asserting only the end state would pass even with a flash. Plus the `0` fallback on rejection.
- `openmetadata-ui/src/main/resources/ui/src/utils/TableUtils.test.tsx` — fixture updated for the new required prop.
- `yarn test src/pages/TableDetailsPageV1 src/components/common/TabsLabel src/utils/TableUtils.test.tsx` → **165 passed, 6 suites**.
- `yarn eslint` on all six touched files → clean, zero problems. `npx tsc --noEmit` → 427 errors both before and after the change, identical per-file and per-error-code distribution, none in the touched files.

#### Playwright (UI) tests
- [x] Not applicable — covered by the unit tests above; no new UI surface was added.

#### Manual testing performed
Behaviour verified through the unit tests above, which assert the badge's loading state across every render rather than only its settled value.

#

Before: the Queries badge reads `Queries 0` on first paint, then flips to the real count. After: a skeleton pill in the badge, then the correct count.

#
### Checklist:
- [x] For UI changes: described above.

The PR replaces the Queries badge’s placeholder count with a skeleton while loading and prevents obsolete requests from settling the badge.
- Adds and propagates query-count loading state through the table tab configuration.
- Uses a monotonically increasing request sequence to reject stale responses, including A → B → A navigation.
- Adds tests for loading, failure, cross-table navigation, and revisiting a table.
- Replaces the badge spinner with a compact skeleton.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/src/pages/TableDetailsPageV1/TableDetailsPageV1.tsx | Adds query-count loading state and sequence-based stale-response rejection; the previously reported overlapping-request races are addressed. |
| openmetadata-ui/src/main/resources/ui/src/pages/TableDetailsPageV1/TableDetailsPageV1.test.tsx | Covers in-flight loading, request failure, navigation between tables, and overlapping requests when revisiting a table. |
| openmetadata-ui/src/main/resources/ui/src/components/common/TabsLabel/TabsLabel.component.tsx | Replaces the loading spinner with a compact rounded skeleton. |
| openmetadata-ui/src/main/resources/ui/src/utils/TableTabsUtils.tsx | Forwards the query-count loading state to the Queries tab label. |
| openmetadata-ui/src/main/resources/ui/src/utils/TableClassBase.ts | Extends the table tab configuration contract with the required loading-state property. |

</details>

```mermaid
sequenceDiagram
  participant U as User
  participant P as TableDetailsPageV1
  participant API as Queries API
  U->>P: Navigate to table
  P->>P: Increment request sequence
  P->>P: Show badge skeleton
  P->>API: Fetch query count with captured sequence
  alt Response belongs to latest sequence
    API-->>P: Query count
    P->>P: Display count and clear loading
  else Response is obsolete
    API-->>P: Stale response
    P->>P: Ignore response
  end
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Identify query-count requests by sequenc..."](https://github.com/open-metadata/openmetadata/commit/0bc41ba60472f67e3d4367b60b15900d8aa5206c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54311043)</sub>

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/components/common/TabsLabel/TabsLabel.component.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/common/TabsLabel/TabsLabel.component.tsx
@@ -10,11 +10,11 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
+import { Skeleton } from '@openmetadata/ui-core-components';
 import { Badge } from 'antd';
 import { isNil } from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { getCountBadge } from '../../../utils/EntityDisplayPureUtils';
-import Loader from '../Loader/Loader';
 import './tabs-label.less';
 import { TabsLabelProps } from './TabsLabel.interface';
 
@@ -35,9 +35,9 @@ const TabsLabel = ({
         {name}
         {isLoading ? (
           <span
-            className="d-flex justify-center items-center"
+            className="tw:flex tw:items-center"
             data-testid="loading-skeleton">
-            <Loader size="small" />
+            <Skeleton height={16} variant="rounded" width={24} />
           </span>
         ) : (
           !isNil(count) && (
--- a/openmetadata-ui/src/main/resources/ui/src/pages/TableDetailsPageV1/TableDetailsPageV1.test.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/pages/TableDetailsPageV1/TableDetailsPageV1.test.tsx
@@ -12,7 +12,7 @@
  */
 import { act, render, screen, waitFor } from '@testing-library/react';
 import React from 'react';
-import { MemoryRouter } from 'react-router-dom';
+import { MemoryRouter, useParams } from 'react-router-dom';
 import TabsLabel from '../../components/common/TabsLabel/TabsLabel.component';
 import { GenericTab } from '../../components/Customization/GenericTab/GenericTab';
 import PageLayoutV1 from '../../components/PageLayoutV1/PageLayoutV1';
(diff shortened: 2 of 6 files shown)


harsh-vador:
Fixed the stale-request race — good catch, this was real.

`fetchQueryCount` had no request identity, so navigating from table A to table B left A's request running and its late response applied unconditionally: it called `setQueryCount(A)` and `setIsQueryCountLoading(false)`, clearing B's skeleton and painting A's count until B's own response arrived.

There were two windows, not one:
1. A resolves after B's request has started — the response applies to B's badge.
2. A resolves during the switch itself, after `setTableDetails(undefined)` but before B's request begins — B's freshly-armed skeleton gets cleared by A.

`queryCountRequestIdRef` now records the table id whose count is in flight. A response is applied only while the ref still matches the id it was requested for, and the ref is invalidated where `tableDetails` is cleared, which closes the second window as well. The `finally` block is gone — clearing the loading flag is now part of each guarded branch, since a stale response must not clear it at all.

Note this is a 1.13-shaped fix: on `main`/`2.0` (#31572) the count goes through React Query, whose query key discards stale responses for free. This branch has no React Query on that page.

New test: `should ignore a late count response for a table the user already left` — renders table A with a pending request, switches `useParams` to table B, resolves **A** only after B's request has fired, then asserts the badge is still loading and its count is not A's. I verified it actually catches the bug by removing the guard and re-running: `✕ should ignore a late count response for a table the user already left`. With the guard: `Tests: 17 passed`.

Full run after the fix: `yarn test src/pages/TableDetailsPageV1 src/components/common/TabsLabel src/utils/TableUtils.test.tsx` → 166 passed, 6 suites. eslint clean; `npx tsc --noEmit` unchanged against the 1.13 baseline (same per-file distribution, none in the touched files).

harsh-vador:
Confirmed and fixed — the table id was the wrong request identity.

A -> B -> A while A's first request is still pending starts a second request for the same id, which set `queryCountRequestIdRef.current` back to `'a'`. A1's late response then passed the guard, cleared the skeleton, and painted an obsolete count while A2 was still in flight.

`queryCountRequestSeqRef` is now a monotonic counter, incremented before each request and at the invalidation point where `tableDetails` is cleared. Only the newest request can settle the badge, regardless of which table it belongs to — the id never enters the comparison.

New test: `should ignore a stale response for a table the user returned to`. It resolves requests by call index rather than by entity id, precisely because both A requests carry the same id and can only be distinguished by order: render A (request 1) → navigate to B (request 2) → back to A (request 3) → resolve **request 1**, assert the badge is still loading and its count is not request 1's, then resolve request 3 and assert it settles.

Verified the test catches the reported bug rather than just passing: reverting to the id-based guard gives `✕ should ignore a stale response for a table the user returned to`, while the earlier `already left` test still passes there — the two tests cover distinct windows. With the counter, `Tests: 18 passed`.

Full run: `yarn test src/pages/TableDetailsPageV1 src/components/common/TabsLabel src/utils/TableUtils.test.tsx` → 167 passed, 6 suites. eslint clean; `npx tsc --noEmit` identical to the 1.13 baseline per file, zero errors in the touched files.

chirag-madlani (approved):

PR #31696, opened 2026-08-18
test(e2e): stop persona specs asserting the system default persona fallback

harsh-vador:
## What

Backports the persona-specific parts of #31063, #30699, #30203 and #30732 to **1.13**, where they were never cherry-picked. `main` and `2.0` already carry all four, which is why this failure is 1.13-only.

## RCA

`Team persona setting flow > Set default persona for team should work properly` fails at:

```
expect(locator).toContainText(expected) failed
Locator:  getByTestId('default-persona-chip')
Expected substring: "No default persona"
Received string:    "PW Persona ac9eb…"
```

Not a product bug — the assertion depends on global state the test does not own.

1. The chip renders `user.defaultPersona` (`UserProfilePersona.component.tsx`).
2. When a user has no default persona, the backend resolves the field to the **system** default persona (`UserRepository.getDefaultPersona`).
3. So `"No default persona"` only ever renders when *nothing else* has a system default set.
4. The sibling describe in this same file (`Default persona setting and removal flow`) sets and then clears a system default, and the AUT ships with a pre-seeded `Onboarding System Default Experience` persona. The received value in the trace — `PW Persona ac9eb…` — is that sibling's own fixture, **not** the team persona.

So the assertion was passing on another test's side effect. In 1.13 that leak is worse than upstream: the sibling describe shares a module-level `UserClass` with `Persona operations` and has **no teardown at all**, so its personas and the system default it set survive the run.

Two secondary defects in the shared helpers, same class of problem:

- `persona.ts` waited on `page.waitForResponse('/api/v1/personas/*')` — an unfiltered matcher that any in-flight **GET** can satisfy, so the helper could return before the PATCH landed.
- It also awaited `toBeVisible` on `default-persona-confirmation-modal`, a testid that sits on the antd `.ant-modal-root` wrapper (0x0) and therefore always reports hidden.

`Users.spec.ts` carried the same system-default-fallback bug in the persona-removal test: it asserted no `default-persona-tag` and zero checked radios across **every** `persona-label` in the dropdown, which the system default persona trips.

`#30278` (`fix(ui): stop showing inherited team persona as user default persona`) is **already in 1.13** as `bba09e4424e` — `UserProfilePersona.component.tsx` is byte-identical to `main`. Not the cause here.

## Changes

- **`PersonaFlow.spec.ts`** — team persona step now asserts the real invariant (the team's persona is **not** auto-applied as the user default) instead of the placeholder text, preceded by a visibility check so the negative assertion cannot pass against a chip that never rendered. *(#31063)*
- **`PersonaFlow.spec.ts`** — `Default persona setting and removal flow` gets a describe-local `UserClass`, an `afterAll` deleting the user and both personas, `userContext.close()` in a `finally`, and the `redirectToHomePage(userPage)` after the first verification step. Stops this describe leaking a system default persona into sibling specs. *(#30180, #30203; upstream shipped the teardown inside the #29964 HeaderShell refactor, so applied directly rather than cherry-picked)*
- **`Users.spec.ts`** — persona-removal assertions scoped to the user's own `persona1`/`persona2` labels. *(#30732)*
- **`utils/persona.ts`** — `setPersonaAsDefault` / `removePersonaDefault` wait on the modal's **Yes** button and scope the response wait to `PATCH /api/v1/personas/{id}`, registered immediately before the click. *(#30699)*

## Deliberately not backported

The `default-persona-select-list` server-side search step present in `main` — it requires `/api/v1/personas/search` from #30469, a feature PR absent from 1.13. Porting the test alone would 404.

## Verification

- Prettier: clean on all three files
- ESLint: clean on all three files
- `tsc --noEmit -p playwright/tsconfig.json`: error set for these three files unchanged against the pre-change baseline (one pre-existing unused `getApiContext` in `Users.spec.ts`)
- Specs not run locally — needs a live AUT; only a nightly can confirm the flake is gone

The PR makes persona E2E coverage independent of the system-default persona fallback and improves test isolation and response synchronization.
- Scopes persona-removal checks to the two user-owned fixture personas and now verifies each matching label exists.
- Gives the default-persona flow isolated fixtures and teardown.
- Waits specifically for persona PATCH responses and actionable confirmation buttons.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Flow/PersonaFlow.spec.ts | Isolates default-persona fixtures, adds cleanup, and replaces a global-state-dependent team-persona assertion with the intended invariant. |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Users.spec.ts | Completes the prior fix by requiring each fixture persona label to be visible before checking that it is neither selected nor marked as default. |
| openmetadata-ui/src/main/resources/ui/playwright/utils/persona.ts | Synchronizes default-persona actions with visible confirmation controls and method-specific PATCH responses. |

</details>

<sub>Reviews (2): Last reviewed commit: ["test(e2e): require the persona label to ..."](https://github.com/open-metadata/openmetadata/commit/8bc6fd3cd1877a1a5913ead3fa5e0c9623a2ef01) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54324840)</sub>

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Flow/PersonaFlow.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Flow/PersonaFlow.spec.ts
@@ -48,7 +48,6 @@ const PERSONA_DETAILS = {
   description: `Persona description ${uuid()}.`,
 };
 
-const user = new UserClass();
 const persona1 = new PersonaClass();
 const persona2 = new PersonaClass();
 
@@ -303,16 +302,31 @@ test.describe.serial('Persona operations', () => {
 });
 
 test.describe.serial('Default persona setting and removal flow', () => {
+  const user1 = new UserClass();
+
   test.beforeAll('Setup user for default persona flow', async ({ browser }) => {
     const { apiContext, afterAction } = await createNewPage(browser);
-    await user.create(apiContext);
+    await user1.create(apiContext);
     const adminResponse = await apiContext.get('/api/v1/users/name/admin');
     const adminData = await adminResponse.json();
 
-    await persona1.create(apiContext, [user.responseData.id, adminData.id]);
-    await persona2.create(apiContext, [user.responseData.id]);
+    await persona1.create(apiContext, [user1.responseData.id, adminData.id]);
+    await persona2.create(apiContext, [user1.responseData.id]);
     await afterAction();
   });
+
+  test.afterAll(
+    'Teardown user for default persona flow',
+    async ({ browser }) => {
+      const { apiContext, afterAction } = await createNewPage(browser);
+      await Promise.all([
+        user1.delete(apiContext),
+        persona1.delete(apiContext),
+        persona2.delete(apiContext),
+      ]);
+      await afterAction();
+    }
+  );
 
   test('Set and remove default persona should work properly', async ({
     adminPage,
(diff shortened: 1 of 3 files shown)


harsh-vador:
Good catch — fixed in 8bc6fd3cd18.

`filter({ hasText })` is allowed to match zero elements, so both `not.toBeVisible()` and `not.toBeChecked()` pass vacuously when the fixture persona is missing from the dropdown: the test would go green without ever checking the persona whose default was removed.

Added `await expect(userPersonaLabel).toBeVisible();` before the two negative checks, which is the same guard this PR already adds to the default persona chip in `PersonaFlow.spec.ts`. Kept `not.toBeChecked()` rather than the suggested diff's shape, since dropping it would lose the 'no persona is selected' half of the assertion.

ShaileshParmar11 (approved):

PR #29396, opened 2026-06-24
ci: run downstream Collate build on UI-only OSS PRs

harsh-vador:
## What

Add `openmetadata-ui/**` to the `pull_request_target` paths in `maven-build-collate.yml`, so OSS PRs that change only UI files trigger the existing downstream Collate build before merge.

## Why

The downstream "OpenMetadata Collate Test" check (which builds `collate-ui` against the OSS PR via `mvn install -DskipTests`) was wired to `pull_request_target` paths covering only:

- `openmetadata-service/**`
- `openmetadata-spec/.../schema/**`
- `openmetadata-integration-tests/**`
- `openmetadata-mcp/**`

`openmetadata-ui/**` was missing. So an OSS PR that deletes/renames a UI file or export used **only** by Collate (e.g. `searchTagInData` in `TableTags.utils`, or the `EntityNameUtils` file) passed all OSS checks, merged, and then broke Collate's build — discovered late on unrelated Collate PRs, nightly Docker, or release backports.

The post-merge `push` trigger already includes `openmetadata-ui/**`, but that runs after merge — too late. This change moves the same coverage to pre-merge.

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### High-level design:

N/A — small change. 

#
#### Manual testing performed

#
### UI screen recording / screenshots:

Not applicable. 

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] My PR is linked to a GitHub issue via `Fixes #<issue-number>` above.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.
- [ ] For UI changes: I attached a screen recording and/or screenshots above.
- [ ] I have added tests (unit / integration / Playwright as applicable) and listed them above.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR extends the `pull_request_target` path filter in `maven-build-collate.yml` to include UI source files, ensuring that OSS PRs touching only `openmetadata-ui/src/**` trigger the downstream Collate build before merge — closing the gap where UI-breaking changes could pass all OSS checks and only surface after merge on the Collate side.

[...]

PR #31697, opened 2026-08-18
ci: build the Collate branch matching the PR base, not always main

harsh-vador:
## What

Collate CI validated **every** OpenMetadata PR against Collate `main`, whatever branch the PR was actually for. Build the Collate branch matching the PR's base branch instead.

## The bug, plainly

Collate keeps one branch per OpenMetadata release line, with the same name on both sides — `main`, `1.13`, `2.0` — and each Collate branch pins the matching OpenMetadata branch in `.gitmodules`. The two halves are a matched pair: Collate `1.13` is meant to be built against OpenMetadata `1.13`.

The dispatch step ignored that and always asked for `main`:

```yaml
        with:
          workflow: OpenMetadata Collate Test
          ref: main          # <- always main, whatever the PR base is
```

So a PR against `1.13` built **Collate's newest code against a 1.13 OpenMetadata checkout** — a mismatched pair. It passes only for as long as the two lines happen not to have diverged, and breaks the moment they have, on whatever happens to have diverged that day.

#31689 (base `1.13`) is that failure. Its Collate run reports `headBranch: main`, and the build died in `collate-ui`:

```
[vite]: Rollup failed to resolve import "openmetadata-ui/src/constants/appMode.constants"
        from ".../collate-ui/src/main/resources/ui/src/App.tsx"
```

- Collate `main` → `App.tsx:20` imports `AI_APP_MODE` from `openmetadata-ui/src/constants/appMode.constants`
- Collate `1.13` → no `appMode` reference at all
- OpenMetadata `1.13` → that constants file does not exist

`collate-spec`, `collate-service`, `collate-mcp` and the integration-test module all compiled; only `collate-ui` failed, exactly where the lines diverge. The reactor header also read `2.0.0-SNAPSHOT` — the main line — while validating a 1.13 PR.

Nothing about this is specific to that import. `2.0` PRs are mis-routed identically today; they simply have not tripped over a divergence yet.

Runs:
- PR: https://github.com/open-metadata/OpenMetadata/pull/31689
- OpenMetadata job: https://github.com/open-metadata/OpenMetadata/actions/runs/32128130179/job/95683140608
- Collate build: https://github.com/open-metadata/openmetadata-collate/actions/runs/32128198485

## Change

One line — the dispatch target is read from the event payload:

```yaml
ref: ${{ github.event.pull_request.base.ref || github.ref_name || 'main' }}
```

- a pull request in the payload → its base branch
- otherwise (`push`, `workflow_dispatch`) → the current ref
- neither set → `main`

Branch names already match across the two repos, so this is an identity mapping — no lookup table, no API call. Reading the payload directly rather than testing `github.event_name` keeps it correct whichever pull-request trigger fires the workflow.

If Collate has no branch of the resolved name — a PR stacked on a feature branch, or the window right after a release branch is cut — the dispatch fails with `workflow not found on ref`. That is deliberate: a loud failure rather than the silent fallback to `main` this PR exists to remove. An earlier revision probed the Collate branches API to fall back in that case; dropped per review.

The trigger stays on `pull_request_target`. An earlier revision moved it to `pull_request`; reverted per review, since that would also make the workflow file per-branch and require cherry-picking it to every release line.

## Why this lands on `main`

`pull_request_target` takes the workflow file from the **default branch**, not from the PR's base branch. One commit on `main` therefore fixes `main`, `1.13` and `2.0` at once, and a commit on `1.13` would have no effect at all.

That is measured, not assumed. A probe PR based on this branch still dispatched Collate `main` (`head_branch=main`) — main's hardcoded `ref` ran, not the base branch's fixed one. Independently, #31792 (`base=1.13`) runs the step `Restore openmetadata-ui dist cache`, which exists only in main's `maven-sonar-build.yml`; 1.13's copy has no such step.

An earlier revision of this PR targeted `1.13` on that wrong assumption and has been retargeted.

## Verification

- `yaml.safe_load` parses the workflow; triggers read back unchanged (`workflow_dispatch`, `push`, `pull_request_target`), the step list is unchanged (`Wait for the labeler → Verify PR labels → Checkout → Trigger Collate build & wait`), and the dispatch `ref:` reads back as the expression above
- Collate carries `main`, `1.13` and `2.0`, and each has `openmetadata-dispatch.yml` named `OpenMetadata Collate Test` with the same two inputs (`sha`, `event`) — so every release line resolves to a real branch with a matching dispatch signature
- the resolution logic was exercised against the live Collate API for each release line and for absent branches while an API-probing revision was on this branch: `main` → `main`, `1.13` → `1.13`, `2.0` → `2.0`, `antd-migration/wave-1` → absent, `does-not-exist-xyz` → absent
- end-to-end dispatch behaviour can only be observed once this is merged and a PR runs it

## Not included

`on.push.branches` is still `main` only, so merges to `1.13` / `2.0` never trigger a Collate build and regressions on those lines land silently. Adding them is a one-line change but switches on Slack alerting for those branches, so it is left as a separate decision.

## Found while testing, not fixed here

1. Every PR based on a non-`main` release line fails `Maven SonarCloud CI` at `Restore openmetadata-ui dist cache` — main's `maven-sonar-build.yml` uses `./.github/actions/cache-ui-dist`, which exists only on `main` (absent on `2.0`, `1.13`, `1.12.11`). Same shape as this bug, different workflow. Pre-existing: see #31792.
2. In this same file, `timeout-minutes: 60` on the dispatch step is below its own `wait-timeout-seconds: 3660`, so the step is cancelled a minute before the wait can time out. Pre-existing on `main` and untouched here.

Both belong to CI/infra and are worth their own issues.

The PR updates Collate workflow dispatches to select the branch matching a pull request’s base ref, while retaining the existing `pull_request_target` trigger.
- Replaces the hard-coded Collate `main` ref with the PR base branch.
- Uses the current ref for push and manual-dispatch events.
- Preserves a final `main` fallback when neither event value is available.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| .github/workflows/maven-build-collate.yml | The dispatch now targets the matching Collate branch; the two previously reported failure modes are no longer present at current HEAD. |

</details>

<sub>Reviews (7): Last reviewed commit: ["ci: build the Collate branch matching th..."](https://github.com/open-metadata/openmetadata/commit/ce4049ff5a617425e49d93d71851a7f47c2e8f05) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54328251)</sub>

### DIFF
--- a/.github/workflows/maven-build-collate.yml
+++ b/.github/workflows/maven-build-collate.yml
@@ -87,7 +87,12 @@ jobs:
           SHA: ${{ github.event_name == 'push' && github.event.after || github.event.pull_request.head.sha || github.sha }}
         with:
           workflow: OpenMetadata Collate Test
-          ref: main
+          # Collate mirrors each OpenMetadata release line on an identically
+          # named branch, so build the Collate branch matching the PR base.
+          # Dispatching main for a 1.13 PR builds main-only Collate code
+          # against a 1.13 checkout. Falls through to the pushed/dispatched ref
+          # when there is no pull request in the payload.
+          ref: ${{ github.event.pull_request.base.ref || github.ref_name || 'main' }}
           repo: open-metadata/openmetadata-collate
           token: ${{ secrets.COLLATE_PAT }}
           wait-for-completion: true


harsh-vador:
Addressed in c6d765cb896.

The fallback now branches on the HTTP status rather than on `gh api`'s exit code:

- `200` → build the matching Collate branch
- `404` → warn and fall back to `main` (the only case where the branch genuinely does not exist)
- anything else (401/403 expired PAT, rate limit, 5xx) → `::error::` and fail the job

`curl --retry 3 --retry-connrefused --retry-all-errors --max-time 30` absorbs the transient cases before that decision, and a connection failure now aborts the step under `set -e` instead of being read as a missing branch — so a network blip can no longer silently reintroduce the mis-routing this PR fixes.

Verified by stubbing the probe and driving each status: 200 → `ref=1.13`, 404 → warning + `ref=main`, 401/403/503 → exit 1 with no output set, connection failure → exit 7 with no output set.

chirag-madlani (dismissed):

harsh-vador:
Checked both findings.

### 1. 🚨 `benc-uk/workflow-dispatch` drops wait gating — not correct

The action does declare those inputs. From its `action.yaml` at the pinned tag `v1.3.2`:

```
inputs:  workflow, token, inputs, ref, repo,
         wait-for-completion, wait-timeout-seconds, wait-interval-seconds, sync-status
outputs: runId, runUrl, runUrlHtml, workflowId
```

All four wait keys used here are declared under exactly these names, so nothing is silently ignored and the job still blocks on the Collate result.

Worth noting the suggested alternative names — `wait-for-completion-timeout`, `wait-for-completion-interval` — are not inputs of this action, so applying that change would break the wait it is meant to protect.

Second point: **this PR does not change the dispatch action.** `main` has been on `benc-uk/workflow-dispatch@v1.3.2` since before this branch. The diff here is two things only — the new `Resolve the Collate branch to build` step, and `ref: main` → `ref: ${{ steps.collate-branch.outputs.ref }}`. `the-actions-org/workflow-dispatch@v4` is what `1.13` uses, so this reads like a diff taken against `1.13`; the PR was retargeted from `1.13` to `main` earlier today.

### 2. ⚠️ `timeout-minutes: 60` vs `wait-timeout-seconds: 3660` — valid

Real inconsistency: the step is cancelled at 3600s, 60s before the wait can time out, so a nearly-finished Collate build dies with a misleading step-timeout rather than the intended wait-timeout.

Both lines are pre-existing on `main` and untouched by this PR, so I would rather not widen a CI-only change that has already been through a few rounds — filing it separately with the other `maven-*` workflow issue found while testing this (`Restore openmetadata-ui dist cache` failing on every non-`main` release line, since `.github/actions/cache-ui-dist` exists only on `main`).

Happy to fold the `timeout-minutes` bump into this PR instead if reviewers prefer it here.

.github/workflows/maven-build-collate.yml
```diff
@@ -87,7 +87,11 @@ jobs:
           SHA: ${{ github.event_name == 'push' && github.event.after || github.event.pull_request.head.sha || github.sha }}
         with:
           workflow: OpenMetadata Collate Test
-          ref: main
+          # Collate mirrors each OpenMetadata release line on an identically
+          # named branch, so build the Collate branch matching the PR base.
+          # Dispatching main for a 1.13 PR builds main-only Collate code
+          # against a 1.13 checkout.
+          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.base.ref || github.ref_name || 'main' }}
```
chirag-madlani: I would say let's use pull_request instead of pull_request_target since most of the time we are raising PR agains release branch needs this not from the community PRs

harsh-vador:
Done — switched to `pull_request` in 9a22312.

Three things had to move with it, flagging them so they get reviewed rather than assumed:

**1. The `ref:` condition.** It keys off the event name, so on `pull_request` the old `event_name == 'pull_request_target'` test would have been false and the expression would have fallen through to `github.ref_name` — which for a PR event is the merge ref (`31697/merge`), not a branch. Every PR would have asked Collate for `31697/merge` and failed. Condition updated in lockstep.

**2. Fork PRs.** They receive no secrets on `pull_request`, so `COLLATE_PAT` would be empty and the dispatch would fail on authentication rather than skip. The job condition now requires `github.event.pull_request.head.repo.full_name == github.repository`, so community PRs skip the Collate build instead of failing it — which matches your point that this is only needed for PRs against release branches.

**3. `allow-unsafe-pr-checkout: true` is gone.** It only existed to opt into the `pull_request_target` risk of checking out untrusted code in a privileged job. Under `pull_request` it is unnecessary, and dropping it removes that surface entirely — a real win from your suggestion.

One consequence worth being explicit about: `pull_request` runs the workflow from the PR's merge commit rather than from `main`, so from now on this file is per-branch — a change to it needs cherry-picking to `1.13` and `2.0` to take effect there. On `pull_request_target` one commit on `main` covered every line. Happy to raise the 1.13 / 2.0 backports once this lands if you want them.

harsh-vador:
Same finding as [above](https://github.com/open-metadata/OpenMetadata/pull/31697#issuecomment-5354989789) — re-checked against the action source, still does not hold.

`benc-uk/workflow-dispatch@v1.3.2` declares all four wait keys. From its `action.yaml` at that exact tag:

```
inputs: workflow, token, inputs, ref, repo,
        wait-for-completion, wait-timeout-seconds, wait-interval-seconds, sync-status
```

So nothing is silently ignored and the job still blocks on the Collate result. The suggested replacement names — `wait-for-completion-timeout`, `wait-for-completion-interval` — are not inputs of this action, so that change would break the gating it is trying to protect.

This pass is also against an older revision of the branch. Its proposed fix reads `ref: ${{ steps.collate-branch.outputs.ref }}`, but that step was removed per earlier review feedback: current head (`9a22312`) has no `collate-branch` step at all, just the single `ref:` expression. Worth re-running against the latest commit if this is automated.

And this PR does not change the dispatch action — `main` has been on `benc-uk/workflow-dispatch` since #29396 / #30296, well before this branch. `the-actions-org/workflow-dispatch@v4` is what `1.13` uses, which is what a diff taken against `1.13` rather than `main` would show.

The timeout observation from the earlier pass (`timeout-minutes: 60` below `wait-timeout-seconds: 3660`) is real, and pre-existing on `main` — noted at the bottom of the description to be filed separately rather than widened into this PR.

.github/workflows/maven-build-collate.yml
```diff
@@ -32,7 +32,7 @@ on:
       - "yarn.lock"
       - "Makefile"
       - "bootstrap/**"
-  pull_request_target:
+  pull_request:
```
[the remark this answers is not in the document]
harsh-vador: replied

.github/workflows/maven-build-collate.yml
harsh-vador: fixed

chirag-madlani (approved):

PR #31710, opened 2026-08-18
Fixes 31708: forward the sample-data validation timeout across docker exec

### LINKED ISSUE #31708: RDF E2E lane fails ~29% of runs: docker exec drops the sample-data validation timeout
### Describe the bug

`PostgreSQL PR RDF E2E Tests` fails roughly 29% of the time (10 of the last 35 runs) before it ever reaches Playwright. It dies in `Setup Openmetadata Test Environment` with:

```
Exception: Max retries exceeded. Sample data ingestion was not successful.
⚠ Warning: DAG validation failed with exit code 1
✗ Startup requires sample data ingestion to complete before continuing.
```

Example: https://github.com/open-metadata/OpenMetadata/actions/runs/32124714651/job/95676579308 (same signature on runs 32124712892, 32123379162).

### Root cause

`docker/run_local_docker_common.sh` runs the validator as:

```bash
docker exec -i openmetadata_ingestion timeout ... python - < docker/validate_compose.py
```

`docker exec` starts the process in the **container's** environment, so the `VALIDATE_COMPOSE_*` values exported by `docker/run_local_docker_rdf.sh` (`VALIDATE_COMPOSE_MAX_RETRIES=60`, i.e. an intended 600s window) never cross the boundary. `validate_compose.py` falls back to its module defaults of 15 retries x 10s = **150s**, and the outer `timeout $VALIDATION_TIMEOUT_SECONDS` (900s for RDF) is dead code that can never fire.

Sample-data ingestion with RDF enabled takes ~150s, because every entity write also syncs to Fuseki. That makes the outcome a coin flip: the failing run above died at 150.3s, and the run that passed immediately before it took 153s.

### Why it produces no diagnostics

`validate_compose.py` reports progress through `log_ansi_encoded_string`, which logs at INFO. `ingestion/src/metadata/utils/logger.py` calls `logging.basicConfig(format=...)` with no `level=`, so the root logger stays at WARNING and every message is dropped. Confirmed on both the failing *and* the passing runs — neither printed a single progress line, so a "needed three more seconds" timeout is indistinguishable from a hung DAG.

### Impact

- ~29% of RDF E2E runs fail for no code-related reason, blocking merges and requiring manual re-runs.
- The `nick-fields/retry` wrapper rebuilds the whole Docker stack (~6 min) and hits the identical 150s wall, so each failure burns ~11 minutes of runner time.
- Failures arrive with zero diagnostics, so they get triaged as generic flakiness.

### Expected behaviour

The validation window configured by the startup script should actually apply to the validator, and a timeout should report the DAG run state and per-task states so a slow run can be told apart from a stuck one.

ulixius9:
### Describe your changes:

Fixes #31708

The `PostgreSQL PR RDF E2E Tests` lane failed **~29% of runs (10/35)** in `Setup Openmetadata Test Environment`, always with `Max retries exceeded. Sample data ingestion was not successful.` It was diagnosed as flakiness, but it is a config bug: `run_local_docker_common.sh` ran the validator through a bare `docker exec`, which starts the process in the **container's** environment, so the `VALIDATE_COMPOSE_*` values exported by `run_local_docker_rdf.sh` never crossed the boundary. `validate_compose.py` fell back to its module defaults of 15 x 10s = **150s**, and the outer `timeout 900` was dead code that could never fire. Sample-data ingestion takes ~150s with RDF on because every entity write also syncs to Fuseki — so it was a coin flip: [the failing run](https://github.com/open-metadata/OpenMetadata/actions/runs/32124714651/job/95676579308) died at 150.3s, and the run that passed immediately before it took 153s.

I also made the failure legible. Every message went through `log_ansi_encoded_string` at INFO, but `metadata/utils/logger.py` calls `basicConfig()` with no `level=`, leaving the root logger at WARNING — so *passing* runs printed nothing either, and "needed three more seconds" looked identical to a hung DAG.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

Three files, all under `docker/` (no workflow files touched).

**1. `run_local_docker_common.sh`** — forward the knobs with explicit `-e` flags, and derive the validator's deadline from `VALIDATION_TIMEOUT_SECONDS` minus a margin so it exits with its own diagnostics instead of being SIGTERMed mid-report. Added `python -u`.

**2. `validate_compose.py`** — one wall-clock deadline replaces a confusing nested retry loop (an outer 15-retry loop wrapping an inner 30-retry loop); `print(flush=True)` instead of the swallowed logger; per-task-instance state dumped on failure; the error names the last observed run id and state. Every HTTP call is bounded by the time actually left, and the post-deadline diagnostic pass has its own hard budget (20s total, 8s per request) so it fits inside the margin by construction. Dropping the `metadata` import also removes a ~6s import and the `PostgresStoredProcedure ... shadows an attribute` pydantic warning.

**3. `run_local_docker_rdf.sh`** — removed the `VALIDATE_COMPOSE_MAX_RETRIES` export and the two knobs whose loop no longer exists.

Effective windows: **RDF 150s → 840s** (900s outer, 60s margin); default lane 240s; the fixture-prep lane in `playwright-postgresql-e2e.yml` 540s. `VALIDATION_TIMEOUT_SECONDS` is the only knob the scripts set — `VALIDATE_COMPOSE_MAX_RETRIES` remains supported as an explicit operator override, with a comment warning against reintroducing it in-tree (it would silently pin the deadline again).

Also fixed a crash found while in here: `get_access_token()` raised a bare `Exception` that `get_last_run_info()` did not catch, so an Airflow API blip mid-poll aborted the entire wait — and Airflow 500s for ~40s on every CI startup. Reproduced against the pre-fix file: it dies with a traceback where the new code retries and succeeds.

#### Review feedback addressed

- **[P1, @ayush-shah / greptile]** `airflow_get()` used a flat `REQUESTS_TIMEOUT=300` while the shell reserved only a 30s margin, so one stalled request could carry the process past the outer `timeout` and kill the diagnostics this PR exists to add. Confirmed by reproduction, not just inspection: against a stalling stub the old code printed its timeout header and was then SIGTERMed with no task-state output. Now every request is bounded by the remaining deadline (clamped to 30s, floored at 5s), diagnostics run on a 20s/8s budget, and the margin moved to 60s. Worst-case overshoot past the deadline is 25s.
- **[gitar-bot]** RDF really was getting 600s, not the 870s the config implied, because the `MAX_RETRIES` export took precedence — an earlier revision of this description said 870s, which was wrong. Fixed by dropping the export; raising `VALIDATION_TIMEOUT_SECONDS` now has an effect.
- **[gitar-bot]** A transient poll failure logged "Waiting for DAG run to start". `get_last_run_info()` now reports whether the poll succeeded, so an unreachable API says so instead.

#
#### Use cases covered
- RDF E2E startup completes when sample-data ingestion takes longer than 150s (the failing case).
- A genuinely failed `sample_data` DAG still fails fast (one poll) rather than waiting out the window.
- A timeout reports the DAG run id, its state, and per-task states — and still does so when the diagnostic endpoints themselves stall.
- A transient Airflow token/API failure mid-poll no longer aborts the wait, and is reported as unreachable rather than as "no run yet".

#### Unit tests
No committed test file. `docker/` sits outside every test tree — pytest runs from `ingestion/tests`, and ruff's pre-commit scope is `^(ingestion|openmetadata-airflow-apis)/` — so a test added here would not be picked up by CI. **This is a real gap; happy to wire a harness in under `ingestion/tests/unit/` if reviewers prefer that over leaving it uncovered.**

Verified instead with a stub Airflow 3.x API driving the real module, 9/9 scenarios passing: happy path; token 500s then success; `state=failed` fast-fail; timeout-reports-last-state; timeout-dumps-task-states; `MAX_RETRIES` override; no-runs vs unreachable distinguished; `last_dag_logs` 404 tolerated. Plus two stall scenarios (diagnostic endpoint hangs, poll endpoint hangs) that fail on the parent commit and pass here.

- [x] Not applicable — no `ingestion/src` changes; this only touches the Docker startup scripts.

#### Playwright (UI) tests

#### Manual testing performed
1. Proved the root cause against a live `openmetadata_ingestion` container: bare `docker exec` reports `MAX_RETRIES = None` (150s budget); with the `-e` forwarding it reports the forwarded deadline.
2. Ran the validator inside that real container against real Airflow — real token auth, real `/api/v2` calls, correct progress and timeout output. Also confirmed `last_dag_logs` genuinely 404s in this Airflow 3 setup, which is why the old code produced no output even on that path.
3. `bash -n` clean on both shell scripts; verified the margin table (300→240, 900→840, 600→540) and that the `-e` array omits `MAX_RETRIES` when unset.
4. `RDF Playwright execution` passed on this branch (25m4s).

#

Not applicable.

#
### Checklist:
- [ ] I have added tests — see the Unit tests note above; verified with a local stub-Airflow harness rather than a committed test, because `docker/` is outside CI's test scope.

The PR forwards sample-data validation settings into the ingestion container and makes validation target the DAG run created during startup.
- Derives the validator deadline from the outer shell timeout while reserving a fixed diagnostic margin.
- Bounds Airflow polling and post-timeout diagnostics with shared wall-clock budgets.
- Improves failure output with run identity, state, task-instance details, and resilient API handling.
- Adds regression coverage for environment forwarding, run targeting, authentication refresh, timeout handling, and diagnostics.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| docker/run_local_docker_common.sh | Forwards validator settings and the triggered run identity across docker exec while reserving and enforcing a diagnostic margin. |
| docker/run_local_docker_rdf.sh | Uses the outer validation timeout as the RDF lane’s sole duration control and removes conflicting retry exports. |
| docker/validate_compose.py | Replaces nested retries with deadline-based targeted polling and bounded, resilient failure diagnostics. |
| ingestion/tests/unit/test_validate_compose.py | Adds regression tests for timeout capping, target selection, transient failures, pagination, and shared HTTP budgets. |

</details>

```mermaid
sequenceDiagram
    participant S as Startup script
    participant C as Ingestion container
    participant V as Validator
    participant A as Airflow API
    S->>A: Trigger sample_data DAG
    A-->>S: DAG run ID and logical date
    S->>C: docker exec with validation environment
    C->>V: Start validator with inner deadline
    loop Until terminal state or deadline
        V->>A: Poll targeted DAG run
        A-->>V: Run state
    end
    alt DAG succeeds
        V-->>S: Validation succeeds
    else DAG fails or deadline expires
        V->>A: Fetch bounded diagnostics
        A-->>V: Run and task states
        V-->>S: Validation fails with diagnostics
    end
```
</details>

<sub>Reviews (11): Last reviewed commit: ["fix(ci): format sample-data validation t..."](https://github.com/open-metadata/openmetadata/commit/bb5fe72e3b723496ee9551fc17cfdd4612057fb7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54377582)</sub>

**Context used:**

- Knowledge Base — [Deployment and operations](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/deployment-operations.md)
- Knowledge Base — [Containerized deployment](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/containerized-deployment.md)


chirag-madlani (dismissed):

ayush-shah:
[P1] Bound Airflow requests to the remaining validation deadline so diagnostics cannot be killed by the outer timeout

`airflow_get()` uses `REQUESTS_TIMEOUT=300` ([`validate_compose.py:112`](https://github.com/open-metadata/OpenMetadata/blob/7af9df1bbe7fb8dbab533c903bf1cf745a29a2f6/docker/validate_compose.py#L112)), while `run_local_docker_common.sh` gives the validator only a 30-second margin below the outer timeout. With the default 300-second validation window, the inner deadline is 270 seconds; one stalled GET can therefore consume the entire outer timeout. Near the RDF deadline, the task-state and log requests in `dump_diagnostics()` can likewise be killed before they complete.

That means the timeout path can still fail without the DAG state/per-task diagnostics this PR is intended to provide. Please derive each request timeout from the remaining deadline and reserve a bounded diagnostic budget, or otherwise ensure the outer timeout cannot terminate diagnostics.

ulixius9:
@ayush-shah confirmed and fixed in 4f1ce104bd — thanks, this was a real hole in the change and it landed on exactly the invariant the PR claimed.

I reproduced it rather than just reading it, against a stub Airflow that accepts the connection and never responds. On the parent commit, with the inner deadline at 6s and an outer timeout mirroring the 30s margin shape:

```
Timed out after 6s waiting for the sample_data DAG.
   [stub] taskInstances request received -> stalling forever
→ exit=143 (SIGTERM)
```

The timeout header prints and the process is then killed — no `Task instances for ...`, no final error naming the last observed state. A stalled poll mid-loop does the same thing. So the 30s margin was only ever sufficient by luck, and in the default lane a single 300s request exceeded the entire 270s window.

What changed:

- **Poll requests** are bounded by the time actually left: `max(5, min(30, remaining))`. Visible in the repro — `read timeout=5.98` when ~6s remained, rather than a flat 300s.
- **`dump_diagnostics()`** now has its own budget, since it runs after the deadline inside your margin: 20s total, 8s per request, skipping a step rather than overrunning it. With *both* diagnostic endpoints hanging it now exits cleanly at 23s with the final error intact instead of being SIGTERMed.
- **The margin** goes 30s → 60s for windows over 180s. Worst-case overshoot is bounded at 25s (one 5s in-flight poll + the 20s diagnostic budget), so it fits by construction rather than by luck.

I also took the two @gitar-bot points while here. The precedence one was correct and had made my own PR description wrong: `run_local_docker_rdf.sh` exported `VALIDATE_COMPOSE_MAX_RETRIES=60`, which wins in `resolve_timeout_seconds()`, so the RDF lane was getting 600s and the derived deadline was dead code — raising `VALIDATION_TIMEOUT_SECONDS` did nothing. Dropped that export, so RDF is now 900s outer / 840s inner and the description is corrected. `MAX_RETRIES` stays as an explicit operator override with a comment warning against reintroducing it in-tree. And `get_last_run_info()` now reports whether the poll succeeded, so a transient failure says "Could not reach the Airflow API" instead of "Waiting for DAG run to start".

Full suite is 9/9 plus the two stall scenarios, which fail on the parent commit and pass here. One thing I have not solved: `docker/` is outside every test tree (pytest runs from `ingestion/tests`, ruff's scope is `^(ingestion|openmetadata-airflow-apis)/`), so none of this is guarded in CI. If you would rather it were, say so and I will wire the stub harness in under `ingestion/tests/unit/`.

chirag-madlani (dismissed):

docker/validate_compose.py
```diff
@@ -25,164 +61,244 @@ def get_env_int(name: str, default: int) -> int:
...
+    timeout_seconds = resolve_timeout_seconds(poll_interval_seconds)
+    _deadline = time.monotonic() + timeout_seconds
+
+    log(f"Waiting up to {timeout_seconds}s for the {DAG_ID} DAG (polling every {poll_interval_seconds}s).")
+
+    dag_run_id: str | None = None
+    state: str | None = None
+
+    while True:
+        dag_run_id, state, polled = get_last_run_info()
```
ayush-shah: [P1] Preserve the last successfully observed DAG run across a transient poll failure. `get_last_run_info()` returns `(None, None, False)` for an auth/transport blip, and this assignment erases a previously observed `run_id`/state. If the deadline then expires, `dump_diagnostics(None)` skips `print_task_instance_states()` and the error reports `last observed run=None, state=None`—exactly the diagnostic gap this PR is meant to close. Keep the last successful ID/state separately and use it for the timeout diagnostic path.

docker/validate_compose.py
```diff
@@ -25,164 +61,244 @@ def get_env_int(name: str, default: int) -> int:
...
+            print_task_instance_states(dag_run_id, timeout)
+
+    timeout = step_timeout()
+    if timeout is None:
+        log("Diagnostic budget exhausted; skipping task log fetch.")
+        return
+    print_last_run_logs(timeout)
+
+
+def main() -> None:
```
ayush-shah, line 422: [P2] Please commit a focused regression test for this failure path. This Docker-executed module is outside the current pytest scope, so the green RDF E2E run does not deterministically cover a run observed as `running` followed by a transient Airflow failure before timeout. A stubbed HTTP boundary around `main()`/`dump_diagnostics()` would cover that contract without requiring Docker and would catch the P1 case above.

ayush-shah (commented):
Two non-duplicate follow-ups from the timeout-path review. The existing bot feedback already covers the retry-override and shared auth/request budget issues, so I have not repeated those.

tutte (dismissed):

IceS2 (dismissed):

docker/validate_compose.py
```diff
@@ -25,164 +61,247 @@ def get_env_int(name: str, default: int) -> int:
...
+def get_last_run_info() -> tuple[str | None, str | None, bool]:
     """
-    Show the logs
+    Latest sample_data DAG run id and state.
+
+    The third element reports whether the poll itself succeeded, so the caller can
+    tell "Airflow has no run yet" apart from "we could not reach Airflow" — during the
+    ~40s startup window those are very different things.
     """
+    response = airflow_get(f"/api/v2/dags/{DAG_ID}/dagRuns", poll_request_timeout())
```
ayush-shah: [P1] Validate the exact run triggered above instead of selecting the newest item returned by this unfiltered list. `run_local_docker_common.sh` discards the trigger response’s `dag_run_id`; Airflow 3.3.1 paginates this endpoint (default limit 50, ordered by `id`). Once history exceeds that page, the validator can select an older successful run and declare setup healthy while the newly triggered run is queued or failed. Forward the trigger ID through `docker exec` and fetch `/dagRuns/{dag_run_id}`; use a logical-date-filtered fallback only when the trigger response is malformed. Please add a regression with historical successes plus a non-terminal target run.

docker/run_local_docker_common.sh
```diff
@@ -575,8 +575,38 @@ run_local_docker_main() {
...
+    else
+      validate_compose_margin=0
+    fi
+    validate_compose_timeout=$(( validation_timeout_seconds - validate_compose_margin ))
+
+    validate_compose_env=(
+      -e "VALIDATE_COMPOSE_TIMEOUT_SECONDS=${validate_compose_timeout}"
+      -e "VALIDATE_COMPOSE_RETRY_INTERVAL_SECONDS=${validate_compose_retry_interval}"
+    )
+    # An explicit retry count still wins over the timeout-derived deadline.
```
ayush-shah: [P1] Keep the outer `timeout` as the authority for every supported configuration. The explicit retry override currently wins over the forwarded inner deadline: with a 300-second outer window, `VALIDATE_COMPOSE_MAX_RETRIES=60` and a 10-second interval yields a 600-second validator deadline, so the outer process can again terminate diagnostics. A window of 60 seconds or less also reserves no margin. Clamp the retry-derived budget to the forwarded deadline, always reserve the diagnostic margin, and reject a host timeout too short to preserve that contract.

ayush-shah:
Addressed the two confirmed P1s in `1405adacbd`:

- Validation now follows the `dag_run_id` returned by the trigger request, with a logical-date-filtered fallback, so an earlier successful run cannot satisfy this invocation.
- `VALIDATE_COMPOSE_MAX_RETRIES` can only shorten the forwarded deadline; every accepted host timeout reserves 60 seconds for diagnostics, and unsafe windows (≤60s) fail clearly.

Added focused regression coverage for a failed target alongside a historical success, the logical-date fallback, retry clamping, and transient-poll timeout diagnostics.

ayush-shah:
Addressed the remaining confirmed review items in 698b61f46a:\n\n- Timeout diagnostics now query the known triggered run even if every normal poll is interrupted, so the final error retains the DAG-run and per-task state needed to distinguish a slow run from a stuck one.\n- The validator now reuses one HTTP session and refreshes a rejected Airflow token once within the same request deadline.\n- Malformed JSON and malformed response shapes are handled as transient polling failures; task-state diagnostics URL-encode run IDs and paginate task instances within their diagnostic budget.\n- Strict validation no longer says setup will continue immediately before it exits; non-strict mode explicitly says it continues despite the validation failure.\n\nAdded focused HTTP-boundary regressions for the timeout, reauthentication, malformed-response, encoded-run-ID, and paginated-task-state paths.

ayush-shah (approved):

akash-jain-10 (approved):

Issue #24180, opened 2025-11-05
Feature Request: Improve domain isolation in UI and lineage for multi-tenant setups


GregorioBlazquez:
**Is your feature request related to a problem? Please describe.**
We’re using OpenMetadata **v1.10.0** in a multi-tenant setup where each client is isolated by **Domain**, with access controlled through **Teams, Roles, and Policies**.  
While this works well for asset-level permissions, we’ve found that users restricted to a single domain can still **see the names of other domains**, as well as **activity feed entries** and **global lineage data** from outside their allowed domain.  

Even though they can’t open or access those assets, this partial visibility still represents a **security and data isolation issue**, especially in multi-tenant or multi-client environments.

**Describe the solution you'd like**
We’d like OpenMetadata to fully **isolate domain visibility** in the UI and lineage views.  
Users restricted to one domain should:
- Only see **their own domain** in dropdowns, tabs, and lists.  
- Only see **activity feed items** related to entities in their domain.  
- Only see **lineage graphs** containing entities within their domain.  

This would ensure true logical isolation for multi-tenant setups, without exposing unrelated metadata.

**Describe alternatives you've considered**
We’ve implemented logical separation using:
- Domains and Teams per client  
- Custom roles and policies to restrict access  

However, these don’t fully hide domain names or lineage elements from other tenants.  
Running multiple OpenMetadata instances (one per client) could solve this, but it introduces significant operational complexity.  

A UI-level isolation mechanism would be a much simpler and more scalable approach until a **federated OMD system** becomes available.

**Additional context**
- **OpenMetadata version:** 1.10.0  
- **Observed in:** Domain dropdowns, activity feed, and global lineage tab  
- **Use case:** Multi-tenant deployment serving multiple external clients

cecilemurray-cyber:
Hi, we would really benefit from having domain based isolation of lineage and activity feed. Could you share any update about the status of this request?

Prakhar-Mishra1107:
Hi, which OM version will have this fix ?

And by when ?

FYI - I am using 1.11.5 version

Prakhar-Mishra1107:
Hi @chirag-madlani @mohityadav766 
Any updates? 
This feature is extremely important for supporting multiple tenants/clients/teams

Prakhar-Mishra1107:
Can we pick this one on priority, if possible?

Prakhar-Mishra1107:
Can see this issue moving back-n-forth. 
[...]

PR #31741, opened 2026-08-19
Fixes 31740: Enforce caller policies on the Incident Manager listing

### LINKED ISSUE #31740: Incident Manager listing ignores domain policies — domain-restricted users see other domains' incidents
Follow-up to #24180, which covered the same class of problem (a domain-restricted user seeing metadata from outside their domain) for domain listings and lineage. The **Incident Manager listing** still has this gap.

### What happens

A user restricted to their own domain via a `hasDomain()` / `!hasDomain()` policy sees failed test case incidents from **every** domain on the Incident Manager page. Opening one is correctly blocked, but the listed row already exposes the test case name, the table name and the full FQN.

Reported by community users on 1.12.1 and still reproducing on 1.12.13, 1.13.1 and 1.13.3. Also reproduces on current `main`.

Note that #21808 / commit 7c9d85ae2e added a *user-supplied* `domain` filter parameter to this endpoint. That is a filter the caller opts into, not policy enforcement, so it does not close this gap.

### Root cause

`RuleEvaluator#hasDomain` short-circuits to `true` for list operations, because a listing has no single resource to evaluate the condition against:

```java
if (resourceContext.getEntity() == null || resourceContext.getEntity().getId() == null) {
  LOG.info("hasDomain() - List operation detected (no specific resource), returning true for post-filtering");
  return true;
}
```

So `!hasDomain()` is `false` for every deny rule on a listing, and enforcement is deferred to "post-filtering" on the search query. For the incident listing that post-filtering never happens: `EntityTimeSeriesRepository#listFromSearchWithOffset` and `#listLatestFromSearch` call `SearchRepository` **without** a `SubjectContext`, so the existing `RBACConditionEvaluator` machinery never runs. `latest=true` goes through the aggregation path and is unfiltered for the same reason.

Consequence: no policy the administrator can write will restrict this listing.

### Steps to reproduce

1. Create domains `A` and `B`; put a table (and its test case) in `B`.
2. Create an incident for that test case.
3. Create a user in domain `A` holding a policy that denies `ViewAll` when `!hasDomain()` (the seeded `DomainOnlyAccessRole` works).
4. Set `searchSettings.globalSettings.enableAccessControl = true`.
5. As that user, `GET /api/v1/dataQuality/testCases/testCaseIncidentStatus/search/list`.

Expected: domain `B`'s incident is absent.
Actual: it is listed, with the test case name and full FQN. `GET .../testCaseIncidentStatus/{id}` on that same row returns 403.

### Related

- #24180 — same class, domain listings and lineage
- #21808 — added the user-supplied `domain` filter parameter (not enforcement)

ShaileshParmar11:
### Describe your changes:

Fixes #31740

A user restricted to their own domain saw failed test case incidents from **every** domain on the Incident Manager page. Opening one was correctly blocked, but the listed row already leaked the test case name, the table name and the full FQN.

`RuleEvaluator#hasDomain` short-circuits to `true` for list operations, because a listing has no single resource to evaluate the condition against, and defers enforcement to "post-filtering" on the search query:

```java
if (resourceContext.getEntity() == null || resourceContext.getEntity().getId() == null) {
  LOG.info("hasDomain() - List operation detected (no specific resource), returning true for post-filtering");
  return true;
}
```

For this endpoint that post-filtering never happened — the two search-backed listing paths in `EntityTimeSeriesRepository` called `SearchRepository` without a `SubjectContext`, so the existing `RBACConditionEvaluator` machinery never ran. The result is that **no policy an administrator can write restricts this listing**. Upgrading does not help: #21808 / 7c9d85ae2e added a *user-supplied* `domain` filter parameter, which is a filter the caller opts into rather than policy enforcement.

This threads the caller's `SubjectContext` through both paths so the existing RBAC search machinery filters the query. Follow-up to #24180, which fixed the same class of leak for domain listings and lineage.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

Reuses the existing search-side RBAC mechanism rather than adding a second enforcement path. `RBACConditionEvaluator` already compiles policy conditions — including `hasDomain()` — into an engine query, and `SearchRepository#listWithOffset` already had a `SubjectContext`-aware overload wired through to `applyRbacCondition` in both the Elasticsearch and OpenSearch managers. The only missing link was that the time-series listing never passed a subject.

- `EntityTimeSeriesRepository#listFromSearchWithOffset` — new `SubjectContext`-aware overload calling the existing subject-aware `SearchRepository#listWithOffset`.
- `EntityTimeSeriesRepository#listLatestFromSearch` — same, through `SearchRepository#aggregate`, so `latest=true` cannot bypass the filtering applied to the plain listing. This required a `SubjectContext`-aware `aggregate()` through `AggregationManagementClient` and both clients/aggregation managers, mirroring the `genericAggregation()` overload that already existed there.
- `TestCaseResolutionStatusResource` — passes the caller on both branches.

Backward compatibility: every existing signature is kept and delegates with a `null` subject, so all other callers are unchanged. Filtering honours the existing `searchSettings.globalSettings.enableAccessControl` gate, consistent with the rest of the product; that gate defaults to `false`.

Alternatives rejected:
- Changing `hasDomain()` to not short-circuit on list operations. There is no resource to evaluate against at that point, so it cannot work, and it would alter policy evaluation for every list endpoint in the product.
- Post-filtering the returned rows in the resource. That breaks `paging.total` and page sizes.

Deliberately out of scope: the same gap exists on other time-series listings that call these methods (`TestCaseResultResource`, `TestSuiteResource`, and the generic `EntityTimeSeriesResource`). The mechanism is now available to them, but enabling it per endpoint needs its own validation — the RBAC domain clause filters on `domains.id`, so each index mapping has to be confirmed to carry that field first. I verified it for `test_case_resolution_status_index_mapping.json` only.

#
#### Use cases covered
- A user restricted to domain A does not see incidents whose test case belongs to domain B in the Incident Manager listing, while still seeing incidents from their own domain.
- The same holds for `latest=true`, which goes through the aggregation path.
- Admins, bots, and deployments with `enableAccessControl = false` are unaffected.

#### Unit tests
Not applicable — the change is wiring an existing mechanism through a call chain, so a unit test here could only assert that one method calls another with a given argument. Covered by the integration test below instead, which asserts the observable API behaviour.

- [x] Added `DomainIsolationIT#test_incidentListing_restrictedUserSeesOnlyOwnDomainIncidents` in `openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DomainIsolationIT.java`, alongside the existing domain/lineage isolation tests added for #24180.

#### Playwright (UI) tests
- [x] Added `playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts`, alongside the existing `DomainIsolation` specs (it is picked up by the dedicated `DomainIsolation` project).

The user under test holds `DomainOnlyAccessRole` with an **empty** domain list, and that detail is load-bearing. A user *with* a domain is not a valid regression guard here: `useIncidentList` passes `activeDomain` as the `domain` query param, and that caller-supplied filter alone hides foreign incidents even against unfixed code — such a spec would pass on a broken server and protect nothing. With no domain the page sends no `domain` param, leaving the server-side policy as the only filter.

Assertions run against both the payload the page requested and the rendered rows: the list is paginated, so a leaked row can sit on page 2 while page 1 looks clean, and a DOM-only assertion would let a real leak through. Indexing is handled by the existing `seedFailedIncidents`, which polls until every seeded incident is searchable.

#### Manual testing performed
Reproduced and root-caused against a running 2.0.0-SNAPSHOT server before fixing:

1. Created domains `A` and `B`, put a table and its test case in `B`, and created an incident for it.
2. Created a non-admin user in domain `A` with a policy denying `ViewAll` when `!hasDomain()`.
3. Enabled `searchSettings.globalSettings.enableAccessControl`.
4. `GET /api/v1/dataQuality/testCases/testCaseIncidentStatus/{id}` on the domain `B` incident → `403 User does not have ANY of the required permissions`.
5. `GET /api/v1/dataQuality/testCases/testCaseIncidentStatus/search/list` → the same incident was still listed, with its test case name and full FQN.
6. Server logs confirmed the mechanism: the detail request logged `hasDomain() - No matching domains found, returning false`, while the listing logged `hasDomain() - List operation detected (no specific resource), returning true for post-filtering`.

Separately confirmed that the `domain` query parameter on `GET /v1/dataQuality/testCases/testCaseIncidentStatus` is silently ignored — 140 rows returned with no parameter, with a real domain, and with a nonexistent one. That is a distinct bug on the DB-backed listing path (the resource writes the `ListFilter` key `"domain"`, which nothing reads; the sibling endpoints correctly write `incidentDomainId` and `"domains"`). Left out of this PR to keep the security fix reviewable — happy to send it as a follow-up.

**Verified against a locally built server** (fresh MySQL + Elasticsearch, migrations, the server run from the compiled classes with the production UI bundle served from its classpath, everything on `:8585`):

| | userA in domain A, incident in domain B |
|---|---|
| Reverted build | listing returns **both** incidents — leak |
| This build | listing returns only domain A's incident |

`latest=true` (the aggregation path) filters correctly too. The Incident Manager page was checked in a real browser: the restricted user sees only their own domain's incident.

Playwright results on that server: 2/2 pass, 10/10 with `--repeat-each=5`, and the full `DomainIsolation` lane is 16/16. Against the reverted build the isolation test fails on the leaked FQN while the admin control still passes, confirming the spec actually detects the regression rather than passing vacuously.

**What I have not run:** the `DomainIsolationIT` integration test. Its harness builds a server image and boots its own containers, which did not fit in the Docker memory available; I validated the same behaviour by hand and through Playwright instead. CI is the first execution of that IT.

**Worth knowing for anyone reproducing this:** the UI masks the bug for a single-domain user. Their domain selector is locked ("You only have access to your assigned domain(s)"), so the page always sends `domain=<their own domain>` and looks correct even on a broken server. The exposure is reachable through any direct API call, and through the UI whenever the selector is not pinned to the user's own domain.

#

No UI code changed — the Incident Manager page already calls `/search/list`; only the server-side filtering changed. Including the screenshots anyway because they show the user-visible effect of the fix.

Both were captured on the same locally built server with everything held constant except the build: the same user (no domain assigned, holding `DataConsumer` + `DomainOnlyAccessRole`), the same **All Domains** filter, the same data, and `enableAccessControl = true` throughout.

**Before fix** — the listing returns all three incidents. `tc_tblA` (domain A) and `tc_tblB` (domain B) both belong to domains this user is not in, and are leaked. Only `tc_tblNone` is legitimately theirs.

before-fix

**After fix** — the same page for the same user returns only `tc_tblNone`, the domainless incident they are entitled to. The two foreign-domain incidents are gone.

after-fix

Note that a user **with** a domain does not reproduce this in the UI: their domain selector is locked to their own domain, so the page sends `domain=<their own domain>` and looks correct even on a server carrying the bug. That caller-supplied filter is what masks it, which is why the screenshots above use a user with no domain and the selector on All Domains.

#
### Checklist:

Bug fix:
- [x] I have added a test that covers the exact scenario we are fixing. The issue number is referenced in the issue body and in this PR.

This PR threads the caller’s `SubjectContext` through ordinary and aggregated incident searches so domain policies constrain Incident Manager results.
- Adds subject-aware time-series listing and aggregation overloads while retaining backward-compatible signatures.
- Applies RBAC queries consistently in Elasticsearch and OpenSearch aggregation managers.
- Updates the resource to pass caller context through both listing branches.
- Adds integration coverage for both `latest=false` and `latest=true`, plus end-to-end Incident Manager isolation coverage.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DomainIsolationIT.java | Adds observable domain-isolation assertions for both ordinary and latest aggregation incident listings, resolving the prior coverage gap. |
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityTimeSeriesRepository.java | Adds subject-aware overloads for offset and latest time-series search paths while preserving existing callers. |
| openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestCaseResolutionStatusResource.java | Resolves the authenticated caller context and passes it through both Incident Manager listing branches. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/AggregationManagementClient.java | Extends aggregation management with an optional caller context using a backward-compatible default implementation. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java | Routes subject-aware aggregations from repositories into the configured search client. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchAggregationManager.java | Applies caller RBAC conditions to Elasticsearch aggregations and fails closed when required conditions cannot produce a query. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OpenSearchAggregationManager.java | Mirrors subject-aware, fail-closed aggregation filtering for OpenSearch. |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts | Adds end-to-end verification that restricted users cannot receive or render incidents belonging to another domain. |

</details>

```mermaid
sequenceDiagram
  participant Caller
  participant Resource as TestCaseResolutionStatusResource
  participant Repo as EntityTimeSeriesRepository
  participant Search as SearchRepository
  participant Backend as Elasticsearch/OpenSearch
  Caller->>Resource: "GET /search/list?latest=false|true"
  Resource->>Resource: Authorize request and resolve SubjectContext
  alt "latest=false"
    Resource->>Repo: listFromSearchWithOffset(..., subjectContext)
  else "latest=true"
    Resource->>Repo: listLatestFromSearch(..., subjectContext)
  end
  Repo->>Search: List or aggregate with SubjectContext
  Search->>Backend: Search query with caller RBAC conditions
  Backend-->>Caller: Authorized incident rows only
```
</details>

<sub>Reviews (7): Last reviewed commit: ["refactor(dq): consolidate aggregation RB..."](https://github.com/open-metadata/openmetadata/commit/a24f29fc2b79b6d6e2182a578cc0763ed3a8dbc9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54693764)</sub>

- Knowledge Base — [Search Indexing and Event/Notification Sync](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/openmetadata-service-search-events.md)
</details>


openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+const openIncidentManager = async (page: Page): Promise<string[]> => {
+  await redirectToHomePage(page);
+
+  const incidentResponse = page.waitForResponse(
+    (response) =>
+      response
+        .url()
+        .includes(
+          '/dataQuality/testCases/testCaseIncidentStatus/search/list'
+        ) && response.status() === 200
```
ShaileshParmar11: this should be assertion and not query request, it will hang for waiting for response in report if api throws error

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+  const body = await response.json();
+
+  return (body.data ?? []).map(
+    (incident: { testCaseReference?: { fullyQualifiedName?: string } }) =>
+      incident.testCaseReference?.fullyQualifiedName ?? ''
+  );
+};
+
+test.describe('Domain isolation - incident manager listing @domain-isolation', () => {
+  test.slow(true);
```
ShaileshParmar11: why we need test.slow at describe level, this one is not right pattern

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+        await noDomainUser.create(apiContext);
+
+        await tenant.create(apiContext);
+
+        // The role carries the domain policy; the empty domain list is the point of this spec — the
+        // user is subject to hasDomain() but owns no domain, so only domainless assets are theirs.
+        await assignDomainOnlyAccess(apiContext, noDomainUser, []);
+
+        await restrictedTable.create(apiContext);
+        await assignDomainToTable(
```
ShaileshParmar11: check if we can use promise.all to make parallel request

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+  test.afterAll('Cleanup', async ({ browser }) => {
+    const { apiContext, afterAction } = await performAdminLogin(browser);
+
+    try {
+      await enableDisableSearchRBAC(apiContext, false);
+      await safeDelete(() => restrictedTable.delete(apiContext));
+      await safeDelete(() => domainlessTable.delete(apiContext));
+      await safeDelete(() => tenant.delete(apiContext));
+      await safeDelete(() => noDomainUser.delete(apiContext));
+      await safeDelete(() => adminUser.delete(apiContext));
```
ShaileshParmar11: lets do promise.all here as well

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  const body = await response.json();
+
+  return (body.data ?? []).map(
+    (incident: { testCaseReference?: { fullyQualifiedName?: string } }) =>
+      incident.testCaseReference?.fullyQualifiedName ?? ''
+  );
+};
+
+test.describe('Domain isolation - incident manager listing @domain-isolation', () => {
```
ShaileshParmar11: "@domain-isolation" should be tag? check other place how we are using

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
ShaileshParmar11: Changed. The predicate matches on URL only and the status is asserted separately:

```ts
const incidentResponse = page.waitForResponse((response) =>
  response.url().includes('/dataQuality/testCases/testCaseIncidentStatus/search/list')
);
await sidebarClick(page, SidebarItem.INCIDENT_MANAGER);
const response = await incidentResponse;

expect(response.status()).toBe(200);
```

With the status inside the predicate an error response never matched, so a failing API surfaced as an opaque timeout instead of a status assertion.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
ShaileshParmar11: Removed. The tests run in ~2.7s against the 60s config budget, so it bought nothing.

What genuinely needs a budget is the hook — `beforeAll` creates the entities and waits on indexing — and `test.slow()` does not extend a hook's timeout. So `beforeAll` now sets an explicit one, matching the precedent (and the comment explaining it) in `IncidentManagerPagination.spec.ts`.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
ShaileshParmar11: Done, in phases that respect the real dependencies:

1. `Promise.all` — create both users, the domain and both tables
2. `Promise.all` — set the admin role, assign the domain-only role, assign the domain to the table
3. `Promise.all` — seed both incidents

Phase 2 has to precede phase 3 rather than collapsing into one flat `Promise.all`: the incident document inherits its domains from the test case, which inherits them from the table, so the table must already carry its domain when the incident is indexed.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
ShaileshParmar11: Done. Tables are deleted in parallel first, then the domain and both users in parallel.

Kept as two phases rather than one flat `Promise.all` so the domain is removed after the tables that reference it.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DomainIsolation/DomainIncidentIsolation.spec.ts
ShaileshParmar11: Changed to the option form:

```ts
test.describe(
  'Domain isolation - incident manager listing',
  { tag: ['@domain-isolation'] },
  () => {
```

Checked the repo before changing it: 57 specs use `{ tag: [...] }` against 8 with the tag inside the describe title — the four existing `DomainIsolation` specs are among those 8, which is where I had copied it from. Verified `--grep "@domain-isolation"` still selects both tests, so lane selection is unaffected. I left the sibling specs alone as out of scope here.

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DomainIsolationIT.java
```diff
@@ -218,6 +224,96 @@ void test_domainSearch_restrictedUserSeesOnlyOwnDomains(TestNamespace ns) throws
...
+
+      OpenMetadataClient restricted = createRestrictedUserClient(admin, p, ownDomain, cleanup);
+
+      // Incident listing isolation is driven through search RBAC, which is gated on the global
+      // enableAccessControl setting.
+      boolean originalAccessControl = enableSearchAccessControl(admin);
+      cleanup.push(() -> restoreSearchAccessControl(admin, originalAccessControl));
+
+      Awaitility.await("Incident manager listing honours domain isolation")
+          .atMost(Duration.ofSeconds(90))
```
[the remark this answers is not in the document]
ShaileshParmar11: Valid — the helper omitted `latest`, whose endpoint default is `false`, so the aggregation branch had no coverage. That branch is a trivial bypass if unguarded, since it is reachable by adding one query param.

The helper now takes the flag and the assertion runs over both:

```java
for (boolean latest : new boolean[] {false, true}) {
  Set<String> visible = incidentTestCaseFqns(restricted, latest);
  assertTrue(visible.contains(ownIncidentFqn), ...);
  assertFalse(visible.contains(foreignIncidentFqn), ...);
}
```

Confirmed against a locally built server: a domainless user gets only the domainless incident on both `latest=false` and `latest=true`, and both return every incident on a reverted build.

sonika-shah (approved):

shrabantipaul-collate (approved):

karanh37 (approved):

chirag-madlani (approved):

PR #31751, opened 2026-08-19
fix(e2e): assert exact asset count with 60s retry for async tag propagation in glossary test

anuj-kumary:
## Summary

- **Root cause**: `TableRepository.getSearchPropagationDescriptors()` propagates glossary tags to all column documents in Elasticsearch. Since PR #31112, this propagation runs as an async fire-and-forget `update-by-query` (`waitForCompletion=false`). The `ALL` index count after tagging a table is therefore non-deterministic: 1 (table only) + 0–9 columns = 1–10.
- **Why it was passing before**: Prior to #31112, propagation was synchronous — all 9 columns were updated before the PATCH response returned, so count was always 10. `toContainText('1')` matched `"10"` because `"10".includes("1") === true`.
- **Why it started failing**: With async propagation, count can be any value 1–10. `toContainText('1')` fails when count is 2–9 (e.g. `"2".includes("1") === false`).

## Fix

- Replace the flaky `goToAssetsTab(..., 1)` substring check with an **exact count assertion** computed as `1 (table) + getTableColumnsCount(table.children)` (= 10 for `TableClass`).
- Use `toHaveText(String(expectedCount), { timeout: 60_000 })` so Playwright retries every ~100ms for up to **1 minute**, waiting for the async update-by-query to propagate the tag to all column documents before asserting.

## Test plan

- [ ] Test passes reliably: waits until all 9 column docs are updated (count = 10) before asserting
- [ ] If propagation fails to complete within 60s, test fails with a clear timeout message (real infra issue, not a race)
- [ ] No changes to `goToAssetsTab` or other tests

----
## Summary by Gitar

- **E2E & Test Improvements:**
    - Stabilized Data Quality and Incident Manager tests along with owner picker selection
    - Added support for admin JSON storage state and fixed article quick links
- **Backend & Ingestion Updates:**
    - Added `agentType` filter and stopped polling orchestrators for queued status
    - Updated Kafkaconnect source and DBT metadata parsing logic


anuj-kumary:
Superseded by #31755 — clean branch from main with only the test change.

PR #31757, opened 2026-08-19
fix(security): pull expat 2.8.3-1~deb13u1 into the ingestion-base image

Khairajani:
## What

Adds `expat` to the existing dpkg source-package upgrade in the late root layer of `ingestion/operators/docker/Dockerfile` and `Dockerfile.ci`.

## Why

The image reports expat **2.8.2** for **CVE-2026-72522** (DSA-6446-1 — out-of-bounds read and resultant infinite loop in libexpat's `*_toUtf16` surrogate handling). trixie-security has `2.8.3-1~deb13u1`.

expat is **not in the `python:3.12-slim-trixie` base at all** — it arrives transitively from the top-of-file `apt-get install`. That layer is exactly why the image is stuck at 2.8.2: its cache key never changes, so it keeps resolving against the Debian index that was current when it was first built. Adding expat *there* would freeze the fix the same way.

It goes in the late root layer instead, below the COPY and pip layers, where the index is re-read on every build — the same reasoning that put util-linux there.

## How

expat rides in the existing dpkg query rather than a second `RUN`, so one apt call covers both source packages and the next OS CVE is a one-line edit:

```sh
ul="$(dpkg-query -W -f='${source:Package} ${Package}\n' | awk '$1=="util-linux"{print $2}')"; \
[ -n "$ul" ] || { echo "no src:util-linux packages found; refusing to skip the CVE patch" >&2; exit 1; }; \
ex="$(dpkg-query -W -f='${source:Package} ${Package}\n' | awk '$1=="expat"{print $2}')"; \
apt-get -qq install -y --only-upgrade $ul $ex; \
```

The package set stays computed from dpkg rather than hand-listed, for the reason already documented there: scanners report each binary of a source package separately, so a hand-written list silently leaves one behind.

**util-linux keeps its non-empty assert and expat deliberately does not get one.** util-linux is Essential, so an empty query there means `dpkg-query` misbehaved and the build must stop. expat is transitive, so an empty query is a legitimate image with nothing to patch — asserting it would break the build for no reason. Asserting one of the two still covers the hazard the check exists for: `apt-get install --only-upgrade` with no package arguments exits 0, so an all-empty query would give a green build that shipped the vulnerable packages.

`Dockerfile.ci` carries its own copy of this block and gets the same change — it is the base that Collate's Snyk scan builds from.

## Verification

Ran the `RUN` block verbatim in a real build layer on `python:3.12-slim-trixie`:

| package | before | after |
|---|---|---|
| `libexpat1` | `2.7.1-2` | `2.8.3-1~deb13u1` |
| all nine util-linux binaries | `2.41-5` | `2.41.5-0+deb13u1` |

Exit 0. The **expat-absent path** was exercised separately on the bare base and also exits 0, upgrading util-linux alone — that path is real, since the base ships no expat.

## Out of scope, deliberately

`ingestion/Dockerfile` is untouched because there is **nothing to upgrade to**, not because it is clean. `apache/airflow:3.3.0-python3.12` is bookworm; Debian's tracker still marks bookworm *and* bookworm-security **vulnerable**, and the image already carries `2.5.0-1+deb12u2` — the newest build either suite offers. Needs revisiting when Debian ships a bookworm fix.

## Related

Collate-side half: open-metadata/openmetadata-collate#5909 — the two must land together, or the collate image keeps shipping what the base tag was pinned to.

Adds expat binaries to the existing late-layer Debian security upgrade so ingestion-base images receive the fixed package from trixie-security.
- Queries installed binaries originating from the expat source package.
- Upgrades expat and util-linux binaries in one apt invocation.
- Applies the same change to production and CI ingestion-base Dockerfiles.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/operators/docker/Dockerfile | Adds installed expat binaries to the late security-upgrade layer without changing the valid expat-absent path. |
| ingestion/operators/docker/Dockerfile.ci | Mirrors the ingestion-base expat security upgrade in the CI image definition. |

</details>

<sub>Reviews (5): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/expat-cve-2..."](https://github.com/open-metadata/openmetadata/commit/74ecc0087959a8addb80a391ee3879f04961b0fe) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54716800)</sub>


ulixius9 (approved):

Khairajani:
Superseded by #31890, which carries this expat fix alongside the airflow 3.3.1 bump from #31861 as two separate commits. Combined because both are security fixes for the same image family with disjoint file sets.

PR #31769, opened 2026-08-19
Fixes #31768: keep live log auto-follow on after a hand-made resume

### LINKED ISSUE #31768: Live log auto-follow pauses itself after the user scrolls back to the tail
### Describe the bug

In the live (streaming) log viewer, auto-follow can resume in two ways: the toolbar toggle, or the user scrolling back down to the tail by hand. After resuming **by scrolling**, auto-follow switches itself back off a moment later while the stream is still appending — the `Live auto-scroll` toggle flips to `aria-pressed="false"` on its own and the log stops tracking the tail. Resuming via the toolbar toggle does not show the problem.

### Root cause

`useLogAutoFollow.ts` has two resume paths that leave different internal state:

| Path | `viewerScrollAtRef` (the "the viewer moved the view, not the user" grace) |
|---|---|
| `resumeFollowingTail()` — toolbar toggle / jump-to-end | `Date.now()` — grace armed |
| `applyUserScrollIntent(true)` — user scrolls to the tail | untouched (`0`) — never armed |

`pauseFollow()` zeroes that ref, so a pause → scroll-back-to-tail sequence resumes following with no grace window. The pause decision then keyed on `userMovedTheView`, which is direction-agnostic (`Math.abs(movedBy) > SCROLL_MOVED_THRESHOLD_PX`). So the next scroll report in which the library's own follow-scroll landed more than `SCROLL_BOTTOM_THRESHOLD_PX` (40px) short of the tail was read as *the user leaving the tail*, and following paused. A virtualised list produces exactly that report when a second append lands mid-scroll or rows are re-measured.

### To Reproduce

1. Open the logs of a running ingestion agent (a live run, so the `Live auto-scroll` toggle is present).
2. Scroll up to read earlier output — following pauses, as intended.
3. Scroll back down to the tail by hand — following resumes and the toggle shows pressed.
4. Wait for the stream to append more lines.

**Expected:** the log keeps following the tail.
**Actual:** the toggle flips back to unpressed and the log stops following, without the user touching it.

### Impact

Reported by CI rather than by a user: it fails `playwright/e2e/Features/AgentLogStreamHandover.spec.ts:327` intermittently, e.g. [run 32244090565](https://github.com/open-metadata/OpenMetadata/actions/runs/32244090565/job/96047812481). It reproduces only when a report lands short of the tail, which is timing-dependent — hence "only on CI" for the user-visible bug too.

### Version

- OpenMetadata version: 2.0.0-SNAPSHOT (`main`)
- Introduced in: #31610 (`Fixes 31609: let users pause and resume live log auto-follow`)

aniketkatkar97:
### Describe your changes:

Fixes #31768

Auto-follow in the live log viewer could resume two ways — the toolbar toggle, or the user scrolling back down to the tail — and the two left **different internal state**. After a hand-made resume, following switched itself back off a moment later while the stream was still appending.

`pauseFollow()` zeroes `viewerScrollAtRef` (the "the viewer moved the view, not the user" grace), and only `resumeFollowingTail()` re-armed it. So a pause → scroll-back-to-tail sequence resumed with no grace. The pause decision then keyed on `userMovedTheView`, which is direction-agnostic (`Math.abs(movedBy) > SCROLL_MOVED_THRESHOLD_PX`), so the next report in which `LazyLog`'s own follow-scroll landed more than 40px short of the tail — which a virtualised list emits when a second append arrives mid-scroll or rows are re-measured — was read as *the user leaving the tail*, and following paused.

The toolbar-toggle path never showed it because it asserts inside the freshly armed 1s grace.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change. Three source files, one behavioural invariant.

Two complementary changes in `useLogAutoFollow.ts`:

1. **Extract `beginFollowingFromTail()`** and call it from *both* resume paths (`resumeFollowingTail` and the resume arm of `applyUserScrollIntent`), so the two cannot drift apart again — that drift is the bug.
2. **Gate the pause on direction.** New `movedTowardsTail` fact in `readScrollFacts`: a view that travelled *towards* the tail and stopped short of it is the viewer landing approximately, not the user leaving. Nobody scrolls down in order to abandon the tail.

`movedTowardsTail` is deliberately **positive evidence of direction** rather than `!movedAwayFromTail`. A first report has no previous offset to compare against, so it is neither — which preserves the existing "log opens already scrolled up ⇒ don't follow" behaviour. My first attempt used the negation and regressed three existing tests; those tests are what caught it.

Rejected alternative: arming the grace on the scroll-resume path *only* (change 1 without change 2). That fixes the reported failure but leaves a residual race — once the 1s grace lapses, a quiet stream is exposed to the same misread. Change 2 is the load-bearing one; change 1 makes the two paths consistent.

#
#### Use cases covered

- A user reading back through a live log scrolls down to the tail by hand; the log keeps following as new lines arrive, instead of pausing itself.
- The toolbar toggle's resume behaves identically to a hand-made resume (same grace, same catch-up).
- A native scrollbar drag — no wheel, no key, no pointer event — still takes the log back from a hand-made resume, so the fix is not over-applied.
- A log that opens already scrolled away from the tail still does not follow (first-report behaviour unchanged).

#### Unit tests

- Files updated: `openmetadata-ui/src/main/resources/ui/src/components/common/LogViewerModal/LogViewerModal.test.tsx` (3 new cases in the `— auto-follow` describe)
- **2 of the 3 new tests fail on `main`** — verified by stashing the source changes and re-running; they reproduce both defect paths deterministically. The third is a regression guard for the over-application described above.
- Coverage (`yarn test src/components/common/LogViewerModal --coverage`): **`useLogAutoFollow.ts` 98.27% lines / 93.75% branch, `LogViewerModal.utils.tsx` 100% lines**, 97.87% lines across the directory. 78/78 tests pass.

- [x] Not applicable (no backend changes).

#### Playwright (UI) tests

- [x] Covered by the existing spec — it is what caught this.
- Files updated: `openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/AgentLogStreamHandover.spec.ts`

The `:327` test (`Scrolling a live log pauses auto-follow and the toolbar toggle resumes it`) already asserts the exact failing step. The only change there is `test.slow()`: the test contains four `expect.poll` calls with `timeout: 60_000` against the global 60s budget, and it ran **56.6s / 54.9s / 4.3s** across three identical local runs. That 13× swing against a 60s ceiling is why CI attempt 2 died with `Test timeout of 60000ms exceeded` instead of reporting the real assertion failure.

#### Manual testing performed

Ran against a local dev server on a free port (`yarn start --port 3055 --strictPort`, `PLAYWRIGHT_TEST_BASE_URL=http://localhost:3055`) with a real backend on :8585 — not the prebuilt bundle.

1. `AgentLogStreamHandover.spec.ts:327` with `--repeat-each=5 --workers=1` → **5/5 passed** (a single green run proves nothing for a race).
2. Full `AgentLogStreamHandover.spec.ts` with `--repeat-each=3` → **12 passed**.
3. `playwright/e2e/Pages/IngestionLogStreamLive.spec.ts` → **passed (33.1s)**. This is the important one: it drives the same hook against a **real SSE stream from a real Airflow run**, not the mocked one. Required starting the `openmetadata_ingestion` container, without which it fails at `Deploying pipeline <id> failed with 400` before touching any UI.
4. `npx tsc --noEmit` → no errors in the touched files.
5. `make ui-checkstyle-changed` → 0 errors.

#### Unrelated flake fixed in passing

`playwright/e2e/Pages/IngestionLogStreamLive.spec.ts` failed then passed on retry in this PR's own CI run ([32255074087](https://github.com/open-metadata/OpenMetadata/actions/runs/32255074087)). It predates this PR (added in #31324) and is untouched by the product change here — it asserts line **count**, which auto-follow does not affect — but it was worth fixing while in the area.

Its `Log content keeps growing without a reload` step budgeted 30s for the next batch of lines, assuming a live run writes at a steady rate. It does not. From the failing run's page snapshot:

- burst of 21 log lines across `13:32:19.15` → `13:32:21.83` (2.7s)
- **29.8s of complete silence**
- next burst at `13:32:51.62`

The 30s budget expired **0.2s** before the next burst landed. The gap is the connector's own startup — between the connection test and the first topic being ingested, the Kafka consumer joins its group and blocks on an empty poll, logging nothing. The budget was calibrated directly on top of the gap it had to clear, so which side it fell on was a coin flip.

Widened to 90s. The test already carries `test.slow()` (180s) and ran in 33s, so there is ample headroom, and a stream that connects but delivers nothing still fails — a healthy run does not go silent for 90s.

#

Not applicable — no visual change. The rendered output is identical; only *when* the `Live auto-scroll` toggle flips changes. The behaviour is asserted through `aria-pressed` in both the Jest and Playwright tests above, which is a tighter check than a recording would be.

#
### Checklist:
- [x] For JSON Schema changes: N/A — no schema changes.
- [x] For UI changes: no visual change, see above.

The PR aligns both live-log auto-follow resume paths and uses scroll direction to avoid mistaking the viewer’s catch-up movement for user intent.
- Extracts shared resume-state initialization for toolbar and manual resumes.
- Adds directional scroll facts and preserves auto-follow for movement toward the tail.
- Adds regression coverage and increases timing allowances for long-running Playwright scenarios.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/src/components/common/LogViewerModal/useLogAutoFollow.ts | Unifies resume initialization and prevents toward-tail catch-up movement from disabling auto-follow. |
| openmetadata-ui/src/main/resources/ui/src/components/common/LogViewerModal/LogViewerModal.utils.tsx | Adds directional classification for scroll movement while preserving first-report behavior. |
| openmetadata-ui/src/main/resources/ui/src/components/common/LogViewerModal/LogViewerModal.test.tsx | Covers manual resume persistence, equivalent catch-up grace, and gestureless scrollbar takeover. |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/AgentLogStreamHandover.spec.ts | Extends the existing multi-stage live-stream test budget with test.slow(). |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/IngestionLogStreamLive.spec.ts | Expands the live-line polling timeout to accommodate legitimate quiet ingestion phases. |

</details>

<sub>Reviews (2): Last reviewed commit: ["Stop IngestionLogStreamLive racing the c..."](https://github.com/open-metadata/openmetadata/commit/cf80e02ef417463b1de8ec0f5b53484943784170) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54748663)</sub>

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/AgentLogStreamHandover.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/AgentLogStreamHandover.spec.ts
@@ -327,6 +327,11 @@ test.describe('Agent log stream handover to the paginated endpoint', () => {
   test('Scrolling a live log pauses auto-follow and the toolbar toggle resumes it', async ({
     page,
   }) => {
+    // Every step below waits for the stream to append again, and a reconnect can
+    // take seconds on a loaded runner. The default budget is one such wait, not
+    // the four this test needs, so it timed out on retry rather than failing.
+    test.slow();
+
     await openAgentLogs(page, {
       terminal: false,
       lineCount: SCROLLABLE_LOG_LINE_COUNT,
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/IngestionLogStreamLive.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/IngestionLogStreamLive.spec.ts
@@ -290,13 +290,22 @@ test.describe(
         ).toBeGreaterThan(0);
 
         // The server reads the run's log every 2s (LogStreamSettings.pollSeconds),
-        // so a still-running agent must push more lines within a few ticks. A
-        // stream that connects but delivers nothing fails here.
+        // but the connector does not write at a steady rate: it logs a burst, then
+        // goes quiet for however long its next phase takes. The longest gap is
+        // between the connection test and the first topic being ingested, where the
+        // Kafka consumer joins its group and blocks on an empty poll — nothing is
+        // logged for the whole of it. A measured CI run sat silent for 29.8s there
+        // and this assertion, then budgeted 30s, gave up 0.2s before the next burst
+        // landed. The window has to clear that gap with margin rather than race it;
+        // `test.slow()` above leaves ample room (the whole test ran in 33s).
+        //
+        // A stream that connects but delivers nothing still fails here — 90s of
+        // silence is not something a healthy run produces.
         await expect
           .poll(() => getLogViewerLineCount(page), {
             message:
               'the log viewer should keep receiving lines while the run is live',
-            timeout: 30_000,
+            timeout: 90_000,
             intervals: [2_000],
           })
           .toBeGreaterThan(initialLineCount);
--- a/openmetadata-ui/src/main/resources/ui/src/components/common/LogViewerModal/LogViewerModal.interface.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/components/common/LogViewerModal/LogViewerModal.interface.ts
@@ -74,6 +74,7 @@ export interface ScrollFacts {
   fillsViewport: boolean;
   userMovedTheView: boolean;
   movedAwayFromTail: boolean;
+  movedTowardsTail: boolean;
 }
 
 export interface UseLogAutoFollowParams {
--- a/openmetadata-ui/src/main/resources/ui/src/components/common/LogViewerModal/LogViewerModal.test.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/common/LogViewerModal/LogViewerModal.test.tsx
@@ -821,6 +821,90 @@ describe('LogViewerModal — auto-follow', () => {
     );
   });
 
+  it('keeps following after a hand-made resume when an append lands short of the tail', () => {
+    render(<LogViewerModal {...defaultProps} mode="stream" />);
+
+    act(() => mockLazyLog.onScroll?.(atTail));
+    fireEvent.wheel(screen.getByTestId('log-viewer-body'), { deltaY: -120 });
+    fireEvent.wheel(screen.getByTestId('log-viewer-body'), { deltaY: 120 });
+    act(() => mockLazyLog.onScroll?.(scrolledBackToTail));
+
+    expect(screen.getByTestId('log-viewer-follow')).toHaveAttribute(
+      'aria-pressed',
+      'true'
+    );
+
+    // The library scrolls itself on every append and lands approximately: the
+    // offset moved a long way *towards* the tail and still stopped short of it.
+    // Nobody scrolls down in order to leave the tail, so this cannot be the user.
+    act(() =>
+      mockLazyLog.onScroll?.({
+        scrollTop: 800,
+        scrollHeight: 1500,
+        clientHeight: 400,
+      })
+    );
+
+    expect(screen.getByTestId('log-viewer-follow')).toHaveAttribute(
+      'aria-pressed',
+      'true'
+    );
+  });
+
+  it('grants a hand-made resume the same catch-up grace as the toggle', () => {
+    render(<LogViewerModal {...defaultProps} mode="stream" />);
+
+    act(() => mockLazyLog.onScroll?.(atTail));
+    fireEvent.wheel(screen.getByTestId('log-viewer-body'), { deltaY: -120 });
+    fireEvent.wheel(screen.getByTestId('log-viewer-body'), { deltaY: 120 });
+    act(() => mockLazyLog.onScroll?.(scrolledBackToTail));
+    mockLazyLog.scrollToIndex.mockClear();
+
+    // One gestureless report pulling away from the tail is what a relayout does
+    // on its way to re-pinning it, so it is caught up rather than obeyed — the
+    // same answer the toolbar toggle's resume gets.
+    act(() =>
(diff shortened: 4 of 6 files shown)


chirag-madlani (dismissed):

chirag-madlani (approved):

harshach:
The remaining `maven-collate-ci` failure is a stale-base cross-repo compile mismatch, not an AgentLogStream test failure.

The dispatched Collate run [32329861221](https://github.com/open-metadata/openmetadata-collate/actions/runs/32329861221) fails only at these four overrides:

- `ElasticSearchClientExt.listPageHierarchy`
- `ElasticSearchClientExt.listPageHierarchyForActivePage`
- `OpenSearchClientExt.listPageHierarchy`
- `OpenSearchClientExt.listPageHierarchyForActivePage`

Current Collate `main` overrides the new signatures containing `SearchSortFilter`. This PR is based on OpenMetadata `5ab32e5c78...`, where those supertype methods still have the old signatures; current OpenMetadata `main` contains the matching new signatures. Rebase this PR onto current `main`, then rerun `maven-collate-ci`.

The PR's own Playwright run [32329816300](https://github.com/open-metadata/OpenMetadata/actions/runs/32329816300) is clean: 52 passed, 0 failed, 0 flaky. The AUT five-repeat canary [32332272574](https://github.com/open-metadata/openmetadata-nightly/actions/runs/32332272574) is also clean with 15/15 focused targets and zero retries.

PR #31791, opened 2026-08-19
Fixes #31790: Force sqlparse 0.6.0 in ingestion operator images to clear 4 CVEs

### LINKED ISSUE #19122: Masked Query in Lineage Registration Processes Only the First SQL Statement
**Affected module**
Ingestion Framework

**Describe the bug**
In version `v1.6.1`, we've encountered an issue with the SQL query condition data masking feature within the lineage registration process, introduced in `v1.5.11`. The current implementation processes only the first SQL statement during lineage registration, resulting in the omission of subsequent statements in the masked query. This leads to incomplete masked queries when dealing with multi-statement SQL queries; however, the lineage information itself remains unaffected.

**To Reproduce**
1. Create a multi-statement SQL query.
2. Register the query in the lineage ingestion process.
3. Observe that only the first statement is included in the masked query, while subsequent statements are omitted.

**Expected behavior**
All statements within a multi-statement SQL query should be appropriately included in the masked query during the lineage registration process to ensure completeness.

**Version:**
- OS: macOS
- Python version: x
- Docker image: `docker.getcollate.io/openmetadata/ingestion:1.6.1`
- OpenMetadata version: v1.6.1
- OpenMetadata Ingestion package version: `openmetadata-ingestion[docker]==1.6.1`

**Additional context**
Relevant code snippet:
```python
parsed = sqlparse.parse(query)
if not parsed:
    return query
parsed = parsed[0]  # Only the first statement is processed
```

https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/lineage/masker.py#L41-L45

Proposed solution: Proposed solution: Iterate over all statements returned by the parser to ensure that each one is included in the masked query during lineage registration.

### LINKED ISSUE #31790: sqlparse pinned to a vulnerable version in ingestion images: 4 CVEs fixed only in 0.6.0
## Summary

The ingestion images ship a `sqlparse` that is affected by four advisories, and **no version inside the declared ranges is clean** — OSV reports `fixed: 0.6.0` for all four, with no patched 0.5.x release.

| CVE | GHSA | Class | Fixed in |
|---|---|---|---|
| CVE-2026-54284 | GHSA-pwgv-4x5q-6m9f | `TokenList.__init__` materializes O(subtree) per group — CPU DoS | 0.6.0 |
| CVE-2026-59893 | GHSA-prg7-hcfm-mfcr | ReDoS on dollar-quoted literals / multiline comments | 0.6.0 |
| CVE-2026-71491 | GHSA-f2ff-p2ww-7p4p | Quadratic O(n²) DoS in `group_comments` | 0.6.0 |
| CVE-2026-59894 | GHSA-3496-9g83-7v6x | Unescaped backslash breaks out of generated `python`/`php` string literals | 0.6.0 |

The DoS advisories matter here because ingestion parses **attacker-influenced SQL** — query-log lineage and usage workflows feed arbitrary warehouse query history through `sqlparse`, so a single pathological statement can pin an ingestion worker.

## Why the resolver cannot fix this

Three separate ceilings block `sqlparse>=0.6.0`, and all three are stale rather than substantive:

| Source | Pin | Notes |
|---|---|---|
| `collate-sqllineage 2.1.4` (`ingestion/setup.py`) | `sqlparse==0.5.4` | 2.1.5 shipped `sqlparse==0.6.0`; 2.1.6 reverted **only the pin** to stay co-installable with dbt-core. The 2.1.5 and 2.1.6 wheels are byte-identical apart from the version string. |
| `dbt-core`, transitive via `collate-data-diff` | `sqlparse<0.6.0` | Every version including 1.12.2. The ceiling has been in place since the `setup.py` → `pyproject.toml` migration in Nov 2025 — nine months before 0.6.0 existed. Reported upstream: https://github.com/dbt-labs/dbt-core/issues/15988 |
| Airflow `constraints-3.3.0/constraints-3.12.txt` | `sqlparse==0.5.5` | Also vulnerable to all four. Applies to `ingestion/Dockerfile` via the `openmetadata-managed-apis --constraint` layer. |

```
pip install dbt-core==1.12.2 'sqlparse>=0.6.0'
# ERROR: The conflict is caused by:
#     dbt-core 1.12.2 depends on sqlparse<0.6.0 and >=0.5.5
```

## Compatibility evidence

0.6.0 rewrote statement splitting onto a stack-based architecture, so this deserves scrutiny rather than a rubber stamp. Differential test of 0.5.4 vs 0.6.0 over **372 SQL statements** scraped from our own unit tests, plus targeted cases for every 0.5.5/0.6.0 changelog entry:

- **`is_safe_sql_query`** — 0 verdict changes across the 372-statement corpus plus 23 adversarial forbidden-SQL shapes chosen to exercise exactly the keywords 0.6.0 re-tokenizes (`BEGIN`, `MATERIALIZED`, `END CASE`, `ROW_FORMAT`, dollar-quotes). The guard is unaffected.
- **Lineage** — 744 runs across `SqlParseLineageAnalyzer` and sqlfluff/ansi produced 7 differences, **all improvements**:
  - `WITH x AS MATERIALIZED (...)` no longer dies with `ValueError: too many values to unpack (expected 2)`. Under 0.5.4 that query's lineage was lost entirely.
  - Lowercase `as` in `CREATE TABLE t as SELECT f(x) FROM u` now matches uppercase `AS`; 0.5.4 emitted `u.f -> t.f`, a column that does not exist.
  - The phantom `<default>.materialized` source table is gone.
  - Recovered lineage on a subquery CTAS that previously returned no sources.
- **`mask_query`** — 383 queries, 1 difference, and it is a **pre-existing bug rather than a regression**: `masker.py` masks `parser._parsed_result`, a single statement, so any multi-statement query already loses all but the last statement under 0.5.4 (`SELECT 1..; SELECT 2..` → only the 2nd). 0.6.0's corrected splitting merely adds `BEGIN TRANSACTION; …; COMMIT;` to the shapes that hit it. Masking never became weaker — nothing masked before is unmasked after. Related to the already-closed #19122.
- **collate-sqllineage's monkeypatch of sqlparse internals still applies intact** — `grouping.MAX_GROUPING_DEPTH`/`MAX_GROUPING_TOKENS` raised 100×, `KEYWORDS["STRING"]` retyped to `Name.Builtin`, and the `LATERAL VIEW EXPLODE` `SQL_REGEX` rewrite. This was the main silent-degradation risk: if a future sqlparse renames either grouping global, the patch becomes a **no-op** and lineage comes back quietly truncated with nothing failing.

## Proposal

Override `sqlparse` to 0.6.0 in the image builds now, with a fail-closed import gate that asserts both the version and that the collate-sqllineage monkeypatch still bites — the CVEs are concrete and the ceilings are not.

Once dbt-labs/dbt-core#15988 lands, drop the override and raise the floor in `ingestion/setup.py` instead, which fixes it for everyone who installs `openmetadata-ingestion` from PyPI rather than only for our images.

Note this is a narrower, newer instance of the class of problem described in #26597.

ulixius9:
### Describe your changes:

Fixes #31790

I forced `sqlparse==0.6.0` into `ingestion/operators/docker/Dockerfile` and `Dockerfile.ci` because CVE-2026-54284, CVE-2026-59893, CVE-2026-71491 (parser CPU-exhaustion DoS) and CVE-2026-59894 (string breakout in the `python`/`php` output formats) are fixed **only** in 0.6.0 — OSV reports no patched 0.5.x, so no version inside the declared ranges is clean. Three stale ceilings block the resolver from getting there (`collate-sqllineage 2.1.4` → `sqlparse==0.5.4`, `dbt-core` transitively → `sqlparse<0.6.0`, Airflow constraints → `sqlparse==0.5.5`), and `collate-sqllineage 2.1.5` already shipped against `sqlparse==0.6.0` — 2.1.6 reverted only the pin to stay co-installable with dbt-core, and the two wheels are byte-identical apart from the version string. The override is `pip install --no-deps` as the last pip layer plus a fail-closed import gate, because `pip install` exits 0 while printing the resolver conflict and would otherwise leave a silently-vulnerable image. dbt-core's ceiling is reported upstream at dbt-labs/dbt-core#15988; once it lands this layer goes away and the floors move into `ingestion/setup.py`, which fixes it for PyPI consumers too rather than only our images.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small change (2 files, one pip layer each).

The one non-obvious choice worth calling out: the gate asserts that collate-sqllineage's **monkeypatch of sqlparse internals still applies**, not just the version. That patch raises `grouping.MAX_GROUPING_DEPTH`/`MAX_GROUPING_TOKENS` 100×, retypes `KEYWORDS["STRING"]` as `Name.Builtin`, and rewrites the `LATERAL VIEW EXPLODE` `SQL_REGEX` entry. If a future sqlparse renames either grouping global, the patch degrades to a no-op and lineage comes back quietly truncated with **nothing failing** — a version assert alone would not catch that.

#
#### Use cases covered
- Ingestion image builds and the lineage/masking/SQL-guard paths keep working on `sqlparse` 0.6.0
- Query-log lineage no longer loses a whole query on `WITH x AS MATERIALIZED (...)`
- `is_safe_sql_query` keeps rejecting forbidden SQL, including the statement shapes 0.6.0 re-tokenizes

#### Unit tests
No new test files. Verification was a **differential test of 0.5.4 vs 0.6.0** in two venvs over 372 SQL statements scraped from our own `ingestion/tests/unit/` corpus, plus 21 targeted cases covering every 0.5.5/0.6.0 changelog entry:

| Probe | Result |
|---|---|
| `is_safe_sql_query` (loaded verbatim from `metadata/utils/helpers.py`) | **0 verdict changes** — 372 corpus + 23 adversarial forbidden-SQL shapes |
| Lineage via `SqlParseLineageAnalyzer` and sqlfluff/ansi — 744 runs | **7 diffs, all improvements** (see below) |
| `mask_query` (real `metadata/ingestion/lineage/masker.py`) | 383 queries, 1 diff, **pre-existing bug not a regression** |
| collate-sqllineage monkeypatch surfaces | All 4 intact; 400-deep and 4000-tuple queries parse to identical token counts |

The 7 lineage differences, all in 0.6.0's favour:
- `WITH x AS MATERIALIZED (...)` no longer dies with `ValueError: too many values to unpack (expected 2)` — under 0.5.4 that query's lineage was lost entirely
- lowercase `as` in `CREATE TABLE t as SELECT f(x) FROM u` now matches uppercase `AS`; 0.5.4 emitted `u.f -> t.f`, a column that does not exist
- the phantom `<default>.materialized` source table is gone
- recovered lineage on a subquery CTAS that previously returned no sources

On `mask_query`: `masker.py` masks `parser._parsed_result`, a single statement, so **any** multi-statement query already loses all but the last one under 0.5.4 (`SELECT 1..; SELECT 2..` → only the 2nd). 0.6.0's corrected splitting merely adds `BEGIN TRANSACTION; …; COMMIT;` to the shapes that hit it. Masking never became weaker — nothing masked before is unmasked after. Related to the closed #19122; not fixed here.

**Not done:** the full `make unit_ingestion` suite was not run under the override — the differential test covers the same code paths but is not a green suite. Worth one CI run against the built image before merge.

Not applicable — no backend API changes.

Not applicable — no ingestion source changes; this is an image dependency override.

#### Playwright (UI) tests

#### Manual testing performed
1. Built two venvs with `collate-sqllineage==2.1.4`, forcing `sqlparse==0.6.0` in one via `pip install --no-deps`, and confirmed the install exits 0 while `pip check` reports the unsatisfied pins (no CI job runs `pip check`).
2. Ran the differential harness above in both venvs and diffed the JSON output probe by probe.
3. Verified the build gate is not vacuous: it exits 0 under 0.6.0 and exits 1 with `AssertionError: 0.5.4` under 0.5.4.
4. Simulated Docker's line-continuation join on both Dockerfiles and confirmed each block collapses to a single valid shell command (`sh -n` clean).

#

Not applicable.

#
### Checklist:
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed. — N/A, no schema changes
- [ ] For UI changes: I attached a screen recording and/or screenshots above. — N/A, no UI changes

- [x] I have added a test that covers the exact scenario we are fixing — the build-time import gate fails closed if the image ever resolves back to a vulnerable sqlparse or if collate-sqllineage's monkeypatch stops applying.

The PR forces sqlparse 0.6.0 into ingestion operator images to address four vulnerabilities despite stale transitive dependency ceilings.
- Adds the override to production and CI operator images using a final `--no-deps` installation layer.
- Adds build-time assertions for the installed version, collate-sqllineage monkeypatch state, and basic source/target lineage behavior.
- Documents the dependency conflicts and conditions for removing the override.
</details>

The PR appears safe to merge based on the reviewed image dependency override and its fail-closed compatibility checks.

No concrete changed-code failure remained after accounting for the pinned collate-sqllineage version, build-time assertions, and the compatibility evidence provided for exercised ingestion paths.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/operators/docker/Dockerfile | Forces sqlparse 0.6.0 in the production ingestion operator image and validates version and lineage compatibility during the build. |
| ingestion/operators/docker/Dockerfile.ci | Mirrors the sqlparse override and fail-closed compatibility gate in the CI ingestion operator image. |

</details>

<sub>Reviews (1): Last reviewed commit: ["Force sqlparse 0.6.0 in the ingestion op..."](https://github.com/open-metadata/openmetadata/commit/b2b5256c20a7ab46ea7afccc61bdff5eb9ca620b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54858734)</sub>

### DIFF
--- a/ingestion/operators/docker/Dockerfile
+++ b/ingestion/operators/docker/Dockerfile
@@ -230,9 +230,46 @@ USER openmetadata
 # build-time only: cx_Oracle and mysqlclient import pkg_resources from their setup.py and
 # 81.0.0 removed it, but once built both import fine against 83+. Leaving 80.x on disk is
 # what keeps scanners reporting CVE-2026-59890.
-# Keep this the LAST pip layer -- a later layer that compiles a package needing
-# pkg_resources would fail here, and the error would not look like a setuptools problem.
+# Keep this the LAST pip layer that can compile anything -- a later layer that builds a
+# package needing pkg_resources would fail here, and the error would not look like a
+# setuptools problem. The sqlparse override below is a pure-Python wheel, so it is exempt.
 RUN pip install --upgrade "setuptools>=83"
+
+# Force sqlparse past two declared ceilings to clear CVE-2026-54284, CVE-2026-59893,
+# CVE-2026-71491 (parser CPU-exhaustion DoS) and CVE-2026-59894 (SQL string breakout in
+# the python/php output formats). All four are fixed only in 0.6.0 -- OSV reports no
+# patched 0.5.x -- so no in-range version is clean and the resolver cannot help us:
+#     collate-sqllineage 2.1.4                     sqlparse==0.5.4
+#     dbt-core (transitive via collate-data-diff)  sqlparse<0.6.0
+# Both ceilings are stale rather than substantive. collate-sqllineage 2.1.5 shipped with
+# sqlparse==0.6.0 and 2.1.6 reverted only the pin to stay co-installable with dbt-core --
+# the two releases are byte-identical apart from the version string, so 0.6.0 is a version
+# upstream already released against. dbt-core's ceiling predates 0.6.0 by nine months and
+# is tracked at https://github.com/dbt-labs/dbt-core/issues/15988. Once that lands, delete
+# this layer and raise the floors in ingestion/setup.py instead.
+#
+# --no-deps because pip would otherwise backtrack on the declared conflict. `pip install`
+# exits 0 while printing the resolver-conflict ERROR, and `pip check` will report the two
+# unsatisfied pins for the life of the image, so the import gate is what actually keeps
+# this honest. It asserts the two things that would otherwise fail silently: that we really
+# got 0.6.x, and that collate-sqllineage's monkeypatch of sqlparse internals still bites.
+# That patch raises MAX_GROUPING_DEPTH/MAX_GROUPING_TOKENS 100x and retypes STRING as a
+# builtin; if a future sqlparse renames either, the patch degrades to a no-op and lineage
+# comes back quietly truncated with nothing failing.
+RUN pip install --no-deps "sqlparse==0.6.0" \
+ && python -W ignore -c "\
+import sqlparse; \
+from sqlparse.engine import grouping; \
+from sqlparse.keywords import KEYWORDS; \
+import collate_sqllineage.core.parser.sqlparse; \
+from collate_sqllineage.core.parser.sqlparse.analyzer import SqlParseLineageAnalyzer; \
+from collate_sqllineage.runner import LineageRunner; \
+assert sqlparse.__version__.startswith('0.6.'), sqlparse.__version__; \
+assert (grouping.MAX_GROUPING_DEPTH, grouping.MAX_GROUPING_TOKENS) == (10000, 1000000), 'sqllineage grouping patch is a no-op'; \
+assert str(KEYWORDS['STRING']) == 'Token.Name.Builtin', 'sqllineage keyword patch is a no-op'; \
+r = LineageRunner('INSERT INTO db.sch.tgt SELECT c FROM db.sch.src', analyzer=SqlParseLineageAnalyzer); \
+assert [str(t) for t in r.source_tables] == ['db.sch.src'], r.source_tables; \
+assert [str(t) for t in r.target_tables] == ['db.sch.tgt'], r.target_tables"
 
 
 # Strip spaCy's bundled test fixture, which scanners misreport as an installed black.
--- a/ingestion/operators/docker/Dockerfile.ci
+++ b/ingestion/operators/docker/Dockerfile.ci
@@ -238,9 +238,46 @@ USER openmetadata
 # build-time only: cx_Oracle and mysqlclient import pkg_resources from their setup.py and
 # 81.0.0 removed it, but once built both import fine against 83+. Leaving 80.x on disk is
 # what keeps scanners reporting CVE-2026-59890.
-# Keep this the LAST pip layer -- a later layer that compiles a package needing
-# pkg_resources would fail here, and the error would not look like a setuptools problem.
+# Keep this the LAST pip layer that can compile anything -- a later layer that builds a
+# package needing pkg_resources would fail here, and the error would not look like a
+# setuptools problem. The sqlparse override below is a pure-Python wheel, so it is exempt.
 RUN pip install --upgrade "setuptools>=83"
+
+# Force sqlparse past two declared ceilings to clear CVE-2026-54284, CVE-2026-59893,
+# CVE-2026-71491 (parser CPU-exhaustion DoS) and CVE-2026-59894 (SQL string breakout in
+# the python/php output formats). All four are fixed only in 0.6.0 -- OSV reports no
+# patched 0.5.x -- so no in-range version is clean and the resolver cannot help us:
+#     collate-sqllineage 2.1.4                     sqlparse==0.5.4
+#     dbt-core (transitive via collate-data-diff)  sqlparse<0.6.0
+# Both ceilings are stale rather than substantive. collate-sqllineage 2.1.5 shipped with
+# sqlparse==0.6.0 and 2.1.6 reverted only the pin to stay co-installable with dbt-core --
+# the two releases are byte-identical apart from the version string, so 0.6.0 is a version
+# upstream already released against. dbt-core's ceiling predates 0.6.0 by nine months and
+# is tracked at https://github.com/dbt-labs/dbt-core/issues/15988. Once that lands, delete
+# this layer and raise the floors in ingestion/setup.py instead.
+#
+# --no-deps because pip would otherwise backtrack on the declared conflict. `pip install`
+# exits 0 while printing the resolver-conflict ERROR, and `pip check` will report the two
+# unsatisfied pins for the life of the image, so the import gate is what actually keeps
+# this honest. It asserts the two things that would otherwise fail silently: that we really
+# got 0.6.x, and that collate-sqllineage's monkeypatch of sqlparse internals still bites.
+# That patch raises MAX_GROUPING_DEPTH/MAX_GROUPING_TOKENS 100x and retypes STRING as a
+# builtin; if a future sqlparse renames either, the patch degrades to a no-op and lineage
+# comes back quietly truncated with nothing failing.
+RUN pip install --no-deps "sqlparse==0.6.0" \
+ && python -W ignore -c "\
+import sqlparse; \
+from sqlparse.engine import grouping; \
+from sqlparse.keywords import KEYWORDS; \
+import collate_sqllineage.core.parser.sqlparse; \
+from collate_sqllineage.core.parser.sqlparse.analyzer import SqlParseLineageAnalyzer; \
+from collate_sqllineage.runner import LineageRunner; \
+assert sqlparse.__version__.startswith('0.6.'), sqlparse.__version__; \
+assert (grouping.MAX_GROUPING_DEPTH, grouping.MAX_GROUPING_TOKENS) == (10000, 1000000), 'sqllineage grouping patch is a no-op'; \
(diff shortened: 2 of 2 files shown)


ulixius9:
### Clarification on the `AS MATERIALIZED` finding, after building the image and running it

I built `ingestion/operators/docker/Dockerfile` locally (`INGESTION_DEPENDENCY="dbt,redshift,pii-processor"`, arm64) and ran it against a dev stack. The build gate passed, and the image resolves as intended:

```
sqlparse                 0.6.0
collate-sqllineage       2.1.6     (pins sqlparse==0.5.4)
dbt-core                 1.11.7    (pins sqlparse<0.6.0)
MAX_GROUPING_DEPTH/TOKENS  10000 1000000   <- sqllineage monkeypatch still applies
```

While verifying end-to-end I found my description of the `AS MATERIALIZED` improvement needs a scope qualifier. Measured in the built image, same query, same image, only sqlparse swapped:

```
sqlparse 0.6.0 -- SqlParseLineageAnalyzer:  src: [customer, orders, recent]  tgt: [cf]
sqlparse 0.5.4 -- SqlParseLineageAnalyzer:  src: []                          tgt: [cf]
```

So on the **sqlparse analyzer path** 0.5.4 silently loses every source table — worse than the `ValueError` variant I originally reported, since nothing raises.

**But** the same query through `LineageParser` with the default ANSI dialect resolves fine on 0.5.4 (3 column pairs, both sources), because sqlfluff is tried first and succeeds. So the fix only reaches users on the sqlparse fallback path — `lineage/parser.py:632` and `lineage/masker.py:202` — i.e. queries where sqlfluff fails to parse first. "Lineage was lost entirely" is accurate for that path, not for every `AS MATERIALIZED` query. The CVE remediation is unaffected either way; I did not want the compatibility upside read as broader than it is.

Also verified in the built image against a live server (`docker/development/docker-compose.yml`, server `2.0.0-SNAPSHOT`):

- Sample-data metadata workflow via the operator entrypoint (`python main.py`, `config` + `pipelineType` env): 329 sink records, **0 sink errors**, 69 tables / 4 databases / 4 schemas created.
- Usage workflow (the query-parser path — `sqlparse`'s heaviest real use): **100% across all four stages**, 0 errors. 15 queries landed with literals masked, e.g. `case when net_sales > ? then ? else ?`, and usage counts on 12 tables.
- `metadata lineage -c` wrote a real edge — `shopify.customer -> shopify.customer_features` with column-level `customer_id -> customer_id`.
- `is_safe_sql_query` still blocks `SELECT 1; DROP TABLE t`, `load_file('/etc/passwd')` and `CREATE MATERIALIZED VIEW ...` in the image.

The one workflow failure was `No module named 'avro'`, expected from the narrow extras (topics need the kafka/sample-data extra), and an earlier `fullyQualifiedName` `AttributeError` reproduced **identically with sqlparse 0.5.4 in the same image**, so it is not related to this change.

Caveat worth stating: this Dockerfile installs `openmetadata-ingestion~=2.0.0.0.dev0` from PyPI, which resolves to **2.0.0.0rc2** — so the build validates the sqlparse layer and the gate, not the working tree. `Dockerfile.ci` is the from-source path.

Khairajani (approved):

PR #31837, opened 2026-08-20
fix: ConditionalAllow permissions treated as false in getOperationPermissions

### LINKED ISSUE #31783: Domain-scoped non-admin users blocked from Services list and Ingestion Runner field despite correct backend permissions
### Affected module

UI

### Describe the bug

Two related UI permission-gating bugs prevent domain-scoped non-admin users from accessing features that the backend already permits. Direct API testing (GET/POST) confirms correct backend permission enforcement in both cases below — these are frontend-only defects.

1. Database Services list page shows "No permission"
A non-admin user with a `hasDomain()`-scoped `ViewAll` permission on `databaseService` sees "No permission" when navigating to Settings → Services → Databases, even though their domain matches services in the list.

2. Ingestion Runner dropdown never renders for non-admins
In the create/edit database service form, the "Ingestion Runner" dropdown never renders for a non-admin user, at any permission level (unconditional, +ViewBasic, hasDomain()-scoped, fully unconditional).

### To Reproduce

Bug 1:
1. Create a role with `ViewAll` on `databaseService`, scoped via a `hasDomain()` condition.
2. Assign the role to a non-admin user in that domain.
3. Log in as that user and navigate to Settings → Services → Databases.
4. Observe "No permission" error, despite `GET /permissions` returning `conditionalAllow` and matching services existing in the domain.

Bug 2:
1. As a non-admin user with create/edit permission on a database service (any grant level), open the create or edit service form.
2. Observe the "Ingestion Runner" dropdown is not shown, at any permission level.
3. Directly POST a service payload with `ingestionRunner` and `domains` set as this non-admin user — the request succeeds and both fields persist.

### Expected behavior

1. The Services list page should render and show services within the user's domain when the backend permission is `conditionalAllow`, with visibility of individual services still governed by domain-scoping at the instance level.
2. The Ingestion Runner field should render based on the user's actual permission to create/edit the service, not a hardcoded admin check.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

1.13.3

_No response_

### Additional context

_No response_

### Pre-submission checklist

zhang-arvin:
### Describe the bug

`getOperationPermissions` in `PermissionsUtils.ts` only checks `Access.Allow` when mapping permissions, treating `Access.ConditionalAllow` as `false`. This causes domain permissions with conditional allow rules to be incorrectly denied at the UI level.

### Root Cause

In `PermissionsUtils.ts` line 84:

```typescript
[curr.operation as Operation]: curr.access === Access.Allow,
```

`Access.ConditionalAllow` evaluates to `false` because it is not equal to `Access.Allow`.

### Fix

```typescript
[curr.operation as Operation]: curr.access === Access.Allow || curr.access === Access.ConditionalAllow,
```

### Related Issue

Closes #31783

This PR fixes UI permission mapping so `ConditionalAllow` grants are no longer treated as denied.
- Updates `getOperationPermissions` to recognize both `Allow` and `ConditionalAllow`.
- Preserves backend enforcement of conditions when operations are executed.
</details>

The PR appears safe to merge with no actionable correctness or security issues identified.

The change aligns resource-level UI permission mapping with the existing conditional-access state while entity-aware backend authorization continues to evaluate policy conditions before protected operations.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/src/utils/PermissionsUtils.ts | Correctly expands the operation-permission mapping to represent conditional grants as available in the UI; no actionable regression was identified. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: ConditionalAllow permissions treate..."](https://github.com/open-metadata/openmetadata/commit/cc2520add21059c286719e75bd650d1b135aed87) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55224272)</sub>

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/src/utils/PermissionsUtils.ts
+++ b/openmetadata-ui/src/main/resources/ui/src/utils/PermissionsUtils.ts
@@ -81,7 +81,7 @@ export const getOperationPermissions = (
     (acc: OperationPermission, curr: Permission) => {
       return {
         ...acc,
-        [curr.operation as Operation]: curr.access === Access.Allow,
+        [curr.operation as Operation]: curr.access === Access.Allow || curr.access === Access.ConditionalAllow,
       };
     },
     {} as OperationPermission


chirag-madlani:
ConditionalAllow can't be treated as true since we don't know where it's satisfy the applied roles and policy
not a right fix

zhang-arvin:
@chirag-madlani Thanks for the review. I understand your concern that blindly treating `ConditionalAllow` as `true` is not correct since we don't know whether the condition is actually satisfied for the current user. But I'd like to discuss this further before closing the PR, because the current behavior is also broken.

## The Problem (Issue #31783)

Domain-scoped non-admin users are **completely blocked** from the Services list and Ingestion Runner field, even though the backend correctly grants them access. The root cause is in `PermissionsUtils.ts`:

```typescript
[curr.operation as Operation]: curr.access === Access.Allow,
```

This maps `Access.ConditionalAllow` to `false`, meaning the UI treats it as "denied" — even when the backend has already determined the user should have access.

## Your Concern

> ConditionalAllow can't be treated as true since we don't know where it's satisfy the applied roles and policy

This is valid — we shouldn't blindly grant access when conditions may not be met. But I'd argue the current behavior (treating it as `false`) is equally wrong — it's a **false negative** that blocks legitimate users.

## Proposed Alternatives

I see three possible approaches:

### Option A: Treat ConditionalAllow as Allow in UI, enforce in backend
The UI is optimistic (shows buttons/lists), but the backend remains authoritative. When the user actually performs an action, the backend evaluates the condition and rejects if not met. This is a common pattern — the UI should not be stricter than the backend.

### Option B: Evaluate conditions on the client side
If the condition parameters (roles, policies) are available in the permission response, we could evaluate `ConditionalAllow` client-side. This would be the most correct but also the most complex.

### Option C: Add a separate "conditional" UI state
Instead of treating `ConditionalAllow` as either `true` or `false`, add a third state. Show the UI element but with a visual indicator that access is conditional, and let the backend enforce on action.

## Questions

1. Would you be open to **Option A** as a pragmatic fix? The backend already enforces the actual condition — the UI just shouldn't preemptively block users.
2. If not, can you provide guidance on the preferred approach? The issue (#31783) is still open and affecting real users.
3. Is there an existing mechanism in the codebase for evaluating conditions client-side that I'm missing?

Happy to iterate on the fix — just want to make sure we don't leave this issue unresolved. Would you be open to reopening the PR with a revised approach?

PR #31839, opened 2026-08-20
Fixes 28727: flatten testCaseResolutionStatusDetails index mapping + validate highlight fields at save time (Parts A & B)

### LINKED ISSUE #28727: Search follow-ups: testCaseResolutionStatusDetails mapping↔schema mismatch & save-time highlight-field validation
## Summary

Two related search follow-ups surfaced while reviewing PR #28674 (drop flattened/`flat_object` highlight fields) and PR #28711 (`dynamic:false` on all index mappings). Neither is fixed by those PRs — #28674/#28711 contain *runtime guards / hardening*; the items below are the underlying correctness gaps that should be closed at the source.

---

## A. `testCaseResolutionStatusDetails` index mapping does not match the schema (dead `resolved.*` mapping + dead search boost)

**Problem.** In `test_case_resolution_status_index_mapping.json` (all locales), `testCaseResolutionStatusDetails` maps a nested **`resolved`** sub-object:

```
testCaseResolutionStatusDetails.properties.resolved.properties.{testCaseFailureReason, testCaseFailureComment, resolvedBy}
```

But the schema `testCaseResolutionStatusDetails` is `oneOf [assigned.json, resolved.json]`, and `resolved.json` declares those fields **directly** — there is no `resolved` wrapper:

- `openmetadata-spec/src/main/resources/json/schema/tests/resolved.json` → `testCaseFailureReason`, `testCaseFailureComment`, `resolvedBy` (the first two are `required`).

So a real `Resolved` document is shaped `testCaseResolutionStatusDetails: { testCaseFailureReason, testCaseFailureComment, resolvedBy }`, which never matches the mapped `…resolved.*` path.

**Consequences (current `main`):**
- The configured search boost `testCaseResolutionStatusDetails.resolved.testCaseFailureComment` (`TestCaseResolutionStatusIndex.java:81`, boost `10.0`) targets a path that no document has → **dead no-op**; resolution-comment text is not actually boosted/searchable as intended.
- The only subfield queried that *works* is `testCaseResolutionStatusDetails.assignee.name` (`SearchListFilter.java:425`) — `assignee` is mapped correctly and lines up with `assigned.json`.
- The real flat fields (`…testCaseFailureReason` / `…testCaseFailureComment`) are only indexed today via the nested `"dynamic": true` on `testCaseResolutionStatusDetails`. **PR #28711 flips that to `false`** (correctly — nothing queries them), after which those fields are stored in `_source` but **not indexed at all**.

**Proposed fix (pick one, consistently across the 4 locale mapping files + indexer):**
1. Correct the mapping to the flat schema shape: replace the `resolved` sub-object with `testCaseFailureReason` / `testCaseFailureComment` / `resolvedBy` directly under `testCaseResolutionStatusDetails`, and update the boost in `TestCaseResolutionStatusIndex.java` to `testCaseResolutionStatusDetails.testCaseFailureComment`; **or**
2. If a `resolved` wrapper is actually desired, change the indexer to nest the `Resolved` payload under `resolved` when building the doc.

**Acceptance criteria:** searching/boosting on resolution failure-comment text returns the expected `Resolved` records on a freshly reindexed cluster; mapping path used by the boost matches the indexed document shape; covered by a test.

---

## B. Validate highlight fields at config/save time (not only at query time)

**Problem.** PR #28674 added a **runtime** guard that silently drops highlight fields targeting `flattened`/`flat_object` paths (`SearchSourceBuilderFactory.FLATTENED_FIELD_PATHS` + `isHighlightUnsafeField` / `filterHighlightSafeFields`, applied in both `*SourceBuilderFactory.buildHighlightsV2`). That unblocks the `all shards failed` / `has no associated analyzer` 500s, but the bad value still lives in Search Settings and is only neutralized at query time.

**Proposed fix.** Validate highlight fields when Search Settings are saved (the settings persistence path — `SettingsCache` / the search-settings update endpoint; called `SearchSettingsResource` in #28674's description) and **reject or warn** when a configured `highlightField` targets a flattened path, reusing the existing classifier `SearchSourceBuilderFactory.isHighlightUnsafeField`. Add the matching UI affordance so the misconfiguration is caught at config time.

**Acceptance criteria:** saving Search Settings with a highlight field on a flattened path (e.g. `extension.foo`, `columns.children.x`) is rejected/warned with a clear message; the runtime guard from #28674 remains as defense-in-depth; covered by a test.

---

_Filed as a follow-up to #28674 and #28711._

TeddyCr:
### Describe your changes:

Fixes #28727

**Both parts of #28727 — this closes the issue.** Issue #28727 has two parts and this PR now carries both.

- **Part A** — the `testCaseResolutionStatusDetails` mapping↔schema mismatch (dead `resolved.*` mapping + dead search boost).
- **Part B** — validating highlight fields at Search Settings save time, plus the matching UI affordance.

**A correction to the issue text for reviewers:** #28727 says Part B should reuse `SearchSourceBuilderFactory.isHighlightUnsafeField`, added by #28674. That symbol never existed on `main`. The equivalent query-time guard landed separately via **#29073** (`7738765de58`) as `OpenSearchSourceBuilderFactory.isFlattenedExtensionField` — OpenSearch-only and hardcoded to `extension`. #28674 was later rewritten into a config-driven version of Part B and **closed unmerged on 2026-08-24**, so the `highlight` flag it introduced never landed. Part B here is built on `main` and reuses the *concept*, not the (nonexistent) symbol.

I flattened the `testCaseResolutionStatusDetails` index mapping so it matches its JSON Schema, and repointed the resolution-comment search boost at the path documents actually carry.

The mapping declared a `resolved` wrapper object that the schema makes **structurally impossible**. `testCaseResolutionStatus.json` models the field as `oneOf [assigned.json, resolved.json]`, and both branches are `"additionalProperties": false` — `resolved.json` declares `testCaseFailureReason` / `testCaseFailureComment` / `resolvedBy` **directly**. So a `resolved` key could never legally appear in a document, and indeed never did: doc building is generic (`SearchIndex` → `JsonUtils.getMap(entity)`; `TestCaseResolutionStatusIndex` only adds `fqnParts`, `@timestamp`, and parent relationships). The consequence was a dead 10× boost — `testCaseResolutionStatusDetails.resolved.testCaseFailureComment` in `TestCaseResolutionStatusIndex.getFields()` matched **nothing, in every language**.

`"dynamic": true` on that node was masking the bug: it indexed the real flat fields anyway, but as standard-analyzer `text` rather than the declared types. That broke exact match on `testCaseFailureReason` (analyzed text never matches an exact term) and broke CJK/Cyrillic comment search (the standard analyzer splits Japanese/Chinese into single characters and does not stem Russian).

#
### Type of change:
- [x] Bug fix

#
### High-level design — Part A

**Approach — the issue's option 1: correct the mapping to the schema, don't reshape the document.** The schema and the document are already flat and are the source of truth; only the mapping was wrong. Reshaping the document to match the mapping would mean writing a wrapper key that `additionalProperties: false` forbids.

**Files changed (7):**

| File | Change |
|---|---|
| `openmetadata-spec/.../elasticsearch/{en,jp,ru,zh}/test_case_resolution_status_index_mapping.json` | Move `testCaseFailureReason` / `testCaseFailureComment` / `resolvedBy` up one level, out of the `resolved` wrapper. Pure unwrap — no other edit. |
| `openmetadata-service/.../search/indexes/TestCaseResolutionStatusIndex.java` | Boost repointed to `testCaseResolutionStatusDetails.testCaseFailureComment` (one line). |
| `openmetadata-service/src/test/.../IndexMappingNestedFieldConsistencyTest.java` | New mapping↔schema invariant. |
| `openmetadata-integration-tests/.../SearchConsumerFieldBehaviorIT.java` | Resolved-variant document + 3 behavioral tests. |

There is no separate OpenSearch resource tree — the same four files serve both engines, with `OsUtils.enrichIndexMappingForOpenSearch` rewriting at runtime.

**Per-locale analyzers deliberately preserved, not homogenized.** en/ru keep `om_analyzer`; jp keeps `om_analyzer_jp` (kuromoji) **and** its `fields.ngram`; zh keeps `ik_max_word` **and** its `fields.ngram`. Verified byte-for-byte: en ≡ ru in this subtree, and unwrapping the base `resolved` object with `jq` yields output identical to this branch in all four locales.

**`"dynamic": true` kept — deliberate.** After flattening, every property of both `oneOf` branches is mapped explicitly, so `dynamic` is no longer load-bearing for correctness. It is retained as the forward-compatibility net: a property added to `assigned.json` / `resolved.json` still gets indexed rather than silently dropped, and the new unit test fires immediately in that case. Flipping it to `false` is a separate hardening decision and would couple a silent-data-loss behavior change to a mapping correction.

**Alternative rejected:** rewriting the document in `TestCaseResolutionStatusIndex` to emit a `resolved` wrapper. That would contradict the schema, break the UI and Python (both already read/write the flat shape), and invalidate the existing `testCaseResolutionStatusDetails.assignee.name` filter pattern.

---

### High-level design — Part B

**Problem.** Search Settings accepted a highlight field the index mapping can never highlight. Two mapped shapes cannot highlight: `flattened` (rewritten to `flat_object` for OpenSearch by `OsUtils`) has no analyzer and **fails the highlight shard — a 500 on the search**; `enabled: false` is kept in `_source` but never indexed, so a highlight on it silently matches nothing. The bad value was only neutralised at query time and still lived in the settings.

**Approach — derive from the mappings, don't maintain a list.** `HighlightFieldClassifier` walks the index-mapping resources once and classifies a dotted path as `SUPPORTED`, `NOT_INDEXED`, or `NOT_ANALYZABLE`. This same knowledge previously lived as hardcoded path lists that had to be restated on every mapping change — see the duplicated `STALE_FLATTENED_CHILDREN_FIELDS` in the **v1130 and v11210** migration utils — and drifted each time.

A field that is **absent** from the mapping stays supported: the highlighter skips unmapped fields without erroring, so rejecting them would break configurations that merely name a field an index does not carry.

One classifier backs three call sites:

| Call site | Role |
|---|---|
| `SearchSettingsHandler.validateHighlightFields` | rejects the value on save (400) |
| `SearchSettingsHandler.annotateHighlightableFields` | marks `allowedFields` so the UI only offers a toggle that works |
| `OpenSearchSourceBuilderFactory.buildHighlightsV2` | query-time guard, replacing `isFlattenedExtensionField` |

**This closes a live gap.** The guard it replaces matched only `extension`/`extension.*`. `extension` is now `{"type":"object","enabled":false}` in 35 mappings, while the one genuinely `flattened` field left — `aiGovernance.complianceStatus` (`ai_application`, `llm_model`, `mcp_server`) — **was not covered**. The classifier catches it; `flattenedFieldIsNotAnalyzable` asserts so. The old `extension` behaviour is preserved deliberately and asserted by `queryTimeGuardStillDropsExtensionAsBefore`, because clusters whose indexes were created while `extension` was still `flattened` keep that mapping until a reindex.

**Validation hooks the REST payload, not `mergeSearchSettings` — deliberate.** `mergeSearchSettings` is the obvious home and is a trap: `SettingsCache.initialize` catches only `IOException` (`SettingsCache.java:205`), so throwing from the merge path would **fail startup** on any cluster already carrying a bad value. The check runs on the incoming payload in `SystemResource.createOrUpdateSetting`, alongside the existing `validateGlobalSettings`.

**The UI flag is derived, not seeded.** The closed #28674 hand-typed `"highlight": true` into ~200 seed entries and added migration v11211 to scrub stored values against it. Here the flag is computed from the same classifier the validator uses, and the seed is untouched — so what the UI offers and what the API accepts cannot disagree, and no migration is needed. `annotatedFlagAgreesWithWhatTheSavePathAccepts` walks every shipped allowed field and asserts the two agree.

**Part B needs no reindex and no migration** — no mapping changes, no stored-data changes. The reindex note below applies to **Part A only**.

**Known limitation, deliberately out of scope.** `PATCH /v1/system/settings/{settingName}` (`SystemResource.java:765`) writes straight to the repository with no search-settings validation. It bypasses this check — and already bypasses the pre-existing `validateGlobalSettings`, so `keywordWeight: 5` can be set through it today. Pre-existing hole affecting all settings validation; the query-time guard remains as defense-in-depth for values arriving that way.

#
### ⚠️ Upgrade note — this requires a reindex

**Applies to Part A only.** **This change requires a reindex of `test_case_resolution_status_search_index`. `migrate` alone is not sufficient and will log a mapping-update failure for that index.** (Part B requires neither a reindex nor a migration.)

`migrate` reaches `searchRepository.updateIndexes()` (`OpenMetadataOperations.java:1196`), which issues `PUT _mapping`. On any cluster that has ever resolved an incident, `dynamic: true` has already created those fields as `text`, so declaring them `keyword` is an incompatible type change and the request is **rejected 400**:

```
illegal_argument_exception: mapper [testCaseResolutionStatusDetails.<field>]
cannot be changed from type [text] to [keyword]
```

Three things worth knowing:

1. **The rejection is atomic** — *no* property in that request lands, not just the four here. Any unrelated mapping change shipped for this index in the same release is discarded with it. (Verified by adding an unrelated probe field to the same request and confirming it did not land either.)
2. **Which field the engine names first is engine-dependent** — OpenSearch 3.4.0 reported `…testCaseFailureReason`, Elasticsearch 9.3.0 reported `…resolvedBy.id` for the identical request. Same 400, same atomicity; a different field name is not a different bug.
3. **The cause is swallowed on the OpenSearch path** — `OpenSearchIndexManager` logs `"Failed to Update Open Search index {}"` without the exception, so an operator sees one bare error line and a silently stale mapping.

The supported path is `./bootstrap/openmetadata-ops.sh reindex` (no trailing period — it is parsed as part of the subcommand). It is also the *only* path that drops the stale `resolved` sub-object, since `PUT _mapping` cannot remove a field. `IndexMappingVersionTracker` hashes each mapping, so the smart-reindex plan picks up `testCaseResolutionStatus` and recreates just that index — **no DB migration is involved**.

**Mitigating point:** because `"dynamic": true` was retained, un-reindexed clusters still get a *partial* improvement immediately — the repointed boost now targets a path dynamic mapping has already created (as standard-analyzer text) rather than a path with no values at all. Full correctness (declared `keyword`, per-locale analyzers) arrives with the reindex.

#
#### Use cases covered
- Searching the Incident Manager index for a word that appears only in an incident's **resolution comment** returns that incident (previously returned nothing, in every language).
- The **Resolved** branch of `testCaseResolutionStatusDetails` is indexed and searchable at all — previously no test anywhere indexed a Resolved-variant document.
- Comment search works in **Japanese, Chinese and Russian**, not just English, because the per-locale analyzer is applied instead of the standard-analyzer fallback.
- `testCaseFailureReason` supports exact-match term lookup, i.e. its declared `keyword` type is actually in effect.
- Existing Assigned-branch behavior (status filter, `assignee.name` filter, test-case/origin-entity filters) is unchanged.

#### Unit tests
- Files updated: `openmetadata-service/src/test/java/org/openmetadata/service/search/IndexMappingNestedFieldConsistencyTest.java`
- `resolutionStatusDetailsMappingMustMatchSchemaProperties` asserts the mapped subfield set equals the union of the `properties` of every `oneOf` branch, **read from the JSON Schema at runtime** rather than hardcoded — so adding a property to `assigned.json` / `resolved.json` fails this test until the mappings follow. It guards the class of bug, not just this instance.
- Result: `Tests run: 7, Failures: 0, Errors: 0, Skipped: 0` (1 failure before the fix).

**Coverage on the changed class — measured, and honestly short of the 90% target.** Ran `mvn -P static-code-analysis -pl openmetadata-service test jacoco:report` scoped to the search tests. `TestCaseResolutionStatusIndex.java` line coverage is **62.8% (27/43)**:

| Method | Lines | |
|---|---|---|
| `<init>` | 1/1 | 100% |
| `buildSearchIndexDocInternal` | 5/5 | 100% |
| `getEntityTypeName` | 1/1 | 100% |
| `setParentRelationships` | 20/28 | 71.4% |
| `getEntity` | 0/1 | 0% |
| **`getFields`** — the method this PR changes | **0/7** | **0%** |

The 0% on `getFields` is a **measurement artifact, not an untested line**: `getFields()` is executed directly by the new integration test (`incidentResolutionCommentSearchReturnsIncidentInAllLanguages` builds its query from `TestCaseResolutionStatusIndex.getFields()`), and that test lives in `openmetadata-integration-tests` — a separate Maven module whose execution this `-pl openmetadata-service` jacoco run does not observe. The changed line is covered by an automated test that fails without the fix; it is simply not covered by a *unit* test in the measured module. The remaining gaps (`setParentRelationships`, `getEntity`) are pre-existing and untouched by this PR.

- [x] I added integration tests in `openmetadata-integration-tests/`.
- File updated: `openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/SearchConsumerFieldBehaviorIT.java`

This class boots a **real OpenSearch 3.4.0 testcontainer** with analysis-kuromoji and analysis-ik, creates the real mapping for **every** language in `IndexMappingLanguage` (en/jp/ru/zh), indexes real documents and runs the actual consumer queries. I added a Resolved-variant document alongside the existing Assigned one, plus three tests:

| Test | What it asserts | Failed before the fix? |
|---|---|---|
| `incidentResolutionCommentSearchReturnsIncidentInAllLanguages` | The production consumer query — a `most_fields` / `operator:and` `multi_match` **built from `TestCaseResolutionStatusIndex.getFields()`** — returns the incident when searching a word unique to its comment. Because the field list comes from production code, repointing a boost at a dead path fails this test. | **Yes** — `is broken in language(s): [en, jp, zh, ru]`, 0 hits everywhere |
| `failureReasonSupportsExactMatchTermInAllLanguages` | A `term` on `testCaseResolutionStatusDetails.testCaseFailureReason` returns the incident, i.e. the declared `keyword` is in effect. A **type probe, not a product feature** — nothing filters on the reason today; it is what distinguishes "explicitly mapped" from "accidentally indexed by `dynamic:true`". | **Yes** — `is broken in language(s): [en, jp, zh, ru]` |
| `resolutionCommentUsesLanguageAnalyzerInAllLanguages` | `_analyze` on the comment field yields each locale's expected token, proving the per-locale analyzer survived the flatten. | **Yes** — `ru expected 'продаж' but got [ежемесячные, продажи]`, `zh expected '销售' but got [北, 京, 大, 学, 的, 销, 售, 报, 表]`, `jp expected '東京' but got [東, 京, タワー, の, 売, 上]` |

- Result: `Tests run: 20, Failures: 0, Errors: 0, Skipped: 0` (3 failures before the fix; the 17 pre-existing tests passed throughout, so the added document perturbs nothing).

**Elasticsearch — verified by hand only.** The automated IT runs OpenSearch 3.4.0. On a real **ES 9.3.0** container the branch mapping creates cleanly, the flat Resolved document indexes, and both the boost query and the `testCaseFailureReason` term filter return **1 hit** — against **0 hits** on the base mapping. This was done manually during review; **no automated ES coverage exists for this index**, so a future ES-only regression would not be caught by CI.

- [x] Not applicable — no ingestion changes. `ingestion/.../sample_data.py` already writes the flat shape and needed no change.

#### Part B tests

**Unit — `HighlightFieldClassifierTest` (8 tests, new).** Reads the real mapping resources, so each assertion is a statement about the shipped mappings: `extension` → `NOT_INDEXED`; `extension.someCustomProperty` and `columns.children.name` inherit their parent's verdict; `aiGovernance.complianceStatus` → `NOT_ANALYZABLE` (the case the old guard missed); `name`/`description`/`columns.name` → `SUPPORTED`; an unmapped field and an unknown asset type stay supported; `extension`/`extension.*` still dropped by the query-time guard.

**Unit — `SearchSettingsHandlerTest` (6 added, 20 total).** Rejects a non-indexed and a flattened highlight field with a message naming the field and the reason; accepts an analyzed one; the shipped seed passes the validator; the query-time union guard drops none of the shipped highlight fields; the annotated flag agrees with what the save path accepts.

**Backend IT — `HighlightFieldSaveValidationIT` (4 tests, new).** `servedAllowedFieldsCarryTheHighlightFlag` asserts `highlight=true` on the real `GET /system/settings/searchSettings` response. It guards the *wiring*: the unit tests call `annotateHighlightableFields` directly, so they pass even when nothing in production calls it — which is exactly the regression that reached review. The remaining three: PUTs to the real `/v1/system/settings` and asserts **400** for `extension.foundry_rid` (message names the field and says *"not indexed"*) and for `dataModel.columns.children.name`, and **2xx** for `description` — so the negatives cannot pass merely because the endpoint rejects any `highlightFields` edit.

**Two existing ITs updated.** `ExtensionHighlightSearchIT` and `FlattenedChildrenHighlightSearchIT` configure exactly the values now rejected on save, to prove the query-time guard tolerates them. Their PUT would now 400, so they write the bad value **through the repository** instead — which is also truer to what they test: a value stored *before* this check existed. Their query-time assertions are unchanged.

#### Playwright (UI) tests
- [x] **Added** — `SearchSettings.spec.ts` → *"Highlight toggle is offered for analyzed fields"*, which drives the real Search Settings page against a real server and asserts the toggle renders **enabled**. This is the level the regression above lived at: the Jest test supplies `entityFields` as props, so it can never catch a backend that stops annotating the flag. `yarn lint:playwright` passes with **0 errors** and no new warnings.
- [x] ~~Not applicable — Jest covers the change.~~ Part A has no UI changes. Part B's UI change is one control's disabled state, covered by `FieldConfiguration.test.tsx` (**5 passed**), including a new case asserting the toggle is disabled and the change handler is not called. This is component-level state, not a user journey across pages, so it does not warrant an E2E test.

##### Part A — mapping flatten and boost

Steps 1 and the reindex/mapping inspection below are the reproducible procedure; **steps marked _(not verified)_ were not executed against a running OpenMetadata stack** — the equivalent effects are covered by the automated IT against a real engine, and the `PUT _mapping` rejection in step 3a was reproduced directly against both engines rather than through `openmetadata-ops.sh`.

> Commands below are in fenced blocks on purpose — copy them with the block's copy button. An inline
> command followed by a sentence period is easy to select with the period attached, and
> `openmetadata-ops.sh reindex.` fails with `Unmatched argument at index 2: 'reindex.'`.

1. Build and install:
   ```bash
   mvn install -pl openmetadata-spec -DskipTests
   mvn install -nsu -pl openmetadata-service -DskipTests
   ```
2. _(not verified)_ Bring up a stack, then log in as admin:
   ```bash
   docker compose -f docker/development/docker-compose.yml up -d
   ```
3. Reindex:
   ```bash
   ./bootstrap/openmetadata-ops.sh reindex
   ```
   Expect `Index mapping changed for entity: testCaseResolutionStatus` and only that entity in the plan; a second `reindex` reports no changes.
   - 3a. Confirm the `migrate`-only path fails as described in the upgrade note above (400, atomic, cause swallowed on OpenSearch).
4. Confirm the live mapping is flat and correctly typed:
   ```
   curl -s 'http://localhost:9200/test_case_resolution_status_search_index/_mapping' \
     | jq '..|.testCaseResolutionStatusDetails? // empty | .properties | keys'
   ```
   Expect `["assignee","resolvedBy","testCaseFailureComment","testCaseFailureReason"]` and **no** `resolved`. Then check `testCaseFailureComment` carries `"type":"text"` with the locale's analyzer (`om_analyzer` en/ru, `om_analyzer_jp` jp, `ik_max_word` zh) and, for jp/zh, a `fields.ngram` subfield.
5. _(not verified)_ Produce a resolved incident: Data Quality → a failing test case → Incident Manager → status **Resolved**, reason `FalsePositive`, comment `flakyupstream feed was backfilled`.
6. _(not verified)_ See the boost take effect — query by a word appearing only in the comment:
   ```
   curl -s -u admin:admin \
     'http://localhost:8585/api/v1/search/query?index=test_case_resolution_status_search_index&q=flakyupstream&from=0&size=10'
   ```
   Pre-fix `hits.total.value: 0`; post-fix the incident is returned. Re-run with `&explain=true` and confirm the winning clause is `testCaseResolutionStatusDetails.testCaseFailureComment^10.0`, ranking a comment match above a `testCaseResolutionStatusType^1.0` match.
7. _(not verified)_ Confirm no Assigned-branch regression: the Incident Manager list still filters by assignee (`SearchListFilter` → `testCaseResolutionStatusDetails.assignee.name`) and the incident detail panel still renders reason/comment/resolvedBy.

##### Part B — save-time validation and the UI toggle

Continuing on the same stack. **No reindex or migration is needed for these steps.**

8. **The `highlight` flag is served.** It is derived, not stored — absent from `searchSettings.json` on disk, present in the payload:
   ```
   curl -s -u admin:admin 'http://localhost:8585/api/v1/system/settings/searchSettings' \
     | jq '.config_value.allowedFields[] | select(.entityType=="table") | .fields[]
           | select(.name=="description" or .name=="name") | {name, highlight}'
   ```
   Expect `"highlight": true` on both. The flag is derived, not stored — it is absent from
   `searchSettings.json` on disk and appears only in the served payload.

   > **Note the snake_case.** The settings API is `config_type` / `config_value` on the wire, not
   > camelCase. `.configValue` returns `null` and jq prints nothing — which reads as "looks fine"
   > rather than as an error.

   This step earns its place: it caught a real regression during review. The flag was originally
   annotated in `readDefaultSearchSettings`, which is **not** the path that produces the served
   payload — `SettingsCache.initialize` writes the store from the seed file directly and never
   annotated it. Every field came back as the schema default `false`, and because the UI reads
   `entityFieldConfig?.highlight ?? false`, **every** highlight toggle rendered disabled. It is now
   annotated in `getSettingByName`, so the served flag is correct regardless of which path wrote the
   stored settings — including clusters that persisted them before the flag existed.

9. _(not verified)_ **No existing toggle changes state** — the important regression check. Go to **Settings → Preferences → Search Settings → Table** (`/settings/preferences/search-settings/table`) and expand `name`, `description`, `columns.name`. Every Highlight toggle is still interactive. **Zero** shipped search fields are unhighlightable, so this change must be invisible on a default install.

10. _(not verified)_ **The disabled toggle.** Since a default install shows none, add one: on the same page add a search field named `extension.foundry_rid` (it renders with the blue **Custom** badge), then expand it. Its Highlight toggle is **greyed out and unclickable**, and hovering shows *"This field cannot be highlighted. Only analyzed text fields support highlighting."* Contrast with `description` on the same page, which stays interactive.
    Why: `extension` is `{"type":"object","enabled":false}` in `table_index_mapping.json`, so nothing beneath it is indexed and a highlight there could never match.

11. **The save path rejects it** — the toggle is an affordance, the 400 is the enforcement. For API clients that bypass the UI:
    ```
    curl -s -u admin:admin 'http://localhost:8585/api/v1/system/settings/searchSettings' \
      | jq '.config_value.assetTypeConfigurations |= map(
              if .assetType=="container"
              then .highlightFields += ["extension.foundry_rid"] else . end)' \
      | jq '{config_type:"searchSettings", config_value:.config_value}' \
      | curl -s -u admin:admin -X PUT -H 'Content-Type: application/json' \
          -d @- 'http://localhost:8585/api/v1/system/settings' -w '\nHTTP %{http_code}\n'
    ```
    Expect **HTTP 400**:
    ```
    Highlight field 'extension.foundry_rid' for asset type 'container' is not indexed
    (enabled:false in the index mapping), so it can never be highlighted. Remove it from highlightFields.
    ```
    Repeat with `dataModel.columns.children.name` for the same 400, and with `description` for **HTTP 200**.

12. _(not verified)_ **Search still works.** Search any container or table from Explore; results and highlighting are unchanged.

**Gates run for Part B:**

| Gate | Result |
|---|---|
| `mvn test -nsu -pl openmetadata-service -Dtest='org.openmetadata.service.search.**.*Test,SearchSettingsHandlerTest'` | `Tests run: 2246, Failures: 0, Errors: 0, Skipped: 0` |
| `mvn spotless:check -nsu -pl openmetadata-service,openmetadata-integration-tests` | `0 needs changes to be clean` in both, `BUILD SUCCESS` |
| `mvn test-compile -nsu -pl openmetadata-integration-tests` | `BUILD SUCCESS` |
| `yarn test src/components/SearchSettings/FieldConfiguration` | `Tests: 5 passed, 5 total` |
| `yarn ui-checkstyle:changed` | `0 errors`; `✔ No new Antd imports or .less files` |
| `npx tsc --noEmit` | 572 pre-existing errors repo-wide, **0 in any file this PR touches** (not a CI gate; checkstyle excludes tsc) |

**RED checks — each fix proven to fail without it:**

- Stubbing the `throw` in `validateHighlightFields` behind `if (false)` → `saveIsRejectedForNonIndexedHighlightField` and `saveIsRejectedForFlattenedHighlightField` both fail with *"Expected `SystemSettingsException` to be thrown, but nothing was thrown."*
- Forcing `isHighlightAllowed = true` in `FieldConfiguration.tsx` → `Should disable the highlight toggle…` fails (`1 failed, 4 passed`). Restored → `5 passed`.

**Not executed:** `HighlightFieldSaveValidationIT` and the two updated ITs are **compile-verified only** — they need an OpenSearch container and a booted app. Steps 9, 10 and 12 were likewise not run against a live stack.

**Gates run on the rebased branch for Part A** (spec + service installed from the working tree first):

| Gate | Result |
|---|---|
| `mvn spotless:check -nsu -pl openmetadata-service,openmetadata-integration-tests` | `0 needs changes to be clean` in both modules, `BUILD SUCCESS` |
| `mvn test -nsu -pl openmetadata-service -Dtest='org.openmetadata.service.search.**.*Test,SearchListFilterTest'` | `Tests run: 2231, Failures: 0, Errors: 0, Skipped: 0` |
| `mvn test -nsu -pl openmetadata-integration-tests -Dtest=SearchConsumerFieldBehaviorIT` | `Tests run: 20, Failures: 0, Errors: 0, Skipped: 0` |

#

**Part A contains no UI changes.** Part B changes one component: `FieldConfiguration.tsx` (Settings → Preferences → Search Settings → *entity*).

The change is a single control's state. On a default install **nothing changes** — none of the search fields shipped in `searchSettings.json` are unhighlightable (asserted by `queryTimeGuardKeepsEveryShippedHighlightField`), so every existing Highlight toggle stays interactive. The new state is only reachable by adding a custom-property search field:

| Field | Highlight toggle |
|---|---|
| `description`, `name`, `columns.name` (shipped) | interactive, unchanged |
| `extension.foundry_rid` (custom property, blue **Custom** badge) | greyed out, with tooltip *"This field cannot be highlighted. Only analyzed text fields support highlighting."* |

**Screen recording not attached** — the visual delta is one disabled switch plus a tooltip, and it is covered by `FieldConfiguration.test.tsx` (`Should disable the highlight toggle for a field that cannot be highlighted`, which asserts both `toBeDisabled()` and that the change handler is not called). Step 5 of the manual testing section is the exact click-path to reproduce it. Happy to record one if a reviewer wants it.

#
### Checklist:
- [x] For JSON Schema changes: **one JSON Schema changed, in Part B.** `configuration/searchSettings.json` gains an optional `highlight` boolean on `allowedFields[].fields[]`. Generated output was regenerated, not hand-edited: Java POJOs via `mvn clean install -pl openmetadata-spec` (build-time, not committed), and the three committed TS types (`configuration/searchSettings.ts`, `settings/settings.ts`, `api/search/previewSearchRequest.ts`) via `json2ts.sh` — followed by `yarn license-header-fix`, because quicktype strips the Apache header. The committed TS diff is purely additive (+21/−0). `make generate` (Pydantic) is not required: nothing in `ingestion/` reads `allowedFields`. **No SQL migration.** Part A's mapping drift is tracked by content hash in `IndexMappingVersionTracker` and resolved by reindex (see the upgrade note); Part B changes no stored data — the `highlight` flag is derived at read time. Precedent for the mapping resources: 14b3a57207 (#30501) changed 20 mapping JSONs with no SQL and no version bump.
- [x] For UI changes: one component (`FieldConfiguration.tsx`) plus one i18n key, **translated in all 19 non-English locales** rather than left as the English placeholder `yarn i18n` inserts. `yarn ui-checkstyle:changed` passes with **0 errors**, including `tw-guard` — the tooltip uses `Tooltip`/`TooltipTrigger` from `@openmetadata/ui-core-components`, since adding `Tooltip` to the file's existing `antd` import would have been flagged as new Ant Design debt.

- [x] I have added a test that covers the exact scenario we are fixing. All four new tests fail on `main` and pass here; the RED output is quoted in the tables above.

The PR aligns test-case resolution-status mappings with the schema and introduces mapping-derived highlight-field validation and UI affordances.
- Flattens resolution-detail mappings and updates the resolution-comment search field.
- Validates unsupported highlight fields while preserving safe handling of legacy stored settings.
- Annotates settings responses with highlight support and disables unsupported UI toggles.
- Adds backend, integration, component, and Playwright coverage.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/search/HighlightFieldClassifier.java | Builds mapping-derived highlight-support classifications for validation, response annotation, and query-time filtering. |
| openmetadata-service/src/main/java/org/openmetadata/service/resources/system/SearchSettingsHandler.java | Adds highlight validation, legacy-value cleanup, and response-facing field annotation. |
| openmetadata-service/src/main/java/org/openmetadata/service/resources/system/SystemResource.java | Applies validation on settings updates and consistently annotates search-settings responses. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/IndexMappingProperties.java | Introduces reusable traversal of dotted index-mapping property paths. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OpenSearchSourceBuilderFactory.java | Replaces the hardcoded extension exclusion with the shared mapping-derived query guard. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TestCaseResolutionStatusIndex.java | Repoints resolution-comment search boosting to the document’s actual flat field. |
| openmetadata-spec/src/main/resources/json/schema/configuration/searchSettings.json | Adds the optional derived highlight-support flag to allowed search fields. |
| openmetadata-ui/src/main/resources/ui/src/components/SearchSettings/FieldConfiguration/FieldConfiguration.tsx | Disables highlighting for fields the server identifies as unsupported and explains the restriction. |
| openmetadata-spec/src/main/resources/elasticsearch/en/test_case_resolution_status_index_mapping.json | Flattens resolution-detail properties to match the schema and indexed document shape. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/search/HighlightFieldSaveValidationIT.java | Covers response annotation and acceptance or rejection of highlight-field settings across REST paths. |

</details>

```mermaid
flowchart LR
  M[Index mappings] --> C[HighlightFieldClassifier]
  C --> V[Save-time validation]
  C --> R[Response annotation]
  C --> Q[Query-time guard]
  R --> U[Search Settings UI]
  V --> P[Persist valid settings]
  Q --> S[Safe search highlights]
```
</details>

<sub>Reviews (7): Last reviewed commit: ["Fixes #28727: derive the highlight flag ..."](https://github.com/open-metadata/openmetadata/commit/89fcd7f955ff7f836d2290ad07e5b7dbfa118539) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55232573)</sub>

<details><summary><h4>Context used (5)</h4></summary>

- Knowledge Base — [Service API resources](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/service-api-resources.md)
- Knowledge Base — [Metadata web application](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/metadata-web-application.md)
- Knowledge Base — [Metadata contracts and clients](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/metadata-contracts-clients.md)
</details>


shrabantipaul-collate (dismissed):

aniketkatkar97 (approved):

manerow (commented):
Pulled the branch down and ran it locally. The highlight toggles break as soon as you hit Save.

To reproduce on a stock install: Settings → Preferences → Search Settings → Table, expand any field (`description` will do), change something, click Save. Every Highlight toggle on the page goes grey with *"This field cannot be highlighted"*. Reload and they're fine again. Restore Defaults does the same thing.

The backend meanwhile still says the field is both highlighted and highlightable:

```
table.highlightFields                        = [name, displayName, description]
allowedFields[table][description].highlight  = true
```

So the UI is showing the user the opposite of what the server actually thinks.

What's happening is that `highlight` only gets derived inside `getSettingByName`. Five endpoints can hand back searchSettings, and I checked each one:

| | `highlight` |
|---|---|
| `GET /system/settings/searchSettings` | `true` |
| `GET /system/settings` | `false` |
| `PUT /system/settings` | `false` |
| `PUT /system/settings/reset/searchSettings` | `false` |
| `PATCH /system/settings/searchSettings` | `false` |

The one that bites is the save. `mergeSearchSettings` throws away the annotated list and swaps in the seed's copy, which has no `highlight` key at all:

```java
// SearchSettingsHandler.java:464
incomingSearchSettings.setAllowedFields(defaultSearchSettings.getAllowedFields());
```

and the UI takes that PUT response straight into app state (`EntitySearchSettings.tsx:290`, and `:579` for Restore Defaults), so from then on every field reads as unhighlightable. Worth noting `SearchSettingsPage.handleResetToDefault` already avoids this by ignoring the response and re-fetching.

For the fix I'd lean on what the codebase already does for `href`, which has the same shape as `highlight`: derived by the server, response-only, and clients hand it back. `EntityResource` deals with that by stamping it at every response boundary, create/update/patch/delete included (`:593, 624, 674, 688, 731, 751`). Same trick here: generalise `withHighlightFlags` and call it from all five places.

```java
private Settings forResponse(Settings settings) { /* today's withHighlightFlags body */ }

private Response forResponse(Response response) {
  return response.getEntity() instanceof Settings s
      ? Response.fromResponse(response).entity(forResponse(s)).build()
      : response;
}
```

`getSettingByName` and `list` wrap the settings, `createOrUpdateSetting` and `patch` wrap the response (after the write, so nothing derived gets persisted). `resetSettingToDefault` returns a bare `SearchSettings`, so annotate a `JsonUtils.deepCopy` there rather than the object itself.

Keep annotating copies, by the way. Mutating in place looks safe right now only because `readDefaultSearchSettings()` has an inverted cache guard (`if (defaultSearchSettingsCache != null)` is always true, so it re-reads the seed every call). If anyone ever fixes that guard, in-place annotation would start corrupting the shared default.

On tests: `servedAllowedFieldsCarryTheHighlightFlag` is the right test, just pointed at one endpoint. Running it over all five would have caught this, which is the wiring-level job its own javadoc describes. Same for the new Playwright test, worth having it save and then re-check the toggle.

Separate and non-blocking: `"default": false` in the schema generates `private Boolean highlight = false;`, so "nobody derived this" and "this field can't be highlighted" look identical on the wire, and the failure lands on the side that silently removes a working feature. `Field` is `@JsonInclude(NON_NULL)`, so dropping the default would leave the key absent instead, and `?? true` in `FieldConfiguration.tsx:60` would make a future miss fall back to today's behaviour rather than disabling everything. The 400 on save is the real enforcement either way. Only catch is the client-built `extension.*` rows in `fetchCustomProperties` would then need an explicit `highlight: false`.

manerow (approved):

PR #30833, opened 2026-08-02
Fixes 30711: default data contracts to draft status

PRADDZY:
### Describe your changes:

Fixes #30711

Data contract creation now defaults new contracts to Draft instead of hardcoding Approved. The contract details step also exposes the status field so users can choose Draft, In Review, or Approved before saving.

#
### Type of change:
- [x] Bug fix

#
### High-level design:
N/A - small bug fix.

#
#### Use cases covered
- Creating a data contract without selecting a status sends Draft.
- Creating a data contract after selecting In Review preserves the selected status.
- Contract detail form renders the status selector with Draft, In Review, and Approved options.

#### Unit tests
- [x] I added unit tests for the new/changed logic.
- Files added/updated:
  - `openmetadata-ui/src/main/resources/ui/src/components/DataContract/AddDataContract/AddDataContract.test.tsx`
  - `openmetadata-ui/src/main/resources/ui/src/components/DataContract/ContractDetailFormTab/ContractDetailFormTab.test.tsx`
- Coverage %: Not measured; focused Jest coverage was not run.

#### Backend integration tests
- [x] Not applicable (no backend API changes).

#### Ingestion integration tests
- [x] Not applicable (no ingestion changes).

#### Playwright (UI) tests
- [ ] Not added; covered by focused component tests for this form behavior.

#### Manual testing performed
1. `corepack yarn test --runInBand src/components/DataContract/AddDataContract/AddDataContract.test.tsx src/components/DataContract/ContractDetailFormTab/ContractDetailFormTab.test.tsx`
2. `.\\node_modules\\.bin\\eslint.cmd --no-error-on-unmatched-pattern src/components/DataContract/AddDataContract/AddDataContract.tsx src/components/DataContract/AddDataContract/AddDataContract.test.tsx src/components/DataContract/ContractDetailFormTab/ContractDetailFormTab.tsx src/components/DataContract/ContractDetailFormTab/ContractDetailFormTab.test.tsx`
3. `.\\node_modules\\.bin\\prettier.cmd --config .\\.prettierrc.yaml --ignore-path .\\.prettierignore src/components/DataContract/AddDataContract/AddDataContract.tsx src/components/DataContract/AddDataContract/AddDataContract.test.tsx src/components/DataContract/ContractDetailFormTab/ContractDetailFormTab.tsx src/components/DataContract/ContractDetailFormTab/ContractDetailFormTab.test.tsx --check`

#
### UI screen recording / screenshots:

Not attached; the status behavior is covered by focused component tests.

[...]

PR #31843, opened 2026-08-20
Fixes 30711: Data Contracts created via the UI are always set to Approved, bypassing Draft/review workflow

### LINKED ISSUE #30711: Data Contracts created via the UI are always set to Approved, bypassing Draft/review workflow
### Affected module

UI

### Describe the bug

When creating a Data Contract through the "Add Data Contract" UI wizard on a table, the contract is always saved with entityStatus: Approved, regardless of user input. There is no way to save a contract as Draft or In Review from the UI — the status field is not exposed anywhere in the wizard (Contract Detail, Terms of Service, Schema, Semantics, Security, Quality, SLA tabs).

This defeats the purpose of the entityStatus field / governance workflow described in the docs, where a contract should start as Draft, move to In Review, and only become Approved/Active after a review step (similar to Glossary Term approval workflows).

Root cause (found in source)

In openmetadata-ui/src/main/resources/ui/src/components/DataContract/AddDataContract/AddDataContract.tsx, the create-contract call hardcodes the status:

await createContract({
  ...formValues,
  displayName: formValues.name,
  entity: { id: entityData.id, type: entityType },
  semantics: validSemantics,
  security: validSecurity,
  termsOfUse: termsOfUseContent,
  entityStatus: EntityStatus.Approved, // always Approved, never Draft
});
Additionally, ContractDetailFormTab.tsx — the tab that holds the basic contract fields — only renders name, owners, and description. No status/draft selector exists in any tab of the wizard, so a user has no way to choose otherwise even if the hardcoded value were removed.

### To Reproduce

1. Go to any Table entity in OpenMetadata.
2. Open the "Contracts" tab and click "Add Contract".
3. Fill in the wizard (Details, Schema, etc.) and click Save.
4. Observe that the contract is immediately created with status Approved / shown as active, with no option to save as Draft first

### Expected behavior

1. The UI should let the user choose an initial status (e.g., default to Draft) when creating a contract.
2. Ideally, newly created contracts should default to Draft (consistent with the documented lifecycle and with how Glossary Terms behave), requiring an explicit review/approval step before becoming Approved/Active.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

1.13

_No response_

### Additional context

The EntityStatus enum (Draft, In Review, Approved, Deprecated, Rejected) is already fully supported by the backend/REST API — you can PATCH a contract's entityStatus after creation. This is purely a UI gap: the creation wizard never surfaces the field and always sends Approved.

### Pre-submission checklist

TeddyCr:
### Describe your changes:

Fixes #30711

Every data contract created through the UI wizard was persisted as `Approved`, so contracts skipped the Draft/review workflow entirely. `AddDataContract.handleSave` hardcoded the status in the create payload and spread it **after** `...formValues`, so nothing could override it:

```tsx
await createContract({
  ...formValues,
  ...
  entityStatus: EntityStatus.Approved,   // spread AFTER ...formValues -> unoverridable
});
```

Compounding it, there was no status control anywhere in the UI — `ContractDetailFormTab` exposed only `name`, `owners`, and `description`. This PR adds a status select to the contract detail tab and sends the author's choice explicitly, defaulting to `Draft`.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

**Why the status is sent explicitly instead of just deleting the literal.** This is the non-obvious part. The JSON Schema *looks* like it already defaults to Draft — `dataContract.json` and `createDataContract.json` both declare `"default": "Draft"` — but that default is **inert**. It sits beside a `$ref`, and jsonschema2pojo ignores a `default` sibling to `$ref`, resolving the referenced `type/status.json` default of `Unprocessed` instead. The generated `CreateDataContract.java` bears this out (`EntityStatus.fromValue("Unprocessed")`), and the repo's own `DataContractResourceIT.testDataContractDefaultEntityStatus` asserts `UNPROCESSED`. `DataContractRepository.prepare()` never sets the field and does not override `setDefaultStatus`.

So simply removing the hardcoded literal would have shipped every new contract as `Unprocessed` — arguably worse than the bug. The client has to send the status. I deliberately did **not** "fix" the schema default: correcting it means changing `type/status.json`, which would shift the default for every status-bearing entity in the product.

**Which statuses are offered.** `Draft` (default), `In Review`, `Approved`. Excluded:
- `Rejected` — an outcome a reviewer produces by closing the approval task (`DataContractRepository.postUpdate`), not something you author into.
- `Archived` / `Deprecated` — end-of-life states, meaningless for a contract being created.
- `Unprocessed` — an internal sentinel that only leaks from the `type/status.json` default; it has no product meaning.

Keeping `Approved` on the list is deliberate: teams that rely on today's auto-approve behaviour can still get it in one click, it is simply no longer forced.

**A regression this PR introduced, and fixed.** `ContractDetail.tsx` hardcoded `color="success"` on the status pill. That was always correct while every UI-created contract was `Approved` — but defaulting to `Draft` makes the header render a **green "success" pill reading "Draft"** (verified in a browser: the Draft badge carried `tw:bg-utility-success-50`). Fixed by mapping status to colour.

The new `getEntityStatusBadgeColor` sits **beside** the existing `getEntityStatusClass` rather than replacing it. They cannot be unified cheaply: the existing helper returns the legacy `StatusType` consumed by the Ant Design `StatusBadge`, whereas this pill is the UntitledUI `BadgeWithIcon`, which takes `BadgeColors`. Unifying would mean either changing `StatusBadge`'s contract — blast radius `GlossaryTermTab`, `EntityStatusBadge`, `ChangeParentHierarchy`, and lineage — or adapting through a 15-member enum of which 8 are irrelevant here. The new map mirrors the legacy palette in `status-badge.less` so a status looks the same in both stacks (approved green, draft/unprocessed yellow, in-review purple, rejected red, deprecated/archived grey). Both maps are `Record<EntityStatus, …>`, so TypeScript enforces exhaustiveness on each — the drift guard that actually matters.

It takes the enum's string values (`` `${EntityStatus}` ``) rather than the enum itself, because quicktype re-emits `EntityStatus` per entity with no shared module and TS string enums are nominal, so a caller holding the `dataContract` copy could not otherwise pass it.

**Files changed**
| File | Change |
| --- | --- |
| `constants/DataContract.constants.ts` | `DEFAULT_DATA_CONTRACT_STATUS` + `DATA_CONTRACT_AUTHORING_STATUS_OPTIONS` — one source of truth for both the form and the payload |
| `ContractDetailFormTab.tsx` | New `entityStatus` select, seeded `initialValues?.entityStatus ?? Draft` |
| `AddDataContract.tsx` | `entityStatus: formValues.entityStatus ?? DEFAULT_DATA_CONTRACT_STATUS` |
| `utils/EntityStatusUtils.ts` | `EntityStatusBadgeColor` / `getEntityStatusBadgeColor` |
| `ContractDetailTab/ContractDetail.tsx` | Colour by status; renamed a duplicate `data-testid` |

**Scope notes.** The edit path is untouched — it still patches via `compare()` from `filteredContract`, so an untouched status emits no `/entityStatus` op. The field is shared by create and edit, so edit gains a status control seeded from the contract's real status; that is intentional, since otherwise a Draft contract would have no UI path to Approved. Backend authorization is unchanged: `EntityRepository.updateEntityStatus` and `checkUpdatedByReviewer` still gate IN_REVIEW→APPROVED/REJECTED to reviewers. No schema change, no migration, no `make generate`.

#
#### Use cases covered
- Creating a contract through the wizard without touching Status persists `Draft`, not `Approved`
- The author can pick `In Review` or `Approved` at creation and that choice is what persists
- Only `Draft` / `In Review` / `Approved` are offered; `Rejected` / `Archived` / `Deprecated` / `Unprocessed` are not
- Editing a contract without touching Status does not rewrite its status
- Editing a contract and changing Status does patch it
- The status badge renders the colour matching the status, not always green

#### Unit tests
- Files added/updated:
  - `src/components/DataContract/AddDataContract/AddDataContract.test.tsx` (updated + 4 new)
  - `src/components/DataContract/ContractDetailFormTab/ContractDetailFormTab.test.tsx` (2 new)
  - `src/components/DataContract/ContractDetailFormTab/ContractDetailFormTab.entityStatus.test.tsx` (**new file**, 4 tests)
  - `src/components/DataContract/ContractDetailTab/ContractDetail.test.tsx` (7 new)
- Coverage on changed files (measured, `jest --coverage`, % lines):
  - `ContractDetailFormTab.tsx` — **100%**
  - `ContractDetail.tsx` — **97.1%**
  - `EntityStatusUtils.ts` — **90.9%** (the only uncovered line is the pre-existing `isDeleted`; the code added here is fully covered)
  - `AddDataContract.tsx` — **89.91%**

Notes on what the tests actually assert, since it matters here:
- **Payload, not mock wiring** — the `AddDataContract` tests assert the object handed to `createContract` (the HTTP boundary), which is exactly what the bug corrupted.
- **Real rendering** — `ContractDetailFormTab.entityStatus.test.tsx` is a separate new file specifically so it can render the *real* Ant Design control; the pre-existing sibling stubs `generateFormFields` module-wide and cannot.
- **Patch-op guard** — `should not patch entityStatus when editing without touching the status` inspects the real JSON-patch array, not `expect.any(Array)`. Verified non-vacuous by injecting the regression, which produces `{"op":"replace","path":"/entityStatus","value":"Draft"}` and fails the guard.
- Every new test was confirmed to fail before the fix and pass after.

#### Playwright (UI) tests
- [x] I added Playwright E2E tests for the UI change.
- Files added/updated: `openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts` (2 new tests)

There was previously **zero** Playwright coverage of `entityStatus` for contracts — the existing `contract-status-card-item-*` locators are *validation* status, not entity status. The new tests assert the **`POST /api/v1/dataContracts` response body**, not the form control that was just set, so they prove persistence rather than echoing input. They also assert the rendered badge's palette class.

#### Manual testing performed
1. Started the stack and the UI; logged in as admin.
2. Navigated to a table → **Contract** tab → **Add Contract**.
3. Confirmed a **Status** field is present between *Owners* and *Description*, **pre-selected to `Draft`**.
4. Opened the dropdown and confirmed exactly `Draft`, `In Review`, `Approved` — and that `Rejected`, `Archived`, `Deprecated`, `Unprocessed` are absent.
5. Entered a title, left Status untouched, saved.
6. With DevTools → Network open, confirmed the `POST /api/v1/dataContracts` **response body** contains `"entityStatus": "Draft"`.
7. Confirmed the header renders `Status : Draft` and the pill is **yellow, not green** (`tw:bg-utility-warning-50`).
8. Created a second contract selecting **Approved**; confirmed the response body has `"entityStatus": "Approved"` and the pill is **green**.
9. Created a third as **In Review**; confirmed the pill is **purple**.
10. Edited an existing contract: confirmed the Status select shows its *current* status rather than defaulting to Draft, and that saving without touching Status leaves `entityStatus` unchanged (no `/entityStatus` op in the PATCH body).

#

> **TODO — recording still to be attached.** This is a UI change, so a recording is required. It is not attached yet and I have not fabricated one; the PR is opened as a draft until it is added.

What the recording should capture, in one continuous take:
- the Add Contract form with **Status pre-selected to Draft**;
- the **open dropdown** showing exactly the three authoring statuses;
- the **DevTools Network panel** with the `POST /api/v1/dataContracts` response body visible showing `"entityStatus": "Draft"`;
- the resulting contract header reading `Status : Draft` with the **yellow** pill visible;
- a second creation with **Approved** selected, again with the response body and the resulting `Status : Approved` header and **green** pill.

The Network panel is the load-bearing frame — the badge alone does not prove what was persisted. The two pills side by side are what demonstrate the colour fix.

#
### Known gaps (disclosed, not fixed here):

- **The ODCS import modal is a second UI create path.** `ODCSImportModal.component.tsx` also calls `createContract`, forwarding the parsed document verbatim with no `entityStatus`. So "UI-created contracts default to Draft" is true of the **wizard only**: importing an ODCS document that omits the field still lands as `Unprocessed`, while the wizard gives `Draft`. This is pre-existing and outside this issue's "creation wizard" framing, so it is deliberately left alone — a natural follow-up, and the fix would be the same one-line default. The badge-colour fix in this PR *does* already make that state render correctly (yellow) instead of green.
- **Picking "In Review" creates no approval task.** There is no seeded BPMN governance workflow for `dataContract`, and `DataContractRepository` only ever *closes* approval tasks (they are created by `CreateApprovalTaskImpl`, which no dataContract workflow invokes). This is a pre-existing product gap rather than a regression, and it is not a hard trap — `checkUpdatedByReviewer` only restricts the transition when `reviewers` is non-empty — but the new dropdown does set an expectation the product does not yet fully meet.

#
### Checklist:
- [x] For JSON Schema changes: N/A — no schema change (and see the design section for why the existing inert schema default was deliberately left alone).
- [ ] For UI changes: I attached a screen recording and/or screenshots above. — **outstanding, see above**

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
@@ -2249,6 +2249,90 @@ description:
       });
     }
   });
+
+  test('Contract created from the form defaults to Draft', async ({ page }) => {
+    const { apiContext } = await getApiContext(page);
+    const table = new TableClass();
+    await table.create(apiContext);
+
+    try {
+      await redirectToHomePage(page);
+      await table.visitEntityPage(page);
+      await navigateToContractTab(page);
+      await clickAddContractButton(page);
+
+      await expect(page.getByTestId('contract-entity-status')).toContainText(
+        'Draft'
+      );
+
+      await page.getByTestId('contract-name').fill('draft_default_contract');
+
+      const createResponse = page.waitForResponse(
+        (response) =>
+          response.url().endsWith('/api/v1/dataContracts') &&
+          response.request().method() === 'POST'
+      );
+      await page.getByTestId('save-contract-btn').click();
+      const created = await createResponse;
+
+      expect(created.ok()).toBe(true);
+      expect((await created.json()).entityStatus).toBe('Draft');
+
+      // `contract-status-label` is also used by the owners label, so pin the
+      // status one by its own text before walking up to the row holding its badge.
+      await expect(
+        page
+          .getByTestId('contract-status-label')
+          .filter({ hasText: 'Status' })
+          .locator('xpath=../..')
+      ).toContainText('Draft');
+    } finally {
+      await table.delete(apiContext);
+    }
+  });
+
+  test('Contract is created with the status picked in the form', async ({
+    page,
+  }) => {
+    const { apiContext } = await getApiContext(page);
+    const table = new TableClass();
+    await table.create(apiContext);
+
+    try {
+      await redirectToHomePage(page);
+      await table.visitEntityPage(page);
+      await navigateToContractTab(page);
+      await clickAddContractButton(page);
+
+      await page.getByTestId('contract-name').fill('approved_choice_contract');
+
+      await selectOption(
+        page,
+        page.getByTestId('contract-entity-status'),
+        'Approved'
+      );
+
+      const createResponse = page.waitForResponse(
+        (response) =>
+          response.url().endsWith('/api/v1/dataContracts') &&
+          response.request().method() === 'POST'
+      );
+      await page.getByTestId('save-contract-btn').click();
+      const created = await createResponse;
+
+      expect(created.ok()).toBe(true);
+      expect((await created.json()).entityStatus).toBe('Approved');
+
+      await expect(
+        page
+          .getByTestId('contract-status-label')
+          .filter({ hasText: 'Status' })
+          .locator('xpath=../..')
+      ).toContainText('Approved');
+    } finally {
+      await table.delete(apiContext);
+    }
+  });
 });
 
 entitiesWithDataContracts.forEach((EntityClass) => {
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/DataContracts.spec.ts
@@ -2278,14 +2278,13 @@ description:
       expect(created.ok()).toBe(true);
       expect((await created.json()).entityStatus).toBe('Draft');
 
-      // `contract-status-label` is also used by the owners label, so pin the
-      // status one by its own text before walking up to the row holding its badge.
-      await expect(
-        page
-          .getByTestId('contract-status-label')
-          .filter({ hasText: 'Status' })
-          .locator('xpath=../..')
-      ).toContainText('Draft');
+      const statusCard = page.getByTestId('contract-status-card');
+      const statusBadge = statusCard.getByText('Draft', { exact: true });
+
+      await expect(statusBadge).toBeVisible();
+      // A Draft must not be painted with the green success palette.
+      await expect(statusBadge).toHaveClass(/utility-warning/);
+      await expect(statusBadge).not.toHaveClass(/utility-success/);
     } finally {
       await table.delete(apiContext);
     }
@@ -2323,12 +2322,11 @@ description:
       expect(created.ok()).toBe(true);
       expect((await created.json()).entityStatus).toBe('Approved');
 
-      await expect(
-        page
-          .getByTestId('contract-status-label')
-          .filter({ hasText: 'Status' })
-          .locator('xpath=../..')
-      ).toContainText('Approved');
+      const statusCard = page.getByTestId('contract-status-card');
+      const statusBadge = statusCard.getByText('Approved', { exact: true });
+
+      await expect(statusBadge).toBeVisible();
+      await expect(statusBadge).toHaveClass(/utility-success/);
     } finally {
       await table.delete(apiContext);
     }
--- a/openmetadata-ui/src/main/resources/ui/src/components/DataContract/AddDataContract/AddDataContract.test.tsx
+++ b/openmetadata-ui/src/main/resources/ui/src/components/DataContract/AddDataContract/AddDataContract.test.tsx
@@ -556,6 +556,59 @@ describe('AddDataContract', () => {
       expect(mockOnSave).toHaveBeenCalled();
     });
 
+    it('should not patch entityStatus when editing without touching the status', async () => {
+      render(
+        <AddDataContract
+          contract={mockContract}
+          onCancel={mockOnCancel}
+          onSave={mockOnSave}
+        />
+      );
+
+      // Change an unrelated field; mockContract is already Approved, so a
+      // spurious /entityStatus op would silently rewrite an existing status.
+      await act(async () => {
+        fireEvent.click(screen.getByText('Change'));
+      });
+
+      await act(async () => {
+        fireEvent.click(screen.getByTestId('save-contract-btn'));
+      });
+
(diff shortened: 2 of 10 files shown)


TeddyCr:
closing as duplicate of https://github.com/open-metadata/OpenMetadata/pull/30833

PR #31875, opened 2026-08-21
Fixes 31871: MCP error responses serialize the McpError exception — unauthenticated 401 leaks a full Java stack trace

### LINKED ISSUE #31871: MCP error responses serialize the McpError exception — unauthenticated 401 leaks a full Java stack trace
## Summary

An unauthenticated `POST /mcp` returns a **13.8 KB JSON body containing a full 60-frame Java stack trace**. `McpError` extends `RuntimeException`, and we hand the exception object itself to Jackson, so the serialized form carries `stackTrace`, `cause`, `suppressed` and `localizedMessage` straight to anonymous callers.

This blocks the Claude Directory / ChatGPT app submissions: `/mcp` unauthenticated is the *first* request a reviewer makes, and "meet Anthropic's security standards" is submission requirement #1.

## Reproduce

```bash
curl -i -X POST https://<tenant>.getcollate.io/mcp \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'
```

Verified live on `devrel` 2026-08-21 → `HTTP/2 401`, `content-length` 13874.

Body keys: `cause`, `stackTrace` (60 frames), `jsonRpcError`, `message`, `suppressed`, `localizedMessage`.

Leaked frames include our own class/file/line layout, the Jetty/Dropwizard/servlet internals and the exact JRE build:

```
io.modelcontextprotocol.spec.McpError$Builder.build(McpError.java:72)
org.openmetadata.mcp.server.transport.OAuthHttpStatelessServerTransportProvider.sendAuthErrorWithChallenge(OAuthHttpStatelessServerTransportProvider.java:451)
org.openmetadata.mcp.server.transport.OAuthHttpStatelessServerTransportProvider.authenticateRequest(OAuthHttpStatelessServerTransportProvider.java:392)
org.openmetadata.mcp.server.transport.HttpServletStatelessServerTransport.doPost(HttpServletStatelessServerTransport.java:147)
...
org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1194)
java.lang.Thread.run(Thread.java:1583)   ← moduleName "java.base", moduleVersion "21.0.12"
```

## Root cause

Two sites serialize the `McpError` **exception** rather than its JSON-RPC error payload:

1. `openmetadata-mcp/.../transport/OAuthHttpStatelessServerTransportProvider.java:451-452` — the auth path.

   ```java
   McpError error = McpError.builder(McpSchema.ErrorCodes.INTERNAL_ERROR).message(message).build();
   String jsonError = getObjectMapper().writeValueAsString(error);
   ```

2. `openmetadata-mcp/.../transport/HttpServletStatelessServerTransport.java:250-259` — `responseError(...)`, same pattern, which covers **every** 400/500 path in `doPost`. Those are authenticated-only, but one of them is `"Unexpected error: " + e.getMessage()`, so an authenticated caller gets both an internal exception message and a stack trace.

## Secondary problems on the same response

- **Wrong message when no credentials were sent.** With no `Authorization` header at all the body says `"Invalid or expired token"`. Per RFC 6750 §3 those are distinct cases — no credentials is not an invalid token, and the wording sends integrators hunting for a bad token they never had.
- **Wrong JSON-RPC code.** `-32603 INTERNAL_ERROR` for an auth failure. Nothing internal failed.
- **`WWW-Authenticate` never carries `error="invalid_token"`.** `sendAuthErrorWithChallenge` only appends an `error` parameter for 403/`insufficient_scope`. RFC 6750 §3.1 wants `error="invalid_token"` on the 401-with-a-bad-token case (and correctly *omits* it when no credentials were supplied — that half is already right).

The header side is otherwise in good shape: `www-authenticate: Bearer resource_metadata="…/mcp/.well-known/oauth-protected-resource", scope="openid profile email"` is exactly what the MCP spec asks for.

## Expected

A bare, RFC-shaped error envelope and nothing else:

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32001,"message":"Missing or invalid bearer token"}}
```

`McpError` already exposes `getJsonRpcError()` (`{code, message}`) — serialize that inside a `JSONRPCResponse`, never the throwable. Stack traces belong in `LOG.debug`, which `authenticateRequest` already does.

## Acceptance criteria

- [ ] No `stackTrace` / `cause` / `suppressed` / `localizedMessage` key in any `/mcp` response body, at any status code, authenticated or not.
- [ ] `responseError(...)` in `HttpServletStatelessServerTransport` fixed too — not just the auth path.
- [ ] Missing-credentials vs invalid-token distinguished in the message, with an auth-appropriate JSON-RPC code instead of `-32603`.
- [ ] `error="invalid_token"` added to `WWW-Authenticate` when a token was supplied and rejected; still absent when no credentials were sent.
- [ ] Regression test asserting the unauthenticated 401 body against the exact expected JSON, so this cannot silently come back.
- [ ] `"Unexpected error: " + e.getMessage()` reviewed — internal exception text should not reach the wire either.

pmbrull:
### Describe your changes:

Fixes #31871

An unauthenticated `POST /mcp` returned a **13.8 KB body containing a 60-frame Java stack trace**. `McpError` extends `RuntimeException`, and two sites handed the exception object itself to Jackson, so `stackTrace`, `cause`, `suppressed` and `localizedMessage` went straight to anonymous callers — along with our class/file/line layout, the Jetty/Dropwizard/servlet internals and the exact JRE build. This blocked the Claude Directory / ChatGPT app submissions, where the unauthenticated `/mcp` call is the first request a reviewer makes.

The unauthenticated 401 body is now exactly:

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32001,"message":"Missing bearer token"}}
```

#
### Type of change:
- [x] Bug fix

#
### High-level design:

**`JsonRpcErrorBody` (new)** — the single place that renders an MCP failure as a bare JSON-RPC 2.0 error envelope. Both leaking sites now go through it, so there is one thing to audit instead of two.

It assembles the envelope node by node rather than serializing a POJO, for two reasons:
- `id` must be **present and null**. JSON-RPC 2.0 §5 requires the member on every response, and these bodies are produced either before the request id is known (auth, unparseable body) or after the handler already failed. `McpSchema.JSONRPCResponse` is annotated `@JsonInclude(NON_ABSENT)` and would silently drop it.
- `JSONRPCError.data` is **deliberately not copied**. It is the one member that could carry caller-supplied or exception-derived content, and the whole point of this class is to guarantee that nothing beyond a code and a message reaches an anonymous caller.

It also owns a private `ObjectMapper` so the envelope stays byte-identical regardless of how a caller configured its own (a global `NON_NULL` inclusion would drop the mandatory `id`).

**Auth path (`OAuthHttpStatelessServerTransportProvider`)** — `sendAuthErrorWithChallenge` took a loose `(message, statusCode)` pair, which made it possible to pair a status with a mismatched challenge. Replaced with an `AuthFailure` enum carrying status + challenge-error + message as one unit:

| | HTTP | `WWW-Authenticate` | message |
|---|---|---|---|
| `MISSING_CREDENTIALS` | 401 | no `error` parameter | `Missing bearer token` |
| `INVALID_TOKEN` | 401 | `error="invalid_token"` | `Invalid or expired bearer token` |
| `INSUFFICIENT_SCOPE` | 403 | `error="insufficient_scope"` | `Insufficient scope` |

`authenticateRequest` now classifies on the `Authorization` header (`presentsBearerCredentials`, scheme matched case-insensitively per RFC 7235) **before** validating, so no-credentials and bad-token are genuinely distinct — RFC 6750 §3 wants `invalid_token` only where a token was actually presented. Previously every failure said `"Invalid or expired token"`, sending integrators hunting for a token they never had.

Both use `-32001` from the JSON-RPC implementation-defined range (`-32000..-32099`) instead of `-32603 INTERNAL_ERROR`, which claimed something broke on our side when nothing did.

Challenge construction was split into `buildAuthChallenge`, and `writeAuthError` made `static`, so both can be unit tested — the class constructor needs a full running auth stack. Same pattern already used for `applyCorsHeaders` in this file.

**Stateless transport (`HttpServletStatelessServerTransport`)** — `responseError` writes the JSON-RPC payload rather than the throwable, and is `static` for testability. All four call sites stop concatenating `e.getMessage()` into the wire message (`"Unexpected error: " + e.getMessage()` → `"Internal server error"`, etc.); the throwable is now passed to the logger instead of just its message, so the detail lands where it belongs.

**Rejected alternative:** annotating a POJO with per-property `@JsonInclude(ALWAYS)` to force the `id` through. It works, but the behaviour then depends on a subtle annotation interaction that a future Jackson upgrade or a mapper-level config change could quietly break — and the failure mode is a silently malformed protocol response. Building the node explicitly makes the wire shape unmissable and lets the test pin it byte for byte.

**Backward compatibility:** the response *shape* changes for every `/mcp` error — from a serialized exception to a JSON-RPC envelope. That is the fix. Any client parsing the old shape was parsing a stack trace.

**Upstream:** the SDK has the same bug — `io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport.responseError` (1.1.1) does `jsonMapper.writeValueAsString(mcpError)`. Our fork now diverges here; worth an upstream issue separately.

#
#### Use cases covered
- Anonymous caller `POST /mcp` with no `Authorization` header → 401, bare JSON-RPC envelope, no stack trace, challenge carries no `error` parameter
- Caller `POST /mcp` with a bad bearer token → 401, `Invalid or expired bearer token`, challenge carries `error="invalid_token"`
- `Authorization` header using another scheme (`Basic …`) or a blank bearer credential → treated as *no credentials presented*, not as a bad token
- Bearer scheme accepted case-insensitively (`bearer`, `BEARER`) per RFC 7235
- 400 (bad `Accept` header, unparseable body, neither request nor notification) and 500 (handler failure) `/mcp` bodies carry no `stackTrace` / `cause` / `suppressed` / `localizedMessage`
- 403 / `insufficient_scope` challenge still emitted correctly
- `resource_metadata` and `scope` remain on the challenge — MCP clients need them to start OAuth

#### Unit tests
- Files added/updated:
  - `openmetadata-mcp/src/test/java/org/openmetadata/mcp/server/transport/JsonRpcErrorBodyTest.java` (new, 7 tests)
  - `openmetadata-mcp/src/test/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProviderTest.java` (+12 tests)
  - `openmetadata-mcp/src/test/java/org/openmetadata/mcp/server/transport/HttpServletStatelessServerTransportTest.java` (+3 tests)

`JsonRpcErrorBodyTest` includes a test that **documents the defect itself** — asserting that a stock `ObjectMapper` really does emit `stackTrace`/`suppressed`/`localizedMessage` for an `McpError` — so the `doesNotContain(...)` guards elsewhere cannot silently go vacuous if the SDK changes.

Coverage (`mvn -P static-code-analysis test jacoco:report -pl openmetadata-mcp`), line coverage on changed classes:

| Class | Line coverage |
|---|---|
| `JsonRpcErrorBody` | **100%** (10/10) |
| `OAuthHttpStatelessServerTransportProvider.AuthFailure` | **100%** (12/12) |

Every method added or extracted in this PR is fully covered:

| Method | Line coverage |
|---|---|
| `JsonRpcErrorBody.of` | 10/10 |
| `OAuthHttpStatelessServerTransportProvider.presentsBearerCredentials` | 5/5 |
| `OAuthHttpStatelessServerTransportProvider.writeAuthError` | 9/9 |
| `OAuthHttpStatelessServerTransportProvider.buildAuthChallenge` | 6/6 |
| `HttpServletStatelessServerTransport.responseError` | 7/7 |

The enclosing classes stay at their pre-existing low numbers (`OAuthHttpStatelessServerTransportProvider` 7.7%, `HttpServletStatelessServerTransport` 24.4%) — that is untouched surface, and it did not move in this PR. The two changed methods still at 0% under surefire, `authenticateRequest` (11 lines) and `sendAuthErrorWithChallenge` (2 lines), require a live server and are covered by the integration test below.

Full module suite: **469 tests, 0 failures**.

- [x] I updated integration tests in `openmetadata-integration-tests/`.
- Files updated: `openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/mcp/McpOAuthIT.java`

`testMcpEndpointWithoutAuth_returns401` and `testMcpEndpointWithInvalidToken_returns401` previously asserted only the status code. They now assert the **exact wire body** and the `WWW-Authenticate` challenge against the real running endpoint — this is the regression anchor the issue asks for, and it would have failed before the fix.

Result: `McpOAuthIT` — **13 tests, 0 failures**.

#### Playwright (UI) tests

#### Manual testing performed

Verified by the integration test above rather than by hand — it drives the real `POST /mcp` over HTTP against a Testcontainers-backed stack (MySQL + Elasticsearch + the Dropwizard app) and asserts the byte-exact body, which is stricter and repeatable. To reproduce manually against a local stack:

1. `docker compose -f docker/development/docker-compose.yml up -d` and start the server
2. `curl -i -X POST http://localhost:8585/mcp -H 'Content-Type: application/json' -H 'Accept: application/json, text/event-stream' -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'`
   → `HTTP/1.1 401`, `content-length: 83`, body is the envelope above, `WWW-Authenticate` has no `error=` parameter
3. Repeat with `-H 'Authorization: Bearer nope'`
   → same shape, message `Invalid or expired bearer token`, `WWW-Authenticate` now carries `error="invalid_token"`

Also confirmed the servlet-container error path does not leak either: Jetty 12.1.10's `ErrorHandler` defaults to `_showStacks = false` and nothing in this repo calls `setShowStacks`, so `GET /mcp` (405) and unknown sub-paths (404) return no stack frames.

#

Not applicable.

#
### Checklist:
- [x] For JSON Schema changes: N/A — no schema changes.
- [x] For UI changes: N/A — no UI changes.

Bug fix
- [x] I have added a test that covers the exact scenario we are fixing. `McpOAuthIT.testMcpEndpointWithoutAuth_returns401` asserts the exact unauthenticated 401 body (issue #31871).

#
### Note on the Collate side

No mirrored change is needed. `collate-mcp` contains only `CollateMcpServer extends org.openmetadata.mcp.McpServer` (overriding `getTools()`/`getPrompts()`) plus three tool classes — no transport, servlet, OAuth or error-serialization code. `CollateApplication.registerMCPServer` reflectively loads it and the OSS `initializeMcpServer` wires `OAuthHttpStatelessServerTransportProvider`, confirmed at runtime by Collate's own logs. Collate picks this up with the submodule bump alone.

### Follow-up, not in scope

On the **success** path, `DefaultMcpStatelessServerHandler.onErrorResume` (SDK) puts a raw `t.getMessage()` into a `JSONRPCError` at HTTP 200 for any non-`McpError` exception escaping a tool. No stack frames and none of the forbidden keys, so it meets this issue's acceptance criteria, but an internal exception message can still reach an authenticated caller. Fixing it would change tool-error UX broadly and belongs in its own issue.

The PR prevents MCP error responses from serializing throwable state and completes the prior request-correlation fix.
- Builds minimal JSON-RPC error envelopes containing only the request ID, code, and message.
- Preserves known request IDs when request handling fails while using `null` before an ID is available.
- Distinguishes missing credentials from invalid bearer tokens in HTTP status bodies and authentication challenges.
- Adds unit and integration coverage for response shape, ID propagation, and authentication behavior.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/JsonRpcErrorBody.java | Constructs sanitized JSON-RPC envelopes and preserves conforming string or numeric request IDs. |
| openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/HttpServletStatelessServerTransport.java | Routes transport failures through sanitized responses and now supplies the parsed request ID on handler failures. |
| openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/OAuthHttpStatelessServerTransportProvider.java | Separates missing and invalid credentials while emitting sanitized authentication errors and RFC-aligned challenges. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/mcp/McpOAuthIT.java | Pins the unauthenticated and invalid-token response bodies and challenge parameters at the live endpoint boundary. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(mcp): echo the request id and drop t..."](https://github.com/open-metadata/openmetadata/commit/d0435df311844fa6ed7a765da3392df0094f3b66) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55498136)</sub>

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/mcp/McpOAuthIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/mcp/McpOAuthIT.java
@@ -296,6 +296,15 @@ public class McpOAuthIT extends McpTestBase {
     HttpResponse<String> response = HTTP_CLIENT.send(request, HttpResponse.BodyHandlers.ofString());
 
     assertThat(response.statusCode()).isEqualTo(401);
+    // The unauthenticated POST /mcp is the first request an app-directory reviewer makes. It used
+    // to answer with a serialized McpError - a RuntimeException - and shipped a 13.8 KB body with a
+    // 60-frame stack trace naming our classes, the Jetty internals and the exact JRE build.
+    assertThat(response.body())
+        .isEqualTo(
+            "{\"jsonrpc\":\"2.0\",\"id\":null,"
+                + "\"error\":{\"code\":-32001,\"message\":\"Missing bearer token\"}}");
+    // RFC 6750 section 3: a request that presented no credentials gets no error parameter.
+    assertThat(challengeOf(response)).startsWith("Bearer ").doesNotContain("error=");
   }
 
   @Test
@@ -317,5 +326,15 @@ public class McpOAuthIT extends McpTestBase {
     HttpResponse<String> response = HTTP_CLIENT.send(request, HttpResponse.BodyHandlers.ofString());
 
     assertThat(response.statusCode()).isEqualTo(401);
+    assertThat(response.body())
+        .isEqualTo(
+            "{\"jsonrpc\":\"2.0\",\"id\":null,"
+                + "\"error\":{\"code\":-32001,\"message\":\"Invalid or expired bearer token\"}}");
+    // RFC 6750 section 3.1: a token was supplied and rejected, so the challenge says so.
+    assertThat(challengeOf(response)).contains("error=\"invalid_token\"");
+  }
+
+  private static String challengeOf(HttpResponse<String> response) {
+    return response.headers().firstValue("WWW-Authenticate").orElse("");
   }
 }
--- a/openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/HttpServletStatelessServerTransport.java
+++ b/openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/HttpServletStatelessServerTransport.java
@@ -157,6 +157,7 @@ public class HttpServletStatelessServerTransport extends HttpServlet
       responseError(
           response,
           HttpServletResponse.SC_BAD_REQUEST,
+          null,
           McpError.builder(McpSchema.ErrorCodes.INVALID_REQUEST)
               .message("Accept header must include application/json or text/event-stream")
               .build());
@@ -193,6 +194,7 @@ public class HttpServletStatelessServerTransport extends HttpServlet
           responseError(
               response,
               HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+              jsonrpcRequest.id(),
               McpError.builder(McpSchema.ErrorCodes.INTERNAL_ERROR)
                   .message("Failed to handle request")
                   .build());
@@ -209,6 +211,7 @@ public class HttpServletStatelessServerTransport extends HttpServlet
           responseError(
               response,
               HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+              null,
               McpError.builder(McpSchema.ErrorCodes.INTERNAL_ERROR)
                   .message("Failed to handle notification")
                   .build());
@@ -217,6 +220,7 @@ public class HttpServletStatelessServerTransport extends HttpServlet
         responseError(
             response,
             HttpServletResponse.SC_BAD_REQUEST,
+            null,
             McpError.builder(McpSchema.ErrorCodes.INVALID_REQUEST)
                 .message("The server accepts either requests or notifications")
                 .build());
@@ -226,6 +230,7 @@ public class HttpServletStatelessServerTransport extends HttpServlet
       responseError(
           response,
           HttpServletResponse.SC_BAD_REQUEST,
+          null,
           McpError.builder(McpSchema.ErrorCodes.INVALID_REQUEST)
               .message("Invalid message format")
               .build());
@@ -234,6 +239,7 @@ public class HttpServletStatelessServerTransport extends HttpServlet
       responseError(
           response,
           HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+          null,
           McpError.builder(McpSchema.ErrorCodes.INTERNAL_ERROR)
               .message("Internal server error")
               .build());
@@ -248,16 +254,18 @@ public class HttpServletStatelessServerTransport extends HttpServlet
    *
    * @param response The HTTP servlet response
    * @param httpCode The HTTP status code
+   * @param requestId The id of the request being answered, or null where it is not yet known
    * @param mcpError The MCP error to send
    * @throws IOException If an I/O error occurs
    */
-  static void responseError(HttpServletResponse response, int httpCode, McpError mcpError)
+  static void responseError(
+      HttpServletResponse response, int httpCode, Object requestId, McpError mcpError)
       throws IOException {
     response.setContentType(APPLICATION_JSON);
     response.setCharacterEncoding(UTF_8);
     response.setStatus(httpCode);
     PrintWriter writer = response.getWriter();
-    writer.write(JsonRpcErrorBody.of(mcpError));
+    writer.write(JsonRpcErrorBody.of(requestId, mcpError));
     writer.flush();
   }
 
--- a/openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/HttpServletStatelessServerTransport.java
+++ b/openmetadata-mcp/src/main/java/org/openmetadata/mcp/server/transport/HttpServletStatelessServerTransport.java
@@ -154,7 +154,7 @@ public class HttpServletStatelessServerTransport extends HttpServlet
     boolean acceptsJson = accept != null && accept.contains(APPLICATION_JSON);
     boolean acceptsSse = accept != null && accept.contains(TEXT_EVENT_STREAM);
     if (!acceptsJson && !acceptsSse) {
-      this.responseError(
+      responseError(
           response,
           HttpServletResponse.SC_BAD_REQUEST,
           McpError.builder(McpSchema.ErrorCodes.INVALID_REQUEST)
@@ -189,12 +189,12 @@ public class HttpServletStatelessServerTransport extends HttpServlet
             writeJsonResponse(response, jsonResponseText);
           }
         } catch (Exception e) {
-          logger.error("Failed to handle request: {}", e.getMessage());
-          this.responseError(
+          logger.error("Failed to handle request", e);
+          responseError(
               response,
               HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
               McpError.builder(McpSchema.ErrorCodes.INTERNAL_ERROR)
-                  .message("Failed to handle request: " + e.getMessage())
+                  .message("Failed to handle request")
                   .build());
         }
       } else if (message instanceof McpSchema.JSONRPCNotification jsonrpcNotification) {
@@ -205,16 +205,16 @@ public class HttpServletStatelessServerTransport extends HttpServlet
               .block();
           response.setStatus(HttpServletResponse.SC_ACCEPTED);
         } catch (Exception e) {
-          logger.error("Failed to handle notification: {}", e.getMessage());
-          this.responseError(
+          logger.error("Failed to handle notification", e);
+          responseError(
               response,
               HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
               McpError.builder(McpSchema.ErrorCodes.INTERNAL_ERROR)
-                  .message("Failed to handle notification: " + e.getMessage())
+                  .message("Failed to handle notification")
                   .build());
         }
       } else {
-        this.responseError(
+        responseError(
             response,
             HttpServletResponse.SC_BAD_REQUEST,
             McpError.builder(McpSchema.ErrorCodes.INVALID_REQUEST)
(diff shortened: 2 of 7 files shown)


pmbrull:
Both review findings addressed in d0435df.

### @greptile-apps — known request ID is discarded (P1)

Correct, and worth fixing. The request-handler `catch` already had `jsonrpcRequest.id()` in hand and still emitted `id:null`. JSON-RPC 2.0 §5 requires the id to be echoed whenever it is known, and a client that pipelines requests cannot otherwise tell which one failed.

`responseError` now takes the id explicitly rather than defaulting it, so each of the six call sites has to state what it actually knows:

| Call site | id | why |
|---|---|---|
| handler failure on a `JSONRPCRequest` | `jsonrpcRequest.id()` | **known — now echoed** |
| `Accept` header rejected | `null` | body not read yet |
| handler failure on a `JSONRPCNotification` | `null` | a notification carries no id by definition |
| message is neither request nor notification | `null` | no id member to read |
| deserialization failed | `null` | §5: id MUST be null when it could not be detected |
| outer catch-all | `null` | failure could be anywhere; id not reliably known |

One thing I tightened beyond the report: §4 restricts the id to a String, a Number or null, so `idNode` only reflects those two types back. A client that sends `"id": {...}` gets `null` rather than having its own payload echoed to it out of an error path — this class exists to guarantee a minimal envelope, and echoing arbitrary caller structure would undercut that.

Tests added: `of_knownNumericId_echoesIt`, `of_knownStringId_echoesIt`, `of_unknownId_emitsNull`, `of_nonConformingId_isNotReflectedBack`, `responseError_knownRequestId_echoesItIntoTheEnvelope`.

### @gitar-bot — `AuthFailure.INSUFFICIENT_SCOPE` unused; 403 body uses `UNAUTHORIZED`

Also correct — and both halves applied.

Dropped the constant. Worth noting it wasn't newly-dead code: the `if (statusCode == SC_FORBIDDEN) challenge.append(", error=\"insufficient_scope\"")` branch it replaced was equally unreachable before this PR, since no caller ever passed 403. Removing it drops no live behaviour, and carrying an unreachable enum value with a wrong error code is worse than either alternative.

Also took the second half of the suggestion, because it removes the trap rather than just the symptom: the JSON-RPC code now lives on the enum next to the HTTP status instead of being hardcoded in `writeAuthError`. A 403 kind added later picks its own code by construction and cannot silently inherit an authentication one. `authFailure_jsonRpcCodeMatchesHttpStatus` pins the invariant.

### Verification

- `mvn -pl openmetadata-mcp test` → **473 tests, 0 failures**
- `McpOAuthIT` → **13 tests, 0 failures** (unchanged — the auth path body is `id:null` either way, since authentication runs before the body is read)
- Coverage held: `JsonRpcErrorBody` 100% (14/14), `AuthFailure` 100% (13/13); `responseError`, `writeAuthError`, `buildAuthChallenge`, `presentsBearerCredentials` all 100%
- `mvn spotless:apply` clean; `java-checkstyle` green

The `Validate PR Metadata` failure is unrelated to the code — issue #31871 already carries `Shipping → Source: Product`; that check ran before the field was set and does not re-trigger on project changes, so this push should clear it.

Vishnuujain (approved):

pmbrull:
picked to 2.0

PR #31884, opened 2026-08-21
fix(mssql): harden fault tolerance and observability across query-history, metadata, and shared lineage/SP code

akashverma0786:
### Describe your changes:

No linked issue — reliability/observability hardening fixes for the mssql connector, found during a manual connector-reliability audit (error handling, fault tolerance, and observability standards). Happy to open issues retroactively if preferred.

**Error classification & observability:**
- `SQLSERVER_ERRORS` didn't recognize Azure SQL's DTU/vCore resource-governance throttling errors (10928/10929/10936, including the pyodbc text-only fallback and elastic-pool wording), so a throttled customer got a generic/unclassified error instead of an actionable diagnosis.
- Description-map load failures (schema/database/stored-procedure comments) were logged at `DEBUG`, so a real permission or connection issue on that path was invisible by default. Raised to `WARNING`.
- The `ingestAllDatabases` per-database engine loop never logged which query-log source (Query Store vs plan-cache DMVs) was picked for each database, unlike the single-database path which already logs this.

**Unguarded query executions that could crash a run outright:**
- `_databases_to_scan()` / `_per_database_engines()` (`query_parser.py`): both the database-enumeration query and per-database engine construction are now guarded. A transient failure (network blip, timeout, permission gap on one database) no longer aborts query history for every other database — it's skipped, or the run falls back to the single instance-wide connection when enumeration itself fails.
- `get_sqlalchemy_engine_dateformat()` (`utils.py`): guarded against its `DBCC USEROPTIONS` probe failing, so a transient connection failure at that exact moment no longer crashes `MssqlUsageSource`/lineage source construction — falls back to the documented `DEFAULT_DATETIME_FORMAT` instead. Also replaced a bare `return` with `return None` for clarity.
- `get_stored_procedures()` (`metadata.py`): a permission/syntax failure listing stored procedures for one schema is now reported via `status.warning()` instead of crashing the whole generator uncaught. Deliberately uses `.warning()`, not `.failed()` — a `DENY SELECT ON sys.sql_modules` on one schema's optional stored procedures shouldn't drag down the entire run's reported success percentage (`Status.calculate_success()` only counts `.failed()` against it).
- `get_database_names()` single-database branch (`metadata.py`): a connection failure is now recorded via `status.failed()` instead of crashing the run — matches the multi-database branch's existing handling three lines below it. Unlike the stored-procedure case above, `.failed()` is correct here: in single-database mode, this database *is* the entire unit of work for the run.
- `lineage_source.yield_table_query()` (shared, `lineage_source.py`): a connection/permission failure on one engine (e.g. `VIEW SERVER STATE` denied on one database in an ingest-all-databases run) is now logged and skipped instead of crashing the whole lineage step — matches the already-correct pattern in `usage_source.py`.
- `stored_procedures_mixin.yield_stored_procedure_queries()` (shared, `stored_procedures_mixin.py`): same guard as above, for stored-procedure query-history fetching. Uses `logger.warning()` + `continue` rather than `status.failed()`, for the same success-percentage reasoning as `get_stored_procedures`.

Each is a real, verified gap — reproduced live against a real SQL Server container via the actual metadata ingest/usage CLI before fixing, and re-verified live afterward where noted.

### Type of change:
- [x] Bug fix (non-breaking which fixes an issue)
- [x] Improvement (increases the code's efficiency/observability)

### Tests:
- Added/updated unit tests for every guard above (`test_mssql.py`, `test_query_log_truncation.py`, `test_stored_procedure_lineage.py`).
- `ruff check` / `ruff format --check` clean on all changed files.
- Full affected suite (mssql + lineage + stored-procedure + query-log-truncation tests): 115/115 passing.

<h2><a href="https://app.greptile.com/api/retrigger?id=55521665"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge.

The PR hardens MSSQL ingestion against connection, permission, and query-history failures while improving diagnostics.
- Adds pyodbc-compatible text fallbacks for Azure SQL throttling errors.
- Skips failing engines and optional metadata queries instead of aborting broader ingestion.
- Falls back to default date formatting when the MSSQL configuration probe fails.
- Adds focused regression tests for the new recovery and classification paths.
</details>

<sub>Reviews (11) · Last reviewed commit: ["Merge branch &#39;main&#39; into mssql-improveme..."](https://github.com/open-metadata/openmetadata/commit/f90795e164b8d2e75edcff05d65c584843da7ca5)</sub>


ingestion/src/metadata/ingestion/source/database/stored_procedures_mixin.py
```diff
@@ -91,9 +91,14 @@ def yield_stored_procedure_queries(self) -> Iterator[QueryByProcedure]:
         Yield query and stored procedure object for lineage processing.
         """
         for engine in self.get_stored_procedure_engines():
-            query = self.get_stored_procedure_sql_statement()
-            with engine.connect() as conn:
-                results = conn.execute(text(query)).all()
+            try:
+                query = self.get_stored_procedure_sql_statement()
```
IceS2, line 95: Should this be within the try?
I think this could hide a code issue and would falsely pass lineage source as Success without us knowing there is a code bug

ingestion/src/metadata/ingestion/source/database/lineage_source.py
```diff
@@ -311,31 +311,35 @@ def yield_table_query(self) -> Iterator[TableQuery]:
...
-                        logger.debug(traceback.format_exc())
-                        logger.warning(f"Error processing query_dict {query_dict}: {exc}")
-                logger.info(f"Processed {row_count} query log entries for lineage")
-                self.warn_if_query_log_truncated(row_count, "lineage")
+            try:
+                with engine.connect() as conn:  # pyright: ignore[reportOptionalMemberAccess]
+                    sql_statement = self.get_sql_statement(
+                        start_time=self.start,
+                        end_time=self.end,
+                    )
```
IceS2, line 319: Like below, should this be within the try? I feel this could hide real bugs that we want to catch easily

IceS2 (changes requested):
Added two comments.

I'd also check if we should catch that broad exception or narrow it and treat different exceptions differently

ulixius9 (approved):

PR #31861, opened 2026-08-21
fix(security): bump apache-airflow 3.3.0 -> 3.3.1 for CVE-2026-67587 / CVE-2026-54183

Khairajani:
CVE-2026-67587 (High) -- the Task SDK rebuilt a Serde `Callback` by re-running its
constructor, which imports the module named by the stored callback path. `SyncCallback`
is an Airflow class, so it passes the default `allowed_deserialization_classes`
allow-list and tightening that setting does not help. A Dag author controls a task
instance's `next_kwargs`, so they can get an arbitrary module imported inside the
scheduler process when the `awaiting_input` timeout sweep deserializes that value. No
non-default configuration is required.

CVE-2026-54183 (Medium) -- the secrets masker's recursion-depth limit did not descend
into values nested inside a list, tuple or set, so a Variable holding a deeply nested
sensitive value rendered unmasked in the Variables UI. Anyone who can see the Variable
in the UI can already read it through the Variables REST API, so this is a
shoulder-surfing defense rather than a disclosure boundary -- hence Medium.

Both are fixed in 3.3.1 and nothing in the 3.3.0 line is clean. The pin, both base
image tags, the vendored constraints file and the integration-test image move together,
as they did for 3.2.2 -> 3.3.0 in #31338.

The constraints file is a straight re-download this time. It carried two hand-patches
marked "keep on regeneration" -- impyla and thrift at 0.24.0 for CVE-2026-66053 /
CVE-2026-41608 / CVE-2026-48586, against upstream constraints-3.3.0 shipping
impyla==0.22.0 and thrift==0.16.0 -- and upstream constraints-3.3.1 now ships both at
0.24.0 natively. Diffing the vendored 3.3.0 file against upstream constraints-3.3.0
confirms those two blocks were the only divergence, and the new file is byte-identical
to upstream constraints-3.3.1 outside comments. The comments are kept, reworded to say
the pin is now upstream-native, so a future regeneration off a branch that regressed
either version still gets caught.

Two comments that name a constraint were checked rather than blind-renumbered:

- tests/integration/airflow/Dockerfile explains the constraints are deliberately not
  applied because they pin chardet against openmetadata-ingestion's chardet==4.0.0.
  Still true; the version moved 6.0.0.post1 -> 7.5.1, so the number is updated.
- Dockerfile.ci's universal-pathlib workaround cites Airflow's >=0.3.8 floor.
  apache-airflow-core 3.3.1 still declares `universal-pathlib>=0.3.8` and constraints
[...]

PR #5909, opened 2022-07-07
UI: Fix browser zoom breaks UI layout

vivekratnavel:
### Describe your changes :
For larger resolutions > 2000px the layout centers itself instead of occupying the full screen.

#
### Type of change :

- [x] Bug fix

#
### Frontend Preview (Screenshots) :

https://user-images.githubusercontent.com/1051198/177706639-cecae38f-1b05-4602-a648-1e0971b18298.mov

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/open-source-community/developer) document.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] All new and existing tests passed.

PR #31890, opened 2026-08-21
fix(security): airflow 3.3.1 (CVE-2026-67587 / CVE-2026-54183) and expat 2.8.3 (CVE-2026-72522)

Khairajani:
## What

Two OS/runtime security fixes for the ingestion images, one commit each:

| Commit | Fix | CVEs |
|---|---|---|
| `fix(security): pull expat 2.8.3-1~deb13u1 into the ingestion-base image` | expat `2.8.2` -> `2.8.3-1~deb13u1` in `ingestion/operators/docker/Dockerfile{,.ci}` | CVE-2026-72522 |
| `fix(security): bump apache-airflow 3.3.0 -> 3.3.1` | airflow pin, both base image tags, vendored constraints, integration-test image | CVE-2026-67587, CVE-2026-54183 |

These were previously #31757 and #31861. They are combined here because both are security fixes for the same image family with **disjoint file sets** — the expat commit touches only `ingestion/operators/docker/`, the airflow commit only `ingestion/Dockerfile{,.ci}`, `setup.py`, the constraints file and the integration-test harness. Nothing overlaps, so the two commits are reviewable independently. Both superseded PRs were green.

---

## Commit 1 — expat CVE-2026-72522

The image reports expat **2.8.2** for **CVE-2026-72522** (DSA-6446-1 — out-of-bounds read and resultant infinite loop in libexpat's `*_toUtf16` surrogate handling). trixie-security has `2.8.3-1~deb13u1`.

expat is **not in the `python:3.12-slim-trixie` base at all** — it arrives transitively from the top-of-file `apt-get install`. That layer is exactly why the image is stuck at 2.8.2: its cache key never changes, so it keeps resolving against the Debian index that was current when it was first built. Adding expat *there* would freeze the fix the same way. It goes in the late root layer instead, below the COPY and pip layers, where the index is re-read on every build — the same reasoning that put util-linux there.

expat rides in the existing dpkg source-package query rather than a second `RUN`, so one apt call covers both source packages and the next OS CVE is a one-line edit:

```sh
ul="$(dpkg-query -W -f='${source:Package} ${Package}\n' | awk '$1=="util-linux"{print $2}')"; \
[ -n "$ul" ] || { echo "no src:util-linux packages found; refusing to skip the CVE patch" >&2; exit 1; }; \
ex="$(dpkg-query -W -f='${source:Package} ${Package}\n' | awk '$1=="expat"{print $2}')"; \
apt-get -qq install -y --only-upgrade $ul $ex; \
```

The package set stays computed from dpkg rather than hand-listed, for the reason already documented there: scanners report each binary of a source package separately, so a hand-written list silently leaves one behind.

**util-linux keeps its non-empty assert and expat deliberately does not get one.** util-linux is Essential, so an empty query there means `dpkg-query` misbehaved and the build must stop. expat is transitive, so an empty query is a legitimate image with nothing to patch — asserting it would break the build for no reason. Asserting one of the two still covers the hazard the check exists for: `apt-get install --only-upgrade` with no package arguments exits 0, so an all-empty query would give a green build that shipped the vulnerable packages.

`Dockerfile.ci` carries its own copy of this block and gets the same change — it is the base that Collate's Snyk scan builds from.

**Verification.** Ran the `RUN` block verbatim in a real build layer on `python:3.12-slim-trixie`:

| package | before | after |
|---|---|---|
| `libexpat1` | `2.7.1-2` | `2.8.3-1~deb13u1` |
| all nine util-linux binaries | `2.41-5` | `2.41.5-0+deb13u1` |

Exit 0. The **expat-absent path** was exercised separately on the bare base and also exits 0, upgrading util-linux alone — that path is real, since the base ships no expat.

**Out of scope, deliberately.** `ingestion/Dockerfile` gets no equivalent expat change because there is **nothing to upgrade to**, not because it is clean. Its base is bookworm; Debian's tracker still marks bookworm *and* bookworm-security **vulnerable**, and the image already carries `2.5.0-1+deb12u2` — the newest build either suite offers. Needs revisiting when Debian ships a bookworm fix. Re-checked against the tag this PR moves to: `apache/airflow` 3.3.1 is still `debian:bookworm-slim`, byte-identical to 3.3.0 on every base-image line and with no trixie reference anywhere, so the airflow bump in commit 2 does not change this conclusion.

---

## Commit 2 — airflow CVE-2026-67587 / CVE-2026-54183

CVE-2026-67587 (High) — the Task SDK rebuilt a Serde `Callback` by re-running its constructor, which imports the module named by the stored callback path. `SyncCallback` is an Airflow class, so it passes the default `allowed_deserialization_classes` allow-list and tightening that setting does not help. A Dag author controls a task instance's `next_kwargs`, so they can get an arbitrary module imported inside the scheduler process when the `awaiting_input` timeout sweep deserializes that value. No non-default configuration is required.

CVE-2026-54183 (Medium) — the secrets masker's recursion-depth limit did not descend into values nested inside a list, tuple or set, so a Variable holding a deeply nested sensitive value rendered unmasked in the Variables UI. Anyone who can see the Variable in the UI can already read it through the Variables REST API, so this is a shoulder-surfing defense rather than a disclosure boundary — hence Medium.

Both are fixed in 3.3.1 and nothing in the 3.3.0 line is clean. The pin, both base image tags, the vendored constraints file and the integration-test image move together, as they did for 3.2.2 -> 3.3.0 in #31338.

The constraints file is a straight re-download this time. It carried two hand-patches marked "keep on regeneration" — impyla and thrift at 0.24.0 for CVE-2026-66053 / CVE-2026-41608 / CVE-2026-48586, against upstream constraints-3.3.0 shipping `impyla==0.22.0` and `thrift==0.16.0` — and upstream constraints-3.3.1 now ships both at 0.24.0 natively. Diffing the vendored 3.3.0 file against upstream constraints-3.3.0 confirms those two blocks were the only divergence, and the new file is byte-identical to upstream constraints-3.3.1 outside comments. The comments are kept, reworded to say the pin is now upstream-native, so a future regeneration off a branch that regressed either version still gets caught.

Two comments that name a constraint were checked rather than blind-renumbered:

- `tests/integration/airflow/Dockerfile` explains the constraints are deliberately not applied because they pin chardet against openmetadata-ingestion's `chardet==4.0.0`. Still true; the version moved `6.0.0.post1` -> `7.5.1`, so the number is updated.
- `Dockerfile.ci`'s universal-pathlib workaround cites Airflow's `>=0.3.8` floor. apache-airflow-core 3.3.1 still declares `universal-pathlib>=0.3.8` and constraints still pin `0.3.10`, so only the Airflow version in the prose changes.

**This commit also folds in the `croniter<3` removal** that was a second commit on #31861 — it is a prerequisite of the bump, not an unrelated change, which is why `setup.py` has a hunk in the dagster extra. apache-airflow-core raised its croniter floor from `>=2.0.2` to `>=6.2.2` between 3.3.0 and 3.3.1, so `uv pip install ingestion[all]` + `[test]` becomes unsatisfiable against the old ceiling. The ceiling was already dead weight: dagster 1.13.18 declares no croniter dependency at all, nothing under `ingestion/` imports croniter, and the one importer in the repo (openmetadata-airflow-apis) already runs croniter 6.2.x via the constraints file. Only the CI test environment resolved croniter 2.x, because it installs from the extras without constraints. Full rationale is in the commit message.

**Verification.** `uv pip compile --python-version 3.12 --extra airflow`: `apache-airflow==3.3.1` co-installs with every transitive floor pin in the airflow extra (providers-http, -opensearch, -elasticsearch, tornado, Werkzeug, starlette, python-multipart) with no conflict. `uv pip compile --extra all --extra test` on both 3.10 and 3.12: resolves to `apache-airflow==3.3.1`, `croniter==6.2.4`, `dagster==1.13.18`, no conflict. sqlparse resolves to 0.5.4 via collate-sqllineage, unchanged by this PR — the release images force 0.6.0 in `ingestion/operators/docker/Dockerfile`, which does not consume this constraints file.

---

## Related

- Supersedes #31757 (expat) and #31861 (airflow).
- Collate-side half of the expat fix: open-metadata/openmetadata-collate#5909 — the two must land together, or the collate image keeps shipping what the base tag was pinned to. Note that #5909 was scoped to the expat half only; it may want the airflow tag bump too now that both ride in one PR here.

The PR updates Airflow and Debian expat remediation paths while also migrating the ingestion dependency stack and associated Pydantic, NumPy, spaCy, PII, and test code.
- Moves ingestion images and constraints from Airflow 3.3.0 to 3.3.1.
- Upgrades installed expat source packages in the ingestion operator images.
- Updates major Python dependency ranges and compatibility call sites.
- Adjusts PII date recognition, deterministic scoring, fixtures, and CI model installation.
</details>

The PR appears safe to merge because no blocking failure remains within the eligible follow-up review scope.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/setup.py | Updates Airflow and several core scientific/runtime dependency ranges, removes the obsolete croniter ceiling, and aligns development stubs. |
| ingestion/airflow-constraints-3.3.1.txt | Replaces the Airflow 3.3.0 constraint snapshot with the substantially updated 3.3.1 dependency set. |
| ingestion/Dockerfile | Moves the production ingestion image and remote constraints URL to Airflow 3.3.1. |
| ingestion/Dockerfile.ci | Moves the CI image, vendored constraints references, and related compatibility documentation to Airflow 3.3.1. |
| ingestion/operators/docker/Dockerfile | Extends the late Debian security-upgrade layer to upgrade installed packages from the expat source package. |
| ingestion/operators/docker/Dockerfile.ci | Mirrors the expat source-package security upgrade in the operator CI image. |
| ingestion/src/metadata/ingestion/models/custom_pydantic.py | Migrates the inherited after-model validator to Pydantic 2.12's supported instance-method form. |
| ingestion/src/metadata/pii/tag_analyzer.py | Filters invalid DATE_TIME recognizer results in the content-analysis path before scores are aggregated. |
| .github/workflows/py-tests-shared.yml | Updates the pinned spaCy model release, checksum, and download source used by Python CI. |

</details>

```mermaid
flowchart LR
  Setup[setup.py dependency ranges] --> Images[Ingestion Docker images]
  Constraints[Airflow 3.3.1 constraints] --> Images
  Airflow[Airflow 3.3.1 base] --> Images
  Expat[Late apt source-package upgrade] --> OperatorImages[Operator images]
  Setup --> Runtime[Ingestion runtime]
  Runtime --> Models[Pydantic models]
  Runtime --> PII[PII and profiler paths]
  ModelWheel[spaCy model 3.8.0] --> CI[Python CI tests]
```
</details>

<sub>Reviews (6): Last reviewed commit: ["fix: filter invalid date recognizer resu..."](https://github.com/open-metadata/openmetadata/commit/32f5760d2193375100afd67e4c1c3e2433afdbc6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55545441)</sub>


mohittilala (dismissed):

IceS2:
Unit tests failing due to stale spacey model on the workflow on main. This PR updates it. You can check the workflow against this branch update here: https://github.com/open-metadata/OpenMetadata/actions/runs/32522341633

pmbrull (approved):

PR #31894, opened 2026-08-21
test(playwright): fix five AUT nightly flakes at their cause

harsh-vador:
Collate's AUT nightly run [32476747983](https://github.com/open-metadata/openmetadata-nightly/actions/runs/32476747983) retried 25 tests on each database lane (13 shared → 37 distinct specs). Fourteen of them live in this repo. Five have a cause the artifacts explain; the rest are listed at the bottom rather than guessed at.

Companion Collate-side PR: open-metadata/openmetadata-collate#6046.

## Fixes

| Spec | Observed | Cause | Change |
|---|---|---|---|
| `LineageInteraction` — node panel opens on click *(both lanes)* | `Test timeout of 60000ms exceeded`, no locator named | `clickLineageNode` clicked the node as soon as the caller's `getLineage` wait returned. React Flow mounts nodes in its own layout pass after that, so the click auto-waited with no timeout of its own and the failure surfaced as a bare test timeout. | Assert the node title is visible, then click. |
| `ServiceIngestion` — Create & Ingest BigQuery | `page.click` on `next-button` timed out; log shows the element visible/enabled/stable with `<span data-testid="alert-message">AutoPilot application is triggered…</span>` intercepting | Creating the service triggers AutoPilot; its toast renders bottom-center, directly over the wizard footer. | `waitForToastToDisappear(page, /AutoPilot/i)` before advancing. |
| `ExplorePageRightPanel_KnowledgeCenter` — remove user owner *(both lanes)* | owner chip never visible (60s) | The Explore summary panel renders owners from the search document, refreshed asynchronously after the owner PATCH. A panel rendered before that refresh will never show the chip — waiting on it waits on the wrong thing. | Re-open the entity until the chip is there. |
| `GlossaryHierarchy` — move term with children | `getByLabel('Select Parent')` click timed out (60s) | The label lookup is page-scoped and also matches the control of a hierarchy modal left in the DOM by an earlier step, so the click spent the test on a hidden element. | Scope to `change-parent-hierarchy-modal`; assert visible + enabled first. |
| `Glossary` — change language to Dutch | `getByRole('menuitem', { name: 'English - EN' })` click timed out (60s) | Click ran against an `ant-dropdown` mid enter-animation — precisely the case `waitForAntdPopupToSettle` was written for (the click point is computed against the 0.8-scaled menu). | Settle the popup, assert the item, then click. Both switches. |

No timeout was raised and no `test.slow()` was added.

## Deliberately not addressed

`AdvancedSearchSuggestions`, `SearchSettings`, `ContextCenterArticles`, `ContextCenterMemories`, `ServiceEntity`, `GlossaryP2Tests`, `SSOConfiguration` (beforeEach) all failed as bare 60s timeouts with no anchor and passed on retry in 2–15s. Marking them slow would hide a regression as easily as it would fix a flake — they need a trace first.

`PlatformLineage` (dead 30s `waitForResponse`) and `DataProductRename` are already fixed on `main` by #31736; both were still failing in the nightly because it builds the `2.0` branch, which does not have that commit. **#31736 is the single highest-value cherry-pick into `2.0`** — its commit message names the PlatformLineage and NotificationTemplate failures verbatim, and its `createOrFetch` change also clears the three duplicate-seed 409s from the same run.

## Verification

- `tsc --project playwright/tsconfig.json --noEmit` — no new errors (the two pre-existing `utils/glossary.ts` errors and the unused-`afterAction` one in `Glossary.spec.ts` are unchanged; verified against a clean tree).
- `eslint` on the touched files — 0 errors; remaining warnings are pre-existing and unrelated to these hunks.
- `organize-imports-cli` + `prettier --write` applied.
- Not executed against a live stack — these are AUT paths; the nightly is the verification.

The PR addresses several nightly test flakes by synchronizing Playwright actions with the UI state that makes them actionable, and also includes profiler fixture-ordering and domain/data-product navigation fixes.
- Bounds the ingestion wizard’s Next click so it retries through the transient AutoPilot toast.
- Reopens stale Explore panels, scopes glossary controls to the active modal, settles animated language menus, and waits for lineage nodes to become visible.
- Moves profiler execution into class fixture setup and re-runs it after profiler settings change.
- Makes domain and data-product name cells invoke the listing navigation handler without duplicating row click propagation.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/playwright/support/entity/ingestion/ServiceBaseClass.ts | The bounded click safely replaces the racy toast-presence wait because the awaited AutoPilot request precedes this step and its five-second toast fits within the click’s retry window. |
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/ExplorePageRightPanel_KnowledgeCenter.spec.ts | Adds retrying navigation so owner assertions read a refreshed search-backed summary panel. |
| openmetadata-ui/src/main/resources/ui/playwright/utils/glossary.ts | Scopes the parent selector to the visible hierarchy modal to avoid stale hidden controls. |
| openmetadata-ui/src/main/resources/ui/playwright/utils/lineage.ts | Waits explicitly for the React Flow node title before clicking it. |
| ingestion/tests/integration/profiler/test_sqa_profiler.py | Establishes profiles during class setup and refreshes them after changing metric settings so tests no longer depend on method order. |
| openmetadata-ui/src/main/resources/ui/src/components/common/atoms/domain/ui/domainFieldRenderers.tsx | Adds optional name-cell navigation while stopping propagation to prevent duplicate row navigation. |

</details>

<sub>Reviews (2): Last reviewed commit: ["test(playwright): bound the wizard Next ..."](https://github.com/open-metadata/openmetadata/commit/3cc69be6043051b34d0ec4fe93a4ce9a1571e7f5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55586931)</sub>

### DIFF
--- a/ingestion/tests/integration/profiler/test_sqa_profiler.py
+++ b/ingestion/tests/integration/profiler/test_sqa_profiler.py
@@ -19,7 +19,7 @@ No sample data is required beforehand
 import json
 import time
 from typing import List  # noqa: UP035
-from unittest import TestCase, TestLoader
+from unittest import TestCase
 
 from _openmetadata_testutils.ometa import int_admin_ometa
 from metadata.generated.schema.configuration.profilerConfiguration import (
@@ -39,8 +39,6 @@ from ..integration_base import (  # noqa: TID252
     METADATA_INGESTION_CONFIG_TEMPLATE,
     PROFILER_INGESTION_CONFIG_TEMPLATE,
 )
-
-TestLoader.sortTestMethodsUsing = None  # type: ignore
 
 
 class TestSQAProfiler(TestCase):
@@ -67,9 +65,42 @@ class TestSQAProfiler(TestCase):
                 ingestion_workflow.execute()
                 ingestion_workflow.raise_from_status()
                 ingestion_workflow.stop()
+
+            # Profile here rather than inside a test method. pytest orders TestCase
+            # methods alphabetically, so test_list_entity_profiles runs *before*
+            # test_profiler_workflow and would otherwise assert on profiles that do
+            # not exist yet. It only ever passed because hard-deleted tables used to
+            # leak their profiler rows into the window it queries (issue #27041, fixed
+            # in #31556). Profiles are class fixture data, so they belong here.
+            cls.run_profiler_workflows()
         except Exception as e:
             cls.container_builder.stop_all_containers()
             raise e  # noqa: TRY201
+
+    @classmethod
+    def run_profiler_workflows(cls):
+        """Run the profiler over every container, using the active profiler settings."""
+        for container in cls.container_builder.containers:
+            config = PROFILER_INGESTION_CONFIG_TEMPLATE.format(
+                type=container.connector_type,
+                service_config=container.get_config(),
+                service_name=type(container).__name__,
+            )
+            profiler_workflow = ProfilerWorkflow.create(json.loads(config))
+            profiler_workflow.execute()
+            profiler_workflow.print_status()
+            profiler_workflow.raise_from_status()
+            profiler_workflow.stop()
+
+    def list_profiled_tables(self):
+        """The tables the fixture ingested, across every container."""
+        tables: List[Table] = []  # noqa: UP006
+        for container in self.container_builder.containers:
+            service_name = type(container).__name__
+            cfg = json.loads(container.get_config())
+            db_name = cfg.get("database") or cfg.get("databaseSchema", "default")
+            tables.extend(self.metadata.list_all_entities(Table, params={"database": f"{service_name}.{db_name}"}))
+        return tables
 
     @classmethod
     def tearDownClass(cls):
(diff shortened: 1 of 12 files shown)


harsh-vador:
Good catch on the no-op — fixed in 3cc69be, though not the way suggested.

Waiting for the toast to become visible first has the mirror-image problem. The toast is fired by the create call several steps earlier (`showSuccessToast(t('message.auto-pilot-triggered-message'), 5000)` in `AddServicePage`) and auto-closes after 5s, so by the time `addIngestionPipeline` reaches the Next click it may equally have already closed — the agents tab, the add-ingestion dropdown, the workflow-form load and `fillIngestionDetails` all run in between. A visible-wait would then burn its bounded timeout on every healthy run, and gating the detach-wait on "the toast actually appeared" only moves the race rather than closing it. Either gate is wrong for one of the two orderings.

So the toast wait is gone entirely and the click is bounded instead:

```ts
await page.click('[data-testid="next-button"]', { timeout: 30_000 });
```

Playwright retries an intercepted click for the whole action timeout, and 30s outlasts the 5s toast in every ordering — not yet rendered, on screen now, or already gone. The original failure was that the click had *no* per-action timeout, so the retry loop ran to the end of the test instead of outliving the toast.

PR #31919, opened 2026-08-23
Fixes #24565: Datalake profiler size and created at

### LINKED ISSUE #24565: Datalake profiler size and created at
# Spec — `sizeInByte` + `createDateTime` for the Datalake (pandas) profiler

**Status:** draft · **Scope:** `PandasProfilerInterface` (Datalake / object-store profiler type) only
**Author:** drafted from codebase inspection at `build/olivaw-runner-image` (2026-08-20)

---

## 1. Goal

`TableProfile.sizeInByte` and `TableProfile.createDateTime` are populated today only by the SQLAlchemy
profiler, via the per-dialect computers in
`ingestion/src/metadata/profiler/orm/functions/table_metric_computer.py`. The Datalake profiler
(`PandasProfilerInterface`) never emits them, so every Datalake table profile shows a blank size and
no creation date in the UI.

Make the Datalake profiler emit both, sourced from object-store metadata (S3 / GCS / Azure Blob),
using the same wire contract the SQA path already uses.

## 2. Current state (verified)

| Piece | Where | State |
|---|---|---|
| Schema fields | `openmetadata-spec/.../entity/data/table.json:917-924` | `sizeInByte` (number), `createDateTime` (date-time) — already exist, no schema change needed |
| Sink mapping | `ingestion/src/metadata/profiler/processor/core.py:510-522` | Already reads `_table_results["sizeInBytes"]` → `TableProfile.sizeInByte` and `_table_results["createDateTime"]` → `createDateTime` |
| SQA producer | `.../sqlalchemy/profiler_interface.py:192-229` | Returns `row._asdict()` from `TableMetricComputer`; the row carries `rowCount`, `columnCount`, `columnNames`, **plus** `sizeInBytes` / `createDateTime` as extra keys |
| Pandas producer | `.../pandas/profiler_interface.py:158-182` | Returns only `{metric.name(): value}` — **no size, no created-at** |
| UI render | `openmetadata-ui/.../TableProfiler/TableProfilerProvider.tsx:169` | `bytesToSize(profile?.sizeInByte ?? 0)` — the value is **bytes**, despite the schema description saying "Table size in GB" (stale description, out of scope) |

**Key insight:** size/created-at are *not* profiler `Metric` classes anywhere in this codebase. They
are extra keys merged into the table-metrics dict. This spec follows that existing pattern rather than
introducing new `Metric` subclasses — consistent with `docs/design-patterns.md` ("extend the
established pattern").

### Available raw material

- `DatalakeSampler.client` (`.../sampler/pandas/sampler.py:52`) → `get_ssl_connection(...)` →
  `get_connection(...)` → `owned.client`, i.e. a **`DatalakeBaseClient`**
  (`.../datalake/clients/base.py`). `PandasProfilerInterface.client` (line 89) is that same object;
  `self.client.client` is the raw boto3 / `google.cloud.storage.Client` / `BlobServiceClient`.
- `DatalakeBaseClient.get_table_names()` already returns `(key, size_bytes)` tuples
  (`clients/base.py:52-58`, S3 `clients/s3.py:87`, GCS `clients/gcs.py:120`, Azure
  `clients/azure_blob.py:74`) — but that's a *listing* API used by the metadata source, not reachable
  per-table from the profiler without re-listing the bucket.
- `DataFrameReader._get_file_size_mb` (`readers/dataframe/base.py:79-108`) already does exactly the
  per-config-source size lookup we need (S3 `head_object`, GCS `get_blob`, Azure
  `get_blob_properties`, Local `os.path.getsize`) — but returns MB, and no timestamp. **This is the
  shape to generalize.**
- Bucket + key for a Datalake table are resolved the same way `get_dataframes` does
  (`mixins/pandas/pandas_mixin.py:186-195`): `bucket = table.databaseSchema.name`,
  `key = table.name.root`.

## 3. Non-goals

- SQA dialects that are still missing size/created-at (Postgres, Timescale, ClickHouse, Cockroach,
  Vertica have size but no created-at; base-fallback + Trino/Hive/Impala/Databricks have neither).
- The NoSQL profiler.
- Container / Storage-service entities (`sampler/storage/**`) — different entity type.
- Fixing the stale `"Table size in GB"` description on `TableProfile.sizeInByte`.
- Backfilling historical profiles.

---

### 4.1 `sizeInByte`
The on-store byte length of the object backing the table. One Datalake table == one object key
(`DatalakeSource.get_tables_name_and_type` skips keys ending in `/`, and
`fetch_dataframe_generator` refuses them), so no summation across parts is needed. Compressed size
as stored — matching the SQA path, which reports storage bytes, not logical/uncompressed bytes.

### 4.2 `createDateTime`

| Store | Field used | True creation? |
|---|---|---|
| GCS | `blob.time_created` | ✅ yes |
| Azure Blob | `BlobProperties.creation_time`, fall back to `last_modified` | ✅ yes (fallback: no) |
| S3 / MinIO / S3-compatible | `head_object()["LastModified"]` | ❌ **no — S3 has no creation timestamp** |

**Decision D2 — map S3 `LastModified` → `createDateTime`.** S3 exposes no creation time at all. For
immutable datalake objects (the overwhelmingly common case) first-write == `LastModified` until the
object is overwritten; for overwritten objects the value tracks the current version's write time,
which is arguably the "creation" of the table as it now exists. The alternative — leave
`createDateTime` null on S3 — makes the feature useless for the most common Datalake backend.
*This is the one judgement call in the spec; flipping it is a one-line change in `_stats_for_s3`.*
Document the approximation in the connector docs and in a code comment (this is exactly the kind of
non-obvious semantics the repo's comment rule says to comment).

### 4.3 Timezone
All values are normalized to **timezone-aware UTC** in the helper.

This matters: `Profiler.get_profile()` (`processor/core.py:511-512`) does
`raw_create_date.replace(tzinfo=timezone.utc)` — a *replace*, not a convert. A tz-aware non-UTC value
arriving there would be silently corrupted. Returning tz-aware UTC makes that call a no-op.
**Do not** "fix" `core.py` to use `astimezone` — SQA dialects return naive datetimes there, and
`astimezone` on a naive datetime assumes local time, which would regress every SQL connector.

---

### 5.1 New: object-stat model + helper

**New file** `ingestion/src/metadata/utils/datalake/object_stats.py`
(Collate Community License header — copy from a sibling under `ingestion/`, **not** Apache).

```python
class ObjectStats(BaseModel):
    """Storage-level stats for a single object backing a Datalake table."""
    size_in_bytes: Optional[int] = None
    create_date_time: Optional[datetime] = None   # tz-aware UTC


@singledispatch
def get_object_stats(config_source, client, bucket_name: str, key: str) -> ObjectStats:
    """Unknown config source (e.g. BurstIQ) -> no stats, no error."""
    return ObjectStats()


@get_object_stats.register
def _(_: S3Config, client, bucket_name: str, key: str) -> ObjectStats:
    response = client.head_object(Bucket=bucket_name, Key=key)
    # S3 has no creation timestamp; LastModified is the closest available signal.
    return ObjectStats(
        size_in_bytes=response.get("ContentLength"),
        create_date_time=_as_utc(response.get("LastModified")),
    )


@get_object_stats.register
def _(_: GCSConfig, client, bucket_name: str, key: str) -> ObjectStats:
    # bucket.blob() does not fetch properties; get_blob() issues the metadata GET.
    blob = client.get_bucket(bucket_name).get_blob(key)
    if blob is None:
        return ObjectStats()
    return ObjectStats(size_in_bytes=blob.size, create_date_time=_as_utc(blob.time_created))


@get_object_stats.register
def _(_: AzureConfig, client, bucket_name: str, key: str) -> ObjectStats:
    props = client.get_blob_client(container=bucket_name, blob=key).get_blob_properties()
    return ObjectStats(
        size_in_bytes=props.size,
        create_date_time=_as_utc(props.creation_time or props.last_modified),
    )
```

`_as_utc(value)` → `None` if falsy, else `value.astimezone(timezone.utc)` for tz-aware input and
`value.replace(tzinfo=timezone.utc)` for naive input (all three SDKs return tz-aware UTC today; the
naive branch is defensive).

Uses `@singledispatch` on the config source — the same dispatch idiom already used for readers
(`readers/dataframe/json.py:204`, `parquet.py:124`).

**`LocalConfig` is deliberately not implemented.** `DatalakeConnection._get_client`
(`.../datalake/connection.py:70`) raises `NotImplementedError` for anything that isn't
S3/GCS/Azure, so no Local Datalake connection can reach the profiler at all. Adding a branch would
be dead code. If Local support returns, it's a 5-line `os.stat` register (`st_size`, and
`st_birthtime` where available else `st_mtime`).

### 5.2 New: shared table-metric key constants

`"createDateTime"` is currently defined twice — `processor/core.py:62` and
`orm/functions/table_metric_computer.py:66` — and `"sizeInBytes"` only in the SQA module, with
`core.py:522` using a bare string literal. The pandas interface must not import from
`profiler/orm/**` (SQA-land; the repo enforces import layering — see commit `8a5106e2e9`).

**New file** `ingestion/src/metadata/profiler/constants.py`:

```python
ROW_COUNT = "rowCount"
COLUMN_COUNT = "columnCount"
COLUMN_NAMES = "columnNames"
SIZE_IN_BYTES = "sizeInBytes"
CREATE_DATETIME = "createDateTime"
```

Re-point `core.py`, `table_metric_computer.py`, and the new pandas code at it. Pure move, no
behaviour change.

### 5.3 Wiring: `PandasProfilerInterface._compute_table_metrics`

`.../profiler/interface/pandas/profiler_interface.py:158`. Merge the stats into the dict the method
already returns, mirroring how the SQA row carries extra keys:

```python
try:
    row_dict = {}
    for metric in metrics:
        row_dict[metric.name()] = metric().df_fn(runner)
    row_dict.update(self._get_object_stats())
    return row_dict
except Exception as exc:
    ...  # unchanged
```

```python
def _get_object_stats(self) -> Dict[str, Any]:
    """Storage-level table stats. Best-effort: a stat failure must not fail the profile."""
    try:
        stats = get_object_stats(
            self.service_connection_config.configSource,
            self.client.client,
            self.table_entity.databaseSchema.name,
            self.table_entity.name.root,
        )
    except Exception as exc:
        logger.warning(f"Could not fetch object stats for {self.table_entity.name.root}: {exc}")
        return {}
    return {
        k: v
        for k, v in (
            (SIZE_IN_BYTES, stats.size_in_bytes),
            (CREATE_DATETIME, stats.create_date_time),
        )
        if v is not None
    }
```

**Failure isolation is a hard requirement.** The existing `_compute_table_metrics` re-raises as
`RuntimeError`, which aborts the whole table's profile. A missing `s3:HeadObject` permission must
degrade to "no size" — never to "no profile". Hence the inner try/except and the `.update()` placed
*after* the metric loop.

**Null-omission is deliberate**: `_table_results.get(...)` in `core.py` already yields `None` for
absent keys, and omitting beats writing explicit nulls into the row dict.

### 5.4 Why here, and what it inherits

- **Gating comes for free.** `_prepare_table_metrics` (`processor/core.py:297-303`) returns `[]` when
  `source_config.computeTableMetrics` is false, so no `MetricTypes.Table` batch is scheduled and
  `_compute_table_metrics` is never called. Size/created-at therefore respect
  `computeTableMetrics` with no extra code — identical to SQA.
  *Known consequence:* if a user narrows `metrics:` to column-only metrics, `table_metrics` is empty,
  the Table batch is dropped by `MetricFilter.filter_empty_metrics`, and no stats are emitted. This
  matches SQA behaviour exactly; not worth diverging.
- **BurstIQ is safe by construction.** `BurstIQProfilerInterface` extends `PandasProfilerInterface`
  and `BurstIQSampler.client` is a `BurstIQClient`, not a `DatalakeBaseClient`
  (`sampler/pandas/burstiq/sampler.py:75`). Its `service_connection_config.configSource` is not an
  S3/GCS/Azure config, so `get_object_stats` hits the `@singledispatch` base and returns empty stats.
  **No BurstIQ override needed** — verify with a unit test rather than trusting it.

### 5.5 Cost

One `HEAD` (or equivalent metadata GET) per table per profiler run. Negligible next to the full
object read the profiler already performs for every table. **No cache is added** — the call happens
once per table per run, so a cache would only add the unbounded-growth risk that `CLAUDE.md` forbids.

### 5.6 Optional cleanup (recommended, separable commit)

Reimplement `DataFrameReader._get_file_size_mb` (`readers/dataframe/base.py:79`) on top of
`get_object_stats`, killing the duplicated per-source dispatch:

```python
def _get_file_size_mb(self, key, bucket_name, file_size=None) -> float:
    if file_size is not None:
        return file_size / BYTES_PER_MB
    try:
        stats = get_object_stats(self.config_source, self.client, bucket_name, key)
    except Exception as exc:
        logger.debug(f"Could not determine file size for {key}: {exc}")
        return 0
    return (stats.size_in_bytes or 0) / BYTES_PER_MB
```

This also fixes a **latent bug**: today, if none of the `isinstance` branches match, the method falls
off the end of the `try` and returns `None`, and `json.py:185` then evaluates `None > 50.0` →
`TypeError`. The rewrite returns `0` on every path.
It would need `LocalConfig` registered on `get_object_stats` after all (the reader path does support
Local even though the Datalake *client* does not) — so if this cleanup is taken, add the Local
register from §5.1.

---

## 6. Edge cases

| Case | Behaviour |
|---|---|
| Object deleted between listing and profiling | SDK raises → warning logged, stats omitted, profile still written |
| No `HeadObject` / `storage.objects.get` permission | Same — warning, stats omitted, profile succeeds |
| Table name is an MD5 hash (key > 256 chars, `datalake/metadata.py:284`) | Key lookup fails. **Pre-existing**: `get_dataframes` uses `table.name.root` too, so the profiler already cannot read those tables at all. No special handling; same failure mode, now with a warning. Do **not** fall back to `displayName` — users can edit it freely |
| S3-compatible stores (MinIO, Ceph) | `head_object` is standard S3 API; works |
| Glacier / cold-storage objects | `head_object` returns metadata for archived objects, so size/`LastModified` still resolve even though the body is unreadable |
| Size is `0` (empty file) | Emitted as `0`, not omitted — `0` is a real size. `if v is not None` in §5.3 is deliberate, `if v` would be wrong |

---

## 7. Test plan

Per `CLAUDE.md` ("prefer integration tests over heavily-mocked unit tests"; mocks are for boundaries).

### Integration — real S3 API against MinIO (primary evidence)
`ingestion/tests/integration/datalake/test_datalake_profiler_e2e.py`
(MinIO container already wired via `integration/containers.py`, see `datalake/conftest.py:31`).

1. Extend `test_datalake_profiler_workflow`, or add a sibling test, that after the profiler run
   fetches the profile and asserts:
   - `profile.sizeInByte == len(<bytes uploaded for profiler_test_.csv>)` (exact — MinIO stores the
     payload verbatim, so this is deterministic; `> 0` only as a fallback if flakiness appears)
   - `profile.createDateTime is not None` and falls inside `[test_start, now]`
   - `profile.createDateTime.tzinfo` is UTC
2. Regression guard: `rowCount` / `columnCount` on the same profile still match today's values —
   proves `row_dict.update(...)` doesn't clobber metric keys.

### Unit — `ingestion/tests/unit/observability/profiler/pandas/test_object_stats.py` (new)
Boundaries only (no GCS/Azure emulator exists):

1. `get_object_stats` with a stub boto3 client → correct `size_in_bytes` and UTC `create_date_time`.
2. Same for `GCSConfig` (mock `client.get_bucket(...).get_blob(...)`) and `AzureConfig`
   (mock `get_blob_client(...).get_blob_properties()`).
3. Azure with `creation_time=None` → falls back to `last_modified`.
4. GCS `get_blob` returns `None` (missing object) → empty `ObjectStats`, no raise.
5. Timezone: provider returns `+02:00` → result is the equivalent instant in UTC (guards the
   `replace` vs `astimezone` trap in §4.3).
6. Unknown config source (a bare `BaseModel`) → empty `ObjectStats` — the BurstIQ guarantee.

### Unit — extend `.../pandas/test_profiler_interface.py`
7. `_compute_table_metrics` merges `sizeInBytes` / `createDateTime` alongside `rowCount` /
   `columnCount`.
8. `get_object_stats` raising → `_compute_table_metrics` still returns the metric keys and does not
   raise (the failure-isolation requirement in §5.3).
9. Stats with both fields `None` → neither key present in the dict.

**RED first.** Per `feedback_prove_tests_fail_first`: run 1, 7, and 8 against unmodified `main` and
confirm each fails for the expected reason before writing the implementation.

---

## 8. Files touched

| File | Change |
|---|---|
| `ingestion/src/metadata/utils/datalake/object_stats.py` | **new** — `ObjectStats` + `get_object_stats` singledispatch |
| `ingestion/src/metadata/profiler/constants.py` | **new** — shared table-metric key constants |
| `ingestion/src/metadata/profiler/interface/pandas/profiler_interface.py` | `_get_object_stats()` + one `.update()` in `_compute_table_metrics` |
| `ingestion/src/metadata/profiler/processor/core.py` | import `CREATE_DATETIME` / `SIZE_IN_BYTES` from the new constants module; drop the local literal |
| `ingestion/src/metadata/profiler/orm/functions/table_metric_computer.py` | import the shared constants; drop local duplicates |
| `ingestion/tests/unit/observability/profiler/pandas/test_object_stats.py` | **new** |
| `ingestion/tests/unit/observability/profiler/pandas/test_profiler_interface.py` | 3 new cases |
| `ingestion/tests/integration/datalake/test_datalake_profiler_e2e.py` | size/created-at assertions |
| `ingestion/src/metadata/readers/dataframe/base.py` | *optional §5.6* — rewrite `_get_file_size_mb` |

No JSON-schema change → **no `make generate` needed**. No Java, no UI, no migration.

TeddyCr:
> **This pull request was opened by an olivaw agent.** It is an input to review, not an approved change — read the diff, not this description.
>
> Fixes #24565

## What olivaw checked (machine-observed, §8.2/A.1)

- diff: non-empty
- build: not_run
- tests: not_run
- base: `a7242502a4d462d88f4c892f80106c9157f11470`
- head: `4c22aa07972e2c5b540499e3da73d2b9e33eecf3`

**No build or test command is configured for this workspace**, so only the diff's existence was checked. Set `run_config.build_cmd`/`test_cmd` to have olivaw verify its own work.

## What the agent says it did (unverified)

> Done. Committed as `4c22aa07972` on `agent/issue-24565`.
> 
> ## What changed
> 
> **New `ingestion/src/metadata/utils/datalake/object_stats.py`** — `ObjectStats` model plus a `get_object_stats` `@singledispatch` on the Datalake config source: `head_object` for S3, `get_bucket().get_blob()` for GCS, `get_blob_properties()` for Azure (with `creation_time or last_modified`). All timestamps normalized to tz-aware UTC by `_as_utc`, so the `replace(tzinfo=utc)` relabel in `Profiler.get_profile` is a no-op instead of a silent shift. Unknown config sources return empty stats. `LocalConfig` is not registered — `DatalakeConnection._get_client` raises for it, so it would be dead code.
> 
> **`PandasProfilerInterface`** — `_compute_table_metrics` now does `row_dict.update(self._get_object_stats())`. The stat lookup has its own try/except so a missing `s3:HeadObject` degrades to "no size", not "no profile"; `None` values are omitted, but `0` is kept as a real size.
> 
> **New `ingestion/src/metadata/profiler/constants.py`** — the five table-metric keys were defined in `table_metric_computer.py`, redefined in `core.py`, and typed as bare literals in `core.py`'s `TableProfile` construction. Now one module; `core.py`, `table_metric_computer.py`, and `sampler/sqlalchemy/sampler.py` import from it.
> 
> ## Deviations from the spec
> 
> - **BurstIQ needed a guard after all.** The spec asserted `BurstIQConnection.configSource` is simply not an S3/GCS/Azure config; it has no `configSource` field at all, so the plain attribute access would have raised `AttributeError` on every BurstIQ table — caught, but logging a warning per table. Uses `getattr(..., "configSource", None)` instead, which dispatches cleanly to the base. Locked in by a new test in `test_burstiq_profiler_interface.py`.
> - **Skipped the optional §5.6 cleanup** (reimplementing `DataFrameReader._get_file_size_mb` on `get_object_stats`) to keep the change scoped. That latent bug is still there: if no `isinstance` branch matches, the method returns `None` and `json.py:185` does `None > 50.0` → `TypeError`.
> - **No docs change** — the Datalake connector docs aren't in this repo. The S3 `LastModified` approximation is documented in a code comment.
> 
> ## Tests
> 
> 13 new unit tests (9 in `test_object_stats.py`, 3 in `test_profiler_interface.py`, 1 BurstIQ guard). RED-first confirmed: with the source changes stashed, the 3 interface tests fail on `does not have the attribute 'get_object_stats'`.
> 
> `tests/unit/observability/profiler` passes: 517 passed, with 2 failures (`exasol`/`teradata` dialect tests) that are missing-driver failures unrelated to this change — `tests/unit/readers`' 25 failures are likewise identical on a stashed baseline.
> 
> **I could not run the integration test** (`test_datalake_profiler_e2e.py`) — it needs Docker/MinIO and a running OpenMetadata server, neither available here. Its new assertions (`sizeInByte == profiler_test_.csv` on-disk size, UTC `createDateTime`, and `rowCount`/`columnCount` unchanged as the regression guard) are written and the file compiles, but they are unverified.
> 
> The commit used `--no-verify`: pre-commit isn't installed in this clone. I ran the Python half of that gate manually — `ruff check` and `ruff format --check` at the pinned v0.15.12 against `ingestion/pyproject.toml`, both clean.

The PR adds provider-specific object-stat retrieval so Datalake profiles can include object size and creation time, while centralizing table metric keys.
- Adds S3, GCS, and Azure implementations for retrieving object metadata.
- Merges storage metadata into Pandas table-profile results with best-effort error handling.
- Adds unit and integration coverage for provider responses, metric merging, UTC normalization, and non-object-store profiler interfaces.
</details>

The PR is not yet safe to merge because Datalake profiling still cannot resolve the physical storage key for object paths transformed into MD5-derived table names.

For object paths longer than 256 characters, ingestion stores the MD5-derived value in Table.name, while both the dataframe reader and the new object-stat lookup use that catalog name as the provider key instead of the original path.

**Files Needing Attention:** ingestion/src/metadata/profiler/interface/pandas/profiler_interface.py, ingestion/src/metadata/mixins/pandas/pandas_mixin.py
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/utils/datalake/object_stats.py | Introduces provider-dispatched retrieval and UTC normalization for Datalake object size and timestamps. |
| ingestion/src/metadata/profiler/interface/pandas/profiler_interface.py | Merges best-effort object statistics into Pandas table metrics, but the previously reported long-object-key handling remains unresolved. |
| ingestion/src/metadata/profiler/processor/core.py | Replaces duplicated table-profile metric literals with shared constants. |
| ingestion/tests/unit/observability/profiler/pandas/test_object_stats.py | Covers provider-specific object-stat retrieval, missing objects, and timestamp normalization. |
| ingestion/tests/integration/datalake/test_datalake_profiler_e2e.py | Extends the Datalake profiler integration test with size, UTC timestamp, and existing-metric assertions. |

</details>

```mermaid
flowchart LR
  T[Datalake table entity] --> P[Pandas profiler interface]
  P --> M[Compute dataframe metrics]
  P --> S[Fetch object statistics]
  S --> C{Config source}
  C -->|S3| A[HeadObject]
  C -->|GCS| G[Get blob]
  C -->|Azure| Z[Get blob properties]
  M --> R[Table metric results]
  A --> R
  G --> R
  Z --> R
  R --> O[TableProfile]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["refactor(profiler): tighten typing and c..."](https://github.com/open-metadata/openmetadata/commit/49f1e77d9820f299f1792f45b8f6a8e15599e0aa) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55963872)</sub>

### DIFF
--- a/ingestion/src/metadata/utils/datalake/object_stats.py
+++ b/ingestion/src/metadata/utils/datalake/object_stats.py
@@ -73,8 +73,10 @@ def _(_: S3Config, client: Any, bucket_name: str, key: str) -> ObjectStats:
 
 @get_object_stats.register
 def _(_: GCSConfig, client: Any, bucket_name: str, key: str) -> ObjectStats:
-    # `bucket.blob()` only builds a reference; `get_blob()` is the call that fetches the properties.
-    blob = client.get_bucket(bucket_name).get_blob(key)
+    # `bucket()` only builds a reference, so it needs no `storage.buckets.get` permission --
+    # unlike `get_bucket()`. `get_blob()` then fetches the properties with object-level read access,
+    # which the profiler already holds to read the object itself.
+    blob = client.bucket(bucket_name).get_blob(key)
     if blob is None:
         return ObjectStats()
     return ObjectStats(size_in_bytes=blob.size, create_date_time=_as_utc(blob.time_created))
--- a/ingestion/src/metadata/utils/datalake/object_stats.py
+++ b/ingestion/src/metadata/utils/datalake/object_stats.py
@@ -15,9 +15,9 @@ Storage-level stats of the object backing a Datalake table
 
 from datetime import datetime, timezone
 from functools import singledispatch
-from typing import Any, Optional
+from typing import TYPE_CHECKING, Any
 
-from pydantic import BaseModel
+from pydantic import BaseModel, field_validator
 
 from metadata.generated.schema.entity.services.connections.database.datalake.azureConfig import (
     AzureConfig,
@@ -29,63 +29,75 @@ from metadata.generated.schema.entity.services.connections.database.datalake.s3C
     S3Config,
 )
 
+if TYPE_CHECKING:
+    from azure.storage.blob import BlobServiceClient
+    from google.cloud.storage import Client as GCSStorageClient
+
 
 class ObjectStats(BaseModel):
     """Storage-level stats for a single object backing a Datalake table."""
 
-    size_in_bytes: Optional[int] = None  # noqa: UP045
-    create_date_time: Optional[datetime] = None  # tz-aware UTC  # noqa: UP045
+    size_in_bytes: int | None = None
+    create_date_time: datetime | None = None
 
+    @field_validator("create_date_time")
+    @classmethod
+    def normalize_to_utc(cls, value: datetime | None) -> datetime | None:
+        """Normalize a provider timestamp to timezone-aware UTC.
 
-def _as_utc(value: Optional[datetime]) -> Optional[datetime]:  # noqa: UP045
-    """Normalize a provider timestamp to timezone-aware UTC.
-
-    `Profiler.get_profile` relabels the value with `replace(tzinfo=utc)` rather than converting it,
-    so a non-UTC timestamp leaving this module would end up silently shifted on the profile.
-    """
-    if not value:
-        return None
-    if value.tzinfo is None:
-        return value.replace(tzinfo=timezone.utc)
-    return value.astimezone(timezone.utc)
+        `Profiler.get_profile` relabels the value with `replace(tzinfo=utc)` rather than converting
+        it, so a non-UTC timestamp leaving this model would end up silently shifted on the profile.
+        Validating here rather than at each call site means no provider below can forget.
+        """
+        if value is None:
+            return None
+        if value.tzinfo is None:
+            return value.replace(tzinfo=timezone.utc)
+        return value.astimezone(timezone.utc)
 
 
 @singledispatch
-def get_object_stats(config_source: Any, client: Any, bucket_name: str, key: str) -> ObjectStats:
+def get_object_stats(config_source: object, client: object, bucket_name: str, key: str) -> ObjectStats:
     """Fetch the size and creation time of the object backing a Datalake table.
 
-    Config sources that are not object stores (e.g. BurstIQ) get empty stats rather than an error.
+    Config sources that are not object stores (e.g. BurstIQ) get empty stats rather than an error,
+    hence the `object` rather than a union of the three supported configs.
     """
     return ObjectStats()
 
 
-@get_object_stats.register
+# The config class is passed to `register` explicitly rather than read off the annotation:
+# `singledispatch` resolves a function's annotations at registration time, which would blow up on
+# the `TYPE_CHECKING`-only client types below.
+@get_object_stats.register(S3Config)
 def _(_: S3Config, client: Any, bucket_name: str, key: str) -> ObjectStats:
+    # boto3 builds its clients at runtime, so `head_object` exists on no static type: `BaseClient`
+    # would type-check the parameter at the cost of an unresolved attribute on the line below.
     response = client.head_object(Bucket=bucket_name, Key=key)
     # S3 exposes no creation time at all. Datalake objects are typically written once, so
     # LastModified is the closest available signal; on overwritten objects it tracks the write time
     # of the current version.
     return ObjectStats(
         size_in_bytes=response.get("ContentLength"),
-        create_date_time=_as_utc(response.get("LastModified")),
+        create_date_time=response.get("LastModified"),
     )
 
 
-@get_object_stats.register
-def _(_: GCSConfig, client: Any, bucket_name: str, key: str) -> ObjectStats:
+@get_object_stats.register(GCSConfig)
+def _(_: GCSConfig, client: "GCSStorageClient", bucket_name: str, key: str) -> ObjectStats:
     # `bucket()` only builds a reference, so it needs no `storage.buckets.get` permission --
     # unlike `get_bucket()`. `get_blob()` then fetches the properties with object-level read access,
     # which the profiler already holds to read the object itself.
     blob = client.bucket(bucket_name).get_blob(key)
     if blob is None:
         return ObjectStats()
-    return ObjectStats(size_in_bytes=blob.size, create_date_time=_as_utc(blob.time_created))
+    return ObjectStats(size_in_bytes=blob.size, create_date_time=blob.time_created)
 
 
-@get_object_stats.register
-def _(_: AzureConfig, client: Any, bucket_name: str, key: str) -> ObjectStats:
+@get_object_stats.register(AzureConfig)
+def _(_: AzureConfig, client: "BlobServiceClient", bucket_name: str, key: str) -> ObjectStats:
     props = client.get_blob_client(container=bucket_name, blob=key).get_blob_properties()
     return ObjectStats(
         size_in_bytes=props.size,
-        create_date_time=_as_utc(props.creation_time or props.last_modified),
+        create_date_time=props.creation_time or props.last_modified,
     )
--- a/ingestion/src/metadata/utils/datalake/object_stats.py
+++ b/ingestion/src/metadata/utils/datalake/object_stats.py
@@ -1,0 +1,89 @@
+#  Copyright 2025 Collate
+#  Licensed under the Collate Community License, Version 1.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/LICENSE
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""
+Storage-level stats of the object backing a Datalake table
+"""
+
+from datetime import datetime, timezone
+from functools import singledispatch
+from typing import Any, Optional
+
+from pydantic import BaseModel
+
+from metadata.generated.schema.entity.services.connections.database.datalake.azureConfig import (
+    AzureConfig,
+)
+from metadata.generated.schema.entity.services.connections.database.datalake.gcsConfig import (
+    GCSConfig,
+)
+from metadata.generated.schema.entity.services.connections.database.datalake.s3Config import (
+    S3Config,
+)
+
+
+class ObjectStats(BaseModel):
+    """Storage-level stats for a single object backing a Datalake table."""
+
+    size_in_bytes: Optional[int] = None  # noqa: UP045
+    create_date_time: Optional[datetime] = None  # tz-aware UTC  # noqa: UP045
+
+
+def _as_utc(value: Optional[datetime]) -> Optional[datetime]:  # noqa: UP045
+    """Normalize a provider timestamp to timezone-aware UTC.
+
+    `Profiler.get_profile` relabels the value with `replace(tzinfo=utc)` rather than converting it,
+    so a non-UTC timestamp leaving this module would end up silently shifted on the profile.
+    """
+    if not value:
+        return None
+    if value.tzinfo is None:
+        return value.replace(tzinfo=timezone.utc)
+    return value.astimezone(timezone.utc)
+
+
+@singledispatch
+def get_object_stats(config_source: Any, client: Any, bucket_name: str, key: str) -> ObjectStats:
+    """Fetch the size and creation time of the object backing a Datalake table.
+
+    Config sources that are not object stores (e.g. BurstIQ) get empty stats rather than an error.
+    """
+    return ObjectStats()
+
+
+@get_object_stats.register
+def _(_: S3Config, client: Any, bucket_name: str, key: str) -> ObjectStats:
+    response = client.head_object(Bucket=bucket_name, Key=key)
+    # S3 exposes no creation time at all. Datalake objects are typically written once, so
+    # LastModified is the closest available signal; on overwritten objects it tracks the write time
+    # of the current version.
+    return ObjectStats(
+        size_in_bytes=response.get("ContentLength"),
+        create_date_time=_as_utc(response.get("LastModified")),
+    )
+
+
+@get_object_stats.register
+def _(_: GCSConfig, client: Any, bucket_name: str, key: str) -> ObjectStats:
+    # `bucket.blob()` only builds a reference; `get_blob()` is the call that fetches the properties.
+    blob = client.get_bucket(bucket_name).get_blob(key)
+    if blob is None:
+        return ObjectStats()
+    return ObjectStats(size_in_bytes=blob.size, create_date_time=_as_utc(blob.time_created))
+
+
+@get_object_stats.register
+def _(_: AzureConfig, client: Any, bucket_name: str, key: str) -> ObjectStats:
+    props = client.get_blob_client(container=bucket_name, blob=key).get_blob_properties()
+    return ObjectStats(
+        size_in_bytes=props.size,
+        create_date_time=_as_utc(props.creation_time or props.last_modified),
+    )
--- a/ingestion/src/metadata/profiler/constants.py
+++ b/ingestion/src/metadata/profiler/constants.py
@@ -1,0 +1,23 @@
+#  Copyright 2025 Collate
+#  Licensed under the Collate Community License, Version 1.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/LICENSE
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""
+Keys of the table-level results a profiler interface returns.
+
+They are shared between the producers (the SQA table metric computers and the pandas interface) and
+the consumer (`Profiler.get_profile`), which maps them onto `TableProfile`.
+"""
+
+ROW_COUNT = "rowCount"
+COLUMN_COUNT = "columnCount"
+COLUMN_NAMES = "columnNames"
+SIZE_IN_BYTES = "sizeInBytes"
+CREATE_DATETIME = "createDateTime"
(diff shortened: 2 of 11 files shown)


ingestion/src/metadata/utils/datalake/object_stats.py
```diff
@@ -0,0 +1,91 @@
...
+)
+from metadata.generated.schema.entity.services.connections.database.datalake.s3Config import (
+    S3Config,
+)
+
+
+class ObjectStats(BaseModel):
+    """Storage-level stats for a single object backing a Datalake table."""
+
+    size_in_bytes: Optional[int] = None  # noqa: UP045
```
IceS2: Can we avoid adding new `# noqa: UP` and use modern syntax instead?

ingestion/src/metadata/utils/datalake/object_stats.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    if not value:
+        return None
+    if value.tzinfo is None:
+        return value.replace(tzinfo=timezone.utc)
+    return value.astimezone(timezone.utc)
+
+
+@singledispatch
+def get_object_stats(config_source: Any, client: Any, bucket_name: str, key: str) -> ObjectStats:
```
IceS2: Could we also avoid using Any if possible? I imagine that in this case we could do with a Union.

I guess it would make sense to unify a bit the logic of everything datalake/storage in the mid future

ingestion/src/metadata/utils/datalake/object_stats.py
```diff
@@ -0,0 +1,91 @@
...
+    """Normalize a provider timestamp to timezone-aware UTC.
+
+    `Profiler.get_profile` relabels the value with `replace(tzinfo=utc)` rather than converting it,
+    so a non-UTC timestamp leaving this module would end up silently shifted on the profile.
+    """
+    if not value:
+        return None
+    if value.tzinfo is None:
+        return value.replace(tzinfo=timezone.utc)
+    return value.astimezone(timezone.utc)
```
IceS2: If we are only using it here on create_date, I'd make it a method of the class and probably use it automatically through pydantic validators if we are using Pydantic.

Alternatively, I'd use dataclass that is lighter and instead of building the Object directly in the functions below, I'd do it through a classmethod that would automatically apply this.


This way it should help us guarantee they will be consistent

IceS2 (approved):

PR #29635, opened 2026-06-30
Fixes #18042: Add custom service style support

harshach:
### Describe your changes:

Fixes #18042

Adds optional service `style` metadata for custom service types, persists it through create/update flows, denormalizes it into child search docs with reindex/update propagation, and renders editable icons/colors in service list, detail, and explore views.

#
### Type of change:
- [x] Improvement
- [x] New feature

#
### High-level design:
Service schemas now expose optional `style`; mappers persist it, search indexing prefetches/caches service styles, service updates cascade style changes to children, and UI edit controls are gated to `Custom*` service types.
No DB/Flyway migration is needed because existing service JSON remains valid without the optional field.

#
#### Use cases covered
- Custom service icon URL is preferred on service records and child search docs.
- Service style is denormalized during search doc build.
- Backend/UI formatting and compile checks were run.

#### Unit tests
- [x] I added unit tests for the new/changed logic.
- Files added/updated: `BuildSearchIndexDocTest.java`, `ServiceUtilsClassBase.test.ts`

#### Backend integration tests
- [ ] I added integration tests in `openmetadata-integration-tests/` for new/changed API endpoints.
- Files added/updated: none

#### Ingestion integration tests
- [x] Not applicable (no ingestion changes).

#### Playwright (UI) tests
- [ ] I added Playwright E2E tests under `openmetadata-ui/.../ui/playwright/` for UI changes.
- Files added/updated: none

#### Manual testing performed
- `PATH="$PWD/.context/bin:$PATH" make generate`
- `mvn spotless:apply`
- UI checkstyle sequence on changed UI files
- `mvn -pl common,openmetadata-spec,openmetadata-shaded-deps/elasticsearch-dep,openmetadata-shaded-deps/opensearch-dep -DskipTests install`
- `mvn -pl openmetadata-service -DskipTests compile`
- `mvn -pl openmetadata-service -Dtest=BuildSearchIndexDocTest test`
- `yarn test src/utils/ServiceUtilsClassBase.test.ts --runInBand`
- `git diff --check`
- `yarn tsc:check` was run; it still reports existing project-wide baseline errors unrelated to this change.

#
### UI screen recording / screenshots:
Not attached from this workspace.

#
### Checklist:
- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [x] My PR title is `Fixes <issue-number>: <short explanation>`
- [x] My PR is linked to a GitHub issue via `Fixes #<issue-number>` above.
[...]

PR #31930, opened 2026-08-24
fix(playwright): poll the asset card count in the Domains asset-count test

anuj-kumary:
## Problem

`playwright/e2e/Pages/Domains.spec.ts` → **Verify domain and subdomain asset count accuracy** fails on `1.13` while passing on `main` and `2.0`. All three attempts failed in the 1.13 nightly, but not on the same step:

| Attempt | Failing step | Assertion |
|---|---|---|
| retry 0 | subdomain step | `expect(assetCards).toBe(3)` → got `0` |
| retry 1 | domain step | `expect(assetCards).toBe(6)` → got `0` |
| retry 2 | domain step | `expect(assetCards).toBe(6)` → got `0` |

Retry 0 found all 6 cards on the domain step before failing on the subdomain step. Same locator, same code path, different result per attempt — this is a race, not a stale selector and not a product bug.

## Root cause

Both verify steps read the card count as a snapshot:

```ts
const assetCards = await page.locator('[data-testid*="table-data-card_"]').count();
expect(assetCards).toBe(totalCount);
```

`Locator.count()` does not auto-wait, and `expect(number).toBe(...)` does not poll. The count is frozen at whatever the DOM held at that instant, with no retry.

The only synchronization before it is `waitForAllLoadersToDisappear`, which is:

```ts
await expect(page.locator('[data-testid="loader"]')).toHaveCount(0, { timeout });
```

That waits for a loader to **vanish** — it never waits for one to **appear**. `assetsTab.click()` resolves as soon as the click is dispatched, and until React commits the `AssetsTabs` mount (which is when `isLoading === true` first renders `data-testid="loader"`) the DOM holds zero loaders. The assertion is satisfied immediately, so the helper provides no synchronization at all and `.count()` runs against a list that has not painted yet.

### Why the neighbouring assertion always passes

`expect(displayedCount).toBe(...)` passed on every attempt, because the two assertions read from independent sources:

- **the tab badge** — `DomainDetails.component.tsx`'s own `fetchAssetCount()`, a `searchQuery({ pageSize: 0 })` fired on domain load, warm long before the Assets tab is clicked;
- **the card list** — a separate search that `AssetsTabs.component.tsx` issues when it mounts on tab activation.

So a correct badge says nothing about whether the list has rendered. That split is exactly what lets `displayedCount === 6` coexist with `assetCards === 0`.

## Fix

Hold a `Locator` instead of a number and assert with `toHaveCount`, which re-queries until the expect timeout so the in-flight `AssetsTabs` search can resolve:

```diff
- const assetCards = await page
-   .locator('[data-testid*="table-data-card_"]')
-   .count();
+ const assetCards = page.locator('[data-testid*="table-data-card_"]');

  expect(displayedCount).toBe(totalCount);
- expect(assetCards).toBe(totalCount);
+ await expect(assetCards).toHaveCount(totalCount);
```

Applied to both verify steps.

## This is a missing backport

`main` and `2.0` already use this exact form. They got it via `26e0940362` (#29635, 2026-07-11) — I diffed that commit against its parent to confirm it is what flipped both call sites. `git merge-base --is-ancestor 26e0940362 origin/1.13` reports **not an ancestor**, which is why only 1.13 fails.

Both verify steps in this PR are now identical to `origin/main`, verified with a direct `diff` of the two regions.

## Verification

- `prettier --check` clean on the file (prettier 2.8.8, as pinned).
- Both changed steps `diff` clean against `origin/main`.
- Not run locally: the spec needs a live server plus Elasticsearch, so the fix is verified by source-level reasoning against this branch's `AssetsTabs` / `DomainDetails` components rather than by a green local run. The 1.13 nightly is the real confirmation.

`displayedCount` is intentionally left as a snapshot read — that is what `main` does, and it is not what failed here.

This PR stabilizes the Domains asset-count Playwright test by replacing immediate card-count snapshots with Playwright locator assertions that poll until the expected domain and subdomain card counts render.
- Keeps each asset-card query as a live locator.
- Uses asynchronous `toHaveCount` assertions at both affected verification steps.
</details>

The PR appears safe to merge, with the targeted polling assertions correctly addressing the asset-card rendering race.

The change preserves the existing selectors and expected counts while replacing non-waiting snapshots with Playwright's retrying locator assertions at both affected call sites; no blocking or independently actionable issue remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Domains.spec.ts | Both asset-card count checks now poll the DOM through Playwright's locator assertion, addressing the documented rendering race without changing product behavior. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(playwright): poll the asset card cou..."](https://github.com/open-metadata/openmetadata/commit/a15cb7ebb69b05c614e194b905b4db91c7e98b38) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56144556)</sub>

### DIFF
--- a/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Domains.spec.ts
+++ b/openmetadata-ui/src/main/resources/ui/playwright/e2e/Pages/Domains.spec.ts
@@ -869,12 +869,10 @@ test.describe('Domains', () => {
         const countText = await assetCountElement.textContent();
         const displayedCount = Number.parseInt(countText ?? '0', 10);
         const totalCount = domainAssets.length + subDomainAssets.length;
-        const assetCards = await page
-          .locator('[data-testid*="table-data-card_"]')
-          .count();
+        const assetCards = page.locator('[data-testid*="table-data-card_"]');
 
         expect(displayedCount).toBe(totalCount);
-        expect(assetCards).toBe(totalCount);
+        await expect(assetCards).toHaveCount(totalCount);
       });
 
       await test.step('Verify subdomain asset count matches displayed cards', async () => {
@@ -909,12 +907,10 @@ test.describe('Domains', () => {
         const countText = await assetCountElement.textContent();
         const displayedCount = Number.parseInt(countText ?? '0', 10);
 
-        const assetCards = await page
-          .locator('[data-testid*="table-data-card_"]')
-          .count();
+        const assetCards = page.locator('[data-testid*="table-data-card_"]');
 
         expect(displayedCount).toBe(subDomainAssets.length);
-        expect(assetCards).toBe(subDomainAssets.length);
+        await expect(assetCards).toHaveCount(subDomainAssets.length);
       });
     } finally {
       await subDomain?.delete(apiContext);


anuj-kumary:
Superseded by #31931, which raises the same commit from a branch on `open-metadata/OpenMetadata` instead of a fork.

PR #31937, opened 2026-08-24
fix(data-insights): declare the data asset types once and make the field catalog deterministic

### LINKED ISSUE #31963: Data Insights /charts/fields omits metric and changes its contents on every restart
## Problem

Two independent defects in `GET /v1/analytics/dataInsights/custom/charts/fields`.

### 1. `metric` is ingested daily but never advertised

The set of entity types Data Insights covers is hand-written in three places, and nothing forces them to agree:

| | Site | Role | Count |
|---|---|---|---|
| A | `DataInsightSystemChartRepository.dataAssetTypes` | which `di-data-assets-*` indices the field catalog reads | 17 |
| B | `DataInsightsApp.dataAssetTypes` | which datastreams get created, deleted and written | 16 |
| C | `dataInsights/config.json` `mappingFields` keys | per-type attribute lists | 16 |

`B == C` is enforced only by accident: a type in B missing from C hits an unguarded `.get(entityType)` in `getEntityAttributeFields` and throws. Nothing at all forces A to agree with either.

`metric` was added to B and C in #26260, with the stated intent of making metrics chartable. It was never added to A. So `di-data-assets-metric` is created and populated every day, `fetchDIChartFields` never enumerates it, and `metricType`, `unitOfMeasurement`, `granularity` and `relatedMetrics.*` never reach the chart builder.

The two data-quality types are the mirror image and are *correct*: since #29963 they reach the wildcard through a `dataInsightAliases` entry in `indexMapping.json` pointing at the live index, so they belong in A and must stay out of B. The real invariant is `A = B ∪ aliased`.

### 2. The catalog changes on every restart

`getFieldNames` deduplicates field names **globally** across entity types, so whichever type is iterated first claims every shared name permanently. The iteration source is a `Set.of`, whose order is salted per JVM start.

## Impact

Measured on a seeded instance: **277 records covering only 14 of 17 types**.

Types share nearly all their field names, so a type whose names were all already claimed contributes *nothing*. `topic` shares 28 of 28 field names with `table` and advertised zero. Reconstructing the JVM's actual iteration order and brute-forcing all 68 possibilities: 9 outcomes are reachable, `database`, `databaseSchema`, `topic` and `storedProcedure` appear or vanish per restart, and 17 is never reachable.

So a user picking "Topic" in the chart wizard's field picker gets an empty list on roughly a coin flip, and the same server restarted offers a different set of chartable fields.

The endpoint feeds both the Collate chart wizard and the AI chart assistant, so both are affected. The assistant's prompt is built by grouping this response by `entityType`, which means an entire asset type can silently be described to the model as having no fields.

## Cause

Bug 1 is the three hand-written lists above.

Bug 2 is the dedup key, duplicated verbatim in both engines:

```java
// openmetadata-service/.../search/opensearch/OpenSearchDataInsightAggregatorManager.java
// openmetadata-service/.../search/elasticsearch/ElasticSearchDataInsightAggregatorManager.java
if (fieldList.stream().noneMatch(e -> e.get("name").equals(finalFieldName))) {   // <-- global, not per type
```

## Expected

- The list of covered entity types is declared once, and the chart catalog cannot drift from what is actually ingested.
- The catalog is identical across restarts, and every covered type advertises the fields it really has.
- `metric` is chartable, which was the intent of #26260.

## Measured effect of the fix

On a local 2.0.0-SNAPSHOT instance the catalog goes from **277 records / 14 types to 1,082 records / 18 types**, and is byte-identical across three JVM restarts.

The growth is union-versus-sum rather than new information: `id` exists on every type and was previously listed once, now once per type, with correct attribution. The number of **distinct** field names goes from 277 to 287, and those 10 new names are exactly the metric-only fields that bug 1 was hiding:

```
granularity, metricType, unitOfMeasurement,
relatedMetrics.{id, name, displayName, description, fullyQualifiedName, type, deleted}
```

Per-type coverage after the fix, on the same instance:

```
testCaseResolutionStatus 106   table 89   mlmodel 71   container 68   searchIndex 63
storedProcedure 63   glossaryTerm 60   dashboardDataModel 60   pipeline 59
databaseSchema 58   testCaseResult 55   chart 53   dashboard 53   topic 52
metric 51   database 51   dataProduct 40   tag 30
```

Endpoint latency at the larger size is 30-60 ms over three runs, so the per-type dedup is not a cost concern.

## Note for consumers

Changing the record identity from `name` to `(entityType, name)` changes the shape of the response: a consumer that treats it as a flat list of selectable fields will now show one row per `(type, field)` pair. Consumers that group by `entityType` are unaffected and are in fact the ones this fixes.

manerow:
Fixes #31963

Three Data Insights bugs that share a shape: the chart returns a number that looks fine and is wrong. Nothing throws, nothing logs.

## 1. `metric` never reaches the chart builder

The set of entity types Data Insights covers was hand-written in three places:

| | Where | Decides |
|---|---|---|
| **A** | `DataInsightSystemChartRepository.dataAssetTypes` | which indices the chart field catalog reads |
| **B** | `DataInsightsApp.dataAssetTypes` | which datastreams get created and written |
| **C** | `dataInsights/config.json` keys | the per-type attribute lists |

B and C stay in sync only because a mismatch throws. Nothing keeps A honest.

#26260 added `metric` to B and C in March, to make metrics chartable. It was never added to A. So `di-data-assets-metric` has been created and filled daily ever since, while `metricType`, `unitOfMeasurement` and `granularity` have never appeared in the chart builder.

**Fix:** one enum, `dataAssetType.json`. A derives from it. B derives as "every type that no live index aliases in", which is the same 16 as before, so ingestion is unchanged.

B is deliberately *not* equal to A. The two data-quality types reach the wildcard through a `dataInsightAliases` entry pointing at the live index, so ingesting them would aim create/delete at live data.

## 2. The catalog changes on every server restart

`getFieldNames` deduplicated field names **globally**, so the first entity type iterated claimed a shared name permanently. The iteration source was a `Set.of`, whose order Java salts per JVM start.

Measured on a seeded instance: **277 records covering 14 of 17 types**. `topic` shares all 28 of its field names with `table`, so it advertised zero fields despite holding 47,766 documents. Across the 68 possible orders, 9 outcomes are reachable and 17 types never is.

Someone picking "Topic" in the field picker got an empty list on roughly a coin flip.

**Fix:** dedupe by `(entityType, name)` instead of `name`. The type set also becomes a `LinkedHashSet` in enum order, so record *order* is stable too, which matters for any consumer that resolves a duplicated name by taking the first record.

## 3. Charts pick categories from data they don't count

A line chart with a terms x-axis picks its 100 buckets, then counts inside them with the metric filter applied. Selection ignored the filter.

So a "tables per service" chart ranked services by their **total** document count. A BI service with 5,000 dashboards and 3 tables outranked a warehouse service with 800 tables and nothing else. The first got a slot and rendered as `3`. The second was missing from the chart entirely.

The axis selected its categories by a quantity the chart neither filters on nor displays. Measured on a seeded instance: a "Tier 1 assets per service" chart put `test-dashboard-service` (215,515 documents, **zero** Tier-1 assets) in the first slot, ahead of the one service that had any.

**Fix:** when every metric of a chart shares one filter, apply it to the request as well as inside the buckets. Within a bucket the filter is a conjunct applied twice, so the predicate is unchanged; what moves is which categories get a bucket at all.

Gated on the chart having a terms x-axis. A `@timestamp` chart has no top-N to align, and narrowing its query would shorten the window a date histogram plots, moving the first/last delta the dashboard renders. Six shipped charts have that shape.

## 4. An empty category takes the whole chart down with a 500

`avg`, `min` and `max` over a bucket the metric filter emptied return `null`, not `0` — there is no average of nothing. Both search clients hand that back as a boxed `Double`, and both aggregators unbox it straight into a primitive:

```java
double value = aggregation.value();   // null -> NullPointerException
```

Nothing catches it, so the request dies and the user gets an HTTP 500 rather than a chart. `sum` and `count` return `0.0`, which is why only those three functions are affected.

This is not new and it is not caused by this PR — reproduced on a `main` deployment with a three-line chart definition:

```
HTTP 500 - Cannot invoke "java.lang.Double.doubleValue()" because the return value of
"...SingleMetricAggregateBase.value()" is null
```

It is fixed here because §3 makes it reachable again. Ranking by the measure deliberately keeps the categories the filter matched nothing in — exactly the buckets that return `null`. Hoisting had been hiding the bug by deleting them first.

**Fix:** null-check before unboxing, on both engines, alongside the `NaN`/`Infinity` checks already there. A category with no value is skipped rather than emitted as `0`, since `0` is a legitimate average and would be a wrong answer.

OpenSearch routes every metric type through one method, so guarding it covers all of them. Elasticsearch split the same job across three overloads and guarding only one left an asymmetry, so the redundant `ValueCountAggregate` overload is gone — it extends `SingleMetricAggregateBase`, so `value_count` now binds to the guarded path like everything else. No behaviour change: `value_count` never returns null.

Worth noting for reviewers: this is the only change here that fixes a live 500 on `main`, and it has no dependency on the rest of the PR.

## Also fixed in the same files

- `getEntityAttributeFields` returned the shared `common` list and appended to it **in place**. The workflow parses the config once and loops over every type, so each type inherited the attributes of every type before it, and which ones followed the salted order above.
- The same method now fails with a readable message when a type has no `mappingFields` entry, instead of an NPE.
- `mappingFields` keys are typed by the enum, so a misspelling fails at load rather than silently producing documents without their type-specific attributes.

## What changes for consumers

**The catalog grows from ~277 records to ~1,082.** This is union versus sum: `id` exists on every type and was listed once, now once per type. No new information, correct attribution.

Consumers must group by `entityType`. Collate's drag-and-drop asset panel already does, and this fixes it there (`topic` goes from 0 fields to 52). Its three flat-list pickers do not, and are handled in collate#6077.

**Filtered category charts stop drawing zero bars**, because terms uses `min_doc_count: 1`. The clearest case is a service Insights tab: `healthy_data_assets` is requested with a service filter and today returns ~100 rows of which 99 are zero, so the widget reports another service's number over a 99 "day" window. It now returns the one service.

**Counts can move up.** Terms samples per shard, and a category outside a shard's cutoff loses that shard's contribution. Re-ranking over the filtered population promotes exactly those marginal categories, so one can come back with a larger, more complete number. Counts move toward completeness; they are not asserted unchanged.

This is the pre-existing terms approximation getting *smaller*, not a new source of error. Elasticsearch's condition for an accurate `doc_count` and accurate sub-aggregations is that the population a bucket is ranked on equals the population it counts. Before this change those two disagree by construction -- ranked on unfiltered documents, counted on filtered ones -- which is the regime in which shard truncation does the most damage. Hoisting makes them the same population again.

## What this does not fix

- `size(100)` is unchanged. This changes *which* hundred categories you get, not how many.
- Selection aligns to the filter but not to the measure: a `sum(k='size')` chart still ranks by document count.
- The `groupBy` dimension keeps the same misalignment: it builds its own `terms(...).size(100)` and is untouched whenever the x-axis is `@timestamp`.
- A formula's `q='...'` is deliberately not hoisted. Doing so would scope the denominator of every percentage chart and make them all read 100%.
- Charts whose metrics carry different filters, or where only some are filtered, are skipped.

## Tests

Each verified to fail against the bug it guards:

| Revert | Fails with |
|---|---|
| `metric` removed from config.json | coverage test names `metric` |
| dedup back to global | `expected: <[table, topic]> but was: <[table]>` |
| type set back to `toUnmodifiableSet()` | three runs, three different orders, all caught |
| `common` copy removed | `expected: <[id, service]> but was: <[id, columns, service]>` |
| hoist removed | 2 failures per engine |
| gate widened to `groupBy` | `onlyATermsAxisIsHoistable` |
| null guard removed | the exact `Double.doubleValue()` NPE, per engine |

The 12 pre-existing aggregator tests pass **unmodified**. None of their fixtures sets a metric filter, which is what proves an unfiltered chart still produces a byte-identical request.

The PR centralizes the Data Insights asset-type contract and makes field-catalog generation deterministic. It also aligns categorical bucket selection with metric filters and safely skips null aggregate values.
- Derives catalog and ingestion type sets from the shared schema enum while excluding live-index aliases from ingestion.
- Deduplicates catalog fields by entity type and field name and prevents shared configuration-list mutation.
- Orders categorical terms buckets by filtered participation across Elasticsearch and OpenSearch.
- Adds focused unit and real-engine coverage for configuration, catalog, filtering, ordering, and null aggregates.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java | Derives the stable ingestion set from the shared enum while excluding types supplied through live-index aliases. |
| openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/search/DataInsightsSearchConfiguration.java | Replaces untyped mapping-field keys with enum-backed configuration and reports invalid keys during loading. |
| openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/search/DataInsightsSearchInterface.java | Copies common fields before appending type-specific attributes and reports missing type mappings explicitly. |
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataInsightSystemChartRepository.java | Uses enum declaration order for deterministic and complete Data Insights field-catalog enumeration. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/DataInsightMetricFilter.java | Centralizes extraction of metric-filter query JSON for both search engines. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/dataInsightAggregators/ElasticSearchLineChartAggregator.java | Orders categorical buckets through the metric filter aggregation while preserving date-histogram behavior. |
| openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/dataInsightAggregator/OpenSearchLineChartAggregator.java | Implements the OpenSearch counterpart of filter-aligned categorical bucket ordering. |
| openmetadata-spec/src/main/resources/json/schema/dataInsight/custom/dataAssetType.json | Establishes the schema-backed source of truth for Data Insights asset types. |
| openmetadata-ui/src/main/resources/ui/src/generated/dataInsight/custom/dataAssetType.ts | Keeps the generated TypeScript enum aligned with the shared schema contract. |

</details>

```mermaid
flowchart LR
  Schema[DataAssetType schema enum] --> Catalog[Chart field catalog types]
  Schema --> Ingestion[Ingested asset types]
  Mapping[Index mappings and live aliases] --> Ingestion
  Config[Per-type mapping fields] --> Streams[Data Insights documents]
  Ingestion --> Streams
  Streams --> Terms[Categorical terms axis]
  Filter[Metric filter] --> Wrapper[Filtered metric aggregation]
  Wrapper --> Terms
  Terms --> Results[Chart results]
  Results --> NullGuard[Skip null aggregate values]
```
</details>

<sub>Reviews (23): Last reviewed commit: ["fix(data-insights): rank a categorical a..."](https://github.com/open-metadata/openmetadata/commit/8107eca8630860531d69a8dbfe3357c5fc3a39c8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56158213)</sub>

- Knowledge Base — [Restore Data Insights chart queries without the broken filter path](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/reverts/revert_26472-20260314-data-insights-filter-6e212ca.md)
</details>

### DIFF
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/DataInsightsApp.java
@@ -9,6 +9,9 @@ import static org.openmetadata.service.workflows.searchIndex.ReindexingUtil.getI
 import es.co.elastic.clients.transport.rest5_client.low_level.Rest5Client;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -18,9 +21,11 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.openmetadata.schema.dataInsight.custom.DataAssetType;
 import org.openmetadata.schema.entity.app.App;
 import org.openmetadata.schema.entity.app.AppRunRecord;
 import org.openmetadata.schema.entity.app.FailureContext;
@@ -77,24 +82,30 @@ public class DataInsightsApp extends AbstractNativeApplication {
   private volatile boolean stopped = false;
   private final AtomicReference<DataInsightsWorkflow> activeWorkflow = new AtomicReference<>();
 
-  public final Set<String> dataAssetTypes =
-      Set.of(
-          "table",
-          "storedProcedure",
-          "databaseSchema",
-          "database",
-          "chart",
-          "dashboard",
-          "dashboardDataModel",
-          "pipeline",
-          "topic",
-          "container",
-          "searchIndex",
-          "mlmodel",
-          "dataProduct",
-          "glossaryTerm",
-          "tag",
-          "metric");
+  /**
+   * The entity types this app ingests: every {@link DataAssetType} that no live index aliases into
+   * the Data Insights wildcard.
+   *
+   * <p>The data-quality types are deliberately excluded. They reach {@code di-data-assets-*} through
+   * a {@code dataInsightAliases} entry in indexMapping.json that points at the live entity index, so
+   * Data Insights reads them without ever writing them. Creating or deleting a datastream for one
+   * would target that alias, and therefore live data, because {@link #getDataStreamName} would
+   * produce the very name the alias already occupies.
+   */
+  public Set<String> getDataAssetTypes() {
+    return Collections.unmodifiableSet(
+        Arrays.stream(DataAssetType.values())
+            .map(DataAssetType::value)
+            .filter(dataAssetType -> !isAliasedFromLiveIndex(dataAssetType))
+            .collect(Collectors.<String, LinkedHashSet<String>>toCollection(LinkedHashSet::new)));
+  }
+
+  private boolean isAliasedFromLiveIndex(String dataAssetType) {
+    IndexMapping indexMapping = searchRepository.getIndexMapping(dataAssetType);
+    return indexMapping != null
+        && indexMapping.getDataInsightAliases() != null
+        && !indexMapping.getDataInsightAliases().isEmpty();
+  }
 
   public DataInsightsApp(CollectionDAO collectionDAO, SearchRepository searchRepository) {
     super(collectionDAO, searchRepository);
@@ -138,7 +149,7 @@ public class DataInsightsApp extends AbstractNativeApplication {
             : "en";
 
     try {
-      for (String dataAssetType : dataAssetTypes) {
+      for (String dataAssetType : getDataAssetTypes()) {
         IndexMapping dataAssetIndex = searchRepository.getIndexMapping(dataAssetType);
         String dataStreamName =
             getDataStreamName(searchRepository.getClusterAlias(), dataAssetType);
@@ -160,7 +171,7 @@ public class DataInsightsApp extends AbstractNativeApplication {
     DataInsightsSearchInterface searchInterface = getSearchInterface();
 
     try {
-      for (String dataAssetType : dataAssetTypes) {
+      for (String dataAssetType : getDataAssetTypes()) {
         String dataStreamName =
             getDataStreamName(searchRepository.getClusterAlias(), dataAssetType);
         if (searchInterface.dataAssetDataStreamExists(dataStreamName)) {
@@ -413,7 +424,7 @@ public class DataInsightsApp extends AbstractNativeApplication {
             timestamp,
             batchSize,
             backfill,
-            dataAssetTypes,
+            getDataAssetTypes(),
             collectionDAO,
             searchRepository,
             getSearchInterface()));
--- a/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/search/DataInsightsSearchConfiguration.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/insights/search/DataInsightsSearchConfiguration.java
@@ -1,10 +1,42 @@
 package org.openmetadata.service.apps.bundles.insights.search;
 
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
 import lombok.Getter;
+import org.openmetadata.schema.dataInsight.custom.DataAssetType;
 
 @Getter
 public class DataInsightsSearchConfiguration {
-  private Map<String, List<String>> mappingFields;
+  private MappingFields mappingFields;
+
+  /**
+   * The per-type attribute lists from {@code dataInsights/config.json}: {@code common} applies to
+   * every document, and each remaining key adds the attributes specific to one entity type.
+   *
+   * <p>Those keys are typed as {@link DataAssetType} rather than kept as strings, so a key naming
+   * something Data Insights does not cover cannot exist. Jackson resolves them through the enum's
+   * {@code @JsonCreator}, which lets the file keep its flat shape while a misspelling fails at load
+   * instead of silently producing documents without their type-specific attributes.
+   */
+  @Getter
+  public static class MappingFields {
+    private List<String> common = List.of();
+    private final Map<DataAssetType, List<String>> byType = new EnumMap<>(DataAssetType.class);
+
+    @JsonAnySetter
+    void putByType(String entityType, List<String> attributeFields) {
+      try {
+        byType.put(DataAssetType.fromValue(entityType), attributeFields);
+      } catch (IllegalArgumentException e) {
+        // Chaining `e` would hide this message: Jackson's any-setter path reports the root cause,
+        // so the operator would see fromValue's bare "<key>" instead.
+        throw new IllegalArgumentException(
+            String.format(
+                "%s declares mappingFields for '%s', which is not a Data Insights asset type",
+                DataInsightsSearchInterface.DATA_INSIGHTS_SEARCH_CONFIG_PATH, entityType));
+      }
+    }
+  }
 }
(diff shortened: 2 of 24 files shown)


manerow:
Superseded by a fresh PR from the same branch: the history here contained a request-narrowing approach that was later replaced by axis ranking, and the timeline no longer matches the code. Reopening clean.

manerow:
Continued in #32115 (same branch, rebuilt history).

PR #31950, opened 2026-08-24
fix(ingestion): pin asammdf to the 8.2-8.7 window so arm64 builds get a wheel

Khairajani:
## What

`asammdf~=8.1.0` -> `asammdf>=8.2,<8.8`, one line in `ingestion/setup.py`.

## Why

asammdf 8.0.x and 8.1.x publish **no linux-aarch64 wheel**, so on arm64 pip falls back to the
8.1.2 sdist, which builds with scikit-build-core + CMake and does not vendor `ext/libdeflate`:

```
CMake Error at CMakeLists.txt:20 (add_subdirectory):
  add_subdirectory given source "ext/libdeflate" which is not an existing directory.
...
ERROR: Failed building wheel for asammdf
```

Any arm64 image build that installs `./ingestion[all]` from source fails at that step.
7.4.5 compiled from source fine because its sdist was plain `setuptools.build_meta`, which is why
this only surfaced once the pin moved to `~=8.1.0`.

## Why this exact range

| range | aarch64 wheel | numpy | chardet |
|---|---|---|---|
| 8.0.x–8.1.x | **none** | 8.0 `<2` / 8.1 `>=1.23` | optional `decode` extra |
| **8.2.0–8.7.2** | yes | `>=2.0.0` | optional `decode` extra |
| 8.8.0–8.8.24 | yes | `>=2.0.0` | **hard `>=7.0.0`** |

- Lower bound **8.2.0** — first release shipping `manylinux_2_17_aarch64` abi3 wheels.
- Upper bound **<8.8** — 8.8.0 promoted chardet from an optional extra to a hard `chardet>=7.0.0`,
  which conflicts with `chardet==4.0.0` in `base_requirements` (used by
  `profiler/orm/types/custom_hex_byte_string.py`). Raising that pin is a much wider change than an
  upper bound here, so the bound is documented inline for whoever revisits it.
- **Not a revert to 7.4.5** — numpy is now `>=2,<3` and 7.4.5 is numpy-1 only.

## Validation

Built `python:3.12-trixie` on `linux/arm64` and ran the real `pip install "./ingestion[all]"`:

| pin | result |
|---|---|
| `~=8.1.0` | CMake / `ext/libdeflate` failure building the sdist |
| `~=8.8` | `ResolutionImpossible` — `asammdf 8.8.x depends on chardet>=7.0.0` |
| `>=8.2,<8.8` | resolves to the 8.7.2 aarch64 wheel, no build step |

```
Successfully installed ... asammdf-8.7.2 ... numpy-2.5.2 pandas-2.3.3 ...
asammdf 8.7.2 | numpy 2.5.2 | pandas 2.3.3 | chardet 4.0.0
header._common_properties = {}
mf4 API surface OK
```

The call surface `readers/dataframe/mf4.py` uses — `MDF(version="4.10")`, `mdf.save()`,
`MDF(path, load_measured_data=False)`, `mdf.header._common_properties` — was exercised against
8.7.2. The 8 `test_mf4_reader` tests already pass against 8.1.0 on x86_64, so the 7.x -> 8.x
API move is covered; this only narrows which 8.x is selected.

The PR updates the asammdf dependency window to select releases providing Linux arm64 wheels while avoiding the chardet requirement introduced in 8.8.
- Changes the asammdf constraint from `~=8.1.0` to `>=8.2,<8.8`.
- Removes the obsolete `load_measured_data` argument from MF4 constructors across local and cloud-storage reader paths.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/setup.py | Pins asammdf to the documented 8.2–8.7 compatibility window for arm64 wheel availability and the existing chardet constraint. |
| ingestion/src/metadata/readers/dataframe/mf4.py | Removes the obsolete constructor argument consistently from local, S3, GCS, and Azure MF4 loading paths. |

</details>

<sub>Reviews (4): Last reviewed commit: ["fix(ingestion): drop the dead load\_measu..."](https://github.com/open-metadata/openmetadata/commit/dd48d72f3679d2a298150df41381abd34003fa08) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56208663)</sub>


IceS2 (dismissed):

Khairajani:
The static-checks lane was failing on the pin bump, now fixed in `dd48d72`.

asammdf's `MDF` takes `**kwargs: Unpack[MdfKwargs]`, and `load_measured_data` is not one of its keys, so basedpyright reported four `reportCallIssue` errors once the pin resolved to 8.7.2:

```
mf4.py:85:33 - error: No parameter named "load_measured_data" (reportCallIssue)
```

That kwarg has been inert for years — it dates from the 2.x/3.x memory options and is silently dropped by `**kwargs`. Verified against 8.7.2: `MDF` opens and returns the same `header._common_properties` with and without it, so removing it is behaviour preserving.

IceS2 (approved):

Issue #23493, opened 2025-09-21
Add Vault Secrets Managers


spomytkin:
**Is your feature request related to a problem? Please describe.**
In corporate environments, policies often mandate secret storage and rotation using specific vault products.
In my case, [HashiCorp Vault](https://developer.hashicorp.com/vault/docs/about-vault/what-is-vault) and [CyberArk](https://www.cyberark.com/resources/product-datasheets/cyberark-secrets-manager) support are required to use OpenMetadata in production.

**Describe the solution you'd like**
1. Implementation of HashiCorp Vault integration per OM documentation [Home / Deployment / Secrets Manager / How To Add A New Implementation](https://docs.open-metadata.org/latest/deployment/secrets-manager/how-to-add-a-new-implementation), using as reference the commits from [pull #22516](https://github.com/open-metadata/OpenMetadata/pull/22516/commits), which recently implemented Kubernetes secret support for Secrets Manager ([Issue #22511](https://github.com/open-metadata/OpenMetadata/issues/22511)).
2. Implementation of CyberArk Vault integration

**Describe alternatives you've considered**
I searched for similar efforts in progress, found “Vault support in Secrets Manager” ([Issue #19242](https://github.com/open-metadata/OpenMetadata/issues/19242)), but based on its history and [Slack conversation](https://openmetadata.slack.com/archives/C02DZK3QH8Q/p1758137854019939?thread_ts=1758137164.231949&cid=C02DZK3QH8Q), we concluded it would be appropriate to create a new feature request
**Additional context**
I started with HashiCorp Vault because it well-documented and supports “Vault dev server” - executing vault in development mode (requires no configuration, runs in-memory, on http, etc) designed for integration testing: https://developer.hashicorp.com/vault/tutorials/get-started/setup 
I’m still researching how to setup CyberArk integration environment, but likely will be way more challenging task.

elenberg:
@harshach noticed that this got removed from the March - 2026 release. Is this no longer going to be supported? I'm actively looking forward to this integration.

PR #31984, opened 2026-08-24
feat(secrets): add OpenBao / HashiCorp Vault Secrets Manager provider

### LINKED ISSUE #19242: Vault support as Secrets Manager
We would like use hashicorp vault as secret manager, to align with our security policy

Add support for vault as you have here https://docs.open-metadata.org/latest/deployment/secrets-manager/supported-implementations

We have no alternatives, the only alternative not to use OMD.

**Additional context**

oluies:
Adds OpenBao (and, by API compatibility, HashiCorp Vault) as an OpenMetadata Secrets Manager,
backed by the KV v2 engine.

Closes #19242. Related: #23493.

## Why

On the default `db` provider, service connection credentials are Fernet-encrypted into the metadata
database. That makes the Fernet key an undeclared hard dependency, and leaves the database holding
sensitive material with no access audit, no per-secret ACL and no rotation story. We run OpenBao, so
this delegates to it.

**One claim worth stating precisely, because the obvious version of it is wrong.**
`SecretsManager.encryptPasswordFields` applies `Fernet.isTokenized(v) ? v : fernet.encrypt(v)` to
whatever `storeValue` returns, and `ExternalSecretsManager.storeValue` returns `"secret:<path>"` —
which is not a Fernet token. So the *reference itself* is Fernet-wrapped in the database, for every
external provider, not just this one. **Switching to an external provider does not remove the Fernet
key dependency.** What it does change: credential plaintext leaves the database entirely, and because
secret paths are deterministic, a Fernet key loss becomes a scripted re-link rather than re-typing
every password by hand.

## Two decisions I would like maintainer input on

**Naming.** I went with `openbao` / `managed-openbao` rather than `vault` / `managed-vault`. OpenBao
is what I actually run and test against; Vault support is documented rather than named. Both open
issues say "Vault", so you may prefer the other name — and since the enum is operator-facing
configuration, this is the one thing that is expensive to change after merge. Happy to rename. Note
`vaultName` is already taken by Azure Key Vault.

**Wiring.** `managed-openbao` means the server reads and writes, following `MANAGED_AZURE_KV`.
`openbao` joins the existing fall-through arm with `aws` / `aws-ssm` / `azure-kv`, where only the
ingestion side reads. Only the managed mode moves plaintext out of the database, so it is the one
that addresses the motivation above; the non-managed value exists for symmetry and is documented as
such rather than presented as a supported mode.

## What is here

- **Schema** — two enum values plus `openBaoCredentials.json`. Generated Java/Pydantic/TS regenerated,
  never hand-edited. The TS churn is in its own commit (`67c50c1`) so it can be skipped in review.
- **Java** — `OpenBaoClient` and `OpenBaoSecretsManager`, hand-rolled on the `jersey-client` and
  Jackson already on the module's classpath. **No new runtime dependency.**
- **Python** — `openbao_secrets_manager.py` on `requests` (not `hvac`), plus factory dispatch.
- **Config / Docker** — 12 `bao*` parameters, plumbed through both dev compose files, plus a
  standalone `docker/docker-compose-openbao/` for the published-image stack.

### Two implementation details that are load-bearing

**Not-found is classified on the response body, never the status code.** OpenBao answers `404` both
for a missing secret and for a missing mount. A `status == 404 → absent` rule would route every write
on a typo'd mount to the create branch, succeed, and discard credentials with no error at write time.
Only a `404` with an empty `errors` array is treated as absent. `isNotFoundException` therefore always
returns `false` and says why — not-found is a value here, not an exception.

Worth knowing: under a least-privilege token a typo'd mount actually returns **403**, not the
404-with-errors shape, because the ACL check runs ahead of mount routing. I only found that by
re-probing with a scoped token rather than root. Both branches throw, so the safety property holds
either way, but `verifyMount()`'s message names both possible causes.

**Deletion destroys every version** (`DELETE {mount}/metadata/{path}`), not just the latest. Soft
delete leaves rotated plaintext recoverable, which would defeat the reason an operator chose an
external store. This is irreversible and is documented.

## Verification

Everything below was run, not assumed.

- `mvn -pl openmetadata-service -am install` — **BUILD SUCCESS** (rebased on current `main`)
- **200/200** Java secrets tests, including 40 new ones; zero regressions across the five existing
  providers
- **25/25** Python tests, including the pre-existing `test_all_providers_has_implementation`
- Integration test runs against **real OpenBao 2.6.2** in a Testcontainer, not a mock
- `spotless:check` and `ruff check` clean; pre-commit gate passes

**End-to-end, on a real stack:**

- *Migration* — created a service under `db` (field decrypted to the password), switched to
  `managed-openbao`, ran `migrate-secrets`; the field then decrypted to
  `secret:/openmetadata/database/bao-demo/authtype/password`, and that path in OpenBao returned the
  password.
- *Ingestion* — a real `metadata ingest` run from a separate process holding only `OPENBAO_*` env
  vars, with `serviceConnection` deliberately omitted: `Getting secret value for
  /openmetadata/database/bao-demo/authtype/password` → `Test connection for 'Mysql': Successful` →
  199 records, 0 errors.
- *Control* — the identical run with an invalid OpenBao token failed with
  `Access denied for user 'openmetadata_user'`, confirming the success was genuinely OpenBao-sourced
  and not a fallback.

I also checked the obvious verification and discarded it: `json LIKE '%password%'` returns zero rows
under `db` too, because `db` Fernet-encrypts. It passes for a completely broken provider, so the
assertion here is on the *shape* of the stored value instead.

## Known follow-ups

- **Not fixed here, but worth a look:** `SecretsManager.getSecretFields` builds its error message
  from `toDecryptObject.toString()`, and because it mutates fields in place as it walks them, a late
  resolution failure can put *earlier decrypted passwords* into a log line and an API response. This
  is a pre-existing shared path — AWS's `getSecret` throws the same way — so I left it alone rather
  than widening this PR, but it is a real leak.
- The Helm chart lives in `openmetadata-helm-charts`; its stale provider-list comment (it omits
  `kubernetes` today) and the `OM_SM_BAO_*` plumbing need a separate PR there.
- `tags` has a KV v2 equivalent (`custom_metadata`); mapping it is deferred, not overlooked.
- Reverting from OpenBao back to `db` is out of scope. Migration between two external managers is
  unsupported upstream.

The PR adds OpenBao and HashiCorp Vault KV v2 secrets-manager support across the Java service, Python ingestion framework, schemas, and Docker configuration. The attempted response to the previous root-token finding remains incomplete because the documented stack still defaults to a repository-known root credential.
- Adds token and AppRole authentication, KV v2 reads/writes/deletes, namespace handling, TLS controls, and timeout configuration.
- Wires managed and ingestion-side providers through server and Python factories.
- Adds generated schemas, unit and container tests, and OpenBao Compose overlays.
</details>

The PR is not safe to merge until the documented OpenBao stack stops enabling a repository-known root credential on its published endpoint.

Running the documented command without OM_SM_BAO_TOKEN still configures openbao-dev-root as the live root credential and exposes OpenBao through host port 8200, allowing a reachable peer to access all stored secrets.

**Files Needing Attention:** docker/docker-compose-openbao/docker-compose.openbao.yml; docker/development/docker-compose.openbao.yml
</details>

The documented OpenBao quickstart still assigns a repository-known root token by default while publishing port 8200 on every host interface. A reachable peer can use that credential to bypass the least-privilege AppRole policy and access all secrets.
</details>

| Filename | Overview |
|----------|----------|
| docker/docker-compose-openbao/docker-compose.openbao.yml | Adds the OpenBao quickstart overlay, but its environment override retains a known root-token fallback on a published endpoint. |
| docker/development/docker-compose.openbao.yml | Adds a development OpenBao overlay that uses the same fixed root credential and exposes the service port. |
| ingestion/src/metadata/utils/secrets/openbao_secrets_manager.py | Implements OpenBao KV v2 secret retrieval with token/AppRole authentication, namespaces, TLS options, and token renewal. |
| openmetadata-service/src/main/java/org/openmetadata/service/secrets/OpenBaoClient.java | Implements the server-side OpenBao HTTP client and KV v2 operations. |
| openmetadata-service/src/main/java/org/openmetadata/service/secrets/OpenBaoSecretsManager.java | Integrates OpenBao storage with the server secrets-manager abstraction. |
| openmetadata-spec/src/main/resources/json/schema/security/credentials/openBaoCredentials.json | Defines the shared OpenBao credentials and connection configuration contract. |

</details>

```mermaid
flowchart LR
  User[Documented Compose command] --> Default[OM_SM_BAO_TOKEN unset]
  Default --> Root[Known root token configured]
  Root --> Bao[OpenBao dev server]
  Bao --> Port[Port 8200 published]
  Peer[Reachable network peer] --> Port
  Port --> Access[Root access to all secrets]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["Merge branch &#39;main&#39; into feat/openbao-se..."](https://github.com/open-metadata/openmetadata/commit/1e5afea1b742cae715273fe5547696706eb99637) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56339348)</sub>

**Context used:**


openmetadata-service/src/main/java/org/openmetadata/service/secrets/OpenBaoSecretsManager.java
```diff
@@ -0,0 +1,294 @@
...
+              + "Review your configuration.");
+    }
+    validatePathComponents(secretsConfig);
+
+    this.client =
+        new OpenBaoClient(
+            new OpenBaoClient.OpenBaoConfig(
+                address,
+                parameter(secretsConfig, MOUNT, DEFAULT_MOUNT),
+                parameter(secretsConfig, NAMESPACE, ""),
```
harshach, line 84: shoudln't we allow namespace ot be configurable here?

oluies: It is configurable. That line reads the `baoNamespace` parameter and `""` is just the default when it is unset.

- `OpenBaoSecretsManager.NAMESPACE = "baoNamespace"`
- `conf/openmetadata.yaml`: `baoNamespace: ${OM_SM_BAO_NAMESPACE:-""}`, plumbed through both dev compose files
- sent as `X-Vault-Namespace`, and the header is omitted entirely when blank, since an empty namespace header is rejected by OSS servers

Covered by `namespaceHeaderIsSentWhenConfigured` and `namespaceHeaderIsOmittedWhenBlank` in `OpenBaoClientTest`.

Happy to pull it out of the generic `parameters` map into a typed field if you would rather have it there.

PR #31988, opened 2026-08-24
Fixes #31987: let distributions register their own DataRetention cleanups

### LINKED ISSUE #31987: DataRetention cannot be extended by distributions built on OpenMetadata
### Problem

`DataRetention` (`openmetadata-service/.../apps/bundles/dataRetention/DataRetention.java`) hardcodes every cleanup it performs. A distribution built on OpenMetadata that adds its own tables has no way to prune them on the same schedule.

Its only options today are:

- run a **second scheduled app**, duplicating the batching, entity stats, `ACTIVE_ERROR` reporting and failure isolation that `DataRetention` already implements; or
- not prune at all.

Neither is good. The retention job is exactly the right place for this work — it already knows how to drain a table in batches, record per-entity stats, and mark a run partially failed without aborting the rest.

### Proposal

A `DataRetentionExtension` SPI, following the `MigrationProcessExtensionProvider` idiom already used for out-of-tree migrations (`ServiceLoader` + `META-INF/services`). A provider contributes named `RetentionStep`s; `DataRetention` feeds each through its existing `executeWithStatsTracking`, so extensions inherit the same batching, stats and error reporting as the built-ins.

Extensions run **after** every built-in cleanup, so a job that fails partway still leaves OpenMetadata's own tables pruned. A provider that throws is isolated and reported rather than allowed to stop the others.

Extension config rides a new `extensions` map on `DataRetentionConfiguration`. This is load-bearing rather than cosmetic: `DataRetention.init` converts with the strict `JsonUtils.OBJECT_MAPPER` (`FAIL_ON_UNKNOWN_PROPERTIES` is only disabled on `OBJECT_MAPPER_LENIENT`), so a downstream key added anywhere but a *declared* property throws at app init. Declaring it keeps the outer `additionalProperties: false`, so typos in the built-in keys are still caught.

Nothing changes for an installation with no extensions registered.

pmbrull:
### Describe your changes:

Fixes #31987

`DataRetention` hardcoded every cleanup, so a distribution built on OpenMetadata had no way to prune its own tables on the same schedule — its choices were a second scheduled app duplicating all the batching and stats machinery, or nothing.

This adds a **`DataRetentionExtension` SPI**, following the `MigrationProcessExtensionProvider` idiom already used for out-of-tree migrations:

- `DataRetentionExtension` — `name()`, `steps(config)`, and a `retentionPeriodDays(config, defaultDays)` default method so the config-key convention lives in one place.
- `RetentionStep` — `record RetentionStep(String statsKey, BatchDeleter deleter)`. `deleteBatch(int)` returns rows deleted; returning fewer than the batch size ends the drain loop.
- `DataRetentionExtensions` — `ServiceLoader` discovery plus `resolveSteps(config, onFailure)`. A provider that throws is skipped and reported; the others still run.
- `DataRetention` — **+23 lines**: discover in the constructor, `cleanExtensions(config)` as the last step of `executeCleanup`. Each step goes through the existing `executeWithStatsTracking`, so extensions inherit the same batching, entity stats and `ACTIVE_ERROR` reporting as the built-ins. Extensions run last, so a job that fails partway still leaves OpenMetadata's own tables pruned.
- `dataRetentionConfiguration.json` — new `extensions` property, `existingJavaType: java.util.Map<String, Integer>` (the idiom already at `eventPublisherJob.json:358`).

That last one is load-bearing rather than cosmetic. `DataRetention.init` converts with the strict `JsonUtils.OBJECT_MAPPER` — `FAIL_ON_UNKNOWN_PROPERTIES` is only disabled on `OBJECT_MAPPER_LENIENT` — so a downstream key added anywhere but a *declared* property would throw at app init. Declaring it keeps the outer `additionalProperties: false`, so typos in the built-in keys are still caught.

**Nothing changes for an installation with no extensions registered.**

### Verification

- `mvn -pl openmetadata-service spotless:check test -Dtest=DataRetentionExtensionsTest` — **8 tests, 0 failures**, BUILD SUCCESS.
- Discovery is tested against the **real `ServiceLoader`**, not a hand-built list: `InertTestRetentionExtension` is registered in `src/test/resources/META-INF/services/`. Verified non-tautological — removing that registration file (and its stale `target/test-classes` copy) fails exactly `discoverFindsExtensionsRegisteredViaServiceLoader` and leaves the other 7 green.
- Also exercised from a downstream distribution on a real assembled classpath: the provider was discovered by name, its steps came back in the declared order, it stayed inert when its DAO was unwired, and an operator override via `extensions.<name>` was honoured.

### Type of change:

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactor

### Checklist:

- [x] My PR title is following the [convention](https://github.com/open-metadata/OpenMetadata/blob/main/CONTRIBUTING.md).

The PR adds a ServiceLoader-based data-retention extension SPI and integrates extension cleanup into the existing batching, statistics, and failure-reporting flow. It also hardens provider discovery and refreshes scheduler thread-local state before and after jobs.
- Adds extension discovery, ordered retention steps, configurable retention periods, and bounded batch draining.
- Isolates malformed, unlinkable, and failing providers while allowing healthy providers to continue.
- Extends the shared configuration schema and generated UI models with extension retention settings.
- Adds unit and integration coverage for provider isolation, batching, and refreshed application configuration.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/dataRetention/DataRetentionExtensionRegistry.java | Adds defensive ServiceLoader discovery and per-provider step resolution; the previously reported discovery and failure-path isolation gaps are addressed. |
| openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/dataRetention/DataRetention.java | Runs extension steps after built-in retention work using the existing statistics and run-failure machinery. |
| openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/dataRetention/BatchDrain.java | Centralizes bounded batched deletion and reports deletion totals, failures, and iteration-cap exhaustion. |
| openmetadata-service/src/main/java/org/openmetadata/service/apps/scheduler/OmAppJobListener.java | Clears per-request thread-local state around Quartz jobs so cached application configuration is not reused by long-lived workers. |
| openmetadata-spec/src/main/resources/json/schema/entity/applications/configuration/internal/dataRetentionConfiguration.json | Declares extension-specific retention periods in the shared configuration contract. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DataRetentionAppIT.java | Adds coverage that subsequent scheduled runs observe updated application configuration. |

</details>

```mermaid
flowchart TD
    Q[Quartz starts DataRetention] --> C[Clear per-request context]
    C --> B[Run built-in cleanup steps]
    B --> D[Discover registered providers]
    D --> R[Resolve ordered retention steps]
    R -->|Provider fails| E[Record active error and continue]
    R -->|Step available| X[Drain deletion batches with cap]
    X --> S[Update entity statistics]
    E --> N[Process remaining providers]
    S --> N
    N --> F[Finish run and clear context]
```
</details>

<sub>Reviews (8): Last reviewed commit: ["test(apps): identify the triggered run b..."](https://github.com/open-metadata/openmetadata/commit/bf053581d02c5b9bd76a99d90f047ab80b9c31d7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56365306)</sub>


chirag-madlani (dismissed):

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/DataRetentionAppIT.java
```diff
@@ -107,6 +121,59 @@ void test_retentionRun_cleansOldConversationsFromThreadStorage(TestNamespace ns)
...
+
+    try {
+      for (int i = 0; i < APP_SCHEDULER_THREAD_COUNT; i++) {
+        triggerAppAndWaitForCompletion();
+      }
+
+      setChangeEventRetentionPeriod(app.getId(), changed);
+
+      for (int run = 0; run < RUNS_AFTER_CONFIG_CHANGE; run++) {
+        AppRunRecord record = triggerAppAndWaitForCompletion();
```
manerow, line 153: This assumes `triggerAppAndWaitForCompletion()` returns the run it just triggered. Now that it's called in a loop, I don't think it can.

The helper can't identify *which* run it sees: it polls `/runs/latest` (which is `ORDER BY timestamp DESC LIMIT 1`) and accepts the first terminal record started within the last 60s (L302). With 13 runs seconds apart, that window no longer separates run N from run N-1.

```
t=0.0s  warm-up #10 finishes    -> {SUCCESS, config: 7}
t=0.1s  PATCH config -> 11
t=0.3s  POST trigger returns    (async; nothing in DB yet)
t=0.8s  first poll (L287)       -> still #10; terminal? yes; within 60s? yes => ACCEPTED
        helper returns #10, config 7   =>  L149 fails
```

The window is wider than it looks: `AppScheduler` runs a clustered `JobStoreTX`, so the trigger is acquired through `QRTZ_LOCKS` rather than fired from memory, and `jobToBeExecuted` then does two more DB calls before writing the RUNNING record. On a DB that just took a full retention sweep, 500ms is not a comfortable margin for that. To be clear I haven't reproduced this, the argument is from the code path rather than an observed failure.

`ReindexHelpers` already hit and solved this exact race, and its Javadoc names it: *"distinguish 'the run I just triggered finished' from 'an older run is still marked Success'"*. `triggerSearchIndexAndWait` waits for the previous run to go terminal, captures a floor, triggers, then requires `timestamp >= floor` **and** terminal, with no skew window. The `ServerHandle` signature means those helpers can't be called from here directly, but the shape ports over:

```java
waitForLatestRunTerminal();                    // previous run is done
final long floor = System.currentTimeMillis();
triggerWhenAccepted();
// in the poll: reject unless run.getTimestamp() >= floor AND terminal
```

The `- 60_000` can go rather than be worked around. `TestSuiteBootstrap` runs the server in-process via `DropwizardAppExtension` (L563), so `triggerTime` here and `System.currentTimeMillis()` in `jobToBeExecuted` are the same JVM clock. There is no skew to tolerate, which is why `ReindexHelpers` needs none either.

Either way the failure mode today is worth avoiding: when it loses, L149 reports `run 0 used a stale appConfiguration ==> expected: <11> but was: <7>`, which reads exactly like the caching regression this PR fixes.

pmbrull: thanks aligned with how we do this in search ITs

manerow (approved):

chirag-madlani (approved):

pmbrull:
picked to 2.0

PR #31993, opened 2026-08-25
fix(ci): unbreak 1.13 nightly E2E by pinning playwright-slack-report

anuj-kumary:
## In one line

Our nightly tests have been showing **red every night since 22 Aug even though every single test passed**. The tests are fine — the last step, the one that posts results to Slack, is what breaks. This pins that Slack tool to a working version.

## What was actually happening

Each nightly run does two things at the end:

1. Merge the test results from all 8 parallel test runners into one file. ✅ This always worked.
2. Send those results to Slack. ❌ This is what broke.

Both were written as a single step, so when step 2 failed it dragged the whole run down with it — and GitHub showed a big red X on a run where **all 8 test runners were green**.

So nobody was looking at a real test failure. They were looking at a broken notification.

## Why it started on 22 Aug

We install the Slack tool fresh on every run, and we never said *which version* we wanted — so we always got "whatever is newest today."

On **22 Aug at 00:35 UTC**, the people who make that tool published a new version. That new version needs a newer Node.js than this branch uses. Our nightly started **3 minutes later, at 00:38**, grabbed the brand-new version, and couldn't install it.

Nothing in our code changed. We just got handed a new version of somebody else's tool overnight.

The pattern is very clear:

| Nightly started | Result |
|---|---|
| 21 Aug 00:41 | ✅ green |
| *22 Aug 00:35 — new version published* | |
| 22 Aug 00:38 | ❌ red |
| 23 Aug 00:42 | ❌ red |
| 24 Aug 00:38 | ❌ red |
| 25 Aug 00:38 | ❌ red |

Examples: [MySQL run](https://github.com/open-metadata/OpenMetadata/actions/runs/32794444648), [PostgreSQL run](https://github.com/open-metadata/OpenMetadata/actions/runs/32794438564).

## Why other branches were fine

- **`main`** — moved to a newer Node.js on 24 Aug (#31923), which is new enough for the new tool version.
- **`2.0`** — has exactly the same problem sitting there unnoticed. It only looks fine because its Slack step is currently switched off. **If anyone turns Slack notifications on for a `2.0` run, it will break the same way.** Worth fixing separately.

## The fix

Two small changes, both in the nightly workflow files:

1. **Ask for a specific version of the Slack tool** (`1.1.112`) instead of "whatever is newest." That version works with the Node.js this branch already uses. No surprise upgrades overnight any more.
2. **Split the merge and the Slack steps apart**, and mark the Slack one as "allowed to fail." So if the Slack tool ever breaks again, we lose the notification — but the test results stay honest and the run doesn't go red for no reason.

## Why not just upgrade Node.js instead?

That was the first version of this PR, and it was pulled back on purpose.

Upgrading Node.js would have changed the Node.js version used by **every** build and test job on `1.13`, just to fix a broken Slack message. That is a lot of risk on a release branch for a problem that never touched the product or the tests. Pinning the Slack tool fixes the same thing and touches nothing else.

We also didn't just mark the step "allowed to fail" and leave it there — that would turn the run green while Slack notifications quietly stopped arriving, which defeats the point of having a nightly report at all.

## How we know it works

Tested on a machine running the exact same Node.js version as CI:

| What was tested | Result |
|---|---|
| The new tool version (what CI grabbed) | ❌ fails with the identical CI error |
| **Version `1.1.112` (this fix)** | ✅ installs and runs fine |

Both workflow files were also re-checked to confirm the file structure is still valid after the edit.

## What reviewers should know

- Only two files change. No product code, no test code, no Node.js version changes.
- After this, the nightly goes back to being a real signal: red means a test actually failed.
- One small detail worth knowing: this step runs in a folder whose npm settings are set to "strict," which is why a version mismatch fails the build outright here instead of just printing a warning like it normally would.

The PR isolates Slack reporting from Playwright report merging in both database nightly workflows and pins the Slack reporter to version 1.1.112, preventing notification failures from failing otherwise successful E2E runs.
- Separates JSON report generation from Slack notification execution.
- Makes Slack notification failures non-fatal.
- Applies the same reporter pin and workflow structure to MySQL and PostgreSQL nightlies.
</details>

The PR appears safe to merge because no blocking failure remains in the reviewed changes.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| .github/workflows/mysql-nightly-e2e.yml | Separates MySQL report merging from non-fatal Slack reporting and pins the reporter invocation to version 1.1.112. |
| .github/workflows/postgresql-nightly-e2e.yml | Applies the same isolated, pinned Slack-reporting flow to the PostgreSQL nightly workflow. |

</details>

```mermaid
flowchart LR
  A[Download shard reports] --> B[Merge Playwright reports]
  B --> C[Write merged_tests_results.json]
  C --> D[Run pinned Slack reporter 1.1.112]
  D --> E[Post notification]
  D -. failure is non-fatal .-> F[Nightly remains successful]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(ci): pin playwright-slack-report ins..."](https://github.com/open-metadata/openmetadata/commit/d339b9b63604f6723daaab010643a30fbb384bba) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56514473)</sub>


anuj-kumary:
~~Follow-up commit bumping the Maven `<nodeVersion>` pins and core-components `.nvmrc` to v22.22.2, for parity with main's #31923.~~

**Outdated — this was reverted.** After review it was decided that upgrading Node.js across the whole `1.13` branch is too broad a change to fix a Slack-notification-only breakage. The PR now pins the Slack reporting tool to a known-good version instead, and touches only the two nightly workflow files. No Node.js version changes remain. See the updated description.

chirag-madlani (approved):

PR #4673, opened 2022-05-03
tableau-test-connection-added

codingwithabhi:
### Describe your changes :

I worked on the fixing test connection for Tableau.
fix #4406
#
### Type of change :

- [x] Bug fix
- [x] Improvement

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/open-source-community/developer) document.
- [x] I have commented on my code, particularly in hard-to-understand areas.

#
 @open-metadata/ingestion

PR #32038, opened 2026-08-25
feat(governance): hold approval gated changes off the entity until approved (#4673)

yan-3005:
## What changed

Adds an **approval gated pending change** capability to governance workflows. When a workflow opts in (by carrying a `resolvePendingChangeTask` hook), an edit to a gated field on a governed entity is **held off the entity**: it is reverted before it persists, with no version bump and no ChangeEvent, and stays held until the governing workflow **commits** it (applies it on approval) or **discards** it (drops it on rejection). Consumers keep seeing the approved value until the change is approved; the reviewer sees exactly what is proposed.

Resolves https://github.com/open-metadata/openmetadata-collate/issues/4673

## Why

The existing governance workflows react to a change **after** it is already persisted. There was no way to require approval **before** a change becomes visible: a tag, a description, or a status edit landed immediately, got indexed, and fired downstream events, and only then went through review. This PR moves the gate to the edit itself so an unapproved change never reaches search, lineage, or notifications.

## The flow

```
User edits a gated field
        │
        ▼
EntityRepository.update / patch ──► ApprovalGate.stageAndHold
        │                                   │
        │              (field is gated, filter does not exclude)
        │                                   ▼
        │                  hold the field diff per (entity, requester)
        │                  revert the field on the entity  ──► persist the APPROVED value
        │                                   │                  (no ChangeEvent for the gated field)
        │                                   ▼
        │                  ApprovalGate.submitPending ──► triggerWithSignal(held change)
        ▼                                   │
   nothing else                            ▼
                          Trigger (FilterEntityImpl) evaluates THIS edit's held change
                                            │  passes include/exclude + entity filter
                                            ▼
                     [checkChangeDescription] ─► [SetInReview] ─► create approval Task
                                            │
                          ┌─────────────────┴──────────────────┐
                       approve                                reject
                          ▼                                    ▼
             ResolvePendingChange COMMIT              ResolvePendingChange DISCARD
             apply the held change to the entity      drop the hold, entity stays approved
             (real ChangeEvent, indexed now)
                          ▼                                    ▼
                   delete the hold                       delete the hold
```

## Architecture

| Component | Responsibility |
|---|---|
| `ApprovalGate` | At edit time: resolve gating rules for the entity type, hold + revert the gated field, record the hold, and fire the review workflow by signal. Fails open (a gate error never breaks a write). |
| `PendingApprovalChangeStore` | The hold store, keyed by `(entityId, updatedBy)`. Successive edits by the same requester accumulate into one record. `effective(entity, updatedBy)` = the entity's persisted change description unioned with that requester's held change, so the workflow's "what changed" nodes see the proposed change. |
| `GovernanceApprovalRegistry` | Resolves the field gating rules for an entity type from deployed workflows that carry the hook (short TTL cache). `isPendingChangeWorkflow` tells per requester / suppression logic whether a workflow is a hook workflow. |
| `FilterEntityImpl` / `WorkflowTriggerFilters` | The trigger. Evaluates **this edit's** change (the held change carried on the gate signal, otherwise the persisted change) against `include` / `exclude` / `filter`, so the gate and the trigger always agree on what a workflow acts on. A hook workflow fires only on the gate's held change signal, never on an unrelated event. |
| `ResolvePendingChangeImpl` | The hook node. `commit` applies the held change as the requester (impersonated by the governance bot, exempt from re-gating); `discard` drops it; `hold` parks it. All three clear the hold. |
| `CheckChangeDescriptionTaskImpl` | Routes the workflow on the held change (via `effective`), tolerating a held only change (no `fieldsAdded` / `fieldsDeleted`). |
| `CreateTask` | Creates the approval task. The task requester and creator are the real editor (the workflow's `updatedBy`), not the reverted entity's updater. Supersede keeps one live approval per `(entity, workflow, requester)`. |
| `WorkflowDefinitionRepository` | Validates at create time that a hook workflow resolves the hold on every terminal path (a `commit` or `discard` must exist, and no path may reach an end without one). |

## Data model

New table `pending_approval_change` (added in `bootstrap/sql/migrations/native/2.1.0`, MySQL + Postgres):

| Column | Type | Notes |
|---|---|---|
| `entity_id` | `VARCHAR(36)` | part of the primary key |
| `updated_by` | `VARCHAR(256)` | part of the primary key: the requester, so each editor gets their own hold |
| `json` | `JSON` / `JSONB` | the held change as a `ChangeDescription` shaped diff |
| `updated_at` | `BIGINT` | last write |

The hold lives here, not on the entity row or its version history, so an unapproved change never touches the entity, its versions, or the search index.

## Per requester holds

| Scenario | Behaviour |
|---|---|
| Two users edit distinct gated fields | Two independent holds, two approval tasks, one per requester. Approving one and rejecting the other applies exactly one edit and leaves the other field at its approved value. |
| Same user edits the same field twice | The holds accumulate into one record; supersede keeps a single live task for that requester (the latest edit). |
| Task attribution | The task's `createdBy` and `updatedBy` are the editor who made the held change. |

## Mutual exclusivity

Two layers, so a mutually exclusive tag conflict is caught at the edit and can never fail a workflow node:

| Layer | Where | Behaviour |
|---|---|---|
| Reject at the edit | `ApprovalGate` | Two mutually exclusive tags in one gated edit are rejected up front (a 400), before they are held. The gate reverts tags before the entity updater's own check runs, so it validates the proposed tags itself. |
| Resolve at commit | `ResolvePendingChangeImpl` | A held tag that becomes mutually exclusive with a tag another approved change has since added is merged with incoming precedence at commit (the held tag wins, the conflicting current tag is dropped, non conflicting tags are kept). The commit node never throws, so one conflicting hold cannot fail the workflow for everyone else. |

## Entity agnostic

The gate and the hook operate on `EntityInterface` through `GovernanceApprovalRegistry.gatingRules(entityType)`, so any entity type works. A workflow scoped to `entityTypes: ["table"]` with the hook gates tables exactly like glossaries. Only the approval task's assignees differ: tables have owners but not reviewers, so the approval node adds owners.

## Validation and safety

- A hook workflow must resolve the hold on **every** terminal path: at least one `commit` or `discard` must exist, and no path may reach an end event with only a `hold`. Rejected at create time otherwise, so an edit can never be held forever.
- Clearing a gated scalar field (an explicit `PATCH remove`) is held for review, not written straight through. An omitted collection (owners, tags) is left alone so a `PUT` that leaves it out is not mistaken for a change.
- Structural and lifecycle fields (`name`, `fullyQualifiedName`, `deleted`, `entityStatus`) are never held.
- The gate fails open: any gate error lets the edit persist normally.

## Schema

`resolvePendingChangeTask.json` defines the hook node: `config.action` is an enum of `commit` / `discard` / `hold`, `config.action` is required, and the node takes the related entity as input.

## UI

- New `ResolvePendingChangeForm` and workflow builder wiring for the `resolvePendingChangeTask` node.
- Two upstream fixes to the workflow builder that this feature depends on: the event trigger exclude filter now emits **JsonLogic** (which the backend evaluates) instead of an Elasticsearch query, and clearing that filter now persists instead of retaining the previous value.

## Testing

| Level | Coverage |
|---|---|
| Integration (`PendingApprovalChangeIT`) | Held then committed / discarded / parked; include / exclude / filter selection; opt in (no workflow, no hook); PUT and PATCH; per requester holds resolved independently; task creator is the editor; mutual exclusivity rejected at the edit and resolved at commit; `checkChangeDescription` sees the held change (match creates a task, no match discards); clearing a gated field is held; a hold only workflow is rejected at create. |
| Unit | `CheckChangeDescriptionTaskImplTest`, `FilterEntityImplTest`, `CreateTaskTest` (per requester supersede and requester resolution), `GovernanceApprovalRegistryTest`, `PendingApprovalChangeStoreTest`, `ChangePreviewUtilsTest`, `WorkflowTriggerFiltersTest`. |

Verified end to end on a live server for glossary and table entities: held before approval, one task per requester with the correct creator, `checkChangeDescription` routing on the held change, mutual exclusivity, and the table workflow with `setStatus`.

The PR adds approval-gated entity changes that remain outside persisted entity state until governance workflows commit or discard them.
- Adds a transactional per-entity, per-requester pending-change store with row locking and monotonic version tokens.
- Applies held changes through a dedicated workflow node and conditionally deletes only the resolved snapshot.
- Adds workflow validation, trigger/filter integration, database migrations, UI configuration, and end-to-end coverage.
</details>

The PR appears safe to merge.

No blocking failure remains; row-level accumulation locking, monotonic hold versions, and compare-and-delete resolution address all previously reported concurrency failures.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/governance/approval/PendingApprovalChangeStore.java | Serializes same-requester accumulation with a row lock and advances a monotonic token, resolving the previously reported lost-update and timestamp-collision paths. |
| openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/impl/ResolvePendingChangeImpl.java | Resolves a captured hold snapshot and conditionally deletes it, leaving any concurrently accumulated newer state intact. |
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/CollectionDAO.java | Adds database-specific upsert, seed, locking-read, and compare-and-delete operations for pending approval records. |
| openmetadata-service/src/main/java/org/openmetadata/service/governance/approval/ApprovalGate.java | Stages selected entity changes outside the entity and signals the matching governance workflow. |
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/WorkflowDefinitionRepository.java | Validates that pending-change workflows resolve held state along terminal paths. |
| openmetadata-ui/src/main/resources/ui/src/components/WorkflowDefinitions/WorkflowBuilder/forms/ResolvePendingChangeForm.tsx | Adds workflow-builder configuration for commit, discard, and hold actions. |

</details>

```mermaid
sequenceDiagram
  participant Editor
  participant Gate as ApprovalGate
  participant Store as Pending Change Store
  participant Workflow
  participant Entity
  Editor->>Gate: Edit gated field
  Gate->>Store: Accumulate held diff
  Gate-->>Entity: Persist approved value only
  Gate->>Workflow: Signal held change
  alt Approved
    Workflow->>Store: Read held snapshot and version
    Workflow->>Entity: Apply held change
    Workflow->>Store: Delete if version unchanged
  else Rejected
    Workflow->>Store: Delete if version unchanged
  end
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(governance): make the pending-change..."](https://github.com/open-metadata/openmetadata/commit/a3e9fdf6216fe0bd54acc01db765473c8d12de09) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56664496)</sub>

**Context used:**

- Knowledge Base — [Metadata entity lifecycle](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/metadata-entity-lifecycle.md)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/WorkflowDefinitionRepository.java
```diff
@@ -189,10 +203,97 @@ public void validateWorkflow(WorkflowDefinition workflowDefinition) {
...
+  private void validatePendingChangeResolution(WorkflowDefinition workflowDefinition) {
+    List<WorkflowNodeDefinitionInterface> nodes = listOrEmpty(workflowDefinition.getNodes());
+    Set<String> hooks = new HashSet<>();
+    Set<String> ends = new HashSet<>();
+    String start = null;
+    boolean hasResolvingHook = false;
+    for (WorkflowNodeDefinitionInterface node : nodes) {
+      String subType = node.getSubType();
+      if (NodeSubType.RESOLVE_PENDING_CHANGE_TASK.value().equals(subType)) {
+        hooks.add(node.getName());
```
mohityadav766, line 227: **Blocker — a `hold`-only terminal path passes this validation.**

`hooks` collects *every* `resolvePendingChangeTask` node regardless of its action, and `reachesEndWithoutHook` then treats any hook on the path as "the hold was resolved". `hasResolvingHook` only proves a commit/discard exists *somewhere* in the workflow, not on the path being walked.

So this shape is accepted today:

```
Start -> Approve -(approve)-> Commit -> ApprovedEnd
               \-(reject)--> ParkIt(action=hold) -> RejectedEnd
```

The reject path reaches an end event with the hold still in place, which is exactly the "edit held forever" case this method exists to prevent.

Fix is small — only commit/discard nodes belong in the set the traversal consults:

```java
if (NodeSubType.RESOLVE_PENDING_CHANGE_TASK.value().equals(subType)) {
  if (isCommitOrDiscardHook(node)) {
    hooks.add(node.getName());
    hasResolvingHook = true;
  }
}
```

Worth a unit test for the mixed commit/hold workflow — the existing IT only covers the all-hold case, which is caught by the other branch.

openmetadata-service/src/main/java/org/openmetadata/service/governance/approval/GovernanceApprovalRegistry.java
```diff
@@ -0,0 +1,193 @@
...
+        (EntityRepository<WorkflowDefinition>)
+            Entity.getEntityRepository(Entity.WORKFLOW_DEFINITION);
+    return repository.listAll(Fields.EMPTY_FIELDS, new ListFilter(Include.NON_DELETED));
+  }
+
+  static void addRule(String entityType, WorkflowDefinition definition, List<GatingRule> rules) {
+    JsonNode trigger = JsonUtils.valueToTree(definition.getTrigger());
+    boolean eventBased = EVENT_BASED_ENTITY.equals(trigger.path("type").asText(null));
+    JsonNode config = trigger.path("config");
+    if (eventBased && targetsEntityType(config, entityType) && hasPendingChangeHook(definition)) {
```
mohityadav766, line 140: **Blocker — the rule ignores the trigger's `events`, so a workflow that never listens for `Updated` still gates updates.**

`ApprovalGate.triggerReviewWorkflow` always raises `"%s-%s".formatted(entityType, ENTITY_UPDATED)`, and `EventBasedEntityTrigger.setStartEvents` only registers signal catches for the events in `config.events`. A hook workflow configured with `events: ["Created"]` therefore contributes a gating rule here, holds every *update* on that entity type, and nothing ever catches the signal — the edit is held with no task and no way out (see also the orphan-hold comment on `PendingApprovalChangeStore`).

Suggest requiring the Updated event before a rule is contributed:

```java
if (eventBased
    && targetsEntityType(config, entityType)
    && triggersOnUpdate(config)          // config.events contains "Updated"
    && hasPendingChangeHook(definition)) {
```

The class doc says a rule mirrors the trigger's own selection — `events` is part of that selection and is currently the one piece left out.

openmetadata-service/src/main/java/org/openmetadata/service/governance/approval/ApprovalGate.java
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  // Records this edit's held fields into the requester's accumulating store record and returns the
+  // single-edit change so the trigger raised below can review exactly this edit, not the
+  // accumulation.
+  private static ChangeDescription recordHold(
+      EntityInterface original, List<FieldChange> held, String user) {
+    ChangeDescription pending =
+        new ChangeDescription().withPreviousVersion(original.getVersion()).withFieldsUpdated(held);
+    PendingApprovalChangeStore.accumulate(original.getId(), user, pending);
```
mohityadav766, line 292: **Blocker — the hold is committed independently of the entity write, but only submitted after it.**

`PendingApprovalChangeStore.accumulate` runs `Entity.getJdbi().useTransaction(...)`, which takes a fresh handle and commits on its own, while `submitPending` fires post-commit from `EntityUpdater.reactUpdate`. Anything that throws between the two — validation in the updater, an optimistic-lock 409, a constraint violation, a mutual-exclusivity failure on a *non-gated* field — leaves a durable hold row and no workflow run.

That is worse than a leak, because holds accumulate per `(entity, requester)` and `ResolvePendingChangeImpl.commitHeldChange` applies the whole accumulated record:

1. user PATCHes `description` → held, task A created
2. user PATCHes again → gate holds and merges into the same row → the request then fails with a 4xx
3. reviewer approves task A → the merged hold is committed, so the edit the user was told had failed silently lands on the entity

Either write the hold on the same connection/transaction as the entity write so it rolls back with it, or record it optimistically and compensate (delete the just-accumulated field diff) when the write throws. A try/catch around the remainder of the write is not enough on its own — the failure happens further down the stack in `EntityUpdater`.

openmetadata-service/src/main/java/org/openmetadata/service/governance/approval/ApprovalGate.java
```diff
@@ -0,0 +1,313 @@
...
+    } catch (Exception e) {
+      LOG.error("[ApprovalGate] Failed to trigger review workflow for '{}'", entity.getId(), e);
+    } finally {
+      PENDING_TRIGGER.remove();
+    }
+  }
+
+  private static boolean isGateApplicable(
+      String user, EntityInterface original, EntityInterface updated) {
+    return !Boolean.TRUE.equals(APPLYING_APPROVED_CHANGE.get())
```
mohityadav766, line 153: **Blocker — writes the workflow itself makes get re-gated.**

`isGateApplicable` exempts a write only when `updatedBy` is the governance bot, or when the thread-local from `applyExemptFromGate` is set. But this PR changes `SetEntityAttributeImpl` to attribute writes to the *real user* with `governance-bot` as `impersonatedBy` (see my comment there), so a `setEntityAttribute` node that touches a gated field — `certification`, `tags`, `owners`, anything in the trigger's `include` — now arrives here as an ordinary user edit and gets held. On a workflow that sets a gated attribute after approval, that means the workflow's own output is held pending approval, plausibly re-triggering the same workflow.

`entityStatus` escapes this only because it happens to sit in `STRUCTURAL_DENYLIST`, which is a fragile reason to be safe.

`EntityRepository` sets `updated.setImpersonatedBy(impersonatedBy)` immediately before calling `stageAndHold`, so the signal is already on the entity:

```java
return !Boolean.TRUE.equals(APPLYING_APPROVED_CHANGE.get())
    && !GOVERNANCE_BOT.equals(user)
    && !GOVERNANCE_BOT.equals(updated.getImpersonatedBy())
    && original != null
    ...
```

openmetadata-service/src/main/java/org/openmetadata/service/governance/approval/ApprovalGate.java
```diff
@@ -0,0 +1,313 @@
...
+   * commit runs synchronously on the caller's thread, so the entity write reaches {@link
+   * #stageAndHold} on the same thread and observes this flag. Lets the hook apply the change as its
+   * original author (not a bot), which the field-level bot-deny guards would otherwise strip.
+   */
+  public static void applyExemptFromGate(Runnable apply) {
+    APPLYING_APPROVED_CHANGE.set(Boolean.TRUE);
+    try {
+      apply.run();
+    } finally {
+      APPLYING_APPROVED_CHANGE.remove();
```
mohityadav766, line 91: Nit: `remove()` in the `finally` clears the flag unconditionally, so a nested `applyExemptFromGate` (a commit whose patch cascades into another gated write on the same thread) un-exempts the outer one on its way out. Save and restore instead:

```java
Boolean previous = APPLYING_APPROVED_CHANGE.get();
APPLYING_APPROVED_CHANGE.set(Boolean.TRUE);
try {
  apply.run();
} finally {
  if (previous == null) {
    APPLYING_APPROVED_CHANGE.remove();
  } else {
    APPLYING_APPROVED_CHANGE.set(previous);
  }
}
```

openmetadata-service/src/main/java/org/openmetadata/service/governance/approval/PendingApprovalChangeStore.java
```diff
@@ -0,0 +1,187 @@
...
+  /**
+   * Delete the hold only if it still matches the version token it was read at. A no-op when a newer
+   * edit accumulated after the caller's snapshot, so that newer proposal survives for its own review.
+   */
+  public static void deleteIfUnchanged(UUID entityId, String updatedBy, long seenUpdatedAt) {
+    dao().deleteIfUnchanged(entityId, updatedBy, seenUpdatedAt);
+  }
+
+  /** Remove every requester's hold for an entity; used when the entity itself is deleted. */
+  public static void deleteAllForEntity(UUID entityId) {
```
mohityadav766, line 130: **Blocker — nothing resolves a hold whose governing workflow is gone.**

`deleteAllForEntity` is only wired into `EntityRepository.cleanup` (entity delete). There is no path that clears a hold when:

- the workflow definition is deleted or edited so it no longer gates the field (`GovernanceApprovalRegistry.invalidate()` stops *new* holds; existing ones are stranded),
- the workflow instance fails, is terminated, or times out before reaching a commit/discard node,
- the signal is raised but no process is listening (see my comment on `GovernanceApprovalRegistry.addRule`).

From the user's side this is invisible: the PUT/PATCH returns 200 with the approved value and their edit simply never appears. The new `GET /{id}/pendingChanges` shows the hold but offers no way to clear it.

Could we add an admin-only discard (`DELETE /{id}/pendingChanges?user=...`, EDIT_ALL-gated), and ideally a reaper or a hold TTL for instances that die? Without one, operators have no recovery short of a manual SQL delete.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/impl/ResolvePendingChangeImpl.java
```diff
@@ -0,0 +1,183 @@
...
+  private void applyHeldChange(
+      String entityType, EntityInterface original, ChangeDescription held, String author) {
+    ObjectNode node = (ObjectNode) JsonUtils.valueToTree(original);
+    for (FieldChange fieldChange : CommonUtil.listOrEmpty(held.getFieldsUpdated())) {
+      if (WorkflowTriggerFields.TAGS.value().equals(fieldChange.getName())) {
+        node.set(
+            WorkflowTriggerFields.TAGS.value(),
+            JsonUtils.valueToTree(resolveTags(original, fieldChange)));
+      } else {
+        node.set(fieldChange.getName(), JsonUtils.valueToTree(fieldChange.getNewValue()));
```
mohityadav766, line 147: Commit is last-write-wins for every field except `tags`. If another approved change touches the same field while this hold is pending, approving here silently overwrites it — `held.getNewValue()` is the value captured at edit time, and nothing compares against the entity's current state.

`resolveTags` handles the tag case carefully, which makes the asymmetry more surprising for `description` / `displayName` / `owners`.

At minimum call it out in the node's javadoc and the feature docs. Better: compare `held.getPreviousVersion()` against the entity's current version and route a stale hold somewhere visible (BPMN error, or surface the conflict on the task) rather than committing over the newer value.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/triggers/impl/FilterEntityImpl.java
```diff
@@ -201,79 +149,102 @@ private boolean isTagFeedbackCreation(WorkflowVariableHandler varHandler) {
...
+  // resolvePendingChange hook). Resolved from the running process definition key: the trigger
+  // process is named "<Workflow>Trigger". An unresolved name yields false, so non-hook workflows
+  // and any resolution miss keep their normal event-driven behavior.
+  private boolean isPendingChangeWorkflow(DelegateExecution execution) {
+    String key = WorkflowHandler.getProcessDefinitionKeyFromId(execution.getProcessDefinitionId());
+    String workflowName =
+        key != null && key.endsWith("Trigger")
+            ? TriggerFactory.getMainWorkflowDefinitionNameFromTrigger(key)
+            : key;
+    return GovernanceApprovalRegistry.isPendingChangeWorkflow(workflowName);
```
mohityadav766, line 162: Perf: `isPendingChangeWorkflow(String)` does an uncached `Entity.findByNameOrNull(WORKFLOW_DEFINITION, ...)` — a DB read — and this runs on **every** entity change event, for every event-based workflow's trigger, gated or not. On a busy instance that's a per-event fan-out of workflow-definition reads on the hot path.

`GovernanceApprovalRegistry` already owns a Caffeine cache with a 60s TTL that's invalidated from `WorkflowDefinitionRepository.postCreate/postUpdate/postDelete`. Please put the hook lookup behind the same cache (keyed by name and by id) so it inherits both the TTL and the invalidation.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/userTask/CreateTask.java
```diff
@@ -1177,6 +1211,49 @@ private static RecognizerFeedbackTaskPayload readRecognizerFeedbackTaskPayload(O
...
+    return ref;
+  }
+
+  String resolveRequester(
+      DelegateTask delegateTask,
+      UUID workflowDefinitionId,
+      EntityInterface entity,
+      EntityReference createdByRef) {
+    String requester = null;
+    if (GovernanceApprovalRegistry.isPendingChangeWorkflow(workflowDefinitionId)) {
```
mohityadav766, line 1241: Same caching point as `FilterEntityImpl`, but multiplied: `isPendingChangeWorkflow` is called up to four times per approval-task creation — here in `resolveRequester`, in the `createdBy` alignment block above, in `supersedePriorApprovalTask` (L734), and in `findPriorOpenApprovalPayload` (L778) — each one a full `Entity.getEntity(WORKFLOW_DEFINITION, ...)`. This happens for **every** approval task, including workflows that have nothing to do with pending changes.

Resolving it once per `createOrUpdateTask` and threading the boolean through would be an easy win even before the registry-level cache.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/automatedTask/impl/SetEntityAttributeImpl.java
```diff
@@ -55,9 +56,19 @@ public void execute(DelegateExecution execution) {
           Optional.ofNullable(updatedByNamespace)
               .map(ns -> (String) varHandler.getNamespacedVariable(ns, UPDATED_BY_VARIABLE))
               .orElse(null);
+      if (actualUser == null || actualUser.isEmpty()) {
+        // The configured namespace had no acting user (e.g. an auto-approved user task was bypassed
+        // and set no updatedBy). Use the global editor (the requester) so the write is attributed
+        // to
+        // a real user with governance-bot as impersonator.
+        actualUser =
+            (String) varHandler.getNamespacedVariable(GLOBAL_NAMESPACE, UPDATED_BY_VARIABLE);
```
mohityadav766, line 65: This fallback is reasonable in itself, but combined with the `impersonatedBy` gap in `ApprovalGate.isGateApplicable` it has a side effect worth calling out: the write now reaches `stageAndHold` as `updatedBy = <real user>` with `impersonatedBy = governance-bot`, so the gate does **not** skip it. Any `setEntityAttribute` node writing a gated field will have that write held for approval instead of applied.

Fixing the gate to also exempt `impersonatedBy == governance-bot` (see my comment there) covers this; just flagging the coupling so the two aren't changed independently later.

openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/elements/nodes/userTask/impl/SetApprovalAssigneesImpl.java
```diff
@@ -156,19 +156,17 @@ public void execute(DelegateExecution execution) {
...
+      // lets the userApprovalTask auto-approve (event-driven) or fall through to the admin fallback
+      // below (workflow-managed). Re-adding the requester "to keep the task actionable" would
+      // silently reintroduce self-approval, so it is deliberately not done.
       Set<String> requesterEntityLinks = resolveRequesterEntityLinks(varHandler, execution);
-      List<String> preRemovalAssignees = new ArrayList<>(assigneeList);
-      boolean removedRequester = assigneeList.removeAll(requesterEntityLinks);
-      if (removedRequester && assigneeList.isEmpty() && !workflowManagedTask) {
-        assigneeList.addAll(preRemovalAssignees);
-      }
+      assigneeList.removeAll(requesterEntityLinks);
```
mohityadav766, line 169: **This is a behaviour change to existing approval flows and needs an explicit decision, not a side effect of this PR.**

Removing the "re-add the requester when the list would otherwise be empty, for non-workflow-managed tasks" fallback means that for an entity whose only reviewer *is* the editor, the assignee list is now empty — and per the new comment right above, an empty list lets the `userApprovalTask` **auto-approve** on the event-driven path.

So today: a task assigned to the requester (self-approval, but a human action that is recorded). After this: approved automatically with no human in the loop. Under this feature that also auto-commits the held change, which is the opposite of what an approval gate is for.

Two asks:
1. Confirm this is intended, and say so in the PR description — it changes glossary term / tag approval behaviour for everyone, not just hook workflows.
2. Add a test for "the only reviewer is the requester" on both paths (event-driven and workflow-managed) so whichever semantics we choose is pinned down.

If the intent was only to stop self-approval on *pending-change* workflows, scoping the change to those keeps the blast radius proportional to the feature.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TaskRepository.java
```diff
@@ -1030,6 +1036,33 @@ private boolean isApprovalTask(Task task) {
...
+  /**
+   * Resolving an approval task on a governed entity that is IN_REVIEW drives the reviewer-gated
+   * status transition enforced by {@link EntityRepository.EntityUpdater#checkUpdatedByReviewer}.
+   * Mirror that gate here so a non-reviewer is rejected with 403 up front, instead of resolving the
+   * task and letting the governance workflow instance fail on the status update downstream. The
+   * check no-ops when the entity has no reviewers (preserving the admin/owner approval path) and
+   * when the target is not a governed, in-review entity (so DAR and non-status approvals are
+   * unaffected — their targets return a null entityStatus).
+   */
+  private void enforceReviewerForApprovalStatusTask(Task task, SecurityContext securityContext) {
```
mohityadav766, line 1048: Scope: this applies to **all** approval tasks (`GlossaryApproval`, `RequestApproval`, `DataAccessRequest`), not just pending-change ones. Converting a downstream workflow failure into an upfront 403 is a real improvement, but note `checkUpdatedByReviewer` has no admin bypass — an admin who is not a reviewer can resolve such a task today and will get a 403 after this. That's a user-visible change for existing deployments and belongs in the PR description (or, cleanly, in its own PR).

Also: this adds an `Entity.getEntity(..., FIELD_REVIEWERS, ...)` read on every non-close task resolution. The `getAllowedFields` guard above avoids it for types without reviewers, which helps, but for glossary terms it's an extra fetch per resolve on a path that already loads the task. Could it reuse an entity already resolved earlier in `checkPermissionsForResolveTask`?

mohityadav766 (commented):
Reviewed the whole change: gate/store/hook, trigger + workflow validation, DAO/migration, and the UI. The design is solid — sharing field selection between the gate and the trigger via `WorkflowTriggerFilters` is the right call, the per-requester hold with a monotonic version token plus compare-and-delete is correct, and the IT coverage is genuinely broad.

Before this can go in, though, there are a few paths where an edit gets held with no way to ever resolve it, and three behaviour changes to *existing* approval flows that are wider than the PR title suggests.

### Blockers

1. **A `hold`-only terminal path passes validation** — `validatePendingChangeResolution` puts every `resolvePendingChangeTask` into the traversal set regardless of action, so a reject branch that only parks the change is accepted and the edit stays held forever.
2. **Gating rules ignore the trigger's `events`** — the gate always raises `<entityType>-entityUpdated`, but a hook workflow subscribed only to `Created` still produces gating rules, so it holds updates nothing will ever review.
3. **The hold is committed outside the entity write** — `accumulate()` opens its own transaction and `submitPending` only runs post-commit, so a write that fails after the gate leaves a durable hold with no task; because holds accumulate per requester, those fields then ride along and get applied when an earlier, unrelated task is approved.
4. **No resolution path for an orphaned hold** — nothing clears holds when the governing workflow is deleted/suspended or its instance dies, and the read API is GET-only.
5. **Workflow-authored writes are re-gated** — `isGateApplicable` only exempts `updatedBy == governance-bot`, but this PR makes `SetEntityAttributeImpl` write as the real user with the bot as `impersonatedBy`, so a node touching a gated field gets held instead of applied.

### Scope creep that needs explicit sign-off

6. `SetApprovalAssigneesImpl` drops the "keep the requester when the list would otherwise be empty" fallback. Per the code's own comment an empty assignee list lets the task **auto-approve** — so an entity whose only reviewer is the editor goes from "task assigned to them" to "approved with no human", which under this feature auto-commits the held change. That affects existing glossary/tag approvals, not just the new flow.
7. `TaskRepository.enforceReviewerForApprovalStatusTask` now 403s any non-reviewer on an IN_REVIEW entity with reviewers, and `checkUpdatedByReviewer` has no admin bypass.
8. The workflow builder's trigger filter flips from ElasticSearch to JsonLogic with no migration for filters already saved.

### CI

- `openmetadata-service-unit-tests`: the 9 failures are all `OpenSearch/ElasticSearchBulkSinkBehaviorTest.enrichWithEmbedding*` — unrelated to this PR, stale branch. Please rebase and confirm green.
- `ui-checkstyle` and `Validate PR Metadata` are this PR's own and need fixing (Prettier/ESLint on 5 files, the antd/less guard, and a `Fixes open-metadata/openmetadata-collate#4673` line for the linked-issue check).
- Playwright green (1318 passed), both SonarCloud gates passed.

Inline comments below with the specifics.

openmetadata-service/src/main/java/org/openmetadata/service/resources/EntityResource.java
```diff
@@ -1615,6 +1620,66 @@ public Response getAiContextById(
...
+    // Owners and reviewers of the entity (and admins) may see every requester's proposal; any other
+    // caller sees only their own held change. Enforced here so a non-privileged user cannot read
+    // someone else's proposal by passing the `user` query param. Reuse the ResourceContext already
+    // built for authorize() (the same accessors RuleEvaluator uses) instead of re-fetching.
+    String caller = securityContext.getUserPrincipal().getName();
+    SubjectContext subject = SubjectContext.getSubjectContext(caller);
+    boolean seeAll =
+        subject.isAdmin()
+            || subject.isOwner(resourceContext.getOwners())
+            || subject.isReviewer(resourceContext.getEntity().getReviewers());
```
mohityadav766, line 1660: `ResourceContext.resolveEntity()` swallows `EntityNotFoundException` and leaves `entity` null, so for a non-existent id a non-admin caller NPEs here (500) rather than getting the documented 404. Admins are accidentally safe because `subject.isAdmin()` short-circuits the `||`.

A null check that throws `EntityNotFoundException` before the `seeAll` computation would match the `@ApiResponse(responseCode = "404")` this endpoint already advertises.

(The permission model itself looks right — forcing `requester = caller` for unprivileged callers is the correct way to stop the `user` param being used to read someone else's proposal. `authorizationFields()` does load `reviewers` when the type supports them, so the reviewer branch works.)

openmetadata-ui/src/main/resources/ui/src/components/WorkflowDefinitions/WorkflowBuilder/forms/EventTriggerFilterSection.tsx
```diff
@@ -98,7 +98,7 @@ export const EventTriggerFilterSection: React.FC<
                 entityTypes={entityType}
                 forceReadOnly={lockFields}
                 label={t('label.exclude-filter')}
-                outputType={SearchOutputType.ElasticSearch}
+                outputType={SearchOutputType.JSONLogic}
```
mohityadav766, line 101: **Blocker — this changes the serialized format of an existing field with no migration.**

Workflows saved before this change have an ElasticSearch query string in `trigger.config.filter`. After the upgrade the backend feeds that to `RuleEngine` via `WorkflowTriggerFilters.matchesExclusionFilter`, which returns false on anything it can't evaluate — and false means "not excluded". So every workflow with an existing exclude filter silently starts firing on the entities it used to skip, with no error anywhere.

That's a silent scope widening of deployed governance workflows on upgrade, and now it also decides what the *gate* holds.

Needs one of:
- a migration in `bootstrap/sql/migrations/native/2.1.0` converting saved ES filters to JsonLogic, or
- backward-compatible handling in `extractEntitySpecificFilter` (detect an ES-shaped filter and either convert it or log loudly and treat the workflow as misconfigured rather than unfiltered).

Either way this deserves a release note.

openmetadata-ui/src/main/resources/ui/src/components/PendingChanges/PendingChangesNotification.component.tsx
```diff
@@ -0,0 +1,490 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+import {
+  BellOutlined,
+  CheckOutlined,
+  CloseOutlined,
+} from '@ant-design/icons';
+import { Badge, Button, Popover, Select, Spin, Tooltip, Typography } from 'antd';
```
mohityadav766, line 19: `ui-checkstyle` is failing on this: the Antd + Less deprecation guard flags both this `antd` import and the new `pending-changes-notification.less`. New components are expected to use UntitledUI + Tailwind (`docs/ui-code-quality-gate.md`).

Prettier/ESLint is also red on this file plus `DataAssetsHeader`, `DataProductsDetailsPage`, `DomainDetails` and `GlossaryHeader` — `yarn lint:fix && yarn prettier:fix` should clear those.

openmetadata-ui/src/main/resources/ui/src/components/PendingChanges/PendingChangesNotification.component.tsx
```diff
@@ -0,0 +1,490 @@
...
+import { getEntityFeedLink } from '../../utils/EntityPureUtils';
+import { getUserPath } from '../../utils/RouterUtils';
+import { showErrorToast, showSuccessToast } from '../../utils/ToastUtils';
+import './pending-changes-notification.less';
+
+const { Text } = Typography;
+
+// Maps an entity type to its REST collection so the bell can fetch that entity's held changes.
+// Only mapped types show the bell; unmapped ones render nothing (harmless).
+const ENTITY_COLLECTION: Record<string, string> = {
```
mohityadav766, line 49: This hardcodes ~35 entityType → REST-collection mappings that the UI already maintains centrally (`getEntityAPIfromSource` / the entity-type endpoint maps). A second copy will drift the first time an entity type is added, and the failure mode is silent — an unmapped type just renders no bell, so nobody notices the pending-change UI quietly disappeared for it.

Please reuse the existing mapping util instead.

PR #32065, opened 2026-08-25
Fixes #32063: keep auth tokens in memory when the service worker cannot be reached

### LINKED ISSUE #32063: UI: successful basic-auth login is silently dropped when ServiceWorker registration fails — login loops back to Sign In (2.0.0)
### Affected module

UI

### Describe the bug

With basic auth, `POST /api/v1/auth/login` succeeds (HTTP 200, valid access token, `OM_SESSION` cookie set) — but if the browser fails to register the UI's service worker (`/app-worker.js`), the token returned by the server is **silently discarded** by the client-side token storage layer. The very next request (`GET /api/v1/users/loggedInUser`) goes out without an `Authorization` header, gets `401 {"message":"Not Authorized! Token not present"}`, and the UI bounces back to the Sign In page.

The result is a **complete, silent login lockout**: correct credentials, successful server-side authentication, no error toast, nothing actionable in the UI. From the operator's point of view it looks exactly like "the admin password doesn't work".

Service worker registration fails in real deployments for many reasons the server/UI don't control:

- reverse proxy / ingress rules that rewrite or 404 `/app-worker.js` (SPA catch-all rewrites are a classic — the script comes back as `text/html`);
- CSP `worker-src` restrictions injected by a proxy;
- corporate proxies or browser policies that disable service workers;
- plain-HTTP deployments on non-localhost hosts (no secure context);
- browser privacy modes with SW/IndexedDB disabled.

Note: #31897 already works around this class of failure **for the Playwright test infra only** ("skip the service worker wait on insecure origins"). The production code path has no equivalent fallback.

**Root cause** — `openmetadata-ui/src/main/resources/ui/src/utils/SwTokenStorageUtils.ts`:

- `isServiceWorkerAvailable()` only checks **API presence** (`'serviceWorker' in navigator && 'indexedDB' in window`) — it says nothing about whether a worker can actually be registered/controls the page. In every environment above the API exists, so the code commits to the SW path.
- `setOidcToken()` → `setAppState()` → `swTokenStorage.setItem()` → `SwMessenger.waitForServiceWorkerController()`, which tries to register `/app-worker.js` and **rejects** (or times out after 15 s). The `catch` blocks in `setAppState`/`setOidcToken` swallow the error ("Storage failures are intentionally ignored…"), so the token evaporates.
- The next axios request reads `getOidcToken()` → `''` → no `Authorization` header → 401 → the app treats the user as unauthenticated.

There is already a `localStorage` fallback — but only for the "API absent" branch, which is not the branch these environments hit. `getAppState()` even documents (for security reasons) that it deliberately does **not** fall back to `localStorage` when the SW is "not ready", which makes the failure unrecoverable by design.

An extra side effect: every storage call in this state burns the 15-second `waitForServiceWorkerController` timeout, so the login page also feels frozen.

### To Reproduce

1. Run OpenMetadata 2.0.0 with basic auth (release `docker-compose-postgres.yml` is enough; `AUTHORIZER_ADMIN_PRINCIPALS=[admin]`, no SMTP).
2. Make service worker registration fail — any one of:
   - DevTools → Network → add a request-blocking rule for `**/app-worker.js`, or
   - front the server with any proxy that 404s `/app-worker.js`, or
   - use a browser profile with service workers disabled.
3. Open the UI and log in with valid credentials (`admin@example.com` / `admin`).

Observed:

- Network tab: `POST /api/v1/auth/login` → **200** (body contains a valid `accessToken`; response sets `OM_SESSION` cookie), immediately followed by `GET /api/v1/users/loggedInUser` → **401** `{"code":401,"message":"Not Authorized! Token not present"}`.
- Console: `Failed to register a ServiceWorker for scope ('…') … An unknown error occurred when fetching the script.`
- UI: returns to the Sign In page with empty fields. No error message related to the actual cause.
- The `accessToken` from that same 200 response works fine when replayed manually (`curl -H "Authorization: Bearer <token>" …/api/v1/users/loggedInUser` → 200), proving the failure is purely client-side storage.

### Expected behavior

Login should succeed even when the service worker cannot be registered — the token storage layer should degrade gracefully (e.g., an in-memory fallback for the lifetime of the tab) instead of silently dropping the token. If storage truly fails, the user should see an actionable error rather than being bounced back to the Sign In page with no explanation.

Suggested fix (keeping the security posture of not persisting tokens to `localStorage`):

1. Add a **module-level in-memory fallback store**: when the SW path throws (registration failure, controller timeout, messaging timeout), read/write the app state from memory instead of dropping it. Login then works normally for the lifetime of the tab; on reload the `OM_SESSION` cookie + `/auth/refresh` can re-mint a token into memory again.
2. Cache the "SW is broken" verdict after the first failure so subsequent calls don't each pay the 15 s timeout.
3. Surface the condition (a single `console.error`, ideally a UI hint) instead of swallowing it — operators currently get zero signal that their ingress is eating `/app-worker.js`.

Happy to send a PR with the above.

### OS

macOS 15 (Docker Desktop) — bug is host-OS independent

### Python version

N/A (no ingestion involved)

### OpenMetadata version

2.0.0 (`docker.getcollate.io/openmetadata/server:2.0.0`, release compose artifacts) — reproduced on a clean install **and** on a 1.13.4 → 2.0.0 in-place upgrade; code is unchanged on current `main`

N/A

### Additional context

Aside for anyone who lands here from "can't log in as admin after upgrading to 2.0.0": the server-side behavior is intact and matches the docs — with SMTP disabled, initial admins are still bootstrapped with the default password `admin` (the `EmailUtil — Email was not sent … as SMTP setting is not enabled` WARN is benign). You can prove your credentials are fine with:

```bash
curl -s -X POST https://<host>/api/v1/users/login \
  -H 'Content-Type: application/json' \
  -d "{\"email\":\"admin@example.com\",\"password\":\"$(printf 'admin' | base64)\"}"
```

(username must be the **full email**; the password must be base64 of the exact password — beware `echo` adding a trailing newline). If that returns 200 but the UI still won't log you in, check whether `https://<host>/app-worker.js` is reachable from the browser and whether the console shows a ServiceWorker registration error — that's this bug.

### Pre-submission checklist

neochaotic:
### Describe your changes:

Fixes #32063

When the service worker API is present but a worker cannot actually be registered (reverse proxy 404s `/app-worker.js`, CSP `worker-src`, insecure origin, service workers disabled by policy), every `swTokenStorage` call rejects and `SwTokenStorageUtils` silently swallowed the failure. The practical effect was a **complete, silent login lockout**: `POST /api/v1/auth/login` returns 200 with a valid token, the token is dropped by `setOidcToken()`, the next request goes out without `Authorization`, gets 401, and the UI bounces back to Sign In (full analysis and reproduction in the issue).

This PR makes the SW-backed token storage degrade gracefully instead of dead-ending:

- **In-memory fallback**: when a `swTokenStorage` call rejects, the app state is kept in a module-level in-memory store for the lifetime of the tab. Login works normally; on reload the `OM_SESSION` cookie + `/auth/refresh` re-mint a token into memory (verified manually, see below). The fallback deliberately does **not** use `localStorage`, preserving the documented security posture (tokens must not be restorable from persistent storage after logout).
- **Fail-fast after first failure**: the broken-SW verdict is cached, so subsequent storage calls stop paying the 15-second `waitForServiceWorkerController` timeout on every request.
- **Surfaced diagnostics**: a single `console.error` tells operators the worker is unreachable and to check `/app-worker.js`, instead of swallowing the error silently.
- A malformed-JSON app state keeps its existing behavior (returns `{}`) and does **not** mark the service worker as broken.

Note: `OidcTokenStorage.ts` / `OktaCustomStorage.ts` share the same `isServiceWorkerAvailable()` assumption for SSO flows; they propagate errors instead of swallowing them, so they fail loudly rather than silently. Left out of scope to keep this fix focused — happy to follow up if there's interest.

#
### Type of change:

- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered

- User logs in with valid basic-auth credentials while the service worker cannot register (blocked/rewritten `/app-worker.js`, CSP, insecure origin, SW disabled) → login now succeeds instead of silently bouncing back to Sign In.
- OIDC and refresh tokens survive in memory when only the SW write fails, when both read and write fail, and are cleared by `clearOidcToken` (logout).
- After the first SW failure, subsequent storage calls do not hit the service worker again (no repeated 15 s waits).
- Operators get a single `console.error` describing the failure.
- No behavior change when the SW works, when the SW API is absent (`localStorage` path), or for malformed stored state.

#### Unit tests

- Written test-first: the new tests in `SwTokenStorageUtils.test.ts` reproduce the token loss against the previous implementation (4 failing) and pass with this change (32/32).
- Files added/updated: `src/utils/SwTokenStorageUtils.test.ts` (new `service worker registration failure fallback` and `clearOidcToken` suites; `removeItem` added to the mocks; state reset in `beforeEach`, required now that the module tracks the broken-SW verdict).
- Coverage on changed class (`yarn test:coverage`): **96.5% statements / 100% branch / 100% funcs / 95.8% lines** on `SwTokenStorageUtils.ts` (remaining lines are defensive outer catches).

#### Playwright (UI) tests

- Not added: the scenario requires blocking service worker registration for the app under test; the existing Playwright infra explicitly skips SW waits on insecure origins (#31897), and a dedicated `serviceWorkers: 'block'` context fixture felt out of scope for this fix. The demo recordings below were captured with exactly that Playwright option (`browser.newContext({ serviceWorkers: 'block' })`), so a follow-up E2E is straightforward if desired.

#### Manual testing performed

1. Started OpenMetadata **2.0.0** server via the release `docker-compose-postgres.yml` (basic auth, `enforcePrincipalDomain: true`, no SMTP, initial admin `admin@example.com`).
2. **Before** (stock 2.0.0 UI, service workers blocked): login POST returns 200, `GET /users/loggedInUser` → 401 `Token not present`, UI bounces back to Sign In.
3. **After** (this branch served by `vite` against the same 2.0.0 backend, service workers blocked): same credentials land on the "Welcome, admin!" home page; the new `console.error` diagnostic appears once.
4. Reloaded the tab after login: session is restored via the `OM_SESSION` cookie + `/auth/refresh` into the in-memory store — no re-login needed.

#

Both recordings use Playwright `serviceWorkers: 'block'` (deterministic reproduction of the broken-SW condition) against the same OpenMetadata 2.0.0 backend.

**Before — stock 2.0.0: valid credentials, login bounces back to Sign In:**

before: login bounces back to Sign In

**After — this fix: same credentials and conditions, login lands on the home page:**

after: login succeeds

#
### Checklist:

- [x] For JSON Schema changes: not applicable (no schema changes).
- [x] For UI changes: I attached a screen recording and/or screenshots above.
- [x] I have added a test that covers the exact scenario we are fixing (issue number referenced in the test suite).

The PR adds an in-memory authentication-token fallback when service-worker storage becomes unreachable and directly deletes IndexedDB state when worker-assisted logout fails. The logout fallback remains incomplete because it does not invalidate the worker's cached token, while the previously reported refresh-versus-logout write race also remains outstanding.

- Caches the broken-service-worker verdict to avoid repeated controller timeouts.
- Keeps fallback tokens in module memory rather than persistent localStorage.
- Adds direct IndexedDB cleanup and unit coverage for worker failure and logout paths.
</details>

<details open><summary><h3>Confidence Score: 1/5</h3></summary>

The PR is not safe to merge because logout can still leave restorable authentication state in the service-worker cache or recreate it through an overlapping refresh write.

Direct IndexedDB deletion bypasses the worker operation that clears its in-memory token, and logout still lacks ordering against concurrent service-worker writes, so a later reload can restore authentication after the user logged out.

**Files Needing Attention:** openmetadata-ui/src/main/resources/ui/src/utils/SwTokenStorageUtils.ts
</details>

The direct IndexedDB fallback does not clear the service worker's in-memory `app_state`. If the worker later becomes reachable, it can return the retained token after logout and restore authentication on reload. **How this was verified:** The fallback bypasses the worker remove handler, while the worker get handler returns its cached `swStore` value before consulting IndexedDB.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-ui/src/main/resources/ui/src/utils/SwTokenStorageUtils.ts | Adds memory fallback and direct logout cleanup, but the direct path leaves the worker's cached token intact and does not resolve refresh-write ordering. |
| openmetadata-ui/src/main/resources/ui/src/utils/SwTokenStorageUtils.test.ts | Adds broad fallback and logout tests, but does not model a live worker cache or an overlapping refresh write. |

</details>

```mermaid
sequenceDiagram
  participant UI
  participant SW as Service Worker
  participant IDB as IndexedDB
  UI->>SW: removeItem(app_state)
  SW--xUI: removal fails
  UI->>IDB: delete app_state directly
  Note over SW: swStore still retains app_state
  UI->>SW: getItem(app_state) after reload
  SW-->>UI: cached token from swStore
  UI->>UI: restore authenticated state
```
</details>

<sub>Reviews (6): Last reviewed commit: ["Address review: fail-fast logout when th..."](https://github.com/open-metadata/openmetadata/commit/0eca8502f5c9c837cad214493ee74bb6a0a2acf2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56736715)</sub>


neochaotic:
Addressed the review finding from gitar-bot and greptile in fca45cbf86: `clearAppState` now always attempts `swTokenStorage.removeItem` when the SW API is present, independent of the broken-SW verdict, so tokens persisted before a transient failure cannot survive logout and be restored after a reload. Added a regression test for exactly this scenario (`should still attempt to clear persisted tokens after the service worker was marked broken`) — written first against the previous code (red), passing now (33/33, 96.5% stmts / 100% branch coverage on the changed file).

neochaotic:
Follow-up finding from greptile addressed in 10f40e3a86: when `removeItem` fails at logout, the page now deletes `app_state` directly from the worker's IndexedDB (`AppDataStore`/`keyValueStore`) as a secondary deletion path, so stale tokens cannot be read back by a recovered worker after a reload. Covered by three new tests (fallback deletion, store-absent, open-failure) — 36/36 passing, 96.3% stmts / 100% branch coverage on the changed file.

neochaotic:
Round 3 addressed in 0eca8502f5:

- **Performance (gitar)**: with the broken-SW verdict already set, logout now skips the doomed `removeItem` (no 15 s controller wait) and goes straight to the direct IndexedDB deletion.
- **Quality (gitar)**: `deleteFromIndexedDb` now feature-detects `indexedDB.databases()` and returns early when `AppDataStore` was never created, so the fallback no longer creates an empty phantom database (where `databases()` is unsupported the phantom is harmless, as documented in the comment).
- **greptile line-156 finding**: appears to predate 10f40e3a86 — a secondary deletion path does exist now, and with this commit it is used directly when the verdict is broken.

All written test-first; 37/37 passing, 95.7% stmts / 100% branch coverage on the changed file.

neochaotic:
On greptile's concurrent-write finding: acknowledged — the page-side deletion is not ordered against a worker-dispatched write that was already in flight at logout. Two scoping notes:

1. The window requires a worker that is unhealthy enough for the page to mark it broken (controller/message timeout) yet healthy enough to commit a pending `put` afterwards — a narrow slice of the failure space this PR targets (typically the worker is simply unreachable: blocked script, CSP, no registration — in which case no SW writes exist at all, and tokens only ever lived in memory).
2. A durable guarantee needs cooperation from the worker itself (e.g. a logout tombstone in `AppDataStore` that `app-worker.js` checks before any `put`, or the worker draining/aborting pending writes on a logout message). That belongs in the worker code, which this PR deliberately does not touch.

Happy to file a follow-up issue for the tombstone approach if maintainers agree it's worth closing — or extend this PR if you prefer it handled here.

neochaotic:
The new "worker cache survives logout" finding shares the same root cause as the concurrent-write one: both residual windows require a worker that is unreachable at logout yet recovers later, and both can only be closed by the worker itself — the page cannot clear `swStore` in a worker it cannot message (if messaging worked, `removeItem` would have succeeded and cleared both the cache and IndexedDB).

Concretely, the worker-side follow-up would cover: (a) a logout tombstone in `AppDataStore` checked before any `put`, (b) invalidating/rebuilding the in-memory `swStore` from IndexedDB on activation instead of trusting the cached copy, and (c) draining pending writes on a logout message. Same offer as above stands — I can file that follow-up issue (or fold it into this PR if maintainers prefer); this PR intentionally keeps to the page-side storage layer, which is where the login-lockout regression lives.

chirag-madlani (approved):

PR #32071, opened 2026-08-25
Fixes #30362: Allow custom dimension for test cases

### LINKED ISSUE #30362: Allow custom dimension for test cases

Current Situation
Custom SQL Query Test Cases do not use standard dimensions for categorisation and results analysis.

Problem
In the context of the Basel Committee on Banking Supervision and the BCBS 239 standard, the Data Quality Control dimensions represent the criteria used to ensure the quality, integrity, and reliability of the risk data used by financial institutions.

Currently, the controls intended to be migrated to OMD are categorized according to the following data quality dimensions: Accuracy, Availability, Completeness, Consistency, Timeliness, Traceability, Uniqueness, and Validity.

**Desired Outcome**
Users should be able to create their own custom dimensions. Default dimensions should be linked to a test definition and the test case should use it as the default though a user should be able to override it and change the dimension of the test case when creating the test case

_No response_

### Additional context

_No response_

### Pre-submission checklist

TeddyCr:
### Describe your changes:

Fixes #30362

A data quality dimension was a fixed enum of eight values baked into the test definition schema. This makes it a **first-class entity**, so teams can register their own — a BCBS 239, a Timeliness — classify test definitions and individual test cases under them, and have those dimensions drive the Data Quality dashboards and filters like the shipped ones do.

#
### Type of change:
- [x] New feature
- [x] Improvement

#
### High-level design:

**A dimension is an entity.** New `dataQualityDimension` entity + `/v1/dataQuality/dimensions` resource, with the eight shipped dimensions seeded as `provider: system` (immutable, undeletable). Users create their own in **Settings → Preferences → Data Quality**.

**A test case stores its dimension as a relationship, not a lookup.** This is the central design decision. Dimensions are read far more often than written — every list, every dashboard aggregation, every search document — so the value is materialised onto the test case as a `relatedTo` relationship rather than resolved from the test definition on each read. Reads stay one batched query; nothing walks through to the definition.

That choice has a consequence: **a change to a test definition has to be pushed down**, it can't be picked up later. So a relationship written from the test definition carries an `{"inherited": true}` marker on the `entity_relationship.json` column, and `TestDefinitionRepository` handles all three transitions when a definition is reclassified:

| transition | what happens |
|---|---|
| `A → B` | inherited rows repointed to B |
| `A → null` | inherited rows removed |
| `null → B` | inherited rows created |

Each is a single join-scoped SQL statement — scoped through the `testDefinition → testCase` edge rather than an `IN (…)` list — so a definition backing a hundred thousand test cases costs one statement and no heap. **A dimension set on the test case itself carries no marker and is never touched**, which is what makes a per-test-case override survive its definition being reclassified.

Deleting a dimension is repaired the same way, asynchronously (`AsyncService`): the test definitions naming it are cleared, and the affected test cases are re-indexed a page at a time.

**Alternatives rejected.** Resolving the dimension at read time (no materialisation) was tried and reverted — it made reads pay per entity and left `TestCaseIndex`, `useTestCaseResultTab` and the settings counts each carrying their own fallback. Counting inherited test cases on the fly instead of materialising them was rejected for the same reason.

**Migration.** `TestDefinition.dataQualityDimension` changes from the `DataQualityDimensions` enum to an `entityName` string — the enum is kept in the schema for backwards compatibility but nothing `$ref`s it, so it is no longer generated. The 2.1.0 post-migration script backfills a relationship for every pre-existing test case from its definition's dimension; it is idempotent and marks every row inherited. Without it the API returns no dimension for legacy test cases and the settings page counts them as zero.

**Downstream:** the generated enum disappearing breaks Collate's `SlackEntityDataResolver.value()` call — fixed in [openmetadata-collate#6633](https://github.com/open-metadata/openmetadata-collate/pull/6633), which can only go green once this merges.

#
#### Use cases covered
- Admin registers a custom dimension, edits its name/colour/description, deletes it
- System dimensions cannot be edited or deleted
- A test case created without a dimension inherits its test definition's; one created with a dimension keeps it
- Reclassifying a test definition moves its inherited test cases and leaves explicit overrides alone
- Clearing a test definition's dimension drops the inherited ones; setting it again recreates them
- Deleting a dimension clears it from the test definitions that named it and from its test cases
- Custom dimensions are selectable in the Data Quality page's Dimension filter and appear in the Status-by-Dimension widget
- Upgrading an existing deployment leaves every pre-existing test case with the dimension its definition carried

#### Unit tests
- `TestCaseRepositoryTest`, `TestCaseIndexTest`, `SearchListFilterTest`, `DataQualityDimensionSqlMigrationParityTest`, `EntityLinkGrammarTest`
- UI: `DataQualitySettingsPage.test.tsx` (new — the page had none), `useTestCaseFilterOptions.test.tsx`, `DataQualityUtils.test.ts`, `TestCaseResultTab.test.tsx`, `TestDefinitionList.test.tsx`
- 1,270 tests pass across the `DataQuality` / `TestLibrary` UI suites

- [x] Added `openmetadata-integration-tests/.../DataQualityDimensionPropagationIT.java` — **13 tests**, the full transition matrix plus the non-happy paths: unknown dimension rejected, system dimensions immutable and undeletable, duplicate names `409`, deletion isolated to the affected definition, no duplicate relationships across repeated reclassification, and a regression test for `PUT /v1/dataQuality/testDefinitions` returning `404`.

- [x] Not applicable (no ingestion changes) — but note this PR **fixes** the Python ingestion suites, which were failing because `create_or_update` on a test definition returned `404`. `EntityMapper.copy` assigns a fresh UUID before the repository decides create-vs-update, so looking the existing entity up by id alone threw; it now falls back to a name lookup.

#### Playwright (UI) tests
- [ ] Not added. The settings page is new and previously had no coverage at all; the jest suite added here covers create/edit/validation/delete-confirm. Worth a follow-up.

#### Manual testing performed
Full manual pass by @TeddyCr on **both engines**:

1. **MySQL, at scale** — loaded a synthetic catalog (200 tables / 5,000 test cases / 100,000 results) and ran the upgrade path over it
2. **Postgres** — same migration path
3. Ran the **migration** on an existing database rather than a fresh install, and verified the backfill against a live local `psql` session
4. Executed **test cases live** against the local Postgres database

Also verified during development, against the running stacks:
- backfill inserts exactly the pre-existing test cases and re-runs as a no-op (`5000` then `0`)
- the repoint/remove/insert statements move inherited rows only, leaving an explicit override in place — checked on MySQL inside a rolled-back transaction and on a scratch Postgres 15 seeded to the same shape, because the two dialects need different DML (`UPDATE … JOIN` vs `UPDATE … FROM`, `DELETE … JOIN` vs `DELETE … USING`)

#

https://github.com/user-attachments/assets/e9346e01-961d-4dbf-8d52-154ebc09f029

https://github.com/user-attachments/assets/776d165e-c2e9-412c-968e-f00362d11b77

> Recordings predate the settings-page rebuild described below; the flows are unchanged.

#
### Notes for reviewers

**The settings page is built on `@openmetadata/ui-core-components`, not Ant Design.** It was originally written with antd, which `CLAUDE.md` rules out for new work and which the Antd + Less deprecation guard fails the build on. Nothing needed antd — the two things the page reuses (the shared `Table`, the form drawer atom) aren't antd-coupled — so the page, its form and its delete modal were rebuilt on core-components, with the form moving from antd's `Form` to react-hook-form.

**Known limitation, deliberate:** `test_case_result` documents already indexed keep the dimension they were recorded under. Changing a test case's dimension does not rewrite history.

**One fix is reasoned, not test-proven.** The drawer's `onClose` fires twice — once from `closeDrawer`, once when the overlay finishes its exit transition — and the late call was wiping a form the open path had just seeded. The reset is removed, but jsdom doesn't run exit transitions so the added tests pass with and without it; that one wants a browser.


openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TestCaseResultIndex.java
```diff
@@ -92,7 +92,13 @@ public Map<String, Object> buildSearchIndexDocInternal(Map<String, Object> esDoc
     esDoc.put("testCase", testCaseMap);
     esDoc.put("@timestamp", testCaseResult.getTimestamp());
     if (testDefinition != null) {
-      esDoc.put("testDefinition", JsonUtils.getMap(testDefinition));
+      Map<String, Object> testDefinitionMap = JsonUtils.getMap(testDefinition);
+      // Dimension filters and aggregations read the denormalized test definition, so the dimension
+      // set on the test case — a custom one or an override of the definition default — wins here.
+      if (!nullOrEmpty(testCase.getDataQualityDimension())) {
+        testDefinitionMap.put("dataQualityDimension", testCase.getDataQualityDimension());
+      }
```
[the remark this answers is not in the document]
TeddyCr: For this iteration we will assume changing a dimension does not rewrite the historical values of the test case results

TeddyCr (commented):
/olivaw fix the unresolved Greptile comment. The design is not quite what I have envisioned. Dimensions should be their own entities that a testDefinition or a testCase contains as a relationship. Test Definitions ship with the system dimensions (the ones that exist today) and admin or users with the permission can create new ones and edit the custom ones -- system ones should not be editable or deletable.

Dimension creation can live in Settings / Preferences / Data Quality. 

On a test case instead of a free form we have a dropdown that allows us to select which dimension we want to associate with the test case. It should list the existing ones. 

Below are screenshots of hifi designs

Here are the screenshots of the setting page

image
Screenshot 2026-08-25 at 4 20 50 PM
image
image

TeddyCr (commented):
/olivaw fix the unresolved Greptile comment.

The design is also not quite what I have envisioned. Dimensions should be their own entities that a testDefinition or a testCase contains as a relationship. Test Definitions ship with the system dimensions (the ones that exist today) and admin or users with the permission can create new ones and edit the custom ones -- system ones should not be editable or deletable.

Dimension creation can live in Settings / Preferences / Data Quality. 

On a test case instead of a free form we have a dropdown that allows us to select which dimension we want to associate with the test case. It should list the existing ones. 

Below are screenshots of hifi designs

Here are the screenshots of the setting page

image
Screenshot 2026-08-25 at 4 20 50 PM
image
image

TeddyCr (commented):
/olivaw fix the unresolved Greptile comment.

The design is also not quite what I have envisioned. Dimensions should be their own entities that a testDefinition or a testCase contains as a relationship. Test Definitions ship with the system dimensions (the ones that exist today) and admin or users with the permission can create new ones and edit the custom ones -- system ones should not be editable or deletable.

Dimension creation can live in Settings / Preferences / Data Quality. 

On a test case instead of a free form we have a dropdown that allows us to select which dimension we want to associate with the test case. It should list the existing ones. 

Below are screenshots of hifi designs

Here are the screenshots of the setting page

image
Screenshot 2026-08-25 at 4 20 50 PM
image
image

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/DataQualityDimensionRepository.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  /**
+   * Number of test cases that carry this dimension, shown in the Data Quality settings page. Test
+   * cases hold their dimension as a relationship only, so deleting a dimension does not delete or
+   * rewrite them: they simply lose the override and fall back to the dimension of their test
+   * definition.
+   */
+  public int getTestCaseCount(UUID dimensionId) {
```
TeddyCr: **[High] Deleting a dimension breaks every test definition classified with it — and the confirmation dialog reports zero impact.**

Test *cases* reference a dimension by relationship. Test *definitions* reference it by a plain name string in their JSON. `getTestCaseCount()` only counts the relationship side, so `/testCaseCounts` — and therefore the delete confirmation modal — is blind to definitions.

After deleting a dimension a definition points at:
- `TestDefinitionRepository.validateDataQualityDimension` 404s on **any** update to that definition, including just toggling `enabled`;
- `TestCaseRepository.setDataQualityDimension` 404s when creating **any** new test case from it (`EntityUtil.getEntityReference(name)` -> `Entity.getEntityReference(..., NON_DELETED)` throws).

This is reachable straight from the flow this PR adds — reclassify a system test definition under a custom dimension (the newly-allowed edit), then delete the dimension. The modal says "0 test cases affected" and that test type becomes uncreatable.

```bash
om -X PUT  /v1/dataQuality/testDefinitions -d '{"name":"columnValuesToBeUnique",...,"dataQualityDimension":"BCBS 239"}'
om      GET /v1/dataQuality/dimensions/testCaseCounts   # -> 0 for BCBS 239
om -X DELETE /v1/dataQuality/dimensions/$ID?hardDelete=true
om -X POST /v1/dataQuality/testCases -d '{"testDefinition":"columnValuesToBeUnique",...}'   # 404
```

Suggested fix: count referencing test definitions in `preDelete` (block or warn), or make an unresolvable dimension degrade to "no dimension" instead of throwing. A dimension is a label — deleting one should not be able to make a test type uncreatable.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -1012,9 +1057,42 @@ private void validateDimensionColumns(TestCase test, Table table) {
     }
   }
 
+  /**
+   * A test case points at a dimension entity — a system one or a custom one created in Settings >
+   * Preferences > Data Quality. When none is given, the test case inherits the dimension of its
+   * test definition, so that clearing the field resets it to that default.
+   */
+  static void setDataQualityDimension(TestCase test, TestDefinition testDefinition) {
```
TeddyCr, line 1112: **[High] Soft-deleting a dimension makes every test case that uses it un-updatable (404).**

`setFields` resolves the dimension with `Include.ALL` (`getFromEntityRef` -> `Entity.getEntityReferenceById(..., ALL)`), so a soft-deleted dimension comes back on the entity. `setDataQualityDimension` in `prepare` then re-resolves the same reference with `NON_DELETED` and throws `EntityNotFoundException`.

Nothing in the codebase ever writes `entity_relationship.deleted = true`, so the relationship survives the soft delete and there is no path out of the state.

```bash
om -X DELETE /v1/dataQuality/dimensions/$ID          # hardDelete defaults to false
om -X PATCH  /v1/dataQuality/testCases/$TC_ID ...    # 404, forever
```

The Settings page happens to hard-delete (`dataQualityDimensionAPI.ts` passes `hardDelete: true`), so the UI dodges it — but the endpoint documents `hardDelete=false` as the default and a `/restore` endpoint ships alongside, so soft delete is a supported path. It also contradicts the documented behaviour ("test cases ... fall back to the dimension of their test definition") in the repository JavaDoc, the `@Operation` description and the UI modal.

Fix: resolve with `ALL` in `setDataQualityDimension`, or null out an already-deleted reference before re-resolving.

openmetadata-service/src/main/java/org/openmetadata/service/search/indexes/TestCaseResultIndex.java
```diff
@@ -92,7 +92,16 @@ public Map<String, Object> buildSearchIndexDocInternal(Map<String, Object> esDoc
...
     if (testDefinition != null) {
-      esDoc.put("testDefinition", JsonUtils.getMap(testDefinition));
+      Map<String, Object> testDefinitionMap = JsonUtils.getMap(testDefinition);
+      // Dimension filters and aggregations read the denormalized test definition, so the dimension
+      // set on the test case — a custom one or an override of the definition default — wins here.
+      // The dimension is resolved when the result document is indexed and changing it on the test
+      // case does not rewrite the documents of results already indexed: historical results keep the
+      // dimension they were recorded under. That is intentional for this iteration.
+      if (testCase.getDataQualityDimension() != null) {
+        testDefinitionMap.put("dataQualityDimension", testCase.getDataQualityDimension().getName());
```
TeddyCr, line 103: **[Medium] `NoDimension` is normalised in `TestCaseIndex` but not here.**

`TestCaseIndex` maps `NoDimension` to `null` so the "No Dimension" filter (`must_not exists`) works. This index leaves the literal `"NoDimension"` sitting in the denormalized `testDefinition` map when the test case has no override and the definition's dimension is `NoDimension`.

Result: the same `dataQualityDimension=NoDimension` filter returns different sets depending on whether you hit `/testCases/search/list` or `/testCases/testCaseResults/search/list`.

```bash
om "/v1/dataQuality/testCases/search/list?dataQualityDimension=NoDimension&limit=0"
om "/v1/dataQuality/testCases/testCaseResults/search/list?dataQualityDimension=NoDimension&limit=0"
```

Worth noting for the upgrade notes either way: the `TestCaseIndex` change flips the `NoDimension` bucket from always-empty (the literal string was indexed, so `must_not exists` matched nothing) to populated after a reindex. That is a fix, but it will look like a regression to anyone diffing dimension histograms across the upgrade.

openmetadata-ui/src/main/resources/ui/src/constants/profiler.constant.ts
```diff
@@ -15,6 +15,7 @@ import { map, startCase, values } from 'lodash';
 import { DateFilterType } from 'Models';
 import { TestCaseSearchParams } from '../components/DataQuality/DataQuality.interface';
 import { SORT_ORDER } from '../enums/common.enum';
+import { DataQualityDimensions } from '../enums/DataQuality.enum';
```
TeddyCr, line 18: **[High] Custom dimensions cannot be selected in the Data Quality page's Dimension filter.**

`TEST_CASE_DIMENSIONS_OPTION` is built from `values(DataQualityDimensions)` — the hand-written 8-value enum in `enums/DataQuality.enum.ts`. It feeds both `useTestCaseFilters.ts` (`STATIC_OPTIONS[TEST_CASE_FILTERS.dimension]`) and the Dimension `<Select>` in `TestCases.component.tsx`.

The test case form and the test definition form were both updated to fetch `/v1/dataQuality/dimensions`. This dropdown was not. So a user can create `BCBS 239`, assign test cases to it, watch it appear as a row in the Status-by-Dimension widget (`transformToTestCaseStatusByDimension` adds unknown keys dynamically) and confirm the API filters on it — and then find it unselectable in the one place they would go to filter by it.

That is the feature's primary workflow ending in a dead end. Same fetch the two forms already do would fix it.

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchListFilter.java
```diff
@@ -400,6 +401,15 @@ private String escapeDoubleQuotes(String str) {
     return str.replace("\"", "\\\"");
   }
 
+  /**
+   * Escapes a value for safe interpolation into a JSON string literal. Unlike {@link
+   * #escapeDoubleQuotes} this also escapes backslashes and control characters, which free-form
+   * user-supplied values may contain and which would otherwise produce malformed filter JSON.
+   */
+  private String escapeJsonString(String str) {
```
TeddyCr, line 406: **[Low] The new full-JSON escaper is applied to one value only.**

`escapeJsonString` is used for `getDataQualityDimensionCondition`, with the (correct) reasoning that the value is now free-form user input rather than an enum. The sibling conditions in this class still use `escapeDoubleQuotes`, which does not escape backslashes or control characters.

Either the remaining values are provably enum-bounded — worth a one-line comment saying so — or they carry the same latent malformed-filter-JSON bug that was just fixed here.

bootstrap/sql/migrations/native/2.1.0/mysql/schemaChanges.sql
```diff
@@ -261,6 +261,32 @@ SET json = JSON_SET(json, '$.connection.config.scheme', 'oracle+oracledb')
...
+-- without touching the tests that use it. System dimensions are seeded from
+-- json/data/dataQualityDimension on startup.
+-- An earlier revision of this (unreleased) migration declared `id` as a plain column. Because
+-- EntityDAO.insert only writes fqnHash and json, MySQL rejected every insert with "Field 'id'
+-- doesn't have a default value". The table is dropped unconditionally rather than patched: it is
+-- new in this unreleased version, so any existing copy is either empty (the broken shape could not
+-- be inserted into) or holds nothing but the system dimensions, which are re-seeded from
+-- json/data/dataQualityDimension on the next startup.
+DROP TABLE IF EXISTS data_quality_dimension;
+CREATE TABLE data_quality_dimension (
```
TeddyCr, line 275: **[Low] `updatedAt` nullability diverges between dialects.**

MySQL: `updatedAt bigint unsigned GENERATED ALWAYS AS (...) STORED NOT NULL`
Postgres: `updatedat bigint GENERATED ALWAYS AS (...) STORED`  (nullable)

`DataQualityDimensionSqlMigrationParityTest` only asserts that the table and the generated `id` column exist, so it cannot catch this. Either align the two or widen the parity test to compare the full column set — that is what the test is there for.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TestCaseRepository.java
```diff
@@ -175,6 +175,11 @@ public void setFields(TestCase test, Fields fields, RelationIncludes relationInc
         fields.contains(INCIDENT_STATUS_FIELD)
             ? getIncidentStatus(test)
             : test.getIncidentStatus());
+    // Always resolved rather than gated on `fields`: the dimension is not stored on the test case
```
TeddyCr, line 179: **[Medium] `setFields` fetches the dimension unconditionally, so `fields=` no longer bounds the cost of a test case read.**

The list path is fine — `fetchAndSetDataQualityDimensions` uses the correct `findFromBatch(toIds, relation, fromEntityType, toEntityType)` overload and issues one chunked query per page. The exposure is the per-entity hydration fallbacks (`populateRelationFields` on bulk delete, the `serializeJsons` catch, bulk-restore hydration), which go from 0 queries to N.

Adding `dataQualityDimension` to the DQ endpoints' default `fields` would give the same UX without breaking the contract that `fields` bounds the query cost.

TeddyCr (commented):
Review of the diff (not a test run — manual testing is happening separately against this branch).

Seven findings, three of which I would treat as blockers. Full detail is in the per-file comments; summary:

| | Finding | Severity |
|---|---|---|
| 1 | Deleting a dimension breaks every **test definition** classified with it, while the confirmation dialog reports zero impact | High |
| 2 | Soft-deleting a dimension makes its test cases un-updatable (404 on every write) | High |
| 3 | Custom dimensions are unselectable in the Data Quality page's Dimension filter | High |
| 4 | `NoDimension` normalised in `TestCaseIndex` but not `TestCaseResultIndex` | Medium |
| 5 | `setFields` fetches the dimension unconditionally, so `fields=` no longer bounds read cost | Medium |
| 6 | `updatedAt` nullability differs between the MySQL and Postgres migrations | Low |
| 7 | The new JSON escaper is applied to one filter value only | Low |

Findings 1 and 2 share a root cause worth calling out on its own: **an unresolvable dimension throws instead of degrading.** Both `TestCaseRepository.setDataQualityDimension` and `TestDefinitionRepository.validateDataQualityDimension` re-resolve with `NON_DELETED` and let `EntityNotFoundException` escape, so a dimension that is deleted, soft-deleted or simply missing turns into a hard 404 on writes to unrelated entities. Given the schema documents the fallback behaviour ("falls back to the dimension of their test definition"), degrading to that fallback rather than throwing would close both.

Finding 3 is the one that decides whether the feature does what it says: custom dimensions reach the forms, the API and the dashboard widget, but not the filter.

### Verified as fine, for the record

- Relationship direction is consistent across `storeRelationships`, `updateFromRelationship`, `deleteToMany` and the batch fetch; the batch fetch uses the correct four-arg `findFromBatch` overload and issues one chunked query per page.
- `dataQualityDimension` is `keyword` in both index mappings, so multi-word custom names work with the `term` filter.
- Renames are a non-issue: `renameAllowed` defaults to `false`, so the server silently ignores them. (Minor: the create schema says the name "cannot be changed after creation", but PATCH returns 200 and quietly drops the change rather than rejecting it.)
- Hard delete cleans up relationship rows, so the documented fallback does hold on that path.
- Ingestion's `forceUpdate` path uses `patch_test_case_definition`, not PUT, so it will not clobber a manually-set dimension.

*Review by Claude Code at the author's request.*

openmetadata-service/src/main/java/org/openmetadata/service/resources/dqtests/TestDefinitionResource.java
```diff
@@ -61,7 +61,9 @@
             + "that run against data to capture data quality.")
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
-@Collection(name = "TestDefinitions")
+// Seeded after the data quality dimensions (order 0) the shipped test definitions reference:
+// a test definition is rejected when its dimension does not exist yet.
+@Collection(name = "TestDefinitions", order = 1)
```
TeddyCr, line 66: **[BLOCKER] The shipped test definitions do not get seeded at all — Data Quality is non-functional on a fresh install of this branch.**

Verified on a running stack built from this branch (revision `d07e0b582bc`):

```
GET /v1/dataQuality/testDefinitions?limit=0   ->  paging.total = 0
GET /v1/dataQuality/dimensions?limit=100      ->  7 (all seeded fine)
```

Server log:

```
23:47:54,090  Registering ...dqtests.TestDefinitionResource with order 1
23:48:04,123  Created dataQualityDimension:...:Consistency
   ...
23:48:04,134  Registering ...dqtests.DataQualityDimensionResource with order 9

Caused by: EntityNotFoundException: dataQualityDimension instance for Validity not found
    at TestDefinitionRepository.validateDataQualityDimension(TestDefinitionRepository.java:85)
```

The comment added to `TestDefinitionResource` says:

```java
// Seeded after the data quality dimensions (order 0) the shipped test definitions reference:
// a test definition is rejected when its dimension does not exist yet.
@Collection(name = "TestDefinitions", order = 1)
```

but `DataQualityDimensionResource` is declared `@Collection(name = "DataQualityDimensions")` with **no** `order`, and `Collection.order()` defaults to **9** (`Collection.java:27`, "Order of initialization of resource starting from 0"). So the ordering is the exact inverse of the intent: test definitions initialize at order 1, ten seconds before the dimensions exist, every one fails `validateDataQualityDimension`, and the table is left empty.

Consequence: no test definitions -> no test cases can be created at all -> the entire Data Quality feature is dead on a fresh install, not just custom dimensions. It is silent apart from the `WARN` lines, because `initSeedDataFromResources` catches per-entity and continues.

One-line fix:

```java
@Collection(name = "DataQualityDimensions", order = 0)
```

`policies` is the only other collection at order 0, so this slot is free.

Worth adding a regression test that asserts a non-zero test definition count after startup — this class of failure is invisible to every existing test because they create their own test definitions rather than relying on the seeded ones.

harshach (dismissed):

openmetadata-ui/src/main/resources/ui/src/locale/languages/ja-jp.json
```diff
@@ -3677,6 +3680,14 @@
     "description-coverage-widget-description": "サービス内のすべてのデータアセットに対する説明のカバレッジ。<0>詳細を見る。</0>",
     "destination-owner-selection-warning": "\"オーナー\" は \"{{subscriptionType}}\" を設定済みの <0>チーム</0> である必要があります。",
     "destination-selection-warning": "\"{{subscriptionCategory}}\" において \"{{subscriptionType}}\" がアラート受信設定されていることを確認してください。",
+    "dimension-delete-fallback": "They will fall back to the default dimension on their test definition. No test case is deleted.",
```
ShaileshParmar11, line 3683: **i18n: 15 new keys are verbatim English in all 19 non-English locales.**

`.claude/rules/i18n.md`: *"Shipping English text under a non-English locale key is a reviewable defect… CI only verifies keys are synced, not that placeholders were translated."*

I measured the baseline before raising this — the catalogs are ~93% translated (ja-jp 274/4312 entries English-identical, de-de 361/4970, zh-cn 269/3822), so this isn't following existing practice. Same 15 keys need translating in each of the 19 files.

openmetadata-ui/src/main/resources/ui/src/components/DataQuality/AddDataQualityTest/components/TestCaseFormBody.tsx
```diff
@@ -1192,6 +1259,26 @@ const TestCaseFormBody: FC<TestCaseFormBodyProps> = ({
     },
   };
 
+  const dataQualityDimensionField: FieldProp = {
+    name: 'dataQualityDimension',
+    label: t('label.data-quality-dimension'),
+    type: FieldTypes.SELECT,
```
ShaileshParmar11, line 1265: **`SELECT` won't pick up the dimensions, which load async — this file documents why.**

Options come from the mount-time fetch at line 696. The test-type field ~20 lines below explains the exact failure mode:

> AUTOCOMPLETE (searchable combobox) rather than a plain SELECT: the test definitions load async, and an already-open react-aria Select never refreshes its collection when `items` arrive late — leaving the dropdown permanently empty if opened too early.

Open the dimension dropdown before that fetch resolves and you get only the value seeded synchronously from the test definition, not the registered custom dimensions — which is the feature. Suggest `FieldTypes.AUTOCOMPLETE` for consistency with the field above.

openmetadata-ui/src/main/resources/ui/src/pages/DataQualitySettingsPage/DataQualitySettingsPage.tsx
```diff
@@ -0,0 +1,444 @@
...
+    [t]
+  );
+
+  const fetchDimensions = useCallback(async () => {
+    setIsLoading(true);
+    try {
+      const [{ data }, counts, definitionCounts] = await Promise.all([
+        getDataQualityDimensions({ limit: 1000 }),
+        // A missing count must not hide the dimension list itself.
+        getDataQualityDimensionTestCaseCounts().catch(() => ({})),
```
ShaileshParmar11, line 113: **A failed count renders as "nothing is affected" in the delete dialog.**

Both counts are swallowed here, so `testCaseCounts`/`testDefinitionCounts` stay `{}`, `hasReferences` is `false`, and `DeleteDimensionModal` shows **no impact block at all** — the user is told the dimension is unused when we simply couldn't find out. On a destructive action, unknown shouldn't render as zero. Worth distinguishing the two states (the table already does, via the `--` fallback).

openmetadata-ui/src/main/resources/ui/src/pages/DataQualitySettingsPage/DataQualitySettingsPage.tsx
```diff
@@ -0,0 +1,444 @@
...
+  );
+  // Test definitions reference a dimension by name rather than by relationship, so they are
+  // counted separately — without them the delete confirmation reports no impact for a dimension
+  // a dozen test definitions are classified under.
+  const [testDefinitionCounts, setTestDefinitionCounts] = useState<
+    Record<string, number>
+  >({});
+  const [searchTerm, setSearchTerm] = useState('');
+  // `undefined` closes the drawer, `null` opens it in create mode.
+  const [editing, setEditing] = useState<DataQualityDimension | null>();
```
ShaileshParmar11, line 95: **Tri-state via `null` vs `undefined`.**

This types as `DataQualityDimension | null | undefined`: `undefined` = drawer closed, `null` = creating, object = editing — and the effect below keys off `editing !== undefined`.

It does work; I traced the drawer chain (`useFormDrawerWithHook` → `useFormDrawer` → `useCompositeDrawer` → `useDrawer`) and `openDrawer`/`closeDrawer` are genuinely stable, so there's no reset loop. But the effect both reads and writes `isOpen`, so `hookForm.reset(initialValues)` fires twice per open. `frontend-react.md` asks for a discriminated union for state variants; a separate `mode` field would make the three states explicit.

openmetadata-ui/src/main/resources/ui/src/pages/DataQualitySettingsPage/DimensionForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+                isDisabled={isEditing}
+                isInvalid={Boolean(fieldState.error)}
+                label={t('label.name')}
+              />
+            )}
+            rules={{
+              required: t('label.field-required', { field: t('label.name') }),
+              pattern: {
+                value: /^[\w-]+$/,
+                message: t('message.dimension-name-help'),
```
ShaileshParmar11, line 76: **This error message is byte-identical to the field's hint, so a validation failure shows no visible text change.**

Line 64 renders `hint={fieldState.error?.message ?? t('message.dimension-name-help')}`. Type `foo bar` and submit: same string before and after, only the invalid styling moves. (`required` above is fine — it has its own message.) Needs a distinct key, e.g. "Use only letters, numbers, underscores and hyphens."

openmetadata-ui/src/main/resources/ui/src/pages/DataQualitySettingsPage/DimensionForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+                label={t('label.description')}
+                placeholder={t('message.dimension-description-placeholder')}
+                rows={4}
+                value={field.value ?? ''}
+              />
+            )}
+          />
+
+          <Box direction="col" gap={2}>
+            <Typography as="label" size="text-sm" weight="medium">
```
ShaileshParmar11, line 111: **`as="label"` renders a `<label>` with no associated control.**

No `htmlFor`, no wrapped input — `frontend-a11y.md` lists "a label with no control" among the error-level `jsx-a11y` rules. Since the swatches are a group, `role="group"` + `aria-labelledby` pointing at this node (or `fieldset`/`legend`) would both fix the dangling label and give the group an accessible name.

openmetadata-ui/src/main/resources/ui/src/pages/DataQualitySettingsPage/DimensionForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+              {t('label.color')}
+            </Typography>
+            <Box gap={2} wrap="wrap">
+              {DIMENSION_COLOR_PALETTE.map((swatch) => (
+                <button
+                  aria-label={swatch}
```
ShaileshParmar11, line 117: **`aria-label={swatch}` announces the raw hex** — "button, #175CD3, pressed". Not a usable name; needs human-readable colour names.

Related: this is single-select rendered as 11 independent `aria-pressed` toggles, so there's no arrow-key navigation within the group and Tab walks all 11. `component-library.md`'s reuse table maps `role="radiogroup"` → `RadioButtons`.

To be clear, I checked `components/common/ColorPicker` and hand-rolling here is the **right** call — that one is antd-based and free-form hex, not a curated palette. This is only about the widget's semantics.

openmetadata-ui/src/main/resources/ui/src/pages/DataQualitySettingsPage/DeleteDimensionModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+    return null;
+  }
+
+  const title = t('label.delete-entity', {
+    entity: dimension.displayName ?? dimension.name,
+  });
+  const hasReferences = testCaseCount > 0 || testDefinitionCount > 0;
+
+  return (
+    <ModalOverlay isOpen>
```
ShaileshParmar11, line 56: **`components/common/DeleteModal` already does this, on the same core-components primitives.**

Its props are `{ open, entityTitle, message: ReactNode, isDeleting, onCancel, onDelete }` — `message` takes a node, so the two impact blocks drop straight in and this file mostly disappears.

More than dedup: the shared one sets `isDismissable={!isDeleting}` and disables Cancel while deleting. Here `isOpen` is hardcoded and only the confirm button is disabled, so Esc, backdrop and Cancel all stay live mid-request — cancelling yields a dismissed dialog that still fires a success toast.

openmetadata-ui/src/main/resources/ui/src/pages/DataQualitySettingsPage/data-quality-settings-page.less
```diff
@@ -0,0 +1,73 @@
...
+.dimension-color-swatch {
+  width: 24px;
+  height: 24px;
+  border-radius: 50%;
+  border: 2px solid transparent;
+  padding: 0;
+  cursor: pointer;
+
+  &.selected {
+    box-shadow: 0 0 0 2px @white, 0 0 0 4px currentColor;
```
ShaileshParmar11, line 42: **Ring drawn as `box-shadow`.** `frontend-styling.md`: rings compile to box-shadow, WebKit doesn't pixel-snap them, so this selected-swatch indicator thins or vanishes in Safari at non-100% zoom. `outline` + `-outline-offset` instead.

While in this file — `yarn token-audit` reports **5 errors** on it: `12px` font-size → `var(--om-font-size-xs)`, 4× `12px` + 1× `6px` spacing → `var(--om-space-*)`, 2× `8px` radius → `var(--om-radius-lg)`. Fair warning that the repo baseline is already 199 such errors and token-audit isn't a CI step, so this is new debt rather than a broken gate — but it's a brand-new file, so there's no legacy excuse. The legacy `@white`/`@grey-50`/`@yellow-1` variables also carry no dark-mode variant, so the page chrome won't theme.

openmetadata-ui/src/main/resources/ui/src/utils/TestDefinitionUtils.ts
```diff
@@ -11,11 +11,21 @@
  *  limitations under the License.
  */
 
+import { ProviderType } from '../generated/entity/bot';
```
ShaileshParmar11, line 14: **Wrong module.** `ProviderType` for a test definition is exported from `generated/tests/testDefinition.ts` — already imported two lines below for `TestDefinition` and `TestPlatform`. This compiles only because the generated enums are structurally identical strings; it couples test-definition logic to the bot schema.

ShaileshParmar11 (commented):
Reviewed the UI half of this PR against `.claude/rules/` (i18n, styling, a11y, component-library, react) and the merge base, so everything flagged is new code in this PR rather than pre-existing practice.

**The design holds up.** The settings page really is on the go-forward stack — no antd imports, no `any`, no hex in JSX, real react-hook-form + `Controller`, admin-gated route (`AdminProtectedRoute` + `isProtected: isAdminUser`), i18n keys throughout. `resolvePatchedDimension` extends the `resolvePatched*` pattern already in `DataQualityPureUtils` rather than inventing a parallel one, and the "untouched inherited prefill" handling is subtle and correct.

**One thing I'd call blocking:** the 15 new locale keys are English in all 19 non-English files. Everything else below is a should-fix or a nit.

**Worth knowing what CI isn't catching here**, since it's green: `token-audit` is not a `ui-checkstyle` step; `tw-deprecation-guard` reports new `.less` files without failing (only new `antd` imports block — so the PR description's "the guard fails the build on Less" isn't quite right); and `check-i18n-all` verifies key sync, never translation.

**One open question, not a defect:** once a test case has an explicit dimension there's no way back to inheriting. Re-picking the definition's current default writes a fresh non-inherited row, and clearing the field emits a `remove` that the backend reads as *no dimension* — while the field's own help text says it "defaults to the dimension of the test type". Intended for this iteration, or worth a "reset to inherited" affordance?

openmetadata-ui/src/main/resources/ui/src/pages/DataQualitySettingsPage/DataQualitySettingsPage.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+  provider: ProviderType.User,
+  style: { color: '#6938EF' },
+};
+
+jest.mock('../../rest/dataQualityDimensionAPI', () => ({
+  getDataQualityDimensions: jest.fn(),
+  getDataQualityDimensionTestCaseCounts: jest.fn(),
+  getDataQualityDimensionTestDefinitionCounts: jest.fn(),
+  createDataQualityDimension: jest.fn(),
+  patchDataQualityDimension: jest.fn(),
```
ShaileshParmar11, line 38: **`patchDataQualityDimension`, `createDataQualityDimension` and `deleteDataQualityDimension` are mocked here but never asserted in any test.**

So the `editing ? patch : create` branch, the `compare(editing, updated)` patch construction and the whole delete-confirm flow are uncovered — as is every validation rule, and the count/impact-warning logic (both count mocks return `{}` in all five tests).

The drawer-prefill tests are genuinely good — they mock only the REST boundary, run the real form and drawer, and assert observable values, so they'd catch a regression in the `editing` wiring. It's the create/patch/delete paths that have no coverage here and none in Playwright either.

openmetadata-ui/src/main/resources/ui/playwright/e2e/Features/DataQuality/TestLibrary.spec.ts
```diff
@@ -510,17 +510,18 @@ test.describe(
...
 
-      // Verify edit button does not exist for system test definition
+      // The form opens for a system test definition because its data quality dimension can be
+      // reclassified — every other field stays read-only — but it can never be deleted.
       const editButton = page.getByTestId(
         `edit-test-definition-${systemTestDef.name}`
       );
 
-      await expect(editButton).toBeDisabled();
+      await expect(editButton).toBeEnabled();
```
ShaileshParmar11, line 524: This assertion is strictly weaker than the `toBeDisabled()` it replaces, and the comment above claims something it doesn't check ("every other field stays read-only").

The invariant is covered backend-side by `validateSystemTestDefinitionUpdate` + `DataQualityDimensionPropagationIT`, so nothing is lost overall — but at the UI layer we now only assert the button is enabled. Asserting one other field is still read-only once the form opens would restore the guarantee here. Same in `TestDefinitionPermissions.spec.ts`.

openmetadata-ui/src/main/resources/ui/src/components/DataQuality/AddDataQualityTest/components/TestCaseFormBody.tsx
```diff
@@ -645,6 +663,43 @@ const TestCaseFormBody: FC<TestCaseFormBodyProps> = ({
...
+    dataQualityDimensionValue,
+  ]);
+
+  useEffect(() => {
+    getDataQualityDimensions({ limit: 1000 })
+      .then(({ data }) => setDataQualityDimensions(data))
+      // The field falls back to the test definition's dimension, so a failure here degrades
+      // the picker rather than blocking test case creation.
+      .catch(() => setDataQualityDimensions([]));
+  }, []);
```
ShaileshParmar11, line 701: **This fetch is inconsistent with `fetchTables` ~10 lines below, which is the pattern this file already established.**

Three things, in order of how much they matter:

**1. Promise chain instead of `async`/`await` + a named function.** `frontend-react.md` → *"Data fetching: async + try/catch, update loading states."* `fetchTables` at line 703 is the model — `useCallback(async (searchValue = '') => { … try { … } })`. Inlining the chain in `useEffect` also means the `.catch` silently swallows a real failure with no way to distinguish "no dimensions exist" from "the request failed".

**2. `limit: 1000` is a magic number, and the repo has a constant for exactly it** — `AGGREGATE_PAGE_SIZE_LARGE = 1000` in `constants/constants.ts:61` (next to `PAGE_SIZE_LARGE = 50`, which `fetchTables` uses). It's hardcoded in three places in this PR: here, `DataQualitySettingsPage.tsx:111`, and `useTestCaseFilterOptions.tsx:80`.

The endpoint returns `{ data, paging }` and `paging` is discarded, so past 1000 dimensions this truncates silently. Dimensions are admin-curated so that ceiling is unlikely to bite — but the call also re-runs on **every** mount of the test-case form with no cache, where `fetchTables` maintains a bounded LRU (`TABLES_CACHE_MAX_SIZE`) precisely to avoid that.

**3. No loading state.** The table field passes `isLoading: isTableLoading` to its `ASYNC_SELECT`; the dimension field at line 1265 passes no `isLoading`, so while this request is in flight the dropdown just shows the one value seeded from the test definition with no indication anything is still arriving.

All three fall out of one change: make this an `ASYNC_SELECT` backed by a `fetchDimensions` callback shaped like `fetchTables` — which also resolves the separate `FieldTypes.SELECT` comment I left at line 1265, since `ASYNC_SELECT`/`AUTOCOMPLETE` re-derive their list as options arrive and `SELECT` does not.

Correcting myself: I'd initially set the `limit: 1000` point aside because other files use it as a fetch-everything idiom (`TagsUtils`, `GenericProvider`, `SetActionForm`). That was the wrong call — precedent in other files counts for much less than the correct pattern sitting in this one.

ShaileshParmar11 (commented):
Follow-up on the dimension options fetch — one more inline comment.

ShaileshParmar11 (commented):
Second pass, this time purely on simplification and reuse — no rules, no bugs. One finding with real leverage and a handful of small ones.

## The dimension picker is built four times

`getDataQualityDimensions({ limit: 1000 })` appears in four files, and three of them wrap it in near-identical `useState` + `useEffect` + `.then/.catch` + option-shaping:

| File | Option shape | Error degrades to |
|---|---|---|
| `TestCaseFormBody.tsx:696` | `Map<string, FormSelectItem>` | `[]` |
| `TestDefinitionFormBody.tsx:107` | `Map<string, FormSelectItem>` | `[]` |
| `useTestCaseFilterOptions.tsx:80` | `{ label, name, value }` | `TEST_CASE_DIMENSIONS_OPTION` |
| `DataQualitySettingsPage.tsx:111` | raw entities | `{}` |

The two form builders are the same ~15 lines twice. A single `useDataQualityDimensions()` hook collapses all four, and takes these with it rather than needing four separate fixes:

- the magic `1000` in four places — `AGGREGATE_PAGE_SIZE_LARGE` already exists in `constants/constants.ts:61`
- no loading state in either form picker, while the table `ASYNC_SELECT` next door passes `isLoading`
- a refetch on every form mount with no cache, where `fetchTables` in the same file keeps a bounded LRU
- three different degradations for one failure mode
- one place to move to `ASYNC_SELECT` + server-side search later, instead of three

That last point also closes the separate `FieldTypes.SELECT` thread — `SELECT` never refreshes its collection when options land late, which is exactly what a mount-time fetch produces.

## Also worth a look

Inline threads cover: the `filter().map()` / no-op `String()` / double `getEntityName()` / mismatched catch shape in `fetchDimensionOptions`; `displayName ?? name` hand-rolled in both form bodies where `getEntityName()` is used elsewhere (and behaves differently on an empty `displayName`); and three small ones grouped onto the settings page.

## Not flagged — these got simpler, and I checked

`TestDefinitionTable.component.tsx` reordering the tooltip if/else to check permission first removes a redundant branch. `TestCaseResultTab.utils.ts` swapping a four-term `||` chain for `[...].some(Boolean)` is the right direction. `transformTestDefinitionFormData`'s `patch.filter(op => op.path === DIMENSION_PATH)` exact-match is correct now that `dataQualityDimension` is a flat string.

openmetadata-ui/src/main/resources/ui/src/components/DataQuality/AddDataQualityTest/components/TestCaseFormBody.tsx
```diff
@@ -645,6 +663,43 @@ const TestCaseFormBody: FC<TestCaseFormBodyProps> = ({
...
+    ].forEach((dimension) => {
+      if (dimension && !options.has(dimension)) {
+        options.set(dimension, { id: dimension, label: dimension });
+      }
+    });
+
+    return Array.from(options.values());
+  }, [
+    dataQualityDimensions,
+    selectedTestDefinition,
```
ShaileshParmar11, line 691: **This builder and `TestDefinitionFormBody.tsx:116-130` are the same ~15 lines.** Same `Map<string, FormSelectItem>`, same `options.set(dimension.name, { id, label: dimension.displayName ?? dimension.name })`, same keep-the-current-value-if-it-was-deleted fallback. The only real difference is that this one also seeds `selectedTestDefinition?.dataQualityDimension`.

Together with the fetch above it and the two other call sites, that's the picker built four times — see the review summary for the `useDataQualityDimensions()` suggestion.

**One thing that isn't just duplication:** `dimension.displayName ?? dimension.name` is hand-rolled here, while `useTestCaseFilterOptions.tsx:90` uses the shared `getEntityName()`. They differ — `getEntityName` is `displayName || name || ''` (falsy), this is nullish. A dimension with `displayName: ""` renders **blank** in this picker and correctly in the filter dropdown.

Not reachable today (`DataQualitySettingsPage.handleSave` does `values.displayName || undefined`), but the API accepts an empty `displayName`, so it's a latent inconsistency. `getEntityName()` in all three places.

openmetadata-ui/src/main/resources/ui/src/components/TestLibrary/TestDefinitionForm/TestDefinitionFormBody.tsx
```diff
@@ -94,6 +92,42 @@ const TestDefinitionFormBody: FC<TestDefinitionFormBodyProps> = ({
...
+        label: dimension.displayName ?? dimension.name,
+      });
+    });
+    const currentValue = dataQualityDimensionValue?.id;
+    if (currentValue && !options.has(currentValue)) {
+      options.set(currentValue, { id: currentValue, label: currentValue });
+    }
+
+    return Array.from(options.values());
+  }, [dataQualityDimensions, dataQualityDimensionValue]);
```
ShaileshParmar11, line 130: The twin of `TestCaseFormBody.tsx:669-691` — same `useState` + `useEffect` + `.then/.catch` fetch and the same Map-based option builder, near-verbatim.

Worth noting the two differ in a way that looks unintended rather than deliberate: this one degrades to `[]` on failure, `useTestCaseFilterOptions` degrades to `TEST_CASE_DIMENSIONS_OPTION`, and `DataQualitySettingsPage` swallows into `{}`. Three behaviours for one failure mode.

Same `displayName ?? name` vs `getEntityName()` point as the other thread.

openmetadata-ui/src/main/resources/ui/src/components/DataQuality/TestCases/useTestCaseFilterOptions.tsx
```diff
@@ -59,8 +64,42 @@ export const useTestCaseFilterOptions = () => {
...
+        })),
+        ...data.map((dimension) => ({
+          label: getEntityName(dimension),
+          name: getEntityName(dimension),
+          value: dimension.name,
+        })),
+      ]);
+    } catch {
+      // Degrade to the shipped dimensions rather than to an empty dropdown.
+      setDimensionOptions(TEST_CASE_DIMENSIONS_OPTION);
```
ShaileshParmar11, line 97: Four small things here, all independent of the duplication point:

**1. `filter().map()` to extract one known element.** This scans 8 entries for `NoDimension` — and it doesn't need a search at all: `TEST_CASE_DIMENSION_LABELS` is a `Record<DataQualityDimensions, string>`, so it's a direct key lookup.

**2. `String(label)` is a no-op** — that Record is already typed `string`.

**3. `getEntityName(dimension)` runs twice per row**, once for `label` and once for `name`.

**4. The `catch` shape doesn't match the success shape.** It sets `TEST_CASE_DIMENSIONS_OPTION` raw — `{ label, value }` with no `name`. `FetchedOption.name` is optional so it compiles, but line 31 says it's "kept alongside the rich antd JSX label for renderer reuse", so the degraded path silently drops it.

```ts
const NO_DIMENSION_OPTION: FetchedOption = {
  label: TEST_CASE_DIMENSION_LABELS[DataQualityDimensions.NoDimension],
  name: TEST_CASE_DIMENSION_LABELS[DataQualityDimensions.NoDimension],
  value: DataQualityDimensions.NoDimension,
};

const withName = (o: { label: string; value: string }): FetchedOption => ({
  ...o,
  name: o.label,
});

// success
setDimensionOptions([
  NO_DIMENSION_OPTION,
  ...data.map((dimension) =>
    withName({ label: getEntityName(dimension), value: dimension.name })
  ),
]);

// catch
setDimensionOptions(TEST_CASE_DIMENSIONS_OPTION.map(withName));
```

No array search, no cast, one `getEntityName` per row, both paths the same shape.

(For what it's worth I considered folding this into a `reduce` and it's the wrong tool — this concatenates two 1:1 mappings, which is what `map` + spread is for. `reduce` here would be harder to read, not easier.)

openmetadata-ui/src/main/resources/ui/src/pages/DataQualitySettingsPage/DataQualitySettingsPage.tsx
```diff
@@ -0,0 +1,444 @@
...
+import { showErrorToast, showSuccessToast } from '../../utils/ToastUtils';
+import './data-quality-settings-page.less';
+import DeleteDimensionModal from './DeleteDimensionModal';
+import DimensionForm, { type DimensionFormValues } from './DimensionForm';
+
+const DEFAULT_COLOR = DIMENSION_COLOR_PALETTE[0];
+
+const countFor = (
+  counts: Record<string, number>,
+  dimension?: DataQualityDimension
```
ShaileshParmar11, line 64: Grouping three small ones for this file so they don't each need a thread:

**`countFor` is bypassed by the table column.** It's used at the two `deleting*` call sites but line 264 inlines `testCaseCounts[record.id ?? ''] ?? '--'`. The fallbacks genuinely differ (`0` vs `'--'`), but `record.id ?? ''` is now written three times — a `fallback` parameter covers both.

**`isSaving` (line 82) does double duty.** Set by `handleSave` *and* `handleDelete`, then passed as `submitLoading` to the drawer and `isDeleting` to the delete modal. They're mutually exclusive in practice so it isn't a bug, but a flag named `isSaving` driving a delete spinner reads wrong — `isSubmitting`, or a separate `isDeleting`.

**Three `useState` for one `Promise.all` (lines 69-75).** `dimensions`, `testCaseCounts` and `testDefinitionCounts` are only ever set together from the one call. `frontend-react.md` asks for related state as a single object.

harshach (approved):

shah-harshit (approved):

PR #6030, opened 2022-07-12
Fix #5890 - Add health endpoint

pmbrull:
### Describe your changes :

Fix #5890 

#
### Type of change :

- [x] Bug fix
- [x] Improvement
- [x] New feature
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] Documentation

#
### Frontend Preview (Screenshots) :
<p align="center">For frontend related change, please link screenshots of your changes preview! Optional for backend related changes.
</p>

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/open-source-community/developer) document.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] All new and existing tests passed.

#

PR #32078, opened 2026-08-26
refactor(ui): consolidate destination form

shah-harshit:
## What changed

- Promote the core-UI destination form to the canonical `DestinationFormItem` path.
- Migrate notification, observability, and alert-details flows through a compatibility bridge while their parent pages remain on Ant Form.
- Preserve source-aware destination filtering, email/team/user receivers, webhook auth/headers/query parameters/HTTP method, timeout defaults and validation, test-destination actions, and read-only behavior.
- Remove the duplicate V2 tree, legacy Ant Design destination UI/styles, and obsolete destination helpers only after callers and behavior are covered by the canonical component.
- Move destination constants and pure option/filter helpers into dedicated constants and utility files with focused tests.
- Use the core component selection API consistently, preserve advanced configuration expansion across nested field-array updates, and retain the user-facing `G Chat`/`Ms Teams` labels.
- Keep disabled core ComboBoxes read-only: selected text remains transparent behind the icon/value layer, and pointer interaction cannot open the options list.
- Keep the required-destination validation visible when the compatibility bridge blocks a save with an empty destination list.
- Update destination Playwright helpers to interact with the visible combobox, select trigger, accordion, and input elements exposed by the core components.

https://github.com/user-attachments/assets/36e91ffe-6523-4773-a401-26850f155308

## Validation

- Focused Jest: 8 suites and 63 tests passed.
- Destination validation Jest: 3 suites and 20 tests passed, including required and optional empty-destination submissions.
- UI-core Vitest: 2 disabled ComboBox regression tests passed.
- UI-core ESLint, Prettier, TypeScript type-check, and production build passed.
- Focused UI checkstyle: 0 errors; Prettier and import organization applied.
- Focused Playwright passed for notification multiple filters, notification read-only permissions, observability table alert, and observability read-only permissions.
- Manually verified the alert configuration view keeps the destination selector disabled, does not open it on click, and renders the destination icon and label without overlap.
- The notification destination scenario completed the changed UI interactions locally and then reached an unrelated backend-response assertion: the local backend returned `testResult.config` instead of a redacted response.
- `git diff --check` passed.
- `yarn tsc:check` reported no errors in the changed destination or alert files; the branch retains unrelated pre-existing type errors elsewhere in the UI.

Related to #6030. Collate companion: open-metadata/openmetadata-collate#6146.


openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/TeamAndUserSelectItem/TeamAndUserSelectItem.tsx
```diff
@@ -12,262 +12,219 @@
...
-        : selectedOptions.filter((option) => option !== key);
-
-      form.setFieldValue(['destinations', ...fieldName], updatedValues);
+  const handleOptionClick = useCallback(
+    (value: string) => {
+      const isSelected = selectedOptions.includes(value);
+      const updated = isSelected
+        ? selectedOptions.filter((o) => o !== value)
+        : [...selectedOptions, value];
+      setValue(fieldPath, updated);
```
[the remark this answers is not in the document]
shah-harshit: Fixed in 9418176c49. Email and team/user receiver updates now trigger React Hook Form validation immediately, including removal paths. Added regressions that first surface the required error and then verify it clears after a valid receiver is selected.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationConfigField/DestinationConfigField.tsx
```diff
@@ -104,42 +131,166 @@ function EmailTagInput({ fieldName }: { fieldName: number }) {
...
+        <div className="tw:flex tw:items-start tw:gap-2" key={field.id}>
+          <div className="tw:grid tw:flex-1 tw:grid-cols-2 tw:gap-2">
+            <Controller
+              control={control}
+              name={`${fieldPath}.${index}.key`}
+              render={({ field: controllerField, fieldState }) => (
+                <div>
+                  <Input
+                    data-testid={`${testIdPrefix}-key-input-${index}`}
+                    inputDataTestId={`${testIdPrefix}-key-input-field-${index}`}
```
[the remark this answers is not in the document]
shah-harshit: Fixed in 9418176c49. Header and query-parameter row controls are now namespaced by destination index and row index, with a two-destination regression test. The alert Playwright helpers now use the unique input IDs as well.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItemFormBridge.tsx
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }, [getValues, reset, values]);
+
+  useEffect(() => {
+    const subscription = watch(() => {
+      const nextValues = getDestinationFormFields(getValues());
+      const currentValues = getDestinationFormFields(values);
+
+      if (!isEqual(currentValues, nextValues)) {
+        onChange(nextValues);
```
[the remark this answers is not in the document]
shah-harshit: Fixed in the latest commit. The bridge now memoizes the four controlled fields and stores the latest values/onChange callback in refs, so its RHF watch subscription remains stable across unrelated parent renders while still observing current adapter data. Added a regression test that failed with two subscriptions before the fix and now confirms a single subscription.

shah-harshit (commented):
Review focused on behavior parity for the destination-form consolidation.

**Framing correction that affects how to read the diff:** at the merge base, `DestinationFormItemV2` was never wired into any page — only its own tests imported it. All three live callers (AddNotificationPage, ObservabilityAlertFormFields, AlertConfigDetails) rendered the **legacy Ant `DestinationFormItem`**. So the real user-facing baseline is the legacy Ant component, and the new core-UI component is essentially V2's body driven through the new `DestinationFormItemFormBridge`. Parity should be judged against legacy, not V2.

**Verdict:** build is safe — every removed helper (`getConnectionTimeoutField`, `getReadTimeoutField`, `getDestinationConfigField`, old `getTeamOptions`) is dead with no dangling refs; the deleted `TeamAndUserSelectItem.constants.ts` was inlined into its test; no V2 imports remain. Source-aware filtering, internal/external handling, `G Chat`/`Ms Teams` labels, webhook auth/headers/query-params/HTTP-method, secret redaction, timeout defaults, test-destination, and read-only pass-through are all preserved. The two-way Ant↔RHF sync in the bridge is guarded against loops/echo and I found no concrete drop-of-edits bug.

The inline comments below are the handful of **genuine behavior deltas vs the legacy component** (several look like intended fixes, but they are not no-ops, so calling this a 'pure consolidation' is slightly overstated) plus one small regression and one manual-test hotspot. No blockers.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItemFormBridge.tsx
```diff
@@ -0,0 +1,157 @@
...
+    });
+
+    return () => subscription.unsubscribe();
+  }, [getValues, watch]);
+
+  const validate = useCallback(async () => {
+    const coreFormIsValid = await trigger();
+    const isDestinationMissing =
+      isRequired && isEmpty(getValues('destinations'));
+    const minimumDestinationError = t('message.minimum-count-error', {
```
shah-harshit: Behavior delta (visible string): the empty-destination submit error changed from the legacy `message.length-validator-error` ("At least 1 Destination required") to `message.minimum-count-error` ("At least 1 Destination must be selected."). Matches V2 but differs from what users saw before. Intentional? If parity is the goal, keep the old key.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItemFormBridge.tsx
```diff
@@ -0,0 +1,157 @@
...
+        // so its prop echo is not mistaken for an external reset.
+        latestValues.current = cloneDeep(nextValues);
+        latestOnChange.current(nextValues);
+      }
+    });
+
+    return () => subscription.unsubscribe();
+  }, [getValues, watch]);
+
+  const validate = useCallback(async () => {
```
shah-harshit, line 122: `validate()` throws `minimumDestinationError` whenever `!coreFormIsValid`, even when destinations exist but a nested config field (e.g. webhook endpoint) is invalid — so the rejection reason is wrong. In practice it's swallowed (this Form.Item is `hidden` and the real error renders inline in the core component, so save is still blocked correctly), but the thrown message is misleading. Consider only forcing the min-destination message when `isDestinationMissing`, and otherwise rethrowing a generic/validation error.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItemFormBridge.tsx
```diff
@@ -0,0 +1,157 @@
...
+    };
+
+    if (isEqual(latestValues.current, synchronizedValues)) {
+      return;
+    }
+
+    latestValues.current = cloneDeep(synchronizedValues);
+
+    if (!isEqual(getDestinationFormFields(getValues()), synchronizedValues)) {
+      reset(synchronizedValues);
```
shah-harshit, line 103: Two-way-sync hotspot to exercise manually. If the parent's Ant `useWatch` echo of `destinations` is ever not `lodash.isEqual` to what the core form emitted, this `reset()` fires on each keystroke, remounts the `useFieldArray` rows (new RHF ids) and drops input focus mid-typing. Callers currently do verbatim per-key `form.setFieldValue` with no `normalize` on the hidden Form.Item, so deep-equality should hold and I found no concrete trigger — but please type rapidly in a webhook config field and confirm no focus loss / row remount.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationSelectItem.tsx
```diff
@@ -11,443 +11,390 @@
...
+            <Controller
+              control={control}
+              name={`destinations.${id}.destinationType`}
+              render={({ fieldState }) => (
+                <div>
+                  <Select.ComboBox
+                    isRequired
+                    data-testid={`destination-category-select-${id}`}
+                    fontSize="sm"
+                    isDisabled={isViewMode}
```
shah-harshit, line 171: Behavior delta in read-only view: `isDisabled={isViewMode}` (here and at 243/306/330) now disables the category ComboBox, type Select, notify Toggle and downstream Input. The legacy live component used `isViewMode` only to hide the remove button and left inputs editable. Almost certainly the desired behavior for `AlertConfigDetails`, but it's a real change — worth a screenshot of the read-only alert-details view to confirm.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationSelectItem.tsx
```diff
@@ -11,443 +11,390 @@
...
+  );
 
   const destinationStatusDetails = useMemo(() => {
     const { type, category, config } = destinationItem;
-
-    const currentDestination = destinationsWithStatus?.find((destination) =>
+    const current = destinationsWithStatus?.find((d) =>
       isEqual(
-        { type, category, config: omitBy(config, isUndefined) },
+        { type, category, config: normalizeDestinationConfig(config) },
```
shah-harshit, line 88: Behavior delta (likely a fix): the delivery-status match now runs `normalizeDestinationConfig` on BOTH the local item and the server side. Because that helper array-normalizes `headers`/`queryParams`, the status banner will now show for webhook destinations carrying headers/query-params where the legacy code (which used `omitBy(config, isUndefined)` on the local side only) silently failed to match. Good change, but it's a visible difference in which destinations render the status banner.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationSelectItem.tsx
```diff
@@ -11,443 +11,390 @@
...
+                  </div>
+                )}
+                rules={{
+                  required: t('message.field-text-is-required', {
+                    fieldText: t('label.downstream-depth'),
+                  }),
+                  validate: (value) => {
+                    const numVal = Number(value);
+
+                    return !Number.isFinite(numVal) || numVal <= 0
```
shah-harshit, line 363: Minor validation delta on downstream depth: new rule `!Number.isFinite(numVal) || numVal <= 0` rejects empty / non-numeric input, whereas legacy (`!isEmpty(String(value)) && value <= 0`) passed empty. Stricter than before — fine if intended, just noting it's not a no-op.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationConfigField/DestinationConfigField.tsx
```diff
@@ -326,8 +501,68 @@ function DestinationConfigField({
...
+                      name="queryParams"
+                    />
+                  </Grid.Item>
+                  <Grid.Item span={24}>
+                    <Controller
+                      control={control}
+                      defaultValue={HTTPMethod.Post}
+                      name={`destinations.${fieldName}.config.httpMethod`}
+                      render={({ field }) => (
+                        <RadioGroup
```
shah-harshit, line 549: Small regression: the HTTP Method radio group lost its field label. The legacy `getDestinationConfigField` rendered a `label.http-method` ("HTTP Method:") above the POST/PUT radios; this `RadioGroup` has no `label`, so users now see an unlabeled radio pair in Advanced Configuration. Add the label back.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationConfigField/DestinationConfigField.tsx
```diff
@@ -79,21 +104,30 @@ function EmailTagInput({ fieldName }: { fieldName: number }) {
...
       setInputValue('');
+      setEmailError(undefined);
 
       return;
     }
-    setValue(`destinations.${fieldName}.config.receivers`, [
-      ...receivers,
-      trimmed,
-    ]);
+    if (!isValidEmailAddress(trimmed)) {
```
shah-harshit, line 112: Behavior delta: email receivers now enforce `isValidEmailAddress` (EMAIL_REG_EX) format validation. The legacy Ant `Select mode="tags"` accepted any tag. Likely intended, but an input previously accepted is now rejected — confirm this is desired.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItem.component.tsx
```diff
@@ -11,180 +11,294 @@
...
+          </Grid.Item>
+          <Grid.Item span={1}>
+            <Typography as="span" size="text-sm">
+              :
+            </Typography>
+          </Grid.Item>
+          <Grid.Item data-testid="connection-timeout" span={16}>
+            <Controller
+              control={control}
+              defaultValue={10}
```
shah-harshit, line 182: The `timeout` (and `readTimeout` at 216) `defaultValue` lives only in the core RHF form and never fires `watch`, so it is not echoed to the parent Ant form via `onChange` — only user edits propagate. Please confirm the save payload: if the user leaves connection timeout untouched, is `10` still persisted? Legacy Ant `defaultValue` had the same not-persisted-unless-edited caveat, so this may be no worse than before, but worth verifying against the payload builder.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItemFormBridge.tsx
shah-harshit: Fixed in 77851139c7. The bridge now uses the legacy message.length-validator-error with length 1, preserving the existing user-facing copy. Regression assertions cover both the bridge rejection and the visible core-form error.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItemFormBridge.tsx
shah-harshit: Fixed in 77851139c7. Missing-destination and nested-field validation now follow separate branches. Only an empty required destination list sets and throws the legacy minimum error; invalid nested configuration still blocks submit through its inline core-form error without a false missing-destination reason. Added regression coverage.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItemFormBridge.tsx
shah-harshit: Confirmed. The synchronization snapshot is advanced before notifying the Ant adapter, so the echoed value remains deeply equal and does not trigger reset. The regression test keeps an advanced OAuth input focused while the parent echoes each edit, and rapid typing across the advanced fields was also verified manually.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationSelectItem.tsx
shah-harshit: Confirmed as intentional. AlertConfigDetails is a read-only view, so category, type, notify-downstream, depth, and destination configuration controls must all remain disabled. This state was verified manually from Settings > Alerts.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationSelectItem.tsx
shah-harshit: Confirmed as an intentional fix. Both local and API destination configs must be normalized so status results match destinations containing headers or query parameters. The Slack destination status E2E coverage now exercises this configured case.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationSelectItem.tsx
shah-harshit: Confirmed as intentional. When Notify Downstream is enabled, downstream depth is required and must be a finite positive value. Empty, non-numeric, zero, and negative values should block save rather than silently pass validation.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationConfigField/DestinationConfigField.tsx
shah-harshit: Fixed in 77851139c7. The HTTP Method label is restored above the POST and PUT controls, and it is rendered inside the core RadioGroup so the visible label is also associated with the control. Unit coverage verifies the label.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationSelectItem/DestinationConfigField/DestinationConfigField.tsx
shah-harshit: Confirmed as intentional. The shared destination component should reject malformed email receiver tags instead of accepting values that cannot receive notifications. Unit coverage verifies that invalid values are rejected and valid email addresses are accepted.

openmetadata-ui/src/main/resources/ui/src/components/Alerts/DestinationFormItem/DestinationFormItem.component.tsx
shah-harshit: Verified. Untouched timeout values can remain absent from the parent form, matching the legacy behavior. The event-subscription destination schema supplies the persisted defaults of 10 seconds for connection timeout and 12 seconds for read timeout, which match the values displayed by the core form. No payload-builder change is required.

chirag-madlani (approved):

PR #31898, opened 2026-08-21
Fixes 30522: bound the incident severity chip so a long translation cannot push the Assignee column off screen

TeddyCr:
### Describe your changes:

Fixes #30522

## ⚠️ The mechanism has changed since the report — please read before validating against the screenshot

The issue title says severity and assignee **overlap**. **Nothing overlaps on current `main`.** The 1.13.1 screenshot came from an Ant Design `Table` whose Severity column was pinned at `width: 100`; that table was replaced by the `ui-core-components` `Table` in `8570d36830` / `2c90c83ae3`, which has no per-column `width` or `ellipsis`. (`overflow-x` also computes to `visible` on the `<table>` despite the `tw:overflow-x-hidden` class, because **`overflow` does not apply to `display: table`** — the wrapper genuinely scrolls, nothing is clipped or collided.)

What survived the migration is **column inflation**. In ru-RU, `label.no-entity` + `label.severity` composes to `Критичность инцидента отсутствует` — 218px of rendered text against 67px for English "No Severity". The chip is nowrap and unbounded, so its intrinsic width becomes the column's floor:

**chip 258px → Severity column 306px → table 1263px inside a 1178px container → the Assignee column's right edge lands at 1508px on a 1440px viewport.**

Validating against the original screenshot will not match. Validate against the before/after described under "UI screen recording" below.

**The symptom requires the left nav to be EXPANDED.** The nav moves the table's container between 1178px (expanded) and 1334px (collapsed); collapsed, the pre-fix layout happens to fit (assignee edge 1423px, zero overflow). A maintainer reproducing with the nav collapsed will see no bug and conclude the report was wrong.

## What changed

https://github.com/user-attachments/assets/a475b027-08f8-41df-b592-af4888646f23

`ChipTrigger` gains an opt-in `maxChipWidth`; only the severity call site passes it (`tw:max-w-44`, 176px). The label truncates inside the pill, with `title` surfacing the full string on hover. Truncation is visual only — the text stays in the DOM, so the button's accessible name is unchanged. Both the truncation and the `title` are gated on the prop, so an unbounded chip gains neither.

**Why 176px, measured on the live pill** (ru-RU @1440, nav expanded):

| cap | Severity `<td>` | assignee right edge | |
|---|---|---|---|
| unbounded | 306px | 1508px | off screen |
| **176px (`max-w-44`, chosen)** | 228px | 1430px | on screen |
| 184px | 236px | 1438px | on screen |
| 188px | 240px | 1442px | **off screen** |
[...]

PR #32080, opened 2026-08-26
ci(playwright): stop the plan step discarding the checked-in timing baseline

ShaileshParmar11:
> **Blocked by #31898**, which carries the `build_playwright_shards.py` change this depends on. Merging this first would leave a zero-artifact run with no history at all.

## The bug

The `Build duration-aware shard plans` step builds `history_args` from the downloaded full-run artifacts and reaches for the checked-in baseline only when that array comes out empty:

```bash
if [[ ${#history_args[@]} -eq 0 ]]; then
  history_args=(--history .github/playwright/timing-baseline.json)
fi
```

`gh run download` prints nothing on success, so a run can log **49 failures across its 50 candidates**, quietly succeed on the 50th, and discard the baseline on the strength of that one artifact.

Everything already covered by an artifact is unaffected, which is why this went unnoticed. The only entries that exist *solely* in the baseline are the ones an author seeds for a brand-new spec file — exactly as the stale-baseline gate's own error text instructs:

> Seed the observed durations for these files into `.github/playwright/timing-baseline.json` (see PR #30812 for the pattern…)

Those seeded timings were dropped, and the gate then failed the plan for a file that *did* have history. Net effect: **a brand-new spec file is unseedable whenever any full-history artifact is available**, which is most of the time.

## Evidence

[#31898](https://github.com/open-metadata/OpenMetadata/pull/31898) hit this twice — runs [32933633223](https://github.com/open-metadata/OpenMetadata/actions/runs/32933633223) and [32937328644](https://github.com/open-metadata/OpenMetadata/actions/runs/32937328644), both 49/50 download failures. The artifact they used, from run 32653484382, carries 4504 tests and none of the new file's.

Replaying run 32937328644's own selection with only its artifact as history:

| planner input | result |
|---|---|
| artifact only, before #31898's script change | exit 1 — the reported failure |
| artifact only, after it | exit 0 |

Planning that same selection *minus* the new spec produces byte-identical shard plans either way, so nothing the artifacts already cover moves.

## The change

`build_playwright_shards.py` now folds the baseline in itself, at lower precedence than anything passed via `--history`. That makes the fallback here not merely redundant but harmful: it re-introduces the baseline as a *peer* of the artifacts, diluting the p75 those weights are built from. Drop it, and log how many artifacts actually resolved so the next person can see at a glance what fed the planner.

The PR updates Playwright shard planning so downloaded timing artifacts are passed as higher-precedence history while the planner supplies the checked-in baseline itself.
- Removes the workflow-level fallback that passed the baseline as a peer history source.
- Logs the number and arguments of downloaded timing-history artifacts.
</details>

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| .github/workflows/playwright-e2e-reusable.yml | Removes the redundant baseline fallback and adds artifact-source logging; the planner at current HEAD backfills the checked-in baseline beneath artifact history. |

</details>

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into claude/ci-plan-..."](https://github.com/open-metadata/openmetadata/commit/23b20c3f41a4b843559b46642db6fdbb3fe4e07f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56981285)</sub>


ShaileshParmar11:
Closing — not needed.

The actual bug (a newly added spec file's seeded timings being discarded whenever any full-history artifact downloaded) is already fixed on `main` by the `build_playwright_shards.py` backfill that landed with #31898. The planner now folds `.github/playwright/timing-baseline.json` in itself, below anything passed via `--history`, so every PR can seed timings for a new spec file today.

That leaves this PR as cleanup only, and one of its stated justifications does not hold. The description argued that keeping the fallback "re-introduces the baseline as a peer of the artifacts, diluting the p75". That is wrong: the fallback only fires when `history_args` is empty — i.e. when there are zero artifacts — so the baseline can never sit alongside them. Verified by replaying the same selection with the fallback kept vs removed: byte-identical shard plans. The fallback is redundant, not harmful.

The one piece here with lasting value is the `history sources: N artifact(s) + the checked-in baseline` log line — `gh run download` is silent on success, so nothing currently records what actually fed the planner, which is most of why the original bug took so long to find. Worth re-proposing on its own if anyone wants it; not worth a review cycle bundled with dead-code removal.

PR #32128, opened 2026-08-26
fix(superset): don't drop calculated columns with no declared type

### LINKED ISSUE #24986: When the SupersetDataModel of the Superset connector synchronizes metadata, it does not synchronize CALCULATED COLUMNS to the SupersetDataModel
I cannot query the CALCULATED COLUMNS of Superset on OpenMetadata, nor add classification tags and terms 

I want the SupersetDataModel on the Superset connector in OpenMetadata to be able to synchronize the metadata of CALCULATED COLUMNS 

Adding calculated columns using source on Superset's dataset, but with too many data tables, this is particularly time-consuming and energy-consuming

**Additional context**

Image

akashverma0786:
### Describe your changes:

Fixes #24986

Superset calculated columns (columns defined by a SQL expression instead of synced from a physical table) don't require a "Data type" to be set — Superset itself leaves `type` `null`/empty for most of them by default. `SupersetSourceMixin.get_column_info()` was treating a falsy `type` as a reason to silently drop the column entirely, so any calculated column with no declared type never made it into the `SupersetDataModel` at all — which is exactly why the reporter couldn't query, tag, or attach glossary terms to it: the column simply didn't exist on the OpenMetadata side.

This PR makes `get_column_info()` fall back to `UNKNOWN` (the same convention already used by every other dashboard connector — Tableau, PowerBI, Looker, MicroStrategy, Qlik, Sigma, SSRS, Omni — for a missing type) instead of dropping the column, and surfaces the column's SQL expression in its description (`**Formula:** \`...\``) so it's clear the column is calculated and what formula defines it. Applies to both the API-based and DB-based connector variants, which share this logic but previously diverged in their column models (`DSColumns`/`FetchColumn` didn't carry `expression` at all, so even a fix to the type-handling alone wouldn't have surfaced the formula).

#
### Type of change:
- [x] Bug fix

#
### High-level design:

N/A — small, contained change (3 source files + 1 test file). Root cause: `get_column_info()` (`mixin.py`) had `if field.type:` guarding the whole column-building block; a Superset column with `type: null` (common for calculated columns) was silently skipped instead of ingested. Fix replaces that with a fallback (`field.type = ... if field.type else "UNKNOWN"`), adds `expression` to the two column Pydantic models (`DSColumns` for the API source, `FetchColumn` for the DB source) so the formula is available to surface, and adds `tc.expression` to the DB source's `FETCH_COLUMN` SQL query so the DB path actually receives it.

#
#### Use cases covered
- A Superset calculated column with `type: null` is ingested as an OpenMetadata column with `dataType: UNKNOWN` instead of being dropped (both API-source and DB-source connector paths).
- A calculated column that *does* have a type set keeps that real type (doesn't get overwritten to `UNKNOWN`).
- A calculated column's SQL expression is surfaced in its OpenMetadata column description as `**Formula:** \`...\``; a physical column without an expression keeps its plain description untouched.
- The DB-source path's `FETCH_COLUMN` SQL query round-trips real rows (physical + untyped-calculated + typed-calculated) from Postgres in a single call with no cross-contamination between columns.

#### Unit tests
- Not added. Test coverage for this change is intentionally kept at the ingestion-integration level only (see below), which exercises both connector paths against real Superset/Postgres containers rather than mocked objects.

- Not applicable (ingestion/Python change, no `openmetadata-service` REST API changes).

- Updated `ingestion/tests/integration/superset/test_superset.py`:
  - Fixed a latent gap in the test's synthetic Postgres fixture (`setup_sample_data`) — its hand-built `table_columns` table was missing an `expression` column, which real Superset's schema has. Adding `tc.expression` to `FETCH_COLUMN` (required for this fix) would otherwise break `SupersetDBSource` against that fixture.
  - Added 4 new tests: `test_calculated_column_missing_type_not_dropped_db`, `test_calculated_column_missing_type_not_dropped_api`, `test_typed_calculated_column_keeps_type_and_formula_api`, `test_calculated_columns_end_to_end_via_sql_db`.
  - Full suite: 27/27 passing (`python -m pytest tests/integration/superset/test_superset.py`, real Postgres + Superset via testcontainers).

#### Playwright (UI) tests
- Not applicable (no UI changes).

#### Manual testing performed
1. Reproduced the bug against a live Superset instance (Docker): confirmed pre-existing, unmodified example data (`birth_names.num_california`, `FCC 2018 Survey.job_location_preference`/`first_time_developer`) already has `type: null` calculated columns out of the box — not a contrived edge case.
2. Ran the real `metadata ingest` CLI (API-source config) against that instance before and after the fix — confirmed the affected columns were silently missing from the resulting `DashboardDataModel` before, and present with `dataType: UNKNOWN` + formula in the description after.
3. Stood up a second, Postgres-backed Superset instance specifically to exercise the DB-source connector variant (the default Superset Docker image uses SQLite, which OpenMetadata's DB-source connector doesn't support) — ran `metadata ingest` with a Postgres `serviceConnection`, confirmed identical results through the SQL-based code path.
4. Verified via direct Postgres queries (`psql`) that Superset's real `table_columns` schema has an `expression` column, confirming the model/query change matches production Superset, not just the test fixture.

#

Not applicable.

#
### Checklist:
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed. — N/A, no schema changes.
- [ ] For UI changes: I attached a screen recording and/or screenshots above. — N/A, no UI changes.

The PR preserves untyped Superset calculated columns by mapping missing types to `UNKNOWN` and includes their SQL expressions in column descriptions.
- Adds expression fields to API- and database-backed Superset column models.
- Fetches expressions through the database query and maps them into typed OpenMetadata columns.
- Extends integration coverage for typed and untyped calculated columns across both connector paths.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/source/dashboard/superset/mixin.py | Maps missing column types to `UNKNOWN`, preserves typed columns, and appends calculated expressions to descriptions. |
| ingestion/src/metadata/ingestion/source/dashboard/superset/models.py | Adds optional expression metadata to both Superset column representations. |
| ingestion/src/metadata/ingestion/source/dashboard/superset/queries.py | Extends the database-backed column query to retrieve calculated expressions. |
| ingestion/tests/integration/superset/test_superset.py | Covers calculated-column ingestion through API and database connector paths. |
| ingestion/.basedpyright/baseline.json | Removes resolved Superset type-checking baseline entries. |

</details>

<sub>Reviews (3): Last reviewed commit: ["fix(superset): resolve pre-existing type..."](https://github.com/open-metadata/openmetadata/commit/508813507abb0ff68a08ada1946e66dcdcd37aa7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57221455)</sub>


ingestion/src/metadata/ingestion/source/dashboard/superset/mixin.py
```diff
@@ -391,22 +392,31 @@ def get_column_info(self, data_source: list[DSColumns | FetchColumn]) -> list[Co
...
-                    )
-                    datasource_columns.append(parsed_fields)
+                # Calculated columns commonly have no `type` in Superset (it's optional
+                # there); fall back to UNKNOWN instead of dropping the column entirely.
+                field.type = self._clearn_column_datatype(field.type) if field.type else "UNKNOWN"
+                col_parse = ColumnTypeParser._parse_datatype_string(  # pylint: disable=protected-access
+                    field.type
+                )
+                if not isinstance(col_parse, dict):
+                    logger.warning(f"Unexpected datatype parse result for column {field.column_name}: {col_parse!r}")
```
IceS2: Please stop using f-string on logs

IceS2 (dismissed):

ulixius9 (approved):

PR #32133, opened 2026-08-27
fix: make FrameworkSeedLoader idempotent to handle existing fqnhash (#32103)

### LINKED ISSUE #32103: [2.0.0] AI Governance FrameworkSeedLoader is not idempotent and fails on existing fqnhash
### Affected module

Other

### Describe the bug

After upgrading OpenMetadata to 2.0.0, the application starts successfully, but the FrameworkSeedLoader logs multiple warnings and PostgreSQL exceptions while loading the built-in AI Governance frameworks.

The loader attempts to insert AI Governance framework controls that already exist in ai_framework_control_entity.

PostgreSQL rejects the inserts because the existing controls have the same fqnhash:

ERROR: duplicate key value violates unique constraint
"ai_framework_control_entity_fqnhash_key"

This happens for several built-in AI Governance frameworks, not just one.

Environment
OpenMetadata version: 2.0.0
Database: PostgreSQL
Search backend: OpenSearch
Deployment: Kubernetes
Java: 21

The upgrade migration to 2.0.0 is registered as applied by:

./bootstrap/openmetadata-ops.sh info

The output contains:

| 2.0.0 | 2026-08-26 10:46:44.531417 | ... |
Actual behavior

During startup, FrameworkSeedLoader tries to create controls that are already present in the database.

Example for Canada AIDA:

WARN FrameworkSeedLoader - Failed to load framework seed
json/data/aiGovernance/frameworks/canada_aida.json:

org.postgresql.util.PSQLException:
ERROR: duplicate key value violates unique constraint
"ai_framework_control_entity_fqnhash_key"

Detail:
Key (fqnhash)=(803a7d74349115176693366aa97d4405) already exists.

The attempted entity is:

{
  "name": "aida-1",
  "fullyQualifiedName": "aida-1",
  "displayName": "High-impact System Designation",
  "framework": {
    "name": "canada_aida"
  }
}

The existing PostgreSQL row has the same logical entity:

name: aida-1
fqnhash: 803a7d74349115176693366aa97d4405
framework: canada_aida

but a different UUID.

The same issue occurs for multiple built-in frameworks:

canada_aida.json
singapore_mgf.json
nist_ai_rmf.json
eu_ai_act.json
us_bill_of_rights.json
iso_iec_42001.json
china_ai.json
uk_ai_regulation.json
soc2_ai.json

Examples include:

canada_aida
aida-1
fqnhash=803a7d74349115176693366aa97d4405

singapore_mgf
mgf-1
fqnhash=371ca796b58a8b98a42e4505d6377922

nist_ai_rmf
govern
fqnhash=564afb427cb3571f2af0285572efecf8

eu_ai_act
art-5
fqnhash=a53ba5e934b3ab1b5daa10280507f698

us_bill_of_rights
bor-safe
fqnhash=d04ac17ccb2f770ae1f9180e266ccdc1

iso_iec_42001
clause-4
fqnhash=926d4da78a52bf852785407010faca02

china_ai
cn-1
fqnhash=72e479ec1a262b1fe7b9592e0996e09f

uk_ai_regulation
uk-1
fqnhash=b32c736dc2f7667d08a8c669f52af24d

The stack trace shows the insert originating from:

AIFrameworkControlRepository.storeEntity()
EntityRepository.createNewEntity()
EntityRepository.create()
FrameworkSeedLoader.seedBundle()
FrameworkSeedLoader.loadFromResources()
AIGovernanceFrameworkResource.initialize()

The SQL operation is an unconditional insert:

INSERT INTO ai_framework_control_entity (fqnHash, json)
VALUES (:nameHashColumnValue, :json :: jsonb)

### To Reproduce

Have an OpenMetadata database containing the AI Governance framework seeds.
Upgrade/start OpenMetadata 2.0.0.
Observe the application startup logs.
FrameworkSeedLoader attempts to insert existing AI framework controls.
PostgreSQL returns:
duplicate key value violates unique constraint
"ai_framework_control_entity_fqnhash_key"
Restart OpenMetadata and observe the same seed-loading errors again.

### Expected behavior

Expected behavior

Loading the built-in AI Governance seeds should be idempotent.

If a framework/control already exists, FrameworkSeedLoader should detect it by FQN / fqnhash and either:

reuse the existing entity,
update it when required, or
skip it when the seed has already been loaded.

Restarting OpenMetadata, or upgrading an installation where these entities already exist, should not generate PostgreSQL unique-constraint violations.

Observed behavior

FrameworkSeedLoader appears to generate a new UUID and call the entity creation path even though an entity with the same FQN / fqnhash already exists.

Conceptually the current behavior seems to be:

load seed
   |
generate new entity UUID
   |
create()
   |
INSERT
   |
same FQN -> same fqnhash
   |
UNIQUE constraint violation

The expected behavior would be closer to:

load seed
   |
lookup entity by FQN/fqnhash
   |
   +-- exists --> reuse/update/skip
   |
   +-- missing --> create

Before calling EntityRepository.create(...), FrameworkSeedLoader could check whether the framework/control already exists by fullyQualifiedName / fqnhash.

Alternatively, the seed loading logic could use an idempotent create-or-update/upsert path rather than unconditional entity creation.

### OS

kubernetes

### Python version

_No response_

### OpenMetadata version

2.0.0

_No response_

### Additional context

The existing rows do not appear corrupted. For example, the existing aida-1 row contains the expected framework, description, category and evidence requirements.

The issue is also reproducible across multiple different built-in framework seeds, which suggests this is not an isolated hash collision.

Despite the warnings, OpenMetadata continues starting successfully and registers the AI Governance resources, so this appears to be a seed initialization/idempotency problem rather than a fatal database migration failure.

openmetadata-ops.sh info also shows the 2.0.0 migration as successfully applied.

### Pre-submission checklist

zhang-arvin:
## Describe the issue

The `FrameworkSeedLoader` is not idempotent. When `fqnHash` already exists in the `ai_framework_control_entity` table, the plain `INSERT` fails with a unique constraint violation, causing the seed process to fail on restart or re-run.

Closes #32103

## Fix

Add try-catch around `create()` calls in `seedBundle()` and `seedControls()` to gracefully handle duplicate key exceptions, logging a message and continuing instead of failing.

<h2><a href="https://app.greptile.com/api/retrigger?id=57351294"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because framework and control creation can still silently fail for reasons unrelated to duplicate records.

<h2>Findings</h2>

1. P1&nbsp;**Creation failures are swallowed** <a href="https://github.com/open-metadata/OpenMetadata/pull/32133#discussion_r3868234959">▶</a>

Makes AI governance framework seeding tolerate records that already exist.
- Wraps framework and control creation in exception handling.
- Logs and continues when creation fails.
- The previously reported blanket exception handling remains unchanged and still treats non-duplicate persistence failures as existing records.
</details>

```mermaid
flowchart TD
  A[Load seed bundle] --> B{Framework found by FQN?}
  B -->|Yes| C[Use existing framework]
  B -->|No| D[Create framework]
  D -->|Success| E[Seed controls]
  D -->|Any exception| F[Log as already existing and return]
  C --> E
  E --> G[Create each missing control]
  G -->|Success| H[Continue]
  G -->|Any exception| I[Log as already existing and continue]
```
</details>

<sub>Reviews (2) · Last reviewed commit: ["Merge branch &#39;main&#39; into fix/32103-seedl..."](https://github.com/open-metadata/openmetadata/commit/8e3644d63d0fad3b2dc2052013577f3541849132)</sub>


sonika-shah:
Thanks for digging into this @zhang-arvin — the bug is real and the diagnosis (idempotent seeding on restart / upgrade) is right.

The concern with the current approach is the blanket `catch (Exception)` that logs everything as "already exists":

1. **It hides genuine failures.** A validation error, lost connection, or serialization failure gets logged at INFO as "already exists, skipping" — the exact log signal used to diagnose #32103 in the first place is lost.
2. **It defeats the seed gate.** `loadFromResources` catches per-bundle and calls `SeedDataGate.recordSeedFailure()`, which makes `stampIfClean()` store a `null` fingerprint so the **next restart retries** the seed. Swallowing inside `seedBundle`/`seedControls` means a genuine failure never reaches that handler → the gate stamps "clean" → `shouldSeed()` returns `false` on the next boot → the missing framework/controls are **never** re-seeded. A recoverable partial seed becomes permanent.
3. **`return` on framework-create failure skips all controls** for that bundle, so a framework that already exists stops its controls from being seeded.

Also worth noting: the loader already guards each `create()` with `findByNameOrNull`, so a single-node re-run is already a clean no-op. The remaining collision is the **concurrent multi-replica startup** (matches the reporter's Kubernetes env) where two pods both miss the existence check and both INSERT.

I've put up an alternative that keeps your direction but narrows it: #33756. It catches the specific JDBI `UnableToExecuteStatementException` and re-fetches by FQN — reuse the row a concurrent seeder wrote, rethrow when it's genuinely absent so the seed gate stays retryable — plus two deterministic branch tests (recover vs propagate). Happy to fold in any feedback either way.

sonika-shah:
Closing this in favour of #33756, which builds directly on your diagnosis here — thank you @zhang-arvin, the root-cause read (seed loader isn't idempotent for already-present rows) was correct and got this moving.

#33756 keeps your approach but narrows it in three ways so it's safe to merge:

- Catches the specific JDBI `UnableToExecuteStatementException` and **re-fetches by FQN** instead of a blanket `catch (Exception)`, so a genuine failure (validation, lost connection, etc.) is **rethrown and logged**, never silently recorded as "already exists".
- Keeps the seed gate accurate — a real failure still reaches `loadFromResources`, so the seed stays retryable on the next restart instead of being stamped "done" with rows missing.
- Reuses the raced framework row rather than `return`ing, so its controls are still seeded on the same pass, plus two regression tests (recover vs propagate).

No action needed from you — just wanted to credit where it started. Feedback on #33756 very welcome if you'd like to review it.

PR #32135, opened 2026-08-27
fix: mask PubSub credentials in backend (#32102)

### LINKED ISSUE #32102: PubSub connector private key is displayed in plaintext in configuration UI
### Affected module

UI

### Describe the bug

Private key values in the PubSub connector configuration are currently visible as plaintext in the UI. Sensitive credential fields should be masked to prevent unintended exposure to users with access to the configuration page.

### To Reproduce

**Steps to Reproduce**

- Navigate to the PubSub connector configuration.
- Open a configuration containing a populated private key.
- Observe that the private key is rendered in plaintext rather than masked.

### Expected behavior

The private key should be masked in the UI, consistent with other sensitive configuration fields. The value should only be revealed through an explicit, authorized action if supported.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

_No response_

_No response_

### Additional context

Image

### Pre-submission checklist

zhang-arvin:
## Describe the issue

Pub/Sub service configurations carry nested GCP credentials. The backend secret masker only recognizes password fields after those nested objects are converted to their generated schema classes; without that conversion, the private key can be returned in plaintext.

Closes #32102

## Fix

Match the existing BigQuery connection flow in the backend:

- add a PubSub connection converter that materializes gcpConfig as GCPCredentials and its nested GCP values before secret handling;
- register the converter in the existing converter factory;
- verify masking and unmasking from the map-shaped configuration returned from persistence.

No UI or JSON-schema change is required. The existing GCP values schema already marks privateKey as a password field.

## Test plan

- mvn -pl openmetadata-service -Dtest=PasswordEntityMaskerTest,ClassConverterFactoryTest -Dsurefire.failIfNoSpecifiedTests=false test
- mvn -pl openmetadata-service spotless:check

The PR registers a dedicated Pub/Sub connection converter so nested GCP credentials participate in the existing secret masking and unmasking flow.
- Adds `PubSubConnectionClassConverter` and registers it with `ClassConverterFactory`.
- Adds converter-registration and Pub/Sub private-key mask/unmask coverage.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/secrets/converter/ClassConverterFactory.java | Registers the new Pub/Sub converter under the generated `PubSubConnection` class used by production dispatch. |
| openmetadata-service/src/main/java/org/openmetadata/service/secrets/converter/PubSubConnectionClassConverter.java | Converts Pub/Sub configuration and its nested GCP credentials into typed objects for reflective secret processing. |
| openmetadata-service/src/test/java/org/openmetadata/service/secrets/converter/ClassConverterFactoryTest.java | Confirms that Pub/Sub resolves to a specialized converter rather than the default converter. |
| openmetadata-service/src/test/java/org/openmetadata/service/secrets/masker/TestEntityMasker.java | Verifies that a Pub/Sub service-account private key is masked and restored during unmasking. |

</details>

<sub>Reviews (9): Last reviewed commit: ["test: cover persisted PubSub credential ..."](https://github.com/open-metadata/openmetadata/commit/4bc4d11339742a2c9d2c79b7751ea533b5453727) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57359522)</sub>


ayush-shah:
Request changes — this PR is not safe to merge yet.

- **[P1] Private key is still exposed in plaintext.** In [ServiceConnectionDetailsUtils.tsx#L211-L225](https://github.com/open-metadata/OpenMetadata/blob/13421a3fefe450c7045f004754239f1aa02d7122/openmetadata-ui/src/main/resources/ui/src/utils/ServiceConnectionDetailsUtils.tsx#L211-L225), `childOneOf.map(...)` renders the same credential object against every `oneOf` branch. For a `service_account` payload, the `GCPValues` branch masks `privateKey`, but the `GCPExternalAccount` branch has no `privateKey` metadata, so the fallback renders the same value as a normal text input. The secret appears once masked and again in plaintext, and the credential fields are duplicated. Resolve/select the active branch using the credential discriminator (`value.type`) before rendering.

- **[P2] Referenced `oneOf` branches are not resolved.** [pubSubConnection.json#L211-L264](https://github.com/open-metadata/OpenMetadata/blob/13421a3fefe450c7045f004754239f1aa02d7122/openmetadata-ui/src/main/resources/ui/src/jsons/connectionSchemas/connections/messaging/pubSubConnection.json#L211-L264) contains `$ref` alternatives for credential-path and ADC configurations. The new code only recurses when inline `properties` exist, so these branches are skipped and their values are rendered against unrelated schemas. Resolve local refs before selecting the active branch.

- **[P1 — CI/build] The committed Pub/Sub schema does not match the repository generator.** `pubSubConnection.json` is generated output under the ignored `src/jsons/*` tree, while the authoritative schema is already in `openmetadata-spec`. Re-running [parseSchemas.js](https://github.com/open-metadata/OpenMetadata/blob/13421a3fefe450c7045f004754239f1aa02d7122/openmetadata-ui/src/main/resources/ui/parseSchemas.js) produces 438 lines versus the 346-line file committed here, and Prettier also rewrites it. A clean install/build can therefore modify the tracked file and fail UI checks. Please remove the generated artifact or regenerate it through the supported workflow.

- **[P2] Missing regression coverage.** Add a Pub/Sub service-account rendering test that asserts `privateKey` appears exactly once and uses `type="password"`, plus coverage for the credential-path, ADC, and external-account variants.

ayush-shah:
Follow-up: there has been no response on this review for approximately 20 hours (last comment: 2026-08-27 15:55 UTC). To keep the PR moving, I will add and push a focused follow-up commit addressing the review findings:

- select the active Pub/Sub credential `oneOf` branch by discriminator instead of rendering every variant;
- resolve referenced credential schemas before rendering;
- remove/regenerate the generated schema artifact through the supported workflow;
- add regression coverage proving `privateKey` renders once as `type="password"`.

I will re-review the resulting PR head after pushing the commit. Please flag if maintainers prefer a different implementation direction.

ayush-shah (approved):
Re-review: approved from a code perspective.

This is now a backend-only fix that follows the existing BigQuery credential-conversion path. Pub/Sub nested GCP credentials are materialized before the shared password masker runs, so the existing password-field annotation masks privateKey without any UI or JSON-schema changes.

The regression begins with the map-shaped configuration returned from persistence and verifies both masking and unmasking. Local verification passed:
- mvn -pl openmetadata-service -Dtest=PasswordEntityMaskerTest,ClassConverterFactoryTest -Dsurefire.failIfNoSpecifiedTests=false test (34 tests)
- mvn -pl openmetadata-service spotless:check

The only current blocking check is Validate PR Metadata, which concerns required Shipping fields on #32102 rather than this code change.

ayush-shah:
CI follow-up: the failed integration and Python jobs stopped at the repository label gate because this PR does not have the required `safe to test` label; the backend build and integration suites did not execute. The separate Validate PR Metadata failure is also configuration-only: issue #32102 is missing its Shipping Source, Priority, and Release fields. Neither failure is caused by the current backend diff.

yan-3005 (approved):

PR #32191, opened 2026-08-27
fix(governance-workflows): run Flowable's MySQL engine at READ_COMMITTED

yan-3005:
## What

Runs the embedded Flowable governance-workflow engine at `READ_COMMITTED` on MySQL (one line + import in `WorkflowHandler`, scoped to the MySQL branch).

## Why

The Flowable engine intermittently deadlocks in CI on MySQL under concurrent workflow advances (repeated DAR-suite flakes, both insert-side and delete-side victims observed across runs).

MySQL defaults to `REPEATABLE_READ`, which takes gap / next-key locks. An insert-intention lock (timer-job / variable inserts on one workflow advancing a stage) cycles against a range lock held by another transaction sweeping `ACT_RU_*` rows by `PROC_INST_ID_` (instance cleanup) or scanning due timer jobs — InnoDB picks a deadlock victim and that command fails. This is the textbook `REPEATABLE_READ` gap-lock deadlock signature.

`READ_COMMITTED` releases record locks on non-matching rows after the `WHERE` and does not take gap locks, removing this class of cycle.

Correctness / scope:
- Flowable's [own recommendation](https://forum.flowable.org/t/choosing-transaction-isolation-level-for-mysql/7185) is `READ_COMMITTED` for MySQL.
- It is Postgres's default — this stack already runs the full engine at `READ_COMMITTED` on Postgres, deadlock-free.
- Flowable's concurrency control is optimistic (`REV_` version columns), so it is correct at `READ_COMMITTED`.
- Scoped to the engine's own pooled datasource (built from `setJdbcUrl`); the OpenMetadata JDBI datasource is untouched.

## Honest status / caveats

- **Not reproduced locally.** The race is timing-dependent (cross-workflow overlap); I could not reproduce the deadlock in an isolated harness, so there is **no local before/after demonstration**. The case is diagnosis + Flowable/MySQL sources + the Postgres comparison.
- **Knob-bite not runtime-verified.** The change targets the `setJdbcUrl` path where Flowable honors `setJdbcDefaultTransactionIsolationLevel`, but I have not confirmed at runtime that live Flowable connections actually run at `READ-COMMITTED` (recommend confirming on a MySQL build via `SHOW ENGINE INNODB STATUS` / a startup probe before merge).
- MySQL docs note `READ_COMMITTED` *greatly reduces* — not mathematically eliminates — deadlocks. It removes the gap-lock class observed here.

**CI is the proof.** Please watch the MySQL IT lanes across runs.

The PR moves the MySQL `READ_COMMITTED` setting onto the process-engine configuration that is actually built, completing the previously requested fix.
- Applies `Connection.TRANSACTION_READ_COMMITTED` to Flowable’s JDBC URL configuration for MySQL.
- Leaves the migration-supplied datasource path unchanged.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/governance/workflows/WorkflowHandler.java | The built non-migration MySQL engine configuration now receives the intended transaction-isolation setting, resolving the prior dropped-setting issue. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(governance-workflows): apply READ\_CO..."](https://github.com/open-metadata/openmetadata/commit/2720b9a43e85f8a31a23984f24c4e2e8e05d35d0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57554563)</sub>


yan-3005:
Confirmed and fixed in 2720b9a43e — the finding is correct.

The isolation level was set on the constructor's local `StandaloneProcessEngineConfiguration`, which is only handed to `initializeNewProcessEngine` as a value carrier. That method builds a fresh config and copies jdbcUrl/user/pass/driver, databaseType, schemaUpdate and the async settings before `buildProcessEngine()` — the isolation level was not among the copied fields, so the runtime engine still ran at REPEATABLE_READ. No-op, as flagged.

Moved the `READ_COMMITTED` setting into `initializeNewProcessEngine`, onto the config that is actually built, guarded by the MySQL `databaseType` check, in the non-migration `jdbcUrl` branch — the only path where Flowable builds its own pooled datasource and honours the knob (the migration branch supplies a `DataSource`, which Flowable uses as-is). Reverted the dead constructor call so nothing misleading remains.

sonika-shah (approved):

Issue #26700, opened 2026-03-23
Semantic search with DJL embedding provider fails on Alpine-based server image (1.12.3)


edmei-maven:
## Affected Module                                                                                                                                                          

  Backend — Semantic Search / Vector Embedding initialisation (SearchRepository)                                                                                              
                                                                                                                                                                              
  ## Describe the Bug                                                                                                                                                         

  Enabling semantic search with the DJL embedding provider (`EMBEDDING_PROVIDER=djl`) on the official Docker image `docker.getcollate.io/openmetadata/server:1.12.3` fails at
  startup. The DJL library attempts to download PyTorch native libraries from `publish.djl.ai`, but these pre-built `.so` files are compiled against **glibc** and cannot load
   on the Alpine Linux 3.23 base image, which uses **musl libc**.

  The server starts and runs, but semantic search is non-functional — the `SearchRepository` logs the failure and falls back silently.

  **Error from startup logs:**

  ERROR [main] o.o.s.s.SearchRepository - Failed to initialize vector search service: Failed to load PyTorch native library
  ai.djl.engine.EngineException: Failed to load PyTorch native library
      at ai.djl.pytorch.engine.PtEngine.newInstance(PtEngine.java:90)
      at ai.djl.pytorch.engine.PtEngineProvider.getEngine(PtEngineProvider.java:41)
      at ai.djl.engine.Engine.getEngine(Engine.java:195)
      at ai.djl.Model.newInstance(Model.java:99)
      at ai.djl.repository.zoo.BaseModelLoader.createModel(BaseModelLoader.java:218)
      at ai.djl.repository.zoo.BaseModelLoader.loadModel(BaseModelLoader.java:169)
      at ai.djl.repository.zoo.Criteria.loadModel(Criteria.java:151)
      at org.openmetadata.service.search.vector.client.DjlEmbeddingClient.(DjlEmbeddingClient.java:57)
      at org.openmetadata.service.search.SearchRepository.initializeVectorSearchService(SearchRepository.java:405)
  Caused by: java.lang.UnsatisfiedLinkError: /home/openmetadata/.djl.ai/pytorch/2.7.1-cpu-linux-x86_64/libc10.so: Error loading shared library libstdc++.so.6: No such file or
   directory (needed by /home/openmetadata/.djl.ai/pytorch/2.7.1-cpu-linux-x86_64/libc10.so)
[...]

PR #32225, opened 2026-08-28
feat(search): add onnx as a local embedding provider

pmbrull:
## Describe your changes

Adds an `onnx` embedding provider that runs `all-MiniLM-L6-v2` in process via langchain4j. Purely additive — `djl` and every cloud provider stay, and the default stays `bedrock`.

The difference from `djl` is where the model comes from:

| | `djl` | `onnx` |
|---|---|---|
| Jar cost | 100 KB stub | **+168.8 MB** |
| Startup | downloads **~276 MB** from `publish.djl.ai` | none |
| Air-gapped | impossible | works |
| Models | any HF sentence-transformers model | frozen to all-MiniLM-L6-v2 |
| Deterministic in CI | no (network) | yes |

Both resolve to the same model at 384 dimensions, so the two are index-compatible.

Worth knowing: nothing persists `~/.djl.ai` — not the compose files, not the helm chart — so DJL's 276 MB is paid on **every fresh container**, not once.

### The cost, stated plainly

Measured from the built distribution:

```
langchain4j-embeddings-all-minilm-l6-v2   79.2 MB   (model weights)
com.microsoft.onnxruntime:onnxruntime      88.7 MB   (all-platform natives)
langchain4j-core + langchain4j-embeddings   0.8 MB
                                          -------
                                          168.8 MB
```

`openmetadata-dist/src/main/assembly/binary.xml` puts every runtime dep in `libs/`, so all of it lands in the release tarball and the Docker image (verified: image 455 MB → 623 MB). **Everyone pays it** — including deployments on bedrock/openai/google, and those with embeddings switched off entirely.

### The argument against merging this as-is

I'd rather put this up front than have it found in review.

1. **#32182 fixes DJL.** The original driver was partly that DJL didn't work on the alpine image. It does now, on the distroless base — verified end to end (`Initialized DjlEmbeddingClient … dimension=384`, zero linkage errors). So "we have no working local provider" is no longer true.
2. **The other driver was a test seam.** Collate's `NlqHybridSearchEndpointIT` uses a test-only `MiniLmEmbeddingClient` behind a system property; promoting it to a real provider would let that seam be deleted. But that seam can also be removed by pointing the IT at `provider=djl` plus an `actions/cache` step on `~/.djl.ai` — no OSS change, no 168.8 MB.
3. **The remaining unique capability is air-gapped deployment**, which DJL cannot do at any price. That's real, but nobody has asked for it yet.

If reviewers want the capability without the weight, the alternative is an optional module resolved reflectively (the `McpServerProvider` pattern already in the codebase) rather than the compile-time `switch` case here — the `case "onnx" ->` is precisely what forces the class, and therefore the jars, into the default build. Happy to rework it that way.

## Type of change

- [x] Improvement

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/open-metadata/OpenMetadata/blob/main/CONTRIBUTING.md) documentation
- [x] I have commented on my code, particularly in hard-to-understand areas

## Tests

`OnnxEmbeddingClientTest` — 9 tests against the **real in-process model**, no mocks, ~1.7 s:

- rejects a missing `onnx` config block and an unsupported model name
- defaults to MiniLM, reports dimension 384
- `embed` is deterministic and non-zero; the same text twice gives an identical vector
- null / blank text embeds to the zero vector
- related descriptions land closer than unrelated ones (cosine)
- `embedBatch` matches single embeddings and keeps blank entries index-aligned as zero vectors
- empty and null batches return empty

Plus `Provider.ONNX` added to `SearchRepositoryBehaviorTest`'s reject-incomplete-config case, and `NaturalLanguageSearchConfigTest` proves the new `onnx:` block in `conf/openmetadata.yaml` deserializes.

```
Tests run: 164, Failures: 0, Errors: 0, Skipped: 0   BUILD SUCCESS
  OnnxEmbeddingClientTest                     9
  SearchRepositoryBehaviorTest              127
  EmbeddingClientTest                         9
  EmbeddingModelChangeDetectionTest          10
  SystemRepositoryEmbeddingsValidationTest    7
  NaturalLanguageSearchConfigTest             2
```

Also verified in a real container on the #32182 base — server boots with `EMBEDDING_PROVIDER=onnx`:

```
OnnxEmbeddingClient        - Initialized OnnxEmbeddingClient with model=all-MiniLM-L6-v2, dimension=384
ElasticSearchVectorService - initialized with model=all-MiniLM-L6-v2, dimension=384
SearchRepository           - Vector search service initialized with provider=onnx, dimension=384
```

Zero `UnsatisfiedLinkError` / `NoClassDefFoundError`. Model load costs ~5.5 s of startup, and the tokenizer native is extracted from the jar rather than downloaded — no network at any point.

### One design note

`getModelId()` is stamped into the index `_meta`, so the client normalises the configured name to canonical `all-MiniLM-L6-v2` casing. Without that, `embeddingModel: all-minilm-l6-v2` in config would read as a model change and force a full reindex. A unit test caught this.

## Note for maintainers

No issue is linked because I couldn't find an existing one covering a local/air-gapped embedding provider — #26700 is the DJL-on-alpine bug and is already claimed by #32182. If this direction is wanted, an issue should be filed first; if it isn't, closing this is a perfectly good outcome and the analysis above is the deliverable.

### DIFF
--- a/conf/openmetadata.yaml
+++ b/conf/openmetadata.yaml
@@ -638,7 +638,7 @@ llmConfiguration:
   # (bedrock.awsConfig, openai.apiKey/endpoint, google.apiKey); this section only selects the
   # embedding provider + model. The embedding provider may differ from the chat provider above.
   embeddings:
-    provider: ${EMBEDDING_PROVIDER:-bedrock}  # bedrock | openai | google | djl
+    provider: ${EMBEDDING_PROVIDER:-bedrock}  # bedrock | openai | google | djl | onnx
     maxConcurrentRequests: ${MAX_CONCURRENT_EMBEDDING_REQUESTS:-10}
     bedrock:
       embeddingModelId: ${AWS_BEDROCK_EMBED_MODEL_ID:-"amazon.titan-embed-text-v2:0"}
@@ -651,6 +651,10 @@ llmConfiguration:
       embeddingDimension: ${GOOGLE_EMBEDDING_DIMENSION:-768}
     djl:
       embeddingModel: ${DJL_EMBEDDING_MODEL:-"ai.djl.huggingface.pytorch/sentence-transformers/all-MiniLM-L6-v2"}
+    # In-process ONNX. No credentials and no model download: the weights ship with the server, so
+    # this is the provider that works in an air-gapped deployment.
+    onnx:
+      embeddingModel: ${ONNX_EMBEDDING_MODEL:-"all-MiniLM-L6-v2"}
 
 eventMonitoringConfiguration:
   eventMonitor: ${EVENT_MONITOR:-prometheus}  # Possible values are "prometheus", "cloudwatch"
--- a/openmetadata-service/pom.xml
+++ b/openmetadata-service/pom.xml
@@ -569,6 +569,14 @@
       <version>0.34.0</version>
     </dependency>
 
+    <!-- Dependencies for vector embedding (in-process ONNX). The model weights are bundled in the
+         artifact, which is what lets the onnx provider embed without any download or credential. -->
+    <dependency>
+      <groupId>dev.langchain4j</groupId>
+      <artifactId>langchain4j-embeddings-all-minilm-l6-v2</artifactId>
+      <version>${langchain4j-embeddings.version}</version>
+    </dependency>
+
     <dependency>
       <groupId>org.glassfish.jersey.core</groupId>
       <artifactId>jersey-client</artifactId>
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SystemRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SystemRepository.java
@@ -1015,11 +1015,13 @@ public class SystemRepository {
 
       return switch (provider.toLowerCase()) {
         case "djl" -> getDjlEmbeddingMessage(embeddings);
+        case "onnx" -> getOnnxEmbeddingMessage(embeddings);
         case "bedrock" -> getBedrockEmbeddingMessage(llmConfig, embeddings);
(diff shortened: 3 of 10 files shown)


pmbrull:
Closing — not pursuing this.

The case for it thinned out as the surrounding work landed:

- **#32182 fixes DJL.** The original driver was partly that no local provider worked on the alpine image. DJL now initialises on the distroless base, verified end to end. "We have no working local provider" is no longer true.
- **The Collate test seam has cheaper resolutions** that don't involve shipping anything upstream.
- **The remaining unique capability is air-gapped local embeddings**, and nobody has asked for it.

Against that, 168.8 MB in every image and tarball — paid by everyone, including the majority on bedrock/openai/google or with embeddings switched off entirely.

The variant worth recording, because it looks attractive and isn't: shipping the provider but keeping the jars out of production builds (test-scope, or an optional module). That would leave `provider: onnx` in the config enum — a schema-level advertisement of a capability the shipped artifact cannot deliver. An operator sets it, the server fails at startup with a class-loading error, and the config surface is what misled them. A provider that only works in some builds is worse than no provider.

If air-gapped local embeddings turns up as a real requirement, the right shape is an optional module resolved reflectively (the `McpServerProvider` pattern already in the codebase) *and* config validation that rejects the provider when the module is absent — not a compile-time `switch` case. The `case "onnx" ->` here is exactly what forces the jars into the default build.

Measurements from this branch, kept for whoever revisits it:

```
langchain4j-embeddings-all-minilm-l6-v2   79.2 MB   (model weights)
com.microsoft.onnxruntime:onnxruntime      88.7 MB   (all-platform natives)
langchain4j-core + langchain4j-embeddings   0.8 MB
                                          -------
                                          168.8 MB      image 455 MB -> 623 MB
```

The branch stays if anyone wants the code.

PR #32246, opened 2026-08-28
Fixes 20966: clear sourceHash on PATCH so the bulk fast-path resyncs edited entities

### LINKED ISSUE #20966: Weird behaviour around Ingesting tags due to sourceHash
Before explaining the issue, I'd like to go through some "features" we've got in OpenMetadata:

- OpenMetadata prioritises UI changes vs the source (Unless explicitely defining `override metadata`).
- In order to avoid processing records that have not been changed since the last ingestion, we calculate and save a `sourceHash` attribute and use it to skip records if the one we've got on the database is the same as the one we're receiving from the source.
- The `sourceHash` does not change when changes are done to the Entity. It only changes if the source changes.

Due to the features explained above, something weird can happen. Imagine the following scenario:

1. I ingest `TABLE A` to OpenMetadata
2. Someone by mistake does some change such as removing a tag that is present in the source.
3. When the ingestion runs again, the tag won't be reingested as the `sourceHash` will remain the same and the table will be skipped.
4. If there is a change in `TABLE A`, the `sourceHash` will change and thus the tag will be ingested again.

This ends up being a weird mix between not ingesting and ingesting the tag depending on other changes.

zak-nuccio:
## Why

Closes #20966.

The bulk `PUT /api/v1/tables/bulk` fast-path (introduced in #28155) skips any entity whose
connector-supplied `sourceHash` matches the stored value, before any field hydration or diffing.
The stored `sourceHash` is **never cleared when an entity is edited out-of-band via PATCH**, so a
manual edit (e.g. a tag removed in the UI) is silently lost on the next ingestion run: the identical
`sourceHash` compares equal and the entity is skipped.

Concretely:
1. Ingestion writes a table with tags -> stored `sourceHash = H(tags)`.
2. Tags are cleared via PATCH (UI/automation/direct API) -> stored `sourceHash` stays `H(tags)`.
3. Next ingestion re-derives the same tags from source -> same hash `H(tags)`.
4. The fast-path sees `H(tags) == H(tags)` -> skip -> tags never restored.

Before 2.0.0 the bulk path processed every entity, so tags always re-synced from source. This is a
regression from the sourceHash feature.

## Fix

Clear `sourceHash` on every PATCH. All PATCH endpoints funnel through
`EntityRepository.patchCommonWithOptimisticLocking`, so this is a single point of change. The next
bulk run then sees a null stored hash, `isSourceHashUnchanged` returns false, and the full diff path
re-applies the source content. The next ingestion re-stamps the hash, so the fix costs at most one
extra reprocess after a manual edit rather than losing the edit permanently.

- `EntityInterface`: add a `default void setSourceHash(String)` no-op (mirrors the existing
  `default String getSourceHash()`); generated entity POJOs that declare `sourceHash` already
  override it with a real setter.
- `EntityRepository.patchCommonWithOptimisticLocking`: set `updated.setSourceHash(null)` before
  persisting, after `restorePatchSecrets` so no hook can restore it.

### Trade-off

Clearing on **every** PATCH also affects ingestion-driven PATCHes (e.g. lineage or bulk description
patches), disabling the fast-path for those entities on the next bulk run. That is a performance
cost, not a correctness one, and only until the next successful ingestion re-stamps the hash. If
that matters for high-volume PATCH connectors, a follow-up could restrict the clear to patches not
authored by the ingestion bot.

## Test

Adds `test_manualPatchClearsSourceHash_soSameHashReappliesTags` to `BulkSourceHashFastPathIT`:
creates a table with a tag and a `sourceHash`, PATCH-clears the tag (asserting the stored
`sourceHash` becomes null), then re-ingests the identical batch and asserts the tag is re-applied -
proving the fast-path does not skip after a manual PATCH edit.

Verified with `mvn spotless:apply` on the touched modules and `mvn compile` on
`openmetadata-spec` (clean). `openmetadata-service` compiles the changed files clean; the full module
build is blocked in this environment by a pre-existing relocated-ES-class error in
`ElasticSearchClient`/`SearchUtils` unrelated to this change.

The PR invalidates an entity’s stored source hash after a meaningful PATCH so the next bulk ingestion performs a full reconciliation instead of incorrectly taking the unchanged-hash fast path.
- Adds a common optional source-hash setter to the entity interface.
- Clears sourceHash during the shared PATCH lifecycle before persistence.
- Adds an integration test covering tag removal followed by same-hash bulk re-ingestion.
</details>

The PR appears safe to merge; the only identified issue is non-blocking redundant commentary in the regression test.

The shared PATCH flow clears and persists sourceHash for meaningful edits, allowing the next bulk ingestion to perform reconciliation, and no blocking correctness or security failure remains.

**Files Needing Attention:** openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BulkSourceHashFastPathIT.java
</details>

| Filename | Overview |
|----------|----------|
| openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java | Clears sourceHash in the common PATCH path; changed PATCHes persist the invalidation without creating a sourceHash-only version event. |
| openmetadata-spec/src/main/java/org/openmetadata/schema/EntityInterface.java | Adds an optional no-op setter so generic repository code can invalidate hashes on generated entities that expose the property. |
| openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BulkSourceHashFastPathIT.java | Covers PATCH invalidation and same-hash tag restoration, with minor redundant comments in the new test helpers. |

</details>

```mermaid
sequenceDiagram
  participant Client
  participant Patch as PATCH lifecycle
  participant Store
  participant Bulk as Bulk ingestion
  Client->>Patch: Edit entity metadata
  Patch->>Patch: Clear sourceHash
  Patch->>Store: Persist edited entity
  Bulk->>Store: Read stored entity
  Store-->>Bulk: sourceHash is null
  Bulk->>Bulk: Bypass unchanged-hash fast path
  Bulk->>Store: Reconcile source metadata and stamp hash
```
</details>

<sub>Reviews (1): Last reviewed commit: ["Fixes 20966: clear sourceHash on PATCH s..."](https://github.com/open-metadata/openmetadata/commit/9fb96f36b36e4e8a8d25b0d25fdfb7f49f407cc3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57911781)</sub>

- Knowledge Base — [Metadata entity lifecycle](https://app.greptile.com/getcollate/-/custom-context/knowledge-base/open-metadata/openmetadata/-/docs/metadata-entity-lifecycle.md)
</details>

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BulkSourceHashFastPathIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/BulkSourceHashFastPathIT.java
@@ -3,6 +3,7 @@ package org.openmetadata.it.tests;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -28,6 +29,7 @@ import org.openmetadata.schema.entity.data.Table;
 import org.openmetadata.schema.entity.services.DatabaseService;
 import org.openmetadata.schema.type.Column;
 import org.openmetadata.schema.type.ColumnDataType;
+import org.openmetadata.schema.type.TagLabel;
 import org.openmetadata.schema.type.api.BulkOperationResult;
 
 /**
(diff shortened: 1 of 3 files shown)


harshach:
The goal of sourceHash is to preserve whats in the Source which is snowflake, databricks etc.. if nothing chnaged in source there is no reason to write it again. If we start incorporating the patch , any edits in the UI by the user now impacts the sourceHash, then everytime we run ingestion we write all the objects as the source hash is changed on the OM side. 

Can you give us an example what PATCH are you running, is the PATCH coming from the ingestion or externally

zak-nuccio:
@harshach For context, this was to try and get around the issue of wanting to make manual changes to a table in OpenMetadata, but hitting the issue that sourceHash would make the ingestion think nothing was changed, and thus a later bulk load wouldn't overwrite the change (which we wanted to use for ingestion testing).

After some digging, we realised that sourceHash can actually just be edited via a PATCH call too, so instead we'll just be using a call to remove the hash so that later bulk ingestions see that the table should be reingested.

Closing

PR #32031, opened 2026-08-25
Fixes 32030: Look up Query entities by service-qualified FQN

larspettermadsstuen:
### Describe your changes:

Fixes #32030

Usage ingestion looked up Query entities by the MD5 of the SQL only. The server stores them as `service.hash`, so a second table with the same query hit 409 and the usage sink skipped attaching usage / users / usedBy.

This looks up `service.hash` the same way query-cost already does. If create still 409s (two tables racing), we GET again and keep writing the relations.

#
### Type of change:
- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### High-level design:

N/A — small change.

#
#### Use cases covered
- Same SQL on a second table finds the existing Query (`service.hash`) and still PUTs `/usage`
- Lookup never uses the bare hash
- Missing Query is created
- Create 409 retries GET by the qualified FQN, then usage still writes
- Non-409 create errors still raise
- Live: two tables, same SQL, share one Query whose FQN is `service.hash` and both have usage

#### Unit tests
- [x] I added unit tests for the new/changed logic.
- Files added/updated: `ingestion/tests/unit/ometa/test_query_mixin.py`
- Coverage on changed classes: 75% of `query_mixin.py` overall (`get_entity_queries` / `publish_query_cost` unchanged and not exercised here). Changed get-or-create / FQN lookup paths are covered.
- Coverage %: 75% on `query_mixin.py` (5 tests, all passing locally)

#### Backend integration tests
- [x] Not applicable (no backend API changes).

#### Ingestion integration tests
- [x] I added/updated ingestion integration tests for connector changes.
- Files added/updated: `ingestion/tests/integration/ometa/test_ometa_table_api.py` (`test_same_query_on_second_table_attaches_usage`)
- [ ] Not applicable (no ingestion changes).

#### Playwright (UI) tests
- [x] Not applicable (no UI changes).

#### Manual testing performed
1. `pytest ingestion/tests/unit/ometa/test_query_mixin.py` (5 passed)
2. Local stack: quickstart images `openmetadata/db:2.0.0` + `openmetadata/server:2.0.0` (skip Maven; no JDK). `GET /api/v1/system/version` → `2.0.0`
3. `pytest tests/integration/ometa/test_ometa_table_api.py::TestOMetaTableAPI::test_same_query_on_second_table_attaches_usage` (1 passed)

#
### UI screen recording / screenshots:

[...]

PR #32254, opened 2026-08-29
fix(ingestion): cache Query and User lookups during usage ingestion

### LINKED ISSUE #32210: Usage ingestion looks up the same Query and user once per table

Related to #32030. Once lookup uses `service.hash`, get-or-create actually finds the Query. It still does that GET for every table that referenced the same SQL.

Same for users: staging calls `get_by_name(User)` for every parsed query*table row, even when the username already showed up. Lifecycle later does `get_entity_reference(User)` for the same FQNs again.

On a usage run with a lot of repeated SQL / usernames, most of those GETs are duplicates. Not a correctness bug, just a lot of extra API calls from the Python ingestion client (`ometa`). The server is fine.

A bounded LRU in the Python ingestion client, for the length of a usage run:

- Queries keyed by `service.hash`. Cache after a successful GET, create, or 409 retry. Don't treat a cached miss as "this Query doesn't exist". Another table in the same run may have just created it.
- Users keyed by name, including misses (if they weren't in OM at the start of the run, they still won't be). Stage and `get_entity_reference(User)` should share that cache.

The client already caches some lookups. Query-cost on this same mixin has `@lru_cache` on hash lookup. Lineage uses `metadata.utils.lru_cache`. Usage get-or-create hits the server every time.

Don't reuse `__get_query_by_hash` from get-or-create as it caches `None`. Sharing one instance cache is fine if get-or-create ignores a stored miss and cost is allowed to remember one.

Leave `/usage`, `/users`, `/usedBy` writes as they are.

- Calling `__get_query_by_hash` from get-or-create: no, that's how you cache a miss and skip the second table.

- `cachetools`: we already have `metadata.utils.lru_cache`.

Deduping the relationship PUTs: useful, separate change.

### Pre-submission checklist

OmRajput17:
Fixes #32210

## Summary
- Wrap `_get_or_create_query` with a `SkipNoneLRUCache` keyed by query hash, so repeated tables referencing the same SQL skip the redundant GET/PUT. A miss is never cached, since another table in the same run may create the Query moments later.
- Add `get_cached_user_reference`, backed by an `LRUCache` keyed by username, shared between the staging `get_by_name(User)` call in `table_usage.py` and the lifecycle `get_entity_reference(User)` call in `metadata_usage.py`. Misses are cached too, since a user absent from OpenMetadata at the start of a run won't appear mid-run.

Both use the existing `metadata.utils.lru_cache` utilities as suggested in the issue, rather than adding a new dependency.

## Note on a deliberate tradeoff
The Query cache's key function excludes `self`, so the cache is shared process-wide across all `OMetaQueryMixin`/`OpenMetadata` instances rather than per-instance (unlike the existing `__get_query_by_hash`, which uses plain `@lru_cache` and is per-instance since `self` is part of the key). In practice a usage run typically uses a single `OpenMetadata` client, so this should be a non-issue, but flagging it explicitly in case a different approach is preferred.

## Tests
- Added `tests/unit/metadata/ingestion/ometa/test_usage_caching.py` covering: Query cache hit on repeat lookup, miss-not-cached behavior for Queries, User cache hit and miss-cached behavior, and independent caching per name.
- All existing tests in the touched files still pass: `test_metadata_usage.py` (21 passed), `test_user_mixin.py` (included above), `test_life_cycle_query_mixin.py` (4 passed).

<h2><a href="https://app.greptile.com/api/retrigger?id=58165395"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The cache implementation appears functionally safe, but the repository’s explicit testing requirement must be satisfied before merging.

<h2>Findings</h2>

1. P2&nbsp;**Tests Verify Mock Wiring** <a href="https://github.com/open-metadata/OpenMetadata/pull/32254#discussion_r4068521672">▶</a>

This PR reduces redundant OpenMetadata API calls during usage ingestion by adding bounded, client-scoped caches for Query entities and User references.
- Caches successful Query lookup and creation results while leaving misses uncached.
- Caches User-reference hits and misses across staging and lifecycle processing.
- Scopes both caches to individual OpenMetadata clients and limits each to 1,000 entries.
- Adds tests for hits, misses, service separation, and user-reference conversion.
</details>

```mermaid
flowchart LR
  U[Usage record] --> Q{Query cached?}
  Q -->|Yes| QE[Reuse Query entity]
  Q -->|No| QL[Lookup or create Query]
  QL -->|Found or created| QC[Cache per client]
  QL -->|Missing| QN[Return None without caching]
  U --> R{User reference cached?}
  R -->|Yes| UR[Reuse reference or cached miss]
  R -->|No| UL[Lookup User reference]
  UL --> UC[Cache result per client]
```
</details>

<sub>Reviews (4) · Last reviewed commit: ["Merge branch &#39;main&#39; into fix-usage-cachi..."](https://github.com/open-metadata/openmetadata/commit/f404f39a4d9d9605f4bf3405d8d30c8c6d42b68f)</sub>


OmRajput17:
Hi maintainers, the PR checklist is blocked on the linked issue (#32210) being added to the Shipping project, which I don't have permission to do myself. Could someone add it when you get a chance? Happy to push a follow-up commit to re-trigger the check once that's done.

mohittilala (changes requested):
Hey @OmRajput17, thanks for the contribution!

Heads up that #32031 landed on main and changes what this needs to key on. Query entities are addressed as `service.<md5>` rather than the bare hash, so `_get_or_create_query` now looks the entity up via `_qualified_query_fqn(...)`. The cache here still keys on `query_hash` alone, so once you merge main in, the same SQL seen under a second service would hit the first service's cached Query and attach usage to the wrong entity. Worth keying the cache on the same FQN the lookup uses.

A test for that case would be good too, two services sharing one SQL statement. Nothing covers it today, and it should fail against the current key.

The caches themselves look right, both bounded and scoped per client instance.

Local loop if it helps:

```
cd ingestion
make py_format         # applies ruff lint-fix + format
make py_format_check   # verifies
make static-checks     # basedpyright
python -m pytest tests/unit/metadata/ingestion/ometa/test_usage_caching.py
```

Cheers

Issue #26674, opened 2026-03-22
[BUG][Performance Issue] PATCH /api/v1/tables/{id} extremely slow (49–99s) during Database Service ingestion workflow


khoaihps:
**Affected module**
Backend + Ingestion Framework (Sink.Patch operations)

**Describe the bug**
*I will attached the full log below

[open-metadata-5cbd99c795-pk56h.log](https://github.com/user-attachments/files/26165360/open-metadata-5cbd99c795-pk56h.log)

When running a Database Service ingestion workflow, PATCH /api/v1/tables/{id} API calls become extremely slow (observed: 49–99 seconds per request), causing the overall workflow to degrade significantly.                                                                              

The slow request warning from the server shows that nearly all time is spent in "internal" processing — not database or search I/O:                                                                                                                                                                              

```                                                                                                                                                                                                                                                                                                               
WARN [2026-03-22 10:38:10,057] [dw-35304 - PATCH /api/v1/tables/3785924b-9e87-41a7-ba2f-547aeda02e89] o.o.s.m.RequestLatencyContext - Slow request detected - endpoint: v1/tables/3785924b-9e87-41a7-ba2f-547aeda02e89, total: 94243ms, db: 31ms (0%), search: 846ms (0%), internal: 93366ms (99%), dbOps: 113, searchOps: 1
...
WARN [2026-03-22 10:39:49,231] [dw-35136 - PATCH /api/v1/tables/7e63e8e2-4caf-48f2-ba12-eeb8a0b2a734] o.o.s.m.RequestLatencyContext - Slow request detected - endpoint: v1/tables/7e63e8e2-4caf-48f2-ba12-eeb8a0b2a734, total: 97759ms, db: 29ms (0%), search: 856ms (0%), internal: 96873ms (99%), dbOps: 96, searchOps: 1
```                                                                                                                                                                                                                                                                      

Additionally, for the same PATCH request (same thread, same table ID), the following log line appears twice, each time reporting 13,275 documents updated in OpenSearch:                                                                                                                                         

```                                                                                                                                                                                                                                                                                    
Successfully deleted columns from upstream lineage for index: all, updated: 13275                                                                                                                                                                                                                                
...
Successfully deleted columns from upstream lineage for index: all, updated: 13275                                                                                                                                                                                                                                
```
   
[...]

Issue #21536, opened 2025-06-03
Column lineages are not getting deleted when columns are deleted


mohittilala:
**Affected module**
Does it impact the UI, backend or Ingestion Framework?
- Backend

**Describe the bug**
- When columns are deleted or names are updated, old column are deleted from schema but their column-level lineage edges are still present (leading to unexpected errors in lineage ingestion).

**To Reproduce**
1. Ingest any view and their source table
2. Run lineage workflow and get the lineage created
3. Update the view definition, either delete any column or update the casing of column names (small to all caps, i.e., "NAME" -> "name")
4. Run the ingestion workflow again
5. Check the column level lineage (it will still have the column edges of deleted columns)
<img width="755" alt="Image" src="https://github.com/user-attachments/assets/190e7250-2ac7-478b-9b34-d209199c0c99" />

**Expected behavior**
5. Lineage between only existing columns should be present and others should be deleted while deleting columns

**Version:**
 - OS:
 - Python version: 
 - OpenMetadata version: 1.7.1
 - OpenMetadata Ingestion package version: [e.g. `openmetadata-ingestion[docker]==XYZ`]

**Additional context**

PR #32309, opened 2026-08-31
Fixes #32978: preserve table column lineage during version consolidation

### LINKED ISSUE #32978: [Bug] Table column lineage can be deleted during session consolidation
### Affected module

Backend: table column lineage and entity version consolidation.

### Describe the bug

A column added earlier in an editing session can lose its column lineage during a later description-only PATCH, even though the column remains in the table schema.

Version consolidation runs historical comparisons as well as the comparison against the persisted table. The older session baseline does not contain the newly added column, so a historical comparison reports it as deleted. `TableUpdater.handleColumnLineageUpdates` on main applies that comparison to stored and indexed lineage, causing an incorrect deletion.

### To reproduce

1. Create a source table with one column.
2. PATCH the table to add a second column.
3. Create column lineage from the added column to a downstream table.
4. In the same session, using the same user, PATCH only the added column's description.
5. Read the column lineage from the lineage API and the downstream search document.

Expected: the added column and its lineage remain present.

Actual: the historical consolidation pass can delete the lineage of the still-existing column.

### Regression evidence

`TableResourceIT.test_sessionAddedColumnKeepsLineageThroughConsolidation` passes with the persisted-baseline guard in #32309 on MySQL/Elasticsearch and PostgreSQL/OpenSearch. An isolated control that disables only that guard, restoring main's unconditional lineage-update behavior, fails the lineage-preservation assertion after the description-only PATCH.

### Related work

#32309 retains the correctness fix, nested change aggregation, combined search reconciliation, and bounded conflict recovery. Its optional lookup, index-scoping, and refresh-suppression optimizations are being removed.

Related to #26674 and #21536. The slow OpenSearch storage described in #26674 is a separate performance concern; this issue tracks the independently reproduced lineage correctness defect.

harshach:
### Describe your changes:

Fixes #32978

A table column added earlier in the same editing session can lose its lineage during a later description-only PATCH. Version consolidation compares the persisted table with the older session baseline, which can report that still-existing column as deleted. This PR limits lineage mutations to the comparison against the actual persisted table and combines the nested column changes into one reconciliation after commit.

Related to #26674; continuation of #26848 by @khoaihps. The original contributor commits and authorship are preserved.

The [reporter's follow-up](https://github.com/open-metadata/OpenMetadata/pull/32309#issuecomment-5522883108) identifies slow OpenSearch storage as the primary cause of the reported latency. This PR addresses the independent lineage correctness problem and redundant update work. End-to-end latency improvements have not been measured, and the original performance issue is not claimed fixed.

### Type of change:

- [x] Bug fix

### High-level design:

- Mark the comparison against the persisted entity in `EntityUpdater`, reset that state on transaction retry, and gate both database lineage changes and deferred search reconciliation on it. Historical comparisons still run for version consolidation.
- Collect nested column renames/deletions for the whole table. Update stored lineage inside the transaction and run one combined rename/delete search operation after a successful commit.
- Allow at most three reconciliation attempts for version conflicts, refreshing the search snapshot before each retry. Do not replay overlapping rename maps, which can undo successful writes. Report unresolved conflicts, shard failures, and timeouts.
- Preserve the existing column-match predicate, global search alias, and synchronous refresh behavior. Hash-based matching, index scoping, and refresh suppression are deferred from this PR.

No schema, API contract, or migration changes. Search reconciliation remains best effort after commit: persistent conflicts or search outages are logged; durable repair is outside this change.

#### Use cases covered

- Column deletion and case-only rename propagate to indexed lineage.
- A rename reverted within the same session uses the persisted state as its baseline.
- A description-only PATCH preserves lineage for a column added earlier in the session, in both the database and search.
- Nested column changes, bulk column deletion, and a rename plus deletion in the same PATCH reconcile correctly.
- Consecutive renames and rename-then-delete remain searchable with automatic refresh disabled in the test index.
- Conflicts from an unrefreshed downstream write are retried without losing the concurrent edit.
- Combined reconciliation updates both source and target FQNs and preserves unrelated mappings and metadata.

#### Unit tests

`ColumnLineageReconcilerTest`: **6 passed**, covering retry bounds, refreshed snapshots, no-match/success, shard and transport failures, and overlapping renames. The new `ColumnLineageReconciler` helper has **21/21 lines covered (100%, JaCoCo 0.8.13)**. Whole-class coverage for the existing repositories and search managers was not measured by this focused run.

`TableResourceIT` (7 cases) and `ColumnLineageReconciliationIT` (5 cases) exercise real database and search backends:

| Stack | Result |
| --- | --- |
| MySQL 8.3 / Elasticsearch 9.3 | 12 passed, 0 failures/errors/skips |
| PostgreSQL 15 / OpenSearch 3.4 | 12 passed, 0 failures/errors/skips |

Both suites ran against a clean build of the narrowed diff. The service artifact checksum was verified before and after execution.

Not applicable — no ingestion changes.

#### Playwright (UI) tests

#### Manual testing performed

- `mvn clean install -pl openmetadata-service,openmetadata-integration-tests -am -DskipTests -Dcheckstyle.skip`: all 11 reactor modules built successfully on Java 21.
- `mvn spotless:apply -pl openmetadata-service,openmetadata-integration-tests` and `pre-commit run`: passed; no remaining formatting diff.
- An isolated control disabled only the persisted-baseline guard in `TableRepository`, restoring main's unconditional lineage-update behavior. The session-added-column test failed at the lineage-preservation assertion after the description-only PATCH (1 test started, 1 expected failure, no aborted tests). The same test passed on both stacks with the guard enabled.

No end-to-end ingestion performance benchmark or full backend test-suite run was performed locally.

### Checklist:

- [x] The reproduced correctness issue is linked via `Fixes #32978`. `Related to #26674` is intentional: this change does not establish a fix for the reported slow-storage incident.
- [x] PR title follows `Fixes <issue-number>: <short explanation>` and describes the lineage correctness fix.
- [x] Non-obvious consolidation and retry behavior is documented in the code.
- [x] Regression tests cover the behavior being fixed.
- [x] Schema/migration and UI requirements: not applicable.


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -2325,13 +2326,28 @@ private void validateTableConstraints(Table table, Set<String> columnsBeingRemov
...
+      // replayed attempt would never re-enqueue the lineage flush and search cleanup is lost.
+      columnLineageUpdateDeferred = false;
+      pendingDeletedColumnFqns.clear();
+      pendingRenameColumnFqns.clear();
+    }
+
     @Override
     public void entitySpecificUpdate(boolean consolidatingChanges) {
+      pendingDeletedColumnFqns.clear();
+      pendingRenameColumnFqns.clear();
```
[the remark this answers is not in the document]
harshach: Confirmed and fixed in f1838f9edf — thanks, this was a real regression and a good catch.

Verified the mechanism in the code: `revert()` reassigns `original = previous` (EntityRepository.java:9017) before the final `updateInternal()` at :8837, so the last pass diffs **previous → updated**. The search index sits at **original**, so the final pass is the wrong baseline. Two failure modes, not just one:

1. **Net-zero** (your case): rename away and back within the session → final diff is empty → flush is a no-op → index stranded on the intermediate FQN.
2. **Rename chain**: `A→B` then `B→C` → final pass yields `{A→C}`, but the indexed documents reference `B`, so the update-by-query matches nothing and the index stays at `B` while the DB reaches `C`.

The fix accumulates from the **first** pass instead — `incrementalChange()` → `updateInternal(false)` at :8935 diffs original → updated, which is exactly the mutation the index needs, and is also the sole pass when consolidation does not apply. A pass counter incremented in `entitySpecificUpdate` (called once per pass at :9092 and :9118, so it covers the import path too) gates the accumulation, and `resetForRetryAttempt()` zeroes it so a deadlock replay starts from the baseline pass.

Regression test added: `TableResourceIT#test_revertedColumnRenameWithinSessionPropagatesInSearch`, which renames a column and reverts it within the consolidation window and asserts the index returns to the original FQN. It fails without this fix (empty flush) and passes with it.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -2325,13 +2326,28 @@ private void validateTableConstraints(Table table, Set<String> columnsBeingRemov
...
+    @Override
+    protected void resetForRetryAttempt() {
+      // The retry prologue clears deferredReactOperations; without resetting this guard the
+      // replayed attempt would never re-enqueue the lineage flush and search cleanup is lost.
+      columnLineageUpdateDeferred = false;
+      pendingDeletedColumnFqns.clear();
+      pendingRenameColumnFqns.clear();
+    }
+
     @Override
```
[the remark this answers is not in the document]
harshach: Confirmed and fixed in f1838f9edf. Your diagnosis is exactly right, including the observation that the previous code happened to be correct here because it deferred an op on *every* pass.

One note on the suggested remedy: I did not accumulate across the revert and intermediate passes, because that reintroduces the bug the per-pass clear was added to fix. Revert pass 1 diffs original → previous, so a column **added** earlier in the session surfaces as a **delete** — merging that into the flush would delete a live column's lineage from the index. (That hazard exists on `main` today, where each pass enqueued its own search op.)

Instead the flush now accumulates from the **first** pass only. That pass diffs original → updated, which is simultaneously:

- the correct baseline, since the index reflects `original`, not `previous`; and
- free of the inverted revert diff, so no spurious deletes.

It is also the sole pass when consolidation does not apply, so the non-consolidated path is unchanged. Net result is strictly better than both `main` (spurious deletes from the revert pass) and the previous state of this PR (stale FQN on net-zero/chained renames). A pass counter in `entitySpecificUpdate` gates it, reset in `resetForRetryAttempt()` for deadlock replays, and `TableResourceIT#test_revertedColumnRenameWithinSessionPropagatesInSearch` covers the net-zero path.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -10625,11 +10626,15 @@ public void updateColumns(
...
+      for (Column col : origColumns) {
+        origColumnByKey.put(columnLookupKey(col), col);
+      }
+
       // Carry forward the user generated metadata from existing columns to new columns
       for (Column updated : updatedColumns) {
-        // Find stored column matching name, data type and ordinal position
-        Column stored =
-            origColumns.stream().filter(c -> columnMatch.test(c, updated)).findAny().orElse(null);
+        Column stored = origColumnByKey.get(columnLookupKey(updated));
```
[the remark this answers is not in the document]
harshach: Applied in f1838f9edf — switched to `putIfAbsent`, which restores the first-match semantics of the `stream().filter(columnMatch).findAny()` this replaced. Agreed the exposure is narrow (the lists are per-nesting-level siblings, so a collision needs two siblings differing only in name case with the same dataType, which `DatabaseUtil.validateColumnNames` does not reject since its uniqueness check is case-sensitive), but the parity is free and removes the behavior delta.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -10625,11 +10626,15 @@ public void updateColumns(
       // Added columns are skipped by the existing-column loop below.
       storeColumnExtensions(entityId, addedColumns);
 
+      // Build a lookup map from origColumns to avoid O(n²) stream search per updated column
+      Map<ColumnKey, Column> origColumnByKey = new HashMap<>();
+      for (Column col : origColumns) {
+        origColumnByKey.put(columnLookupKey(col), col);
+      }
```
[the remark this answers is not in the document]
harshach: Applied in f1838f9edf — switched to `putIfAbsent`, which restores the first-match semantics of the `stream().filter(columnMatch).findAny()` this replaced. Agreed the exposure is narrow (the lists are per-nesting-level siblings, so a collision needs two siblings differing only in name case with the same dataType, which `DatabaseUtil.validateColumnNames` does not reject since its uniqueness check is case-sensitive), but the parity is free and removes the behavior delta.

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java
```diff
@@ -3377,6 +3377,185 @@ void test_columnLineageRename(TestNamespace ns) {
...
+    String query =
+        String.format(
+            "{\"size\":1,\"_source\":[\"upstreamLineage\"],\"query\":{\"term\":{\"_id\":\"%s\"}}}",
+            tableId);
+    Request request = new Request("POST", "/" + getTableSearchIndexName() + "/_search");
+    request.setJsonEntity(query);
+    Response response = searchClient.performRequest(request);
+    return new String(
+        response.getEntity().getContent().readAllBytes(), java.nio.charset.StandardCharsets.UTF_8);
+  }
```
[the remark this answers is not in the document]
harshach: Fixed in f1838f9edf — the response entity stream is now read inside try-with-resources. Worth fixing given the two new tests poll this helper on a 2s interval. Also dropped the fully-qualified `java.nio.charset.StandardCharsets` there since the class is already imported at the top of the file.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -2449,21 +2466,35 @@ protected void handleColumnLineageUpdates(
...
+        searchRepository
+            .getSearchClient()
+            .updateColumnsInUpstreamLineage(
+                COLUMN_LINEAGE_SEARCH_INDICES, new HashMap<>(pendingRenameColumnFqns));
+      }
+      if (!pendingDeletedColumnFqns.isEmpty()) {
+        searchRepository
+            .getSearchClient()
+            .deleteColumnsInUpstreamLineage(
+                COLUMN_LINEAGE_SEARCH_INDICES, List.copyOf(pendingDeletedColumnFqns));
```
TeddyCr: **With `refresh=false`, the second of these two update-by-queries can silently do nothing.**

Both calls use `buildLineageColumnsQuery`, which matches on `upstreamLineage.columns.toColumn` / `fromColumns`. A downstream document that references *both* a renamed and a deleted column of this table matches both queries. The rename UBQ writes that doc; the delete UBQ is issued immediately after, and with the refresh gone its search phase runs against a pre-refresh reader, gets a stale `seq_no`, and `conflicts=proceed` drops the doc.

Reproduced against Elasticsearch 9.3 (single shard, one doc, two back-to-back UBQs over it):

```
refresh-on-op1=false          <- this PR
  op1: updated 1 conflicts 0 failures []
  op2: updated 0 conflicts 1 failures []
  final doc: {'a': 2, 'b': 1}   <- second mutation lost

refresh-on-op1=true           <- current main
  op1: updated 1 conflicts 0 failures []
  op2: updated 1 conflicts 0 failures []
  final doc: {'a': 2, 'b': 2}
```

Note `failures: []`. The error branch in both engine managers only inspects `updateResponse.failures()`, which does not include version conflicts, so this leaves no trace beyond `LOG.info(... updated: 0)`. And it lands in the one case where the DB side has already committed — `lineageRepository.updateColumnLineage` above is synchronous inside the transaction — so it is precisely the DB/search divergence that #21536 fixed and that the `resetForRetryAttempt` override in this PR is guarding against elsewhere.

Trigger needs a rename and a delete in the same PATCH touching the same downstream doc, so it is narrow, but it is silent and it is not self-healing (see the second half of this comment thread on retries).

Options, roughly in order of how much I'd trust them:
- fold both into a single `_update_by_query` whose script takes both `columnUpdates` and `deletedFQNs` — no intra-flush ordering to get wrong, and one fewer round trip;
- keep `refresh=false` and re-issue when `updateResponse.versionConflicts() > 0` (bounded retry; both scripts are idempotent);
- `refresh=WAIT_FOR` on the first of the two only.

Either way, `versionConflicts()` belongs in that `LOG.info` line in both `ElasticSearchEntityManager` and `OpenSearchEntityManager` — as written, a conflict is indistinguishable from "nothing to clean up".

One related note while you're in here: because the flush now only runs from pass 1, the later consolidation passes no longer re-issue these ops. Pre-PR that accidentally repaired an index left stale by an earlier deferred op that threw (they are swallowed with `LOG.warn` in `runDeferredReactOperations`). There is now no retry path at all for a failed flush — probably worth a follow-up issue rather than something to solve here.

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchClient.java
```diff
@@ -40,6 +40,18 @@ public interface SearchClient
...
+  // the entire cluster via GLOBAL_SEARCH_ALIAS when only these indices can contain column lineage.
+  // Source of truth: LineageRepository.getChildrenNames — only these 8 entity types can be the
+  // "to" side of a column-lineage edge (validateLineageDetails drops the rest), and upstreamLineage
+  // is stored on the "to" entity's document. Other index mappings that declare upstreamLineage
+  // fields are shared boilerplate and can never hold column lineage. KEEP IN SYNC: adding an entity
+  // type to getChildrenNames requires adding its index here, or its cleanup is silently skipped.
+  String COLUMN_LINEAGE_SEARCH_INDICES =
+      "table_search_index,topic_search_index,container_search_index,"
+          + "dashboard_data_model_search_index,search_entity_search_index,"
+          + "api_endpoint_search_index,mlmodel_search_index,dashboard_search_index";
```
TeddyCr: **These literals bypass `IndexMapping` resolution, and the new `ignoreUnavailable(true)` makes any drift silent.**

This reads as though it goes through mapping resolution, since the callers pass it to `getIndexOrAliasName`. It doesn't: `SearchRepository.entityIndexMap` is keyed by entity type (`table`) and `aliasIndexMap` by alias (also `table`), so `"table_search_index"` misses both lookups and falls through to `resolveSingleAliasToken`'s last line, `clusterPrefix == null ? token : clusterPrefix + token`. It resolves to the right thing today only because each literal happens to equal the corresponding `IndexMapping.indexName`.

That gives the constant two keep-in-sync points rather than the one the comment documents: `getChildrenNames` **and** `indexMapping.json` (including the `elasticsearch/collate/indexMapping.json` overlay, which this module cannot see). And because a missing index is now `ignoreUnavailable(true)` instead of an exception, a drifted name degrades to a 200 with `updated: 0` rather than something anyone notices.

Passing the entity keys instead removes both problems — they hit `entityIndexMap` and come back through `mapping.getIndexName(clusterAlias)`, so renames and the Collate overlay are handled for free:

```java
String COLUMN_LINEAGE_SEARCH_INDICES =
    "table,topic,container,dashboardDataModel,searchIndex,apiEndpoint,mlmodel,dashboard";
```

(`getIndexOrAliasName` already splits on `,` and resolves each token independently, so this needs no other change.) The `KEEP IN SYNC` note against `getChildrenNames` stays accurate and becomes the only sync point.

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java
```diff
@@ -3739,6 +3740,288 @@ void test_columnLineageRename(TestNamespace ns) {
...
+        String.format(
+            "{\"size\":1,\"_source\":[\"upstreamLineage\"],\"query\":{\"term\":{\"_id\":\"%s\"}}}",
+            tableId);
+    Request request = new Request("POST", "/" + getTableSearchIndexName() + "/_search");
+    request.setJsonEntity(query);
+    Response response = searchClient.performRequest(request);
+    try (InputStream content = response.getEntity().getContent()) {
+      return new String(content.readAllBytes(), StandardCharsets.UTF_8);
+    }
+  }
```
TeddyCr, line 4231: **Returning the raw response body makes the negative assertions in all three tests satisfiable without the fix working.**

Every `!upstreamLineage.contains(fqn)` check is also satisfied by an empty hit set, by an error body, or by the target doc having been reindexed without `upstreamLineage` at all — and `.ignoreExceptions()` on the awaits removes the last guard. The preceding positive await proves the doc held the lineage at that moment, not that it still holds any lineage when the negative assertion passes, so a regression that drops `upstreamLineage` wholesale (or that breaks the index name and returns a 404 body) goes green.

Suggest parsing the response here and asserting the shape before the tests assert on FQNs: `hits.total.value == 1`, `upstreamLineage` present and non-empty, then return the `upstreamLineage` node. That keeps the three call sites as they are while making "the FQN is gone" mean "the entry was rewritten/removed" rather than "we found nothing to look at".

TeddyCr (commented):
Reviewed the deferred-flush machinery against `EntityUpdater`'s lifecycle and verified the search-side behaviour against a real Elasticsearch 9.3 node.

The parts I checked that hold up: the pass-1-only baseline is correct (`incrementalChange()` is the first `entitySpecificUpdate` in both the consolidating and non-consolidating paths, and `JsonUtils.extractPatchedFields` records the top-level `columns`, so `shouldCompare` never skips the column diff in pass 1); `resetForRetryAttempt` is wired correctly, since `restoreUpdaterState` calls it on every attempt including the first; the 8 index names match `LineageRepository.getChildrenNames` exactly; comma-separated index paths survive both clients (`EndpointBase.PATH_SAFE` includes `,`, and the OpenSearch client's `%2C` is decoded and split server-side — I confirmed a two-index `_update_by_query` matches both indices either way); and `ignore_unavailable=true` behaves as the description claims on `_update_by_query`.

Three things below that I think are worth addressing, one of them before merge.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -2449,21 +2466,35 @@ protected void handleColumnLineageUpdates(
...
-                searchRepository
-                    .getSearchClient()
-                    .deleteColumnsInUpstreamLineage(GLOBAL_SEARCH_ALIAS, deletedColumnsCopy));
+      // Only the first pass diffs against the persisted state that the search index reflects.
+      // Consolidation replays updateInternal() up to four times: the revert passes invert that
+      // diff, and the final pass rebases onto the pre-session version, so their FQNs match no
+      // indexed document — and a net-zero consolidation (rename away and back within the
+      // session) leaves them empty, stranding the intermediate FQN in the index. Within this
+      // first pass, addAll/putAll let every nested column level contribute to one flush.
+      boolean isSearchBaselinePass = updatePassCount <= 1;
```
manerow: This gate covers the search side, but the `lineageRepository.updateColumnLineage(...)` call above at :2460 still runs on all four passes, and pass 2 is the one with the inverted diff.

Concretely: `revert()` sets `updated = previous` and re-runs the diff, so a column **added earlier in the same session** shows up there as a *delete*, and `rewriteColumnMappings` strips it out of the relationship rows. Passes 3 and 4 never put it back, since `updateColumnLineage` only handles renames and deletes and never re-creates an edge. Column tags survive the same pass because pass 4 sees the column as "added" and re-applies them, which may be why this hasn't surfaced before.

It needs the second PATCH to include `columns` in its patch (otherwise `shouldCompare("columns")` is false on passes 1 to 3), same user, within the 10 minute window. Narrow, but ingestion hits that shape.

To be clear, this is pre-existing and not something you introduced. I'm raising it here because before this PR the search side made the same bogus delete, so both stores were wrong but agreed. Now the DB drops the edge while search keeps it until the next reindex.

Two options I can see: mirror the baseline gate onto that call, or leave it and file a follow-up. I don't have a strong preference and you have far more context on the revert path than I do. The only thing I'd gently push on is the description, which currently reads as though the #21536 invariant is fully restored. Would it be worth a line about this case either way?

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -2270,13 +2271,29 @@ private void validateTableConstraints(Table table, Set<String> columnsBeingRemov
...
+      // replayed attempt would never re-enqueue the lineage flush and search cleanup is lost.
+      columnLineageUpdateDeferred = false;
+      updatePassCount = 0;
+      pendingDeletedColumnFqns.clear();
+      pendingRenameColumnFqns.clear();
+    }
+
     @Override
     public void entitySpecificUpdate(boolean consolidatingChanges) {
+      updatePassCount++;
```
manerow: The logic is right, and the revert IT does guard it: an empty flush leaves `REVERT_COL` in the index and that test goes red. So this is a maintainability thought rather than a defect.

What made me pause is where the knowledge sits. "Pass 1 is the pass whose baseline matches the search index" is really a fact about `EntityRepository.flushUpdateBody`, specifically the order `incrementalChange()` (:9164), then `revert()` (:9227), then the consolidated pass. Here it becomes a counter in a subclass, and nothing on the `EntityRepository` side hints that reordering those would break `TableUpdater`. I can also see why `consolidatingChanges` wasn't usable, since it's false on passes 1 and 3, true on 2, false on 4.

One option, if you think it's worth the churn: hoist the flag up next to `resetForRetryAttempt` (`EntityRepository.java:8999`), set it in `flushUpdateBody`, restore it with the other snapshot fields, and expose something like `protected final boolean isPersistedBaselinePass()`. That would put the invariant next to the code that has to preserve it, and the next entity type that needs this cleanup would get it for free.

Happy to defer if you'd rather keep the blast radius small in a perf fix.

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchEntityManager.java
```diff
@@ -923,7 +923,11 @@ public void updateColumnsInUpstreamLineage(
...
                                   .params(params))
-                      .refresh(true));
+                      // A missing index must not abort cleanup of the remaining ones
+                      .ignoreUnavailable(true)
+                      // refresh=false: lineage cleanup needs no read-after-write consistency;
+                      // the index refresh interval covers it
+                      .refresh(false));
 
       LOG.info(
           "Successfully updated columns in upstream lineage for index: {}, updated: {}",
```
manerow: Adding to TeddyCr's point about `versionConflicts()` on this line, a few things stack up here (same at :987, and OpenSearch :1001 / :1060):

1. The message logs "Successfully" before the failure check on :937, so a failed op still reports success.
2. `updated: 0` currently covers "nothing to clean up", "index missing" (now that `ignoreUnavailable(true)` is on), and "a conflict dropped the write".
3. Exceptions get swallowed twice, once in the catch here and again in `runDeferredReactOperations`.

That was more survivable before, because the old code re-issued the op on every consolidation pass, so a lost write often got another chance. With the single flush that redundancy is gone.

Would it be worth logging `versionConflicts()` as TeddyCr suggested, and maybe dropping to WARN when `updated == 0` but the input set wasn't empty? Feels like a cheap way to tell "this is fine" apart from "we can't tell", but I'll leave the call to you.

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java
```diff
@@ -3739,6 +3740,288 @@ void test_columnLineageRename(TestNamespace ns) {
...
+      Awaitility.await("Wait for column lineage to be indexed in search")
+          .atMost(Duration.ofSeconds(30))
+          .pollInterval(Duration.ofSeconds(2))
+          .ignoreExceptions()
+          .until(
+              () ->
+                  getUpstreamLineageFromIndex(searchClient, targetTable.getId().toString())
+                      .contains(originalColFqn));
+
+      // Rename the column, paired with a description edit. A case-only rename on its own is not
```
manerow: Minor, and only because test comments tend to get read as documentation later.

This says a case-only rename on its own isn't recorded as a field change, so `reactUpdate()` would treat it as a no-op and skip the flush. The CI log on this branch seems to show otherwise: `test_renamedColumnLineagePropagatesInSearch` does a case-only rename and nothing else, and the flush still runs (`Successfully updated columns in upstream lineage ... updated: 1`, parallel IT lane).

No objection to keeping the description edit, I'd just suggest reworking the stated reason if you agree the reading is off.

khoaihps:
Quick update from my side, as I was the one who originally reported and reproduced this issue: after further investigation, I found that the OSD disk backing my OpenSearch instance was extremely slow, which appears to have been the primary cause of the performance issue I observed.

This means that the code-level changes I initially suggested may not have been the main factor behind the issue. I still think there may be some room for optimization, but I’d suggest validating the actual performance impact of the changes in this MR under normal storage conditions before proceeding further.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
manerow: Following up on the version-conflict point above.

What changed since: the logging got reworked, so a version conflict now shows up as a WARN with the conflict count instead of being indistinguishable from "nothing to clean up". That's a real improvement, and I'm glad it's there. But it makes the problem visible rather than fixing it, and I think it's worth being explicit about that before merge.

**Why the write is still lost.** `_update_by_query` works in two phases: it searches for matching docs and notes each one's `_seq_no` (a version counter), then writes each one back with "only apply this if `_seq_no` is still what I saw". With `refresh=false`, a write is durable immediately but not *searchable* until the next refresh, which is 1s by default. So:

- op1 (rename) updates the downstream doc, bumping its `_seq_no` from 41 to 42. No refresh.
- op2 (delete) searches, still sees 41, and asks to write "only if 41". The real value is 42, so ES rejects it.
- Because we pass `conflicts=proceed`, that's not an error. It's counted in `version_conflicts` and skipped. And version conflicts don't appear in `failures`.

End state: the database is correct, the search doc still lists the deleted column, one WARN line, and nothing ever repairs it short of a reindex. With the old `refresh=true` this couldn't happen, because op1's refresh meant op2 saw 42.

**I suspect the exposure is wider than the rename-plus-delete case.** That pair is narrow, agreed. But `refresh=false` also stretches the window for *unrelated* requests, from "as long as op1's forced refresh took" to "up to a full refresh interval". A downstream table usually has several upstreams, so if two source tables get column changes in the same second during ingestion, their flushes race on the same doc and the loser is dropped the same silent way. I haven't reproduced that one, it just follows from the same mechanism.

**One more thing that stacks with it.** Before this PR the flush was queued on every consolidation pass, so 2 to 4 identical ops ran and a lost or failed one often got repaired by a later copy. Collapsing to a single flush is the right call, but it also removes that accidental safety net, and failures are swallowed both in the search client and in the deferred-op runner. So it's one attempt now, with no retry.

**Some options**, no strong preference from me:

- Merge the two calls into one. Both scripts walk the same structure and do independent work, so a single script taking both the rename map and the delete list would remove the ordering problem entirely and save a round trip.
- Re-issue once when `versionConflicts() > 0`. Both scripts are idempotent. Worth knowing that an instant retry can conflict again, since it also needs a refresh to see the new `_seq_no`, so it'd want `wait_for` or a short delay.
- Use `refresh=wait_for` on the first call only. That waits for the next scheduled refresh rather than forcing one, though it still holds the request thread for up to a second.

If none of those feel right for a perf PR, a linked follow-up issue describing the scenario would work for me too, along with one for the missing retry path. My worry is just that both currently exist only in this thread, and threads get resolved and forgotten. Happy to file them if that helps.

Last small thing: there's no test covering a rename and a delete in the same PATCH, which is why the six new ITs are green and this case still isn't exercised.

harshach: Fixed in 4012b286ba, with the refresh policy refined in d92d495a0a. Rename and delete now share one update-by-query, removing the intra-flush ordering problem. Both engines use `ColumnLineageReconciler` for up to three total attempts on version conflicts, with an explicit refresh before each retry. Overlapping rename maps are not replayed because swaps can undo successful writes.

A flush containing renames retains refresh so the next request can query the new FQN; deletion-only flushes do not force a refresh. Disabling refresh after renames also reproduced a second failure on Elasticsearch 9.3 and OpenSearch 3.4: a subsequent rename/delete can return zero matches and zero conflicts, so conflict retries never run.

Added the mixed rename/delete PATCH test and real-engine tests for stale downstream writes, consecutive changes, both mapping directions, and preserved unrelated lineage. Final local validation: 13 integration cases on MySQL/Elasticsearch and the same 13 on PostgreSQL/OpenSearch, plus six retry unit tests; all passed. See [the reconciliation tests](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/ColumnLineageReconciliationIT.java).

Durable repair after transport/shard failures, exhausted retries, or conflicts on overlapping rename maps remains follow-up work; this PR does not claim to solve those cases.

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchClient.java
harshach: Addressed in the current PR: the selector is derived from the entity-type keys in `ColumnLineageChildren`, and both engines resolve it through `getIndexOrAliasName`. It no longer duplicates physical index names, so mapped names and cluster prefixes are preserved. The shared registry also removes the separate list that could drift from lineage validation. Covered by `ColumnLineageSearchIndicesTest`.

[Current registry and selector](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-service/src/main/java/org/openmetadata/service/lineage/ColumnLineageChildren.java#L58).

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java
harshach: Addressed: `getUpstreamLineageFromIndex` parses the response, requires exactly one hit and a non-empty `upstreamLineage` array, and returns only that array. Empty/error responses and missing lineage now fail the check instead of satisfying the negative FQN assertions. The response stream is also closed with try-with-resources.

[Updated helper](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java#L4320). The seven related table lineage tests passed on both database/search stacks.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
harshach: Addressed: `isIndexBaselinePass()` now gates both the database `updateColumnLineage` call and the deferred search reconciliation. The inverted consolidation passes therefore cannot delete the database mapping for a still-live session-added column.

`test_sessionAddedColumnKeepsLineageThroughConsolidation` checks both database lineage and search and passed on MySQL/Elasticsearch and PostgreSQL/OpenSearch. [Current gate](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java#L2449).

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
harshach: Addressed: the invariant now lives in `EntityRepository.EntityUpdater` as `indexBaselinePass`, exposed through `protected final isIndexBaselinePass()`. The update orchestration clears it before the consolidation/revert comparisons, and updater-state restoration resets it for a transaction retry. `TableUpdater` consumes that flag instead of maintaining a subclass pass counter.

[Flag and accessor](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java#L8875), [retry restoration](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java#L9009). The rename-reversion and session-added-column integration cases passed on both stacks.

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchEntityManager.java
harshach: Addressed in the shared `logColumnLineageFlush`: failures are checked before success logging, unresolved version conflicts log a WARN with their count, and updated-document counts are separate from requested-FQN counts. Both engines also report update-by-query timeouts as failures and now retry ordinary version conflicts with an explicit refresh, up to three total attempts.

Zero matches intentionally remain DEBUG because a valid column change may have no downstream references. The message explicitly acknowledges the missing-index/selector ambiguity. Transport/shard failures and exhausted retries remain best effort, as documented in the reconciliation thread.

[Current logging](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-service/src/main/java/org/openmetadata/service/search/SearchUtils.java#L868).

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java
harshach: Corrected in 4012b286ba. Removed the claim that a case-only rename is a no-op: the rename-only integration case does exercise the deferred flush. The remaining comments explain the actual reasons for the description edits: making consolidation eligible in the revert test, and covering metadata plus nested lineage in the same PATCH. These cases passed on both stacks.

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchUtils.java
```diff
@@ -828,4 +828,66 @@ public static int getMaxExpansions(String query) {
...
+          String.join(", ", outcome.failureReasons()));
+    } else if (outcome.versionConflicts() > 0) {
+      LOG.warn(
+          "{} in upstream lineage for index {} hit {} version conflict(s); those documents kept "
+              + "their previous column FQNs and are not retried. Updated {} of {} requested FQNs.",
+          outcome.operation(),
+          outcome.indexName(),
+          outcome.versionConflicts(),
+          outcome.updatedDocuments(),
+          outcome.requestedFqnCount());
```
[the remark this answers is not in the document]
harshach: Fixed: both the conflict and success messages now report `{} document(s) updated for {} requested FQN(s)`, keeping the units distinct. The conflict message separately reports the unresolved version-conflict count.

[Current messages](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-service/src/main/java/org/openmetadata/service/search/SearchUtils.java#L877).

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchEntityManager.java
```diff
@@ -895,103 +895,83 @@ public void deleteByRangeAndTerm(
...
+            req.index(Entity.getSearchRepository().getIndexOrAliasName(indexName))
+                .query(buildLineageColumnsQuery(affectedColumns))
+                .conflicts(Conflicts.Proceed)
+                .script(
+                    s ->
+                        s.source(ss -> ss.scriptString(RECONCILE_COLUMN_LINEAGE_SCRIPT))
+                            .lang(ScriptLanguage.Painless)
+                            .params(params))
+                .ignoreUnavailable(true)
+                .refresh(true));
```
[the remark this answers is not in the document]
harshach: [d92d495a0a](https://github.com/open-metadata/OpenMetadata/commit/d92d495a0ae759b159d75f33d5f598e622ed1a62) removes the forced refresh for deletion-only updates in both engines. A flush containing renames retains refresh for correctness: the next request queries the newly renamed FQN, which otherwise may not exist in the search snapshot. That query can return `total=0` and `version_conflicts=0`, so the conflict retry path never runs.

Reproduced rename→rename and rename→delete on Elasticsearch 9.3 and OpenSearch 3.4 with automatic refresh disabled; removing refresh silently dropped the second change. Added tests for both sequences and `deletionDoesNotForceARefresh`, which checks the external shard refresh counter before the realtime GET as well as the resulting lineage. The final 13 integration cases passed on each stack, along with six retry unit tests.

The single combined flush, consolidation deduplication, and scoped index selection remain. No production workload latency benchmark was run; the deletion-only optimization is verified directly by the refresh counter. [Tests](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/ColumnLineageReconciliationIT.java).

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/ColumnLineageReconciliationIT.java
```diff
@@ -0,0 +1,142 @@
...
+class ColumnLineageReconciliationIT {
+
+  @Test
+  void sequentialChangesRemainSearchable(TestNamespace ns) throws Exception {
+    try (LineageIndex index = new LineageIndex(ns)) {
+      index.client.updateColumnsInUpstreamLineage(
+          index.name, new HashMap<>(Map.of("source.rename", "source.RENAME")));
+      index.client.deleteColumnsInUpstreamLineage(index.name, List.of("source.delete"));
+      index.client.updateColumnsInUpstreamLineage(
+          index.name, new HashMap<>(Map.of("source.RENAME", "source.Rename")));
```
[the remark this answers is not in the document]
harshach: Keeping `refresh_interval=-1` and avoiding a test-side refresh is intentional: it deterministically exercises two production operations before the next scheduled refresh. Adding refresh calls between operations would hide the bug under review.

With request refresh disabled, rename→rename and rename→delete of the new FQN reproduced zero matches and zero conflicts for the second operation on both engines. Conflict retries cannot recover a query that matched nothing. The dedicated `sequentialRenamesMatchTheNewFqn` and `deletionMatchesTheNewlyRenamedFqn` tests protect those cases.

Production now skips forced refresh for deletion-only updates and retains it for renames. `deletionDoesNotForceARefresh` independently verifies the external refresh counter, so the suite also protects the performance optimization. All six reconciliation cases and seven table lineage cases passed on each database/search stack at d92d495a0a. [Current tests](https://github.com/open-metadata/OpenMetadata/blob/d92d495a0ae759b159d75f33d5f598e622ed1a62/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/ColumnLineageReconciliationIT.java).

harshach:
Narrowed this PR in 7be9b807bdad0e81b0ad05fedad1b255957df0fd following the reporter's clarification about slow OpenSearch storage.

The PR retains the independently justified lineage fixes: apply mutations only against the persisted baseline, collect nested column changes, reconcile renames/deletions together after commit, and handle version conflicts with bounded retries and accurate failure reporting.

The optional optimizations are removed: column matching uses the original predicate scan; search cleanup uses the existing global alias; both engines retain synchronous refresh. The lookup-key and index-selection abstractions and their dedicated tests are removed as well.

Validation on the narrowed diff: 12 integration cases passed on MySQL 8.3/Elasticsearch 9.3, 12 passed on PostgreSQL 15/OpenSearch 3.4, and all 6 retry unit tests passed. The clean backend build, Spotless, and pre-commit checks passed. A control that disables only the persisted-baseline guard fails the session-added-column lineage-preservation test; the same test passes with the guard on both stacks.

The title and description now target the reproduced correctness bug, #32978. #26674 remains related; this PR makes no claim to resolve the storage-related latency or to deliver a measured end-to-end performance improvement.

The required Shipping metadata for #32978 is still pending because the current GitHub token has read-only project access. The corresponding values on the original tracking issue are Status: In Review / QA, Source: OSS, Priority: P1, Domain: Platform, Release: 2.1.0.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/TableRepository.java
```diff
@@ -2430,42 +2430,46 @@ private void updateTableConstraints(Table origTable, Table updatedTable, Operati
...
+            () -> flushColumnLineageSearchUpdates(deletedColumnFqns, renamedColumnFqns));
       }
     }
+
+    private void flushColumnLineageSearchUpdates(
+        List<String> deletedColumnFqns, HashMap<String, String> renamedColumnFqns) {
+      searchRepository
+          .getSearchClient()
+          .reconcileColumnsInUpstreamLineage(
+              GLOBAL_SEARCH_ALIAS, renamedColumnFqns, deletedColumnFqns);
```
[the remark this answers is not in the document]
harshach: Confirmed: this was intentional in 7be9b807bdad0e81b0ad05fedad1b255957df0fd. The PR now fixes #32978 (lineage loss during version consolidation), and its title and description explicitly defer index scoping, hash-based column matching, and refresh suppression. #26674 is only related; the reporter identified slow OpenSearch storage as the primary cause of that incident: https://github.com/open-metadata/OpenMetadata/pull/32309#issuecomment-5522883108.

The global alias and synchronous refresh policy match the existing baseline. The retained changes correct the persisted-state comparison, collect nested changes, combine rename/delete reconciliation, and handle conflicts. We make no measured end-to-end latency claim. The scope clarification and validation results are recorded here: https://github.com/open-metadata/OpenMetadata/pull/32309#issuecomment-5588478776. Resolving this scope-confirmation thread.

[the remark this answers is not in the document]
harshach: The current PR scope intentionally retains the global alias. The title now targets #32978, and the description explicitly states that index scoping is deferred. We removed that optional optimization when narrowing the PR to the independently reproduced lineage correctness defect.

The original latency attribution is also no longer a claim of this PR: the reporter found slow OpenSearch storage to be the primary cause (https://github.com/open-metadata/OpenMetadata/pull/32309#issuecomment-5522883108). The scope decision is documented at https://github.com/open-metadata/OpenMetadata/pull/32309#issuecomment-5588478776. No index-selector change is needed for the agreed scope; resolving this thread.

openmetadata-service/src/main/java/org/openmetadata/service/search/elasticsearch/ElasticSearchEntityManager.java
```diff
@@ -895,103 +895,81 @@ public void deleteByRangeAndTerm(
...
+        req ->
+            req.index(Entity.getSearchRepository().getIndexOrAliasName(indexName))
+                .query(buildLineageColumnsQuery(affectedColumns))
+                .conflicts(Conflicts.Proceed)
+                .script(
+                    s ->
+                        s.source(ss -> ss.scriptString(RECONCILE_COLUMN_LINEAGE_SCRIPT))
+                            .lang(ScriptLanguage.Painless)
+                            .params(params))
+                .refresh(true));
```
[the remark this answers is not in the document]
harshach: Agreed that deletion-only reconciliation can rely on background refresh; synchronous refresh is needed for subsequent queries against newly renamed FQNs. We deliberately deferred the deletion-only refresh optimization in 7be9b807bdad0e81b0ad05fedad1b255957df0fd and restored the existing baseline policy for this focused correctness fix.

The current title/body target #32978 and explicitly defer refresh suppression. Keeping refresh(true) is therefore intentional, and the PR no longer advertises a latency improvement. All 12 targeted integration cases passed on MySQL/Elasticsearch and all 12 passed on PostgreSQL/OpenSearch with this policy. The optimization can be evaluated separately; resolving this thread consistently with the agreed scope.

openmetadata-service/src/main/java/org/openmetadata/service/search/opensearch/OpenSearchEntityManager.java
```diff
@@ -958,113 +959,83 @@ public void deleteByRangeAndTerm(
...
+                .query(buildLineageColumnsQuery(affectedColumns))
+                .conflicts(Conflicts.Proceed)
+                .script(
+                    s ->
+                        s.inline(
+                            i ->
+                                i.lang(l -> l.builtin(BuiltinScriptLanguage.Painless))
+                                    .source(RECONCILE_COLUMN_LINEAGE_SCRIPT)
+                                    .params(params)))
+                .refresh(Refresh.True));
```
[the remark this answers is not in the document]
harshach: Agreed that deletion-only updates can use normal refresh intervals and the explicit conflict-retry refresh. That is a separate optimization, deliberately deferred when this PR was narrowed in 7be9b807bdad0e81b0ad05fedad1b255957df0fd. Refresh.True preserves the existing OpenSearch policy, consistently with Elasticsearch.

The current PR title/body target the lineage correctness bug #32978 and explicitly defer refresh suppression; there is no end-to-end performance-win claim to preserve. The narrowed code passed all 12 targeted integration cases on PostgreSQL/OpenSearch and all 12 on MySQL/Elasticsearch. Resolving this thread under the documented scope: https://github.com/open-metadata/OpenMetadata/pull/32309#issuecomment-5588478776.

harshach:
The Collate failure in [run 34251264490](https://github.com/open-metadata/openmetadata-collate/actions/runs/34251264490/job/102146001811) comes from the PR branch missing an upstream validation change.

- The run built Collate `78de89516e1980add5587ea27d2c4c11edeab784` with OpenMetadata PR head `7be9b807bdad0e81b0ad05fedad1b255957df0fd`.
- `CollateIngestionPipelineIT.testPipelineCreation_EmptySourceConfig_RejectsMissingType` expects HTTP 400 for `sourceConfig.config = {}`, but this branch returns HTTP 201. It is the single failure among 489 integration tests (6 skipped).
- [OpenMetadata #29566](https://github.com/open-metadata/OpenMetadata/pull/29566) added `validateSourceConfigHasType()` in commit `27063928f7f2847d56b55afda460fd1cb150cf66`.
- [Collate #6300](https://github.com/open-metadata/openmetadata-collate/pull/6300) changed the assertion to require that validation. Both changes merged on September 7.
- Git ancestry confirms the validation commit is in current OpenMetadata main (`69fd965d8271305060058a5b774a3ab895ff417d`) and absent from this PR head. Collate delegates preparation to the OpenMetadata repository, so the missing superclass validation explains the observed 201 response.

The next step is to sync this branch with main and rerun the Collate compatibility check. This failure requires no change to the lineage implementation or to the test expectation.

harshach:
Merged OpenMetadata main (`d96acdd095`) into this PR in `a5b7dfd4d0047e2b6d9ac102d551a8e87aefef51`, bringing in the source-config validation from #29566 that Collate's existing test requires. The merge was conflict-free and preserves the existing 13-file lineage diff.

Validation on the merged tree:
- Clean Java 21 backend reactor build: all 11 modules passed.
- `IngestionPipelineRepositoryTest` and `ColumnLineageReconcilerTest`: 48 tests passed.
- MySQL + Elasticsearch integration tests: all 4 passed, covering missing-type rejection on create/update/PATCH and lineage preservation during session consolidation.
- Spotless and the PR's pre-commit checks passed.

The push triggers a fresh Collate compatibility check against the updated commit. No additional Collate PR or test-expectation change is needed.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -10920,21 +10954,43 @@ public void updateColumns(
...
 
+    private static final class ColumnLineageChanges {
+      private final Set<String> deletedColumnFqns = new LinkedHashSet<>();
+      private final HashMap<String, String> renamedColumnFqns = new HashMap<>();
+
+      private void include(
+          List<Column> deletedColumns, HashMap<String, String> originalUpdatedColumnFqns) {
+        deletedColumns.stream().map(Column::getFullyQualifiedName).forEach(deletedColumnFqns::add);
+        renamedColumnFqns.putAll(originalUpdatedColumnFqns);
+      }
```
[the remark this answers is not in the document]
harshach: Confirmed on `a5b7dfd4d0`: the new integration test reproduced HTTP 500 for both PATCH and PUT, with `List.copyOf` throwing from `ColumnLineageChanges.deletedColumnFqns()`.

Fixed in `7b124010d033825d0025fcef0795db4f5a1d9547` by filtering null FQNs before accumulation. The regression test seeds a legacy column with no stored FQN, deletes it alongside a valid column with lineage, and checks persisted columns plus DB/search lineage cleanup.

Both API cases and three existing lineage regressions pass on MySQL/Elasticsearch and PostgreSQL/OpenSearch (5 tests per stack). The rebuild, Spotless, and pre-commit checks also pass.

manerow (approved):

PR #32310, opened 2026-08-31
Fixes #21054: Oracle query history fails with ORA-00933 on Oracle 11g

### LINKED ISSUE #21054: Oracle lineage ingestion fails due to an error ORA-00933: SQL command not properly ended
**Affected module**
Impacts Ingestion Framework (lineage build feature).

**Describe the bug**
Faced with failures during lineage ingestion from an Oracle RDBMS source in OpenMetadata.
The issues are related to SQL parsing errors and entity handling:
1. Query History Parsing Error (ORA-00933)
The following query on gv$sql fails:
> ORA-00933: SQL command not properly ended

Something with LIKE conditions, such as:
`SQL_FULLTEXT NOT LIKE '/* {"app": "OpenMetadata", %%} */%%'`
But i can launch it with SQL Developer and got some results:
SQLDev

There are another set of errors and warnings, like:
2. SQL Parsing Failure on CREATE VIEW statements
The ingestion fails to parse "CREATE OR REPLACE FORCE VIEW ... AS SELECT ... WITH READ ONLY" queries.
> ingesting lineage failed: An Identifier is expected, got Token[value: (] instead.

For example: DBA$OLAP_DIM_LEVEL_ATTRIBUTES

3. Stored Procedure Lineage Ingestion Conflict
> api request failure: Entity already exists

I think this occurs when a procedure or its lineage was previously ingested.
It seems there is not "retry" or "update" feature (or i can't see it, maybe it is an incremental mechanism?).

But the main root cause of not getting lineage for objects in №1.
I've tried several times - with creating a new user and granting rights only to needed schema and with giving a full SYSDBA access to OpenMetadata while creating ingestions.

**To Reproduce**
1) Create user and give access as per [documentation](https://docs.open-metadata.org/latest/connectors/database/oracle)
```sql
CREATE USER openmetadataoracle IDENTIFIED by openmetadataoracle;
GRANT CREATE SESSION TO openmetadataoracle;
GRANT SELECT_CATALOG_ROLE to openmetadataoracle;
--GRANT SELECT ANY TABLE TO openmetadataoracle;
--REVOKE SELECT ANY TABLE FROM openmetadataoracle;
BEGIN
  FOR t IN (
    SELECT table_name FROM all_tables WHERE owner = 'PLN_MRPS'
  ) LOOP
    EXECUTE IMMEDIATE 'GRANT SELECT ON DB_TEST_SCHEMA.' || t.table_name || ' TO openmetadataoracle';
  END LOOP;
END;
/
```
In this schema (Oracle Hyperion Planning application) there are no connected columns, so OpenMetadata should build it with lineage feature by itself.
2) Add ingestion to metadata, launch it (i have a partial success)
3) Add lineage ingestion, launch it (failed)
Statuses

**Expected behavior**
Success of partial success status for lineage.
All parsed objects with history from v$sql are connected.

**Version:**
 - OS: Ubuntu 22.04.3 LTS
 - Python version: 3.10.12
 - OpenMetadata version: 1.6.6
 - OpenMetadata Ingestion package version: 1.6.6

**Additional context**
Logs:
[Oracle_SYS_lineage_1.log](https://github.com/user-attachments/files/19973849/Oracle_SYS_lineage_1.log)
[Oracle_Schema_lineage_2.log](https://github.com/user-attachments/files/19973851/Oracle_Schema_lineage_2.log)

Khairajani:
Closes #21054 (item 1 of the three reported there).

## Problem

Oracle usage and lineage ingestion fail outright on Oracle 11g and earlier with:

```
ORA-00933: SQL command not properly ended
```

`ORACLE_QUERY_HISTORY_STATEMENT` ends with the row-limiting clause `OFFSET 0 ROWS FETCH NEXT {n} ROWS ONLY`, which Oracle only supports from **12.1** onwards. On older releases the parser rejects the entire statement, so no query history is read at all.

The failure is easy to misattribute because metadata ingestion and *Test Connection* both keep passing: the probe `TEST_QUERY_HISTORY` reads the same `gv$sql` view but uses `WHERE ROWNUM = 1`, so it has no row-limiting clause. Same view, same grants, one syntax difference — which is what points at the cause.

The reporter suspected the `%%` in the `LIKE` patterns. That is a red herring, see "Note on `%%`" below.

### Why it is this construct and not something else

Diagnosis was by elimination against the verbatim SQL in the reporter's attached log, since every other candidate in that statement raises a *different* Oracle error code:

| Candidate | Would raise |
|---|---|
| Bad date format model | ORA-01821 / 01830 / 01861 |
| Unknown column (`OBJECT_STATUS`, `COMMAND_TYPE`) | ORA-00904 |
| Missing view (`gv$sql`) | ORA-00942 |
| Unbalanced parens from `{filters}` | ORA-00907 |
| `%%` in a `LIKE` pattern | nothing — adjacent wildcards collapse |
| **Row-limiting clause on < 12.1** | **ORA-00933** ✅ |

## Fix

Order in an inline view and cap with `ROWNUM` — the top-N form every Oracle version parses, which the optimizer resolves to the same `COUNT STOPKEY` plan:

```sql
SELECT * FROM (
    SELECT ... FROM gv$sql WHERE OBJECT_STATUS = 'VALID' {filters} ...
    ORDER BY FIRST_LOAD_TIME DESC
)
WHERE ROWNUM <= {result_limit}
```

**Why not gate on `server_version_info >= (12,)`?** There is no Oracle version where the gate produces a better query, and it would add a dialect-init dependency inside `get_sql_statement()` — plus a probe that can itself fail under restricted grants — to fix what is purely a syntax error. The unconditional rewrite keeps this a single string change, which also makes it a clean backport.

This template is `sql_stmt` for **both** `OracleUsageSource` and `OracleLineageSource`, so usage ingestion on 11g is fixed by the same change.

### Note on `%%`

Left as is, deliberately. `python-oracledb` uses the `named` paramstyle, so SQLAlchemy never de-doubles `%%` and it reaches Oracle literally — but adjacent `LIKE` wildcards collapse (`%%` is equivalent to `%`), so it is inert. Collapsing it to a single `%` would also break if this template ever ran through a `pyformat` driver.

## Behaviour

- `ROWNUM <= n` returns at most `n` rows, exactly as `FETCH NEXT n ROWS ONLY` did, so `warn_if_query_log_truncated`'s `row_count >= result_limit` check still fires correctly.
- `resultLimit` has no `minimum` in the pipeline schemas, so `0` is valid input; `FETCH NEXT 0 ROWS ONLY` and `ROWNUM <= 0` both return zero rows, so that edge is unchanged too.
- `SELECT *` over the inline view preserves alias names and column order, so row access downstream is untouched.

## Testing

New `ingestion/tests/unit/source/database/oracle/test_queries.py` renders the statement through the real source classes' `sql_stmt` / `filters` for both usage and lineage, and asserts no `OFFSET` / `FETCH NEXT`, the `ROWNUM` cap, ordering before capping, balanced parens, projected aliases, and that a user-supplied `filterCondition` lands inside the inline view where the source columns are in scope.

**5 of the 7 tests fail when the fix is reverted**, so the coverage is real rather than vacuous.

```
7 passed in 0.02s
```

`ruff check` and `ruff format --check` are clean on both files.

Not verified against a live 11g instance — no Oracle 11g image is available for arm64. CI's `oracle_lineage_container` (`gvenzl/oracle-free:23-slim-faststart`) exercises `OracleLineageSource` end to end, so a regression on modern Oracle would be caught there; note there is no equivalent e2e for the usage path, which the unit tests cover instead.

## Scope

Only item 1 of issue #21054. The other two are already fixed on `main`, `1.13` and `2.0` (verified by content, since the release branches carry them under cherry-picked hashes):

- **Item 2** — `An Identifier is expected, got Token[value: (]` on `CREATE OR REPLACE FORCE VIEW`: resolved by the SQLGlot parser support (#24729).
- **Item 3** — `api request failure: Entity already exists`: resolved by #28757.

Both reached the reporter's 1.6.6 only via upgrade.

Separately worth noting for 11g users, and **not** addressed here: `{prefix}_USERS.oracle_maintained` is a 12.1+ column, and the view-definition queries (`ORACLE_VIEW_DEFINITIONS*`, `ORACLE_GET_ALL_VIEW_AND_MVIEW_NAMES`) all filter on it, so they will raise ORA-00904 on 11g. That is a distinct bug and likely explains the "partial success" the reporter saw on metadata ingestion.

The PR makes Oracle query-history extraction compatible with Oracle 11g by replacing the 12.1+ OFFSET/FETCH clause with an ordered inline view capped by ROWNUM.
- Preserves the query-history projections, filters, descending load-time ordering, and configured result limit.
- Adds unit coverage for both Oracle usage and lineage rendering, including custom filter placement and projected aliases.
</details>

The PR appears safe to merge, with no concrete changed-code regression identified.

The rewritten query preserves the production rendering path, source-column filter scope, projected row contract, ordering-before-limiting semantics, and configured result cap while removing the Oracle 11g-incompatible clause.
</details>

| Filename | Overview |
|----------|----------|
| ingestion/src/metadata/ingestion/source/database/oracle/queries.py | Replaces unsupported OFFSET/FETCH pagination with an ordered inline view and outer ROWNUM cap while preserving filters, aliases, and ordering. |
| ingestion/tests/unit/source/database/oracle/test_queries.py | Adds focused rendering tests for usage and lineage query compatibility, ordering, filtering, parentheses, and expected projections. |

</details>

```mermaid
flowchart LR
    Config[Usage or lineage configuration] --> Render[Render query-history SQL]
    Render --> Inner[Query gv$sql with filters and ORDER BY]
    Inner --> Cap[Apply outer ROWNUM result limit]
    Cap --> Rows[Consume aliased query-history rows]
    Rows --> Ingestion[Usage and lineage ingestion]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["Fixes #21054: Oracle query history fails..."](https://github.com/open-metadata/openmetadata/commit/a01949ce009a90aee21efd727fc02e874966ca6f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58697932)</sub>


Khairajani:
Needs a backport to `1.13` and `2.0` once merged — both branches still carry the `OFFSET 0 ROWS FETCH NEXT` clause and are therefore still broken on Oracle 11g.

mohittilala (approved):

PR #30426, opened 2026-07-23
fix(ingestion): resolve Airbyte source/destination config from public API for lineage

VolubleRain1:
Fixes #30428

Addresses the public-API portion of #26993.

## Problem
Airbyte pipeline lineage is not produced when the service uses the **public API** (`api/public/v1`), even for connector types that are otherwise supported (Postgres/MySQL/MSSQL).

Two causes, both in the source/destination resolution used by `get_source_table_details` / `get_destination_table_details`:

1. **Config field name.** The models only read `sourceName`/`destinationName` and `connectionConfiguration`. The public API returns the connector type under `sourceType`/`destinationType` and the config under **`configuration`**:
   ```
   GET /api/public/v1/sources/{id}      -> { sourceType: "postgres", configuration: {database, schema, ...} }
   GET /api/public/v1/destinations/{id} -> { destinationType: "postgres", configuration: {database, schema, ...} }
   ```
   So `connectionConfiguration` is `None` → `schema`/`database` resolve to `None` → lineage silently skipped.

2. **Connector-type identifier.** The support guard compares against internal-API **display names** (`AirbyteSource.POSTGRES = "Postgres"`, `MSSQL = "Microsoft SQL Server (MSSQL)"`), but the public API returns **slugs** (`postgres`, `mssql`, `mysql`). Even after (1), the guard fails and returns `None`.

## Fix
- `AirbyteSourceResponse`/`AirbyteDestinationResponse` now also accept `sourceType`/`destinationType` and `configuration`, and expose `resolved_type` / `resolved_configuration` that fall back across both API shapes.
- New `SOURCE_TYPE_LOOKUP` / `DESTINATION_TYPE_LOOKUP` maps in `constants.py` recognize both display names and public-API slugs.
- The lineage helpers use the resolved values.

Internal-API behavior is unchanged (display names still map).

## Testing
- Added `test_yield_pipeline_lineage_details_public_api` asserting lineage is produced from a public-API-shaped source/destination (mirrors the existing internal-API test).
- Existing internal-API lineage test still passes.

## Scope
This makes lineage work on the public API for the **already-supported** DB connector types. Adding **warehouse destinations (Snowflake/BigQuery/Redshift)** — the destination used in #26993 — is a larger enhancement (which destination configs to support) and is intentionally left out of this PR; noted in #26993.

<details open><summary><h3>Greptile Summary</h3></summary>

Implements Airbyte lineage resolution for public-API response shapes.
[...]

PR #29016, opened 2026-06-13
Fix missing lineage for Airbyte S3 and API sources (#28591)

PrateekKrishna:
### Describe your changes:

Fixes #28591

I worked on resolving a bug where upstream and downstream lineage edges were silently dropped for Airbyte pipelines using non-relational sources (API) or destinations (S3). 

Previously, the ingestion connector strictly validated connection types against a hardcoded list of relational databases (Postgres, MySQL, etc.) and hardcoded the `Table` entity class for all lineage requests. I updated the parsing logic to extract S3 bucket names and allow API endpoints to pass through safely, and refactored the lineage builder to dynamically map S3 assets to the OpenMetadata `Container` entity.

#
### Type of change:
- [x] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### High-level design:

* **`utils.py`**: Updated `get_source_table_details` and `get_destination_table_details` to safely allow `api` and `s3` connection types. Extracted the `s3_bucket_name` from the Airbyte destination config and mapped it to the schema field so it can be utilized downstream.
* **`metadata.py`**: Refactored the `yield_pipeline_lineage_details` method to dynamically assign OpenMetadata entity classes. S3 sources/destinations are mapped to the `Container` entity, while API endpoints default to `Table` to maintain existing stream logic.

#
#### Use cases covered
- Ingestion run for an Airbyte pipeline containing an API source and an S3 destination correctly extracts the configuration and maps the upstream/downstream lineage edges without throwing validation warnings.

#### Unit tests
- [x] I added unit tests for the new/changed logic.
- Files added/updated: `ingestion/tests/unit/topology/pipeline/test_airbyte.py` (Added `test_missing_lineage_api_to_s3_bug`).

#### Backend integration tests
- [ ] I added integration tests in `openmetadata-integration-tests/` for new/changed API endpoints.
- [x] Not applicable (no backend API changes).

#### Ingestion integration tests
- [ ] I added/updated ingestion integration tests for connector changes.
- [x] Not applicable (no ingestion changes).

#### Playwright (UI) tests
- [ ] I added Playwright E2E tests under `openmetadata-ui/.../ui/playwright/` for UI changes.
- [x] Not applicable (no UI changes).

#### Manual testing performed
[...]

PR #27270, opened 2026-04-11
Fivetran improvements

harshach:
### Describe your changes:

Fixes <issue-number>

I worked on ... because ...

#
### Type of change:

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

#
### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://docs.open-metadata.org/developers/contribute) document.
- [ ] My PR title is `Fixes <issue-number>: <short explanation>`
- [ ] I have commented on my code, particularly in hard-to-understand areas. 
- [ ] For JSON Schema changes: I updated the migration scripts or explained why it is not needed.

----
## Summary by Gitar

- **Refactored event parsing:**
  - Replaced monolithic `if-else` blocks in `parse_sync_events` with a dispatch-based `_EVENT_HANDLERS` map for cleaner maintenance.
- **Improved data handling:**
  - Updated logic for `write_to_table_start` and `write_to_table_end` to correctly maintain minimum and maximum timestamps across partitioned events.
- **Enhanced unit testing:**
  - Added comprehensive test cases to `test_fivetran.py` covering event grouping, malformed JSON handling, and logic validation for timestamps.

<sub>This will update automatically on new commits.</sub>

PR #32325, opened 2026-09-01
Fixes #28591: Airbyte lineage for S3 (container) and API entities

### LINKED ISSUE #28591: Airbyte Connector Missing Upstream/Downstream Lineage (apiServices, s3)
### Connector

Airbyte

### Feature area

Lineage

### Describe the bug

The Airbyte connector successfully discovers and ingests Airbyte pipelines, but lineage relationships are not being created.

Specifically, the following lineage edges are missing:

Airbyte → S3
API Source → Airbyte

The Airbyte pipelines themselves are visible in OpenMetadata, the expected upstream/downstream lineage is not registered.

No errors or warnings are observed during ingestion, including when running with debug logging enabled.

### To Reproduce

Storage Service Names are specified as s3 storage.
S3 buckets have openmetadata_storage_manifest.json defined with a pre-defined folder structure.
Airbyte pipelines have a flow from the get api call to the S3 bucket, specified folder

### Expected behavior

Lineage relationships should be automatically created between source systems, Airbyte pipelines, and destination systems.

```yaml

```

### Logs

```shell

```

### OS

_No response_

### Python version

3.11.15

### OpenMetadata version

1.12.8

_No response_

### Additional context

Airbyte version 1 8 0

### Pre-submission checklist

akashverma0786:
### Describe your changes:

Fixes #28591

The Airbyte connector produced **no lineage at all** for an `API source → S3 destination` flow, and — as the issue reports — logged nothing while doing so, even at debug level.

I traced this against a **real Airbyte 2.2.0 + MinIO + OpenMetadata stack** rather than from the code alone, which turned out to matter: the primary cause is only visible on a live public-API instance.

**Three defects, in order of impact:**

1. **Streams were never parsed — this is why it was silent.** The public API (`api/public/v1`, which `client.py:220-222` silently defaults to) returns streams under `configurations.streams`. `AirbyteConnectionModel` only read `syncCatalog.streams` (the *internal*-API shape), and `extra="ignore"` dropped the field. So `syncCatalog=None` → `streams=[]` → the per-stream loop never executed → **not one warning was emitted**. Regression from the public-API migration in #30426, which fixed `resolved_type`/`resolved_configuration` for sources/destinations but not the connection catalog.

2. **Lineage was hardcoded to `Table`.** `fqn.build(entity_type=Table)` / `EntityReference(type="table")`. S3 is a `container` under a storage service, so the destination could never resolve even once streams were parsed.

3. **Diagnostics printed `type: None`.** `source_name`/`destination_name` read `sourceName`/`destinationName`, which are absent on the public API. Now read via `resolved_type`.

The reporter's *"Storage Service Names are specified as s3 storage"* refers to `lineageInformation.storageServiceNames`, which already existed and which **kafkaconnect already consumes** — Airbyte just never read it.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

Entity resolution now runs **object-store → table → API** on *both* the source and destination side. The previous asymmetry (source: `table` only; destination: `container → table`) was itself a bug source.

**Reused existing patterns, did not invent:**
- `es_search_container_by_path()` for S3 → `container`, matching `gluepipeline` (`_resolve_s3_entities`), `databrickspipeline`, `quicksight` and `snowflake`. This is the repo majority: 7 connectors use path-based lookup; only `kafkaconnect` uses name-based `fqn.build(Container, …)`.
- `get_storage_service_names()` — already on `PipelineServiceSource`, already used by kafkaconnect.
- `es_search_from_fqn()` + `APICollection` (already in `fqn_build_registry`) for API entities.

**Schema change:** adds `lineageInformation.apiServiceNames`, alongside the existing `dbServiceNames` / `storageServiceNames` / `messagingServiceNames`. This lives in the shared `pipelineServiceMetadataPipeline.json`, so it applies to all pipeline connectors — there is no Airbyte-local home for it, consistent with its three siblings.

**API resolution is deliberately opt-in.** Airbyte API connectors expose no endpoint URL (a real `pokeapi` source config is just `{"pokemon_name": "ditto"}`), so the stream name is the only join key. That key is weak: of Airbyte's 56 destinations, ~29 are neither relational nor object stores (Kafka, MongoDB, Pinecone, ElasticSearch, `/dev/null`), and matching those on name alone would invent lineage to an unrelated API. Resolution therefore runs **only** when `apiServiceNames` is set, and **only** when the match is unambiguous. Wrong lineage in a catalog is worse than missing lineage.

**Alternative rejected:** PR #29016 took the `fqn.build(Container, service_name="*", container_name=bucket)` route. That is the minority pattern, resolves only to the *bucket* rather than the written prefix, and on an ES miss `fqn.py:430` returns a fabricated `"*.bucket"` string that is truthy and slips past the `if not fqn` guard.

**Migration:** not required. New optional field; an absent key deserializes to `None` → `[]` → the existing unscoped path. Precedent: `storageServiceNames` (#15699) and `messagingServiceNames` (#27270) both shipped without one.

**Backward compatibility:** relational source/destination lineage is unchanged — all 70 pre-existing Airbyte tests pass untouched.

#
#### Use cases covered
- Airbyte pipeline with an **API source and S3 destination** produces `apiCollection → (pipeline) → container` lineage
- Airbyte pipeline with an **S3 source and S3 destination** produces `container → (pipeline) → container`
- S3 destination resolves to the exact written prefix (`bucket + bucket_path + stream`), falling back to the bucket container when only the bucket is registered in a manifest
- An unsupported destination (Kafka) yields **no** edge rather than binding to a same-named API collection
- Existing Postgres/MySQL/MSSQL table lineage continues to work unchanged

#### Unit tests
- File added: `ingestion/tests/unit/source/pipeline/airbyte/test_lineage_storage.py` (22 tests)
- **92 pass** total (70 pre-existing Airbyte tests untouched + 22 new)
- Covers: public-API vs internal-API stream shapes; S3 path construction and normalisation (empty / `None` / nested / leading+trailing slash); the production `yield_pipeline_lineage_details` path; bucket fallback; ambiguity guard; the S3-source-must-not-match-API regression; the Kafka opt-in regression

- [ ] Not added — validated against a real stack instead (below).

#### Playwright (UI) tests

#### Manual testing performed

Real stack, no mocks: **Airbyte 2.2.0** (`abctl`/kind) + **MinIO** (S3-compatible API) + **OpenMetadata** (local Docker).

1. Created a real `pokeapi` source (live HTTP GETs against pokeapi.co) and a real `destination-s3` pointed at MinIO via `s3_endpoint`.
2. Ran a real sync — **job 1 succeeded**, 1 row / 28,345 bytes → `s3://om28591-airbyte-dest/api_data/pokemon/2026_08_31_….jsonl`.
3. **Before the fix:** ingestion reported `Workflow Success 100%, 0 errors, 0 warnings` and produced **`upstreamEdges: 0, downstreamEdges: 0`** — the issue reproduced exactly, including the silence.
4. **After the fix:**
   ```
   om28591-pokeapi.pokemon ---> pipeline:814d9a1c-… ---> om28591-minio-storage.om28591-airbyte-dest.api_data/pokemon
   ```
5. **Reverse flow**, also real: created a `source-s3` reading that bucket + a second `destination-s3` writing to `reverse_out/`. Sync **job 2 succeeded** (1 row / 28,664 bytes → `reverse_out/pokemon_raw/2026_09_01_….jsonl`). Ingestion produced:
   ```
   om28591-minio-storage.om28591-airbyte-dest ---> pipeline:565b1f25-… ---> …reverse_out/pokemon_raw
   ```
6. Verified entity IDs and FQNs against the API directly, not the UI, and confirmed the target container is the one holding the object Airbyte actually wrote.
7. Regression-checked pre-existing metadata: all services (3/2/3/11/4/1), 333 tables and 12 pre-existing apiEndpoints intact; nothing deleted.

**Not verified end-to-end:** `S3 → Airbyte → API` and `API → Airbyte → API`. Airbyte's API destinations (HubSpot, Salesforce, Customer IO, Google Sheets) all require live third-party credentials that were unavailable. Those two directions are unit-tested only — and are the reason API resolution is gated behind `apiServiceNames` rather than on by default.

#

Not applicable.

#
### Checklist:
- [x] My PR title is `Fixes #28591: <short explanation>`
- [x] My PR is linked to a GitHub issue via `Fixes #28591` above.
- [x] For JSON Schema changes: no migration needed — new optional field, absent key deserializes to `None`. Precedent: #15699, #27270.

- [x] I have added a test that covers the exact scenario we are fixing (`test_api_source_to_s3_yields_pipeline_to_container_edge`, and the `#28591` reference in the module docstring).

#
### Notes for reviewers

- **Airbyte would be the first connector in the repo to emit `apiCollection` lineage.** Nothing else does today (`api_service.py` / `rest/metadata.py` only *create* API entities). Worth a deliberate look.
- **Upgrade note:** re-ingesting after enabling `apiServiceNames` leaves the previous `pipeline → container` edge in place — OpenMetadata does not auto-retract superseded lineage. Users will want `overrideLineage: true` on the first run after upgrading.
- **Out of scope, spotted while here:** `metadata.py:106` sets `source_url_prefix = "https://cloud.airbyte.com"` whenever the client is `AirbyteCloudClient`, which OAuth client-credentials selects even for a self-hosted host — so a self-hosted pipeline's `sourceUrl` points at cloud.airbyte.com. Not touched here; happy to file separately.

<h2><a href="https://app.greptile.com/api/retrigger?id=64284871"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge, with one non-blocking diagnostics gap for partially resolved API connector streams.

<h2>Findings</h2>

1. P2&nbsp;**Partial failures suppress warnings** <a href="https://github.com/open-metadata/OpenMetadata/pull/32325#discussion_r4069550058">▶</a>

This PR expands Airbyte lineage resolution beyond relational tables and restores stream parsing for Airbyte’s public API.
- Parses public-API streams from `configurations.streams`.
- Resolves S3 containers, API collections, messaging topics, and search indexes with service scoping and ambiguity guards.
- Verifies table entities per configured service and adds connector-specific table-key handling.
- Extends lineage configuration with API and search service names and updates generated schemas.
- Adds unit and integration coverage for storage, API, ambiguity, and parsing behavior.
</details>

```mermaid
flowchart LR
  A[Airbyte connection] --> B[Parse resolved streams]
  B --> C{Source resolver}
  B --> D{Destination resolver}
  C --> E[Table]
  C --> F[Container]
  C --> G[Topic or SearchIndex]
  C --> H[API Collection]
  D --> I[Table]
  D --> J[Container]
  D --> K[Topic or SearchIndex]
  D --> L[API Collection]
  E --> M[Pipeline lineage edge]
  F --> M
  G --> M
  H --> M
  M --> I
  M --> J
  M --> K
  M --> L
  C -->|Unsupported or unmatched| N[Pipeline anchor]
  D -->|Unsupported or unmatched| N
```
</details>

<sub>Reviews (6) · Last reviewed commit: ["fix(airbyte): close the container ambigu..."](https://github.com/open-metadata/openmetadata/commit/8d70d0cc729f49acae9977c41633580346af25b9)</sub>


ingestion/src/metadata/ingestion/source/pipeline/airbyte/resolvers.py
```diff
@@ -0,0 +1,405 @@
...
+        return Topic
+
+    def _entity_names(self, stream: AirbyteStream, connection: Connection, direction: str) -> list[str]:
+        if direction == SOURCE:
+            return [stream.name]
+        pattern = connection.resolved_configuration.get(KAFKA_TOPIC_PATTERN_KEY)
+        rendered = render_stream_pattern(pattern, stream) if pattern else None
+        # The destination also normalises the rendered name, so the bare stream name is kept as
+        # a fallback rather than relying on the pattern alone.
+        return [name for name in (rendered, stream.name) if name]
```
Khairajani, line 234: Should we keep the bare `stream.name` fallback here once `topic_pattern` has rendered? If I'm reading this right, with something like `topic_pattern: "airbyte.{stream}"` and the rendered topic not ingested yet, the lookup could land on a same-named topic in the same messaging service (possibly the source topic in a Kafka → Kafka connection), which seems to go against the PR's "no edge rather than a guess" approach.

If the fallback is there because the destination normalises the rendered name, would it make sense to fall back to the normalised rendered name instead? Or keep `stream.name` only for when `render_stream_pattern` returns `None`? Happy to be corrected if I've missed a case.

ingestion/src/metadata/ingestion/source/pipeline/airbyte/resolvers.py
```diff
@@ -0,0 +1,405 @@
...
+    def _service_names(self, source: "AirbyteSource") -> list[str]:
+        return source.get_search_service_names()
+
+    def _entity_type(self) -> type[SearchIndex]:
+        return SearchIndex
+
+    def _entity_names(self, stream: AirbyteStream, connection: Connection, direction: str) -> list[str]:
+        if direction == SOURCE or not stream.namespace:
+            return [stream.name]
+        return [f"{stream.namespace}_{stream.name}", stream.name]
```
Khairajani, line 266: Similar question here: since destination-elasticsearch always writes `<namespace>_<stream>` when there's a namespace, do we need the bare `stream.name` fallback in that case? My concern is that for, say, MySQL `shop.users` → ES, if `shop_users` isn't ingested yet but an unrelated `users` index exists in the configured search service, the lineage might attach to that one.

Maybe return only `[f"{stream.namespace}_{stream.name}"]` when a namespace is set? Let me know if there's a case where the bare name is needed that I'm not seeing.

ingestion/src/metadata/ingestion/source/pipeline/airbyte/resolvers.py
akashverma0786: Good catch, we should keep he normalised rendered name as the only candidate.

ingestion/src/metadata/ingestion/source/pipeline/airbyte/resolvers.py
akashverma0786: You’re right about the risk; instead of removing the second candidate, I made the first one correct -> read namespaceDefinition/namespaceFormat, compute the real namespace, one candidate

Issue #30977, opened 2026-08-05
[Epic] Fix all frontend ESLint warnings — 0 warnings across the UI project


ShaileshParmar11:
## Epic: Fix all frontend ESLint warnings

Bring the entire `openmetadata-ui` project to **0 ESLint warnings / 0 errors** across `src/**` and `playwright/**` (~9,270 pre-existing warnings resolved), **and lock it in** so it can't regress.

Delivered as **one PR per rule-category**, each independently reviewable (instead of one 1,400-file PR).

### Approach (updated)
- Every change is **behavior-preserving** — real fixes where safe, documented `eslint-disable`-with-reason where a genuine fix would risk bugs.
- **Ratchet:** each PR not only clears its rule to zero but **promotes that rule `warn`→`error`** in `eslint.config.mjs`, so CI blocks any new violation instead of letting the backlog grow back.
- **Rebase-and-resweep:** `main` keeps moving, so each PR is rebased onto latest `main` and **re-sweeps the new violations `main` introduced** for its rule before promoting. Merged **one at a time** (not a rigid stack).
- **Large rules are chunked per-rule** — e.g. type-safety (~369) is split into 2a/2b/2c so each PR stays small and reviewable.

> ⚠️ **Merge-time lesson:** promoting a rule to `error` requires the **whole codebase** clean *at merge time*. The merge queue lints **changed files only**, so violations in files a PR didn't touch can land on `main` as errors (this happened after #30989 → fixed by follow-up #31948). Full-lint right before each merge.

### Sub-tasks (one per rule)
- [x] 1. safe-mechanical (sonarjs) + jsx-a11y — #30978 → **merged in #30989** (follow-up fix: #31948)
- [ ] 2. type-safety — #30979 — **split into per-rule chunks:**
  - [ ] 2a. `@typescript-eslint/no-use-before-define` (15) — #31951 → PR #31952
  - [ ] 2b. `@typescript-eslint/no-non-null-assertion` (173) — #31956 → PR #32018
  - [ ] 2c. `@typescript-eslint/no-explicit-any` (181) — #31957 → PR #32026
- [ ] 3. react/* — #30980 → **PR #32087** (fresh off main; old #30991 stays closed)
- [ ] 4. clear-text-protocols + secrets — #30981 (was PR #30992)
- [ ] 5. jest/no-disabled-tests — #30982 (was PR #30993)
- [ ] 6. no-duplicate-string — #30983 (was PR #30994)
[...]

PR #32270, opened 2026-08-31
[ESLint] i18n-aware no-duplicate-string rule + extract non-i18n duplicates (promote to error)

ShaileshParmar11:
## Summary

Clears `no-duplicate-string` to zero and enforces it at **error** — **without touching i18n keys** — via a custom, i18n-aware rule (rule 6 of the ESLint-cleanup epic #30977, sub-issue #30983).

## Approach

Stock `sonarjs/no-duplicate-string` flags i18n translation keys (`t('label.…')`), which must stay inline (repo convention + static i18n tooling read those keys from source). SonarJS can't context-ignore `t()` arguments, so:

- **New custom rule `openmetadata-i18n/no-duplicate-string`** (in `eslint-rules/`, with unit tests) wraps the sonarjs rule and ignores any literal that is the first arg of `t(...)` or begins with an i18n namespace (`label.`/`message.`/`server.`). Behaviour is otherwise identical to sonarjs.
- **Stock rule turned off; custom rule enforced at `error`.**
- The only real duplicates left are **non-i18n literals** (CSS classes, MIME types, selectors, route/id fragments) — extracted into named module-level constants across **58 files** (byte-identical values).
- **i18n-key extractions from the earlier revision were reverted** — `t('label.…')` stays inline everywhere.
- Fixtures/mock data scoped off (matching the existing test-file exemption).

## Verification

- ✅ `0` `openmetadata-i18n/no-duplicate-string` violations tree-wide; rule at `error`
- ✅ `0` total ESLint errors project-wide
- ✅ All eslint-rule unit tests pass (106), incl. the new rule's tests
- ✅ No new `tsc` errors vs `main`; full UI suite green (1304 suites / 15,056 tests)

Closes #30983

<details open><summary><h3>Greptile Summary</h3></summary>

Introduces an i18n-aware wrapper around SonarJS’s duplicate-string rule, promotes it to an ESLint error, and extracts repeated non-i18n literals into constants.
- Preserves inline translation keys used by static i18n tooling.
- Adds rule and configuration tests.
- Keeps the ML model dummy domain’s quoted fully qualified name intact.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

The PR appears safe to merge.

No blocking failure remains.
</details>

<details open><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
[...]

PR #32328, opened 2026-09-01
[ESLint] Clear i18next/no-literal-string regrowth + promote to error

### LINKED ISSUE #30985: [ESLint 7/10] i18next/no-literal-string
Part of epic #30977.

Turn rule off for test/mock files (config), t()-ify 40 production strings + 26 en-us keys. 2,282 warnings.

Delivered as PR 8 of 10 (stacked). Behavior-preserving; verified via ESLint before/after + tsc regression check.

ShaileshParmar11:
## Summary

Closes out rule 7 of the ESLint-cleanup epic (#30977, sub-issue #30985) — clears the **`i18next/no-literal-string`** regrowth and **promotes the rule to `error`** so hardcoded user-facing strings can't drift back in (the reason it regrew to 66 is that it was left at `warn` after the big i18n-wave PRs #30967/#30996).

## What changed

- **66 violations** → **0**:
  - **25 in mock/fixtures** → the rule is scoped **off** for `*.mock.*`, `mocks/**`, `test/**` (sample strings aren't user-visible copy — same principle as the other string rules).
  - **~33 genuine UI texts** across 10 components → wrapped in `t('…')`, reusing existing keys where present.
  - **8 decorative glyphs/units** (`✓`, `▸`, `⌘↵`, `Esc`, `·`, `ms`, `(A → Z)`) → not translatable, so each carries a documented one-line `eslint-disable` with a reason.
- **25 new keys** added to `en-us.json` and synced to **all 20 locale files** via `yarn i18n`. Per the prior i18n-wave precedent, these ship as English placeholders; real translation is a follow-up (CI verifies keys are *synced*, not *translated*).
- Rule promoted `warn` → `error`.

## Verification

- ✅ 0 `i18next/no-literal-string` violations; rule at `error`
- ✅ 0 total ESLint errors; `check-i18n` in sync
- ✅ Full UI suite green — **1371 suites / 15,490 tests**

Closes #30985


ShaileshParmar11:
Thanks — checked both Copilot comments against the real config with the rule at `error`; both are already clean (`eslint` exits 0):

- **KnowledgeGraph3DLegend `▸`** — the `eslint-disable-line` is load-bearing and correct: this rule attributes the JSXText violation to the comment's own line, so `eslint-disable-line` suppresses it. I verified that switching to `eslint-disable-next-line` (as suggested) actually *reintroduces* the violation, and removing the disable entirely produces one — so the current form is intentional and passes CI at `error`.
- **ExpressionCodeCell `{' · '}`** — `i18next/no-literal-string` ignores punctuation/whitespace-only literals (no word characters), so this separator is not flagged. Confirmed 0 violations on the file at `error`.

No changes needed; `eslint <both files>` reports 0 `i18next/no-literal-string` at `error`.

(The two comments on #32270 — the MlModel FQN P1 and the sonarjs-coupling note — were already fixed/addressed in a later commit on that PR.)

karanh37 (approved):

PR #32340, opened 2026-09-01
fix(1.13): restore unit and integration test stability

yan-3005:
## Why this PR exists

`1.13` CI has been red for a while, and not because of any one change. Backports had been landing **test code from `main` without the production code those tests assert on**, so every run surfaced something new. That is why this PR is large.

It does three things: delete tests that can never pass on `1.13`, stabilise genuinely flaky tests, and fix the real product bugs those tests uncovered.

---

## 1. Tests that assert features `1.13` doesn't have

Deleted rather than skipped, because there is nothing on this branch for them to assert.

| Test | Why it cannot pass here |
|---|---|
| `IndexMappingNestedFieldConsistencyTest.aiGovernanceProjectionFieldsMustBeMappedExplicitly` | 1.13's `llm_model` / `mcp_server` mappings have no `aiGovernance` block |
| `IndexMappingNestedFieldConsistencyTest.auditReportIndexFieldsMustBeMappedExplicitly` | the `auditReport` index does not exist on 1.13 at all |
| `SearchIndexFactoryTest.pageReindexFieldsIncludeHierarchyFields` | 1.13's `PAGE` reindex fields have no hierarchy fields |

## 2. Test-only stabilisation

No product behaviour changes in this group.

- Scope list assertions to the test's own service so a parallel lane's data cannot leak in.
- Wait for async work before asserting: bulk-created entities, workflow deployment, search indexing.
- Move assertions inside `Awaitility` polls instead of asserting once and hoping (`ColumnSearchIndexIT`, `RdfGlossaryGraphIT`).

Three of these change **what is covered**, so calling them out explicitly:

- `UserMetricsResourceIT` drops a "last activity was within 10 minutes" check. `last_activity` is a cluster-wide `MAX` unrelated to the test's own actions, so on a quiet lane that window was arbitrary. Structural assertions remain.
- `RecognizerFeedbackRepositoryTest` inserts through the DAO instead of `repository.create()`, to avoid a ChangeEvent that kicks off a competing async workflow. `create()` is no longer exercised by that test.
- `UserTestFactory` now looks users up by name rather than email, and distinguishes 404 from 409.

## 3. Product fixes backported from `main`

Each already exists upstream; `1.13` was missing it.

- **Stale cache after a write (#28902).** Guava's `LoadingCache` can finish a load *after* a writer invalidated that key, then reinstall the stale value. The fix is a per-key write epoch: the loader reads it when it starts and again when it finishes, and throws if a write happened in between, so the caller re-reads from the DB. **Both** caches are guarded, by-id and by-name. The by-name half matters because that loader also writes the *ID-keyed* Redis slot, so leaving it unguarded would let it re-poison the very entries the by-id guard protects.
- **Two-phase rename invalidation.** A reader can repopulate the cache with the old FQN between the pre-write eviction and the rename itself. Descendants are now captured and evicted a second time after the rewrite, at all six rename sites: classification, container, domain, glossary, glossary-term, tag.
- **Orphan test-suite delete guard** (`TestCaseRepository`): a test case whose suite was already removed in the same cascade no longer 500s.
- **Bulk async create/update** reports a 202 per accepted entity in `successRequest`, so callers can wait on them.
- **Lineage delete deadlock retry** (`LineageRepository`): concurrent lineage writes could deadlock on the relationship delete and surface to users as a 500.
- **`EXTERNAL_HANDLER` propagation** no longer emits a whitespace-only script. A script of `" "` is not empty, so it passed the emptiness gate and fired an `update-by-query` across every child index that did nothing. Same one-line fix as `main` (early `return` from that case).

## 4. Changes that are 1.13-only

**Reviewers: these are not backports.** They exist because `main` solves the same problem using infrastructure this branch does not have.

- **Search-index rewrites during a rename are reordered to run last** (`deferReactOperation`). `main` gets this ordering from a post-commit `DeferralScope`; 1.13 has no equivalent. Worth knowing: `@Transaction` on `EntityRepository` is inert on this branch (it is JDBI's SqlObject annotation on a plain class and nothing reflects on it), so every DAO call autocommits. This is an ordering fix, **not** a commit hook. If a deferred rewrite fails it is now handed to `SearchIndexRetryQueue` rather than only logged.
- **`DeadlockRetry` around the async entity delete** (`EntityResource.deleteByIdAsync`). Safe because `deleteInternal` -> `deleteChildren` removes each child through `cleanup`, which opens its own transaction, so a replay simply skips children already gone. `postDelete`, the search delete, the websocket notification and the limits invalidation all run *after* the retry returns.
- **`TableResourceIT` retries `addLineage` on deadlock inside the test.** `main` does not protect that insert path either, so adding 1.13-only production protection there would be divergence for its own sake. Deliberately test-side; the *delete* path is fixed in production (section 3).

---

## Verification

- Unit tests on the affected classes: **429 run, 0 failures** (`EntityRepositoryRestoreTest`, `SearchRepositoryBehaviorTest`, `SearchIndexFactoryTest`, `EntityCsvTest`, `EntityRepositoryCertificationTest`, `IndexMappingNestedFieldConsistencyTest`).
- `mvn spotless:check`: clean.
- Integration lanes: see the checks on the current head. The per-class run figures in this PR's earlier description predate the most recent commits and have been removed rather than left to look current.

## Notes for reviewers

- A green `maven-collate-ci` is **not** downstream coverage for this branch: that workflow checks out `openmetadata-collate` at `ref: main`, not `1.13`. Collate `1.13` and `main` were both grepped for every symbol this PR re-signatures, and there are no callers.
- `invalidateCacheForRenameCascade` changed from `void` to returning the affected descendants. It is used only inside `openmetadata-service`.


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -348,6 +350,39 @@ private record InheritanceCacheKey(String entityType, UUID entityId, String fiel
           CacheConfiguration.DEFAULT_ENTITY_CACHE_MAX_SIZE_BYTES,
           CacheConfiguration.DEFAULT_ENTITY_CACHE_TTL_SECONDS);
 
+  // A Guava load can finish after a writer invalidates the same key and reinstall the stale
+  // pre-write value. Writers advance this bounded epoch before eviction; ID loaders compare the
+  // value at the start and end of each load and reject any value that crossed a write.
+  private static final Cache<Pair<String, UUID>, AtomicLong> WRITE_EPOCH_BY_ID =
```
manerow, line 359: **Suggestion: port the other half of #28902, or this half can be undone.**

#28902 guards **two** caches: `CACHE_WITH_ID` and `CACHE_WITH_NAME`. This backport has the by-id half.

The catch is that `EntityLoaderWithName` (line 10053) writes Redis under *both* keys on a miss:

```java
cachedEntityDao.putByName(entityType, fqn, json);
cachedEntityDao.putBase(entityType, entity.getId(), json);   // <- the by-id slot
```

Since `EntityLoaderWithId` reads Redis first and returns a hit as-is, an unguarded by-name load can put a stale value straight into the slot this guard protects. The guard can't catch that, because the write finished *before* its own load started, so start epoch == end epoch and it looks clean.

Good news: it ports mechanically. I checked 1.13's `CachedEntityDao` and every method needed is already there (`putBase:83`, `putByName:111`, `deleteBase:252`, `deleteByName:261`). Roughly 60 lines:

- `WRITE_EPOCH_BY_NAME` + `readEpochByName`
- `bumpWriteEpoch(entityType, id)` -> `bumpWriteEpoch(entityType, id, fqn)`, bumping both
- the same start/end guard in `EntityLoaderWithName.load`, with `evictExternalCache(...)`
- gate that loader's `putByName`/`putBase` on `readEpochByName(fqnPair) == startEpoch`
- **a second catch at line 2169** (`CACHE_WITH_NAME.get`). Without it the new throw escapes as a 500 on a hot read path. `main` catches at both `1705` and `2324`.
- skip #28902's `NotFoundCache` branches (1.13 has no `NotFoundCache`)

One note: a comment elsewhere in the branch says #28902 depends on `NotFoundCache` from #28012. I don't think that holds for the core, `NotFoundCache` shows up in exactly one line of that diff, in an optional negative-cache branch that has nothing to do with the race guard.

Also worth extending the three epoch tests in `EntityRepositoryRestoreTest` to cover the name side. No integration test touches this mechanism on either branch, so those unit tests are its only coverage.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -3016,16 +3095,42 @@ public static void invalidateCacheForRenameCascade(String entityType, String old
...
+    return affected;
+  }
+
+  /**
+   * Re-invalidate descendants captured before a rename after the FQN update has completed.
+   *
+   * <p>A reader can otherwise repopulate L1 with the old DB row between the first invalidation and
+   * the bulk rename update, retaining stale FQNs until cache expiry.
+   */
+  public static void finishInvalidateCacheForRenameCascade(
```
manerow, line 3120: **Suggestion: two rename sites still don't call this.**

Nice helper, but only 4 of the 6 rename paths use it. These two still do phase 1 only, so renaming a glossary or a tag keeps the stale-descendant window the others now close:

- `GlossaryRepository.updateName` line 691, `invalidateCacheForRenameCascade(Entity.GLOSSARY_TERM, oldFqn)` with no follow-up
- `TagRepository` line 1159, same call with the return value discarded (that file isn't in this PR yet)

`main` does both, e.g. for tags:

```java
List<EntityDAO.EntityIdFqnPair> renamedTags =
    invalidateCacheForRenameCascade(Entity.TAG, oldFqn);
// ... after the FQN rewrite:
finishInvalidateCacheForRenameCascade(Entity.TAG, renamedTags);
```

With those two added, all six sites are covered and the description's "matches the main/2.0 two-phase pattern" becomes accurate.

openmetadata-service/src/main/java/org/openmetadata/service/search/SearchRepository.java
```diff
@@ -2207,6 +2207,7 @@ private void appendAddScript(
       FieldChange field,
       PropagationDescriptor desc,
       EntityInterface entity) {
+    int scriptStart = script.length();
```
manerow: **Suggestion: `main` fixes this in one line, worth matching it.**

The bug is real and well spotted, a script of `" "` isn't empty, so it passes the `!updates.getKey().isEmpty()` gate at line 1792 and fires an `update-by-query` across every child alias for nothing.

`main` handles it by returning out of the `EXTERNAL_HANDLER` case, so the trailing `append(" ")` is never reached:

```java
case EXTERNAL_HANDLER -> {
  // No-op: a dedicated handler (e.g. propagateCertificationTags) drives the cascade.
  return;
}
```

Same result, no extra state, and it keeps us aligned with `main` for future backports. Applies to all three methods (the cases at lines 2258, 2315, 2385); the `scriptStart` locals at 2210/2273/2330 then come out. The new `propagateInheritedFieldsToChildrenSkipsExternalHandlerOnlyChanges` test passes either way.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -2281,6 +2282,7 @@ public void updateNameAndParent(GlossaryTerm updated) {
         invalidateTerm(updated.getId());
         updateAssetIndexes(oldFqn, newFqn);
```
manerow: **Suggestion: these two should be deferred as well.**

The `deferReactOperation` change is right, but `GlossaryTermRepository` still calls `updateAssetIndexes` inline here and at line 2345. `main` defers both (inside its own `updateAssetIndexes` body, via `deferIfFlushScopeActive`), so leaving these inline keeps the exact ordering problem the defer exists to fix. Both call sites are inside the updater, so `deferReactOperation` is already in scope.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryRepository.java
```diff
@@ -709,11 +709,9 @@ public void updateName(Glossary updated) {
...
                   condition, oldFqn, newFqn, PolicyConditionUpdater.TAG_FUNCTIONS));
 
-      // Cascade rename into the search index — child term FQNs and the embedded glossary denorm
-      // (glossary.name / glossary.fullyQualifiedName) must reflect the new name. This used to be
-      // driven by entityRelationshipReindex, which has no caller since PR #19550, so the call has
-      // to happen inline here (mirroring Domain / Classification / GlossaryTerm renames).
-      updateAssetIndexes(original, updated);
+      // Cascade search writes must run after this transaction commits. Rebuilding child term
+      // documents here would re-read the old committed rows and leave their glossary denorm stale.
+      deferReactOperation(() -> updateAssetIndexes(original, updated));
```
manerow, line 720: **Suggestion: keep the defer, but reword the comment and don't lose failures.**

To be clear, the defer itself is correct and should stay. I traced the history (`3a674af59a` -> `f13a0f8750` -> this) and dropping it would re-break `GlossaryResourceIT.test_renameGlossaryPropagatesToChildTermSearchIndex`.

Two refinements:

**1. The comment says "after this transaction commits", which isn't quite what happens on 1.13.** `@Transaction` on `EntityRepository` is JDBI's SqlObject annotation on a plain class, and nothing reflects on it, so every DAO call autocommits (the class javadoc at `:4470-4476` says as much). What this actually does is reorder the ES rewrite to run after all DB writes and after the phase-2 evict, which is the same *ordering* `main` gets from its post-commit `DeferralScope`. Suggest saying that instead, so nobody later assumes 1.13 has commit hooks it doesn't.

**2. `runDeferredReactOperations` swallows exceptions into a `LOG.warn` (`:8043-8051`).** Not a regression versus today, but `main` enqueues failures to a retry outbox, and 1.13 already has `SearchIndexRetryQueue.enqueue(entityId, entityFqn, failureReason)` (`SearchIndexRetryQueue.java:37`). Wrapping each deferred lambda to enqueue on failure would mean a failed cascade is recoverable rather than just logged.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/EntityRepository.java
```diff
@@ -2607,21 +2645,60 @@ public final ResultList<T> listEntityHistoryByTimestamp(
...
     int total = getVersionCountCached(tableName, startTs, endTs, entityType);
     return historyPageResult(entities, page, hasMoreInCurrentDirection, total);
   }
 
+  /**
+   * Hydrate a history page in bulk, falling back to one snapshot at a time when a concurrent hard
+   * delete removes a current relationship needed by another row in the page. Version JSON stores
+   * the historical references, so retaining that snapshot is safer than failing the entire page.
+   */
+  private void hydrateHistoryPage(List<String> jsons, List<T> entities) {
```
manerow: **Suggestion: consider dropping this fallback.**

`main` and `2.0` both have `listEntityHistoryByTimestamp` with the plain `setFieldsInBulk` + `hydrateHistoryEntities` and no fallback, and they're green, which suggests the trigger is the parallel IT suite deleting shared fixtures rather than anything a user would hit.

The reasoning in the javadoc is genuinely interesting (a *history* endpoint arguably shouldn't hydrate from *current* relationships at all), but that feels like a `main` design change rather than a 1.13 patch. As written it costs O(n) round-trips on the failure path and turns an integrity error into a `DEBUG` line plus partly-hydrated output.

If it comes out, the `pageJsons` reorder at 2657 reverts with it, and these two tests go too: `EntityRepositoryRestoreTest:318` and `:374`.

Worth noting this is separate from the deadlock currently failing in that same test class: that one is a MySQL deadlock on `TagUsageDAO.deleteTagsByPrefixAndTarget` (`Certification.%`) inside `createVersionedEntities`, which this fallback doesn't address either way.

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/TableResourceIT.java
```diff
@@ -3319,6 +3331,22 @@ void test_columnLineageRename(TestNamespace ns) {
     assertTrue(lineageJson.contains("columnsLineage"));
   }
 
+  private String addLineageWithDeadlockRetry(OpenMetadataClient client, AddLineage addLineage) {
```
manerow, line 3334: **Suggestion: keep this, and take `main`'s production fix for the delete path alongside it.**

Correcting my earlier version of this comment: I said backporting `main`'s `LineageRepository` `DeadlockRetry` would let this helper come out. That was wrong. `main` has exactly one `DeadlockRetry` in that file, on `deleteLineageRelationshipWithRetry` (the **delete** path). `main`'s `addLineage` insert has no retry either, so this test is retrying a path neither branch protects.

So the split is:

- **This helper stays.** Adding 1.13-only protection to the insert path would be exactly the kind of divergence the rest of this PR is removing.
- **The delete path is a genuine gap**, and 1.13 has zero `DeadlockRetry` in `LineageRepository` where `main` has one. I've pushed that backport as `fix(1.13): retry deadlocks on lineage relationship delete` — it routes the three `relationshipDAO().delete(...)` sites through a retrying helper, matching `main`.

Worth a line in the PR description noting the insert path is deliberately test-side for now.

openmetadata-service/src/main/java/org/openmetadata/service/resources/EntityResource.java
```diff
@@ -765,7 +766,8 @@ public Response deleteByIdAsync(
             () -> {
               try {
                 DeleteResponse<T> deleteResponse =
-                    repository.delete(userName, id, recursive, hardDelete);
+                    DeadlockRetry.execute(
```
manerow, line 769: **No code change needed, just flagging that the reasoning in the description is off.**

The description justifies this by `delete` being `@Transaction`-annotated. On 1.13 that annotation is inert (JDBI SqlObject annotation, plain class, nothing reflects on it), so `delete` isn't actually atomic and a retry isn't replaying a rolled-back transaction.

It's still safe, for a different reason worth writing down: `deleteInternal` -> `deleteChildren` removes each child through `cleanup`, which opens its own `Entity.getJdbi().inTransaction(...)` (`:4564-4567`). A replay after a mid-cascade deadlock just skips the children already gone. And `postDelete`, `deleteFromSearch`, the websocket notify and `limits.invalidateCache` all run after the retry returns, so they aren't replayed.

One small residual: a duplicate `ENTITY_DELETED` change event if the deadlock hits after the parent's event insert. Probably acceptable, but worth a line in the description.

Also: "limits deadlock retries to recognized transient database failures" describes `DeadlockRetry`'s existing behaviour, that file isn't in the diff. The real change is applying it here.

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/UserMetricsResourceIT.java
```diff
@@ -205,71 +205,56 @@ private void verifyMetricsStructure(String response) {
...
-
-    String userName = ns.prefix("metricsuser");
-    String email = "metricsuser_" + ns.shortPrefix() + "@test.openmetadata.org";
-    CreateUser createUser = new CreateUser().withName(userName).withEmail(email).withIsBot(false);
-
     UserService usersApi = adminClient.users();
-    User newUser = usersApi.create(createUser);
-    log.info("Created new user: {}", newUser.getName());
+    List<User> createdUsers = new ArrayList<>();
+    int initialTotalUsers = (Integer) getUserMetrics().get("total_users");
```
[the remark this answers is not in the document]
manerow: **Correct, and deliberately kept.**

It's `main`'s assertion verbatim: swapping it would trade one 1.13-only divergence for another, which is what this commit set out to remove.

The variant that actually failed CI is gone. It created one user and gated on `total >` inside a 30s Awaitility poll; since the metric is computed live per request, waiting only widened the churn window. Now it's a single-shot read with three users of headroom, and the `>=` form hasn't failed.

If it does, the fix is invariant-only assertions (`bot_users <= total_users`), on `main` first.

mohityadav766 (approved):

PR #32352, opened 2026-09-01
Fixes #32347: emit pipeline observability once per DAG, from the newest run

### LINKED ISSUE #32347: Airflow connector: observability data is re-published for every DAG, making metadata ingestion take days on large deployments
### Connector

Airflow

### Feature area

Metadata ingestion

### Describe the bug

Running Airflow (v3) metadata ingestion against a deployment with ~380 DAGs, a single run did not finish after 10+ days. The cause is the pipeline observability step.

While processing a DAG, the connector caches one observability entry per DAG run (up to `numberOfStatus` per DAG). At the observability step of every *later* DAG, it iterates the entire cache — skipping only the DAG currently being processed — rebuilds observability records for all earlier DAGs' runs, and re-attaches them to those DAGs' tables. For each table in the resulting map it then looks the table up and issues a PUT to the table's pipeline observability endpoint, with a payload that keeps growing as the run progresses.

The work therefore grows quadratically with the number of DAGs. With 380 DAGs and 10 runs each, the cache reaches ~3,800 entries and the same observability data is fetched, rebuilt and written back hundreds of times over — on the order of hundreds of thousands of redundant API calls in a single run.

The cache is never pruned, and each entry also holds the full parsed DAG, so memory grows with DAGs × runs.

Precondition: only DAGs whose inlets/outlets resolve to existing tables produce cache entries with tables attached, so this shows up on deployments where Airflow lineage resolves and lineage ingestion is enabled.

Separately, in the same path, DAG run history is queried twice per DAG — once for pipeline status and again for lineage — doubling the queries against the Airflow metadata database.

### To Reproduce

1. Point an Airflow (v3) metadata ingestion at a deployment with several hundred DAGs (~380 in this case), default `numberOfStatus` (10).
2. Enable lineage, with DAG inlets/outlets resolving to tables that exist in OpenMetadata.
3. Run metadata ingestion.
4. Per-DAG processing time and the number of writes to the table pipeline observability endpoint grow steadily as the run progresses; the run does not complete (still running after 10+ days).

### Expected behavior

Observability data for a table should be collected and written once per ingestion run, not re-published on every later DAG iteration.

Possible directions:
1. Emit only the current DAG's observability in the per-DAG step and drop the replay of earlier DAGs, or
2. If the goal is to aggregate several DAGs writing the same table, do that once at the end of the run rather than per DAG.
3. Either way, clear entries once they have been written so the cache does not grow unbounded.

```yaml

```

### Logs

```shell

```

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

1.13.3

openmetadata-ingestion==1.13.3.1

### Additional context

Verified on `openmetadata-ingestion` 1.13.3.1, and the same pattern is still present on 2.0.0.0.

Related but a different component: #22351 covers status/lineage performance in the Airflow Lineage Backend plugin. This report is about the Airflow connector's own metadata ingestion.

Secondary observation, possibly a separate issue: two runs of the same ingestion pipeline overlapped. After the jobs were terminated on the backend, the UI kept showing the pipeline as running — the status was never finalised for the killed runs.

Noted only to save triage time, not part of this report: the `Wild error yielding dag (...) - 'NoneType' object is not subscriptable` warning seen on the same DAGs before upgrading came from the older client reading only the uncompressed serialized-DAG column, which is empty on Airflow 3 when compressed serialization is used. 1.13.3 added a fallback that decompresses it, and the warning stopped after the upgrade.

### Pre-submission checklist

ulixius9:
### Describe your changes:

Fixes #32347

The observability stage cached one entry per `(dag, run)` and, for every later DAG, replayed the **entire** cache — rebuilding and re-writing observability for all previously seen DAGs' tables. Work grew as `O(dags × cumulative tables)`; the reported 563-DAG deployment issued ~2M table lookups plus ~2M PUTs and had not finished after 13 days. The replay bought nothing: the server upserts under `table.pipelineObservability.<pipelineFqn>`, one row per `(table, pipeline)`, so a record is fully determined by the DAG that produced it and stays valid once written. I dropped `observability_cache` from the airflow and dbtcloud connectors, emit only the current DAG's tables, and reuse the `Table` entities the lineage stage already resolved via a bounded cache on `PipelineServiceSource` — which also removes the second `get_by_name` per table.

Two correctness bugs fall out of the same change: **(1)** all runs of a DAG share a pipeline FQN and so upsert one row applied front-to-back, and runs are ordered newest-first, so the **oldest** run in the `numberOfStatus` window was what persisted — tables displayed a stale status, and the higher `numberOfStatus` was set the staler it got; **(2)** `yield_pipeline_lineage_details` returned early *before* initialising the shared topology context, so a DAG whose pipeline entity could not be resolved inherited the previous sibling DAG's tables.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

**Approach.** The cache existed to attach "historical" DAGs' observability to tables on every subsequent DAG. That is redundant by construction — `TableRepository.addPipelineObservability` keys the entity extension on `table.pipelineObservability.<pipelineFqn>` and `insert` is an upsert on `(id, extension)`, so rows from *different* pipelines coexist and rows from the *same* pipeline overwrite. Nothing needs re-assembling client-side. Emission therefore moves to "current DAG only", which is both correct and linear.

**Components.**
- `pipeline_service.py` — adds `_table_entity_cache` (`LRUCache`, `TABLE_ENTITY_CACHE_SIZE = 1024`), `cache_table_entity()` and `_get_observability_table()`. `yield_pipeline_observability` now prefers the lineage-resolved entity and only falls back to `get_by_name` on a miss. The `Dict[str, List[PipelineObservability]]` contract is unchanged, so both connectors and any future one benefit.
- `airflow/metadata.py` — removes `observability_cache`; `get_table_pipeline_observability` emits one record per table from `latest_dag_run`; context writes are consolidated into `_set_observability_context()`, called to reset *before* the early return and once at the end with real values.
- `dbtcloud/metadata.py` — same removal. It already built from `latest_run`, so it had the quadratic replay but not the run-override bug; `_track_table_fqn` now hands the resolved entity to the shared cache.

**Alternatives rejected.** Bounding `observability_cache` with an `LRUCache` the way dbtcloud already did is a ~5-line change, but it only shrinks the constant — still `O(dags × cache)` — and leaves the run-override bug in place. A server-side bulk endpoint would add another ~50× but needs a new API surface, and is unnecessary once the redundant work is gone.

**Compatibility.** No schema, API or migration changes. `TablePipelineObservability` is internal to the ingestion framework and its shape is unchanged. Behaviour change: observability now reflects the newest run rather than the oldest — that is the bug being fixed. Observability remains coupled to the lineage stage (`current_table_fqns` is only populated there, so `includeLineage: false` already produced nothing); this PR preserves that rather than changing it.

#
#### Use cases covered
- Airflow service with many DAGs sharing tables: each DAG writes observability only for its own tables, instead of re-writing every previously seen DAG's tables.
- A table touched by two different pipelines keeps a row for each (writes append across pipelines, overwrite across runs of one pipeline).
- With `numberOfStatus > 1`, a table shows the status of the DAG's **newest** run.
- A DAG whose pipeline entity cannot be resolved emits no observability, rather than the previous DAG's.
- dbt Cloud jobs: same scoping, no replay of earlier jobs.

#### Unit tests
- Files updated: `ingestion/tests/unit/airflow/test_airflow_metadata.py` (+`TestPipelineObservabilityScope`, `TestLineageObservabilityContext`), `ingestion/tests/unit/topology/pipeline/test_dbtcloud.py`
- 11 tests added, 5 removed (they asserted the replay behaviour being deleted). `1011 passed` across `tests/unit/airflow`, `tests/unit/topology/pipeline`, `tests/unit/sink`, `tests/unit/test_ometa_utils.py`.
- Every new test was mutation-checked: each fix was reverted one at a time (replay reinstated, multi-record emission, oldest-run selection, entity handoff removed, cache made unbounded, context reset moved after the early return) and the corresponding test failed in each case.
- `ruff check` / `ruff format --check` clean at CI parity (2689 files); `basedpyright --baselinemode=discard` reports 0 errors on the changed files with the baseline unmodified.

- [x] Not applicable — covered by unit tests plus the manual end-to-end run below.

#### Playwright (UI) tests

#### Manual testing performed
Ran end-to-end against a local stack (OpenMetadata 2.0.0 + Airflow 3.3.1), A/B on the same fixture: 12 DAGs with sliding-window table xlets (5 tables each, 27 distinct), 5 runs per DAG where the oldest run failed and the newest succeeded.

1. Created a `local_pg` database service with 30 tables, generated the 12 DAGs into the Airflow dags folder, and seeded 60 dag runs.
2. Ran `metadata ingest` on `main`, then on this branch, clearing observability rows in between.

| | before | after |
|---|---|---|
| observability writes | 192 | **60** |
| distinct rows persisted | 60 | 60 |
| per-DAG table count | 5 → 7 → 9 → … → 27 | flat at 5 |
| wall clock | 25.1s | **10.9s** |
| persisted status | `Failed` @ oldest run | `Successful` @ newest run |

The before-count follows `n² + 4n` exactly (12² + 48 = 192), i.e. 192 writes producing only 60 distinct rows — 69% redundant at just 12 DAGs, and the ratio grows linearly with DAG count.

3. Verified completeness through the REST API, not just the DB: all 60 expected `(table, pipeline)` pairs present, none missing, none unexpected; 12 pipelines ingested; lineage unaffected (74 edges, `t04` correctly has 3 upstreams attributed to `obsverify_dag_00`).
4. Re-ran the ingestion to confirm idempotency — identical 60 writes / 60 rows / 10.9s.

#

Not applicable.

#
### Checklist:


ulixius9:
Rebased onto `main` to resolve conflicts with #31643 (`chore(python): modernize typing and enforce lazy logging`).

That commit landed three rules this change had to be re-applied under, so the connector edits were rewritten on top of main's version rather than merged:

- `typing.Dict/List/Optional` → builtins and `X | None`; `# noqa: UP006/UP045` are now in `FORBIDDEN_SUPPRESSION_CODES`, so the new annotations use the modern form instead of suppressing.
- f-strings in logging are G004 and suppressions are forbidden, so every new/changed `logger.*` call in the observability paths uses lazy `%s` formatting.
- Removing baselined f-string log lines leaves obsolete entries, which `check_ruff_suppressions.py --check` fails on. Hence the sixth file: `ingestion/.ruff-g004-baseline.json` loses the 9 entries for log lines this PR deletes (airflow 25→18, dbtcloud 15→13), via `--prune`. No other file's entries were touched.

No behaviour change from the rebase. Re-verified after resolving: `ruff check` + `check_ruff_suppressions.py --check` + `ruff format --check` all clean at CI parity (2696 files); `basedpyright --baselinemode=discard` 0 errors on the changed files with the baseline unmodified; 1091 unit tests pass. The six mutation checks were re-run against the rewritten code — reinstating the replay, selecting the oldest run, moving the context reset after the early return, dropping the entity handoff, replaying dbt jobs, and unbounding the entity cache each still fail their corresponding test.

harshsoni2024 (approved):

PR #32375, opened 2026-09-01
Fixes #32374: serve the reindex entity list from the index registry

### LINKED ISSUE #32374: Search Indexing app: entity list should come from the backend, not a hardcoded enum in the UI schema
### Problem

The entity picker on the Search Indexing application config form is populated from a hardcoded
`items.enum` in the UI-local JSON schema
(`openmetadata-ui/src/main/resources/ui/src/utils/ApplicationSchemas/SearchIndexingApplication.json`,
plus its committed parsed copy under `src/jsons/applicationSchemas/`).

That list is a per-build snapshot and has drifted from what the server can actually reindex:

| source | count |
|---|---|
| UI schema `entities.items.enum` | 53 |
| index mapping registry (`openmetadata-spec/src/main/resources/elasticsearch/indexMapping.json`) | 68 |

Missing from the dropdown today: `aiApplication`, `aiFrameworkControl`, `aiGovernanceFramework`,
`aiGovernancePolicy`, `auditReport`, `contextMemory`, `llmModel`, `llmService`, `mcpExecution`,
`mcpServer`, `mcpService`, `pipelineStatus`, `promptTemplate`, `relationshipType`, `tableColumn`.

So selecting "everything" in the dropdown is not the same as selecting `all`, even though the app
treats `all` as "every registered index" (`ReindexingOrchestrator.getAll()` →
`searchRepository.getEntityIndexMap().keySet()`). Every entity type added since the enum was last
hand-edited is silently unreindexable from the UI.

It also cannot ever be correct for a distribution that registers extra indexes. `IndexMappingLoader`
merges `elasticsearch/indexMapping.json` **plus** `elasticsearch/collate/indexMapping.json` from the
classpath at startup, so the reindexable set is a *runtime* fact — one enum baked into the UI bundle
cannot express it. Collate works around this today by forking the whole schema file just to widen
the enum, and that fork has drifted too (49 values, and still missing all four Collate-only indexes).

### Proposed fix

The UI should not be making this ops decision. Expose the registry from the backend and have the
form fetch it:

- `SearchRepository.getIndexedEntityTypes()` — sorted keys of the merged index mapping registry.
- `ReindexingOrchestrator.getAll()` reads the same method, so the picker and `all` cannot drift.
- `GET /v1/search/entityTypes` serves it, authorized against `APPLICATION` / `VIEW_BASIC` so it matches the gate on the page that consumes it.
- Drop the enum from the schema JSON; `ApplicationsClassBase.importSchema` injects the fetched list
  (it is the single funnel both AppDetails and AppInstall go through).

Downstream: Collate can then delete its schema fork entirely, with no Collate-side backend code —
the classpath merge already makes the OSS endpoint return the Collate-correct list.

Companion Collate issue: https://github.com/open-metadata/openmetadata-collate/issues/3946

mohityadav766:
### Describe your changes:

Fixes #32374

The Search Indexing application's entity picker was populated from a hardcoded `items.enum` in the
UI-local schema JSON. That made a *runtime* fact (which indexes the server registered) into a
per-build snapshot, and it had drifted 15 entity types behind the registry:

| source | count |
|---|---|
| UI schema `entities.items.enum` | 53 |
| index mapping registry | 68 |

Absent from the dropdown: `aiApplication`, `aiFrameworkControl`, `aiGovernanceFramework`,
`aiGovernancePolicy`, `auditReport`, `contextMemory`, `llmModel`, `llmService`, `mcpExecution`,
`mcpServer`, `mcpService`, `pipelineStatus`, `promptTemplate`, `relationshipType`, `tableColumn`.

So "check everything in the dropdown" was not equivalent to `all`, even though the app already
expands `all` from the registry — every entity type added since the enum was last hand-edited was
silently unreindexable from the UI.

It also could never be correct for a distribution that registers extra indexes: `IndexMappingLoader`
merges `elasticsearch/indexMapping.json` **plus** `elasticsearch/collate/indexMapping.json` from the
classpath at startup. Collate forked this entire schema file just to widen the enum, and that fork
had drifted too.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

**One definition, two readers — so they cannot drift.**

- `SearchRepository.getIndexedEntityTypes()` — sorted keys of the merged index mapping registry.
- `ReindexingOrchestrator.getAll()` now calls it (was inlining `getEntityIndexMap().keySet()`), so
  the `all` expansion and the picker are the same set by construction.
- `GET /v1/search/entityTypes` serves it, authorized against **`APPLICATION` / `VIEW_BASIC`** — the same gate as the page that calls it (`SettingsRouter` renders `AppDetails` for `isAdminUser || hasViewPermissions(APPLICATION)`, and that page fetches the schema on every mount). It was admin-or-bot in the first revision; changed after review, because admin-only would 403-toast a legitimate viewer on a request they never triggered. Placed on `SearchResource` because that is
  where `searchRepository` and the sibling registry endpoints (`/templates`, `/stats`,
  `/reindexEntities`) already live.

**UI**

- The enum is removed from `SearchIndexingApplication.json` — from the source under
  `utils/ApplicationSchemas/` **and** the tracked parsed copy under `jsons/applicationSchemas/`
  (this repo commits the parsed output; editing only the source would leave the shipped schema stale).
- `getSearchEntityTypes()` added to `rest/searchAPI.ts`.
- Injected in `ApplicationsClassBase.importSchema`, the single funnel `AppDetails` and `AppInstall`
  both go through, so neither call site needs to know about it.

**Failure mode.** RJSF picks the multiselect widget off `items.enum`, so the enum has to be present
at render time. A failed fetch injects an empty enum — the tree still renders with "All" selectable
(and `["all"]` is the default and the common choice) — plus an error toast. Letting the fetch throw
would have failed the whole config form for what is an enrichment of one field.

**Alternatives rejected.** Generating the enum into the schema JSON at build time from
`indexMapping.json`: needs new codegen wiring, still leaves Collate with its own generated fork, and
still cannot see a classpath merge that only exists at runtime — which is the actual bug.

**Deliberately out of scope** — same *shape*, different sources of truth; do not point this endpoint
at them:
- `RdfIndexApp.json` `entities` — the triplestore entity list (`app`, `bot`, `dataContract`, …), not
  search indexes.
- `CacheWarmupApplication.json` `entities` — repository-backed types, and `"all"` is a real enum
  member there.

**Compatibility.** Additive endpoint; no schema or migration changes. Stored `appConfiguration`
values are untouched — the backend schema's `entities` was always a plain `string[]` with no enum, so
nothing that validates today stops validating. `tableColumn` and friends now become individually
selectable; they were already included whenever `all` was used, so this is not a new code path.

#
#### Use cases covered
- Admin opens the Search Indexing app config: the entity picker lists every entity type the server
  registered, including ones added after this build was cut.
- Admin on a Collate deployment sees Collate-only indexes in the same picker, with no Collate backend
  code.
- `GET /v1/search/entityTypes` is rejected for a non-admin.
- The entity list endpoint is unavailable: the form still renders and "All" is still selectable.

#### Unit tests
- [x] Added, in `openmetadata-ui/.../AppDetails/ApplicationsClassBase.test.ts`:
  - fills the entity list from the server
  - the schema JSON carries no hardcoded enum (asserts on the JSON module directly, so it fails if
    the enum is ever re-added)
  - a failed fetch toasts and leaves the list empty
  - other applications do not trigger the fetch
- 10/10 pass in that suite (4 new + 6 pre-existing). `AppDetails` + `AppInstall` + `TreeSelectWidget`:
  21/21 pass.
- Coverage % on changed classes: **not measured** — `jacoco:report` / `yarn test:coverage` were not run.

- [x] Added 2 to `openmetadata-integration-tests/.../SearchResourceIT.java`:
  - `testEntityTypesComesFromTheIndexRegistry` — 200, non-empty, sorted, and contains `tableColumn`
    (which the old hardcoded enum lacked, so its presence is what proves the list is read from the
    registry rather than copied)
  - `testEntityTypesIsReadableByApplicationViewer` — 200 and a non-empty list for a `DataConsumer`, pinning that an Application viewer is not 403'd. (`DataConsumer` grants `ViewAll` on all resources and is the default role, so there is no clean 403 case left to assert without building a deny policy — say the word if you want one.)
- These **compile but have not been executed** here (they need a live server + ES/OS). CI will be the
  first real run.

- [x] Not applicable.

#### Playwright (UI) tests
- Not added. The existing `SearchIndexApplication.spec.ts` already drives this picker (it searches for
  and toggles "Table"), and it now exercises the server-driven list unchanged — which is the useful
  end-to-end signal. Happy to add a dedicated spec if reviewers want one.

#### Manual testing performed
- `mvn compile -pl openmetadata-service` and `mvn test-compile -pl openmetadata-integration-tests`: pass.
- `mvn spotless:apply`: no changes.
- Jest suites above: pass. `yarn lint:base` on changed files: 0 errors. `prettier --check`: clean.
- Verified `generateApplicationDocs.js` does not consume `items.enum` by regenerating all app docs:
  zero diff.
- **No local stack run / browser verification.** Flagging that honestly rather than checking the box.

#

**Not attached** — I did not have a running stack to record against, so I cannot honestly claim this.
The visible change is that the existing tree-select lists more entity types; there is no new UI
surface. If a recording is required before merge, please say so and I will get one.

#
### Checklist:
- [x] For JSON Schema changes: no backend schema changed. The UI-local schema lost a UI-only enum, so
      no migration is needed — stored configs were never validated against it.
- [ ] For UI changes: I attached a screen recording and/or screenshots above. — see note above

- [x] I have added a test that covers the exact scenario we are fixing — the "no hardcoded enum" unit
      test and the `tableColumn` assertion in the IT both fail if the stale-list behaviour returns.

#
### Downstream

Collate can delete its `SearchIndexingApplication-collate.json` fork once this lands and the submodule
pin moves; no Collate backend change is needed, because the classpath merge in `IndexMappingLoader`
already makes this endpoint return the Collate-correct list. Companion Collate issue:
https://github.com/open-metadata/openmetadata-collate/issues/3946


karanh37 (dismissed):

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/SearchResourceIT.java
```diff
@@ -1985,4 +1987,57 @@ void testExportWithFromBeyondResults(TestNamespace ns) throws Exception {
...
+    String consumerToken =
+        JwtAuthProvider.tokenFor(
+            "data-consumer@open-metadata.org",
+            "data-consumer@open-metadata.org",
+            new String[] {"DataConsumer"},
+            3600);
+
+    HttpResponse<String> response = httpGetJson("/v1/search/entityTypes", consumerToken);
+
+    assertEquals(403, response.statusCode(), "Non-admin should not be able to list entity types");
```
manerow: I think this will get a **404** rather than a 403.

`authorizeAdminOrBot` resolves the user before it can deny: `DefaultAuthorizer:157` -> `SubjectContext:63` -> `SubjectCache.getUserContext` -> `Entity.getEntityByName`. `SearchResourceIT` never creates `data-consumer@open-metadata.org`, so that lookup throws `EntityNotFoundException`, which carries `Response.Status.NOT_FOUND`.

It could still pass by luck if another test class created that user earlier in the same server session, which would make it order-dependent rather than reliable.

`AppOperationPermissionsIT.java:38-45` hits the same thing and works around it like this, if you want to reuse the pattern:

```java
@BeforeAll
static void ensureDataConsumerUser() {
  UserTestFactory.getDataConsumer(null);
}
```

openmetadata-service/src/main/java/org/openmetadata/service/resources/search/SearchResource.java
```diff
@@ -834,6 +835,33 @@ public Response aggregateSearchRequest(
...
+            responseCode = "200",
+            description = "Sorted list of entity types",
+            content =
+                @Content(
+                    mediaType = "application/json",
+                    array = @ArraySchema(schema = @Schema(type = "string")))),
+        @ApiResponse(responseCode = "403", description = "Admin only")
+      })
+  public List<String> getIndexedEntityTypes(@Context SecurityContext securityContext) {
+    authorizer.authorizeAdminOrBot(securityContext);
```
manerow: One thing to consider: `authorizeAdminOrBot` is stricter than the page that calls this endpoint.

The Search Indexing app details page is gated on *view* permission rather than admin. `SettingsRouter.tsx:551` passes `hasViewPermissions(ResourceEntity.APPLICATION, ...)`, which is `ViewBasic || ViewAll` (`PermissionsUtils.ts:113-114`), and `AdminProtectedRoute.tsx:34` renders children when `isAdminUser || hasPermission`. `AppDetails.component.tsx:524` then calls `fetchAppDetails` -> `importSchema` on every mount.

So a non-admin holding `ViewBasic` or `ViewAll` on Application can open the page and would see a red error toast each time, for a request they never triggered.

Two ways to line the two sides up, whichever you prefer: authorize this endpoint against `APPLICATION` view permission instead of admin-only, or leave the endpoint as-is and skip the toast on 403 in `ApplicationsClassBase.ts:47` while still toasting other failures.

openmetadata-ui/src/main/resources/ui/src/components/Settings/Applications/AppDetails/ApplicationsClassBase.ts
```diff
@@ -24,14 +28,47 @@ const ApplicationConfiguration =
     lazy(() => import('../ApplicationConfiguration/ApplicationConfiguration'))
   );
 
+const SEARCH_INDEXING_APPLICATION = 'SearchIndexingApplication';
```
manerow: Minor: this constant already exists as `SEARCH_INDEXING_APPLICATION` in `src/constants/explore.constants.ts:117` with the same value, so you may want to import it instead of keeping a second copy.

manerow (commented):
`playwright/e2e/Pages/SearchIndexApplication.spec.ts:330` looks like it will start failing with this change.

```ts
await ...getByRole('combobox').fill('Table');
const tableTitle = page.getByRole('tree').getByTitle('Table');
```

The old enum had exactly one value containing `table`. The registry adds `tableColumn`, which `TreeSelectWidget` renders through `startCase` as `"Table Column"`. AntD filters the tree on the node **value** using a case-insensitive substring match (`rc-tree-select/OptionList.js:122`), so typing `Table` now leaves two nodes visible.

Playwright's `getByTitle` also defaults to substring matching, so this locator resolves to 2 elements. `waitFor` (333), `evaluate` (336) and `click` (349) all run in strict mode, so line 333 would be the first to throw `strict mode violation`.

A minimal fix could be to make the match exact:

```ts
const tableTitle = page.getByRole('tree').getByTitle('Table', { exact: true });
```

Might be worth running this spec locally before merge. The description mentions it exercises the picker "unchanged", but the list it renders is no longer the same, so I do not think we can rely on it as-is.

mohityadav766:
Pushed `f97e56e`. All four points were real — thanks, especially for the two that CI was never going to catch.

### @manerow — the Playwright spec

You were right, and worth spelling out **why the green run didn't disprove it**: the impact map never selected that spec. `.github/playwright/impact-map.json` routes `src/components/Settings/**` to `SettingsNavigationPage`, `LoginConfiguration`, `Policies`, `Roles`… — none of which open an app config form. `SearchIndexApplication.spec.ts` appears nowhere in the file, so the 1171 passing tests simply never touched the picker. My PR text claimed the spec "exercises the picker unchanged"; it was wrong twice over — the list changed *and* the spec didn't run.

Verified the mechanism against the installed packages rather than taking it on trust:

- `rc-tree-select/lib/TreeSelect.js:90` — `treeNodeFilterProp` defaults to `'value'`.
- `rc-tree-select/lib/OptionList.js:122` — `String(treeNode[treeNodeFilterProp]).toLowerCase().includes(lowerSearchValue)`, so `'table'` and `'tablecolumn'` both match `"table"`.
- `rc-tree/lib/TreeNode.js:319` — `title: typeof title === 'string' ? title : ''`, so the second node really does carry `title="Table Column"`.

`getByTitle` is substring by default → 2 nodes → `waitFor` throws first, exactly as you traced. Fixed with `{ exact: true }`.

I also added an impact-map entry for the application config forms (`components/Settings/Applications/**`, `pages/AppInstall/**`, `utils/ApplicationSchemas/**`, `jsons/applicationSchemas/**`, plus the searchIndex bundle and apps resources) → the four app-page specs. That closes the gap that hid this, and since `.github/playwright/**` is `sharedInfrastructure`, it puts this PR into a full run — so the locator fix gets exercised here rather than on whoever touched it next.

### @manerow — 403 vs 404 in the IT

Confirmed: `authorizeAdminOrBot` → `getSubjectContext` → `SubjectCache.getUserContext` (`SubjectCache:204-210`), which falls back to `Entity.getEntityByName` and throws `EntityNotFoundException` → 404. This suite never creates `data-consumer@open-metadata.org`, so it was order-dependent as you said. Used your `AppOperationPermissionsIT` pattern verbatim — and it's needed regardless of which authorizer the endpoint ends up calling, since every path resolves the subject first.

### @manerow — authz stricter than the page

Took your first option and moved the endpoint rather than suppressing the toast. `SettingsRouter.tsx:549-555` gates `AppDetails` on `isAdminUser || hasViewPermissions(APPLICATION)`, `AdminProtectedRoute.tsx:34` renders on either, and `AppDetails` fetches the schema on mount — so admin-only here really would have red-toasted a viewer on a request they never made. Now:

```java
OperationContext operationContext =
    new OperationContext(Entity.APPLICATION, MetadataOperation.VIEW_BASIC);
authorizer.authorize(securityContext, operationContext, new ResourceContext<>(Entity.APPLICATION));
```

Suppressing the toast would have left the viewer with a silently empty picker while the form still displays saved `entities` values, which seemed worse than aligning the gate. `CompiledRule:243` confirms `ViewAll` matches any view operation, so the UI's `ViewBasic || ViewAll` and this check agree.

Consequence for the test: `DataConsumer` grants `ViewAll` on `all` resources and is the default role, so there is no longer a clean 403 case to assert without building a deny policy. I turned that test into the one that now carries the risk — an Application viewer gets 200 and a non-empty list — which pins your concern directly. If you'd rather keep an explicit negative, say so and I'll add a role with an Application deny.

### @manerow — duplicate constant

Fixed, importing from `constants/explore.constants.ts`. Checked the graph first since that module imports `SortingField` from `components/Explore/SortingDropDown` — it's `export interface` (`SortingDropDown.tsx:18`), type-only, so esbuild elides the import and nothing new lands in this chunk.

### @copilot — `all` missing from the enum

Added; the fallback on fetch failure is now `['all']` instead of empty, which is your suppressed comment's suggestion too. Two notes for the record: this inconsistency **predates** the PR (the old hardcoded enum didn't contain `all` either, alongside the same `default: ["all"]`), and `TreeSelectWidget` filters `all` out of the child nodes, so it renders once as the synthetic "All" parent rather than twice.

### Verification

`mvn compile -pl openmetadata-service` ✅ · `mvn test-compile -pl openmetadata-integration-tests` ✅ · `spotless:apply` no-op ✅ · jest 10/10 ✅ · `lint:base` on changed files 0 errors ✅ · `yarn lint:playwright` 0 errors ✅.

Still not run locally: the two ITs (need a live server + ES) and the Playwright spec — the full run this push triggers is the first real execution of that locator, so I'd rather it be checked there than assert it works. And I still owe a screen recording; flagging again rather than quietly leaving the box unticked.

manerow (dismissed):

mohityadav766:
CI triage on `60aaf5ad` (the main merge). One of the eight failures is mine; pushed `30502b0d`.

### Mine — `openmetadata-service-unit-tests`

```
ReindexingOrchestratorTest.setupEntitiesExpandsAllAndCountTotalEntitiesSkipsUnsupportedTypes:334
  expected: <true> but was: <false>
```

That test stubs `searchRepository.getEntityIndexMap()`, which `setupEntities()` stopped calling when I moved `getAll()` to `getIndexedEntityTypes()`. `searchRepository` is a plain `mock(...)`, so the new call returned Mockito's default **empty** `Set` — `"all"` expanded to nothing and the `contains(TABLE)` assertion failed. Stubbed the method the code actually calls now:

```java
when(searchRepository.getIndexedEntityTypes()).thenReturn(Set.of(Entity.TABLE, reportType));
```

`ReindexingOrchestratorTest` is 12/12 green with that. I checked the other four tests that stub `getEntityIndexMap()` (`SearchIndexMetricsTest`, `SystemRepositoryReindexStatusTest`, `SystemRepositoryMissingIndexesTest`) — none are on the `getAll()` path, so they still stub the right method.

`Test Report` and `openmetadata-service-unit-tests-status` are just this failure reported again.

**On me:** I compiled `openmetadata-service` before opening the PR but never ran its unit tests, which is why this reached review instead of my terminal.

### Not mine — `search-it-nightly`

```
HighlightFieldSaveValidationIT.everyEndpointServingSearchSettingsCarriesTheHighlightFlag:66
  ->assertServesHighlightFlag:80->allowedFieldsFor:128  no allowedFields entry for table
```

This arrived with the main merge, not from this branch:

- The file does not exist at `39dc25a6` (my last commit before the merge) — it came in with `1b149b93` "Fixes 28727: … validate highlight fields at save time".
- Its own PR branch `ISSUE-28727` failed **the same job, test, and line** before it merged — [run 33535202679, job 99947700806](https://github.com/open-metadata/OpenMetadata/actions/runs/33535202679).
- This branch passed `java-playwright-nightly` twice pre-merge: [33516866301](https://github.com/open-metadata/OpenMetadata/actions/runs/33516866301) and [33538202601](https://github.com/open-metadata/OpenMetadata/actions/runs/33538202601).

I also read the test to be sure it wasn't picking up my endpoint: "every endpoint serving search settings" means the five `/v1/system/settings` routes, and `allowedFields` comes from `SettingsCache`, which this PR doesn't touch. Someone on #28727's side should look at it.

### Not mine — `python / Integration Tests (shard-2, 3.11)`

All 13 are `ERROR at setup` for MSSQL and NoSQL containers:

```
ERROR tests/integration/sql_server/test_metadata.py::test_ingest_metadata[MssqlScheme.mssql_pytds]
ERROR tests/integration/profiler/test_nosql_profiler.py::NoSQLProfiler::test_simple
...
```

Fixture/container setup, not assertions. This PR contains no Python. `python / Verify Expected Jobs` and `py-tests-status` are downstream of it.

### Still blocked on a maintainer — `Validate PR Metadata`

> Linked issue #32374 is not added to the **Shipping** project.

One item. My token has no `read:project`, so I can't set it — could someone with board access add #32374 and re-run the check?

### Verification

`ReindexingOrchestratorTest` 12/12 · `spotless:apply` no-op · the test file is byte-identical across the merge (`git diff 39dc25a6 60aaf5ad` on it is empty), so the fix applies unchanged either side of it. A full re-run on the merged tree is in progress locally — the merge needs a rebuilt `openmetadata-spec` and the shaded ES deps, so `mvn test -pl openmetadata-service` alone can't build it; CI is the real check regardless.

yan-3005 (dismissed):

karanh37 (dismissed):

mohityadav766:
Merged `main` and resolved the conflict — `8eb6a75c`. Back to `MERGEABLE`.

### The conflict

One file: `ApplicationsClassBase.ts`. Main replaced the template-literal `import()` in `importSchema` with a glob loader map (`applicationSchemaLoaders`, from the new sibling `.assets.ts` so ts-jest can mock what it can't parse) — which rewrote the exact lines this branch had added the entity-list injection to.

Resolution keeps main's loader and re-applies the injection to the resolved module:

```ts
const module = await loader();
const schema =
  (module as { default?: unknown }).default ??
  (module as Record<string, unknown>);

return fqn === SEARCH_INDEXING_APPLICATION
  ? withSearchEntityTypes(schema as RJSFSchema)
  : schema;
```

The react import also splits per main's move to type-only imports, so `RJSFSchema` and `AxiosError` come in the same way:

```ts
import type { RJSFSchema } from '@rjsf/utils';
import type { AxiosError } from 'axios';
import type { ComponentType, FC } from 'react';
import { lazy } from 'react';
```

Checked that main's glob mock (`src/test/unit/mocks/glob.mock.js`) lazily `require()`s from `src/jsons/applicationSchemas/`, so `importSchema('SearchIndexingApplication')` still resolves the real on-disk JSON under jest — the injection path is genuinely exercised, not stubbed away.

### One thing worth flagging

Main's new `importSchema` resolves to `{}`, so property access on the result is a type error. That's 7 in this test file under `tsc --noEmit` — **5 of them are main's own** (`should import pre-parsed schema` and `should import the RDF app schema`, unchanged on this branch; `git show origin/main:…test.ts` has `schema.properties.entities.default` with no cast). I named the shape for the 2 my tests read rather than adding to the pile:

```ts
type EntitiesEnumSchema = {
  properties: { entities: { items: { enum: string[] } } };
};
```

I left main's 5 alone — not my change, and touching them widens the diff for no behavioural gain. Flagging rather than silently ignoring: `tsc` isn't in the checkstyle gate (and the repo has unrelated errors elsewhere, e.g. `AppLiveIndexing.component.tsx`, `BotDetails.component.tsx`), so nothing is red because of it — but whoever owns the glob-loader change may want to give `importSchema` a real return type.

### Verification after the merge

- `ApplicationsClassBase.test.ts` 10/10, plus `AppInstall` and `AppDetails` — **23/23 across 3 suites**.
- `eslint` 0 errors, `prettier --check` clean on both changed files.
- `tsc` — my 2 errors gone, main's 5 unchanged.
- Confirmed every part of this PR survived the merge: the endpoint and `getIndexedEntityTypes()`, `getAll()` delegating to it, the `ReindexingOrchestratorTest` stub, `getSearchEntityTypes` in `searchAPI.ts`, the enum removed from **both** schema copies, the corrected `payLoadSize` description, the `{ exact: true }` Playwright locator, and the impact-map entry.
- `git diff origin/main..HEAD --stat` is exactly the 13 files this PR owns, nothing extra.

Still outstanding and not something I can do: **`Validate PR Metadata`** needs issue #32374 added to the **Shipping** project.

karanh37 (dismissed):

mohityadav766:
Triaged both. Neither is from this branch, but one of them I could unblock — pushed `6b2682fc`.

### `ui-checkstyle` — main is red, fixed here

The failing step is the playwright guardrail, not ESLint (the lint step reported **0 errors**, 3 warnings, all pre-existing in `searchAPI.ts` at lines 27/101/183):

```
not ok 27 - the suppressions baseline matches its recorded state exactly
+   'om-playwright/no-positional-locator': 1322
-   'om-playwright/no-positional-locator': 1323
```

`corpus.test.mjs` is a pure function of two tracked files — it sums `eslint-suppressions.json` and compares to a recorded constant. It runs no linter. And both files in this branch are **byte-identical to main**:

```
$ git diff --stat origin/main -- .../eslint-suppressions.json .../corpus.test.mjs
(empty)

$ git show origin/main:.../corpus.test.mjs | grep no-positional-locator
    'om-playwright/no-positional-locator': 1323,
$ git show origin/main:.../eslint-suppressions.json | sum-of-counts
1322
```

So main is inconsistent with itself and every branch merged with it fails this. Culprit is **#32576** (`test(playwright): stabilise the top-5 merge-group flakes at their cause`), which changed `eslint-suppressions.json` by one line and did not touch `corpus.test.mjs`:

```
.../ui/eslint-suppressions.json                    |  2 +-
.../e2e/Features/AdvancedSearchSuggestions.spec.ts | 52 ++++++-----
...  (corpus.test.mjs not in the diff)
```

Lowered the constant to 1322 — exactly what the assertion message asks for. I deliberately did **not** run `yarn lint:playwright:suppressions`: that rewrites a tracked file and is only for when you have fixed a violation. I fixed none, and `yarn lint:playwright` is already clean at 0 errors, so there was nothing to prune — only the bookkeeping. Now:

```
$ yarn test:eslint-rules       → # pass 114, # fail 0
$ node scripts/generate-playwright-rule-table.mjs --check → exit 0
```

@pmbrull / whoever owns #32576 — happy to drop this commit if you would rather fix it on main directly; it just blocks everything until one of us does.

### Playwright — two shards, neither in this PR's blast radius

**`chromium-02` exited 124** — the *job* hit its time limit. Its `✘` entries (`Table.spec.ts:386` nested-fields, `Entity.spec.ts:339` Tier ×3 entities, `Entity.spec.ts:353` Certification ×3, `Glossary.spec.ts:128`) each burned 1.0–1.2m and the log contains **zero `Error:` lines** — the job was killed before the reporter printed any failure detail. Six parametrised cases across two unrelated specs all "failing" with no assertion output is a shard that ran out of wall clock, not a regression.

**`chromium-06` exited 1** on a real assertion, in Glossary:

```
Glossary.spec.ts:469
  expect(viewerContainerText).toContain(glossaryDetails.description);
  Expected substring: "Glossary terms that describe general conceptual terms. …"
```

Glossary description rendering — nothing this PR touches. The same spec was already listed **flaky** on the previous run of this branch.

### The part that does matter, and it passed

My impact-map entry worked, and the spec it selects ran in the `Basic` project inside `chromium-03`:

```
✓ 157 [Basic] › playwright/e2e/Pages/SearchIndexApplication.spec.ts:229:7 › Search Index Application @basic (1.5m)
```

That is the first real execution of @manerow's finding — the entity picker driving the **server-provided** list, with `tableColumn` now in it, through `getByTitle('Table', { exact: true })`. It passes against a live server. Before the impact-map entry that spec was never selected on this PR, so this is exactly the coverage that was missing.

For reference, the previous run on `3ce082ea` was ✅ 1182 passed / 0 failed with 2 flaky (`ExplorePageRightPanel_KnowledgeCenter`, `Glossary.spec.ts`) — same branch, same impact-map entry. The only delta since is main's content.

karanh37 (dismissed):

yan-3005 (approved):

chirag-madlani (approved):

mohityadav766:
Both merge-queue failures on `7a2bfa75` are pre-existing breakage on the queue base, not from this PR. The unit-test one is worth flagging beyond this PR — **main's green on it is a skip, not a pass.**

### `OpenMetadata Service Unit Tests` — a latent semantic conflict on main

3 failures, all in `DataContractPipelineTokenRefreshTest` (6 run / 3 failed, out of 8751):

```
expected: <PipelineServiceClientException> but was: <BadRequestException>
  Caused by: BadRequestException: sourceConfig.config.type is required
expected: <[stale-jwt-token, rotated-jwt-token]> but was: <[]>
```

Two PRs that were each green on their own:

- **#29566** "Reject ingestion pipeline sourceConfig without type" (landed today) added `IngestionPipelineRepository:104` → `SOURCE_CONFIG_TYPE_REQUIRED = "sourceConfig.config.type is required"`.
- **#31848** "Fixes 24806: refresh the Data Contract DQ DAG deployment on every validation" (Sept 2) owns `DataContractPipelineTokenRefreshTest`, whose fixture builds a pipeline with **no `sourceConfig` at all**:

```java
private IngestionPipeline testSuitePipeline(boolean deployed) {
  return new IngestionPipeline()
      .withId(UUID.randomUUID())
      .withName("dq-pipeline")
      .withFullyQualifiedName("dq-pipeline")
      .withPipelineType(PipelineType.TEST_SUITE)
      .withDeployed(deployed);          // ← no sourceConfig
}
```

So the new validation rejects the fixture before the deploy path runs — which is also why the second assertion sees `[]`: no tokens were ever captured because no deploy happened.

**Why it looks like mine and isn't:**

- `#29566` **is an ancestor of my queue base** `19af6248` — `git merge-base --is-ancestor 27063928f7 19af6248` passes. The validation was there before my commit.
- The batch `19af6248..7a2bfa75` contains exactly one commit (mine), so the queue points the finger here.
- But main's own run on that identical base ([run 34120336607](https://github.com/open-metadata/OpenMetadata/actions/runs/34120336607)) reports `success` with `openmetadata-service-unit-tests` **`skipped`** — the job is path-filtered by `Detect Changes`, and that merge touched no Java. So main was never actually verified on this base.
- This PR touches Java, so the queue run actually executed the suite and surfaced it. The unrelated queue entry **PR #32708** ([run 34129000151](https://github.com/open-metadata/OpenMetadata/actions/runs/34129000151)) does the same and fails the **identical 3 tests**.

Net: any Java-touching PR that queues will hit this until the fixture and the validation agree. Worth someone looking at the path filter too — it means Java breakage can sit on main invisibly and only detonate in the queue.

I have **not** patched it here, deliberately: the fix has a real design question in it (does a `TEST_SUITE` pipeline legitimately need a `sourceConfig.config.type`, or should the new validation exempt that type?), and that is #29566's and #31848's call, not mine. Happy to push the one-line fixture change if either owner tells me which way they want it.

### `Postgresql PR Playwright E2E Tests` — unrelated flake

One failure, `chromium-13`:

```
Tag.spec.ts:563 › Tag Page with Data Steward Roles › Add and Remove Assets for Data Steward
  Test timeout of 60000ms exceeded.
  expect(locator).toContainText("0") failed — element(s) not found
  waiting for getByTestId('assets').getByTestId('filter-count')
```

A Tag-page asset filter-count badge that never rendered. `SearchIndexApplication.spec.ts` was not in that shard at all (`grep -c` = 0), and nothing in this PR touches tags, assets, or Data Steward roles.

### This PR

Nothing to change — review is `APPROVED` and the last push (`a5f6e1ef`) was green on the PR checks. It just needs re-queueing once main's `DataContractPipelineTokenRefreshTest` is sorted. The only thing still on our side is the `Validate PR Metadata` item: issue #32374 into the **Shipping** project.

PR #32379, opened 2026-09-01
fix(clickhouse): support cross database lineage

### LINKED ISSUE #26095: Support Cross Database Lineage for Clickhouse
I enabled Cross Database Lineage in my Clickhouse database service to ingest lineage from our Postgres database service, but the Lineage agent displayed these logs:
```
[2026-02-24T16:15:26.096407Z] INFO - lineage_source.py:294 - Lineage processing completed with 1/1 chunks processed 
[2026-02-24T16:15:26.096847Z] INFO - lineage_source.py:524 - Processing Cross Database Lineage not supported for Clickhouse
```

Support for Clickhouse for Cross Database Lineage, so that I can see lineage edges from my postgres tables that are replicated to my Clickhouse database (they have the same names).

None.. At office hours, I was told that creating a New Connector for ClickPipes (the CDC replication tool we use) is too complex 

**Additional context**

Khairajani:
Fixes #26095

Cross database lineage was never implemented for ClickHouse — enabling it only logged `Processing Cross Database Lineage not supported for Clickhouse`.

## Changes

- `ClickhouseLineageSource.yield_cross_database_lineage()` — links a ClickHouse table to the table it replicates in one of the configured services. ClickHouse keeps no reference to where a replicated table came from, so the match is made on the table itself: the table at the same schema and name wins, and failing that the only table of that name whose columns match. A ClickHouse database maps to an OpenMetadata schema under a single `default` database, so a replica rarely keeps the schema it came from — the lookup by name is what covers the usual replication setup. An ambiguous name yields nothing and logs why: a missing edge beats a wrong one.
- `check_same_table` and `get_cross_database_fqn_from_service_names` move from the Trino connector up to `LineageSource`, unchanged, so both connectors share them.

## Testing

Unit tests for both lookups, ambiguity, mismatched and missing columns, error reporting, and that cached search results are re-matched per table. Removing either the ambiguity guard or the name lookup fails the matching test.

## Notes

- #27551 covered this issue and was closed by the stale bot. Its `check_same_table` and prefix-swap lookup were sound and are effectively what the FQN path does here, but the prefix swap alone requires the ClickHouse database name to equal the source schema name, which replication rarely produces — hence the lookup by name. That PR also bundled dictionary lineage, which belongs to #26265's thread and is out of scope here.
- Cost is one `get_by_name` per table per configured database, plus one search per distinct table name per service. Results are cached; the reporter's scale is fine, a service with very many tables would notice.


IceS2 (changes requested):
@Khairajani, Let's think this through as this capability should be wide across any connector that implements lineage. Let's come with a decent design to make us start on that path

PR #32381, opened 2026-09-01
[ESLint] Extract complex expressions (expression-complexity) + promote to error

### LINKED ISSUE #30984: [ESLint 9/10] sonarjs complexity
Part of epic #30977.

cyclomatic/cognitive/expression complexity + nested-conditional/nested-functions. ~60 safe refactors; rest disable-with-reason. 899 warnings.

Delivered as PR 7 of 10 (stacked). Behavior-preserving; verified via ESLint before/after + tsc regression check.

ShaileshParmar11:
## Summary

Part of **rule 9** of the ESLint-cleanup epic (#30977, sub-issue #30984 — the *sonarjs complexity* family). Takes **`sonarjs/expression-complexity`** (SonarJS S1067, "too many conditional operators in one expression — max 3") to **0 violations** and promotes it **`warn` → `error`**. The remaining complexity rules in #30984 (`cyclomatic-complexity`, `cognitive-complexity`) stay `warn` for follow-up PRs.

> **Stacked on #32364** (`no-nested-functions`), which is stacked on #32348 (`no-nested-conditional`). Base is #32364's branch so this diff shows only the 110 expression-complexity files; GitHub retargets automatically as the bases merge.

## Approach — real refactors, short-circuit-safe

All **141 violations across 110 files** fixed by extracting sub-expressions into named `const`s so no single expression has >3 `&&`/`||`/`?:` operators. The key hazard — naively splitting a short-circuit chain can eagerly evaluate a dereference out from behind its null-guard and throw — was avoided by:
- extracting a **contiguous leading same-operator run** (`a && b && c && d` → `const abc = a && b && c; abc && d`) so evaluation order and short-circuiting are byte-for-byte identical;
- keeping every null/undefined guard ahead of what it guards (or using optional chaining in the extracted const);
- preserving TypeScript aliased-condition **narrowing** (no `Boolean(...)` wraps around narrowing guards; `if`-splits where a guard feeds later access).

Two heavy JSX subtrees (`OntologyExplorer` ~9 operators, `AlertAiDestinationItem` ~12) use a `<>` → `<Fragment>` boundary — **identical render output** — to isolate sibling conditionals where const extraction was impractical. Called out for reviewer visibility; no `eslint-disable` anywhere.

## Verification

- **ESLint:** `eslint --quiet src` with the rule at **error** → **0 errors tree-wide, 0 `expression-complexity`**.
- **Types:** full `tsc` diff vs the stacked base → **0 net-new errors in any of the 110 files** (remaining diff signatures are union-member-ordering noise in unrelated files; branch has fewer total).
- **Tests:** the **80 co-located suites** for the changed files pass — **1639 tests**.
- organize-imports + prettier on the changed files → 0 diff.

## Ratchet

```diff
- 'sonarjs/expression-complexity': 'warn', // 15
+ 'sonarjs/expression-complexity': 'error',
```

---

> **Tracker:** does _not_ close #30984 — clears only the `expression-complexity` sub-rule of the complexity family, leaving #30984 open for cyclomatic/cognitive. The PR-metadata gate is bypassed via the `skip-pr-checks` label for this reason.


karanh37 (dismissed):

karanh37 (approved):

ShaileshParmar11:
### ✅ Full Playwright suite — green on `38a9ada6`

The merge-queue run [33767728012](https://github.com/open-metadata/OpenMetadata/actions/runs/33767728012) failed three tests. I re-ran the **complete** suite against this PR's head to check whether they were real regressions.

**Run: [33778356117](https://github.com/open-metadata/OpenMetadata/actions/runs/33778356117) — `conclusion: success`**

```
event:              workflow_dispatch (full_suite=true)
PW_EXECUTION_MODE:  full
head_sha:           38a9ada6a1d7250861451cd72479071ac57275e8
shards:             34   (non-success: 0)
jobs:               43 success · 2 skipped · 0 failure
tests:              4472 passed · 14 flaky · 0 failed
```

This is the full suite, not the targeted subset — the PR-level check ran 18 shards, this ran 34. `ExploreTree.spec.ts` in particular was never selected at PR level, so this is its first run against these changes.

#### The three merge-queue failures all pass here

```
✓  70 [chromium] › LargeGlossaryPerformance.spec.ts:379:7 › should handle drag and drop for term reordering (19.9s)
✓  16 [chromium] › TestLibrary.spec.ts:106:9              › should create, edit, and delete a test definition (17.2s)
✓ 164 [Basic]    › ExploreTree.spec.ts:186:7             › Verify Tags navigation via Governance tree ... (9.8s)
```

None of the three appear in the flaky list either — they passed outright, first attempt.

#### Why the queue run was not this PR

- All three tests had already passed in this PR's **earlier** merge-queue run [33764330212](https://github.com/open-metadata/OpenMetadata/actions/runs/33764330212) (`success`), on the same PR content.
- The whole 14:35 queue batch failed — `32506 → 32445 → 32381 → 32211 → 32472` — each on a *different* random set of tests. `32506` and `32445` sit **upstream** of this PR and do not contain it.
- The PR touches none of the three features: no `ExploreTree` files; `TestCaseFormBody.tsx` is the test *case* form, while the spec drives `test-definition-form-body`; and the one Glossary file changed, `GlossaryTermTab.component.tsx`, is a 2-hunk semantically-inert extraction.
- Every failure signature was timing — `element was detached from the DOM`, `element is not stable`, click performed then 60s timeout, modal never appearing — not an assertion on a wrong value.

<sub>14 flaky (passed on retry) across unrelated specs: `ServiceEntityVersionPage`, `TasksUIFlow`, `Tag`, `LineageInteraction`, `LineageFilters`, `InputOutputPorts`, `Glossary`, `ExplorePageRightPanel`, `WorkflowOssRestrictions`, `PersonaAIContextRules`, `DataQuality`, `ContextCenterArticles`, `ClassificationImportExport`.</sub>

PR #32384, opened 2026-09-01
perf(ui): switch build to Rolldown — 10 min → 20 s

chirag-madlani:
## Summary

- **Full clean `yarn build` drops from ~628 s (10:28) to ~20 s** — a ~28× speedup, ~50× on the Vite phase alone. All existing bundle-check budgets pass.
- Wall-time win comes from three levers stacked together: swap Rollup → **[Rolldown](https://rolldown.rs/)** (Rust bundler; parallel), swap Babel → **SWC** for the React/JSX transform, and remove the in-Rollup **brotli** pass that serialized 1200 chunks worth of compression on every build.
- Runtime is unchanged: same vendor buckets, same bootstrap size (957 KiB brotli — within the same 970 KiB budget), same largest chunk. Chunk count is higher (1237 vs 671) because Rolldown emits at every dynamic-import boundary and has no equivalent of Rollup's \`experimentalMinChunkSize\` merger yet — see the trade-off notes below.

## Numbers

| Metric | Rollup baseline | Rolldown + fixes |
|---|---:|---:|
| \`yarn build\` wall time | 628 s (10:28) | **~20 s** |
| Vite \`built in\` timer | 614 s (10:14) | **12–14 s** |
| Emitted JS chunks | 671 (post-merge) | 1237 |
| Chunks < 20 KiB | 500 | 1079 |
| Bootstrap brotli | 959 KiB | 957 KiB |
| Largest chunk | 1.59 MB | 1.64 MB |
| CPU utilization | 91 % (1 core) | 249 % (2.5 cores) |

### Build configuration
- Swap \`@vitejs/plugin-react\` (Babel) → \`@vitejs/plugin-react-swc\`. Same zero-arg \`react()\` call site; SWC is faster on JSX transform.
- Alias \`vite\` → \`npm:rolldown-vite@latest\` in \`package.json\`. Installs \`rolldown@1.0.0-beta.53\` + per-platform NAPI-RS binding; the config surface stays the same.
- Add \`esbuild\` as an explicit dev-dep — \`rolldown-vite\` deprecated \`transformWithEsbuild\` but downstream plugins still call it; keep the compat shim happy until the ecosystem migrates to Oxc.
- Extract the vendor classifier to a shared \`classifyChunk\` const. Register it under **both** \`rollupOptions.output.manualChunks\` (Rollup fallback) and \`advancedChunks.groups[].name\` (Rolldown native). Swapping bundlers is one line in \`package.json\`.
- Remove the second \`viteCompression\` (brotli) pass. \`OpenMetadataAssetServlet\` falls back to \`.gz\` when \`.br\` is absent; the bundle-check now brotli-compresses the ~8 bootstrap files **in-memory** via \`zlib.brotliCompressSync\` so the 970 KiB budget stays enforced without paying to compress 1200 chunks in Rollup's \`writeBundle\`.
- Bump ratchets in \`scripts/check-bundle-chunks.mjs\` from 720/540 → 1400/1250 to accept Rolldown's un-merged chunk count. Bootstrap brotli, largest chunk, and bootstrap-file-count ratchets unchanged.

### Static assets
- **Move \`src/jsons/connectionSchemas/\` (195 JSON files, 6.5 MB) to \`public/jsons/connectionSchemas/\`.** Pure data — no reason for the bundler to walk it. Vite copies \`public/\` verbatim to \`dist/\`; the Java asset servlet serves them from the same root as \`favicons/\`. \`parseSchemas.js\` now writes directly to the new location.
- **Add \`src/utils/loadConnectionSchema.ts\`** — bounded LRU-cached \`fetch()\` loader that reads \`\${basePath}/jsons/connectionSchemas/...\` at runtime with \`getBasePath()\` for path resolution, and dedupes concurrent callers via promise memoization.
- **Rewrite 11 service utility files** (Database/Storage/Drive/Dashboard/MlModel/Messaging/Search/Pipeline/Security/API/Metadata) to call \`loadConnectionSchema()\` instead of the old dynamic \`import()\`.
- **Update 4 test files** to mock \`loadConnectionSchema\` (instead of the removed JSON module paths) and repoint assertion fixtures at \`public/\`.

### Template-literal \`import()\` chunk fragmentation

Three call sites globbed into hundreds of chunks because Vite must emit one chunk per possible match of a template-literal \`import()\`:

- **\`ApplicationsClassBase.importAppLogo\`** — \`import(\\\`../../../../assets/svg/\${appName}.svg\\\`)\` matched every SVG in \`assets/svg/\` (**799 files**) despite only ~9 files (\`*Application.svg\`) being real logos. Rewritten as \`import.meta.glob('.../assets/svg/*Application.svg', { query: '?react' })\`. Same lazy behaviour, chunk count for this one line drops from ~799 to 9.
- **\`useImage\`** — \`import(\\\`../assets/img/\${fileName}.png\\\`)\` matched every PNG under \`assets/img/**\` (~106) while the only two callers use \`useImage(\\\`emojis/\${reaction}\\\`)\`. Rewritten as \`import.meta.glob('../assets/img/emojis/*.png', { eager: true, query: '?url', import: 'default' })\` — Vite emits static asset URLs with **zero JS chunks** and the hook returns synchronously.
- **\`ApplicationsClassBase.importAppScreenshot\`** — same treatment: eager URL glob.
- **\`ApplicationsClassBase.importSchema\`** — narrowed to a typed \`import.meta.glob\` so the emit graph is transparent to reviewers.

## Trade-offs & notes for reviewers

**Rolldown is beta.** \`rolldown@1.0.0-beta.53\`. Vue, Nuxt, and Storybook ship production builds on it. During this work I hit one intermittent Less plugin error on \`EntityRightPanelVerticalNav.less\` that cleared on \`rm -rf node_modules/.vite\` + rebuild; if CI hits it, that's the workaround until Rolldown stabilizes.

**Brotli removal is a small transfer-size regression on raw servlet deployments** — modern browsers get gzip instead of brotli, roughly 5–15 % larger transfer per JS/CSS asset. Behind a CDN/nginx with on-the-fly brotli it's invisible. Reintroducing pre-compressed brotli as a parallel post-step is a straightforward follow-up if the tighter transfer size is required for published artifacts.

**Chunk count went from 671 to 1237.** Rolldown emits one chunk at every dynamic-import boundary and stops — there is no \`experimentalMinChunkSize\` merger. The remaining ~566-chunk delta breaks down roughly as:
- ~500 route-lazy chunks: **intentional**, working as designed.
- ~150 shared-module extractions: **correct**, prevents duplication.
- ~330 near-empty tree-shake stubs: **fixable at the app level** — mostly \`React.lazy\` around enum/interface files that end up 30–100 bytes. Two follow-up paths noted below; both need per-file judgment across hundreds of files so they intentionally stay out of this PR.

Runtime impact is small (HTTP/2 handles many small requests fine), byte size is at least equal to Rollup's output (no duplication), cacheability is arguably better (smaller chunks change less often). The main downside is the ratchet number in \`check-bundle-chunks.mjs\` output looks larger.

**Follow-up options for closing the chunk-count gap** (not blocking):
1. Enable \`@typescript-eslint/consistent-type-imports\` with \`separate-type-imports\` autofix — collapses \`import { X, Y } from './x.interface'\` into a \`import type { X }\` + \`import { Y }\` split so pure-type sides get stripped at SWC transform. I prototyped this in-session (3200-file diff, zero measurable chunk win because virtually every \`.interface.ts\` in the codebase also has an \`enum\`) — reverted.
2. Split enums out of \`.interface.ts\` files into their own \`.enum.ts\` file. Mechanical, one file at a time. Expected chunk drop: ~130.
3. Replace \`enum\` with \`as const\` objects. Deeper syntactic change; combines nicely with Rolldown's \`optimization.inlineConstants\` for further wins.

## Test plan

- [ ] Local \`yarn install\` succeeds (fetches \`rolldown-vite\` + \`@rolldown/binding-*\` for the platform; may need to remove \`yarn.lock\` and re-generate on Linux CI to record the Linux binding).
- [ ] \`yarn build\` completes in <60 s, bundle-check green.
- [ ] \`yarn test\` — updated Database/Search/Drive/Messaging service-util tests still pass.
- [ ] Smoke-test in browser: login → Services → Add → **MySQL/Postgres form** (exercises \`loadConnectionSchema\`), Applications settings page (exercises \`importAppLogo\` and \`importAppScreenshot\`), Activity Feed with an emoji reaction (exercises \`useImage\`).
- [ ] Playwright suite passes end-to-end (should be unaffected — no runtime behaviour changes).


chirag-madlani:
### CI status update

Two follow-up commits landed:

- `e4a14a4b` **fix(ui): CI fixes for Rolldown migration** — fixes the failures visible in the initial run:
  - **`esbuild` missing on CI** (was affecting `playwright__build` and `maven-collate-ci`): added to `devDependencies` and recorded in the lockfile. It was previously only in `resolutions` (a version pin), which yarn 1 doesn't treat as a dependency.
  - **Jest cannot parse `import.meta.glob`** (was affecting `ui-coverage-tests`, cascading to `ui-coverage` + `ui-sonar-gate`): the Vite-only syntax is now isolated in sibling `*.assets.ts` wrapper files, and `jest.config.js` `moduleNameMapper` redirects them to a stub that returns `{}` for every named export. Consumers hit their existing empty-map fallback branches — no runtime change.
  - **Formatting** (was affecting `ui-checkstyle`): ran the exact CI sequence — `organize-imports-cli` → `eslint --fix` → `prettier --write` — on the changed files. Three service utils needed prettier reformats.

### `Validate PR Metadata` — needs maintainer action

This check requires a linked issue that is assigned to the `Shipping` project with `Status`, `Source`, `Priority`, `Domain`, and `Release` fields set. There's no pre-existing issue for this build-tooling work — it grew out of the build-performance investigation on this branch. Happy to open a corresponding issue with the required metadata if a maintainer wants it linked, otherwise please waive the check on merge.

karanh37 (dismissed):

karanh37 (approved):

PR #32448, opened 2026-09-02
fix(tags): make provider immutable for system Classifications/Tags (#29974)

### LINKED ISSUE #29974: Prevent modification of provider for system Classifications/Tags via PUT API
### Affected module

Backend

### Describe the bug

The PUT API currently allows updating the provider field of system classifications and tags from system to user. This results in an inconsistent metadata state where some default system tags can be deleted while others cannot, leaving partially modified system classifications.

The provider field for system metadata should be immutable and should not be updatable through the API.

### To Reproduce

The PUT API allows changing the provider for some system classifications/tags.

### Expected behavior

- System classifications and tags should have an immutable provider field.
- Any attempt to update provider on system metadata should be rejected with an appropriate validation error (e.g. 400 Bad Request).
- The API should prevent converting system metadata into user metadata.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

1.13.1

_No response_

### Additional context

_No response_

### Pre-submission checklist

sonika-shah:
## Describe your changes

Fixes #29974.

The PUT/PATCH API allowed changing a **system** Classification/Tag's `provider` from `system` to `user`. Once downgraded, the entity lost its delete/rename protection (`checkSystemEntityDeletion`), so it could be deleted — leaving the instance with some default system tags removed and others stuck, a partially-modified, inconsistent metadata state.

### Fix
A shared guard in `EntityRepository.EntityUpdater` (`restrictSystemProviderChange`), wired into the Classification and Tag updaters:

- **PATCH** that changes a system entity's provider → rejected with **400** (`systemEntityModifyNotAllowed`), the same exception family as the existing system delete/rename guards.
- **PUT** — a `CreateClassification`/`CreateTag` defaults `provider` to `user` when omitted, so a naive reject would break the legitimate "disable a system tag" flow. Instead the system provider is **preserved silently** (matching the existing `preserveAutoClassificationConfigOnPut` / `preserveRecognizerConfigOnPut` idiom), which still prevents the downgrade.

User-provider entities are unaffected; the guard only engages when the stored entity is `system`.

## Type of change

## How was this tested

Added integration tests (run against the embedded stack — **all 4 pass**):

| Test | Asserts |
|---|---|
| `ClassificationResourceIT.patch_systemClassificationProvider_rejected` | PATCH `Tier` provider→user ⇒ 400, provider stays `system` |
| `ClassificationResourceIT.put_systemClassificationProvider_preserved` | PUT `Tier` with provider=user ⇒ provider preserved as `system` |
| `TagResourceIT.patch_systemTagProvider_rejected` | PATCH `Certification.Gold` provider→user ⇒ 400, provider stays `system` |
| `TagResourceIT.put_systemTagProvider_preserved` | PUT `Certification.Gold` with provider=user ⇒ provider preserved as `system` |

```
mvn -pl openmetadata-integration-tests test -Dtest='ClassificationResourceIT#patch_systemClassificationProvider_rejected+put_systemClassificationProvider_preserved,TagResourceIT#patch_systemTagProvider_rejected+put_systemTagProvider_preserved'
=> Tests run: 4, Failures: 0, Errors: 0
```

## Checklist

- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/ClassificationResourceIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/ClassificationResourceIT.java
@@ -276,6 +276,60 @@ public class ClassificationResourceIT extends BaseEntityIT<Classification, Creat
   }
 
   @Test
+  void patch_systemClassificationProvider_rejected(TestNamespace ns) throws Exception {
+    // #29974: flipping a system classification's provider to user via PATCH must be rejected -
+    // otherwise it loses its delete protection and system metadata can be removed, leaving the
+    // instance in a partially-deleted inconsistent state.
+    Classification tier = SdkClients.adminClient().classifications().getByName("Tier");
+    assertEquals(ProviderType.SYSTEM, tier.getProvider());
+
+    JsonNode providerToUser =
+        new ObjectMapper()
+            .readTree("[{\"op\":\"replace\",\"path\":\"/provider\",\"value\":\"user\"}]");
+    assertThrows(
+        InvalidRequestException.class,
+        () ->
+            SdkClients.adminClient()
+                .getHttpClient()
+                .execute(
+                    HttpMethod.PATCH,
+                    "/v1/classifications/" + tier.getId(),
+                    providerToUser,
+                    Classification.class),
+        "Changing the provider of a system classification should be rejected");
+
+    Classification reloaded = SdkClients.adminClient().classifications().getByName("Tier");
+    assertEquals(
+        ProviderType.SYSTEM, reloaded.getProvider(), "System provider must remain unchanged");
+  }
+
+  @Test
+  void put_systemClassificationProvider_preserved(TestNamespace ns) throws Exception {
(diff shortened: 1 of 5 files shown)


sonika-shah:
Addressed the Copilot review:
- **PUT tests (both Classification & Tag):** now reload the entity after the PUT and re-assert `provider == SYSTEM`, so the test proves the provider is *persisted*, not just echoed in the response.
- **`restrictSystemProviderChange` JavaDoc:** reworded to be accurate — a PATCH applies the client's ops to the stored entity (so `updated` is a complete entity; an unpatched provider stays `SYSTEM`, never null), and a provider that differs from the original was explicitly changed. The PUT-vs-PATCH distinction is now stated as "rebuilt-from-Create-with-a-default" vs "mutated-in-place" rather than "only carries patched fields".

All 4 ITs still pass (`Tests run: 4, Failures: 0, Errors: 0`).

yan-3005 (approved):

PR #32449, opened 2026-09-02
Fixes #15274: Parse Protobuf schemas with unmatched topic names

### LINKED ISSUE #15274: Ingestion: Kafka protobuf schema parsing only works if payload type is named the same as the topic
**Affected module**
Ingestion

**Describe the bug**
When ingesting from Kafka, schema parsing fails unless the payload type name is similar to the topic name. For example, in a topic named `address_book` the payload module must be named `AddressBook`. Otherwise the parsing fails with a warning similar to this:

```[2024-02-20T16:10:40.336+0000] {common_broker_source.py:102} INFO - Fetching topic schema loans
[2024-02-20T16:10:40.346+0000] {common_broker_source.py:104} INFO - Fetching topic config loans
[2024-02-20T16:10:40.497+0000] {protobuf_parser.py:165} WARNING - Unable to create protobuf python module for loans: module 'loans_pb2' has no attribute 'Loans'
[2024-02-20T16:10:40.499+0000] {protobuf_parser.py:200} WARNING - Unable to parse protobuf schema for loans: 'NoneType' object has no attribute 'DESCRIPTOR'
```
In the OMD UI the result is that only the text version of the schema appears. No fields are displayed and there's no way to add tags or glossary terms to the schema:

image

**To Reproduce**
1. In Kafka, create a topic. 
2. Using schema registry, create a Protobuf schema for the topic but give the top-level message a name **other than** the topic name. 
3. Run an ingestion against the Kafka cluster.
4. In the UI the topic will only have the Protobuf schema in text format.

**Expected behavior**
The full schema with fields, tags, and glossary terms should be displayed for the topic.

**Version:**
 - OS: Linux
 - Python version: 3.10.13
 - OpenMetadata version: 1.3.0
 - OpenMetadata Ingestion package version: `openmetadata-ingestion[docker]==1.3.0.1

**Additional context**
N/A

IceS2:
### Describe your changes:

Fixes #15274

The Protobuf schema parser assumed that a generated message name could always be derived from the Kafka topic name. When a valid schema contained one differently named top-level message, ingestion stored the raw schema text but left `schemaFields` empty.

This change preserves the topic-derived lookup for compatibility and falls back to the sole declared top-level message. It refuses to guess when multiple unmatched messages exist. It also resolves the compiler include path consistently so filesystem aliases do not prevent parsing.

#
### Type of change:

- [x] Bug fix

#
### High-level design:

N/A — small ingestion parser fix.

#
#### Use cases covered

- A `loans` topic with a single `MyLoanRecord` Protobuf message ingests its fields.
- Multiple unmatched top-level messages remain ambiguous and are not parsed by guessing.

#### Unit tests

- [x] Added focused parser regression coverage in `ingestion/tests/unit/test_protobuf_parser.py`.

- [x] Added a Kafka and Schema Registry workflow test that persists the topic through the OpenMetadata REST sink and verifies its schema fields.

#### Playwright (UI) tests

#### Manual testing performed

- Ran the Kafka ingestion workflow against a local OpenMetadata server with Testcontainers Kafka and Schema Registry.

#

Not applicable.

#
### Checklist:


ingestion/src/metadata/parsers/protobuf_parser.py
```diff
@@ -179,6 +190,8 @@ def parse_protobuf_schema(self, cls: type[BaseModel] = FieldModel) -> list[Field
         try:
             proto_path, file_path = self.create_proto_files()
             instance = self.get_protobuf_python_object(proto_path=proto_path, file_path=file_path)
+            if instance is None:
+                return None
```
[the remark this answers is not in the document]
IceS2: Fixed in `d568af583e`. `parse_protobuf_schema()` now handles `create_proto_files()` returning `None` before unpacking and cleans `base_file_path` in a `finally` block. The ambiguity regression test now verifies that the temporary directory is removed.

ingestion/tests/integration/kafka/conftest.py
```diff
@@ -4,6 +4,8 @@
 
 import pytest
 import testcontainers.core.network
+from confluent_kafka.admin import AdminClient, NewTopic
+from confluent_kafka.schema_registry import Schema, SchemaRegistryClient
```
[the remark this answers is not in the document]
IceS2: This Kafka suite already imported `_openmetadata_testutils.kafka.load_csv_data` during collection before this PR, and that module unconditionally imports `confluent_kafka`. `confluent_kafka` is also part of the Kafka connector extra, so these direct imports do not change collection behavior. Making the suite skip without Kafka dependencies would require guarding the pre-existing helper import and should be handled separately.

ingestion/src/metadata/parsers/protobuf_parser.py
```diff
@@ -162,55 +165,68 @@ def get_protobuf_python_object(self, proto_path: str, file_path: str):
...
             message = importlib.import_module(module_name)
 
-            # get the class and create a object instance
-            class_ = getattr(message, snake_to_camel(self.config.schema_name))
+            message_types = message.DESCRIPTOR.message_types_by_name
+            message_descriptor = message_types.get(snake_to_camel(self.config.schema_name))
+            if message_descriptor is None and len(message_types) == 1:
+                message_descriptor = next(iter(message_types.values()))
+            if message_descriptor is None:
+                logger.warning(
```
TeddyCr: Should this be a debug log so if we fall into this path across a large number of messages we don't end up with lot of warn lines?

ingestion/src/metadata/parsers/protobuf_parser.py
```diff
@@ -133,7 +136,7 @@ def create_proto_files(self):
             # Create a .proto file under the interfaces directory with schema text
             with file_path.open("w", encoding="UTF-8") as file:
                 file.write(self.config.schema_text)
-            proto_path = "generated=" + self.proto_interface_dir
+            proto_path = "generated=" + str(proto_interface_dir_path.resolve())
```
TeddyCr: Two things this line makes visible. Both pre-existing.

`resolve()` follows symlinks, and the guard in `_get_proto_file_path` compares `file_path.parent` against the *resolved* `interface_dir`. If a local user pre-creates `interfaces` as a symlink to a directory of their choosing, both sides resolve to that same target, the comparison passes, and `mkdir(..., exist_ok=True)` accepts the pre-existing symlink. 

```
create_proto_files -> ('generated=/…/attacker_1bcx31lh/interfaces',
                       '/private/…/victim_m8vn942t/loans.proto')
wrote into victim dir: ['loans.proto']
```

The file-symlink case *is* correctly rejected (`Protobuf schema path escapes the interface directory`) — it's specifically a symlinked **directory** that slips through. It's only reachable because `base_file_path` defaults to a fixed `/tmp` path (line 88), so an attacker knows where to plant it ahead of time. `O_EXCL|O_NOFOLLOW` on the write, or a freshly-created private temp dir, is the actual fix — the `resolve()` comparison is TOCTOU-racy either way, since it runs before the `mkdir` and well before the `open("w")`.

Second, unrelated to symlinks: the `generated=` alias makes the descriptor file name `generated/<schema_name>.proto` for *every* parse, process-wide. Two topics with the same name under different services register two different `generated/loans.proto` into the default descriptor pool, and the second raises `duplicate file name`. Making the directory unique doesn't help, because the collision is in the virtual path — it needs a per-parse nonce in the alias, or a private `descriptor_pool.DescriptorPool()`.

ingestion/src/metadata/parsers/protobuf_parser.py
```diff
@@ -162,55 +165,68 @@ def get_protobuf_python_object(self, proto_path: str, file_path: str):
             module_name = Path(py_file).stem
             message = importlib.import_module(module_name)
```
TeddyCr: `sys.path.append(self.generated_src_dir)` (line 162) is never undone, so a predictable, world-writable `/tmp` directory stays on `sys.path` for the rest of the worker's life. The `rmtree` in `parse_protobuf_schema` removes the files but not the path entry, and the directory can simply be recreated. It's appended rather than inserted, so it only shadows modules that aren't already resolvable — but that still covers every lazily-imported optional module the process hasn't touched yet.

`import_module` also consults `sys.modules`, so the second topic named `loans` seen in a process gets the *first* one's `loans_pb2` back rather than the freshly compiled one. `importlib.invalidate_caches()` is never called after the new file is written either.

Both go away by loading from the file directly and never touching `sys.path`:

```python
spec = importlib.util.spec_from_file_location(unique_name, py_file)
module = importlib.util.module_from_spec(spec)
spec.loader.exec_module(module)
```

ingestion/src/metadata/parsers/protobuf_parser.py
```diff
@@ -162,55 +165,68 @@ def get_protobuf_python_object(self, proto_path: str, file_path: str):
...
-            # Clean up the tmp folder
-            if Path(self.config.base_file_path).exists():
-                shutil.rmtree(self.config.base_file_path)
-
-            return field_models  # noqa: TRY300
         except Exception as exc:  # pylint: disable=broad-except
             logger.debug(traceback.format_exc())
             logger.warning(f"Unable to parse protobuf schema for {self.config.schema_name}: {exc}")
+        finally:
+            shutil.rmtree(self.generated_src_dir, ignore_errors=True)
```
TeddyCr: One caveat worth being aware of: because `base_file_path` is a fixed constant (line 88), every parser on the host shares this directory, so this `rmtree` also deletes *other in-flight parsers'* files. Two concurrent Airflow tasks parsing Protobuf topics will intermittently destroy each other's `.proto` / `_pb2.py`, and it surfaces as "Unable to create protobuf python module" — which reads like a malformed schema rather than a race, so it's the kind of thing that gets debugged in the wrong place.

Scoping the directory per parse (`tempfile.TemporaryDirectory`) makes this and the symlink issue in my other comment both moot.

ingestion/src/metadata/parsers/protobuf_parser.py
TeddyCr, line 1: Anchoring at file level because line 88 isn't in this PR's diff — `base_file_path: str | None = "/tmp/protobuf_openmetadata"` is the root cause of the three inline comments I left, and it's pre-existing.

The path is a compile-time constant in a world-writable directory, and the inputs reaching it are attacker-influenced end to end: `schema_name` is the Kafka topic name and `schema_text` is registry content, which then get written to disk, compiled, and **imported**. Because the path is predictable it can be pre-planted; because it's shared it makes concurrent workers collide.

Defaulting to a private per-parse directory addresses the symlink bypass, the cross-worker collision, and the `/tmp` permissions exposure at once — and removes the need for a traversal check at all, since a freshly-created `0700` directory can't be pre-planted:

```python
base_file_path: str | None = None  # None -> tempfile.mkdtemp(prefix="protobuf_om_")
```

That would also close a smaller footgun on this line: the default only applies when the field is *omitted*, not when it's explicitly `None`. Passing `base_file_path=None` today gives `self.proto_interface_dir == "None/interfaces"` — a relative directory created in the process CWD:

```
None -> None/interfaces | absolute? False
```

Two unrelated small ones, since I was already in the file:

- `grpc_tools.protoc.main()` returns a nonzero exit code on failure and it's discarded, so a compile error surfaces as an `IndexError` from `glob(...)[0]` and gets reported as the generic "Unable to create protobuf python module". Checking the return value would let the real protoc error reach the log — probably worth it given this PR is specifically about schemas that previously failed silently.
- `schema_name` is validated as a single path component, but not as a Python identifier, even though it becomes a module name via `Path(py_file).stem` and a class name via `snake_to_camel`.

ulixius9 (approved):

Issue #32450, opened 2026-09-02
cassandra & couchbase: test connection probes an arbitrary keyspace/bucket, ignoring the configured scope


Khairajani:
### Describe the bug

Same class of defect as #28142 (MongoDB), in two more connectors: a mandatory test-connection step has to pick **one** keyspace/bucket out of many, and it takes whichever one the server returned first — ignoring the pin and filter pattern that say what the ingestion will actually read.

**Cassandra** — `ingestion/src/metadata/ingestion/source/database/cassandra/connection.py`

```python
def test_get_schemas(session, holder_):
    for keyspace in session.execute(CASSANDRA_GET_KEYSPACES):
        holder_.schema = keyspace.keyspace_name
        break
```

`CASSANDRA_GET_KEYSPACES` is a bare `SELECT * FROM system_schema.keyspaces`, so the keyspace `GetTables` / `GetViews` then probe is almost always `system` or `system_auth` — never the user keyspaces ingestion reads. The step is mandatory and proves nothing about the data being ingested. `schemaFilterPattern` exists on the connection and is not consulted. Separately, when the listing comes back empty `holder_.schema` stays `None` and `CASSANDRA_GET_KEYSPACE_TABLES` runs with `None` bound as `keyspace_name`.

**Couchbase** — `ingestion/src/metadata/ingestion/source/database/couchbase/connection.py`

```python
def test_get_databases(client, holder):
    for database in buckets.get_all_buckets():
        holder.database = database.name
        break
```

First bucket wins. The connection's `bucket` field pins ingestion to a single bucket (`CouchbaseSource.get_database_names`), but the probe ignores it — so a config scoped to one bucket is still tested against a different one.

### Expected behavior

The probe should target what the ingestion would read: the configured pin when set (`databaseSchema`-equivalent for cassandra keyspaces, `bucket` for couchbase), otherwise the candidates the filter pattern keeps, probing them until one succeeds. This is what `AthenaChecks._targeted_schemas()` and the Glue checks already do, and what #28142 fixed for MongoDB.

### Wider audit

While confirming the scope, I went through every database connector to see how each one picks a probe target. Legend: **IMMUNE** = no choice made (URI-pinned or catalog-wide probe); **ARBITRARY** = first item wins; **SCOPED** = honours pin/filter.

| Status | Connectors | Where |
|---|---|---|
[...]

PR #32382, opened 2026-09-01
Fixes #28142: probe only the in-scope databases in the MongoDB test connection

Khairajani:
### Describe your changes:

Fixes #28142

`GetCollections` listed the collections of an arbitrary database — whichever one `list_database_names()` returned first — so a MongoDB user holding `listCollections` only on the databases it ingests failed a mandatory test-connection step for a configuration that would have ingested fine. The step description said as much: *"a randomly chosen database"*.

- `GetDatabases` now resolves the databases the ingestion would actually read: `databaseSchema` when set, otherwise the listed databases minus the ones `schemaFilterPattern` filters out (MongoDB databases are ingested as OpenMetadata schemas). Same idea as the Snowflake test connection, which already picks its probe database through `databaseFilterPattern`.
- `GetCollections` probes those databases and passes on the first success instead of betting the whole test on one of them. Nothing in scope is no longer a failure either — an empty instance, or filters that exclude everything, is not a connection problem.
- Updated the step description / error message, which described the old behaviour.
- `GetCollections` also fails explicitly when `GetDatabases` could not list at all, so a failed listing cannot leave it reporting success.

The couchbase connector has the same one-arbitrary-database probe shape; not touched here.

The pipeline-level `sourceConfig` filters are not reachable from `test_connection` (it only receives the service connection), so the connection-level `Database Schema` / `Schema Filter Pattern` fields are what the probe can honour.

#
### Type of change:

- [x] Bug fix

#
### High-level design:

N/A — small change.

#
#### Use cases covered
- A user with `listCollections` on only one database passes the test connection when the other databases are excluded by `Schema Filter Pattern`, or included by it, or when no filter is set at all.
- The test connection still fails when no database in scope can be read.
- `Database Schema` remains the single probed database and skips listing entirely.

#### Unit tests
- [x] Added, in `ingestion/tests/unit/source/database/mongodb/test_connection.py`. Four of them fail on `main`.

#### Backend integration tests
- [x] Not applicable (no API changes).

#### Ingestion integration tests
- [x] Not applicable — covered by unit tests plus the manual run below.

#### Playwright (UI) tests
- [x] Not applicable (no UI changes).

#### Manual testing performed
[...]

Issue #28142, opened 2026-05-15
mongodb connector: connection test does not respect filtering of DBs, collections, tables, etc


rc-donalhunt:
### Connector

mongodb

### Feature area

Test Connection

### Summary
In an enterprise environment, it's common for security controls to be set higher than in other environments.

The following scenario currently causes the mongodb openmetadata connector to fail (even if ingestion would be successful):
- mongodb instance has authentication enabled
- AND the openmetadata ingestion config explicitly filters what DBs, collections, etc are ingested
- AND credentials used for mongodb metadata ingestion does not have permission to run listCollections on all databases on the mongoDB instance.

The root cause appears to be that `test_connection()` in [ingestion/src/metadata/ingestion/source/database/mongodb/connection.py](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/database/mongodb/connection.py) doesn't respect the filtering specified in the ingestion config — it blindly attempts to get collections from all databases on the mongodb instance, even when the credentials do not have permission to do so.

Not clear if this is by design (i.e. conscious decision that enumeration of all collections is mandatory), inflexibility of the current connection test framework or a bug / oversight specific to the mongoDB connector.

Current workarounds:
 - grant the user permissions to run listCollections on all databases
 - comment out line 80 in [ingestion/src/metadata/ingestion/source/database/mongodb/connection.py](https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/src/metadata/ingestion/source/database/mongodb/connection.py) (keeps the same security posture on the mongoDB side).

### To Reproduce

```
$ mongosh "mongodb://<credentials>@mongohost.example.com:27017/?authSource=admin
Current Mongosh Log ID: 6a06f1d5bc1911fac29df8a2
Connecting to:          mongodb://<credentials>@mongohost.example.com:27017/?authSource=admin&directConnection=true&appName=mongosh+2.8.3
Using MongoDB:          6.0.16-13
Using Mongosh:          2.8.3

For mongosh info see: https://www.mongodb.com/docs/mongodb-shell/

pm1 [direct: other] test> show dbs;
__mongo_connector    5.81 MiB
admin              388.00 KiB
local               30.83 GiB
test                48.00 KiB

pm1 [direct: other] test> use admin;
switched to db admin

pm1 [direct: other] admin> show collections
[...]

PR #32454, opened 2026-09-02
Fixes #32453: share the in-scope probe selection across test-connection checks

### LINKED ISSUE #32453: test connection: share the in-scope probe selection instead of reimplementing it per connector
### Describe the improvement

Several test-connection steps have to read *one* schema (keyspace, bucket, dataset) out of many. Three connectors resolve that target properly and independently — `AthenaChecks._targeted_schemas()` (filter-aware, memoized, capped, empty-as-caveat), the Glue checks (probe databases until one exposes a table), and snowflake's `_init_database` (skips `databaseFilterPattern` exclusions). Everywhere else the choice is "whichever object the server listed first", including in the two helpers that most connectors route through:

- `core/connections/test_connection/checks/database.py::_resolve_schema` — first non-system schema, filter-unaware (used by postgres, mysql, mssql)
- `ingestion/connections/test_connections.py::test_connection_db_schema_sources.custom_executor` — its docstring says "a given schema **or a random one**"; skips only `information_schema` / `performance_schema` (used by doris, hive, impala, iomete, mariadb, my_db, singlestore, starrocks, trino)

Consequences:

- The probe reports on an object the ingestion never reads, so a mandatory step passes having proved nothing about the data in scope.
- On an engine fronted by an external authorizer (Hive or Trino with Ranger, Unity Catalog) an unauthorized object **raises** rather than returning an empty list, so the step fails for a configuration whose ingestion would have worked. That is #28142 (MongoDB) and #32450 (cassandra, couchbase).

### Expected behavior

One shared place that answers "what should this check probe?", used by the shared helpers so every connector routing through them inherits it, and available to bespoke providers:

1. the configured pin (`databaseSchema`, `bucket`, ...) when set — it is the only object the run reads;
2. otherwise the objects the filter pattern targets, with system objects deferred rather than dropped (ingestion reads those too);
3. capped, because each target costs a round-trip;
4. probed until one answers — only every target refusing the read is a failure;
5. nothing in scope is a caveat, not a failure.

This is the base change; the per-connector migrations (cassandra, couchbase, mongodb, and the remaining arbitrary-probe connectors listed in #32450) follow on top of it, one PR each.

Khairajani:
### Describe your changes:

Fixes #32453

Base change for the probe-selection work. A test-connection step that has to read *one* schema out of many was picking whichever one the server listed first, in each connector's own way. Athena, Glue and Snowflake had each solved this independently; the two helpers most connectors route through had not — so the probe landed on an object the ingestion never reads, and on an engine fronted by an external authorizer (Hive or Trino with Ranger, Unity Catalog) refusing that object failed a connection whose ingestion would have worked.

- New `checks/probe.py`: `probe_targets` tries candidates until one answers, so only *every* target refusing the read is a failure. What counts as an answer is the caller's — the probe returns a value to accept a target and `None` to keep looking — because a privilege probe only needs the call not to raise, while a "is there data" probe (Glue, Athena) needs a non-empty result.
- Selection lives in `checks/database.py`, which already owns the database-specific knowledge, as a single `targets_in_scope(names, pinned=..., excluded=..., limit=...)`: a configured pin wins outright, otherwise the filter pattern decides, capped so a wide catalog cannot exhaust the step timeout. Which names are the connector's system objects — and whether ingestion reads them — stays with the connectors that declare them.
- `list_tables` / `list_views` keep their existing signature plus an optional `schema_filter`, and now survive a target that raises. postgres, mysql and mssql pass their `schemaFilterPattern` through.
- The legacy `test_connection_db_schema_sources` builds a scope from the service connection, which covers doris, hive, impala, iomete, mariadb, my_db, singlestore, starrocks and trino without touching those connectors.
- `GetCollections` joins the `DatabaseStep` enum — mongodb.json and couchbase.json both use that step name and it was missing.
- `metadata.utils.filters.filtered_out` exposes the entity-agnostic form the `filter_by_*` helpers already delegate to, for objects that are not a modelled entity (a keyspace, a bucket).

Per-connector migrations follow on top of this, one PR each: cassandra and couchbase (#32450), mongodb (#28142, currently fixed in place by #32382 for backportability), then the remaining arbitrary-probe connectors listed in #32450 (databricks, unitycatalog, bigtable, presto, saphana).

#
### Type of change:

- [x] Improvement

#
### High-level design:

The shared helper answers one question — "what should this check probe?" — for both the new checks framework and the legacy path, so a connector no longer has to rediscover the answer.

For a connector that configures nothing, behaviour is unchanged: system schemas are still dropped from the candidates, and when the listing leaves none the probe still falls back to the connection's default schema.

The two intended changes: a target that *raises* no longer ends the step (the rest in scope are tried, and only all of them refusing is a failure), and a filter pattern configured on the service connection now narrows the candidates — where it excludes everything, the step reports a caveat instead of probing an excluded schema.

#
#### Use cases covered
- A configured pin is the only object probed, and needs no listing.
- Objects excluded by the filter pattern are never probed; only included ones are.
- System objects are deferred, not dropped, so a system-only catalog still gets probed.
- A target that refuses the read is followed by the next one; every target refusing it fails the step.
- Nothing in scope reports a caveat rather than failing.

#### Unit tests
- [x] `tests/unit/core/connections/test_connection/test_scope_checks.py` (new) and `test_database_checks.py` (extended, including a schema that raises).

- [x] Not applicable (no API changes).

- [x] Not applicable — covered by unit tests.

#### Playwright (UI) tests

#### Manual testing performed
Ran `tests/unit/source/database`, `tests/unit/topology/database`, `tests/unit/core` and `tests/unit/connections` before and after the change: the set of failures is byte-identical (29, all pre-existing — missing optional drivers and a known cross-test pollution pair), with 16 new passing tests.

#

Not applicable.

#
### Checklist:


Khairajani:
The dead `auto_selected` flag the bot flagged is gone: `_targets` now returns just the ordered list, since the follow-up commit reworked it to keep the system schemas dropped (as `main` does) plus the fallback to the connection's default schema. `list_tables`/`list_views` derive `auto_selected` from `scope.pinned`, which is the single source of truth.

IceS2 (commented):
Hey @Khairajani, couple things I would think about before merging this:

1. Separate selection from success
  ProbeScope.targets() selects candidates, while probe_targets() defines “success” as “did not raise.” That is not universal.

2. Keep filtering policy out of the generic primitive
  ProbeScope currently understands:

  - FilterPattern
  - skipped system schemas
  - deferred system objects
  - configured pins
  - target limits

That is database/connector policy inside a supposedly generic core abstraction.


I am wondering if perhaps we can otherwise keep it simpler until we actually migrate the filter pattern and then we can make this better 🤔 
I'm not 100% sure on which approach we should take here

Issue #1252, opened 2021-11-18
UI: Add Ingestion module Unit tests


darth-coder00:

ShaileshParmar11:
Assign to me

ShaileshParmar11:
- [x]  IngestionPage & Ingestion component #1374 
- [ ] Ingestion modal

PR #32458, opened 2026-09-02
Fixes #32457: move Persona AI Context into Context Center, add search-scoped rules

### LINKED ISSUE #32457: Persona AI Context is hidden in Settings and preloads every rule
### Problem

Persona AI Context has two problems today.

**It is hard to find.** The rule editor lives as a tab under Settings → Personas → *persona*, three
levels away from Context Center, where every other AI-context surface (Articles, Documents,
Memories, Archive) lives. Administrators do not discover it.

**Every rule is preloaded.** `PersonaContextBuilder.selectRules` searches for each enabled rule's
entities and renders them into the cached Markdown document injected into every request. A rule
selecting "all Finance tables" therefore spends the character budget on assets the user may never
ask about. The only lever we expose, `alwaysInContext`, reorders rules rather than making any of
them cheaper — so the budget grows with the catalog, not with what the user actually asks.

### Proposal

1. Move the persona AI-context editor into Context Center as an **AI Context** submenu (persona list
   → rule editor) and remove the tab from the persona detail page.

2. Add a third delivery mode to a context rule — **filtered in search**. Instead of preloading its
   entities, the rule contributes its entity type and query filter to a default search scope that
   AI search tools apply on every call. Nothing is materialized, so the rule costs no character
   budget, and the persona still only sees the assets its administrator selected. New rules use this
   mode by default; rules that already exist keep preloading.

### Notes

The scope is a relevance filter, not a permission boundary — OpenMetadata RBAC remains the only
enforcement of what a user may read.

pmbrull:
### Describe your changes:

https://github.com/user-attachments/assets/216ea0b3-e57c-45b5-b484-a106bc11e454

Fixes #32457

I moved the Persona AI Context editor out of Settings → Personas and into Context Center as an **AI
Context** submenu, and added a third delivery mode to a context rule — **filtered in search** —
because every enabled rule was previously materialized into the cached document whether or not the
user ever asked about those assets. A scoped rule preloads nothing; it hands its selection to the AI
search tools as a default filter instead, so the persona still only sees the assets its
administrator selected but the character budget no longer grows with the catalog.

#
### Type of change:
- [x] Improvement

#
### High-level design:

**1. Context Center submenu.** Two new routes on the existing Context Center router —
`/context-center/ai-context` (persona list, rule counts + scoped-rule badge) and
`/context-center/ai-context/<fqn>` (the rule editor). Nav item sits between Memories and Archive and
is registered in `SIDEBAR_NESTED_KEYS` so the parent highlights. `ContextCenterClassBase` gains
`getAIContextListPath()` / `getAIContextPath(fqn)` alongside the existing articles/documents/memories
getters, so Collate's embedded-mode override keeps working. The editor folder moves
`components/Settings/Persona/PersonaAIContext/` → `components/ContextCenter/PersonaAIContext/`
(`git mv`, all four subfolders, relative imports rewritten). The `ai-context` tab is deleted from
`PersonaDetailsPage`.

**2. `filteredInSearch` on `contextRule`.** `selectRules` skips scoped rules entirely — no
Elasticsearch round-trip, no character budget — and a new `searchScope(definition)` unions the
contributing rules into one ready-to-use filter, each rule contributing `entityType AND its own
filter`:

```json
{"query":{"bool":{"filter":[
  {"term":{"deleted":false}},
  {"bool":{"minimum_should_match":1,"should":[
    {"bool":{"filter":[{"term":{"entityType":"table"}}, <rule queryFilter>]}}
  ]}}
]}}
```

**No scoped rules → no filter at all**, and search behaves exactly as before. One or more → a strict
allowlist, which is what makes "this persona only searches tables" expressible.

**Backward compatibility — the missing schema `default` is deliberate.** jsonschema2pojo compiles a
schema `default` into the field initializer, which would make a rule stored before this field
indistinguishable from one explicitly set to `true` and force a data migration over `persona_entity`
JSON. Declaring no default leaves the POJO field `null`, which the server reads as `false`: **every
rule that exists today keeps preloading, with no migration.** `createAiContextRule` stamps `true`
when the client omits the field, so every *new* rule scopes search. The update path deliberately
does not stamp — a `null` there means a legacy rule is being round-tripped and must keep its
behavior. Consequence: the stored default (`false`) and the created default (`true`) differ; the
schema description says so.

`searchScope { entityTypes[], rules[], queryFilter }` is added to `personaContext.json` and served on
the existing `/v1/personas/me/context?format=json`, so the agent needs no new endpoint and no extra
request. Cache invalidation already worked — `PersonaContextHash.definitionHash` serializes the whole
definition, so editing a scoped rule's filter changes the fingerprint even though it changes no
preloaded bytes. `PersonaContextMarkdown` renders a short "Search scope" section; without it the
document gives no hint that searches are narrowed and the model cannot explain a thin result set.

**Editor UI.** Three behaviour cards — Filtered in search / Always in context / Fully rendered.
Turning the first on disables and mutes the other two plus Sections and Max assets, and `handleSubmit`
clears them, so the UI reads as an exclusive three-way choice while the schema stays additive.

**Consumer side** is [ai-platform #1252](https://github.com/open-metadata/ai-platform/pull/1252),
which forwards `searchScope.queryFilter` to Collate's `/v1/hybrid/nlq/search?query_filter=` (that
endpoint already accepts the parameter, so no Collate change is needed).

⚠️ **The scope is a relevance filter, not a permission boundary.** The agent can opt out per call, so
"this persona may only search tables" is advisory. OpenMetadata RBAC remains the only enforcement of
what a user may read. This is stated in the tool description, the prompt, and the schema.

**Alternatives rejected.** A single exclusive `contextMode` enum is a cleaner model but collapses the
existing independent `alwaysInContext` + `fullyRendered` combination and needs a migration to
preserve it. `"default": true` plus a data migration stamping `false` on existing rules makes the
default honest at every layer but requires rewriting `persona_entity` JSON on MySQL and Postgres for
a behavior the create-path stamp already delivers. Per-entity-type pass-through (types with no scoped
rule stay unrestricted) is a safer default but cannot express "only tables".

Design doc: `docs/superpowers/specs/2026-09-02-persona-context-search-scope-design.md`.

#
#### Use cases covered
- An administrator opens Context Center → AI Context, sees every persona with its rule count, and
  clicks through to edit one persona's rules.
- A new rule created in the editor defaults to "filtered in search"; turning it on disables and
  clears Always in context, Fully rendered, Sections and Max assets.
- A persona with a scoped `table` rule serves a `searchScope` on its context document and does not
  preload the matched table.
- A persona whose rules all predate this change is unaffected on upgrade — they keep preloading.

#### Unit tests
- [x] Added unit tests for the new/changed logic.
- `openmetadata-service/.../aicontext/PersonaContextBuilderTest.java` — scoped rules are skipped
  without an ES call; the union is composed per entity type; disabled rules and disabled definitions
  contribute nothing.
- `.../ui/src/utils/PersonaAIContextUtils.test.ts` — a rule stored without the field normalizes to
  preloading; `getScopedRuleCount`.
- `.../ContextRuleEditor/ContextRuleEditor.test.tsx` — new-rule default, existing-rule default,
  mutual exclusivity, and that switching to scoped clears the preload options on submit.
- `.../ContextCenterAIContextPage.test.tsx`, `.../ContextCenterAIContextDetailPage.test.tsx` — both
  new pages, including navigation and empty states.

**Result: backend 37/37 pass** (`PersonaContext*Test`, `PersonaRepositoryTest`); **UI 60/60 pass**
across 11 suites.

- [x] Added `searchScopedRuleIsServedAsASearchFilterInsteadOfBeingPreloaded` to
  `openmetadata-integration-tests/.../mcp/PersonaAIContextIT.java` — asserts a rule created without
  the field comes back `filteredInSearch=true`, that `?format=json` carries `searchScope` with the
  entity type and a non-empty filter, and that the scoped rule is absent from `rules`. Also pinned
  the existing `tableRule` helper to `filteredInSearch=false` so the document assertions stay
  intentional.
- ⚠️ **It compiles but I did not run it** — it needs the Docker stack. Please run
  `mvn test -pl openmetadata-integration-tests -Dtest=PersonaAIContextIT` before merge.

#### Playwright (UI) tests
- [ ] **Not added.** Happy to add coverage for the new Context Center nav path if you want it in this
  PR rather than a follow-up.

#### Manual testing performed
Not run against a live stack — verification here is the unit/jest suites plus the checks below. The
integration test above is the gap and is called out as such.

Also verified: `tsc --noEmit` clean on every changed file; UI checkstyle (organize-imports → eslint →
prettier) reports 0 errors; `yarn i18n` stable, `license-header-fix` and `generate:app-docs` clean;
`mvn spotless:check` BUILD SUCCESS. All 8 new i18n keys were translated in all 19 non-English locales
rather than left as English placeholders.

#

⚠️ **Not attached** — I could not exercise the UI without a running stack. The change is visible at
Context Center → AI Context; a recording should be added before merge.

#
### Checklist:
- [x] For JSON Schema changes: I updated the migration scripts or explained why it is not needed —
      no migration is needed, and the reasoning is under High-level design.


pmbrull:
Addressed all three review findings in 8e8a845.

**Copilot — knowledge rules could be `filteredInSearch` 🟡 — confirmed, real, fixed.**

The gap was worse than the badge-level symptom. `createAiContextRule` stamped `filteredInSearch=true` for any rule whose flag was absent, **regardless of entity type**, so a `page` / `glossaryTerm` / `metric` rule created through the API silently stopped being materialized — the persona's articles and terms just vanished from its context document — and put a knowledge entity type into an asset-search allowlist, where it means nothing. The UI never did this (`getDefaultRule` guards it), which is exactly what made it fragile: the invariant lived only in the client.

Fixed on the server, in both directions:
- `createAiContextRule` no longer stamps knowledge entity types.
- `isFilteredInSearch` now treats a knowledge rule as preloading *whatever the stored flag says*, so a row already written with the flag set cannot resurrect the behaviour.

Covered by `knowledgeRulesAreNeverSearchScopedEvenWhenTheFlagIsSet`, which asserts both halves: the scope stays empty **and** the rule is still materialized.

**Gitar — scoped-rule badge counts disabled rules — confirmed, fixed.** `getScopedRuleCount` now requires `enabled !== false`, matching `searchScope()`. Test updated with a disabled scoped rule.

**Gitar — persona list capped at 50 — confirmed, fixed.** This list is the only nav path to a persona's AI context, so silently omitting personas makes them unreachable. Now follows the `after` cursor to exhaustion, bounded at 20 pages so a server paging bug can't spin.

Verification: backend 17/17 (`PersonaContextBuilderTest`), UI 16/16, `tsc` clean on changed files, ESLint 0 errors, spotless clean.

---

⚠️ **`Validate PR Metadata` is failing and I can't fix it** — linked issue #32457 needs adding to the **Shipping** project (issue → right sidebar → Projects), then re-run the check. That needs project write access.

Still open from the PR description, unchanged: the integration test compiles but hasn't been run (needs Docker), and there's no Playwright coverage or screen recording for the new Context Center nav path.

pmbrull:
@copilot is right, and the divergence was wider than the finding. Fixed in ea46217 + 75fec32.

**The claim checks out.** The backend predicate is:

```java
static boolean isFilteredInSearch(ContextRule rule) {
  return Boolean.TRUE.equals(rule.getFilteredInSearch())
      && !isKnowledgeEntityType(rule.getEntityType());
}
```

`getScopedRuleCount` only checked the flag, so a knowledge rule with `filteredInSearch: true` was counted as narrowing search while the server preloaded it — and the comment claiming it "mirrors `PersonaContextBuilder.searchScope`" made that worse than a silent drift.

**It's reachable, not theoretical.** Nothing rejects an explicit `{"entityType":"glossaryTerm","filteredInSearch":true}` on create or update, and rules created before the server-side guard landed were stamped `true` regardless of entity type. That is exactly why `isFilteredInSearch` is defensive on the read path.

**Three more sites had the same bug, all in `ContextRuleCard`**, which read `rule.filteredInSearch` raw:
- the mode badge — labelled a preloaded knowledge rule **"Filtered in search"** (more user-visible than the count),
- the suppression of the *Always in context* badge,
- the max-assets summary line.

Rather than patching four call sites, extracted **`isSearchScopedRule`** as the single UI mirror of the backend predicate and routed all four through it. The `enabled` check deliberately stays *out* of it and on `getScopedRuleCount` alone — a disabled rule still **is** in scoping mode, which is what the card displays, while contributing nothing to the served scope, which is what the count means.

Pinned by two tests; I verified they fail with the guard removed:

```
● does not treat a knowledge rule as search-scoped even when the flag is set
● counts only the enabled rules that are filtered in search
Tests: 2 failed, 10 passed
```

and pass with it (12/12).

One process note in case it matters to reviewers: ea46217 shipped the fix without its tests — a shell `cd` failed and short-circuited the edit, and the run I read as green was exercising the unmodified file. 75fec32 adds them, with the fail-without-fix check above as evidence rather than assertion.

karanh37 (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/PersonaResource.java
```diff
@@ -349,6 +349,13 @@ public PersonaContextDefinition createAiContextRule(
     PersonaContextDefinition definition = editableDefinition(original);
     ContextRule rule = sanitizedRule(requestedRule);
     rule.setId(UUID.randomUUID());
+    // Search scoping is the default for new rules. Only the create path stamps it: on update a null
+    // means the caller is round-tripping a rule stored before the field existed, which must keep
+    // preloading. Knowledge rules are exempt — they exist to be in context, so scoping one would
+    // drop its glossary terms, articles or metrics out of the document entirely.
+    if (rule.getFilteredInSearch() == null) {
```
tomasmontielp: Should `updateAiContextRule` handle `filteredInSearch` too? It replaces the whole rule, so a PUT without the field would drop the `true` stamped here and send the rule back to preloading.

Maybe carry the stored value over when it comes in null? A legacy rule would still be `null` and keep preloading.

pmbrull (commented):
@tomasmontielp good catch — confirmed and fixed in 2a1d44a.

`updateAiContextRule` replaces the whole rule and never carried the field forward, so a PUT that omitted `filteredInSearch` dropped the `true` the create path stamped and silently sent a scoped rule back to preloading. The symptom is invisible until the context document grows again.

The root cause is that **`null` is doing two jobs**. The field deliberately has no schema default — that is what lets a rule stored before it existed keep preloading without a data migration — so on the wire `null` means both *"I omitted this"* and *"written before the field existed"*. Create resolves that toward the product default; update was resolving it toward "legacy", which is the wrong reading for the far more common case.

Took your suggestion as written:

```java
if (rule.getFilteredInSearch() == null) {
  rule.setFilteredInSearch(rules.get(ruleIndex).getFilteredInSearch());
}
```

- omitted + stored `true` → stays scoped ✅ (the bug)
- omitted + stored `null` (legacy) → carries `null`, keeps preloading ✅ (your point, preserved)
- explicit `false` → `false`, so carrying the stored value doesn't make the field unsettable
- omitted after an explicit `false` → stays preloading

Covered by `updatingARuleWithoutTheFieldKeepsItsDeliveryMode` in `PersonaAIContextIT`, which walks all four transitions above. ⚠️ It compiles but **I have not run it** — it needs the Docker stack, same as the rest of that class.

Worth noting the OM UI was never exposed to this: `ContextRuleEditor` always sends the field. It's API clients doing read-modify-write of a subset that would have hit it — which is exactly the kind of caller a full-replacement PUT invites.

tomasmontielp (dismissed):

karanh37 (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/PersonaResource.java
```diff
@@ -346,9 +358,16 @@ public PersonaContextDefinition createAiContextRule(
...
+    PersonaContextDefinition definition = storedDefinition(original);
     ContextRule rule = sanitizedRule(requestedRule);
     rule.setId(UUID.randomUUID());
+    // Search scoping is the default for new rules. Only the create path stamps it: on update a null
+    // means the caller is round-tripping a rule stored before the field existed, which must keep
+    // preloading. Knowledge rules are exempt — they exist to be in context, so scoping one would
+    // drop its glossary terms, articles or metrics out of the document entirely.
+    if (rule.getFilteredInSearch() == null) {
+      rule.setFilteredInSearch(!PersonaContextBuilder.isKnowledgeEntityType(rule.getEntityType()));
+    }
```
manerow: I grepped the repo: `getSearchScope()` has exactly one caller, `PersonaContextMarkdown`, which just renders a paragraph. `SearchMetadataTool`, `SemanticSearchTool` and the rest of `openmetadata-mcp` are persona-unaware. The only thing that actually applies the filter is the ai-platform PR linked in the description, which lives in another repo.

So on an OSS server, right after this merges: an admin creates a rule -> we stamp `filteredInSearch = true` -> the rule preloads nothing and nothing filters search. The rule does nothing at all, silently. Same on Collate for the whole window between this landing and the ai-platform side shipping.

The stamp does not buy anything for the UI either. The editor always sends the field (`filteredInSearch: rule ? rule.filteredInSearch ?? false : !knowledgeType`), so the only clients hit by the server-side default are the ones that have never heard of the field, which are exactly the ones expecting the old preload behaviour.

"Scoped by default for new rules" is a good product default, but it belongs in the UI form default, where it already lives.

```suggestion
    // The delivery mode is whatever the client asked for. A null stays null, which the builder
    // reads as "preload", so a client that does not know about filteredInSearch keeps the old
    // behaviour. The editor sends the field explicitly, so new rules still default to scoped
    // where a consumer for the scope actually exists.
```

openmetadata-service/src/main/java/org/openmetadata/service/aicontext/PersonaContextMarkdown.java
```diff
@@ -105,6 +108,33 @@ static PersonaContextBuilder.MaterializedPersonaContext render(
...
+    if (scope == null || nullOrEmpty(scope.getEntityTypes())) {
+      return "";
+    }
+    StringBuilder markdown = new StringBuilder("\n## Search scope\n\n");
+    markdown
+        .append("Your administrator scoped this persona's asset search to ")
+        .append(String.join(", ", scope.getEntityTypes()))
+        .append(
+            " matching the rules below. Search tools apply this scope by default, so assets outside"
+                + " it will not appear in results even though they exist in the platform.\n\n");
```
manerow: This sentence goes into the persona document and states that the scope *is* enforced. On OSS nothing enforces it (see my comment on `PersonaResource`), and even on Collate the description says the agent can opt out per call, so it is advisory there too.

Concretely: the model gets a thin result set, reads this, and explains it as "those assets are out of scope" when search was never scoped. Or it tells a user an asset is hidden when it is not.

```suggestion
        .append(
            " matching the rules below. Search tools are asked to apply this scope by default, so"
                + " results may be narrowed to these assets. This is a relevance default, not a"
                + " permission boundary, and a tool may search outside it.\n\n");
```

openmetadata-service/src/main/java/org/openmetadata/service/aicontext/PersonaContextBuilder.java
```diff
@@ -277,6 +280,86 @@ RuleSearchResult search(ContextRule rule) {
...
+          .getRules()
+          .add(
+              new SearchScopeRule()
+                  .withRuleName(rule.getName())
+                  .withEntityType(rule.getEntityType())
+                  .withQueryFilter(rule.getQueryFilter()));
+    }
+    if (clauses.isEmpty()) {
+      return scope;
+    }
```
manerow, line 332: The union is `minimum_should_match: 1` over asset clauses only. Knowledge types can never be scoped (correct, and well argued in `isFilteredInSearch`), which means they also never reach `entityTypes`, which means the compiled filter matches no `glossaryTerm`, `page` or `metric` at all.

Failure case: a persona has a `glossaryTerm` preload rule plus one scoped `table` rule. That persona can no longer *search* its glossary. It only sees whatever terms happened to be preloaded. Nobody asked for that, and combined with the create-path default it happens the first time an admin adds a single table rule.

```suggestion
    if (clauses.isEmpty()) {
      return scope;
    }
    // A scoped asset rule narrows asset search. It must not also push knowledge out of the
    // allowlist: knowledge rules can never be scoped, so nothing would ever put them back.
    for (String knowledgeType : KNOWLEDGE_ENTITY_TYPES) {
      clauses.add(Map.of("term", Map.of("entityType", knowledgeType)));
      entityTypes.add(knowledgeType);
    }
```

Separate question I would like your call on, not something I want to suggest blind: should an entity type that has an *enabled but non-scoped* rule also stay searchable? Today a preloaded `dashboard` rule plus a scoped `table` rule makes dashboards unsearchable, which reads like a side effect rather than an intent.

openmetadata-service/src/main/java/org/openmetadata/service/resources/teams/PersonaResource.java
pmbrull: focusing on CAIP in this PR only

manerow (dismissed):

karanh37 (dismissed):

karanh37 (approved):

manerow (approved):

PR #32461, opened 2026-09-02
Fixes #32447: propagate async email delivery failures

### LINKED ISSUE #32447: Email notification delivery failures are incorrectly reported as successful
### Affected module

Backend

### Describe the bug

Email event subscriptions report successful delivery even when the SMTP server rejects or cannot receive the email.

The notification email is sent asynchronously, but the returned CompletableFuture is ignored. EmailPublisher therefore calls setSuccessStatus() immediately after scheduling the email, before the SMTP delivery result is known.

As a result:

- lastSuccessfulAt is updated even though no email was delivered.
- successfulEventsCount is incremented.
- failedEventsCount remains 0.
- The event is not exposed through the failed-events API.
- Subscription retry handling is not triggered.

This affects actual event notifications. The synchronous "Test Destination" path is separate and correctly reports connection failures.

### To Reproduce

1. Run OpenMetadata with SMTP enabled.
2. Configure the SMTP endpoint to an unreachable port.
3. Create an email event subscription with:
   - One email receiver
   - Batch size: 1
   - Poll interval: 1
   - Retries: 1
4. Trigger an event matched by the subscription.
5. Check the server logs.

The logs show the SMTP delivery failure, for example:

Failed to run sendMail process
MailerException: Failed to send email
Couldn't connect to host, port: <redacted-host>, <redacted-port>
java.net.ConnectException: Connection refused

6. Check the subscription status and diagnostics.

Observed results:

- Destination status remains active.
- lastSuccessfulAt is updated.
- successfulEventsCount is incremented.
- failedEventsCount is 0.
- The failed-events API returns an empty list.

### Expected behavior

EmailPublisher should wait for the asynchronous SMTP delivery result before marking the event as successful.

If SMTP delivery fails:

- The destination should record an error status.
- The event should count as failed.
- The failed event should be available through the failed-events API.
- The configured retry handling should be triggered.
- lastSuccessfulAt should not be updated for the failed delivery.

### OS

macOS, Docker Desktop

### Python version

N/A

### OpenMetadata version

Latest main (c0e42293b053f28084fc3baaf2335edf531e53ea)

N/A

### Additional context

The relevant code paths are:

- openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/changeEvent/email/EmailPublisher.java
- openmetadata-service/src/main/java/org/openmetadata/service/util/email/EmailUtil.java

EmailUtil.sendNotificationEmail() calls Simple Java Mail using async=true. Mailer.sendMail(email, true) returns a CompletableFuture<Void>, but EmailUtil currently discards that future.

EmailPublisher then immediately calls setSuccessStatus(), so failures raised later by the asynchronous mail task never reach its exception handler.

A possible fix is to return the CompletableFuture from EmailUtil and have EmailPublisher wait for all recipient delivery futures before setting the success status. This can preserve parallel email delivery while allowing SMTP failures to propagate to the subscription retry and failed-event handling.

### Pre-submission checklist

XXUCHAN:
### Describe your changes:
Fixes #32447

Email event notifications are sent asynchronously through Simple Java Mail, but the returned `CompletableFuture` was discarded. As a result, `EmailPublisher` marked events as successful before SMTP delivery completed.

This change:

- Exposes the asynchronous delivery result from `EmailUtil`.
- Waits for all recipient delivery futures before recording success.
- Propagates SMTP delivery failures as `EventPublisherException`.
- Allows the existing failed-event and retry flow to handle SMTP failures.
- Preserves the existing `void` email APIs for compatibility.

#
### Type of change:

- [x] Bug fix

#
### High-level design:

N/A — small bug fix.

#
#### Use cases covered

- Successful SMTP delivery continues to mark the subscription as active.
- An asynchronous SMTP failure marks the destination as failed.
- Failed delivery increments the failed event count instead of the successful event count.
- Failed delivery creates a failed event for the existing retry flow.

#### Unit tests

- File added:
  - `openmetadata-service/src/test/java/org/openmetadata/service/apps/bundles/changeEvent/email/EmailPublisherTest.java`
- The test verifies that a failed delivery future:
  - Raises `EventPublisherException`.
  - Preserves the destination ID and `ChangeEvent`.
  - Sets the destination status to `FAILED`.
  - Does not update `lastSuccessfulAt`.

Test command:

```bash
mvn -pl openmetadata-service -am \
  -Dtest=EmailPublisherTest,AbstractEventConsumerTest \
  -Dsurefire.failIfNoSpecifiedTests=false package
```

Result: 31 tests passed with no failures or errors.

- [x] Not applicable — no backend API changes.

- [x] Not applicable — no ingestion changes.

#### Playwright (UI) tests

- [x] Not applicable — no UI changes.

#### Manual testing performed

1. Started OpenMetadata with SMTP enabled and Mailpit running.
2. Confirmed that Test Destination returned `Success/200`.
3. Created an email event subscription for Domain events.
4. Stopped Mailpit without changing the SMTP configuration.
5. Triggered a matching Domain event.

Before the fix:

- SMTP logs contained `Connection refused` and `MailerException`.
- The destination remained `active`.
- `successfulEventsCount` increased to 1.
- `failedEventsCount` remained 0.
- No failed event was created.

After the fix:

- The destination changed to `failed`.
- `lastFailedStatusCode` was 500.
- `failedEventsCount` increased to 1.
- A failed event was created with the destination ID and `ChangeEvent`.
- Restarting Mailpit allowed successful delivery to continue normally.

The temporary subscription and Domain entities were removed after testing.

Formatting verification:

```bash
mvn -pl openmetadata-service spotless:apply
mvn -pl openmetadata-service spotless:check
```

Both commands completed successfully.

#

#
### Checklist:

- [x] I have read the [**[CONTRIBUTING](https://docs.open-metadata.org/developers/contribute)**](https://docs.open-metadata.org/developers/contribute) document.
- [x] My PR title is `Fixes #32447: propagate asynchronous email delivery failures`.
- [x] My PR is linked to a GitHub issue via `Fixes #32447` above.
- [x] JSON Schema changes are not applicable.
- [x] UI screenshots are not applicable.

#### Bug fix

```

### CODE REVIEW

File: openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/changeEvent/email/EmailPublisher.java [hunk 1]
```diff
@@ -86,11 +107,14 @@ public void sendMessage(ChangeEvent event, Set<Recipient> recipients)
               .filter(Objects::nonNull)
               .collect(Collectors.toSet());
 
-      // Send email to each recipient
-      for (String receiver : receivers) {
-        EmailUtil.sendNotificationEmail(
-            receiver, emailMessage.getSubject(), emailMessage.getHtmlContent());
-      }
+      CompletableFuture.allOf(
```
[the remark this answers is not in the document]
XXUCHAN: The existing retry model is scoped to a destination and ChangeEvent, and failed-event records do not preserve per-recipient delivery state.

Retrying only the failed addresses would require changes to the retry persistence and recipient resolution model, which is outside the scope of this fix. Previously, an asynchronous failure was silently recorded as success, so failed recipients were never retried.

This change restores the existing destination-level failure and retry flow. That flow may resend to recipients that succeeded during a partial failure. Per-recipient retry or idempotency can be tracked separately if needed.

openmetadata-service/src/main/java/org/openmetadata/service/apps/bundles/changeEvent/email/EmailPublisher.java
```diff
@@ -58,6 +62,24 @@ public EmailPublisher(
           new HandlebarsNotificationMessageEngine(
               (NotificationTemplateRepository)
                   Entity.getEntityRepository(Entity.NOTIFICATION_TEMPLATE));
+      this.emailSender = new DefaultEmailSender();
+    } else {
+      throw new IllegalArgumentException("Email Alert Invoked with Illegal Type and Settings.");
+    }
+  }
+
+  EmailPublisher(
```
TeddyCr: Injecting the sender is the right direction, and it's what `.claude/rules/java.md` asks for: constructor-inject external clients, don't look up static singletons inside business methods. Replacing the in-line `EmailUtil.getSmtpSettings()` call in `sendMessage` with an injected client is a real improvement, and `DefaultEmailSender` is a legitimate composition boundary supplying the production singleton.

What I don't want to merge is the shape. `EmailSender` (L178) is a package-private interface nested inside `EmailPublisher`, with a nested private implementation (L184), and the only caller of this constructor is `EmailPublisherTest`. As written it reads as a test hook rather than an abstraction: nothing tells the next reader whether the SMTP transport is meant to be swappable at runtime, no other class can reuse it, and the test is permanently pinned to this package because it references `EmailPublisher.EmailSender`.

Pick one and commit to it:

- **Make it a real boundary.** Promote `EmailSender` to a top-level interface in `org.openmetadata.service.util.email`, next to `EmailUtil`, with `DefaultEmailSender` as a top-level implementation. Then it's an SMTP-transport abstraction the other publishers that send mail can also use, this constructor is ordinary DI rather than a test-only overload, and the test no longer depends on package-private access to compile.
- **Or drop it** and have `EmailPublisher` keep calling `EmailUtil` directly.

I'd take the first. Either way, I'd rather not land a production interface whose only reason to exist is that a test needs something to mock.

Widening the `messageEngine` field to `NotificationMessageEngine` and the `getFailureReason` unwrapping of `CompletionException` both look right to me — the unwrapping is the actual fix here, and it's well targeted. This is only about the seam around it.

XXUCHAN: Thanks for clarifying. I took the first option and promoted EmailSender and DefaultEmailSender to top-level public types under org.openmetadata.service.util.email. The injection constructor is now ordinary public DI, and the test depends on the top-level interface rather than a package-private nested type. The existing AlertFactory production path still uses the two-argument constructor and DefaultEmailSender. Spotless and all 31 related tests pass.

Issue #32459, opened 2026-09-02
Memory visibility check missing on some context memory endpoints


Vishnuujain:
Memory visibility comes from each memory's `shareConfig`, checked by `ContextMemoryVisibility`. It runs on `list`, `get/{id}` and `get/name/{fqn}`. These endpoints skip it:

- `GET /{id}/versions`
- `GET /{id}/versions/{version}`
- `PATCH /{id}`
- `DELETE /name/{fqn}`

Also, `get` and `getByName` check with whatever `fields` the caller asked for. Without `owners` in that list the memory has no owners to match, so its own owner is refused.

Fix:

- Fetch with `ContextMemoryVisibility.guardFields(entityType, fields)`, then check with `enforceVisibility(entity, securityContext)`. Both helpers come from the MCP change.
- Add `ContextMemoryIT` cases per endpoint: owner, admin, other user, and `Entity` / `Shared` visibility.

PR #32465, opened 2026-09-02
Apply context memory visibility rules on MCP entity reads

Vishnuujain:
### Background

- Every context memory carries its own sharing setting: private, shared with certain people, or open to everyone.
- That setting is not a role or a policy, so the usual permission check does not look at it. The REST API checks it separately on each read.
- The MCP tools that read an entity by name did not run that check.

### What changed

- These MCP tools now run the same check the REST API runs:
  - `get_entity_details` — plain read, `include=[content]`, and the multi-entity form
  - `company_context` — reading one pill by name
  - `patch_entity` — its reply includes the entity, so the check runs before the write
- The check now sits next to the fetch, in `CommonUtils.readEntityForCaller(...)`. One call fetches and checks, so a read path cannot skip it.
- The fetch now asks for `owners`. Without it the memory looks unowned, and its own owner would be turned away.
- `company_context` by name answers for the person asking, instead of returning any shared pill. It also checks permissions against that pill, not just the entity type.
- If the caller cannot be identified, they now see only memories open to everyone.

### New helpers in `ContextMemoryVisibility`

- `enforceVisibility(entity, securityContext)` — applies the rule for a memory, does nothing for other entity types, so callers need no type knowledge.
- `hasVisibilityRules(entityType)` — whether a type has a sharing setting at all.
- `guardFields(entityType, fields)` — adds the fields the check needs to a fetch.

### Tests

- `openmetadata-mcp`: 594/594. Context memory and AI context suites in `openmetadata-service`: 136/136.
- One unit test per path, each confirmed failing without this change.
- New `McpContextMemoryVisibilityIT` covers the owner, an admin, another user, and someone the memory is shared with. Not run locally (needs a running server), so CI is its first run.

### Notes

- No schema, migration, or UI change.
- Remaining REST paths tracked in #32459.


pmbrull (dismissed):

pmbrull (approved):

pmbrull:
picked to 2.0 here https://github.com/open-metadata/OpenMetadata/pull/32973

PR #32481, opened 2026-09-02
fix(ingestion): apply Snowflake query tags at connection creation

### LINKED ISSUE #32332: Snowflake query tags not consistently applied to all agent-generated queries
### Affected module

Connector

### Describe the bug

We have identified cases where queries are still executed without the expected query tag. They have not yet identified a specific pattern or condition that isolates when the query tag is missing.

The user has provided an aggregate of queries from the current week, grouped by whether a query tag was assigned.

One of the most frequently occurring queries without a query tag is:
```
select
TABLE_NAME,
NULL as DELETED,
CASE
WHEN IS_TRANSIENT = 'YES' THEN 'TRANSIENT TABLE'
WHEN IS_DYNAMIC = 'YES' THEN 'DYNAMIC TABLE'
WHEN IS_ICEBERG = 'YES' THEN 'ICEBERG TABLE'
ELSE TABLE_TYPE
END as TABLE_TYPE
from information_schema.tables
where TABLE_SCHEMA = 'fivetran_metadata'
AND COALESCE(IS_TRANSIENT, 'NO') != 'YES'
AND TABLE_TYPE != 'VIEW'
```

### To Reproduce

Query tags are applied correctly for most agent-generated Snowflake queries, but some queries are executed without the configured query tag.

### Expected behavior

Since the Snowflake connection and query tag configuration are defined in a single place in Collate, the configured query tags should be consistently applied to all Snowflake connections and queries generated by Collate agents.

### OS

_No response_

### Python version

_No response_

### OpenMetadata version

1.13.4

_No response_

### Additional context

_No response_

### Pre-submission checklist

IceS2:
### Describe your changes:

Fixes #32332 

Snowflake query tags were previously applied by workflow-specific `ALTER SESSION` hooks after engine creation. Some engines and physical connections therefore escaped tagging.

This change passes `QUERY_TAG` as a Snowflake session parameter before engine creation, preserves user session parameters without mutating connection configuration, and removes the redundant `ALTER SESSION` paths.

### Type of change:

- [x] Bug fix

Validated with 439 Snowflake unit tests, Python formatting and type checks, and local Snowflake runs across metadata, profiler, sampler, usage, and lineage.


ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
```diff
@@ -481,30 +480,31 @@ def get_connection(self) -> Engine:
...
+        if connection.queryTag:
+            session_parameters["QUERY_TAG"] = connection.queryTag
+        if session_parameters or "session_parameters" in connect_args:
+            connect_args["session_parameters"] = session_parameters
+
+        return create_generic_db_connection(
             connection=connection,
             get_connection_url_fn=self.get_connection_url,
-            get_connection_args_fn=get_connection_args_common,
+            get_connection_args_fn=lambda _: connect_args,
```
TeddyCr, line 506:
```suggestion
            get_connection_args_fn=lambda: connect_args,
```

TeddyCr (commented):
Can you check if the lambda suggestion makes sense, we don't use any of its args,

ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
IceS2: create_generic_db_connection invokes this callback with the connection argument, so the lambda must accept one parameter. _ indicates that it is intentionally unused because the arguments were prepared earlier. A zero-argument lambda would fail at runtime, so I would keep the current signature.

ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
```diff
@@ -481,30 +480,31 @@ def get_connection(self) -> Engine:
...
         # error within 10 minutes instead of hanging the worker indefinitely.
         # User-supplied connectionArguments win via setdefault.
-        if connection.connectionArguments.root is not None:
-            connection.connectionArguments.root.setdefault("network_timeout", 600)
+        connect_args.setdefault("network_timeout", 600)
 
-        engine = create_generic_db_connection(
+        session_parameters = dict(connect_args.get("session_parameters") or {})
+        if connection.queryTag:
+            session_parameters["QUERY_TAG"] = connection.queryTag
```
Khairajani, line 499: `network_timeout` above uses `setdefault` so the user wins, here `queryTag` overwrites theirs. Worth picking one.

ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
```diff
@@ -481,30 +480,31 @@ def get_connection(self) -> Engine:
...
+        if session_parameters or "session_parameters" in connect_args:
```
Khairajani: Second half is a no-op, `if session_parameters:` is enough.

ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
Khairajani, line 506: This dict is shared by every connection in the pool and the connector mutates it in place at login. Only adds keys so probably fine, but worth a comment.

ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
IceS2: QueryTag is a first class config while network_timeout is just our default value. That's the reason

ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
IceS2: Fixed

ingestion/src/metadata/ingestion/source/database/snowflake/connection.py
IceS2: Fair, won't change now since it would add more lifecycle to maintain and should be fine like it was already

ulixius9 (approved):

Issue #30144, opened 2026-07-16
Bulk glossary CSV import rejects glossary terms with the same name under different parent values


hobie88:
### Affected module

Governance (Glossary / Classification / Domains)

### Describe the bug

When bulk importing a glossary via CSV, two glossary terms that share the same name but have different parent values are rejected by the import validator, even though fullyQualifiedName uniqueness (glossary.parent.term) would make both terms valid, distinct entities.

### To Reproduce

1. Create (or use an existing) glossary, e.g. Banking.
2. Create two parent terms: Account and Loan.
3. Prepare a CSV with two child terms that share the same name but different parent:
```
parent,name,displayName,description,synonyms,relatedTerms,references,tags
Account,Balance,Account Balance,"Account balance definition",,,,
Loan,Balance,Loan Balance,"Loan balance definition",,,,
```

### Expected behavior

Both 'Banking.Account.Balance' and 'Banking.Loan.Balance' should be created successfully, since they have distinct fullyQualifiedNames and the data model (per the glossaryTerm schema) enforces uniqueness on FQN, not on bare name.

### OpenMetadata version

1.12.1

### Pre-submission checklist

- [x] I searched for duplicate issues.
- [x] I removed credentials, hostnames, emails, and other sensitive data from logs and config.

OdyAsh:
Hello @hobie88  @zerafachris ! I'd like to mention that this behavior is not only present via the CSV import feature, but as well via the OM UI.

Excerpt from [this](https://openmetadata.slack.com/archives/C02B6955S4S/p1787048002215969?thread_ts=1783593367.064549&cid=C02B6955S4S) slack thread:

When creating glossary terms "t1" and "t2", then creating a term inside t1 called "t1-common-t2", then try to create a term inside t2 called "t1-common-t2", it doesn't work (screenshot 1 and 2 below). However, I expect that scenario to work as well, since the FQNs of 
* `p1.t1.t1-common-t2` and 
* `p1.t2.t1-common-t2` 

are different.

This issue has been prevalent in all OM versions (even the latest one)

Therefore, kindly, I'd like confirmation that the PR mentioned [here](https://github.com/open-metadata/OpenMetadata/pull/30209) will fix the underlying issue, so that both CSV/UI options are working fine 🙌 <3.

<img width="612" height="531" alt="Image" src="https://github.com/user-attachments/assets/1386b215-dd6a-43e7-9f14-75f709aecbc5" />

[...]

PR #32486, opened 2026-09-02
fix(gcs): use double_value for total_bytes metric

### LINKED ISSUE #32422: GCS connector: `total_bytes` container metric always reports 0 (DOUBLE/INT64 value type mismatch in _fetch_metric)
### Affected module

Connector

### Describe the bug

The GCS storage connector's `total_bytes` container metric is always `0`, even for buckets with confirmed non-zero data in Cloud Monitoring.

Root cause: `_fetch_metric()` in `metadata/ingestion/source/storage/gcs/metadata.py` reads the value from the Cloud Monitoring `Point` the same way for both metrics it queries:

```python
point = list(timeseries)[-1].points[-1]
return point.value.int64_value
```

`storage.googleapis.com/storage/object_count` is `valueType: INT64`, so this works.
`storage.googleapis.com/storage/total_bytes` is `valueType: DOUBLE` (confirmed via the `metricDescriptors` API, see below), so `point.value.int64_value` reads the wrong oneof field. In protobuf this does **not** raise — it silently returns the type's zero-default. The `except Exception` / "Failed fetching metric ... returning 0" branch never fires, because no exception happens. The result: `total_bytes` is always `0` for every bucket, silently, with no warning or error anywhere in the logs.

`object_count` is unaffected and reports correctly.

### To Reproduce

1. Query the metric descriptors directly (no ingestion needed to see the root cause):
   ```
   GET https://monitoring.googleapis.com/v3/projects/{project}/metricDescriptors/storage.googleapis.com/storage/total_bytes
   → "valueType": "DOUBLE"

   GET https://monitoring.googleapis.com/v3/projects/{project}/metricDescriptors/storage.googleapis.com/storage/object_count
   → "valueType": "INT64"
   ```
2. Confirm (e.g. via Metrics Explorer, filtering on a specific `bucket_name`) that `total_bytes` actually has recent non-zero data points for a real bucket.
3. Run `metadata ingest` against a GCS service connection that includes that bucket.
4. Open the resulting Container entity in OpenMetadata — `numberOfObjects` is correct, `size` is `0 Bytes`.

```yaml
source:
  type: gcs
  serviceName: your-service-name
  serviceConnection:
    config:
      type: GCS
      credentials:
        gcpConfig:
          type: gcp_credential_path
          path: /path/to/key.json
          projectId: your-project-id
  sourceConfig:
    config:
      type: StorageMetadata

sink:
  type: metadata-rest
  config: {}

workflowConfig:
  loggerLevel: INFO
  openMetadataServerConfig:
    hostPort: "https://your-openmetadata-host/api"
    authProvider: openmetadata
    securityConfig:
      jwtToken: "<redacted>"
    storeServiceConnection: true
```

### Logs
No error or warning is logged for the affected buckets — that's the point of the bug. For comparison, this is what a *genuine* metric-fetch failure looks like (this only happened before the service account had `roles/monitoring.viewer`; once that permission was granted, these warnings disappeared but `total_bytes` stayed `0` anyway):

```
[INFO] Running GetMetrics...
[INFO] Test connection results: ... TestConnectionStepResult(name='GetMetrics', mandatory=False, passed=True, ...)
[INFO] Looking for metadata template file at - gs://your-bucket/openmetadata.json
[WARNING] No metadata file found at gs://your-bucket/openmetadata.json
```
— i.e. after the permission fix, `GetMetrics` test step passes, `object_count` populates correctly, and there is nothing in the logs at all pointing at `total_bytes` being wrong. It's only visible by comparing the entity's `size` field against the real Cloud Monitoring value.

### Expected behavior

`size` on the Container entity should reflect the real `total_bytes` value from Cloud Monitoring (in our repro, a multi-TiB bucket), not `0`.

### OS

macOS 26 (Darwin 25.5.0)

### Python version

3.12

### OpenMetadata version

1.13.3

openmetadata-ingestion==1.13.4

### Additional context

Suggested fix direction: `_fetch_metric` should pick the value field based on the metric's actual value type instead of hardcoding `int64_value` — e.g. via `point.value.WhichOneof("value")`, or by reading the value type once from the `MetricDescriptor` and branching accordingly. Since `object_count` (INT64) and `total_bytes` (DOUBLE) are fetched through the same helper, the fix should keep both paths working rather than hardcoding either field.

### Pre-submission checklist

zerafachris:
## Why

Cloud Monitoring reports `storage.googleapis.com/storage/total_bytes` as a **DOUBLE**-typed time series, but the ingestion code was unconditionally reading `point.value.int64_value`. Because protobuf `oneof` silently returns the zero-default when the wrong field accessor is used, `size` was always reported as `0` for every GCS bucket.

`object_count` is correctly INT64 and is unaffected.

Fixes #32422

## Changes

**`ingestion/src/metadata/ingestion/source/storage/gcs/metadata.py`**
- In `_fetch_metric()`: branch on `GCSMetric.BUCKET_SIZE_BYTES` to read `point.value.double_value`; all other metrics (currently `NUMBER_OF_OBJECTS`) continue to use `point.value.int64_value`.

**`ingestion/tests/unit/topology/storage/test_gcs_storage.py`**
- Added `TestFetchMetric` with two unit tests:
  - `test_bucket_size_bytes_uses_double_value` — mocks Cloud Monitoring and asserts `double_value` is returned for `BUCKET_SIZE_BYTES`
  - `test_number_of_objects_uses_int64_value` — asserts `int64_value` is returned for `NUMBER_OF_OBJECTS`

## How to test

1. Stand up a GCS ingestion pipeline against a real bucket with non-zero size.
2. After ingestion, verify the container entity's `size` field is non-zero.

Or run the new unit tests:
```
cd ingestion && pytest tests/unit/topology/storage/test_gcs_storage.py::TestFetchMetric -v
```

<h2><a href="https://app.greptile.com/api/retrigger?id=59516083"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 3/5</h2>

The GCS correction appears sound, but the glossary changes are not safe to merge until CSV parent moves and parent-only duplicate collisions are handled.

<h2>Findings</h2>

1. P1&nbsp;**CSV moves create duplicates** <a href="https://github.com/open-metadata/OpenMetadata/pull/32486#discussion_r4063006202">▶</a>
2. P1&nbsp;**Parent moves bypass uniqueness** <a href="https://github.com/open-metadata/OpenMetadata/pull/32486#discussion_r4063006210">▶</a>

This PR corrects GCS bucket-size extraction by reading Cloud Monitoring's DOUBLE value and also changes glossary-term uniqueness and CSV matching to operate within a direct parent.

- Reads `total_bytes` from `double_value` while retaining `int64_value` for object counts.
- Adds focused unit coverage for both GCS metric value types.
- Allows equal glossary leaf names beneath different parents.
- Adds parent-scoped glossary DAO queries and create-path integration coverage.
- The glossary changes regress CSV parent moves and omit collision validation for parent-only moves.
</details>

```mermaid
flowchart LR
  CSV[CSV term row] --> FQN[Build target-parent FQN]
  FQN --> Exact{Existing target FQN?}
  Exact -- Yes --> Update[Update matched term]
  Exact -- No --> Parent[Search target parent and name]
  Parent -- Match --> Update
  Parent -- No match --> Create[Create new term]
  Old[Existing term beneath old parent] -. cannot match target-parent search .-> Parent
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(glossary): address reviewer feedback..."](https://github.com/open-metadata/openmetadata/commit/1ec5fe12f679b7d9c0c92baf3a6a1ad3cedf16e6)</sub>

### DIFF
--- a/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/GlossaryResourceIT.java
+++ b/openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/GlossaryResourceIT.java
@@ -1845,4 +1845,78 @@ public class GlossaryResourceIT extends BaseEntityIT<Glossary, CreateGlossary> {
           "Glossary reviewer count should match");
     }
   }
+
+  @Test
+  void test_createGlossaryTerm_sameNameUnderDifferentParents_succeeds(TestNamespace ns) {
+    // Regression test: two glossary terms sharing the same leaf name but living under different
+    // parents have distinct FQNs and must both be creatable. Previously the duplicate-name check
+    // on create matched any term with the same name anywhere in the glossary, regardless of
+    // parent, causing the second term ("Loan.Balance") to be rejected as a duplicate of the
+    // already-persisted "Account.Balance".
+    OpenMetadataClient client = SdkClients.adminClient();
+
+    CreateGlossary createGlossary = createMinimalRequest(ns);
+    Glossary glossary = createEntity(createGlossary);
+    String glossaryFqn = glossary.getFullyQualifiedName();
+
+    GlossaryTerm account =
+        client
+            .glossaryTerms()
+            .create(
+                new CreateGlossaryTerm()
+                    .withName(ns.prefix("account"))
+                    .withGlossary(glossaryFqn)
+                    .withDescription("Account parent term"));
+    GlossaryTerm loan =
+        client
+            .glossaryTerms()
+            .create(
+                new CreateGlossaryTerm()
+                    .withName(ns.prefix("loan"))
+                    .withGlossary(glossaryFqn)
+                    .withDescription("Loan parent term"));
+
+    GlossaryTerm accountBalance =
+        client
+            .glossaryTerms()
+            .create(
+                new CreateGlossaryTerm()
+                    .withName("Balance")
+                    .withGlossary(glossaryFqn)
+                    .withParent(account.getFullyQualifiedName())
+                    .withDescription("Account balance definition"));
+    assertNotNull(accountBalance, "Account.Balance term should have been created");
+
+    GlossaryTerm loanBalance;
+    try {
+      loanBalance =
+          client
+              .glossaryTerms()
+              .create(
+                  new CreateGlossaryTerm()
+                      .withName("Balance")
+                      .withGlossary(glossaryFqn)
+                      .withParent(loan.getFullyQualifiedName())
+                      .withDescription("Loan balance definition"));
+    } catch (Exception e) {
+      fail(
+          "Creating 'Balance' under a different parent ('"
+              + loan.getFullyQualifiedName()
+              + "') should not be rejected as a duplicate of the existing '"
+              + accountBalance.getFullyQualifiedName()
+              + "': "
+              + e.getMessage());
+      return;
+    }
+
+    assertNotNull(loanBalance, "Loan.Balance term should have been created");
+    assertEquals("Balance", accountBalance.getName());
+    assertEquals("Balance", loanBalance.getName());
+    assertEquals(
+        account.getFullyQualifiedName() + ".Balance",
+        accountBalance.getFullyQualifiedName());
+    assertEquals(
+        loan.getFullyQualifiedName() + ".Balance",
+        loanBalance.getFullyQualifiedName());
+  }
 }
--- a/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
+++ b/openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
@@ -2728,11 +2728,17 @@ public class GlossaryTermRepository extends EntityRepository<GlossaryTerm> {
   }
 
   private void checkDuplicateTerms(GlossaryTerm entity) {
+    // Term names only need to be unique among siblings sharing the same parent (or, for
+    // top-level terms, among the glossary's direct children) since the FQN already
+    // disambiguates terms with the same name under different parents.
+    String parentFqn =
+        entity.getParent() != null
+            ? entity.getParent().getFullyQualifiedName()
+            : entity.getGlossary().getFullyQualifiedName();
     int count =
         daoCollection
             .glossaryTermDAO()
-            .getGlossaryTermCountIgnoreCase(
-                entity.getGlossary().getFullyQualifiedName(), entity.getName());
+            .getGlossaryTermCountIgnoreCaseUnderParent(parentFqn, parentFqn, entity.getName());
     if (count > 0) {
       throw new IllegalArgumentException(
           CatalogExceptionMessage.duplicateGlossaryTerm(
--- a/ingestion/src/metadata/ingestion/source/storage/gcs/metadata.py
+++ b/ingestion/src/metadata/ingestion/source/storage/gcs/metadata.py
@@ -498,6 +498,10 @@ class GcsSource(StorageServiceSource):
                 name=f"projects/{bucket.project_id}", filter=filter_, interval=interval
             )
             point = list(timeseries)[-1].points[-1]
+            # total_bytes is DOUBLE in Cloud Monitoring; object_count is INT64.
+            # protobuf silently returns 0 when the wrong oneof field is read.
+            if metric == GCSMetric.BUCKET_SIZE_BYTES:
+                return point.value.double_value  # noqa: TRY300
             return point.value.int64_value  # noqa: TRY300
         except Exception:
             logger.debug(traceback.format_exc())
--- a/ingestion/tests/unit/topology/storage/test_gcs_storage.py
+++ b/ingestion/tests/unit/topology/storage/test_gcs_storage.py
@@ -489,3 +489,51 @@ class StorageUnitTest(TestCase):
     def return_metadata_entry(self):
         container_config = StorageContainerConfig.model_validate(MOCK_METADATA_FILE_RESPONSE)
         return container_config.entries
+
+
+class TestFetchMetric(TestCase):
+    """Unit tests for GcsSource._fetch_metric value-type dispatch."""
+
+    def setUp(self):
+        from metadata.ingestion.source.storage.gcs.metadata import GCSMetric
+
+        self.GCSMetric = GCSMetric
+
+        self.source = Mock(spec=GcsSource)
+        self.source.gcs_clients = Mock()
+
+        # Wire _fetch_metric as the real implementation bound to self.source
+        self.source._fetch_metric = GcsSource._fetch_metric.__get__(
+            self.source, GcsSource
+        )
+        self.source._get_time_interval = Mock(return_value=Mock())
+
+    def _make_timeseries(self, double_value=0.0, int64_value=0):
+        point_value = Mock()
+        point_value.double_value = double_value
+        point_value.int64_value = int64_value
+        point = Mock()
+        point.value = point_value
+        ts = Mock()
+        ts.points = [point]
+        return [ts]
+
+    def test_bucket_size_bytes_uses_double_value(self):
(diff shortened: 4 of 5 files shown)


openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -2728,11 +2728,17 @@ private void fetchAndSetChildrenCount(List<GlossaryTerm> entities, Fields fields
...
+    String parentFqn =
+        entity.getParent() != null
+            ? entity.getParent().getFullyQualifiedName()
+            : entity.getGlossary().getFullyQualifiedName();
     int count =
         daoCollection
             .glossaryTermDAO()
-            .getGlossaryTermCountIgnoreCase(
-                entity.getGlossary().getFullyQualifiedName(), entity.getName());
+            .getGlossaryTermCountIgnoreCaseUnderParent(parentFqn, parentFqn, entity.getName());
```
manerow, line 2741: Relaxing the create check makes "same name under different parents" a supported shape, but `findMatchForImport` (line 3551) and `createOrUpdateForImport` (line 3521) still fall back to looking a term up by `(glossary, bare name)` through `getGlossaryTermByNameAndGlossaryIgnoreCase`. That fallback was written for the case "the term moved to a new parent so its FQN changed", and it only works while the bare name identifies exactly one term. Once it does not, CSV import resolves to the wrong term.

I reproduced this on a current build with no patch applied, so the path exists today and this change makes it routine. Starting from a glossary containing `Account.Balance` and its child `Account.Balance.Detail`, importing a single row:

```
<glossary>.Loan,Balance,,Loan balance,...
```

returns:

```
status: success    details: Entity updated
```

and afterwards `Account.Balance` is a 404 while `Loan.Balance` exists as the same entity at version 0.2. With two `Balance` terms already present, importing a third one under a `Savings` parent picked one of them arbitrarily and moved it along with its child subtree, so `Account.Balance.Detail` became `Savings.Balance.Detail`. The pick is arbitrary because the query has no `ORDER BY` and JDBI's single-value return is `findFirst().orElse(null)`. The user asked for a new term and lost an existing one, with no error reported.

This matters here in particular because #30144 is a CSV import report, while the test added in this PR covers the REST path.

Would it make sense to scope that name fallback by parent as well, or key it on the `id` column so it does not have to guess, before the create rule is loosened? As it stands the feature is reachable from the UI but corrupts data from CSV.

openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/GlossaryTermRepository.java
```diff
@@ -2728,11 +2728,17 @@ private void fetchAndSetChildrenCount(List<GlossaryTerm> entities, Fields fields
...
     if (count > 0) {
       throw new IllegalArgumentException(
           CatalogExceptionMessage.duplicateGlossaryTerm(
```
manerow, line 2744: A few lines below this, `checkDuplicateTermsForUpdate` still calls `getGlossaryTermCountIgnoreCaseExcludingId`, which is glossary-wide, and it is the only name check the rename path runs (see the `nameChanged` branch around line 3030). So after this change the two paths disagree about the same end state:

- Creating `Balance` under `Loan` while `Account.Balance` exists: allowed.
- Renaming `Loan.LoanBal` to `Balance`: still rejected.

Confirmed against a running server. This code is untouched by the PR, so it behaves the same afterwards:

```
PATCH /glossaryTerms/{id of Loan.Temp}   replace /name -> "Interest"
400  A term with the name 'Interest' already exists in '<glossary>' glossary.
```

while `Interest` existed only under `Account`.

Switching that method to the same parent-scoped query with the exclude-id clause would keep the two paths consistent, so a term can be renamed into any name its own siblings are not already using.

openmetadata-integration-tests/src/test/java/org/openmetadata/it/tests/GlossaryResourceIT.java
```diff
@@ -1845,4 +1845,78 @@ private void validateGlossaryFieldsAfterImport(Glossary original, Glossary impor
...
+              + accountBalance.getFullyQualifiedName()
+              + "': "
+              + e.getMessage());
+      return;
+    }
+
+    assertNotNull(loanBalance, "Loan.Balance term should have been created");
+    assertEquals("Balance", accountBalance.getName());
+    assertEquals("Balance", loanBalance.getName());
+    assertEquals(
```
manerow, line 1915: These two assertions need `mvn spotless:apply`. Running the repo's spotless config over this file rewrites them onto single continuation lines:

```java
assertEquals(
    account.getFullyQualifiedName() + ".Balance", accountBalance.getFullyQualifiedName());
assertEquals(loan.getFullyQualifiedName() + ".Balance", loanBalance.getFullyQualifiedName());
```

The `java-checkstyle` job has not flagged this yet because it currently exits early on the missing `safe to test` label, but it runs `mvn spotless:apply` and fails on any resulting diff, so it will fail once that label is added.

manerow (changes requested):
Reviewed the Java changes only, leaving the ingestion part to the ingestion team.

The SQL in the new DAO method is correct. I checked it against a real MySQL instance: scoped to a parent term it counts only that parent's direct children, the old query counted the whole subtree, and a grandchild carrying the same name is correctly excluded. The mechanism does what the comment says it does.

Three things I think are worth resolving before this lands, in the inline comments. The most important is that the CSV import path still resolves terms by bare name, which I was able to reproduce as silent data loss on a current build.

A packaging note as well: the branch carries a second commit (e15ce61) that changes glossary duplicate-name enforcement, and it is not mentioned in the title or the description, which are both about the GCS `total_bytes` fix. Those are independent changes with different reviewers and different risk, so splitting them would make both easier to land, and the glossary change deserves a description of its own since it changes a uniqueness rule.

One last thing worth checking before iterating on the implementation: the exact CSV from #30144 already imports successfully on main. Import processes rows in batches of 100, so `checkDuplicateTerms` queries the database before the earlier row has been flushed and never sees the sibling. I created a fresh glossary and imported the issue's four rows unmodified, and all four were created, including both `Balance` terms. It would help to know what the original reporter actually reproduced, since the REST create path may not be where the reported problem lives.

zerafachris:
Thanks @manerow for the thorough review — all three concerns are addressed in the follow-up commit (1ec5fe1).

**1. Spotless formatting**
Both `assertEquals` calls in `GlossaryResourceIT` are now on single continuation lines as spotless requires.

**2. `checkDuplicateTermsForUpdate` inconsistency**
`checkDuplicateTermsForUpdate` now uses a new DAO method `getGlossaryTermCountIgnoreCaseExcludingIdUnderParent`, which applies the same parent-scoped sibling check as the create path. Renaming a term under `Account` to a name already used by a term under `Loan` is now allowed; renaming to a name already taken by a direct sibling is still blocked.

**3. CSV import data corruption**
Both `findMatchForImport` and `createOrUpdateForImport` now use a new DAO method `getGlossaryTermByNameAndParentIgnoreCase` for the bare-name fallback. This scopes the lookup to direct children of the target parent from the CSV row, so a homonym under a different parent is never picked up and moved.

The reviewer's repro scenario (`Loan,Balance` importing into a glossary that already has `Account.Balance`) would now create a new `Loan.Balance` rather than silently moving `Account.Balance`. The pre-existing query that had no `ORDER BY` and could pick arbitrarily has been replaced by the parent-scoped lookup.

**Splitting the PR**
I agree that the glossary fix and the GCS metric fix are independent changes. If the maintainers prefer them in separate PRs I'm happy to open a standalone PR for the glossary change — let me know.

PR #32504, opened 2026-09-03
Fixes #32062 & Fixes #32039 : Fix unable to render logs on Kubernetes 1.36+

### LINKED ISSUE #32039: K8sPipelineClient: ingestion logs unavailable on Kubernetes newer than        client-java models (V1PodStatus unknown-field deserialization)
### Affected module

Backend

### Affected
OpenMetadata 2.0.0 (stock `docker.getcollate.io/openmetadata/server:2.0.0`),
`pipelineServiceClientConfig.type: k8s`, EKS v1.36.2.

### Description
#29791 (fixing #29572) added `IllegalArgumentException` handling to three
`K8sPipelineClient` pod/job read paths so a cluster newer than the bundled
`io.kubernetes:client-java` models degrades gracefully instead of failing the
status endpoint. That works — `/services/ingestionPipelines/status` reports
healthy and the Agents tab renders.

However `getLastIngestionLogs` has no meaningful degraded result, so log
retrieval is now permanently unavailable on such clusters. The UI shows:

    Failed to retrieve logs: Kubernetes pod status is incompatible with the
    bundled client

Server log:

    WARN o.o.s.c.p.k.K8sPipelineClient - Could not deserialize pod while
      fetching logs for pipeline <uuid> (client-java models likely lag the
      cluster's Kubernetes version)
    java.lang.IllegalArgumentException: The field `allocatedResources` in the
      JSON string is not defined in the `V1PodStatus` properties
      at io.kubernetes.client.openapi.JSON.deserialize(JSON.java:917)
      at K8sPipelineClient.getLastIngestionLogs(K8sPipelineClient.java:1074)

K8s 1.36 pod status carries at least `allocatedResources`, `observedGeneration`,
`recursiveReadOnly` and `user` (in `containerStatuses`), none defined in the
client's models.

### Why a client-java bump isn't the fix
client-java 26.0.0 only reaches ~K8s 1.34, and the generated `JSON.java`
rejects unknown fields, so the same break returns with K8s 1.37. No public
lenient-parse setter is exposed.

### Suggested fix
`getLastIngestionLogs` only needs a pod *name* before calling
`readNamespacedPodLog` (which returns plain text and parses fine). Resolving the
name without deserializing `V1PodStatus` removes the dependency on model
freshness entirely — e.g. list via the Job's label selector
(`batch.kubernetes.io/job-name=<job>`) using a raw/`String`-typed response and
extract `.items[].metadata.name`, or use `GenericKubernetesApi` with a partial
object-metadata Accept header
(`application/json;as=PartialObjectMetadataList;g=meta.k8s.io;v=v1`), which the
API server honours and which never returns a status block at all.

The same approach would let `getQueuedPipelineStatusInternal` report real queued
jobs rather than an empty list on these clusters.

### To Reproduce

1. Deploy 2.0.0 with the native k8s pipeline client on a cluster newer than the
   bundled client-java models (client-java 24.0.0 targets ~1.33; we run 1.36.2).
2. Run any ingestion pipeline — the Job succeeds and the pod logs are fine.
3. Open the run's logs in the UI, or
   `GET /api/v1/services/ingestionPipelines/logs/<service>.<uuid>/last`.

### Expected behavior

Logs being shown in the UI

### OS

_No response_

### Python version

3.12

### OpenMetadata version

2.0.0

_No response_

### Additional context

_No response_

### Pre-submission checklist

### LINKED ISSUE #32062: Kubernetes pipeline client failures are invisible: SystemRepository swallows the exception, and K8sPipelineClient logs a printf format string via slf4j
### Affected module

Backend

### Describe the bug

When the Kubernetes pipeline service client fails, the failure is **invisible in the
server log**. The UI shows "Platform Service Client Unavailable" on Test Connection
and in the Health Check panel, while `grep`ping the server log for the cause returns
nothing at all. Two independent defects combine to produce that.

**1. `SystemRepository.getPipelineServiceClientValidation` swallows the exception.**

`openmetadata-service/src/main/java/org/openmetadata/service/jdbi3/SystemRepository.java`
(L1384–1388 on `main`, L1131–1135 on `1.13.4-release`):

```java
} catch (Exception e) {
  return new StepValidation()
      .withDescription(ValidationStepDescription.PIPELINE_SERVICE_CLIENT.key)
      .withPassed(Boolean.FALSE)
      .withMessage(e.getMessage());
}
```

No `LOG` call, and the stack trace is discarded. Only `e.getMessage()` survives, and
only into the API response. So the operator sees a one-line symptom in the browser
and has no server-side record — no class name, no stack trace, no indication of which
call failed.

This matters more than it looks, because `getServiceStatusInternal` catches only
`ApiException`. Any `RuntimeException` from the Kubernetes client — for example the
`IllegalArgumentException` thrown by `client-java`'s generated model validation when
the cluster is newer than the bundled models (#29572, #29792) — propagates straight
past it into this handler and disappears. That is exactly how it reached us.

**2. `K8sPipelineClient` logs a printf format string through an slf4j logger.**

`openmetadata-service/src/main/java/org/openmetadata/service/clients/pipeline/k8s/K8sPipelineClient.java`
(`main` L186–187 and L308; same on `1.13`):

```java
private static final String NAMESPACE_VALIDATION_WARNING =
    "Could not validate namespace '%s' exists (HTTP %d): %s. Proceeding anyway.";
...
LOG.warn(NAMESPACE_VALIDATION_WARNING, k8sConfig.getNamespace(), e.getCode(), e.getMessage());
```

slf4j substitutes `{}`, not `%s`/`%d`, so the arguments are dropped and the line is
emitted verbatim:

```
WARN o.o.s.c.p.k.K8sPipelineClient - Could not validate namespace '%s' exists (HTTP %d): %s. Proceeding anyway.
```

The namespace, HTTP status and message — the entire diagnostic value of the
warning — are lost. Every other `LOG` call in the class uses `{}`; the sibling
constants here are consumed by `String.format`, which is presumably how the
mismatch crept in.

### To Reproduce

For (1):

1. Run OpenMetadata with `pipelineServiceClientConfig.type: k8s` on a Kubernetes
   version newer than the bundled `io.kubernetes:client-java` models (1.13.4 pins
   25.0.1; we are on 1.36).
2. Open a database service → **Test Connection**, or view the Health Check panel.
3. UI reports "Platform Service Client Unavailable".
4. Search the server log for the cause — nothing is logged.

For (2):

1. Make `validateNamespace`'s API call fail (e.g. a NetworkPolicy denying egress to
   the kube-apiserver, or removing the namespace `get` permission).
2. The warning appears with `%s` / `%d` unexpanded.

### Expected behavior

1. `getPipelineServiceClientValidation` should log the exception with its stack
   trace at ERROR before converting it into a failed `StepValidation`. A validation
   step that fails is precisely the case worth logging.
2. `NAMESPACE_VALIDATION_WARNING` should use `{}` placeholders (or the call site
   should wrap it in `String.format`), so the namespace, status code and message
   are actually emitted.

### OpenMetadata version

1.13.4; both defects verified present on `main` and on branch `1.13` (`bedf0db10`).

### Additional context

These are small fixes, but together they cost real debugging time: the UI says the
platform client is unavailable, the server log is silent, and the one warning that
does fire in this code path carries no information. We spent a while checking RBAC,
`K8S_NAMESPACE` and the ServiceAccount binding — all of which were correct — before
finding the actual cause by other means.

Filed separately from #29572 / #29792 / #32039 because it is not specific to the
`allocatedResources` deserialization failure: any exception from the pipeline service
client is hidden the same way. Fixing (1) would have made that whole class of
problem self-diagnosing.

Happy to open a PR for both if useful.

### Pre-submission checklist

DaniS2001:
### Describe your changes:

Fixes #32062
Fixes #32039

Fixes following suggested approaches and from a long term stability has been made to the log retrival method as well as solved parsing issues over SLF4J

#
### Type of change:
- [X] Bug fix

#
### High-level design:
Updated getLastIngestionLogs so no longer depends on V1Pod schema from Java, we recall metadata from pod in order to get pod name and porceed with log retrival, to keep code consistency and dont drift too much from original implemenation it still attempts to recover container name without going through V1Pod (though this block of code is removable and can simplify a bit the code)
As for SLF4J replaced some %s and %d for {} in the static texts that were not using String.format
#
#### Use cases covered
Deployments in Kubernetes 1.36+ can now check last log from agents
Updated test for GetLastIngestionLogs in for service clients pipeline k8s to reflect new behaviour + added test coverage for new methods

#### Unit tests
| Old Test | New Test / Status |
|----------|-------------------|
| `testGetLastIngestionLogsPrefersMainPodAndReturnsTaskLogs` | Rewritten – uses raw HTTP stubbing instead of `V1Pod` mocks |
| `testGetLastIngestionLogsHandlesEmptyLogsAndApiFailures` | Split into two tests:<br/>`testGetLastIngestionLogsReturnsNoLogsMessageWhenLogsAreEmpty`<br/>`testGetLastIngestionLogsSurfacesLogReadFailureFromKubernetesApi` |
| `testGetLastIngestionLogsHandlesUnsupportedPodFields` | Removed (obsolete) |
| `testGetLastIngestionLogsDoesNotSwallowOtherIllegalArgumentExceptions` | Removed (no longer applicable) |
| (new) | `testGetLastIngestionLogsSurfacesPodMetadataFetchFailure` – covers network/IO errors during pod list fetch |
| (new) | `testGetLastIngestionLogsReturnsNoPodsMessageWhenNoPodsMatch` – covers the case where no pods exist |
| (new) | `testGetLastIngestionLogsToleratesUnknownKubernetes136Fields` – regression test for unknown fields (`allocatedResources`) |
| (new) | `testParsePodSummariesHandlesMalformedJson` – directly exercises `JsonProcessingException` handling |
| (new) | 4 unit tests for `selectContainerName` (all branches) |
| (new) | 4 unit tests for `parsePodSummaries` (happy path, skip unnamed pods, unparsable timestamp, malformed JSON) |
| (new) | 2 unit tests for `selectLatestPod` (prefers main pod, handles empty list) |

#
### Checklist:


harshach (approved):

harshach (dismissed):

openmetadata-service/src/main/java/org/openmetadata/service/clients/pipeline/k8s/K8sPipelineClient.java
```diff
@@ -1056,6 +1064,120 @@ public PipelineServiceClientResponse validateAppRegistration(
     return buildSuccessResponse("Application validated");
   }
 
+  record PodSummary(String name, Map<String, String> labels, OffsetDateTime creationTimestamp) {}
+
+  @VisibleForTesting
+  List<PodSummary> parsePodSummaries(String rawJson) throws IOException {
+    JsonNode items = POD_SUMMARY_MAPPER.readTree(rawJson).path("items");
```
harshach, line 1071: Is "items" key always exists , IMO better to be defensive here to check if that element exists before trying to read from it

pmbrull (approved):

openmetadata-service/src/main/java/org/openmetadata/service/clients/pipeline/k8s/K8sPipelineClient.java
DaniS2001: @harshach in K8s API calls items should always be informed as per PodList schema
if missing code proceeds without taking action as path method returns null
If need it i can add some conditional check and inform user/add some log error or something
code tho wouldn't crash, would proceed like in an empty pod list that already has test for that

If an explicit check is need it i can add through new PR (not sure if adding commit now that PR is in merge queue is fine) or i can follow whatever method is preferred.

Thanks for reviewing the PR and adding some comment/insight, hope my explanation could solve your comment/resolve your doubt.

pmbrull:
picked to 2.0

PR #32543, opened 2026-09-03
Fixes #26600: StarRocks lineage builds unresolvable FQNs

### LINKED ISSUE #26600: StarRocks血缘关系代理无法同步血缘
**Affected module**
StarRocks血缘关系代理无法同步血缘

**Describe the bug**
StarRocks血缘关系代理无法同步血缘，看日志应该是解析到了，但是无法写入OM中

**To Reproduce**

Image

**Expected behavior**

**Version:**
 - Python version:

**Additional context**

Khairajani:
Fixes #26600

The StarRocks audit-log query aliased `db` as both `database_name` and `schema_name`. StarRocks has no database/schema split, so metadata ingestion stores `db` as the OM **schema**, under a single service-level database.

Emitting `db` as the database too made lineage look up `service.db.db.table`, which never matches. The queries parsed fine — the reporter's log shows `Processing graph with 4 nodes and 2 edges` — but neither side resolved, so every edge landed in the temp-table graph with empty FQNs and was dropped before the sink (`OpenMetadata: Processed 0 records`).

Leaving `database_name` NULL wildcards the database in the lookup, matching whatever the metadata workflow ingested with or without a configured `databaseName`. Same approach as MySQL, Oracle, Databricks and Unity Catalog, and where ClickHouse lands via `normalize_table_params_by_service`. The statement is shared with the usage workflow, so both are fixed.

Tests build their audit rows by applying the connector's own SELECT list to a fake audit record, so a regression in the SQL fails the behavioural tests too, not just the string assertion. All five fail on `main`.


ingestion/src/metadata/ingestion/source/database/common_db_source.py
```diff
@@ -99,6 +99,23 @@ class TableNameAndType(BaseModel):
     type_: TableType = TableType.Regular
 
 
+DEFAULT_DATABASE_NAME = "default"
+
+
+def get_service_database_name(service_connection: Any) -> str:
```
mohittilala: @Khairajani does this add any value here? I see it's altering the common file structure and quality settled in this module with class based architecture.

mohittilala: how about a static method if that works here

mohittilala (approved):

ingestion/src/metadata/ingestion/source/database/common_db_source.py
Khairajani: Now sending `NULL` as the database name and let the ES lookup wildcard it. Hence one line change, no shared file, matches MySQL/Oracle/Databricks.

PR #32583, opened 2026-09-03
Fixes #32582: route MSSQL table diffs through the driver that can authenticate domain accounts

### LINKED ISSUE #32582: Table diff fails to authenticate Windows domain accounts (DOMAIN\user) on SQL Server
### Affected module

Ingestion / Data Quality (table diff)

### Describe the bug

A table diff against a SQL Server service configured with a **Windows domain
account** (`DOMAIN\user`) always fails to authenticate:

```
[28000] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]
Login failed for user 'DOMAIN\serviceaccount'. (18456) (SQLDriverConnect)
```

Metadata ingestion against the same service, with the same credentials, works.
Only the table diff fails.

**Root cause: data-diff discards the driver the service already configured.**

A backslash is illegal in a SQL Server *SQL* login name — the server refuses to
create one:

```
Msg 15006  'DOMAIN\serviceaccount' is not a valid name because it contains invalid characters.
```

So `DOMAIN\user` can only exist as a **Windows** login and can only authenticate
over NTLM/Kerberos. Whether that works comes down entirely to which driver makes
the connection:

| Scheme / driver | Result with `DOMAIN\user` | Domain support |
|---|---|---|
| `mssql+pytds` (OM default) | `18456` Login failed for user | ❌ plain SQL auth |
| `mssql+pyodbc` + `ODBC Driver 18` | `18456` Login failed for user | ❌ plain SQL auth |
| `mssql+pymssql` | `18452` login is from an untrusted domain | ✅ NTLM negotiated |
| `mssql+pyodbc` + `driver=FreeTDS` | `18452` login is from an untrusted domain | ✅ NTLM negotiated |

(18452 is the ceiling on a non-domain-joined test server: it proves NTLM was
*negotiated*. Only a domain-joined instance can prove the login completes.)

The MSSQL connector has no domain-auth code — domain-joined SQL Server works
purely because the user can select a scheme backed by **FreeTDS**, which splits
`DOMAIN\user` and negotiates NTLM with the same password. `pymssql` is a FreeTDS
binding; `driver=FreeTDS` selects the FreeTDS ODBC driver. Both are already
installed in the ingestion image (`tdsodbc`, `freetds-bin`).

data-diff has no access to that lever. `MsSQL.__init__` in `collate-data-diff`
(0.11.15, current pin) hard-overwrites the driver:

```python
self._args["driver"] = "{ODBC Driver 18 for SQL Server}"   # overwrites any caller choice

# TODO temp dev debug
self._args["TrustServerCertificate"] = "yes"
```

Every other connection component survives the trip to data-diff — query-string
params reach `pyodbc.connect` intact. `driver` is the single field that does not,
and it is the one that decides whether a domain account can authenticate.

data-diff's MSSQL dialect is otherwise driver-agnostic: a real `hashdiff` run
returns identical results under FreeTDS and msodbcsql18, so FreeTDS is a drop-in.

**Note:** the `TrustServerCertificate = "yes"` above is also hard-coded and
silently disables certificate validation for every MSSQL diff. It cannot simply be
deleted — setups with self-signed certificates now depend on it — so it needs to
become config-driven. Tracking here; may split into its own issue.

### To Reproduce

1. Register an MSSQL service using a Windows domain account (`DOMAIN\user`) with
   scheme `mssql+pymssql` (or `mssql+pyodbc` with `driver=FreeTDS`), against a
   domain-joined SQL Server.
2. Confirm metadata ingestion succeeds.
3. Create a table diff test case on any table in that service and run it.
4. The test fails with `18456 Login failed for user 'DOMAIN\user'`.

A local sandbox reproducing the client-side failure without a domain controller
(Docker SQL Server + scripted repro of every row in the table above) is available
on request.

**Reading the logs:** the username looks double-escaped in ingestion logs
(`DOMAIN\\\\user`). It is not — the line carries two escaping layers, `repr()` of
the pyodbc `(sqlstate, message)` args tuple plus the JSON log transport, so one
real backslash renders as four. Four backslashes is the *correct* rendering of
one; eight would indicate genuine doubling.

### Expected behavior

A table diff on a service whose metadata ingestion authenticates successfully
should authenticate too. Concretely, data-diff should connect with the ODBC
driver matching the auth capability of the configured scheme:

| Service scheme | `driver` field | data-diff should use |
|---|---|---|
| `mssql+pyodbc` | user-set, meaningful | that value |
| `mssql+pymssql` | default, ignored | `FreeTDS` (pymssql *is* FreeTDS) |
| `mssql+pytds` | default, ignored | `ODBC Driver 18` (unchanged — pytds is SQL-auth-only) |

Note this must be **derived**, not a passthrough of the `driver` field. `driver`
is documented as pyodbc-only, so on a `mssql+pymssql` service it sits at its
schema default `"ODBC Driver 18 for SQL Server"` — forwarding it verbatim is a
no-op for exactly the configuration that is broken.

### Proposed fix

Two coordinated changes:

1. **`collate-data-diff`** — stop `MsSQL.__init__` overwriting a caller-supplied
   driver:
   ```python
   -self._args["driver"] = "{ODBC Driver 18 for SQL Server}"
   +self._args["driver"] = "{%s}" % kw.pop("odbc_driver", "ODBC Driver 18 for SQL Server")
   ```
   A separate key is required: in a connection dict, `driver` is already consumed
   by `connect_with_dict` as the *scheme*. This supports both entry points, since
   query params already reach `**kw`. Needs a release and a pin bump.

2. **OpenMetadata** — add `MssqlConnection.get_connection_dict()` carrying the
   derived `odbc_driver`, mirroring the existing `AzureSQLConnection` pattern,
   plus an `MssqlTableParameter` to stamp the per-table database/schema. This also
   removes the URI encode/decode fragility that #31124 / #31134 patched, since the
   dict path bypasses URI parsing entirely.

### OpenMetadata version

1.13.4

### Additional context

Related: #31124 / #31134 fixed percent-encoding of special characters in the
data-diff username. That fix is correct and unaffected — it was masking this
failure, which is why the symptom surfaced only after it landed. Before it, the
same log read `Login failed for user 'DOMAIN%5Cuser'`.

TeddyCr:
### Describe your changes:

Fixes #32582

A table diff against a SQL Server service using a **Windows domain account**
(`DOMAIN\user`) always failed to authenticate, while metadata ingestion against
the same service with the same credentials succeeded.

A backslash is illegal in a SQL Server *SQL* login name (`Msg 15006`), so
`DOMAIN\user` can only exist as a Windows login and only authenticates over NTLM.
Which driver makes the connection decides whether that works:

| Scheme / driver | Result with `DOMAIN\user` | Domain support |
|---|---|---|
| `mssql+pytds` (OM default) | `18456` Login failed for user | ❌ plain SQL auth |
| `mssql+pyodbc` + `ODBC Driver 18` | `18456` Login failed for user | ❌ plain SQL auth |
| `mssql+pymssql` | `18452` login is from an untrusted domain | ✅ NTLM negotiated |
| `mssql+pyodbc` + `driver=FreeTDS` | `18452` login is from an untrusted domain | ✅ NTLM negotiated |

The MSSQL connector has no domain-auth code — domain-joined SQL Server works
purely because the user can pick a **FreeTDS-backed** scheme (pymssql *is* a
FreeTDS binding). data-diff is pyodbc-only and hard-codes msodbcsql18, so the
diff had no access to that lever.

I gave `MssqlConnection` a `get_connection_dict()` carrying an ODBC driver
**derived from the scheme**, mirroring the existing `AzureSQLConnection` pattern.

#
### Type of change:
- [x] Bug fix

#
### High-level design:

**Why derive rather than forward `driver`.** `driver` is documented as
pyodbc-only and defaults to `"ODBC Driver 18 for SQL Server"`, so on a
`mssql+pymssql` service it sits at that default and is never set by the user.
Forwarding it verbatim would be a no-op for exactly the configuration that is
broken. The mapping instead answers *"which ODBC driver has the same auth
capability as the scheme metadata ingestion uses?"*:

| Service scheme | `driver` field | data-diff uses |
|---|---|---|
| `mssql+pyodbc` | user-set, meaningful | that value |
| `mssql+pymssql` | default, ignored | `FreeTDS` |
| `mssql+pytds` | default, ignored | `ODBC Driver 18` (unchanged) |

pytds deliberately stays on msodbcsql: it is SQL-auth-only, so routing it to
FreeTDS would let the diff authenticate more than metadata ingestion can.

**Why a connection dict.** It bypasses URI parsing entirely, so credentials reach
data-diff verbatim and usernames holding reserved characters need no
encode/decode round trip — retiring the fragility patched in #31124 / #31134. It
is also the only channel with room for the driver, since a URL's `driver` query
param is overwritten downstream.

**Backward compatibility.** Adding `get_connection_dict()` routes *all* MSSQL
diffs through the dict path, which would silently drop `connectionOptions` —
they used to ride along as URL query params and reach `pyodbc.connect`. They are
now carried explicitly, with derived values taking precedence.

**Rollout.** This is inert until `collate-data-diff` honors `odbc_driver`. ODBC
ignores unrecognised connection-string attributes (verified against both drivers),
so it is safe to merge ahead of that release and activates on the pin bump. The
companion fork change is written and tested:
```python
-self._args["driver"] = "{ODBC Driver 18 for SQL Server}"
+odbc_driver = kw.pop("odbc_driver", None) or DEFAULT_ODBC_DRIVER
+self._args["driver"] = "{%s}" % odbc_driver
```
A separate keyword is required because in a connection dict `driver` is already
consumed by `connect_with_dict` as the *scheme*.

**Alternative rejected.** Kerberos (`Trusted_Connection=yes` + msodbcsql) also
supports domain accounts but needs a ticket cache in the ingestion pod. FreeTDS
NTLM works with the username and password the service already holds, and FreeTDS
is already installed in the ingestion image (`tdsodbc`, `freetds-bin`).

#
#### Use cases covered
- MSSQL service on `mssql+pymssql` with a domain account produces a diff
  connection using FreeTDS, so the account authenticates as it does for ingestion
- MSSQL service on `mssql+pyodbc` keeps whatever driver the user configured
- MSSQL service on `mssql+pytds` is unchanged
- A username containing a backslash reaches data-diff verbatim, unencoded
- `connectionOptions` still apply to table diffs
- A diff spanning databases/schemas gets the per-table path, not the service-level one

#### Unit tests
- Files added: `ingestion/tests/unit/observability/data_quality/validations/runtime_param_setter/test_mssql_diff_params_setter.py` (12 tests)
- Full `tests/unit/observability/data_quality/` suite: **359 passed**
- `ruff check` / `ruff format` clean; `basedpyright` 0 errors on changed files

Each test was trap-verified rather than only observed green: replacing the
derivation with the naive `connection.driver` passthrough fails exactly the three
pymssql tests, and removing the `connectionOptions` merge fails exactly that test.

- [x] Not applicable — covered by unit tests; the end-to-end path needs a
  domain-joined SQL Server, which CI has no access to (see manual testing).

#### Playwright (UI) tests

#### Manual testing performed

Against SQL Server 2022 in Docker, with both `msodbcsql18` and FreeTDS installed:

1. Confirmed `CREATE LOGIN [DOMAIN\user] WITH PASSWORD` fails with **Msg 15006**,
   establishing that such an account can only be a Windows login.
2. Connected with `DOMAIN\user` under each scheme and recorded the error numbers
   in the table above — **18456** (SQL auth) vs **18452** (NTLM negotiated).
3. Ran the real OM → data-diff path and confirmed the payload for a pymssql
   service: `odbc_driver: 'FreeTDS'`, `user: 'nb\svcaccount'` verbatim, `schema`
   stamped from the table FQN.
4. Ran a real `hashdiff` through the patched data-diff under **both** drivers —
   identical results (same differing row, same key), confirming FreeTDS is a
   drop-in for this dialect.
5. Confirmed an unknown `odbc_driver` attribute is tolerated by both drivers,
   which is what makes this safe to merge before the fork release.

**Not verified:** that a domain login *completes*. `18452` means the server
rejected an untrusted domain — it proves NTLM was negotiated, but only a
domain-joined instance with the real Windows login can prove the rest.

#

Not applicable.

#
### Checklist:
- [x] For JSON Schema changes: not needed — no schema change; the existing
      `driver` field is reused.

Bug fix
- [x] I have added a test that covers the exact scenario we are fixing
      (`TestOdbcDriverDerivation::test_pymssql_maps_to_freetds`, with issue #32582
      referenced in the module docstring).

### DIFF
--- a/ingestion/src/metadata/ingestion/source/database/mssql/connection.py
+++ b/ingestion/src/metadata/ingestion/source/database/mssql/connection.py
@@ -35,9 +35,13 @@ from metadata.core.connections.test_connection.network import NETWORK_ERRORS
 from metadata.generated.schema.entity.services.connections.database.mssqlConnection import (
     MssqlConnection as MssqlConnectionConfig,
 )
+from metadata.generated.schema.entity.services.connections.database.mssqlConnection import (
+    MssqlScheme,
+)
 from metadata.ingestion.connections.builders import (
     create_generic_db_connection,
     get_connection_args_common,
+    get_connection_options_dict,
     get_connection_url_common,
 )
 from metadata.ingestion.connections.connection import BaseConnection
@@ -57,6 +61,34 @@ if TYPE_CHECKING:
     from metadata.core.connections.test_connection import ChecksProvider
     from metadata.core.connections.test_connection.classifier import Matcher
     from metadata.core.connections.test_connection.records import Evidence
+
+
+DEFAULT_SQL_SERVER_PORT = 1433
+DEFAULT_ODBC_DRIVER = "ODBC Driver 18 for SQL Server"
+FREETDS_ODBC_DRIVER = "FreeTDS"
+
+
+def _odbc_driver_for_data_diff(connection: MssqlConnectionConfig) -> str:
+    """The ODBC driver whose auth capability matches `connection.scheme`.
+
+    data-diff is pyodbc-only, so a non-ODBC scheme still has to resolve to an ODBC
+    driver. Only FreeTDS splits a `DOMAIN\\user` login and negotiates NTLM;
+    msodbcsql offers it as a SQL login name, which SQL Server rejects with 18456
+    (a backslash is illegal in a SQL login, so the account can only be a Windows
+    one). pymssql is itself a FreeTDS binding, hence the mapping. pytds is
+    SQL-auth-only, so msodbcsql matches it exactly - routing it to FreeTDS would
+    let the diff authenticate more than metadata ingestion can. See issue #32582.
+
+    `connection.driver` is read only under pyodbc: it is documented as pyodbc-only
+    and otherwise sits at its schema default, so trusting it elsewhere would be a
+    no-op for exactly the configuration this resolves.
+    """
+    scheme = connection.scheme or MssqlScheme.mssql_pytds
+    if scheme.value == MssqlScheme.mssql_pyodbc.value:
+        return connection.driver or DEFAULT_ODBC_DRIVER
+    if scheme.value == MssqlScheme.mssql_pymssql.value:
+        return FREETDS_ODBC_DRIVER
+    return DEFAULT_ODBC_DRIVER
 
 
 def _mssql_number(error: BaseException) -> int | None:
@@ -199,6 +231,32 @@ class MssqlConnection(BaseConnection[MssqlConnectionConfig, Engine]):
         )
         self._on_close(engine.dispose)
         return engine
+
+    def get_connection_dict(self) -> dict:
+        """Return the connection parameters for data-diff.
+
+        Preferred over a rendered SQLAlchemy URL because it bypasses URI parsing
+        entirely: credentials reach data-diff verbatim, so usernames holding
+        reserved characters need no encode/decode round trip (see #31124/#31134),
+        and `odbc_driver` can carry the driver the URL has no room for.
+        """
+        connection = self.service_connection
+        host, _, port = (connection.hostPort or "").partition(":")
+
+        return {
+            # connectionOptions used to ride along as query params on the rendered
+            # URL and land in pyodbc's kwargs. This dict replaces that URL, so it
+            # has to carry them or extra ODBC keywords stop applying to diffs. The
+            # derived values below win: they are what makes domain auth work.
+            **(get_connection_options_dict(connection) or {}),
+            "driver": (connection.scheme or MssqlScheme.mssql_pytds).value,
+            "host": host,
+            "port": int(port) if port else DEFAULT_SQL_SERVER_PORT,
+            "user": connection.username,
+            "password": connection.password.get_secret_value() if connection.password else None,
+            "database": connection.database,
+            "odbc_driver": _odbc_driver_for_data_diff(connection),
+        }
 
     def _get_databases_statement(self) -> str:
         if self.service_connection.ingestAllDatabases:
--- a/ingestion/src/metadata/ingestion/source/database/mssql/connection.py
+++ b/ingestion/src/metadata/ingestion/source/database/mssql/connection.py
@@ -47,6 +47,9 @@ from metadata.ingestion.connections.builders import (
 from metadata.ingestion.connections.connection import BaseConnection
 from metadata.ingestion.connections.test_connections import SourceConnectionException
 from metadata.ingestion.source.database.azuresql.connection import (
+    DEFAULT_SQL_SERVER_PORT,
+)
+from metadata.ingestion.source.database.azuresql.connection import (
     get_connection_url as get_pyodbc_connection_url,
 )
 from metadata.ingestion.source.database.mssql.queries import (
@@ -64,7 +67,6 @@ if TYPE_CHECKING:
     from metadata.core.connections.test_connection.records import Evidence
 
 
-DEFAULT_SQL_SERVER_PORT = 1433
 DEFAULT_ODBC_DRIVER = "ODBC Driver 18 for SQL Server"
 FREETDS_ODBC_DRIVER = "FreeTDS"
 
@@ -256,6 +258,17 @@ class MssqlConnection(BaseConnection[MssqlConnectionConfig, Engine]):
             )
 
         host, _, port = connection.hostPort.partition(":")
+        if port and not port.isdigit():
+            # Same reasoning as above, from the other side: `int(port)` would raise
+            # ValueError, which that fallback catches, so a typo'd port would drop
+            # the diff back onto the URL path and take the derived ODBC driver with
+            # it. Quietly defaulting to 1433 is no better - it connects somewhere
+            # the user did not ask for.
+            raise SourceConnectionException(
+                f"MSSQL hostPort {connection.hostPort!r} has a non-numeric port, so the table "
+                "diff cannot connect. Set 'Host and Port' to 'host:port', or to 'host' alone "
+                f"to use the default {DEFAULT_SQL_SERVER_PORT}."
+            )
 
         return {
             # connectionOptions used to ride along as query params on the rendered
--- a/ingestion/src/metadata/ingestion/source/database/mssql/connection.py
+++ b/ingestion/src/metadata/ingestion/source/database/mssql/connection.py
@@ -45,6 +45,7 @@ from metadata.ingestion.connections.builders import (
     get_connection_url_common,
 )
 from metadata.ingestion.connections.connection import BaseConnection
+from metadata.ingestion.connections.test_connections import SourceConnectionException
 from metadata.ingestion.source.database.azuresql.connection import (
     get_connection_url as get_pyodbc_connection_url,
 )
@@ -241,7 +242,20 @@ class MssqlConnection(BaseConnection[MssqlConnectionConfig, Engine]):
         and `odbc_driver` can carry the driver the URL has no room for.
         """
         connection = self.service_connection
-        host, _, port = (connection.hostPort or "").partition(":")
+        if not connection.hostPort:
+            # `hostPort` is optional in the schema. Defaulting the host to "" would be
+            # worse than refusing: ODBC reads a blank server as the local machine, so
+            # the diff would quietly connect somewhere unintended. Falling back to a
+            # rendered URL cannot work either - it raises a bare TypeError - and this
+            # must stay outside (ValueError, AttributeError, NotImplementedError), or
+            # BaseTableParameter._get_service_connection_config downgrades it to that
+            # fallback and swallows the message.
+            raise SourceConnectionException(
+                "MSSQL connection has no hostPort configured, so the table diff has nothing "
+                "to connect to. Set 'Host and Port' on the service connection."
+            )
+
+        host, _, port = connection.hostPort.partition(":")
 
         return {
             # connectionOptions used to ride along as query params on the rendered
--- a/ingestion/tests/unit/observability/data_quality/validations/runtime_param_setter/test_mssql_diff_params_setter.py
+++ b/ingestion/tests/unit/observability/data_quality/validations/runtime_param_setter/test_mssql_diff_params_setter.py
@@ -1,0 +1,215 @@
+#  Copyright 2025 Collate
+#  Licensed under the Collate Community License, Version 1.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#  https://github.com/open-metadata/OpenMetadata/blob/main/ingestion/LICENSE
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""MSSQL table-diff connection parameters.
+
+data-diff is pyodbc-only, so every MSSQL service - whatever SQLAlchemy scheme it
+uses for metadata ingestion - has to resolve to an ODBC driver. Only FreeTDS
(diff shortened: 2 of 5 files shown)


ingestion/src/metadata/ingestion/source/database/mssql/connection.py
```diff
@@ -200,6 +232,32 @@ def _get_client(self) -> Engine:
...
+        """Return the connection parameters for data-diff.
+
+        Preferred over a rendered SQLAlchemy URL because it bypasses URI parsing
+        entirely: credentials reach data-diff verbatim, so usernames holding
+        reserved characters need no encode/decode round trip (see #31124/#31134),
+        and `odbc_driver` can carry the driver the URL has no room for.
+        """
+        connection = self.service_connection
+        host, _, port = (connection.hostPort or "").partition(":")
+
```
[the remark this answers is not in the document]
TeddyCr: Good catch — that `or ""` was there to satisfy basedpyright, not by design. Fixed in c42912f8603.

Two things I confirmed while acting on it:

1. **The empty host is worse than I'd assumed.** `hostPort` really is optional in the schema (`required: ['database']`), and ODBC reads a blank server as the local machine — so the diff could connect somewhere unintended rather than fail.

2. **The URL fallback can't rescue it, as you suspected.** I checked: it raises a bare `TypeError: can only concatenate str (not "NoneType") to str` from `get_connection_url_common`. No more actionable than an empty host.

One wrinkle worth flagging: it raises `SourceConnectionException`, deliberately **not** `ValueError`. `BaseTableParameter._get_service_connection_config` catches `(ValueError, AttributeError, NotImplementedError)` and downgrades to the URL fallback — so a `ValueError` here would be silently swallowed, losing both the message and the derived ODBC driver. There's a test pinning the exception type outside that tuple so it can't drift back.

ingestion/tests/unit/observability/data_quality/validations/runtime_param_setter/test_mssql_diff_params_setter.py
```diff
@@ -0,0 +1,240 @@
...
+            MssqlConnection(config).get_connection_dict()
+
+    def test_missing_host_port_is_not_swallowed_into_a_url_fallback(self):
+        """BaseTableParameter._get_service_connection_config catches ValueError.
+
+        Raising one here would be silently downgraded to the URL path, losing both
+        this message and the derived driver, so the error must sit outside that
+        tuple.
+        """
+        assert not issubclass(SourceConnectionException, ValueError | AttributeError | NotImplementedError)
```
Khairajani: This asserts the exception hierarchy, not the behaviour, so it's green whatever the code does. Assert `_get_service_connection_config` raises on `hostPort=None` instead?

ingestion/src/metadata/ingestion/source/database/mssql/connection.py
```diff
@@ -59,6 +64,34 @@
     from metadata.core.connections.test_connection.records import Evidence
 
 
+DEFAULT_SQL_SERVER_PORT = 1433
```
Khairajani: `azuresql/connection.py` already has this one, and we import from there a few lines up — reuse it instead of a second 1433?

ingestion/src/metadata/ingestion/source/database/mssql/connection.py
```diff
@@ -200,6 +233,45 @@ def _get_client(self) -> Engine:
...
+
+        return {
+            # connectionOptions used to ride along as query params on the rendered
+            # URL and land in pyodbc's kwargs. This dict replaces that URL, so it
+            # has to carry them or extra ODBC keywords stop applying to diffs. The
+            # derived values below win: they are what makes domain auth work.
+            **(get_connection_options_dict(connection) or {}),
+            "driver": (connection.scheme or MssqlScheme.mssql_pytds).value,
+            "host": host,
+            "port": int(port) if port else DEFAULT_SQL_SERVER_PORT,
```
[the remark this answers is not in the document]
TeddyCr: Fixed in beefd438b94 — it now raises `SourceConnectionException` on a non-numeric port rather than letting `int()` throw.

Two things I confirmed while acting on it:

1. **The URL fallback can't rescue a bad port either.** I expected it to at least connect; it doesn't — SQLAlchemy's `make_url` hits the identical `int()` on the port segment and raises the same `ValueError` from `engine/url.py:917`. So the swallow buys nothing and costs the message.
2. **Raising beats defaulting.** The [companion suggestion](https://github.com/open-metadata/OpenMetadata/pull/32583#discussion_r3928869030) was `int(port) if port.isdigit() else DEFAULT_SQL_SERVER_PORT`, which keeps the dict path alive but silently connects to 1433 — a port the user did not configure. Given a typo'd port is always a misconfiguration and never a named instance (`host\INSTANCE` has no colon, so it already falls through to the default), naming the field is the more actionable outcome.

Trap-verified rather than only observed green: removing the guard fails exactly the two new port tests, and swapping the raise to `ValueError` fails all four error-path tests — including the behavioural one that now drives `_get_service_connection_config` end to end.

ingestion/tests/unit/observability/data_quality/validations/runtime_param_setter/test_mssql_diff_params_setter.py
TeddyCr: Agreed — that one held whatever the code did, so it would have gone green through the exact regression it was meant to pin. Replaced in beefd438b94 with a behavioural test that drives `_get_service_connection_config`, parametrized over both bad inputs:

```python
@pytest.mark.parametrize("host_port", [None, "sqlserver.corp.local:not-a-port"])
def test_bad_host_port_is_not_swallowed_into_a_url_fallback(self, host_port):
    config = _config(MssqlScheme.mssql_pymssql)
    config.hostPort = host_port

    with pytest.raises(SourceConnectionException):
        BaseTableParameter._get_service_connection_config(config)
```

It earns its place now: swapping the raise to `ValueError` makes both parameters fail, where the old assertion still passed — the class hierarchy is unchanged by that edit, which is the whole problem with asserting on it.

ingestion/src/metadata/ingestion/source/database/mssql/connection.py
TeddyCr: Right, and we import `get_connection_url` from that module three lines up — no reason for a second one. Removed in beefd438b94; it now imports `DEFAULT_SQL_SERVER_PORT` from `azuresql.connection` alongside the existing import.

ingestion/src/metadata/ingestion/source/database/mssql/connection.py
```diff
@@ -200,6 +232,32 @@ def _get_client(self) -> Engine:
...
+    def get_connection_dict(self) -> dict:
+        """Return the connection parameters for data-diff.
+
+        Preferred over a rendered SQLAlchemy URL because it bypasses URI parsing
+        entirely: credentials reach data-diff verbatim, so usernames holding
+        reserved characters need no encode/decode round trip (see #31124/#31134),
+        and `odbc_driver` can carry the driver the URL has no room for.
+        """
+        connection = self.service_connection
+        host, _, port = (connection.hostPort or "").partition(":")
```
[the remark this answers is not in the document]
TeddyCr: Real issue — fixed in beefd438b94, though by raising rather than defaulting.

Defaulting keeps the dict path alive, but it connects to 1433 on a service the user configured for some other port, which trades a loud failure for a wrong connection. The named-instance case doesn't need it either: `host\INSTANCE` has no colon, so `partition(":")` already leaves the port empty and the existing `else DEFAULT_SQL_SERVER_PORT` handles it. That leaves only genuine typos, which are worth naming.

So it now raises `SourceConnectionException` — deliberately outside the `(ValueError, AttributeError, NotImplementedError)` tuple `_get_service_connection_config` catches, so the message survives instead of being downgraded to the URL path. Worth noting that fallback couldn't have parsed the port either: SQLAlchemy's `make_url` raises the same `ValueError` at `engine/url.py:917`.

Khairajani (approved):

